PR #19948, opened 2025-11-10
Fix node bootstrap error when enable stream transport and remote cluster state

bowenlan-amzn:
### Description

This PR fixes a bootstrap failure that occurs when streaming transport is used with remote cluster state. Without this fix, nodes fail to start with the error:
```
can't overwrite as repositories are already present
	at org.opensearch.repositories.RepositoriesService.updateRepositoriesMap(RepositoriesService.java:885)
	at org.opensearch.node.remotestore.RemoteStoreNodeService.createAndVerifyRepositories(RemoteStoreNodeService.java:163)
	at org.opensearch.node.Node$LocalNodeFactory.apply(Node.java:2355)
	at org.opensearch.node.Node$LocalNodeFactory.apply(Node.java:2323)
	at org.opensearch.transport.TransportService.doStart(TransportService.java:402)
	at org.opensearch.common.lifecycle.AbstractLifecycleComponent.start(AbstractLifecycleComponent.java:77)
	at org.opensearch.node.Node.start(Node.java:1850)
```

### Root Cause

When streaming transport is enabled, the node bootstrap process creates **two separate `LocalNodeFactory` instances**:
1. One for the regular `TransportService`
2. Another for the `StreamTransportService` 

During node startup, both transport services are started sequentially:
```java
if (streamTransportService != null) {
    streamTransportService.start();  // First call to LocalNodeFactory.apply()
}
transportService.start();  // Second call to LocalNodeFactory.apply()
```

Both services inherit `TransportService.doStart()` which calls:
```java
localNode = localNodeFactory.apply(transport.boundAddress());
```

Each call to `LocalNodeFactory.apply()` triggers:
1. `DiscoveryNode` creation
2. **Remote store repository creation and verification** via `remoteStoreNodeService.createAndVerifyRepositories()`

Since both factories attempt to register the same repositories (configured via node attributes), the second call fails with "can't overwrite as repositories are already present".

### Solution

We don't do Remote store repository creation and verification when it's for stream transport

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/node/Node.java
+++ b/server/src/main/java/org/opensearch/node/Node.java
@@ -1292,7 +1292,7 @@ public class Node implements Closeable {
                         streamTransport,
                         threadPool,
                         networkModule.getTransportInterceptor(),
-                        new LocalNodeFactory(settings, nodeEnvironment.nodeId(), remoteStoreNodeService),
+                        new LocalNodeFactory(settings, nodeEnvironment.nodeId()),
                         settingsModule.getClusterSettings(),
                         transportService.getTaskManager(),
                         transportService.getRemoteClusterService(),
@@ -2327,12 +2327,10 @@ public class Node implements Closeable {
         private final SetOnce<DiscoveryNode> localNode = new SetOnce<>();
         private final String persistentNodeId;
         protected final Settings settings;
-        protected final RemoteStoreNodeService remoteStoreNodeService;
 
-        private LocalNodeFactory(Settings settings, String persistentNodeId, RemoteStoreNodeService remoteStoreNodeService) {
+        private LocalNodeFactory(Settings settings, String persistentNodeId) {
             this.persistentNodeId = persistentNodeId;
             this.settings = settings;
-            this.remoteStoreNodeService = remoteStoreNodeService;
         }
 
         @Override
@@ -2357,12 +2355,15 @@ public class Node implements Closeable {
      */
     private static class RemoteStoreVerifyingLocalNodeFactory extends LocalNodeFactory {
 
+        private final RemoteStoreNodeService remoteStoreNodeService;
+
         private RemoteStoreVerifyingLocalNodeFactory(
             Settings settings,
             String persistentNodeId,
             RemoteStoreNodeService remoteStoreNodeService
         ) {
-            super(settings, persistentNodeId, remoteStoreNodeService);
+            super(settings, persistentNodeId);
+            this.remoteStoreNodeService = remoteStoreNodeService;
         }
 
         @Override


server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -2325,11 +2325,22 @@ private static class LocalNodeFactory implements Function<BoundTransportAddress,
...
 
         private LocalNodeFactory(Settings settings, String persistentNodeId, RemoteStoreNodeService remoteStoreNodeService) {
+            this(settings, persistentNodeId, remoteStoreNodeService, false);
+        }
+
+        private LocalNodeFactory(
+            Settings settings,
+            String persistentNodeId,
+            RemoteStoreNodeService remoteStoreNodeService,
+            boolean useStreamTransport
```
andrross: The weird thing here is that "useStreamTransport" doesn't make any sense because there's nothing about stream transport in the LocalNodeFactory. I think you can use composition to make this more clear, if a bit more verbose. You can refactor LocalNodeFactory into two separate functions. First, remove RemoteStoreNodeService from this implementation, then create one that looks like:

```java
private static class RemoteStoreVerifyingLocalNodeFactory extends LocalNodeFactory {
    private final RemoteStoreNodeService remoteStoreNodeService;
    
    private RemoteStoreVerifyingLocalNodeFactory(Settings settings, String persistentNodeId, RemoteStoreNodeService remoteStoreNodeService) {
        super(settings, persistentNodeId);
        this.remoteStoreNodeService = remoteStoreNodeService;
    }

    @Override
    public DiscoveryNode apply(BoundTransportAddress boundTransportAddress) {
        final DiscoveryNode discoveryNode = super.apply(boundTransportAddress);
        if (isRemoteStoreAttributePresent(settings)) {
            remoteStoreNodeService.createAndVerifyRepositories(discoveryNode);
        }
        return discoveryNode;
    }
}
```

In the normal case you construct a `RemoteStoreVerifyingLocalNodeFactory` instance, and in the stream transport case you just construct a plain `LocalNodeFactory` instance.

bowenlan-amzn: Nice point! Better readability would help here

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -2320,11 +2320,14 @@ protected List<AuxTransport> newAuxTransports(NetworkModule networkModule) {
...
+    /**
+     * Base factory for creating DiscoveryNode instances during node initialization.
+     */
     private static class LocalNodeFactory implements Function<BoundTransportAddress, DiscoveryNode> {
         private final SetOnce<DiscoveryNode> localNode = new SetOnce<>();
         private final String persistentNodeId;
-        private final Settings settings;
-        private final RemoteStoreNodeService remoteStoreNodeService;
+        protected final Settings settings;
+        protected final RemoteStoreNodeService remoteStoreNodeService;
```
andrross: Nitpick, but this field can move to `RemoteStoreVerifyingLocalNodeFactory` since it's not used here, right?

bowenlan-amzn: Yep, removed.

Issue #11749, opened 2024-01-04
[Feature Request] allow field names with dots in the name


rursprung:
### Is your feature request related to a problem? Please describe

sometimes we need to index documents where the field names contain dots but shouldn't be turned into subobjects (e.g. `foo.bar` should stay `foo.bar` and not become a field `foo` containing a nested field `bar`).
this is esp. relevant when there's a second field which contains part of the path of the other (e.g. here `foo`) which then wouldn't work.

### Describe the solution you'd like

it must be possible to define that some fields should not be evaluated as subobjects.

### Related component

Indexing

### Describe alternatives you've considered

one workaround is to append a custom suffix to the fields which should (hopefully) not be part of the path of others, e.g. instead of `foo` and `foo.bar` one could use `foo_x` and `foo.bar_x` (thus avoiding the name collision with `foo`).

### Additional context

Elasticsearch introduced the possibility to define `"subobjects": false` in ES 8.x

msfroh:
This should be doable. 

I believe we need to add a new property to [`ObjectMapper`](https://github.com/opensearch-project/OpenSearch/blob/09bacee5fc85676e97bee6b4ad87dec35c6aa8cc/server/src/main/java/org/opensearch/index/mapper/ObjectMapper.java), which will be inherited by [`RootObjectMapper`](https://github.com/opensearch-project/OpenSearch/blob/09bacee5fc85676e97bee6b4ad87dec35c6aa8cc/server/src/main/java/org/opensearch/index/mapper/RootObjectMapper.java). Based on that property, we can how we resolve sub-mappers.

(I could be wrong about the above, but that's where I would start looking to work on this.)

stowns:
We have the same request though it's for a different reason. We are trying to index json objects which have potentially large nested values. `flat_object` was not working for us because some of the nested values exceeded 32kb which throws an error in lucene as `flat_object`'s nested fields get indexed as keywords. As a work-around I tried to flatted all objects before indexing and have each nested field treated as text with `dynamic_templates` ie)
```
"dynamic_templates": [
  {
    "context": {
      "path_match": "context.*",
      "mapping": {
        "type": "text"
      }
    }
  },
  {
    "message": {
      "path_match": "message.*",
      "mapping": {
        "type": "text"
      }
    }
  }
]
```

However, now I'm experiencing
```
class org.opensearch.index.mapper.TextFieldMapper cannot be cast to class org.opensearch.index.mapper.ObjectMapper
```
[...]

PR #19958, opened 2025-11-11
Support for fields containing dots in their name as literals

mohit10011999:
### Description
Add the `disable_objects` field with default value of false to ObjectMapper
  - Implement builder methods for setting disable_objects
  - Implement getter methods (disableObjects() and disableObjectsExplicit())
  - Add serialization/deserialization support for disable_objects parameter

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/11749
RFC - https://github.com/opensearch-project/OpenSearch/issues/20125

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added public getter method in `SourceFieldMapper` to return included field ([#20290](https://github.com/opensearch-project/OpenSearch/pull/20290))
 - Support for HTTP/3 (server side) ([#20017](https://github.com/opensearch-project/OpenSearch/pull/20017))
 - Add circuit breaker support for gRPC transport to prevent out-of-memory errors ([#20203](https://github.com/opensearch-project/OpenSearch/pull/20203))
+- Add index-level-encryption support for snapshots and remote-store ([#20095](https://github.com/opensearch-project/OpenSearch/pull/20095))
 
 ### Changed
 - Handle custom metadata files in subdirectory-store ([#20157](https://github.com/opensearch-project/OpenSearch/pull/20157))
@@ -34,6 +35,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fix X-Opaque-Id header propagation (along with other response headers) for streaming Reactor Netty 4 transport ([#20371](https://github.com/opensearch-project/OpenSearch/pull/20371))
 - Allow removing plugin that's optionally extended ([#20417](https://github.com/opensearch-project/OpenSearch/pull/20417))
 - Fix indexing regression and bug fixes for grouping criteria. ([20145](https://github.com/opensearch-project/OpenSearch/pull/20145))
+- LeafReader should not remove SubReaderWrappers incase IndexWriter encounters a non aborting Exception ([#20193](https://github.com/opensearch-project/OpenSearch/pull/20193))
+- Fix Netty deprecation warnings in transport-reactor-netty4 module ([20429](https://github.com/opensearch-project/OpenSearch/pull/20429))
 
 ### Dependencies
 - Bump `com.google.auth:google-auth-library-oauth2-http` from 1.38.0 to 1.41.0 ([#20183](https://github.com/opensearch-project/OpenSearch/pull/20183))
@@ -62,6 +65,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Bump `org.jline:jline` from 3.30.5 to 3.30.6 ([#20369](https://github.com/opensearch-project/OpenSearch/pull/20369))
 - Bump `lycheeverse/lychee-action` from 2.6.1 to 2.7.0 ([#20370](https://github.com/opensearch-project/OpenSearch/pull/20370))
 - Bump `opensearch-protobufs` from 1.0.0 to 1.1.0 and update transport-grpc module compatibility ([#20396](https://github.com/opensearch-project/OpenSearch/pull/20396))
+- Bump `ch.qos.logback:logback-classic` from 1.5.23 to 1.5.24 ([#20407](https://github.com/opensearch-project/OpenSearch/pull/20407))
 - Bump `org.wiremock:wiremock-standalone` from 3.6.0 to 3.13.2 ([#20406](https://github.com/opensearch-project/OpenSearch/pull/20406))
 - Bump `tj-actions/changed-files` from 47.0.0 to 47.0.1 ([#20408](https://github.com/opensearch-project/OpenSearch/pull/20408))
 - Bump `reactor-netty` to 1.3.2, `reactor` to 3.8.2 ([#20419](https://github.com/opensearch-project/OpenSearch/pull/20419))
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -20,6 +20,7 @@ This document contains a list of maintainers in this repo. See [opensearch-proje
 | Binlong Gao       | [gaobinlong](https://github.com/gaobinlong)             | Amazon      |
 | Gaurav Bafna      | [gbbafna](https://github.com/gbbafna)                   | Amazon      |
 | Jay Deng          | [jed326](https://github.com/jed326)                     | Amazon      |
+| Jialiang          | [guojialiang92](https://github.com/guojialiang92)       | ByteDance   |
 | Karen Xu          | [karenyrx](https://github.com/karenyrx)                 | Uber        |
 | Ke Wei            | [kkewwei](https://github.com/kkewwei)                   | ByteDance   |
 | Kunal Kotwani     | [kotwanikunal](https://github.com/kotwanikunal)         | Amazon      |
--- a/plugins/repository-azure/src/main/java/org/opensearch/repositories/azure/AzureBlobContainer.java
+++ b/plugins/repository-azure/src/main/java/org/opensearch/repositories/azure/AzureBlobContainer.java
@@ -40,6 +40,7 @@ import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionRunnable;
 import org.opensearch.action.support.GroupedActionListener;
 import org.opensearch.action.support.PlainActionFuture;
+import org.opensearch.cluster.metadata.CryptoMetadata;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.blobstore.BlobContainer;
 import org.opensearch.common.blobstore.BlobMetadata;
@@ -134,6 +135,25 @@ public class AzureBlobContainer extends AbstractBlobContainer {
         } catch (URISyntaxException | BlobStorageException e) {
             throw new IOException("Can not write blob " + blobName, e);
         }
+    }
+
+    @Override
+    public void writeBlobWithMetadata(
+        String blobName,
+        InputStream inputStream,
+        long blobSize,
+        boolean failIfAlreadyExists,
+        @Nullable Map<String, String> metadata,
+        @Nullable CryptoMetadata cryptoMetadata
+    ) throws IOException {
+        if (cryptoMetadata != null) {
+            throw new UnsupportedOperationException(
+                "Azure Blob Storage repository does not currently support CryptoMetadata. "
+                    + "Consider using repository-level encryption settings instead."
+            );
+        }
+        // Azure does not support custom metadata, so we just delegate to writeBlob
+        writeBlob(blobName, inputStream, blobSize, failIfAlreadyExists);
     }
 
     @Override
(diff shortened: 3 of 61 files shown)


server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -561,6 +697,191 @@ private static void parseObjectOrField(ParseContext context, Mapper mapper) thro
...
+    /**
+     * Parses fields with dots as literal field names when disable_objects is enabled.
+     * This method treats dotted field names (e.g., "user.name", "metrics.cpu.usage") as flat fields
+     * rather than expanding them into nested object hierarchies.
+     * 
+     * @param context the parse context
+     * @param mapper the object mapper with disable_objects enabled
+     * @throws IOException if parsing fails
+     */
+    private static void parseFieldWithDots(ParseContext context, ObjectMapper mapper) throws IOException {
```
bharath-techie: Kindly use consistent naming `disable objects` everywhere and also update the PR description.

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -494,6 +549,80 @@ private static void generateGroupingCriteria(ParseContext context) {
...
+        XContentParser parser = context.parser();
+        XContentParser.Token token = parser.nextToken(); // Move to first field or END_OBJECT
+        
+        while (token != XContentParser.Token.END_OBJECT) {
+            if (token == XContentParser.Token.FIELD_NAME) {
+                String fieldName = parser.currentName();
+                String flattenedFieldName = prefix + "." + fieldName;
+                token = parser.nextToken(); // Move to the value
+                
+                if (token == XContentParser.Token.START_OBJECT) {
```
bharath-techie: Reuse this code between this block and `flattenArray` method ?

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -561,6 +697,191 @@ private static void parseObjectOrField(ParseContext context, Mapper mapper) thro
...
+                        token = parser.nextToken();
+                        
+                        // When disable_objects is enabled and we encounter a nested object, flatten it
+                        if (token == XContentParser.Token.START_OBJECT) {
+                            flattenObject(context, mapper, currentFieldName);
+                        } else if (token == XContentParser.Token.START_ARRAY) {
+                            flattenArray(context, mapper, currentFieldName);
+                        } else {
+                            // Handle primitive values with dynamic mapping
+                            ObjectMapper.Dynamic dynamic = dynamicOrDefault(mapper, context);
```
bharath-techie: Why not use reuse parts of `parseDynamicValue` for this ?

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -561,6 +697,191 @@ private static void parseObjectOrField(ParseContext context, Mapper mapper) thro
...
+            // if we are just starting an OBJECT, advance, this is the object we are parsing, we need the name first
+            token = parser.nextToken();
+        }
+
+        try {
+            assert token == XContentParser.Token.FIELD_NAME || token == XContentParser.Token.END_OBJECT;
+            context.incrementFieldCurrentDepth();
+            context.checkFieldDepthLimit();
+            
+            while (token != XContentParser.Token.END_OBJECT) {
```
bharath-techie, line 821: Is it possible to reuse existing methods for this ? 
Lets take what we can from existing flows such that originate from `parseObjectOrNested` - this has lot of duplicate code

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -244,7 +288,7 @@ static Mapping createDynamicUpdate(Mapping mapping, DocumentMapper docMapper, Li
         Iterator<Mapper> dynamicMapperItr = dynamicMappers.iterator();
         List<ObjectMapper> parentMappers = new ArrayList<>();
         Mapper firstUpdate = dynamicMapperItr.next();
-        parentMappers.add(createUpdate(mapping.root(), splitAndValidatePath(firstUpdate.name()), 0, firstUpdate));
+        parentMappers.add(createUpdate(mapping.root(), splitPathRespectingDisableObjects(firstUpdate.name(), docMapper), 0, firstUpdate));
```
bharath-techie: `splitPathRespectingDisableObjects` - this method has made the flow heavy though ? is there a way to exit faster instead of looping through finding all the objectMappers ?

bharath-techie:
Kindly add benchmarks for this change , lets benchmark atleast `big5` / `nyc_taxis`.

bharath-techie:
{"run-benchmark-test": "id_1"}

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: PR description is updated, but the function name makes sense, in comment I have already stated `disable_objects` param

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: 1. Created processFieldWithDisableObjects() method to handle field processing logic
This consolidates the mapper lookup and processing logic
2. Removed the getFieldType() method that was no longer needed after using parseDynamicValue

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: Done
Replaced the entire dynamic mapping switch statement in parseFieldWithDots with a call to parseDynamicValue

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: Before: Both flattenObject and flattenArray had identical token processing logic duplicated in each method.

After:
1. Extracted shared logic into processFlattenedToken() method
2. Eliminated code duplication - the token handling logic is now in one place

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: Scenarios optimized:-
1. Optimized no dots: If the field path doesn't contain dots, return immediately without any mapper lookups
2. For single part: If after splitting there's only one part, return immediately
3. Single map lookup: Get the objectMappers map once instead of calling docMapper.objectMappers().get() repeatedly
4. Optimized string building: More efficient StringBuilder usage for the common case
5. Early termination: The method returns as soon as it finds a parent with disable_objects=true

bharath-techie:
{"run-benchmark-test": "id_1"}

bharath-techie:
{"run-benchmark-test": "id_1"}

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
bharath-techie: Lets resolve this after running benchmarks please. 

We need to do benchmarks with Big5 workload which has lots of fields with dots.

bharath-techie:
{"run-benchmark-test": "id_1"}

CHANGELOG.md
```diff
@@ -5,6 +5,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.x]
 ### Added
+- Add support for fields containing dots in their name as literals ([#19958](https://github.com/opensearch-project/OpenSearch/pull/19958))
+- Add support for a ForkJoinPool type ([#19008](https://github.com/opensearch-project/OpenSearch/pull/19008))
```
bharath-techie: Lets remove CHANGELOG diff unrelated to this PR

mohit10011999: Done

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -957,6 +957,22 @@ static Map<String, Object> parseV1Mappings(
                     mappings = templateMapping;
                 } else {
                     XContentHelper.mergeDefaults(mappings, templateMapping);
```
bharath-techie: How about V2 templates, are those tested with the change?

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -494,6 +572,117 @@ private static void generateGroupingCriteria(ParseContext context) {
...
+        XContentParser.Token token
+    ) throws IOException {
+
+        // Handle structured tokens (objects and arrays) with recursive flattening
+        switch (token) {
+            case START_OBJECT:
+                flattenObjectForDisableObjects(context, mapper, fieldName);
+                return;
+            case START_ARRAY:
+                flattenArrayForDisableObjects(context, mapper, fieldName);
```
bharath-techie: Why do we need custom logic here ? This method skips validations such as this : https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java#L709

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -427,6 +476,49 @@ static void parseObjectOrNested(ParseContext context, ObjectMapper mapper) throw
...
+    private static void processToken(
+        ParseContext context,
+        ObjectMapper mapper,
+        String fieldName,
+        String[] paths,
+        XContentParser.Token token
+    ) throws IOException {
+        switch (token) {
+            case START_OBJECT:
+                if (mapper.disableObjects()) {
```
bharath-techie: Is it possible to remove if/else checks and handle this in existing methods ? this is error prone as logic added in one method could be missed in other method.

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
mohit10011999: Added similar handling for V2 templates
added extensive UTs for V1 and V2 templates

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: Used parseNonDynamicArray method and modified to cater to disableObjects case

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: Done used existing methods parseObject and parseArray here and other places wherever possible

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -427,6 +476,41 @@ static void parseObjectOrNested(ParseContext context, ObjectMapper mapper) throw
         }
     }
 
+    /**
+     * Processes different token types during object parsing
+     */
+    private static void processToken(
```
mgodwan: We have some tokens handled in innerParseObject already. This method doesn't seem to add a lot of value. Can we remove this and move the logic back?

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -494,6 +564,117 @@ private static void generateGroupingCriteria(ParseContext context) {
...
+        }
+        // For null values in disable_objects mode, we typically don't create mappers
+    }
+
+    /**
+     * Functional interface for processing flattened fields that can throw IOException.
+     */
+    @FunctionalInterface
+    private interface FieldProcessor {
+        void accept(String fieldName, XContentParser.Token token) throws IOException;
```
mgodwan: Is this needed? I see same implementation being passed to the flattenObjectFields method?

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -561,129 +753,299 @@ private static void parseObjectOrField(ParseContext context, Mapper mapper) thro
...
-                    parseObjectOrField(context, mapper);
-                } else {
-                    parseNonDynamicArray(context, parentMapper, lastFieldName, arrayFieldName);
-                }
+        // Handle both flat and nested modes internally based on mapper's disable_objects setting
+        if (mapper.disableObjects()) {
+            // disable_objects mode: incorporate object flattening logic directly
+            parseObjectForDisabledObjects(context, mapper, currentFieldName);
+        } else {
+            // Nested mode: use existing nested object parsing logic
```
mgodwan: [nit] can we call it object mode. nested has a different meaning.

mohit10011999: Done , changed the comment to Object mode and at other places

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: Its redundant w.r.t functionality now, removed it and handled `flattenObjectFields` method accordingly

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: Removed the method `processToken` and moved the logic to `innerParseObject` method

Issue #19225, opened 2025-09-03
[Refactor] Use Builder pattern instead of constructors for Stats API classes


sandeshkr419:
### Is your feature request related to a problem? Please describe

As a follow-up on https://github.com/opensearch-project/OpenSearch/issues/18723, I noticed other stats related classes should also utilize Builder pattern to initialize stats objects.

For example IndexingStats currently has 2 constructors: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/shard/IndexingStats.java#L191-L246

Stats related classes are exposed as PublicApi, and stats are frequently (or maybe once in a while but need to be) incremented, which causes backward-compatibility to break. This makes it difficult to add new metrics and evolve the API without causing significant disruption for developers who have adopted it. Each new stat forces a major version bump and requires users to update their code, even if they don't need the new metric.

Using this issue to track all such metric/stats classes to adopt to builder pattern and mark the existing public constructors to be on deprecation path. 

Reference from SearchStats: https://github.com/opensearch-project/OpenSearch/pull/18707

### Describe the solution you'd like

- List down all classes to be actioned from stats API / code dive deep
- Create builder pattern for those classes
- Mark existing constructors as deprecated

### Related component

Other

pado0:
Hi @sandeshkr419 , can I work on this issue?

sandeshkr419:
Thanks @pado0 for taking this, feel free to post any questions if you need help.

pado0:
@sandeshkr419 Thanks for assigning this to me! I’ll reach out if I run into any difficulties. 👍

pado0:
Hello, @sandeshkr419 I’m planning to work on this and would like to confirm a few details:

**First**  
I identified a few different patterns of how `Stats` classes are currently defined:  

1. **Pattern 1**: Using `Stats` as an inner class & exposed as Public API  
   - [`IndexStats`  ](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/admin/indices/stats/IndexStats.java)
   - [`RemoteTranslogTransferTracker` ](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/remote/RemoteTranslogTransferTracker.java) 

2. **Pattern 2**: Using `Stats` as an inner class & marked as internal API  
[...]

PR #19961, opened 2025-11-11
Refactor TranslogStats and RequestCacheStats with Builder pattern

pado0:
### Description
This PR refactors the `TranslogStats`  and `RequestCacheStats` class to use the Builder pattern instead of relying on multiple constructors.  

By adopting the Builder pattern, it becomes easier to evolve the stats API, add new metrics, and maintain backward compatibility without forcing disruptive constructor changes.

Based on the related issue:
1. Added a Builder and deprecated the existing constructors.
2. Replaced usages of constructors in code and tests with the new Builder.

There are multiple stats-related classes that need similar refactoring, and we are addressing them in priority order. This PR covers `TranslogStats`  and `RequestCacheStats`  as part of that effort.

### Related Issues
Related to #19225 

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/translog/TranslogStats.java
+++ b/server/src/main/java/org/opensearch/index/translog/TranslogStats.java
@@ -193,26 +193,41 @@ public class TranslogStats implements Writeable, ToXContentFragment {
         public Builder() {}
 
         public Builder numberOfOperations(int operations) {
+            if (operations < 0) {
+                throw new IllegalArgumentException("numberOfOperations must be >= 0");
+            }
             this.numberOfOperations = operations;
             return this;
         }
 
         public Builder translogSizeInBytes(long size) {
+            if (size < 0) {
+                throw new IllegalArgumentException("translogSizeInBytes must be >= 0");
+            }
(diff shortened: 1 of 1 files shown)


sandeshkr419 (approved):

server/src/main/java/org/opensearch/index/translog/TranslogStats.java
```diff
@@ -159,6 +178,69 @@ public RemoteTranslogStats getRemoteTranslogStats() {
...
+            }
+            if (uncommittedOperations < 0) {
+                throw new IllegalArgumentException("uncommittedOperations must be >= 0");
+            }
+            if (uncommittedSizeInBytes < 0) {
+                throw new IllegalArgumentException("uncommittedSizeInBytes must be >= 0");
+            }
+            if (earliestLastModifiedAge < 0) {
+                throw new IllegalArgumentException("earliestLastModifiedAge must be >= 0");
+            }
```
sandeshkr419: nit: you can actually pull these checks in the setters instead to throw exception early. Basically don't let anyone set the values in the builder in the first place before waiting to build.

PR #19963, opened 2025-11-11
[Pull-based Ingestion] Update ingestion stream params to be dynamic

varunbharadwaj:
### Description

1. Update ingestion source params to be dynamic, to make it easy to tune/update consumer settings. On updating ingestion params, a new consumer will be created and start polling from the latest batchStartPointer at that time.
2. Integration tests are refactored to extract pause and resume methods into the parent base test class.

Note: If multiple writer threads are used, it is possible that the new consumer reprocess some messages as it starts from batchStartPointer, similar to shard recovery scenario. This aligns with pull-based ingestion's atleast once message processing guarantee.

### DIFF
--- a/server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
+++ b/server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
@@ -576,11 +576,11 @@ public class DefaultStreamPoller implements StreamPoller {
     private void closeConsumer() {
         if (this.consumer != null) {
             try {
-                logger.info("Closing consumer for shard {}", shardId);
+                logger.info("Closing consumer for index {} shard {}", indexName, shardId);
                 this.consumer.close();
                 this.consumer = null;
             } catch (Exception e) {
-                logger.warn("Error closing consumer for shard {}: {}", shardId, e.getMessage());
+                logger.warn("Error closing consumer for index {} shard {}: {}", indexName, shardId, e.getMessage());
             }
         }
     }


andrross:
> The changes will only reflect when the consumer is reinitialized on reset, or on shard reinitialization

Does this mean the update settings API will accept the new values but nothing will actually happen until the shard is somehow reinitialized? Seems potentially quite confusing as most (all?) dynamic settings in OpenSearch are applied right away.

> to avoid the need for recreating index whenever consumer settings need to be updated

Is it possible to close the index, update the settings, and reopen the index as an alternative that is less painful than recreating?

varunbharadwaj:
>Seems potentially quite confusing as most (all?) dynamic settings in OpenSearch are applied right away.

Makes sense, i'll see if we can add a listener and swap the consumer dynamically similar to reset API

>Is it possible to close the index, update the settings, and reopen the index as an alternative that is less painful than recreating?

That should work, but will result in some downtime.

andrross:
> > Is it possible to close the index, update the settings, and reopen the index as an alternative that is less painful than recreating?
> 
> That should work, but will result in some downtime.

@varunbharadwaj Doesn't the approach in this PR require some downtime as well since it requires a disruptive action like reset or reinitialization to apply the new settings? If we can make them fully dynamic then that is great, it's just this in-between state where you can update them but need to do something else to make them take effect that I have concerns about.

varunbharadwaj:
> @varunbharadwaj Doesn't the approach in this PR require some downtime as well since it requires a disruptive action like reset or reinitialization to apply the new settings? If we can make them fully dynamic then that is great, it's just this in-between state where you can update them but need to do something else to make them take effect that I have concerns about.

Yeah, thinking we can create a new consumer with the new settings and swap with existing consumer. There will be downtime only on indexing side. Query should not be impacted.

With the current approach, we will have to restart the nodes or relocate the shards. It should not have 100% downtime as in the case of closing the index. But I agree that it can be confusing, if it doesn't reflect immediately.

varunbharadwaj:
Have added a listener to re-create the consumer on the fly with the new settings. We use the batchStartPointer as the start point for the new consumer. 
cc @andrross  @msfroh @yupeng9

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -519,6 +524,37 @@ private void updateErrorHandlingStrategy(IngestionErrorStrategy.ErrorStrategy er
...
+            // Initialize the factory with updated params
+            ingestionConsumerFactory.initialize(updatedIngestionSource);
+
+            // Request consumer reinitialization in the poller
+            streamPoller.requestConsumerReinitialization();
+
+            logger.info("Successfully processed ingestion source params update");
+        } catch (Exception e) {
+            logger.error("Failed to update ingestion source params", e);
+            throw new OpenSearchException("Failed to update ingestion source params", e);
```
andrross: Log-and-throw is generally an anti-pattern as the thrown exception should ultimately lead to a log message so this usually leads to double logging. Also, catching Exception can be problematic in general as it can lead to swallowing interrupts if there is anything block here. Do you need the try-catch at all? Why not just let the underlying exception bubble up to be handled?

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -459,6 +476,21 @@ public IngestionShardConsumer getConsumer() {
...
+     */
+    @Override
+    public void requestConsumerReinitialization() {
+        if (closed) {
+            logger.warn("Cannot reinitialize consumer for closed poller of shard {}", shardId);
+            return;
+        }
+
+        logger.info("Configuration parameters updated for index {} shard {}, requesting consumer reinitialization", indexName, shardId);
+        reinitializeConsumer = true;
```
msfroh, line 478: There is a potential race condition on `reinitializeConsumer`. 

If you update the parameters twice in quick succession, the sequence could be as follows:
1. First param update sets `reinitializeConsumer` to `true`.
2. We enter the `if` block in `startPoll` and run `handleConsumerInitialization` to open the new consumer.
3. The second parameter update comes in. Since `reinitializeConsumer` is already true, the `requestConsumerReinitialization` method does nothing.
4. We set `reinitializeConsumer = false` at the end of the `if` block in `startPoll`.

I think we can fix it as follows:
1. Move the `reinitializeConsumer = false` statement into `handleConsumerInitialization`.
2. Add `synchronized` to both `handleConsumerInitialization` and `requestConsumerReinitialization`.

I'm wondering if there's a way that we can do it with a reference-counted, lock-free data structure, like `ReferenceManager`. Thinking out loud: What if `consumer` was a `ReferenceManager<IngestionShardConsumer>`? In each iteration in `startPoll`, we would acquire the current `IngestionShardConsumer`, use it, then release it. Then if a parameter change comes in, we would trigger a refresh. That will decref the current `IngestionShardConsumer`, so it will automatically get closed when the current poll releases it. On the next iteration, we'll pick up the new consumer. Maybe I'm overcomplicating it, but what you're doing here feels very similar to refreshing a searcher to me.

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
varunbharadwaj: Updated to remove the try-catch.

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
varunbharadwaj: I like the `reference-counted, lock-free data structure` idea, but thinking we might need to refactor the DefaultStreamPoller class first before that. Updated to handle the race condition and visibility scenarios.

Issue #19816, opened 2025-10-29
[Feature Request] Preserve Original Search Request Body


KishoreKicha14:
### Is your feature request related to a problem? Please describe

The Query Insights plugin needs access to the original search request body as submitted by users, but currently OpenSearch QueryBuilder classes automatically populate all fields with default values during construction. This makes it impossible to distinguish between user-specified values and system defaults.

For example, when a user submits:

`{"query": {"match": {"field": "value"}}}
`
The MatchQueryBuilder constructor adds default fields like operator, prefix_length, max_expansions, etc., polluting the original request with defaults that the user never specified.

### Describe the solution you'd like

Add an originalRequestBody field to SearchRequest that captures the raw JSON request before QueryBuilder processing. This preserves the exact user input before default value pollution, enabling Query Insights to analyze actual user query.

### Describe alternatives you've considered

1. Modify QueryBuilder classes to track user-specified fields
2. Parse request body separately in Query Insights

### Additional context

Related issue: https://github.com/opensearch-project/query-insights-dashboards/issues/30

AmandaBalderas20:
Hi, I'd like to work on this enhancement.

The proposal to add an `originalRequestBody` field to `SearchRequest` makes sense, it would allow Query Insights to distinguish between user-specified values and system defaults.  
I can implement this change by capturing the raw JSON request before QueryBuilder processing and adding the new field to `SearchRequest` (with serialization and tests).

Can you assign this issue to me? plss

kkhatua:
Hi @AmandaBalderas20 ... Thanks for taking this up. I've just assigned this to you.

epugh:
I wanted to mention that tracking the the original user query (what they typed in) as well as the parsed query that OpenSearch gets is what the https://github.com/opensearch-project/user-behavior-insights plugin does.

sandeshkr419:
[Search Triage] Some early clarifying questions came up while discussing this issue.

- What if the request body is not a json and just binary content?
- What if the request raw body is too big?
- Do we intend to do this only at the coordinator level and not at the shard level?
- Also, this might be a good addition specially at the coordinator level when emitting out slow logs.

PR #19970, opened 2025-11-12
Add originalRequestBody field to SearchRequest and handle it in RestSearchAction

AmandaBalderas20:
### Description
This PR introduces a new field `originalRequestBody` to the `SearchRequest` class.  
It allows OpenSearch to retain the original raw JSON request body sent through the REST layer, before any parsing or system modification occurs.  

This enhancement enables downstream components (like Query Insights or analytics tools) to distinguish between:
- **User-provided query values**, and
- **System defaults or injected parameters**

#### Key changes:
- Added a new field `BytesReference originalRequestBody` to `SearchRequest`.
- Updated `writeTo()` and the constructor from `StreamInput` for proper serialization/deserialization.
- Modified `RestSearchAction` to capture and store the original body from `request.contentOrSourceParam().v2()`.
- Added two unit tests in `SearchRequestTests`:
  - `testOriginalRequestBodySerialization`
  - `testOriginalRequestBodyIsNullByDefault`

---

### Related Issues
Resolves [#19816](https://github.com/opensearch-project/OpenSearch/issues/19816)

---

### Check List
- [x] Functionality includes testing (`SearchRequestTests`)
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable. *(N/A — internal field only)*
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable. *(N/A — internal change)*

---

### DIFF
--- a/server/src/main/java/org/opensearch/action/search/SearchRequest.java
+++ b/server/src/main/java/org/opensearch/action/search/SearchRequest.java
@@ -42,6 +42,8 @@ import org.opensearch.common.Nullable;
 import org.opensearch.common.annotation.PublicApi;
 import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.common.bytes.BytesArray;
 import org.opensearch.core.common.Strings;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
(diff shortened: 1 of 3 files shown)


kaushalmahi12:
Thanks @AmandaBalderas20 for making this change but can we think of alternate ways to capture this info outside core.
Since there is no usecase for this new field in the core opensearch, we should try to intercept the request in plugin before it reaches the core. 

We can maybe try to look into UBI plugin for reference
https://github.com/opensearch-project/OpenSearch/issues/19816#issuecomment-3523087491

AmandaBalderas20:
Okay, thanks, so I’ll review the User Behavior Insights (UBI) plugin to see how it captures the original request body.  
If this logic fits better there instead of the core, I can explore implementing it as part of that plugin.

PR #19973, opened 2025-11-12
Added index.optimized setting for configuring the indexing engine at index level

cocosz:
### Description
Created a new index setting `index.optimized` setting for configuring the indexing engine at index level

### Testing

To verify the new index setting functionality, you can create an index with a custom configuration using the following `curl` command:

```bash
curl -X PUT "http://localhost:9200/my-index" \
  -d '{
    "settings": {
      "index": {
        "number_of_shards": 3,
        "number_of_replicas": 1,
        "refresh_interval": "30s",
        "optimized": true
      }
    },
    "mappings": {
      "properties": {
        "message": { "type": "text" },
        "timestamp": { "type": "date" }
      }
    }
  }'
```
Response :

```
{"acknowledged":true,"shards_acknowledged":true,"index":"my-index"}
```

After creating the index, ran the following command to verify that the custom settings were applied:

```bash
curl -X GET "http://localhost:9200/my-index/_settings?pretty"
```
Response :
```
{
  "my-index": {
    "settings": {
      "index": {
        "replication": {
          "type": "DOCUMENT"
        },
        "refresh_interval": "30s",
        "number_of_shards": "3",
        "optimized": "true",
        "provided_name": "my-index",
        "creation_date": "1762954402418",
        "number_of_replicas": "1",
        "uuid": "PzoDM1SHTpmyek3ON-0mnA",
        "version": {
          "created": "137247827"
        }
      }
    }
  }
}

```

### DIFF
--- a/server/src/main/java/org/opensearch/index/IndexSettings.java
+++ b/server/src/main/java/org/opensearch/index/IndexSettings.java
@@ -598,8 +598,8 @@ public final class IndexSettings {
         Property.Dynamic
     );
 
-    public static final Setting<Boolean> INDEX_OPTIMIZED = Setting.boolSetting(
-        "index.optimized",
+    public static final Setting<Boolean> INDEX_ENGINE_LEGO = Setting.boolSetting(
+        "index.engine.lego",
         true,
         Property.IndexScope,
         Property.Dynamic
@@ -906,7 +906,7 @@ public final class IndexSettings {
     private volatile String requiredPipeline;
     private volatile boolean searchThrottled;
     private volatile boolean shouldCleanupUnreferencedFiles;
-    private volatile boolean indexOptimized;
+    private volatile boolean indexEngineLego;
     private volatile long mappingNestedFieldsLimit;
     private volatile long mappingNestedDocsLimit;
     private volatile long mappingTotalFieldsLimit;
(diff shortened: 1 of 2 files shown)


msfroh:
> Created a new index setting index.optimized setting for configuring the indexing engine at index level

@cocosz  -- Can you please come up with a different name that reflects what's actually happening? Why would anyone ever want their index non-optimized?

Lucene used to have an API called `IndexWriter#optimize`, which everyone always assumed was a great thing to call. In Solr, there was a big "Optimize" button in the web UI, and people would push it all the time (since you obviously want stuff to be optimized). To reduce confusion, the name was changed to `forceMerge` to reflect what it really does. (Sometimes you do want to force merge, but sometimes you don't.)

Bukhtawar:
Maybe `index.engine.lego` would be a better fit to reflect a composable engine that supports pluggable data formats and query execution engine.

cocosz:
We can rename it to `index.engine.lego` considering the nature of the plugin

server/src/main/java/org/opensearch/index/IndexSettings.java
```diff
@@ -2018,6 +2028,14 @@ private void setShouldCleanupUnreferencedFiles(boolean shouldCleanupUnreferenced
         this.shouldCleanupUnreferencedFiles = shouldCleanupUnreferencedFiles;
     }
 
+    public boolean getIndexEngineLego() {
+        return indexEngineLego;
+    }
+
+    private void setIndexEngineLego(boolean indexEngineLego) {
```
darjisagar7: Once the server is up, I don't think this settings should be changeable. Also check if updating this setting throws the appropriate error message.

cocosz: We can make the setting `index.engine.lego` as static index setting. This will allow the setting to throw an error if user tries to update it.

cocosz: Tried to change setting after code change getting error as expected  :

```
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Can't update non dynamic settings [[index.engine.lego]] for open indices [[my-index/rdhhVM2aR6Kf5yyJBi4iag]]"}],"type":"illegal_argument_exception","reason":"Can't update non dynamic settings [[index.engine.lego]] for open indices [[my-index/rdhhVM2aR6Kf5yyJBi4iag]]"},"status":400}% 
```

msfroh:
Does this change need to go into the server module? 

If you implement an `EnginePlugin`, you can define a custom setting (returned from `Plugin#getSettings()`) and then react to it in your implementation of `getEngineFactory`.

PR #19974, opened 2025-11-12
Avoid eager getByName in pom validation plugin

andrross:
The `getByName()` call is eager, which means this can fail non-deterministically if the dependent gradle tasks have not yet been configured. I ran into this error when doing other gradle refactorings. This change avoids the eager call while also avoiding the nested publications.all() call.

### Check List

### DIFF
--- a/buildSrc/src/main/java/org/opensearch/gradle/precommit/PomValidationPrecommitPlugin.java
+++ b/buildSrc/src/main/java/org/opensearch/gradle/precommit/PomValidationPrecommitPlugin.java
@@ -59,13 +59,8 @@ public class PomValidationPrecommitPlugin extends PrecommitPlugin {
             validateTask.configure(task -> {
                 task.dependsOn(generateMavenPom);
                 task.getPomFile().fileProvider(generateMavenPom.map(GenerateMavenPom::getDestination));
-                publishing.getPublications().all(publicationForPomGen -> {
-                    task.mustRunAfter(
-                        project.getTasks()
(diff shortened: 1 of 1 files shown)


andrross:
@cwperks @zelinh Can you take a look? You were both involved in the original change here.

I did test this locally by removing the lines [here in the geospatial plugin](https://github.com/opensearch-project/geospatial/blob/430077dc11ec2d77f9ba2c0f4e4f60fae8a3def3/libs/h3/build.gradle#L76-L77) and verified that `./gradlew :libs:h3:validatePom` still worked. Prior to @zelinh's change those extra dependencies were sometimes required in plugins to get things to work correctly.

cwperks:
> The getByName() call is eager, which means this can fail non-deterministically if the dependent gradle tasks have not yet been configured.

Any idea what changed in gradle? I ran into issues updating the security plugin to gradle 9 due to eager loading as well.

andrross:
> > The getByName() call is eager, which means this can fail non-deterministically if the dependent gradle tasks have not yet been configured.
> 
> Any idea what changed in gradle? I ran into issues updating the security plugin to gradle 9 due to eager loading as well.

@cwperks  I think this is just fragile because there is an implicit dependency between tasks/projects that isn't fully modeled. As I understand it, you need to make sure every thing is lazy to avoid these issues. Alternatively, you can add `evaluationDependsOn(...)` to model dependencies between projects but that is generally not recommended and can lead to other problems.

PR #19976, opened 2025-11-12
Add case sensitivity as an argument to XContentMapValues.filter

cwperks:
### Description

This PR solves an issue in the RestRequestFilter extension point that is used in cases to mark request params as sensitive so that the request is filtered in cases where the request body may be logged somewhere. Currently, it uses [XContentMapValues.filter](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/RestRequestFilter.java#L76-L80) to perform the filtering, but implementers of this extension point would need to add all permutations of the field in [getFilterFields](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/RestRequestFilter.java#L96-L99) due to the case sensitivity of the filtering.

I propose to make this case insensitive to filter out regardless of the casing.

This is particular useful for endpoints like https://docs.opensearch.org/latest/ml-commons-plugin/remote-models/index/#step-2-create-a-connector 

### Check List


andrross:
I believe the existing `filter` implementation is used for [selecting which fields to return from source](https://docs.opensearch.org/latest/search-plugins/searching-data/retrieve-specific-fields/#specifying-the-fields-to-retrieve), and probably many other user-facing behaviors. I think we'll need to be more surgical about introducing a case-insensitive option and only use it cases that do not change existing semantics.

jainankitk (commented):
I am wondering if we should add an option for case insensitive filter instead of changing the default behavior. That should allow addressing the issue for RestRequestFilter extension point in the short term, while keeping the default behavior. Changing default behavior can be separate decision in the longer term

cwperks:
Ack, I'll push a commit so that only the path within RestRequestFilter is case insensitive.

cwperks:
@andrross @jainankitk I pushed a commit to add case sensitivity as a param so it can be used within RestRequestFilter.

server/src/main/java/org/opensearch/common/xcontent/support/XContentMapValues.java
```diff
@@ -215,18 +216,37 @@ private static Object extractValue(String[] pathElements, int index, Object curr
...
+        return filter(includes, excludes, caseSensitive).apply(map);
     }
 
     /**
      * Returns a function that filters a document map based on the given include and exclude rules.
      * @see #filter(Map, String[], String[]) for details
      */
-    public static Function<Map<String, ?>, Map<String, Object>> filter(String[] includes, String[] excludes) {
+    public static Function<Map<String, ?>, Map<String, Object>> filter(String[] includes, String[] excludes, boolean caseSensitive) {
         if (hasNoWildcardsOrDots(includes) && hasNoWildcardsOrDots(excludes)) {
```
jainankitk, line 246: Given this class is annotated as @opensearch.internal, we should be okay just adding this new parameter. Else, we probably need to maintain method with new and old signature?

jainankitk (approved):
Looks good to me!

server/src/main/java/org/opensearch/common/xcontent/support/XContentMapValues.java
cwperks: There are 20+ usages (mostly in tests) for the other signature so I opted to keep it as default behavior and add a second with a parameter. This same pattern is present in many places (i.e. Setting.listSetting) which has many different overrides. I can follow up this PR with another to remove the old signature if you think that would be beneficial.

jainankitk: I don't have strong opinion on this, was just an observation. We should be okay either way due to the class annotation.

PR #19982, opened 2025-11-13
Fixing the infinite loop issue and avoiding the merge if the file is already merged

darjisagar7:
### Description
Fixing the infinite loop issue and avoiding the merge if the file is already merged

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/engine/exec/merge/MergeHandler.java
+++ b/server/src/main/java/org/opensearch/index/engine/exec/merge/MergeHandler.java
@@ -55,7 +55,7 @@ public abstract class MergeHandler {
 
     public abstract Collection<OneMerge> findForceMerges(int maxSegmentCount);
 
-    public void updatePendingMerges() {
+    public synchronized void updatePendingMerges() {
         Collection<OneMerge> oneMerges = findMerges();
 //        System.out.println("Found merges : " + oneMerges);
         for (OneMerge oneMerge : oneMerges) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/opensearch/index/engine/exec/merge/MergeHandler.java
```diff
@@ -72,6 +72,20 @@ public void updatePendingMerges() {
     }
 
     public synchronized void registerMerge(OneMerge merge) {
+        try (CompositeEngine.ReleasableRef<CatalogSnapshot> catalogSnapshotReleasableRef = compositeEngine.acquireSnapshot()) {
+            Set<FileMetadata> catalogFiles = new HashSet<>();
+            for (CatalogSnapshot.Segment segment : catalogSnapshotReleasableRef.getRef().getSegments()) {
+                catalogFiles.addAll(segment.getSearchableFiles(compositeDataFormat.getPrimaryDataFormat().name()));
```
dhwanilpatel: Why primary data format? It should be oneMerge's dataformat.

Also we can directly fetch the searchableFiles for oneMerge's dataformat and update set instead of iterating over segments. It will provide files for all segments.

darjisagar7: if we are saying something like this ```catalogSnapshotReleasableRef.getRef().getSearchableFiles(merge.getDataFormat().name())``` then this returns `WriteFileSet` and not FileMetadata

PR #19988, opened 2025-11-13
Test odd behavior of suppress warnings?

andrross:
Testing only, will close


andrross:
Failure seems to be caused by [this call to disable assertions](https://github.com/opensearch-project/OpenSearch/blob/e52ebe4cb541d9fbf57c91a4653ae3b29b24fabf/server/src/test/java/org/opensearch/index/engine/MergedSegmentWarmerTests.java#L145) in the test. That should be removed since it changes the assertion behavior for that JVM instance which can impact other tests.

Issue #19741, opened 2025-10-23
[Feature Request] Remove maps from hot loop in matrix_stats agg for performance


peteralfonsi:
### Is your feature request related to a problem? Please describe

The matrix_stats agg is somewhat slow, and dominated by the per-doc loop where we add stats for each doc to a `RunningStats` object. This object keeps track of its numeric stats in Map<String, Double> (or Long) where the keys are the fields provided in the agg. I noticed a decent amount of CPU time (6%) spent just on unnecessary checks to HashMap.containsKey(), and larger chunks are spent on .put() and .get(). That made me suspect the use of maps here was slowing things down. 

<img width="2518" height="740" alt="Image" src="https://github.com/user-attachments/assets/8adbe36f-5354-49e9-81fc-a7cc096bfd9b" />

### Describe the solution you'd like

Since the fields in a given agg don't change, we know them up front and they're the same for all shards. So we can store the stats for each field in double[] and long[], where the index corresponds to the field index, instead of maps. This spares us from Map.get(), .put(), and also unboxing.

 In a proof-of-concept branch, I found the OSB p50 latency for the below agg on nyc_taxis goes **from 15.639 seconds to 3.093 seconds, an 80% speedup.** Based on the flamegraphs I suspect most of this speedup is from not having to unbox. 

```
curl -XGET "localhost:9200/_search?pretty" -H 'Content-Type: application/json' -d'
          {
          "size": 0,
        "query": {
          "range": {
              "dropoff_datetime": {
              "gte": "01/01/2015",
              "lte": "31/12/2015",
              "format": "dd/MM/yyyy"
            }
          }
        },
        "aggs": {
          "matrix_stats_result": {
            "matrix_stats": {
              "fields":["trip_distance", "tip_amount"]
            }
          }
        }
        }'
```

I took a brief look around the per-doc loops for other aggs to see if any of them also use unnecessary maps like this, and I didn't find any, but we should look more thoroughly since the performance cost from unboxing actually seems really high. 

One complication is the BWC for serialization. If a node gets a RunningStats from an old version, it won't know what the field names were. For example, maybe by freak coincidence none of the docs on that shard had a value for one of the fields in the query, and then we can't even check the key sets of the maps. I guess if we can't think of anything better, we could keep the old map-based logic around and use it only when we have to add more docs or merge with RunningStats from old versions. 

### Related component

Search:Performance

jainankitk:
[...]

PR #19989, opened 2025-11-13
Optimize matrix_stats agg by using primitive arrays instead of maps

peteralfonsi:
### Description
As described in the initial issue, the use of Map<String, Double/Long> in the hot loop of matrix_stats is slow mostly because of un/boxing. Replacing it with double[] / long[] speeds up the aggregation by 80%, from 15.639 seconds to 3.093 seconds on nyc_taxis. 

As discussed with @jainankitk , we unfortunately do have to keep the older map-based logic around for BWC/serialization reasons. I wasn't able to think of a good way to remove it - while it is true that we'll never add RunningStats that are for different aggregation requests, and therefore the fieldNames keys will actually always be the same in practice, we still do need a RunningStats instance constructed with the StreamInput ctor to be functional, even before we add it to the RunningStats from another segment/shard. So we have to keep the old code around as a fallback. Sad since it adds a lot of bloat + logic around switching between the two. However I still think the performance improvement makes the overall change worthwhile. 

Note we still use map-based logic in the other objects like InternalMatrixStats/MatrixStatsResults. This should have ~no perf impact since it only happens at the end, not in the per-doc hot loop. Doing it this way reduces the complexity of having multiple cases for those objects too. 

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/19741

### Check List
- [N/A] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [N/A] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.


peteralfonsi:
distribution:bwc:maintenance:buildBwcLinuxTar is failing for 2.19.4 but not 2.19.3. I think this isn't related to serialization BWC test failures or anything like that since I think those show up as test failures. Not sure why this is happening.

PR #19714, opened 2025-10-22
Replace explicit type conversion with instanceof pattern variable in org.opensearch.search package

gaobinlong:
### Description

Use the modern instanceof pattern variables instead of explicit type conversions in the package org.opensearch.search, this can improve the code readability and reduce potential ClassCastException risks.

Most code are generated by LLM, and I checked them manually and reverted the unnecessary and wrong changes.

### Related Issues
No issue, just [mentioned this in a previous PR](https://github.com/opensearch-project/OpenSearch/pull/19595#discussion_r2428849476).

### Check List
<del> - [ ] Functionality includes testing.
<del>- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
<del>- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #20012, opened 2025-11-15
Replace explicit type conversion with instanceof pattern variable in plugins dir and libs dir

gaobinlong:
### Description
Similar to https://github.com/opensearch-project/OpenSearch/pull/19714.

Use the modern instanceof pattern variables instead of explicit type conversions in all moduels, this can improve the code readability and reduce potential ClassCastException risks.

### Related Issues
No issue.

### Check List

### DIFF
--- a/plugins/telemetry-otel/src/main/java/org/opensearch/telemetry/OTelAttributesConverter.java
+++ b/plugins/telemetry-otel/src/main/java/org/opensearch/telemetry/OTelAttributesConverter.java
@@ -44,7 +44,9 @@ public final class OTelAttributesConverter {
             case Long l -> attributesBuilder.put(key, l);
             case Double v -> attributesBuilder.put(key, v);
             case String s -> attributesBuilder.put(key, s);
-            default -> throw new IllegalArgumentException(String.format(Locale.ROOT, "Span attribute value %s type not supported", value));
+            case null, default -> throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "Span attribute value %s type not supported", value)
+            );
         }
     }
 


sandeshkr419 (approved):
LGTM! 

Thanks @gaobinlong for putting efforts in refactoring this!

gaobinlong:
> ❌ Gradle check result for [33e2d12](https://github.com/opensearch-project/OpenSearch/commit/33e2d12d0000b72983d9f8092b6401e259fcaf09): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/67365/)
> 

Need to handle potential NPE for `switch`, will check previous refactor PRs.

sandeshkr419 (approved):

plugins/telemetry-otel/src/main/java/org/opensearch/telemetry/OTelAttributesConverter.java
```diff
@@ -44,7 +44,9 @@ private static void addSpanAttribute(String key, Object value, AttributesBuilder
             case Long l -> attributesBuilder.put(key, l);
             case Double v -> attributesBuilder.put(key, v);
             case String s -> attributesBuilder.put(key, s);
-            default -> throw new IllegalArgumentException(String.format(Locale.ROOT, "Span attribute value %s type not supported", value));
+            case null, default -> throw new IllegalArgumentException(
```
sandeshkr419, line 47: Wondering if you might still get an NPE if `value == null`?

Should we segregate the `null` case from `default`?

```
case null -> throw new UnSupportedOperation/IllegalArgumentException("Required value for ...."); //without trying to access null value
```

Although its a private util and we may not even have the need for it, but I'd prefer to make it safe.

gaobinlong: Yeah, if no null case, then swith (value ) throws NPE. Segregate the null case from default sometimes works fine, for example, the default case doesn't expect the value to be null, but for the case here, I think the behavior of  null case and default case should be same, changing the exception message should be a breaking change I think.

gaobinlong: I've raised another [PR](https://github.com/opensearch-project/OpenSearch/pull/20044) to add `null` case for `switch` added in previous refactor PRs, please help to review, thanks!

PR #20013, opened 2025-11-15
Restructure native handling and integrate with the Cleaner interface

Bukhtawar:
### Description
✅ Type Safety - Typed handles prevent pointer misuse
✅ Resource Management - Automatic cleanup with Cleaner API
✅ Clear Structure - Single source of truth in NativeBridge
✅ Better Error Handling - Custom exceptions and consistent patterns
✅ Maintainable - Clear separation of concerns

### Check List

### DIFF
--- a/libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/NativeHandle.java
+++ b/libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/NativeHandle.java
@@ -9,6 +9,7 @@
 package org.opensearch.vectorized.execution.jni;
 
 import java.lang.ref.Cleaner;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Base class for type-safe native pointer wrappers.
@@ -19,7 +20,7 @@ import java.lang.ref.Cleaner;
 public abstract class NativeHandle implements AutoCloseable {
 
     protected final long ptr;
-    private volatile boolean closed = false;
+    private final AtomicBoolean closed = new AtomicBoolean(false);
     protected static final long NULL_POINTER = 0L;
     private final Cleaner.Cleanable cleanable;
 
@@ -43,7 +44,7 @@ public abstract class NativeHandle implements AutoCloseable {
      * @throws IllegalStateException if the handle has been closed
      */
     protected void ensureOpen() {
-        if (closed) {
+        if (closed.get()) {
             throw new IllegalStateException("Handle already closed");
         }
     }
@@ -59,14 +60,9 @@ public abstract class NativeHandle implements AutoCloseable {
     }
 
     @Override
-    public final void close() {
-        if (!closed) {
-            synchronized (this) {
-                if (!closed) {
-                    closed = true;
-                    cleanable.clean();  // Triggers cleanup immediately
-                }
-            }
+    public void close() {
+        if (closed.compareAndSet(false, true)) {
+            cleanable.clean();
         }
     }
 
--- a/libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/RefCountedNativeHandle.java
+++ b/libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/RefCountedNativeHandle.java
@@ -1,0 +1,57 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.vectorized.execution.jni;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Reference-counted native handle for shared native resources.
+ * Allows multiple owners to safely share a native pointer.
+ * The native resource is released only when the reference count reaches zero.
+ */
+public abstract class RefCountedNativeHandle extends NativeHandle {
+
+    private final AtomicInteger refCount = new AtomicInteger(1);
+
+    /**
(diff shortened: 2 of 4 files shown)


plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/jni/NativeException.java
```diff
@@ -0,0 +1,23 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.datafusion.jni;
+
+/**
+ * Exception thrown for DataFusion JNI errors.
+ */
+public class NativeException extends RuntimeException {
```
bharath-techie: Should we keep this in core ? we can reuse in both parquet / datafusion plugins for example

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/RecordBatchStream.java
```diff
@@ -131,9 +131,15 @@ public void close() throws Exception {
...
+        NativeBridge.streamNext(runtime, pointer, callback);
+    }
 
-    private static native void getSchema(long pointer, ObjectResultCallback callback);
+    private static void getSchema(long pointer, ObjectResultCallback callback) {
+        NativeBridge.streamGetSchema(pointer, callback);
+    }
 
-    private static native void closeStream(long pointer);
+    private static void closeStream(long pointer) {
```
mgodwan, line 142: Should we think about making this class extend from an OpenSearch interface? Beyond the scope of this PR, but would be useful to have something to denote a result set in general as and when we start supporting multiple engines

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/jni/NativeException.java
Bukhtawar: Makes sense

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/core/GlobalRuntimeEnv.java
```diff
@@ -8,25 +8,29 @@
...
 
     /**
      * Creates a new global runtime environment.
      */
     public GlobalRuntimeEnv() {
-        this.ptr = createGlobalRuntime();
-        this.tokio_runtime_ptr = createTokioRuntime();
+        this.ptr = NativeBridge.createGlobalRuntime();
+        this.tokioRuntimePtr = NativeBridge.createTokioRuntime();
+        this.cleanable = CLEANER.register(this, new CleanupAction(ptr, tokioRuntimePtr));
```
alchemist51: Instead of creating a cleaner, should we create sub classes for all the pointers and let the handlers handle the cleaning of them? Since globalRuntimeEnv will have more ptr in upcoming PRs.

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/jni/NativeLibraryLoader.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+
+/**
+ * Handles loading of the native DataFusion JNI library.
+ */
+final class NativeLibraryLoader {
```
alchemist51: We can move this loader as well in core.

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/core/GlobalRuntimeEnv.java
```diff
@@ -8,25 +8,29 @@
...
+    private final long tokioRuntimePtr;
+    private final Cleaner.Cleanable cleanable;
 
     /**
      * Creates a new global runtime environment.
      */
     public GlobalRuntimeEnv() {
-        this.ptr = createGlobalRuntime();
-        this.tokio_runtime_ptr = createTokioRuntime();
+        this.ptr = NativeBridge.createGlobalRuntime();
```
alchemist51: How are we thinking to handle the error handling? In the rust side we are returning back 0, but there is no cache here for it.

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -235,12 +216,14 @@ pub extern "system" fn Java_org_opensearch_datafusion_DataFusionQueryJNI_createD
...
-pub extern "system" fn Java_org_opensearch_datafusion_DataFusionQueryJNI_closeDatafusionReader(
-    mut env: JNIEnv,
+pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_closeDatafusionReader(
+    _env: JNIEnv,
     _class: JClass,
     ptr: jlong,
 ) {
-    let _ = unsafe { Box::from_raw(ptr as *mut ShardView) };
+    if ptr != 0 {
+        let _ = unsafe { Box::from_raw(ptr as *mut ShardView) };
```
alchemist51, line 225: In Java, when we try to decRef an closed ptr. we general throw an exception. Should we do something similar here to signify an wrong action?

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/RecordBatchStream.java
bharath-techie: There is one in `org/opensearch/vectorized/execution/search/spi/RecordBatchStream.java` but we are not using it as its not concrete implementation. Maybe we should use that ?

bharath-techie:
Merging for now , there will be an iterative PR to incrementally make more changes to this.

PR #20014, opened 2025-11-16
Refactor HttpStats, ScriptStats, AdaptiveSelectionStats and OsStats with Builder pattern

pado0:
### Description
This PR refactors the `HttpStats`, `ScriptStats`, `AdaptiveSelectionStats`  and `OsStats` class to use the Builder pattern instead of relying on multiple constructors.  

By adopting the Builder pattern, it becomes easier to evolve the stats API, add new metrics, and maintain backward compatibility without forcing disruptive constructor changes.

Based on the related issue:
1. Added a Builder and deprecated the existing constructors.
2. Replaced usages of constructors in code and tests with the new Builder.

There are multiple stats-related classes that need similar refactoring, and we are addressing them in priority order. This PR covers `HttpStats`, `ScriptStats`, `AdaptiveSelectionStats`  and `OsStats`   as part of that effort.

### Related Issues
Related to #19225 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,8 +57,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Refactor the FieldDataStats and CompletionStats class to use the Builder pattern instead of constructors ([#19936](https://github.com/opensearch-project/OpenSearch/pull/19936))
 - Thread Context Preservation by gRPC Interceptor ([#19776](https://github.com/opensearch-project/OpenSearch/pull/19776))
 - Update NoOpResult constructors in the Engine to be public ([#19950](https://github.com/opensearch-project/OpenSearch/pull/19950))
+- Make XContentMapValues.filter case insensitive ([#19976](https://github.com/opensearch-project/OpenSearch/pull/19976))
 - Refactor the TranslogStats and RequestCacheStats class to use the Builder pattern instead of constructors ([#19961](https://github.com/opensearch-project/OpenSearch/pull/19961))
 - Refactor the IndexPressutreStats, DeviceStats and TransportStats class to use the Builder pattern instead of constructors ([#19991](https://github.com/opensearch-project/OpenSearch/pull/19991))
+- Refactor the Cache.CacheStats class to use the Builder pattern instead of constructors ([#20015](https://github.com/opensearch-project/OpenSearch/pull/20015))
 - Refactor the HttpStats, ScriptStats, AdaptiveSelectionStats and OsStats class to use the Builder pattern instead of constructors ([#20014](https://github.com/opensearch-project/OpenSearch/pull/20014))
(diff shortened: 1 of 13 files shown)


server/src/main/java/org/opensearch/script/ScriptStats.java
```diff
@@ -102,6 +129,30 @@ public void writeTo(StreamOutput out) throws IOException {
...
+            compilationLimitTriggered += stat.getCompilationLimitTriggered();
+        }
+
+        return new Builder().contextStats(Collections.unmodifiableList(new ArrayList<>(ctxStats)))
+            .compilations(compilations)
+            .cacheEvictions(cacheEvictions)
+            .compilationLimitTriggered(compilationLimitTriggered)
+            .build();
+    }
+
```
pado0, line 155: I moved the metric aggregation logic from the [original constructor](https://github.com/opensearch-project/OpenSearch/pull/20014/files#diff-75c086335e09e5e91cc2513fd256480aed3d115bf65d77b1f43b4472a9d2b93bL61) into a static factory method.
I initially considered delegating the aggregation to the [stats()](https://github.com/opensearch-project/OpenSearch/pull/20014/files#diff-bda471ea96de71c670c4d2b053e1e2d9fc07c38f5b5f290de13252025453dc75R806) method, but it became too verbose, so I decided to handle it this way instead.

Since it’s only used in that place, and a similar pattern exists in [ScriptCacheStats.java](https://github.com/opensearch-project/OpenSearch/pull/20014/files#diff-30b41f2a3931e3685ade56131ec749953603662d9df3b8e553843ff48351299eR157) where the aggregation logic also lives in the caller,
I think moving it into the stats() method could also make sense — curious to hear your thoughts on this part.

sandeshkr419: I think I like the idea of having a static util called `aggregate` exposed as you did in current change. Its a much cleaner and meaningful abstraction.

server/src/main/java/org/opensearch/script/ScriptStats.java
```diff
@@ -58,6 +58,23 @@ public class ScriptStats implements Writeable, ToXContentFragment {
...
+    private ScriptStats(Builder builder) {
+        this.contextStats = builder.contextStats;
+        this.compilations = builder.compilations;
+        this.cacheEvictions = builder.cacheEvictions;
+        this.compilationLimitTriggered = builder.compilationLimitTriggered;
+    }
+
+    /**
+     * This constructor will be deprecated starting in version 3.4.0.
+     * Use {@link Builder} instead.
```
sandeshkr419: Can you point to the new `aggregate` method you defined for accuracy.

sandeshkr419 (commented):
Minor comments, rest LGTM!

Thanks for working on this! :D

PR #20016, opened 2025-11-16
Added validation for parquet data fields

cocosz:
### Description

This PR adds validation to ensure that field types defined in index mappings are compatible with the Parquet data format. When Parquet format is enabled, the system will now reject field mappings that use data types not supported by Parquet during the mapping definition phase.

### Testing

1. Valid Parquet mapping

```
curl -X PUT "localhost:9200/test-allowed-types" -H "Content-Type: application/json" -d '{
  "mappings": {
    "properties": {
      "name": {"type": "text"},
      "category": {"type": "keyword"},
      "age": {"type": "integer"},
      "salary": {"type": "long"},
      "score": {"type": "float"},
      "rating": {"type": "double"},
      "active": {"type": "boolean"},
      "created_date": {"type": "date"},
      "ip_address": {"type": "ip"}
    }
  }
}'
```
Response : 
```
{"acknowledged":true,"shards_acknowledged":true,"index":"test-allowed-types"}% 
```
2. Invalid Parquet Mapping 

```
curl -X PUT "localhost:9200/test-geo-point" -H "Content-Type: application/json" -d '{
  "mappings": {
    "properties": {
      "name": {"type": "text"},
      "location": {"type": "geo_point"}
    }
  }
}'
```
Response :
```
{"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"Field [user_agent] of type [geo_point] is not supported by data format [parquet]; Field [client_ip] of type [geo_point] is not supported by data format [parquet]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: Field [user_agent] of type [geo_point] is not supported by data format [parquet]; Field [client_ip] of type [geo_point] is not supported by data format [parquet]","caused_by":{"type":"mapper_parsing_exception","reason":"Field [user_agent] of type [geo_point] is not supported by data format [parquet]; Field [client_ip] of type [geo_point] is not supported by data format [parquet]"}},"status":400}
```

### DIFF
--- a/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/engine/ParquetDataFormat.java
+++ b/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/engine/ParquetDataFormat.java
@@ -47,7 +47,7 @@ public class ParquetDataFormat implements DataFormat {
 
     @Override
     public boolean isDataFormatSupported(String fieldTypeName) {
-        return ArrowFieldRegistry.getRegisteredFieldNames().contains(fieldTypeName);
+        return ArrowFieldRegistry.isFieldTypeSupported(fieldTypeName);
     }
 
     public static ParquetDataFormat PARQUET_DATA_FORMAT = new ParquetDataFormat();
--- a/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/engine/ParquetDataFormat.java
+++ b/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/engine/ParquetDataFormat.java
@@ -1,5 +1,6 @@
 package com.parquet.parquetdataformat.engine;
 
+import com.parquet.parquetdataformat.fields.ArrowFieldRegistry;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.index.engine.exec.DataFormat;
@@ -42,6 +43,11 @@ public class ParquetDataFormat implements DataFormat {
     @Override
     public void configureStore() {
 
+    }
+
+    @Override
+    public boolean isDataFormatSupported(String fieldTypeName) {
+        return ArrowFieldRegistry.getRegisteredFieldNames().contains(fieldTypeName);
     }
 
     public static ParquetDataFormat PARQUET_DATA_FORMAT = new ParquetDataFormat();
--- a/server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
+++ b/server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
@@ -91,6 +91,7 @@ import org.opensearch.index.IndexSettings;
 import org.opensearch.index.compositeindex.CompositeIndexSettings;
 import org.opensearch.index.compositeindex.CompositeIndexValidator;
 import org.opensearch.index.compositeindex.datacube.startree.StarTreeIndexSettings;
+import org.opensearch.index.engine.exec.DataFormat;
 import org.opensearch.index.mapper.DocumentMapper;
 import org.opensearch.index.mapper.MapperService;
 import org.opensearch.index.mapper.MapperService.MergeReason;
--- a/server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
+++ b/server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
@@ -91,7 +91,6 @@ import org.opensearch.index.IndexSettings;
 import org.opensearch.index.compositeindex.CompositeIndexSettings;
 import org.opensearch.index.compositeindex.CompositeIndexValidator;
 import org.opensearch.index.compositeindex.datacube.startree.StarTreeIndexSettings;
-import org.opensearch.index.engine.exec.DataFormat;
 import org.opensearch.index.mapper.DocumentMapper;
 import org.opensearch.index.mapper.MapperService;
 import org.opensearch.index.mapper.MapperService.MergeReason;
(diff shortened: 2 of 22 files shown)


server/src/main/java/org/opensearch/index/mapper/ObjectMapper.java
```diff
@@ -572,6 +573,9 @@ protected static void parseProperties(ObjectMapper.Builder objBuilder, Map<Strin
                         }
                     }
 
+                    // Validate parquet data format compatibility
+                    validateParquetDataFormatCompatibility(type, fieldName);
```
mgodwan: Core should not be aware of parquet. We need to have a way for data formats to tell if certain field is compatible with them. For fields coming from core/plugins, the data format registry should be used as a source of truth there.

server/src/main/java/org/opensearch/index/mapper/ObjectMapper.java
```diff
@@ -606,7 +610,40 @@ protected static void parseProperties(ObjectMapper.Builder objBuilder, Map<Strin
...
+         * Checks if the given field type is compatible with Parquet data format.
+         */
+        private static boolean isParquetCompatibleType(String fieldType) {
+            return getAllowedParquetTypes().contains(fieldType);
+        }
+
+        /**
+         * Returns the set of field types that are allowed for Parquet data format.
+         */
+        private static Set<String> getAllowedParquetTypes() {
```
mgodwan: This should be coming from the data format plugin's implementation. We already have a method there. Could you see if you can extend the implementation there?

mgodwan (changes requested):

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -91,6 +91,7 @@
 import org.opensearch.index.compositeindex.CompositeIndexSettings;
 import org.opensearch.index.compositeindex.CompositeIndexValidator;
 import org.opensearch.index.compositeindex.datacube.startree.StarTreeIndexSettings;
+import org.opensearch.index.engine.exec.DataFormat;
```
darjisagar7: Unused Import

modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/engine/ParquetDataFormat.java
```diff
@@ -44,6 +45,11 @@ public void configureStore() {
 
     }
 
+    @Override
+    public boolean isDataFormatSupported(String fieldTypeName) {
+        return ArrowFieldRegistry.getRegisteredFieldNames().contains(fieldTypeName);
```
darjisagar7: Can we check if the field is present in the ArrowFieldRegistry and not fetch complete map here

server/src/main/java/org/opensearch/index/engine/exec/DataFormat.java
```diff
@@ -43,6 +45,11 @@ public String name() {
         public void configureStore() {
 
         }
+
+        @Override
+        public boolean isDataFormatSupported(String fieldTypeName) {
+            return true;
```
darjisagar7: For now we can make this false as lucene won't be supported.

server/src/main/java/org/opensearch/index/engine/exec/coord/Any.java
```diff
@@ -48,6 +48,11 @@ public List<DataFormat> getDataFormats() {
         return dataFormats;
     }
 
+    @Override
+    public boolean isDataFormatSupported(String fieldType) {
+        return dataFormats.stream().anyMatch(df -> df.isDataFormatSupported(fieldType));
```
darjisagar7: Why are we using anyMatch here, shouldn't it be matching with dataformat opted for that data type?

cocosz: We can actually just keep it as false since doesn't have any use anywhere.

server/src/main/java/org/opensearch/index/engine/exec/DataFormat.java
cocosz: Sure I will cover it in the next commit

modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/engine/ParquetDataFormat.java
cocosz: We can move the logic to ArrowFieldRegistry

server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
```diff
@@ -218,8 +222,14 @@ public T meta(Map<String, String> meta) {
     protected MultiFields multiFields;
     protected CopyTo copyTo;
     protected DerivedFieldGenerator derivedFieldGenerator;
+    protected Set<String> dataFormats;
 
     protected FieldMapper(String simpleName, FieldType fieldType, MappedFieldType mappedFieldType, MultiFields multiFields, CopyTo copyTo) {
+        // Hardcoded to validate all fields against "parquet" data format
+        this(simpleName, fieldType, mappedFieldType, multiFields, copyTo, Collections.singleton("parquet"));
```
darjisagar7, line 229: Little confused here, can you help clarify, here dataformats is taken as input and later in checkDataFormat the dataformats is taken from plugin service. Why do we need this?

server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
```diff
@@ -639,6 +650,47 @@ public void canDeriveSource() {
...
+        if (pluginsService == null) {
+            return;
+        }
+
+        if (this.dataFormats == null || this.dataFormats.isEmpty()) {
+            return;
+        }
+
+        Map<String, DataFormat> availableDataFormats = new HashMap<>();
+        for (DataFormatPlugin plugin : pluginsService.filterPlugins(DataFormatPlugin.class)) {
```
darjisagar7, line 665: Can we do this once and store it to reuse it later as this will not change once the opensearch server is up. Here you are using this code for every indexing call which is not required.

PR #20032, opened 2025-11-18
Injecting count sub agg if absent to ensure doc_count is correct

expani:
Equivalent SQL plugin Commit : https://github.com/vinaykpud/sql/commit/99bdc4869f46208ba76382d6a7447faf5bc934a3

The merging of InternalAggregation at the Coordinator heavily relies on `doc_count` parameter. 

Consider a PPL query that doesn't have count sub aggregation
```
source=hits | stats sum(EventID) by Browser
```

In this case, we can only mock the doc_count as the Datafusion query doesn't have the sub agg. 

Now, we are adding count sub agg in the SQL plugin if it's absent and the request contains aggregation.

In this PR, we are handling the same in core opensearch to populate `doc_count` correctly.

### DIFF
--- a/server/src/main/java/org/opensearch/search/SearchService.java
+++ b/server/src/main/java/org/opensearch/search/SearchService.java
@@ -1310,6 +1310,7 @@ public class SearchService extends AbstractLifecycleComponent implements IndexEv
             if (request.scroll() != null) {
                 context.scrollContext().scroll = request.scroll();
             }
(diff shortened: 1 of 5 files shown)


expani:
cc: @mch2 @bharath-techie

expani:
Still working on this. 
There are a few issues causing DataFusion E2E to fail, will update once ready for merging

expani:
@mch2 @bharath-techie Simplified the code a lot and handled populating `doc_count` 
please review

PR #20048, opened 2025-11-19
add stream and allocator closure post fetch phase

mch2:
Add missing stream and allocator closure post fetch phase.

### DIFF
--- a/plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
+++ b/plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
@@ -186,12 +186,13 @@ public class DatafusionEngine extends SearchExecEngine<DatafusionContext, Datafu
         return false;
     }
 
-    RootAllocator allocator = null;
-    RecordBatchStream stream = null;
+
     @Override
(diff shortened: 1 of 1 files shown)


plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
```diff
@@ -186,12 +186,13 @@ public boolean assertSearcherIsWarmedUp(String source, Engine.SearcherScope scop
...
     }
 
-    RootAllocator allocator = null;
-    RecordBatchStream stream = null;
+
     @Override
     public Map<String, Object[]> executeQueryPhase(DatafusionContext context) {
         Map<String, Object[]> finalRes = new HashMap<>();
         List<Long> rowIdResult = new ArrayList<>();
+        RootAllocator allocator = null;
```
expani, line 194: Found some discussions that this is Thread safe and can be created once https://www.mail-archive.com/issues@arrow.apache.org/msg75219.html

mch2: yeah working on some changes to make this a shared allocator

PR #20055, opened 2025-11-19
fix bug of warm index : FullFileCachedIndexInput was repeatedly closed

### LINKED ISSUE #20054: [BUG]  write warm index error because execption of  index file Already closed
### Describe the bug

when we writing a warm index，the index break whith execption：
Image

### Related component

Storage

### To Reproduce

1. config and start server：
```
## this feature is not release
opensearch.experimental.feature.writable_warm_index.enabled: true

node.search.cache.size: 1gb
cluster.filecache.remote_data_ratio: 5

s3.client.default.endpoint: **** # S3 has alternate endpoints, but you probably don't need to change this value.
s3.client.default.max_retries: 3 # number of retries if a request fails
s3.client.default.path_style_access: true # whether to use the deprecated path-style bucket URLs.
s3.client.default.protocol: http # http or https
s3.client.default.read_timeout: 50s # the S3 connection timeout
s3.client.default.use_throttle_retries: true # whether the client should wait a progressively longer amount of time (exponential backoff) between each successive retry
s3.client.default.region: **** # AWS region to use. For non-AWS S3 storage, this value is required but has no effect.

# Repository name
node.attr.remote_store.segment.repository: warm-repo-1
node.attr.remote_store.translog.repository: warm-repo-1
node.attr.remote_store.state.repository: warm-repo-1
# Segment repository settings
node.attr.remote_store.repository.warm-repo-1.type: s3
node.attr.remote_store.repository.warm-repo-1.settings.bucket: test-warm
#node.attr.remote_store.repository.warm-repo-1.settings.base_path: data1013
node.attr.remote_store.repository.warm-repo-1.settings.region: ****

cluster.remote_store.state.enabled: false #stores index metadata in the remote store
```
2. continuously writing data 
3. error：
```
org.apache.lucene.store.AlreadyClosedException: Already closed: MemorySegmentIndexInput(path="/opt/workspace/apps/test/ssnap/opensearch-3.4.0-SNAPSHOT/data/nodes/0/indices/XIZHZhb3R06MfB_vjsg1Rw/0/index/_2e_Lucene103_0.tip")
```

### Expected behavior

no error

### Additional Details

**Plugins**

**Screenshots**

**Additional context**

liuyonghengheng:
### Description
check closed  value in function  FullFileCachedIndexInput.IndexInputHolder.run()，prevent FileCachedIndexInput  closed repeatedly  and duplicate deref the  reference count of cache entry

### Related Issues
Resolves #20054  https://github.com/opensearch-project/OpenSearch/issues/20054

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved warm index caching lifecycle so cached index inputs and their clones are cleaned up reliably and closing is idempotent, preventing premature or repeated closure.

* **Tests**
  * Added end-to-end lifecycle and GC-related tests to validate clone/slice closing and cleanup behavior.

* **Documentation**
  * Added a changelog entry describing the warm index caching fix.


server/src/main/java/org/opensearch/index/store/remote/filecache/FileCachedIndexInput.java
```diff
@@ -139,7 +140,7 @@ public IndexInput slice(String sliceDescription, long offset, long length) throw
 
     @Override
     public void close() throws IOException {
-        if (!closed) {
+        if (!closed.get()) {
```
rayshrey, line 143: Nit - can we use `compareAndSet` for all such operations

liuyonghengheng: Thank you, We should consider that exceptions are thrown when call luceneIndexInput.close() ，then we can not  set the value to true？

server/src/main/java/org/opensearch/index/store/remote/filecache/FullFileCachedIndexInput.java
```diff
@@ -113,8 +116,11 @@ private static class IndexInputHolder implements Runnable {
         @Override
         public void run() {
             try {
-                indexInput.close();
-                if (isClone) cache.decRef(path);
+                if (!closed.get()) {
+                    indexInput.close();
+                    if (isClone) cache.decRef(path);
+                    closed.set(true);
```
himkak, line 129: is it required to set closed to true here, as during the invocation of indexInput.close the closed would already be set to true

server/src/main/java/org/opensearch/index/store/remote/filecache/FileCachedIndexInput.java
rayshrey: In that case can we put the `luceneIndexInput.close()` inside a try-catch block and then reset the closed variable back to false and rethrow the error ?

The problem with the current code is that it does not seem thread safe, if multiple concurrent close calls come then we can again run into the same issue of mutiple decRefs. (although it's very unlikely that multiple concurrent close calls will come, but still let's ensure we handle it)

server/src/main/java/org/opensearch/index/store/remote/filecache/FullFileCachedIndexInput.java
rayshrey: It's not mandatory that the close method will always be called(especially for clones). That is the main reason we had added this CLEANER fallback, but we failed to take into account cases where close was also being called and our cleaner also did the decRef. This is being fixed now with this PR, thanks to @liuyonghengheng  

@liuyonghengheng Can we also add a simple UT where we call close and run method of this IndexInputHolder to assert the refCounting ?

server/src/main/java/org/opensearch/index/store/remote/filecache/FileCachedIndexInput.java
liuyonghengheng: >  (although it's very unlikely that multiple concurrent close calls will come, but still let's ensure we handle it)

I aggre with you，I didn't consider thread safe  because I assumed that IndexInputHolder.run is invoked only in a GC context, whereas FullFileCachedIndexInput.close is called explicitly only by the IndexReader (I'm not entirely sure—if this is indeed , then concurrent invocation between IndexInputHolder.run and FullFileCachedIndexInput.close would be impossible). But if we  need thread safe，we should add synchronized block or lock，because if thread-1 called closed.compareAndSet  ，and then thead-2 will assume that luceneIndexInput has been closed and skip the code block ，then if thread-1 throw exception ，but may be thead-2 can close luceneIndexInput  successfully，so we should add lock or synchronized block  to ensuer the block will not be skiped

liuyonghengheng: can we depend on the AlreadyClosedException  of luceneIndexInput.close()?  if indexInput.close() throw exception( I assume that when input has been closed it will allways throw AlreadyClosedException)  then skip cache.decRef.
```
        @Override
        public void run() {
            try {
                if (!closed.get()) {
                    indexInput.close();
                    if (isClone) cache.decRef(path);
                    closed.set(true);
                }
            } catch (IOException e) {
                logger.error("Failed to close IndexInput while clearing phantom reachable object");
            }
        }
```

```
    @Override
    public void close() throws IOException {
        if (!closed.get()) {
            try {
                luceneIndexInput.close();
                if (isClone) {
                    cache.decRef(filePath);
                }
            } catch (AlreadyClosedException e) {
                logger.trace("FullFileCachedIndexInput already closed");
            }
            luceneIndexInput = null;
            closed.set(true);
        }
    }
```

server/src/main/java/org/opensearch/index/store/remote/filecache/FullFileCachedIndexInput.java
liuyonghengheng: OK, I will add the UT

server/src/main/java/org/opensearch/index/store/remote/filecache/FileCachedIndexInput.java
rayshrey: > can we depend on the AlreadyClosedException of luceneIndexInput.close()

I don't think it will work as we may get a NullPointerException as well since we are explicitly setting the luceneIndexInput as null in next line.

---

On second thoughts, I don't think we need to handle thread safety as [IndexInput by contract is not thread safe](https://github.com/apache/lucene/blob/cb9f2974df61e599759089ce4b208b64ddb3a554/lucene/core/src/java/org/apache/lucene/store/IndexInput.java#L28), the only race can be between cleaner's run() and index input's close() methods but that also will not happen as run() will be executed only if instance is phantom reachable. So I think we are good on that part. Making any additional changes will be over-engineering in my opinion. And apologies as I only seem to have started the thread safety discussion :(

Let's just add an UT and close this PR.

liuyonghengheng: @rayshrey  I have added the UT，please review it

rayshrey:
@gbbafna can you please review and merge this if it looks good

gbbafna:
Changes LGTM . Thanks @liuyonghengheng for this and @rayshrey for the review. Will merge once the build passes .

gbbafna (approved):

PR #19573, opened 2025-10-08
Combining filter rewrite and skip list to optimize sub aggregation

jainankitk:
### Description
This change aims at optimizing the sub aggregation by leveraging multi range traversal for top level aggregation, and skip list for the sub aggregation. This optimization helps improve the performance for such aggregation requests upto 10x (needs sorting on field for which skip list is utilized). Queries `range-auto-date-histo` and `range-auto-date-histo-metrics` from the `big5` benchmark workload should directly benefit from this optimization once change is extended for auto histogram aggregations as well - https://github.com/opensearch-project/OpenSearch/pull/20057 

~I have copied over the `BitSetDocIdStream` class in OpenSearch from Lucene for now as it is not public, but should look at eventually getting rid of it.~ Copying over `BitSetDocIdStream` was resulting in 3 implementations of `DocIdStream` making some of the calls polymorphic. Hence, instead added helper class to initialize `BitSetDocIdStream` in the lucene namespace. We have followed the same pattern for few other extensions in Lucene

### Related Issues
Related to #17447, #19384

### Check List
- [x] Functionality includes testing.
- ~[ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.~
- ~[ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

## Summary by CodeRabbit

* **New Features**
  * Enhanced aggregation sub-processing with combined optimization strategies for improved performance in histogram and range aggregations.

* **Tests**
  * Added test coverage for skiplist-optimized collection with range and date histogram sub-aggregations across multiple segments.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

Issue #19384, opened 2025-09-23
[META] Skip List Based Optimization for Aggregations in OpenSearch


jainankitk:
### Motivation ###

Efficient aggregations are critical for OpenSearch use cases involving analytics, dashboards, and exploratory queries. Recent work such as *Filter Rewrite* and *Multi Range Traversal* have shown orders-of-magnitude improvements for specific cases like `date_histogram`. However, these optimizations are limited to scenarios where the aggregation field aligns closely with query filters and document structures.

This RFC proposes a new optimization leveraging **Lucene’s skip index on doc values** (introduced in Lucene 10) to accelerate aggregations even when:
* Filters apply on fields other than the aggregation field
* Sub-aggregations are involved
* Segments contain deleted documents

Our approach achieves up to **2x improvements** across broader workloads than Multi Range Traversal, by enabling efficient doc ID skipping and population counting using CPU-level instructions.

### Skip Index in Lucene ###

<img width="2520" height="980" alt="Image" src="https://github.com/user-attachments/assets/dd8868fd-14f5-4289-989f-8a6b2105205c" />

* Built on top of doc values, conceptually similar to a skip list.
* Maintains min/max values per interval at multiple levels (e.g., skipping every 5, 25, 125 documents).
* Enables efficient skipping of doc IDs when combined with sorted doc values.

### Problem Statement ###

Given a query that:
* Filters on one field (`trip_distance: [0, 50]`)
* Aggregates on another (`dropoff_datetime` in a monthly `date_histogram`)

Existing Multi Range Traversal cannot optimize this case because the filter field (`trip_distance`) has no correlation with the aggregation field (`dropoff_datetime`).

We need a mechanism to **bridge filter doc ID sets with sorted aggregation fields** efficiently.

#### Core Idea ####
Overlay the **doc ID set iterator** (from the filter field) with the **skip index** of the aggregation field, enabling block-level skipping and efficient counting.

#### Execution Flow ####
1. Query execution produces a **doc ID set iterator** for the filter (e.g., `trip_distance: [0,50]`).
2. Retrieve the **skip index** for the aggregation field (`dropoff_datetime`).
3. Align the doc ID set with skip intervals:
  * If all doc IDs in an interval fall into the same aggregation bucket (e.g., a single month in `date_histogram`), skip iteration over individual doc IDs.
  * Instead, count them in bulk using **bitset operations**.
[...]

Issue #18882, opened 2025-07-31
[META] Skiplist plan for 3.2


asimmahmood1:
## Please describe the end goal of this project

As follow up to skiplist meta https://github.com/opensearch-project/OpenSearch/issues/17967, this issue captures the introduction of skiplist in OS 3.2.

There are few points to consider:
1. Should skiplist be enabled by default? 
This depends on index size impact. Enabling skiplist on just timestamp field shows a 0.5% increase in bytes, which is very minimal. https://github.com/opensearch-project/OpenSearch/issues/17710#issuecomment-2801746980. Another experiment on enabled for all numeric fields in big5 bechmark shows [63%, 22 to 36gb] https://github.com/opensearch-project/OpenSearch/issues/18751

2. What is the performance impact?
Some numbers are in https://github.com/opensearch-project/OpenSearch/issues/17710 with synthenic data. big5 and http_logs benchmarks are in https://github.com/opensearch-project/OpenSearch/issues/18751 [Summary TBD]

Based on above

# Plan for 3.2 is to: 

1. Run benchmarks and find speed up in existing queries:  https://github.com/opensearch-project/OpenSearch/issues/18751
1. Add skiplist mapping option with default false:  https://github.com/opensearch-project/OpenSearch/issues/17965

## Follow up

- [ ] https://github.com/opensearch-project/OpenSearch/issues/17964

PR #20057, opened 2025-11-19
Add skiplist optimization to auto_date_histogram aggregation

asimmahmood1:
### Description
This is a built on top https://github.com/opensearch-project/OpenSearch/pull/19573. 

Compare to date histogram skiplist, this change needs to hook into dynamic rounding during collection. There are 2 variables to keep track of:

* `bucketOrds` - seen rounded dates
* `preparedRounding` - starts at lowest interval: MINUTES and goes up

When a new ord is created, `increaseRoundingIfNeeded` function is called to determine if new `preparedRounding` needs to kick in (e.g. from HOURS to DAYS), and may also merge dates in `bucketOrds`. Thus, both are need to be *supplied* via lambda. 

In the future skiplist can be enhanced to keep track of multiple `owningBucketOrd`, for now it only works when auto date histogram is root (parent == null), or within range filter rewrite context that guarantees new auto date histogram is created per range. 

 

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/19827
Part of https://github.com/opensearch-project/OpenSearch/issues/18882
Also https://github.com/opensearch-project/OpenSearch/issues/19384

### Check List
- [n/a] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [TODO] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

## Summary by CodeRabbit

* **New Features**
  * Skiplist optimization for auto_date_histogram to speed up bucketing.
  * Public helper to evaluate skiplist eligibility at runtime.

* **Improvements**
  * Dynamic rounding support in histogram aggregations to handle runtime interval changes.
  * Performance gains and added debug counters when the skiplist path is used.

* **Tests**
  * New tests covering skiplist behavior, rounding-change scenarios, bucket merging, and sub-aggregation equivalence.

### DIFF
--- a/server/src/main/java/org/opensearch/search/aggregations/bucket/histogram/AutoDateHistogramAggregator.java
+++ b/server/src/main/java/org/opensearch/search/aggregations/bucket/histogram/AutoDateHistogramAggregator.java
@@ -280,23 +280,7 @@ abstract class AutoDateHistogramAggregator extends DeferableBucketAggregator {
             skipper = ctx.reader().getDocValuesSkipper(this.fieldName);
         }
 
-        final LeafBucketCollector iteratingCollector = getLeafCollector(values, skipper, sub);
-        return new LeafBucketCollectorBase(sub, values) {
-            @Override
-            public void collect(int doc, long owningBucketOrd) throws IOException {
-                iteratingCollector.collect(doc, owningBucketOrd);
-            }
-
-            @Override
-            public void collect(DocIdStream stream, long owningBucketOrd) throws IOException {
-                iteratingCollector.collect(stream, owningBucketOrd);
-            }
-
-            @Override
-            public void collectRange(int min, int max) throws IOException {
-                iteratingCollector.collectRange(min, max);
-            }
-        };
+        return getLeafCollector(values, skipper, sub);
     }
 
     protected final InternalAggregation[] buildAggregations(


asimmahmood1:
Functionally correct but not showing improvement

```
diff 17_big5_auto_date_filter_baseline.json 17_big5_auto_date_filter_candidate.json
2c2
<   "took": 114,
---
>   "took": 214,
```

### Query
```
curl -X POST "http://localhost:9200/big5/_search"     -H "Content-Type: application/json"     -d '{
      "size": 0,
      "query": {
        "bool": {
          "must": [
            {
              "term": {
                "process.name": "systemd"
              }
            }
          ]
        }
      },
      "aggs": {
        "by_hour": {
          "auto_date_histogram": {
            "field": "@timestamp",
            "buckets": 3
          }
        }
      }
    }'
```

### Result 
```
{
  "took": 127,
  "timed_out": false,
  "terminated_early": true,
  "_shards": {
    "total": 1,
    "successful": 1,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": {
      "value": 10000,
      "relation": "gte"
    },
    "max_score": null,
    "hits": []
  },
  "aggregations": {
    "by_hour": {
      "buckets": [
        {
          "key_as_string": "2023-01-01T00:00:00.000Z",
          "key": 1672531200000,
          "doc_count": 2488712
        },
        {
          "key_as_string": "2023-01-08T00:00:00.000Z",
          "key": 1673136000000,
          "doc_count": 824998
        }
      ],
      "interval": "7d"
    }
  }
}
```

asimmahmood1:
[auto-date-agg-baseline.html](https://github.com/user-attachments/files/23636137/auto-date-agg-baseline.html)
[auto-date-agg-skiplist.html](https://github.com/user-attachments/files/23636139/auto-date-agg-skiplist.html)

jainankitk:
@asimmahmood1 - Have you verified that the regression is only with `AutoDateHistogramAggregator` and not when using `DateHistogramAggregator` for same query?

asimmahmood1:
So I figured out why performance was not up to par. Short answer is when auto date moves onto large intervals, we need to track of not just preparedRounding but also bucketOrds. If preparedRounding has changed since last time, we need to restart the skiplist logic. Otherwise, we'll collect too many docs, and although the end result doesn't change (i.e. unit test passes), performance is too low.

Auto date histo has two modes: FromSingle and FromMany. FromSingle is often used in top level aggregation, so It's similar to Date Histogram where parent is null. FromMany is used e.g in big5's `range-auto-date-histo`, which would normally handle interleaving owningBucketOrd. In the special case where filter rewrite logic is used, then we can safely assume that only 1 owningBucketOrd will be called per leaf collector, thus we can use skiplist histogram. 

See validation below. 

Note: will update this PR after https://github.com/opensearch-project/OpenSearch/pull/19573 is merged.

asimmahmood1:
## range auto date: from 1335 to 139 (89%)


```
11_range_auto_date_histo.sh
#!/bin/bash

curl -XGET 'http://localhost:9200/big5/_search' \
-d '{
  "size": 0, "profile": false,
  "aggs": {
    "tmax": {
      "range": {
        "field": "metrics.size",
        "ranges": [
          {
            "to": -10
          },
          {
            "from": -10,
            "to": 10
          },
          {
            "from": 10,
            "to": 100
          },
          {
            "from": 100,
            "to": 1000
          },
          {
            "from": 1000,
            "to": 2000
          },
          {
            "from": 2000
          }
        ]
      },
      "aggs": {
        "date": {
          "auto_date_histogram": {
            "field": "@timestamp",
            "buckets": 20
          }
        }
      }
    }
  }
}'

```


```
diff 11_range_auto_date_histo_candidate.json 11_range_auto_date_histo_baseline.json
2c2
<   "took": 139,
---
>   "took": 1335,

```

asimmahmood1:
## range-auto-date-with-metrics (22% lower)

this is similar to the date-with-metrics, since the time is bounded by tavg stat

```
[ec2-user@ip-172-31-61-197 ~]$ diff 11_range_auto_date_histo_with_metrics_candidate.json 11_range_auto_date_histo_with_metrics_baseline.json
2c2
<   "took": 2920,
---
>   "took": 3781,
```

asimmahmood1:

asimmahmood1:
{"run-benchmark-test": "id_11"}

asimmahmood1:
@coderabbitai review

asimmahmood1:
### Valdiation

```
[ec2-user@ip-172-31-61-197 ~]$ diff 11_range_auto_date_histo_baseline.json 11_range_auto_date_histo_candidate2.json
2c2
<   "took": 1335,
---
>   "took": 136,
```


```
[ec2-user@ip-172-31-61-197 ~]$ diff 11_range_auto_date_histo_with_metrics_baseline.json 11_range_auto_date_histo_with_metrics_candidate2.json
2c2
<   "took": 3781,
---
>   "took": 2956,

```

server/src/test/java/org/opensearch/search/aggregations/bucket/histogram/AutoDateHistogramAggregatorTests.java
```diff
@@ -1052,6 +1070,151 @@ private Map<String, Double> maxAsMap(InternalAutoDateHistogram result) {
...
+            // Verify total doc count matches input
+            long totalDocs = histogram.getBuckets().stream().mapToLong(Histogram.Bucket::getDocCount).sum();
+            assertEquals(54, totalDocs); // 24 + 30 = 54 docs
+
+            // Verify buckets are in ascending order (requirement for histogram)
+            List<? extends Histogram.Bucket> buckets = histogram.getBuckets();
+            for (int i = 1; i < buckets.size(); i++) {
+                assertTrue(((ZonedDateTime) buckets.get(i - 1).getKey()).isBefore((ZonedDateTime) buckets.get(i).getKey()));
+            }
+        });
```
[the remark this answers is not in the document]
asimmahmood1: `SortedNumericDocValuesField.indexedField` actually creates skiplist, sorted is an additional benefit but not neccessary.

asimmahmood1:
{"run-benchmark-test": "id_11"}

asimmahmood1:

asimmahmood1:
{"run-benchmark-test": "id_11"}

asimmahmood1:

asimmahmood1:
benchmarks failed with insufficient ec2:

```
opensearch-infra-stack-5271 | 6:12:38 PM | CREATE_FAILED        | AWS::EC2::Instance                        | single-node-instance (singlenodeinstance5DE833AF133b4ae8cbb89594) Resource handler returned message: "We currently do not have sufficient r5.xlarge capacity in the Availability Zone you requested (us-east-1a). Our system will be working on provisioning additional capacity. You can currently get r5.xlarge capacity by not specifying an Availability Zone in your request or choosing us-east-1b, us-east-1c, us-east-1d, us-east-1f. (Service: Ec2, Status Code: 500, Request ID: df795b2f-5086-4a1a-8b8f-26e5fcdbe1cf) (SDK Attempt Count: 5)" (RequestToken: 36f81b51-c440-6f04-bc9f-0dc97485ee5e, HandlerErrorCode: GeneralServiceException)
```

asimmahmood1:
{"run-benchmark-test": "id_11"}

asimmahmood1:

asimmahmood1:
## Service time p90 improvement: range-auto-date-histo: 90%; range-auto-date-histo-with-metrics: 21%

from https://github.com/opensearch-project/OpenSearch/pull/20057#issuecomment-3614912554


Metric | Task | Baseline | Contender | Diff | Unit
-- | -- | -- | -- | -- | --
50th percentile service time | range-auto-date-histo | 3435.59 | 347.067 | -3088.52 | ms
90th percentile service time | range-auto-date-histo | 3503.87 | 385.106 | -3118.77 | ms
99th percentile service time | range-auto-date-histo | 3559.24 | 419.822 | -3139.41 | ms
100th percentile service time | range-auto-date-histo | 3582.41 | 430.952 | -3151.46 | ms
50th percentile service time | range-auto-date-histo-with-metrics | 9816.78 | 6973.65 | -2843.13 | ms
90th percentile service time | range-auto-date-histo-with-metrics | 9918.76 | 7051.46 | -2867.3 | ms
99th percentile service time | range-auto-date-histo-with-metrics | 10062.4 | 7197.56 | -2864.88 | ms
100th percentile service time | range-auto-date-histo-with-metrics | 10067.6 | 7239.29 | -2828.31 | ms

server/src/test/java/org/opensearch/search/aggregations/bucket/histogram/AutoDateHistogramAggregatorTests.java
```diff
@@ -95,7 +95,7 @@
 import static org.hamcrest.Matchers.hasSize;
 
 public class AutoDateHistogramAggregatorTests extends DateHistogramAggregatorTestCase {
-    private static final String DATE_FIELD = "date";
```
bowenlan-amzn, line 98: Could add a comment explain why need to use `@timestamp` specifically

server/src/main/java/org/opensearch/search/aggregations/bucket/HistogramSkiplistLeafCollector.java
```diff
@@ -13,25 +13,39 @@
...
 import org.opensearch.search.aggregations.bucket.terms.LongKeyedBucketOrds;
 
 import java.io.IOException;
+import java.util.function.LongFunction;
+import java.util.function.Supplier;
 
 /**
  * Histogram collection logic using skip list.
  *
+ * Currently, it can only handle one owningBucketOrd at a time.
```
bowenlan-amzn, line 29: Could be better to save the owningBucketOrd into a SetOnce reference

server/src/main/java/org/opensearch/search/aggregations/bucket/HistogramSkiplistLeafCollector.java
```diff
@@ -136,15 +188,14 @@ public void collect(DocIdStream stream) throws IOException {
...
-        // This will only be called if its the sub aggregation
         for (;;) {
             int upToExclusive = upToInclusive + 1;
             if (upToExclusive < 0) { // overflow
                 upToExclusive = Integer.MAX_VALUE;
             }
 
             if (upToSameBucket) {
-                if (sub == NO_OP_COLLECTOR) {
+                if (isSubNoOp) {
```
bowenlan-amzn, line 198: Below line 204, does it make sense to use `collect(DocIdStream stream, long owningBucketOrd)` api instead of iterating over docs?

server/src/test/java/org/opensearch/search/aggregations/bucket/histogram/AutoDateHistogramAggregatorTests.java
```diff
@@ -1019,11 +1029,19 @@ private void testSearchCase(
...
+        indexSampleData(dataset, indexWriter, false);
+    }
+
+    private void indexSampleData(List<ZonedDateTime> dataset, RandomIndexWriter indexWriter, boolean enableSkiplist) throws IOException {
         final Document document = new Document();
         int i = 0;
         for (final ZonedDateTime date : dataset) {
             final long instant = date.toInstant().toEpochMilli();
-            document.add(new SortedNumericDocValuesField(DATE_FIELD, instant));
+            if (enableSkiplist) {
```
bowenlan-amzn, line 1044: Is it possible have a test that compares the results of skiplist vs non-skiplist on some random data?

asimmahmood1: Good idea, I actually just did that for min agg: https://github.com/opensearch-project/OpenSearch/pull/20175

Let me try for this one as well.

server/src/main/java/org/opensearch/search/aggregations/bucket/HistogramSkiplistLeafCollector.java
asimmahmood1: That is a good idea. The current DocIdStream api doesn't way to specify `upTo`, e.g.

```
sub.collect(steam.copy(upToExclusive), upToBucketIndex)
```
Plus we still need to count the them and stream is a read once. Will need to think about this more.

server/src/main/java/org/opensearch/search/aggregations/bucket/HistogramSkiplistLeafCollector.java
asimmahmood1: I plan to extend to use multiple bucket, like in https://github.com/opensearch-project/OpenSearch/pull/20175.

server/src/main/java/org/opensearch/search/aggregations/bucket/histogram/AutoDateHistogramAggregator.java
```diff
@@ -262,7 +275,12 @@ public final LeafBucketCollector getLeafCollector(LeafReaderContext ctx, LeafBuc
         }
 
         final SortedNumericDocValues values = valuesSource.longValues(ctx);
-        final LeafBucketCollector iteratingCollector = getLeafCollector(values, sub);
+        DocValuesSkipper skipper = null;
+        if (this.fieldName != null) {
+            skipper = ctx.reader().getDocValuesSkipper(this.fieldName);
+        }
+
+        final LeafBucketCollector iteratingCollector = getLeafCollector(values, skipper, sub);
```
jainankitk: Can we not return `iteratingCollector` directly now? Why do we need to wrap it?

asimmahmood1: Good point, I don't why it was ever wrapped.

Issue #19843, opened 2025-10-31
[BUG] Modification to cluster state object has backward compatiblity issue for remote publication


SwethaGuptha:
### Describe the bug

In transport cluster state publication since active cluster manager sends cluster state updates individually to all nodes in the cluster, when new fields are introduced to cluster state it is easier to handle backward compatibility by serializing fields based on the version of the node receiving the update and hence different nodes in will receive the version of the cluster state that they understand. Whereas, in remote publication the cluster state serialized only once and uploaded to remote, which thereby limits the active cluster manager from creating different versions of the cluster state for the different node versions and can fail the remote cluster download in such cases because of unread data.

We need to explore and introduce a mechanism for maintaining the backward compatibility that allows ClusterState class modification in remote publication.

### Related component

Cluster Manager

### To Reproduce

N/A

### Expected behavior

Support for making modifications to ClusterState class with backward compatibility in remote publication mode.

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross) [3](https://github.com/cwperks)

PR #16505, opened 2024-10-28
[Remote Publication] Add rolling upgrade bwc with remote publication enabled

shiv0408:
### Description
Add bwc test in rolling upgrade with remote publication enabled.

### Check List
- [ ] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #20080, opened 2025-11-24
Adding version checks to remote entities using bytestream ser/de

gargharsh3134:
### Description
RemotePublication entities which use bytestream for ser/de of clusterState stored in remote were not setting Opensearch version in the bytestream. This was leading to failures during version upgrades whenever a new attribute was getting added to any of those remoteState entities as the new nodes in the latest version were failing to join the existing older version nodes while trying to deserialize the state which did not have new attributes.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/19843

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * Version-aware serialization/deserialization for remote cluster state artifacts, fixing version incompatibility and ensuring remote reads/writes use manifest version.
  * Logs informational messages when reading manifests uploaded with non-current versions for clearer diagnostics.

* **Tests**
  * Tests updated to exercise versioned read/write paths and validate behavior across multiple OpenSearch versions.

* **Documentation**
  * CHANGELOG updated with the fix.


server/src/main/java/org/opensearch/gateway/remote/model/RemoteDiscoveryNodes.java
```diff
@@ -49,11 +47,16 @@ public RemoteDiscoveryNodes(
...
         this.discoveryNodes = discoveryNodes;
         this.stateVersion = stateVersion;
+        this.discoveryNodesFormat = new ChecksumWritableBlobStoreFormat<>("nodes", is -> DiscoveryNodes.readFrom(is, null));
     }
 
-    public RemoteDiscoveryNodes(final String blobName, final String clusterUUID, final Compressor compressor) {
+    public RemoteDiscoveryNodes(final String blobName, final String clusterUUID, final Compressor compressor, final Version version) {
         super(clusterUUID, compressor, null);
         this.blobName = blobName;
+        this.discoveryNodesFormat = new ChecksumWritableBlobStoreFormat<>("nodes", is -> {
```
rajiv-kv: Should the version be field in `ChecksumWritableBlobStoreFormat` and exposed by getter/setter. We can default to Version.Current.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateService.java
```diff
@@ -1296,7 +1296,8 @@ ClusterState readClusterStateInParallel(
             remoteRoutingTableService.getAsyncIndexRoutingReadAction(
                 clusterUUID,
                 indexRouting.getUploadedFilename(),
-                routingTableLatchedActionListener
+                routingTableLatchedActionListener,
+                manifest.getOpensearchVersion()
```
rajiv-kv, line 1309: Can we emit WARN log if the version is not equal to Version.current ?

server/src/main/java/org/opensearch/gateway/remote/model/RemoteClusterStateCustoms.java
```diff
@@ -65,16 +66,17 @@ public RemoteClusterStateCustoms(
...
         super(clusterUUID, compressor, null);
         this.blobName = blobName;
         this.customType = customType;
         this.namedWriteableRegistry = namedWriteableRegistry;
-        this.clusterStateCustomsFormat = new ChecksumWritableBlobStoreFormat<>(
-            "cluster-state-custom",
-            is -> readFrom(is, namedWriteableRegistry, customType)
-        );
+        this.clusterStateCustomsFormat = new ChecksumWritableBlobStoreFormat<>("cluster-state-custom", is -> {
+            is.setVersion(version);
```
rajiv-kv: can the version value be null ? What happens if the version int is greater than Version.Current ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateService.java
gargharsh3134: Added info log at the start of the method if version from manifest doesn't match the current version.

server/src/main/java/org/opensearch/gateway/remote/model/RemoteClusterStateCustoms.java
gargharsh3134: We always upload the manifest with `opensearchVersion` set as VERSION.CURRENT. During deserialization as well, if a node on older version tries to construct a new version from manifest file, that too will succeed. If the new version is absent from the list of versions on the old node, it will construct a new version object with the ID received from manifest.

server/src/main/java/org/opensearch/gateway/remote/model/RemoteDiscoveryNodes.java
gargharsh3134: Moved version attribute to `ChecksumWritableBlobStoreFormat`. As of now, it doesn't require getters and setters, so just added a private member variable.

server/src/main/java/org/opensearch/repositories/blobstore/ChecksumWritableBlobStoreFormat.java
```diff
@@ -102,6 +108,7 @@ public T deserialize(String blobName, BytesReference bytes) throws IOException {
             BytesReference bytesReference = bytes.slice((int) filePointer, (int) contentSize);
             Compressor compressor = CompressorRegistry.compressorForWritable(bytesReference);
             try (StreamInput in = new InputStreamStreamInput(compressor.threadLocalInputStream(bytesReference.streamInput()))) {
+                in.setVersion(opensearchVersion);
```
shwetathareja, line 117: this deserialize might be used for other flows other than remote state, could there be implication of setting OpenSearch version here?

server/src/test/java/org/opensearch/cluster/routing/remote/RemoteRoutingTableServiceTests.java
```diff
@@ -608,7 +621,8 @@ public void testGetAsyncIndexRoutingTableDiffReadAction() throws Exception {
         remoteRoutingTableService.getAsyncIndexRoutingTableDiffReadAction(
             "cluster-uuid",
             uploadedFileName,
-            new LatchedActionListener<>(listener, latch)
+            new LatchedActionListener<>(listener, latch),
+            Version.CURRENT
```
shwetathareja: should we have test with version other than current?

server/src/main/java/org/opensearch/repositories/blobstore/ChecksumWritableBlobStoreFormat.java
gargharsh3134: I checked usages of this non-static method, and currently its just being used by RemoteState entities. In the earlier revision, we did try setting this version in the remote state entity itself, but then decided to move it here instead. It helps to avoid missing out on setting this version in future for any new entity which might get introduced.

As of now, setting the version here seems like a reasonable choice. I have added a javadoc for this new member in the revision, explaining its requirement.

Please let me know if you still feel this might cause confusion.

server/src/test/java/org/opensearch/cluster/routing/remote/RemoteRoutingTableServiceTests.java
gargharsh3134: Modified this test to include mixed versions. However, since no incompatible attribute has been introduced for most of the entities it doesn't really help. I have added another UT `testSerializationOnOldVersionDeserializationOnNewVersions` for `RemoteDiscoveryNodes`, which had a new attribute causing failures. 

End to end solution to proactively catch these incompatibility issues is being tracked by this PR: https://github.com/opensearch-project/OpenSearch/pull/16505. I have already tested this locally and established that the IT fails while upgrading from 3.1 to 3.4 and passes with the current fix present. Tracking revising and merging of that PR independent of this current change.

Hope this addresses the concern, please let me know if otherwise.

shwetathareja (approved):

Issue #20004, opened 2025-11-14
[Feature Request] Forbid the use of Thread.sleep in tests


andrross:
### Is your feature request related to a problem? Please describe

Using a fixed sleep in tests is a great way to introduce a flaky test. It's almost always better to poll for the condition you're waiting for rather than waiting a fixed amount of time. Here are a handful of flaky tests from the past that were caused by fixed sleeps:

- #19422 
- #14288
- #18820

This list is certainly not exhaustive.

### Describe the solution you'd like

Use forbidden APIs to forbid the usage of `Thread#sleep` in test code. There are certainly cases where sleeping will still be needed so this can be overridden with the `@SuppressForbidden` annotation, but making the default behavior to fail the build will make it clear to contributors and reviewers that any usage of sleep needs scrutiny.

### Additional context

Because usage of Thread.sleep is currently quite pervasive, my rollout plan for this is to chip away one at a time by making it an optional property that can be turned on one test suite at a time. For example, the following can be added to `server/build.gradle` to enable this for internal cluster tests:

```
tasks.named('forbiddenApisInternalClusterTest').configure { forbidSleep() }
```

This will avoid a huge PR to annotate or remove hundreds of usages across all the projects. Once we reach a critical mass we can switch the default in this repo.

andrross:
This request is inspired by similar work that happened in Lucene: https://github.com/apache/lucene/issues/12946

andrross:
My plan is to submit a PR that forbids the use of sleep in `:server:internalClusterTest` as the first step. There are 70+ usages of sleep that have to be annotated or fixed so it's taking me a bit to chip away at that. Here's the branch I'm working on: https://github.com/andrross/OpenSearch/tree/forbid-sleep

PR #20088, opened 2025-11-24
Forbid use of Thread.sleep in :server:internalClusterTest

andrross:
### Description
This commit introduces enforcement to cause any future usage of Thread.sleep within :server:internalClusterTest to fail by default. It can still be overridden when necessary by using `@SuppressForbidden`.

For the existing usages of Thread.sleep(), I have either annotated them with some sort of rationale, replaced them with a polling alternative, or just removed them all together. I ran `./gradlew :server:internalClusterTest` 100 times over the weekend, and while it failed nearly half the time, none of the tests modified here were implicated.

### Related Issues
Related to #20004

### Check List

### DIFF
--- a/server/src/internalClusterTest/java/org/opensearch/search/sort/FieldSortIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/search/sort/FieldSortIT.java
@@ -47,6 +47,7 @@ import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.search.ShardSearchFailure;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.common.Numbers;
+import org.opensearch.common.SuppressForbidden;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.xcontent.XContentFactory;
 import org.opensearch.common.xcontent.XContentType;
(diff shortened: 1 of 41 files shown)


server/src/internalClusterTest/java/org/opensearch/search/sort/FieldSortIT.java
```diff
@@ -1249,6 +1250,7 @@ public void testSortMissingNumbersMinMax() throws Exception {
         assertThat(searchResponse.getHits().getAt(2).getId(), equalTo("3"));
     }
 
+    @SuppressForbidden(reason = "WTF?")
```
reta, line 1253: :-o probably a bit too much :-D  (but I feel you)

reta (approved):
Nice cleanup, thanks @andrross !

server/src/internalClusterTest/java/org/opensearch/search/sort/FieldSortIT.java
andrross: Check out [line 1282](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/internalClusterTest/java/org/opensearch/search/sort/FieldSortIT.java#L1282) in this class :)

Issue #20078, opened 2025-11-23
[BUG] DataStreams: Can't delete index template that matches the data stream but is unused


jpmat296:
### Describe the bug

This issue is similar to #9194. But in #9194 the two index templates were not data streams templates (no `"data_stream": {}` for the first).

In this new issue, the two index templates are both for data streams.

### Related component

Indexing

### To Reproduce

```
PUT _index_template/test
{
  "index_patterns": [
    "test*"
  ],
  "data_stream": {},
  "template": {},
  "composed_of": [],
  "priority": 50
}

PUT _index_template/test-data-stream
{
  "index_patterns": [
    "test"
  ],
  "data_stream": {},
  "template": {},
  "composed_of": [],
  "priority": 51
}

PUT _data_stream/test

DELETE _index_template/test
```

Get following error at `DELETE` time:
```
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "unable to remove composable templates [test] as they are in use by a data streams [test]"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "unable to remove composable templates [test] as they are in use by a data streams [test]"
  },
  "status": 400
}
```

### Expected behavior

It should be possible to delete the index template as it is not used.

### Additional Details

Setup: Docker on macOS
Opensearch: v3.3.2

gaobinlong:
This is really a bug, very similar to https://github.com/opensearch-project/OpenSearch/issues/9194, taking look into this.

PR #20102, opened 2025-11-26
Fix deletion failure/error of unused index template when the index template matches a data stream but has lower priority

gaobinlong:
### Description

When there're multiple index templates matching data stream, the highest priority template will be used for the new created data stream, we can check the current used template by GET _data_stream/{name} API, but for the low priority template which are not used by any data streams, it cannot be deleted, this PR fixes this bug and add some tests for this case.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/20078

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where deleting an index template could fail when the template matched a data stream pattern but had lower priority than another template.

* **Tests**
  * Added tests validating template deletion behavior and priority-based template selection to prevent regressions.


server/src/main/java/org/opensearch/cluster/metadata/MetadataIndexTemplateService.java
```diff
@@ -1156,36 +1157,38 @@ public static List<IndexTemplateMetadata> findV1Templates(Metadata metadata, Str
...
 
-        if (matchedTemplates.size() == 0) {
+        if (winner == null) {
             return null;
         }
 
-        final List<ComposableIndexTemplate> candidates = new ArrayList<>(matchedTemplates.keySet());
-        CollectionUtil.timSort(candidates, Comparator.comparing(ComposableIndexTemplate::priorityOrZero, Comparator.reverseOrder()));
-
-        assert candidates.size() > 0 : "we should have returned early with no candidates";
```
gaobinlong, line 1185: Since we only need the highest priority template, we can get it in the above for loop, no need to sort.

CHANGELOG.md
```diff
@@ -88,6 +88,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fix ClassCastException in FlightClientChannel for requests larger than 16KB ([#20010](https://github.com/opensearch-project/OpenSearch/pull/20010))
 - Fix GRPC Bulk ([#19937](https://github.com/opensearch-project/OpenSearch/pull/19937))
 - Fix node bootstrap error when enable stream transport and remote cluster state ([#19948](https://github.com/opensearch-project/OpenSearch/pull/19948))
+- Fix delete not using index template failed when the index template matches a data stream but has low priority ([#20102](https://github.com/opensearch-project/OpenSearch/pull/20102))
```
sandeshkr419: nit: Fix deletion of unused index template; case when an index template matches a data stream but has a lower priority.

peterzhuamazon:
@coderabbitai review

PR #20113, opened 2025-11-28
[Derived Source] Support ignore_above/normalizer for keyword/wildcard field and optimise text field lookup

tanik98:
- For keyword and wildcard field handle `ignore_above` by explicitly storing values exceeding that and for `noarmalizer` store value in all cases, and use stored field for using it as a fallback 
- For text field, use sub keyword fields for deriving source(using `doc_values`) instead of explicitly storing the value, this way we can avoid storing duplicates(as a `doc_values` under sub keyword and `stored field` under text field). 
- In case of multiple keyword fields, we will be going through each one to find first one, and in case of ingored field, we are choosing keyword with max value of ignore_above to save the dupplication of docValue and fallback stored field. Updation of sub keyword field is append only, we can't remove sub keyword once added, these would get appended in applicable keyword fields for derived source as and when mapping gets updated.

## Release Notes

* **New Features**
  * Added support for `ignore_above` parameter on keyword and wildcard fields.
  * Optimized text field retrieval in derived source scenarios.

* **Bug Fixes**
  * Improved handling of values exceeding the `ignore_above` threshold for keyword and wildcard fields with derived source enabled.


server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -936,4 +956,51 @@ protected String contentType() {
     public ParametrizedFieldMapper.Builder getMergeBuilder() {
         return new Builder(simpleName(), indexAnalyzers).init(this);
     }
+
+    static final class DerivedSourceHelper {
```
mgodwan: We don't need this separate namespace class

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -872,11 +881,22 @@ protected void parseCreateField(ParseContext context) throws IOException {
...
-        if (value == null || value.length() > ignoreAbove) {
+        if (value == null) {
             return;
         }
 
         NamedAnalyzer normalizer = fieldType().normalizer();
+
+        // Explicitly add value as a stored field if value is getting ignored, to be able to derive the source
+        if (value.length() > ignoreAbove) {
+            if ((normalizer == null || Lucene.KEYWORD_ANALYZER.equals(normalizer))
```
mgodwan: Would this not work if splitting queries on whitespace is set as well? Should we just check for default?

server/src/main/java/org/opensearch/index/mapper/StringFieldType.java
```diff
@@ -255,4 +256,8 @@ public Query rangeQuery(Object lowerTerm, Object upperTerm, boolean includeLower
             includeUpper
         );
     }
+
+    public String derivedSourceIgnoreFieldName() {
+        return name() + IGNORED_VALUE_FIELD_SUFFIX;
```
mgodwan: This can cause issues for existing mappings where users may have created field with this name? Can we only support this feature for index created above the new version?

mgodwan: Also, we should block users from creating field with these naming conventions for indices created beyond this version to avoid naming conflicts.

server/src/main/java/org/opensearch/index/mapper/MatchOnlyTextFieldMapper.java
```diff
@@ -139,21 +139,22 @@ public Builder(String name, Version indexCreatedVersion, IndexAnalyzers indexAna
...
-                multiFieldsBuilder.build(this, context),
+                multiFields,
                 copyTo.build(),
                 this
             );
         }
 
         @Override
-        protected MatchOnlyTextFieldType buildFieldType(FieldType fieldType, BuilderContext context) {
+        protected MatchOnlyTextFieldType buildFieldType(FieldType fieldType, MultiFields multiFields, BuilderContext context) {
```
mgodwan, line 157: Can this break any place where this method is used in plugins?

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -936,4 +956,51 @@ protected String contentType() {
...
+                Collections.emptyMap()
+            ) {
+                @Override
+                public String typeName() {
+                    return "keyword";
+                }
+
+                @Override
+                public ValueFetcher valueFetcher(QueryShardContext context, SearchLookup searchLookup, String format) {
+                    return null;
```
mgodwan: Is this not expected to be called ever? Can we throw UnsupportedOperationException instead?

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -295,11 +294,21 @@ protected void canDeriveSourceInternal() {
...
     protected DerivedFieldGenerator derivedFieldGenerator() {
-        return new DerivedFieldGenerator(
-            mappedFieldType,
-            new SortedSetDocValuesFetcher(mappedFieldType, simpleName()),
-            new StoredFieldFetcher(mappedFieldType, simpleName())
+        final FieldValueFetcher primaryFieldValueFetcher = KeywordFieldMapper.DerivedSourceHelper.getPrimaryFieldValueFetcher(
+            this,
+            simpleName()
+        );
+        final FieldValueFetcher fallbackFieldValueFetcher = KeywordFieldMapper.DerivedSourceHelper.getFallbackFieldValueFetcher(this);
```
mgodwan: Do we always need a fallbackValueFetcher? If normalizer/ignore_above is not set, would it only need primary?

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
tanik98: Removed this from all three mappers.

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
tanik98: Normalizer is not being controlled by splitting queries on whitespace mapping parameter, if normalizer is not configured explicitly then default keyword(doesn't change anything on the string) one is being used.

server/src/main/java/org/opensearch/index/mapper/StringFieldType.java
tanik98: Added version check as well as, added mapping validation in DocumentMapper to not allow fields matching with internal field name .ingored_value.

server/src/main/java/org/opensearch/index/mapper/MatchOnlyTextFieldMapper.java
tanik98: Validated only MatchOnlyTextFieldMapper is extending TextFieldMapper across packages.

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
tanik98: Updated

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
tanik98: Updated it to not always consider fallbackValueFetcher and only for normalizer/ignore_above case.

PR #20123, opened 2025-11-29
Make crypto store settings immutable

RajatGupta02:
### Description
[Opensearch-storage-encryption](https://github.com/opensearch-project/opensearch-storage-encryption) plugin introduces some index settings which should be immutable after index creation. This PR adds the validation to not allow these crypto settings update.

### Check List

## Summary by CodeRabbit

* **New Features**
  * Enforces immutability of crypto-related store settings so sensitive store configuration cannot be altered after index creation.
  * Prevents invalid transitions of index store types involving encrypted store variants, rejecting attempts to switch to or from those types.

* **Bug Fixes**
  * Validations now run during settings updates to fail early on disallowed crypto-store changes.


server/src/main/java/org/opensearch/cluster/metadata/MetadataUpdateSettingsService.java
```diff
@@ -589,4 +590,32 @@ public static void validateIndexTotalPrimaryShardsPerNodeSetting(Settings indexS
...
+        // Crypto settings are completely immutable - reject any attempt to modify them
+        for (String settingKey : restrictedCryptoSettings) {
+            if (indexSettings.keySet().contains(settingKey)) {
+                throw new IllegalArgumentException("Cannot update [" + settingKey + "] - crypto settings are immutable");
+            }
+        }
+
+        // Validate store type changes
+        String newStoreType = indexSettings.get("index.store.type");
+        if ("cryptofs".equals(newStoreType)) {
```
kumargu: and vice versa?  if you have ```cryptofs``` as store type it cannot be modified.

kumargu: this could be included in ```restrictedCryptoSettings```.

RajatGupta02: yeah we should also prevent `cryptofs` -> `non-cryptofs` update. But it shouldn't be included in `restrictedCryptoSettings` as it will prevent prevent ALL store type changes for all indices.

kumargu:
cc @cwperks , could you please review?

server/src/main/java/org/opensearch/cluster/metadata/MetadataUpdateSettingsService.java
```diff
@@ -589,4 +590,41 @@ public static void validateIndexTotalPrimaryShardsPerNodeSetting(Settings indexS
...
     }
+
+    /**
+     * Validates crypto store settings are immutable after index creation.
+     */
+    public static void validateCryptoStoreSettings(Settings indexSettings, Index[] indices, ClusterState clusterState) {
+        final String[] restrictedCryptoSettings = {
+            "index.store.crypto.key_provider",
+            "index.store.crypto.kms.key_arn",
+            "index.store.crypto.kms.encryption_context" };
```
cwperks: I think this can be done through properties like `Property.Filtered` or `Property.InternalIndex`

See https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/settings/Setting.java#L160-L164

The last param in https://github.com/opensearch-project/OpenSearch/blob/main/plugins/crypto-kms/src/main/java/org/opensearch/crypto/kms/KmsService.java#L53 is a vararg list of properties.

RajatGupta02: Oh, yes, this is neat, didn't know about this, thanks!

cwperks:
@RajatGupta02 would it make sense to add any tests here? Is `cryptofs` defined in the core or is that only available when ILE plugin is installed? If its only available when ILE is installed I wonder if we can somehow enforce within ILE. I know there are extension points to listen to cluster settings updates and I imagine index settings update have a similar listener to hook into.

The changes in the PR itself look fine to me.

RajatGupta02:
> @RajatGupta02 would it make sense to add any tests here? Is `cryptofs` defined in the core or is that only available when ILE plugin is installed? If its only available when ILE is installed I wonder if we can somehow enforce within ILE. I know there are extension points to listen to cluster settings updates and I imagine index settings update have a similar listener to hook into.
> 
Yeah, its available only when ILE is installed.

server/src/main/java/org/opensearch/cluster/metadata/MetadataUpdateSettingsService.java
```diff
@@ -589,4 +590,33 @@ public static void validateIndexTotalPrimaryShardsPerNodeSetting(Settings indexS
             );
         }
     }
+
+    /**
+     * Validates crypto store settings are immutable after index creation.
+     */
+    public static void validateCryptoStoreSettings(Settings indexSettings, Index[] indices, ClusterState clusterState) {
```
gbbafna, line 597: Isn't it already static via 

https://github.com/opensearch-project/opensearch-storage-encryption/blob/83495d6d484d22c3d48f03166a8aa4fc463165c1/src/main/java/org/opensearch/index/store/CryptoDirectoryFactory.java#L123C41-L123C67

PR #20129, opened 2025-11-30
HLL field Mapper for Cardinality Rollups

sandeshkr419:
### Description

The `hll` field type stores pre-aggregated cardinality data using HyperLogLog++ (HLL++) sketch data structures. This field type is designed for internal use by OpenSearch and its plugins (such as the ISM plugin for multi-tier rollup operations) to enable efficient cardinality aggregation on rolled-up data.

#### Key Features

- **Efficient Storage**: Stores approximate cardinality estimates in compact binary format sketches
- **Merge Support**: Supports merging multiple HLL++ sketches during aggregation
- **Configurable Precision**: Allows tuning the accuracy-memory tradeoff via precision parameter
- **Rollup Integration**: Designed for use with ISM plugin rollup operations

### Parameters

- `precision` (optional): Integer between 4 and 18 (consistent with OpenSearch) that controls the accuracy-memory tradeoff
  - Default: 12 (keeping consistent with OpenSearch Cardinality Aggregation)
  - Higher precision = better accuracy but more memory

### Related Issues

### Check List

## Summary by CodeRabbit

* **New Features**
  * Added HyperLogLog field type support for efficient cardinality aggregations on pre-aggregated sketches.
  * Cardinality aggregations now support merging pre-aggregated HyperLogLog data from HLL fields.

* **Bug Fixes**
  * Enhanced error messaging for sketch precision mismatch scenarios during merge operations.

* **Tests**
  * Added comprehensive test coverage for HyperLogLog field mapping and cardinality aggregation behavior.

### DIFF
--- a/server/src/test/java/org/opensearch/index/mapper/HllFieldMapperIntegrationTests.java
+++ b/server/src/test/java/org/opensearch/index/mapper/HllFieldMapperIntegrationTests.java
@@ -8,6 +8,7 @@
 
 package org.opensearch.index.mapper;
 
+import org.opensearch.action.search.SearchPhaseExecutionException;
 import org.opensearch.action.search.SearchResponse;
 import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.util.BigArrays;
@@ -574,5 +575,100 @@ public class HllFieldMapperIntegrationTests extends OpenSearchSingleNodeTestCase
         } finally {
             expectedMerged.close();
         }
+    }
+
+    public void testPrecisionMismatchAcrossIndices() throws IOException {
+        // Test aggregating across multiple indices with same field name but different precisions
+        // This validates that OpenSearch properly rejects aggregations across indices with
+        // mismatched HLL precision configurations
+
+        String index1 = "test-hll-precision-12";
+        String index2 = "test-hll-precision-14";
+        int precision1 = 12;
+        int precision2 = 14;
+
+        // Create first index with precision 12
+        XContentBuilder mapping1 = XContentFactory.jsonBuilder()
+            .startObject()
+            .startObject("properties")
+            .startObject("user_sketch")
+            .field("type", "hll")
+            .field("precision", precision1)
+            .endObject()
+            .endObject()
+            .endObject();
+
+        client().admin().indices().prepareCreate(index1).setMapping(mapping1).get();
+
+        // Create second index with precision 14 (different precision, same field name)
+        XContentBuilder mapping2 = XContentFactory.jsonBuilder()
+            .startObject()
+            .startObject("properties")
+            .startObject("user_sketch")
+            .field("type", "hll")
+            .field("precision", precision2)
+            .endObject()
+            .endObject()
+            .endObject();
+
+        client().admin().indices().prepareCreate(index2).setMapping(mapping2).get();
+
+        // Create and index sketch with precision 12 in first index
+        HyperLogLogPlusPlus sketch1 = new HyperLogLogPlusPlus(precision1, BigArrays.NON_RECYCLING_INSTANCE, 1);
+        try {
+            for (int i = 0; i < 50; i++) {
+                sketch1.collect(0, mix64(i));
+            }
+
+            BytesStreamOutput out = new BytesStreamOutput();
+            sketch1.writeTo(0, out);
+            org.apache.lucene.util.BytesRef bytesRef = out.bytes().toBytesRef();
+            byte[] sketchBytes = Arrays.copyOfRange(bytesRef.bytes, bytesRef.offset, bytesRef.offset + bytesRef.length);
+
+            client().prepareIndex(index1)
+                .setId("1")
+                .setSource(XContentFactory.jsonBuilder().startObject().field("user_sketch", sketchBytes).endObject())
+                .get();
+        } finally {
+            sketch1.close();
+        }
+
+        // Create and index sketch with precision 14 in second index
+        HyperLogLogPlusPlus sketch2 = new HyperLogLogPlusPlus(precision2, BigArrays.NON_RECYCLING_INSTANCE, 1);
+        try {
+            for (int i = 0; i < 50; i++) {
+                sketch2.collect(0, mix64(i + 100));
+            }
+
+            BytesStreamOutput out = new BytesStreamOutput();
+            sketch2.writeTo(0, out);
+            org.apache.lucene.util.BytesRef bytesRef = out.bytes().toBytesRef();
+            byte[] sketchBytes = Arrays.copyOfRange(bytesRef.bytes, bytesRef.offset, bytesRef.offset + bytesRef.length);
+
+            client().prepareIndex(index2)
+                .setId("1")
+                .setSource(XContentFactory.jsonBuilder().startObject().field("user_sketch", sketchBytes).endObject())
+                .get();
+        } finally {
+            sketch2.close();
+        }
+
+        client().admin().indices().prepareRefresh(index1, index2).get();
+
+        // Attempt to aggregate across both indices using wildcard pattern
+        // This should fail because the indices have different precisions for the same field
+        SearchPhaseExecutionException exception = expectThrows(SearchPhaseExecutionException.class, () -> {
+            client().prepareSearch("test-hll-precision-*").addAggregation(cardinality("total_users").field("user_sketch")).get();
+        });
+
+        // Verify the error mentions precision mismatch
+        String errorMessage = exception.getMessage();
+        Throwable cause = exception.getCause();
+        String fullMessage = errorMessage + (cause != null ? " " + cause.getMessage() : "");
+
+        assertTrue(
+            "Error should mention precision mismatch. Got: " + fullMessage,
+            fullMessage.contains("Cannot merge HLL++ sketches with different precision")
+        );
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add circuit breaker support for gRPC transport to prevent out-of-memory errors ([#20203](https://github.com/opensearch-project/OpenSearch/pull/20203))
 - Add index-level-encryption support for snapshots and remote-store ([#20095](https://github.com/opensearch-project/OpenSearch/pull/20095))
 - Adding BackWardCompatibility test for remote publication enabled cluster ([#20221](https://github.com/opensearch-project/OpenSearch/pull/20221))
-
 - Support for hll field mapper to support cardinality rollups ([#20129](https://github.com/opensearch-project/OpenSearch/pull/20129))
 
 ### Changed


asimmahmood1:
@coderabbitai review

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * Validates that the binary data is a valid HLL++ sketch by attempting to deserialize it.
+     *
+     * @param sketchBytes the binary sketch data to validate
+     * @throws MapperParsingException if the data is not a valid HLL++ sketch
+     */
+    private void validateSketchData(BytesRef sketchBytes) throws MapperParsingException {
+        try (StreamInput in = new BytesArray(sketchBytes.bytes, sketchBytes.offset, sketchBytes.length).streamInput()) {
+            AbstractHyperLogLogPlusPlus sketch = AbstractHyperLogLogPlusPlus.readFrom(in, BigArrays.NON_RECYCLING_INSTANCE);
```
asimmahmood1: My understanding is that OS's HLLPP doesn't have a version number.

Since we are going to persist this now, I think it'll be worth adding a version number so can handle backward compatibility. 

Or can be use the existing index version number?

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
```diff
@@ -0,0 +1,234 @@
...
+                        + "]: "
+                        + "expected "
+                        + precision
+                        + ", got "
+                        + sketch.precision()
+                );
+            }
+
+            // Close the sketch to release resources
+            sketch.close();
```
asimmahmood1: Put this in try finally block to make sure its closed on exception.

server/src/test/java/org/opensearch/search/aggregations/metrics/HllCardinalityAggregatorTests.java
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
sandeshkr419: deleted file - not needed.

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
sandeshkr419: Ack, done.

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
sandeshkr419: I think like most of the filed mappers, we will have to rely on OpenSearch version compatibility.

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.function.Supplier;
+
+/**
+ * A {@link FieldMapper} for HyperLogLog++ sketch fields.
+ * This field type stores pre-aggregated cardinality data using HLL++ sketch data structures.
+ * It is intended for internal use by OpenSearch and its plugins (such as ISM for multi-tier rollup).
+ *
+ * @opensearch.internal
+ */
+public class HllFieldMapper extends ParametrizedFieldMapper {
```
navneet1v, line 42: For this field do we support parameters like `index`, `doc_values`, `stored`, `copyTo` ? if not should we throw error in mapping if someone tries to set it?

Should this field support Derived Source? I think it should since internally it is storing info in the BDV.

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
```diff
@@ -0,0 +1,231 @@
...
+     * @opensearch.internal
+     */
+    public static class Builder extends ParametrizedFieldMapper.Builder {
+
+        private final Parameter<Integer> precision = Parameter.intParam(
+            "precision",
+            false,
+            m -> toType(m).precision,
+            HyperLogLogPlusPlus.DEFAULT_PRECISION
+        ).setValidator(Builder::validatePrecision);
```
navneet1v, line 62: do we allow changing the precision of the field as mapping update?

navneet1v:
@sandeshkr419 another thing as we were discussing offline to consider here is the new mapper which we are adding will be exposed to customers. I don't think we want this mapper to be used by customers for their fields. So question we should ask ourselves is do we want users to use this mapper? if yes how we want them to use it. Because if ans is no, we should see how we can hide this mapper to be added by users(something similar to SourceFieldMapper). 

Questions to ans:
1. Is this a good idea to expose the new mapper to user?
2. If we want to hide the mapper then how system will come to know that it has use the Mapper for a specific field/index.

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
sandeshkr419: > For this field do we support parameters like index, doc_values, stored, copyTo ? if not should we throw error in mapping if someone tries to set it?

Good find. Only `doc_values` needs to be true, `copyTo` can be supported, rest will throw an error in revision changes.

sandeshkr419: No, HLL fields should not support Derived Source.

- HLL sketches are binary data structures, not human-readable values
- The "source" would just be base64-encoded binary data
- No meaningful reconstruction possible - you'd just get the same binary blob back
- The default behavior (throwing UnsupportedOperationException) is correct for this field type. If someone tries to use HLL fields with derived source enabled, they'll get a clear error message explaining that it's not supported, which is the appropriate behavior.

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
sandeshkr419: No, precision also defines the length of binary ingested fields. So this cannot be changed after creation. Adding a test for validation as well.

sandeshkr419 (commented):
Thanks @navneet1v for reviewing, I have addressed your comments.


On your questions on making this an internal field mapper, I think we can make it as internal field `_hll` which can help, let me check on it once to see if that satisfies the requirements.

sandeshkr419:
@navneet1v I was looking into making this field an internal field `_hll` but the problem with that is it will be a metadata field and only a single such field can exist in the index. That will bring the limitation that one index can only be configured with this field type. So this needs to be a regular field type so that in case multiple fields of hll can be stored in an index.

sandeshkr419:
@navneet1v Can you please take a look again. I'll add the changelog next. CI (`gradle check`) is succeeding now. :D

server/src/main/java/org/opensearch/index/fielddata/plain/HllFieldData.java
```diff
@@ -0,0 +1,177 @@
...
+
+        HllLeafFieldData(LeafReader reader, String fieldName) {
+            this.reader = reader;
+            this.fieldName = fieldName;
+        }
+
+        @Override
+        public long ramBytesUsed() {
+            return 0;
+        }
```
navneet1v, line 140: any reason for having this value as 0.

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
navneet1v: @sandeshkr419 I think the idea of derived source is not make it human readable, but more so to reduce the storage size. Since we already support DV for this field, we should support derived source for this field. Another reason for that is if derived source is not supported on this field then any index that uses the field will not be able to use derived source feature, as per my understanding.

server/src/test/java/org/opensearch/index/mapper/HllFieldMapperIntegrationTests.java
```diff
@@ -0,0 +1,405 @@
...
+import static org.opensearch.search.aggregations.AggregationBuilders.terms;
+import static org.opensearch.search.aggregations.bucket.terms.Terms.Bucket;
+
+/**
+ * Integration tests for HLL field mapper demonstrating programmatic field creation
+ * and usage patterns for ISM plugin rollup scenarios.
+ */
+public class HllFieldMapperIntegrationTests extends OpenSearchSingleNodeTestCase {
+
+    public void testProgrammaticFieldCreation() throws IOException {
```
navneet1v, line 38: can we also add an IT where some docs has this HLL field and some docs doesn't and validate if indexing, search and aggregations work as expected.

navneet1v:
> @navneet1v Can you please take a look again. I'll add the changelog next. CI (`gradle check`) is succeeding now. :D

most of the code looks good. Just few minor comments

server/src/main/java/org/opensearch/index/fielddata/plain/HllFieldData.java
sandeshkr419: Return 0 because HllLeafFieldData doesn't hold any persistent heap memory:
1. The underlying doc values are managed by Lucene (stored off-heap, accessed via memory mapping)
2. When getSketch() is called, bytes are temporarily copied to heap and deserialized, but the caller is responsible for immediately closing the sketch
3. No sketches are cached or held by this HllLeafFieldData instance
4. Aggregator memory (HyperLogLogPlusPlus counts) is tracked separately via BigArrays

This follows the same pattern as BinaryDVLeafFieldData and other doc values implementations

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
sandeshkr419: Got it, added the implementation and test for it. Implementation is straight - just read binary doc value and base64 encode it, similar to how binary fields work.

server/src/main/java/org/opensearch/search/aggregations/metrics/HllCardinalityAggregator.java
```diff
@@ -0,0 +1,133 @@
...
+                        counts.merge(bucket, sketch, 0);
+                    }
+                } catch (IllegalArgumentException e) {
+                    // Log precision mismatch or other merge errors
+                    logger.warn(
+                        "Failed to merge HLL++ sketch for field [{}] in document {}: {}",
+                        fieldData.getFieldName(),
+                        doc,
+                        e.getMessage()
+                    );
```
asimmahmood1, line 97: Overall looks good. 

Codecov shows good code coverage, I think this is only one that's missing: https://app.codecov.io/gh/opensearch-project/OpenSearch/pull/20129?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=opensearch-project#91629e37ef00f5fe68212507af353c8f-R90

I'm not sure different precisions can be indexed/merged into same index, see if you can add a test case for this as well.

server/src/main/java/org/opensearch/search/aggregations/metrics/HllCardinalityAggregator.java
```diff
@@ -0,0 +1,133 @@
...
+
+                        // Grow if needed to accommodate this bucket
+                        if (bucket >= counts.maxOrd()) {
+                            HyperLogLogPlusPlus newCounts = new HyperLogLogPlusPlus(precision, context.bigArrays(), bucket + 1);
+                            for (long i = 0; i < counts.maxOrd(); i++) {
+                                if (counts.cardinality(i) > 0) {
+                                    newCounts.merge(i, counts, i);
+                                }
+                            }
+                            counts.close();
```
asimmahmood1, line 83: I think this close is correct. If newCounts on line 77 fails (e.g. due to bigArray circuit breaker), it'll close newCounts first, then throw exception.

this.counts will be closed when agg.close is called (line 131).

Am I right?

asimmahmood1:
We trying to restart gradle check but I don't have permission to reopen. @sandeshkr419 @prudhvigodithi

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
navneet1v: Thanks

navneet1v:
Approved the code. Please fix the CIs

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        // HLL fields are always stored as doc values and cannot be indexed or stored separately
+        private final Parameter<Boolean> index = Parameter.indexParam(m -> false, false).setValidator(v -> {
+            if (v) {
+                throw new MapperParsingException("Cannot set [index] on field of type [" + CONTENT_TYPE + "]");
+            }
+        });
+
+        private final Parameter<Boolean> store = Parameter.storeParam(m -> false, false).setValidator(v -> {
```
bharath-techie, line 84: For non derived source cases, we will need store right ? otherwise how will we reindex for example

sandeshkr419: No, the store parameter is not needed for reindexing.
Here’s why:
- Reindex uses `_source`, not stored fields - The reindex API retrieves document data from _source, as evidenced by the code in ReindexBasicTests.java which shows reindex working with derived source enabled.
- HLL fields support derived source - When derived source is enabled, the HLL field values are reconstructed from doc values using DocValueFetcher. When derived source is disabled (default), the values are in the regular `_source`.
- Store would be redundant - If we enabled store, the data would be stored in 3 places:
  - _source (or reconstructed from doc values)
  - Doc values (for aggregations)
  - Stored fields (never accessed!)

Consistent with similar fields - dense_vector also disables store for the same reasons.

bharath-techie:
```
private final Parameter<Boolean> store = Parameter.storeParam(m -> toType(m).store, false);
```
Also, do we need to set it this way ? otherwise won't this be always false ?

server/src/main/java/org/opensearch/search/aggregations/metrics/HllCardinalityAggregator.java
sandeshkr419: Done.

server/src/main/java/org/opensearch/index/mapper/HllFieldMapper.java
sandeshkr419: The pattern Parameter.storeParam(m -> false, false) means:

- First parameter m -> false: Lambda to extract value from existing mapper (always returns false since we block it)
- Second parameter false: Default value

This is consistent with how we handle index and docValues parameters that are also blocked/forced to specific values. The implementation is correct as-is.

PR #20132, opened 2025-12-01
BulkRequest: Make sure that indices Set is properly initialized when being deserialized from TransportRequest stream

nibix:
### Description

In the class `BulkRequest`, the `indices` property provides a short-cut to retrieve the names of all indices referenced in the bulk items. 

The index names are added to the `Set` when one of the `add()` methods is being called. However, the property is neither being serialized, nor the constructor for deserialization makes sure that the property is being properly initialized again. Thus, the property will be an empty set after the request was serialized and deserialized.

Fortunately, this happens only in rare cases, specifically when ingestion pipelines are present and the request did not hit a node with the role "ingest". Compare this code:

https://github.com/opensearch-project/OpenSearch/blob/1ee30dcc8334f63c51bed276397fa4d50406a8e6/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L254-L277

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed bulk request deserialization so indices are correctly populated and tracked after a serialization round-trip.

* **Tests**
  * Added a serialization/deserialization test to verify that indices are preserved after serializing and deserializing bulk requests.

* **Documentation**
  * Added a changelog entry noting the deserialization fix for bulk requests.

### DIFF
--- a/server/src/test/java/org/opensearch/action/bulk/TransportBulkActionTests.java
+++ b/server/src/test/java/org/opensearch/action/bulk/TransportBulkActionTests.java
@@ -53,14 +53,11 @@ import org.opensearch.cluster.metadata.Metadata;
 import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.cluster.node.DiscoveryNodeRole;
 import org.opensearch.cluster.service.ClusterService;
-import org.opensearch.common.io.stream.ReleasableBytesStreamOutput;
+import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.unit.TimeValue;
-import org.opensearch.common.util.MockBigArrays;
-import org.opensearch.common.util.MockPageCacheRecycler;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.concurrency.OpenSearchRejectedExecutionException;
-import org.opensearch.core.indices.breaker.NoneCircuitBreakerService;
 import org.opensearch.index.IndexNotFoundException;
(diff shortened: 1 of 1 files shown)


server/src/test/java/org/opensearch/action/bulk/TransportBulkActionTests.java
```diff
@@ -379,6 +384,18 @@ public void testRejectionAfterCreateIndexIsPropagated() throws Exception {
         }
     }
 
+    public void testSerializationDeserialization() throws Exception {
+        BulkRequest bulkRequest = new BulkRequest().add(new IndexRequest("index").id("id").source(Collections.emptyMap()));
+        MockBigArrays mockBigArrays = new MockBigArrays(new MockPageCacheRecycler(Settings.EMPTY), new NoneCircuitBreakerService());
+
+        try (ReleasableBytesStreamOutput out = new ReleasableBytesStreamOutput(mockBigArrays)) {
```
cwperks: nit: I I believe this can be simplified to something like:

```
BytesStreamOutput out = new BytesStreamOutput();
out.setVersion(Version.V_3_4_0);
```

ref: https://github.com/opensearch-project/OpenSearch/blob/1ee30dcc8334f63c51bed276397fa4d50406a8e6/modules/aggs-matrix-stats/src/test/java/org/opensearch/search/aggregations/matrix/stats/MatrixStatsAggregatorTests.java#L165-L182

nibix: Thanks for the hint! Fixed.

PR #20140, opened 2025-12-01
Fix duplicate registration of FieldDataCache dynamic setting

sgup432:
### Description
We encountered an issue where fieldDataCache dynamic setting was getting registered multiple times ([link](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/indices/fielddata/cache/IndicesFieldDataCache.java#L126)) causing settingUpdaters list to bloat up ([link](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/settings/AbstractScopedSettings.java#L76)) and eventually causing memory issues on active cluster manager node.

This is a sample screenshot which shows multiple such settingUpdaters, all belonging to FieldDataCache and specifically `indices.fielddata.cache.size` setting.

Screenshot 2025-11-27 at 2 09 42 PM

#### Why this happens?
This is happening as during every index creation/updates, we verify whether metadata is sane or not by creating temporary FieldDataCache objects, and later closing them on. By doing this, the setting getting registered multiple times and causing this issue. See logic [here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/indices/IndicesService.java#L1228)

### Related Issues

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate registration of fielddata and query cache settings so each is registered only once.

* **New Features**
  * Exposed a read-only view for inspecting registered setting updaters.

* **Tests**
  * Added an integration test verifying settings are registered exactly once across multiple index creations.

* **Changelog**
  * Updated release notes to include the duplicate-registration fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,7 +93,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fix GRPC Bulk ([#19937](https://github.com/opensearch-project/OpenSearch/pull/19937))
 - Fix node bootstrap error when enable stream transport and remote cluster state ([#19948](https://github.com/opensearch-project/OpenSearch/pull/19948))
 - Fix deletion failure/error of unused index template; case when an index template matches a data stream but has a lower priority. ([#20102](https://github.com/opensearch-project/OpenSearch/pull/20102))
-- Fix toBuilder method in EngineConfig to include mergedSegmentTransferTracker([20105](https://github.com/opensearch-project/OpenSearch/pull/20105))
+- Fix toBuilder method in EngineConfig to include mergedSegmentTransferTracker([#20105](https://github.com/opensearch-project/OpenSearch/pull/20105))
 - Fixed handling of property index in BulkRequest during deserialization ([#20132](https://github.com/opensearch-project/OpenSearch/pull/20132))
 - Fix negative CPU usage values in node stats ([#19120](https://github.com/opensearch-project/OpenSearch/issues/19120))
 - Fix duplicate registration of FieldDataCache dynamic setting ([20140](https://github.com/opensearch-project/OpenSearch/pull/20140))


server/src/internalClusterTest/java/org/opensearch/index/fielddata/FieldDataLoadingIT.java
```diff
@@ -240,6 +247,37 @@ public void testFieldDataCacheClearConcurrentFields() throws Exception {
...
+        List<AbstractScopedSettings.SettingUpdater<?>> settingUpgraders = clusterSettings.getSettingUpdaters();
+        for (AbstractScopedSettings.SettingUpdater<?> settingUpgrader : settingUpgraders) {
+            String input = settingUpgrader.toString();
+            // Trying to fetch key value as there is no other way
+            Pattern p = Pattern.compile("\"key\"\\s*:\\s*\"([^\"]+)\"");
+            Matcher m = p.matcher(input);
+            String key = "";
+            if (m.find()) {
+                key = m.group(1);
+            }
```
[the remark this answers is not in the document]
sgup432: SettingUpdater doesn't have setting field, so this won't work.

server/src/main/java/org/opensearch/indices/IndicesService.java
```diff
@@ -1228,10 +1228,13 @@ public synchronized MapperService createIndexMapperService(IndexMetadata indexMe
...
         try {
+            // For FieldDataCache and QueryCache, we deliberately don't pass clusterService/clusterSettings to avoid registering dynamic
+            // setting each time this method is called. As otherwise, it will bloat up the settingUpdater list and cause memory exhaustion
+            // in cluster manager node. This is safe as eventually these are temporary objects and are closed eventually.
             IndicesFieldDataCache indicesFieldDataCache = new IndicesFieldDataCache(settings, new IndexFieldDataCache.Listener() {
-            }, clusterService, threadPool);
+            });
             closeables.add(indicesFieldDataCache);
-            IndicesQueryCache indicesQueryCache = new IndicesQueryCache(settings, clusterService.getClusterSettings());
+            IndicesQueryCache indicesQueryCache = new IndicesQueryCache(settings);
```
peteralfonsi: This will prevent duplicate registrations during metadata updates, but won't we still see them during IndicesService.withTempIndexService (which happens on index creation)?

sgup432: I don't think so. Inside `IndicesService.withTempIndexService`,  it seems like it is reusing the global `IndicesFieldDataCache` object and not creating them every time. Inside `Indices.verifyIndexMetadata`, the problem is that it was creating temporary `IndicesFieldDataCache` objects every time, and it was the main culprit.

Issue #19919, opened 2025-11-07
[Context Aware Segments] Fix indexing regression for Context Aware Segments due to contention on DocumentWriterFlushControl


RS146BIJAY:
### Is your feature request related to a problem? Please describe

Fix indexing regression for Context Aware Segments due to contention on DocumentWriterFlushControl. We are observing 30% regression on throughput and latency because of this issue

### Related component

Indexing

cwperks:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/cwperks) [3](https://github.com/imdhemy)

Thank you for filing this issue @RS146BIJAY . Are you actively working on this? What is the status?

PR #20145, opened 2025-12-02
Fixing indexing regression and bug fixes for grouping criteria

RS146BIJAY:
### Description
Fixing indexing regression and bug fixes for grouping criteria. For testing grouping criteria changes, enabled the grouping criteria on local and tested with setting criteria. Wil raise the changes for integ test enablement for CAS in a separate PR as that require decent changes in integ test as well.

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/19919

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  - Fixed indexing regression and grouping criteria handling; removed a now-unused internal lock-related exception.

* **Improvements**
  - Better RAM and flush accounting for index writers with per-child document tracking.
  - Improved state and error handling during concurrent indexing and refresh operations.

* **Tests**
  - Expanded tests covering tragic-exception scenarios, RAM/flush behavior, and concurrent delete/refresh cases.

* **Documentation**
  - Changelog updated to reflect these fixes and improvements.


server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -468,36 +466,41 @@ String getCriteriaForDoc(BytesRef key) {
...
                 success = true;
                 return writer;
             } finally {
-                if (success == false && current != null) {
-                    assert current.mapReadLock.isHeldByCurrentThread() == true;
+                if (success == false && current != null && current.mapReadLock.isHeldByCurrentThread() == true) {
                     current.mapReadLock.close();
                 }
             }
         }
```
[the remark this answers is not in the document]
RS146BIJAY: Resolved this comment.

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -468,16 +466,31 @@ String getCriteriaForDoc(BytesRef key) {
...
+                    // This function acquires a first read lock on a map which does not have any write lock present. Current keeps
+                    // on getting rotated during refresh, so there will be one current on which read lock can be obtained.
+                    // Validate that no write lock is applied on the map and the map is not closed. Idea here is write lock was
+                    // never applied on this map as write lock gets only during closing time. We are doing this instead of acquire,
+                    // because acquire can also apply a read lock in case refresh completed and map is closed.
+                    current = this.current.mapReadLock.tryAcquire();
+                    if (current != null && current.isClosed() == true) {
+                        current.mapReadLock.close();
+                        current = null;
+                    }
```
Bukhtawar: Should we move this logic centrally in `close()`.  Also didn't quite understand why the close operation done as a part of write lock acquisition doesn't handle this logic?

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -813,24 +794,15 @@ public final synchronized Iterable<Map.Entry<String, String>> getLiveCommitData(
...
-            Collection<IndexWriter> oldWriterSet = liveIndexWriterDeletesMap.old.criteriaBasedIndexWriterMap.values()
-                .stream()
-                .map(DisposableIndexWriter::getIndexWriter)
-                .collect(Collectors.toSet());
-            for (IndexWriter indexWriter : oldWriterSet) {
-                if (indexWriter.isOpen() == true) {
-                    indexWriter.rollback();
+            for (DisposableIndexWriter disposableIndexWriter : liveIndexWriterDeletesMap.old.criteriaBasedIndexWriterMap.values()) {
+                if (disposableIndexWriter.getIndexWriter().isOpen() == true) {
+                    disposableIndexWriter.getIndexWriter().rollback();
```
Bukhtawar: This might not be thread-safe for instance the index writer might be closed while we are doing a rollback?

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
RS146BIJAY: This handles the scenario where try acquire succeded in obtaining the lock on the current writer but the map itself rotated and the writer got closed. In that case, we close the old writer as we retry to obtain lock again on the current. As this ensures the lock is correctly released on the old writer before we try acquiring lock on new writer.

CHANGELOG.md
```diff
@@ -100,6 +100,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fixed handling of property index in BulkRequest during deserialization ([#20132](https://github.com/opensearch-project/OpenSearch/pull/20132))
 - Fix negative CPU usage values in node stats ([#19120](https://github.com/opensearch-project/OpenSearch/issues/19120))
 - Fix duplicate registration of FieldDataCache dynamic setting ([20140](https://github.com/opensearch-project/OpenSearch/pull/20140))
+- Fix toBuilder method in EngineConfig to include mergedSegmentTransferTracker([20105](https://github.com/opensearch-project/OpenSearch/pull/20105))
```
mgodwan: Rebase?

server/src/main/java/org/opensearch/index/IndexSettings.java
```diff
@@ -498,9 +498,9 @@ public static IndexMergePolicy fromString(String text) {
      */
     public static final Setting<Integer> INDEX_MAX_RETRY_ON_LOOKUP_MAP_LOCK_ACQUISITION_EXCEPTION = Setting.intSetting(
         "index.context_aware.max_retry_on_lookup_map_acquisition_exception",
-        15,
-        5,
         100,
```
mgodwan: Why is the default so high?

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -690,7 +691,9 @@ public boolean isCompositeIndexPresent() {
     }
 
     public Set<CompositeMappedFieldType> getCompositeFieldTypes() {
```
bharath-techie, line 693: Lets rename to `getStarTreeFieldTypes` ?

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
RS146BIJAY: Just wondering if requires to be thread safe. I mean if we are anyways closing writers here. In case the writers are already closed, we can move to closing next writes in the rollback flow. We can keep the entire code block in a try catch block and simply ignore AlreadyClosedException for the rollback flow?? 

In all other cases where any of the writer is closed, we can close all the writer as well?? As this means we encountered some tragic errors while writing to lucene

server/src/main/java/org/opensearch/index/IndexSettings.java
RS146BIJAY: So just added stats around the time for which we were unable to occupy block on current map, number of retires made and number of times map rotated from old to new:
`"max_retry_count" : 518,
"time_taken_in_nano_sec" : 11517,
"map_rotated_count" : 0,`

So adjusted this value accordingly.

It seems about 500 retries were made in 11517 `nano sec`, while number of time maps were rotated in this duration was 0. Only logical explanation I can figure out is it is taking this much time for changes made on a volatile object to be reflected for this variable [during map rotation](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java#L516).

mgodwan: We can get rid of the retry limit given the retries are like CAS. We should always expect to be in a terminal state.

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -686,11 +687,13 @@ public Iterable<MappedFieldType> fieldTypes() {
     }
 
     public boolean isCompositeIndexPresent() {
-        return this.mapper != null && !getCompositeFieldTypes().isEmpty();
+        return this.mapper != null && !getStarTreeFieldTypes().isEmpty();
     }
 
-    public Set<CompositeMappedFieldType> getCompositeFieldTypes() {
-        return compositeMappedFieldTypes;
+    public Set<CompositeMappedFieldType> getStarTreeFieldTypes() {
```
mgodwan: @bharath-techie Would it make sense for this to be tied to data-cube abstraction?

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -209,12 +207,12 @@ public CriteriaBasedIndexWriterLookup getLookupMap() {
...
      *
      */
-    public static final class CriteriaBasedIndexWriterLookup implements Closeable {
+    public static class CriteriaBasedIndexWriterLookup implements Closeable {
         private final Map<String, DisposableIndexWriter> criteriaBasedIndexWriterMap;
         private final Map<BytesRef, DeleteEntry> lastDeleteEntrySet;
         private final Map<BytesRef, String> criteria;
         private final ReentrantReadWriteLock mapLock;
-        private final CriteriaBasedWriterLock mapReadLock;
+        CriteriaBasedWriterLock mapReadLock;
```
mgodwan: This can continue to be private final?

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -765,27 +747,19 @@ public Throwable getTragicException() {
...
-            .map(DisposableIndexWriter::getIndexWriter)
-            .collect(Collectors.toSet());
 
         try (ReleasableLock ignore = liveIndexWriterDeletesMap.current.mapWriteLock.acquire()) {
-            for (IndexWriter indexWriter : currentWriterSet) {
-                if (indexWriter.isOpen() == true) {
-                    ramBytesUsed += indexWriter.ramBytesUsed();
+            for (DisposableIndexWriter disposableIndexWriter : liveIndexWriterDeletesMap.current.criteriaBasedIndexWriterMap.values()) {
+                if (disposableIndexWriter.getIndexWriter().isOpen() == true) {
+                    ramBytesUsed += disposableIndexWriter.getIndexWriter().ramBytesUsed();
```
mgodwan: [minor] Instead of looping through it, we can probably maintain bytes used separately. If this is called too often, we may see a penatly when we are at high memory as this may end up being called always. Its alright to take this ins a separate PR

mgodwan (approved):
Overall LGTM. Few minor comments

server/src/main/java/org/opensearch/index/mapper/MapperService.java
bharath-techie: Yes data cube abstraction makes sense.

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
RS146BIJAY: Actually writer keeps on rotating as well so may need to keep this bytes used inside Map itself if we want to precompute it. But I would want to keep this as a separate discussion and not linked to this PR.

RS146BIJAY: Also discussed this offline, ramBytesUsed calculation will only become an issue when number of tenants is higher. Will create a separate github issue for this.

server/src/main/java/org/opensearch/index/mapper/MapperService.java
RS146BIJAY: It seems changing fieldTypes name is breaking backward compatiblity test cases.

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
RS146BIJAY: Created a gihub issue to track this: https://github.com/opensearch-project/OpenSearch/issues/20380

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -533,14 +546,17 @@ private void refreshDocumentsForParentDirectory(CriteriaBasedIndexWriterLookup o
         final Map<String, CompositeIndexWriter.DisposableIndexWriter> markForRefreshIndexWritersMap = oldMap.criteriaBasedIndexWriterMap;
         deletePreviousVersionsForUpdatedDocuments();
         final List<Directory> directoryToCombine = new ArrayList<>();
+        final AtomicLong pendingNumDocsByOldChildWriter = new AtomicLong();
```
Bukhtawar: I believe we don't need this be atomic since this is not a shared resource?

server/src/main/java/org/opensearch/index/engine/InternalEngine.java
```diff
@@ -1239,7 +1239,7 @@ private boolean mayHaveBeenIndexedBefore(Index index) {
 
     private void addDocs(final List<ParseContext.Document> docs, final DocumentIndexWriter indexWriter, Term uid) throws IOException {
         if (docs.size() > 1) {
-            indexWriter.addDocuments(docs, uid);
+            indexWriter.addDocuments(docs, uid, docs.size());
```
Bukhtawar: We are already passing the reference for docs, why do we need to change the interface to additionally pass a size. This can lead to an inconsistent state if the size isn't correctly passed

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -533,14 +546,17 @@ private void refreshDocumentsForParentDirectory(CriteriaBasedIndexWriterLookup o
...
         for (CompositeIndexWriter.DisposableIndexWriter childDisposableWriter : markForRefreshIndexWritersMap.values()) {
             directoryToCombine.add(childDisposableWriter.getIndexWriter().getDirectory());
             childDisposableWriter.getIndexWriter().close();
+            pendingNumDocsByOldChildWriter.addAndGet(childDisposableWriter.getIndexWriter().getPendingNumDocs());
         }
 
         if (!directoryToCombine.isEmpty()) {
             accumulatingIndexWriter.addIndexes(directoryToCombine.toArray(new Directory[0]));
             IOUtils.closeWhileHandlingException(directoryToCombine);
+            childWriterPendingNumDocs.addAndGet(-pendingNumDocsByOldChildWriter.get());
```
mgodwan: Can we evaluate if it makes sense to add index iteratively, one by one? This can cause excessive flushes, but reduce how much we overshoot reserved docs

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
RS146BIJAY: Addressed the comments.

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
RS146BIJAY: I think it will reduce the overshot for reserved docs. All the flushes will happen for parent writer. This will happen during refresh flow (in async flow) and parentWriter does not accept any active operation except delete. So do not think we will observe too much impact due to parent flush. 
Addressed the comments.

server/src/main/java/org/opensearch/index/engine/InternalEngine.java
RS146BIJAY: I think we easily fix this by making CompositeIndexWriter accept a list. I think I have added this because we are passing Iterable<ParseContext.Document> to CompositeIndexWriter: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java#L887C30-L887C66. Getting size of iterable will require us to iterate all values. 

Addressed the comment.

Issue #18362, opened 2025-05-23
[Workload Management][Rule Based Autotagging] Scroll API support in autotagging


kaushalmahi12:
### Is your feature request related to a problem? Please describe

Currently Rule based autotagging is done by a `ActionFilter` and it mainly works by matching the target indices against the admin created Rules. But `SearchScrollRequest` doesn't provide a way to access target indices for this request at `ActionFilter`.

### Describe the solution you'd like

Introduce a mechanism to expose target indices in scroll requests.

### Related component
Search, Rule based autotagging

andrross:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

tigermint:
Hi! I'm new to OpenSearch but would like to work on this issue. 
Could you please assign it to me?

I'm planning to investigate how to expose target indices in SearchScrollRequest 
at the ActionFilter level. 

Any guidance on the approach would be appreciated!

laminelam:
Hi @tigermint  
I have already started working on this and have a PR almost ready for review. Forgot to ask to assign the issue to me. Do you mind I get the issue assigned to me?

PR #20151, opened 2025-12-02
[Workload Management][Rule Based Autotagging] Scroll API support in autotagging

laminelam:
Resolves [#8362](https://github.com/opensearch-project/OpenSearch/issues/18362)

### Description
 

### Check List

## Summary by CodeRabbit

* **New Features**
  * Scroll requests are now tagged by workload management rules and propagate workload headers.

* **Improvements**
  * Scroll identifiers now carry original index expressions so index metadata is preserved across scrolls.
  * Scroll request handling was extended to extract original indices from request metadata for consistent processing.

* **Tests**
  * Added tests covering tagging and metadata behavior for scroll flows.

### DIFF
--- a/server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
+++ b/server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
@@ -57,7 +57,7 @@ final class TransportSearchHelper {
 
     private static final String INCLUDE_CONTEXT_UUID = "include_context_uuid";
 
-    public static final Version INDICES_IN_SCROLL_ID_VERSION = Version.V_3_4_0;
+    public static final Version INDICES_IN_SCROLL_ID_VERSION = Version.V_3_6_0;
 
     static InternalScrollSearchRequest internalScrollSearchRequest(ShardSearchContextId id, SearchScrollRequest request) {
         return new InternalScrollSearchRequest(request, id);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add getWrappedScorer method to ProfileScorer for plugin access to wrapped scorers ([#20548](https://github.com/opensearch-project/OpenSearch/issues/20548))
 - Support expected cluster name with validation in CCS Sniff mode ([#20532](https://github.com/opensearch-project/OpenSearch/pull/20532))
 - Add security policy to allow `accessUnixDomainSocket` in `transport-grpc` module ([#20463](https://github.com/opensearch-project/OpenSearch/pull/20463))
+- [Workload Management] Enhance Scroll API support for autotagging ([#20151](https://github.com/opensearch-project/OpenSearch/pull/20151))
 
 ### Changed
 - Move Randomness from server to libs/common ([#20570](https://github.com/opensearch-project/OpenSearch/pull/20570))


plugins/workload-management/src/main/java/org/opensearch/plugin/wlm/AutoTaggingActionFilter.java
```diff
@@ -80,14 +85,44 @@ public <Request extends ActionRequest, Response extends ActionResponse> void app
...
+                public LogicalOperator getLogicalOperator() {
+                    return LogicalOperator.AND;
+                }
+            });
+        } else if (isSearchRequest) {
+            attributeExtractors.add(new IndicesExtractor((IndicesRequest) request));
+        } else {
+            chain.proceed(task, action, request, listener);
+            return;
+        }
```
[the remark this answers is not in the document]
laminelam: I followed the same logic than IndicesExtractor which uses AND.


```
public class IndicesExtractor implements AttributeExtractor<String> {
    private final IndicesRequest indicesRequest;

    /**
     * Default constructor
     * @param indicesRequest
     */
    public IndicesExtractor(IndicesRequest indicesRequest) {
        this.indicesRequest = indicesRequest;
    }

    @Override
    public Attribute getAttribute() {
        return RuleAttribute.INDEX_PATTERN;
    }

    @Override
    public Iterable<String> extract() {
        return List.of(indicesRequest.indices());
    }

    @Override
    public LogicalOperator getLogicalOperator() {
        return AND;
    }
}
```

laminelam:
@kaushalmahi12 
When you have time, would you please take a look into this?
Thx

laminelam:
Hi @andrross 
Can you please help on this?

andrross:
@kaushalmahi12 Can you take a look?

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -78,6 +84,19 @@ static String buildScrollId(AtomicArray<? extends SearchPhaseResult> searchPhase
...
+                // To keep autotagging consistent between the initial SearchRequest
+                // and subsequent SearchScrollRequests, we store exactly the same
+                // index targets that were visible to the indices attribute during
+                // the "search" phase
+                if (originalIndices != null && originalIndices.length > 0) {
+                    out.writeVInt(originalIndices.length);
+                    for (String index : originalIndices) {
+                        out.writeString(index);
+                    }
+                }
```
andrross: Do we need to be worried about the size of the scroll ID? Index names can be up to 255 bytes. Is there a bound to the number of indices that can be included here?

laminelam: The scroll id already scales with number of shards, so adding _originalIndices_ is extra but typically small because we store request.indices() patterns/aliases (what the user provided), not the expanded concrete indices. 

We can cap it and fall back to _unknown_ when we exceed some limit; but in that case auto-tagging won’t be applied, which is a tradeoff to avoid unbounded size. 

What do you think/recommend? Should we enforce a size limit here?

andrross: Is it possible to avoid including extra information in the scroll id? If we already know the shards then we can get the indexes, at least in theory. Is the problem that we need the user-provided pattern/alias and not the concrete indexes?

kaushalmahi12:
I will look into it this week. @dzane17 Can you do the first round of review on this ?

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
laminelam: >  Is the problem that we need the user-provided pattern/alias and not the concrete indexes?

Yes, that's exactly the issue here. The  current _AutoTaggingActionFilter_ logic runs _SearchRequest_/_SearchScrollRequest_ against an _IndicesExtractor_ which uses _RuleAttribute.INDEX_PATTERN_

So the key here is "pattern". The user sends a _pattern_ in the SearchRequest (it could be alias, wildcard, concrete indices, etc.) and we need that information (in the following _SearchScrollRequest_) for the rule engine to map patterns attributes to tags (I am simplifying). It doesn't run against the real physical indices names.

andrross: I think including original indices is probably reasonable here. The size of this data should be modest relative to all the shard information in large indices.

I've pinged @jainankitk to help find an autotagging subject matter expert to review this as well.

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -60,7 +60,13 @@ static InternalScrollSearchRequest internalScrollSearchRequest(ShardSearchContex
         return new InternalScrollSearchRequest(request, id);
     }
 
+    public static final Version INDICES_IN_SCROLL_ID_VERSION = Version.V_3_3_2;
```
dzane17: Update to latest version before merge

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -78,6 +84,19 @@ static String buildScrollId(AtomicArray<? extends SearchPhaseResult> searchPhase
                     out.writeString(searchShardTarget.getNodeId());
                 }
             }
+
+            if (version.onOrAfter(INDICES_IN_SCROLL_ID_VERSION)) {
+                // To keep autotagging consistent between the initial SearchRequest
+                // and subsequent SearchScrollRequests, we store exactly the same
+                // index targets that were visible to the indices attribute during
+                // the "search" phase
+                if (originalIndices != null && originalIndices.length > 0) {
```
dzane17: Can you use `writeArray` instead of `writeInt` + `writeString` loop? Also if `originalIndices == null` write an empty array so that there is always something in the stream.

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -114,10 +133,22 @@ static ParsedScrollId parseScrollId(String scrollId) {
...
+            String[] originalIndices;
+            if (in.getPosition() < bytes.length) {
+                final int numOriginalIndices = in.readVInt();
+                originalIndices = new String[numOriginalIndices];
+                for (int i = 0; i < numOriginalIndices; i++) {
+                    originalIndices[i] = in.readString();
+                }
+            } else {
+                originalIndices = new String[0];
+            }
```
dzane17: I believe a version check is needed and use `readArray`

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
laminelam: Done.
Thx for the suggestion. Didn't notice the existence the StringArray version.

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
laminelam: It may sound counterintuitive, but you don’t need the version check when reading, because:
- The reader doesn’t know its version.
- The reader is supposed to be "old" in a scenario of an upgrade with a version mismatch. If it's "old" it's not aware of the new version you’e introducing. If it did it would't be “old” but new and the check would be unless.
- "_if (in.getPosition() < bytes.length)_" already plays that role. If the received chunk was written by a new version this conditional code would be execute, if not that part gets ignored.
- This is the reason why _parseScrollId_ does not have a "_version_" parameter contrary to the _buildScrollId_  method.

laminelam: Thx for the review @dzane17  btw

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -114,10 +129,13 @@ static ParsedScrollId parseScrollId(String scrollId) {
                 }
                 context[i] = new SearchContextIdForNode(clusterAlias, target, new ShardSearchContextId(contextUUID, id));
             }
+
+            final String[] originalIndices = in.getPosition() < bytes.length ? in.readStringArray() : Strings.EMPTY_ARRAY;
+
             if (in.getPosition() != bytes.length) {
                 throw new IllegalArgumentException("Not all bytes were read");
```
andrross, line 135: Related to @dzane17's question, if a mixed-version cluster is running then it's possible that a new node will generate a scroll ID which will get sent to an old node. Since the old node does not know anything about the new field, won't it fail to parse by hitting this condition?

In the "normal" node-to-node usage of StreamInput/StreamOutput, the sender always knows the version of the node it is talking to and can craft a message it knows the receiver will be able to parse (this is what all the version.onOrAfter code does). In this case, we're sending this binary data as base64 all the way to the end user to be sent back in a subsequent request. We have no control over which node will handle that subsequent request. Am I missing something?

dzane17: @andrross Sending scroll ID from new node to old is covered because there is a version check on the sender's side in `buildScrollId(AtomicArray<? extends SearchPhaseResult> searchPhaseResults, Version version)`. 

For parsing scroll id, receiving node does NOT know the sender version. That is why `in.getPosition() < bytes.length` is used in place of `version.onOrAfter`. I believe this pattern works and is actually more flexible than the traditional check because it does not make any assumptions based on sender's version, only looks at the actual data in the stream.

andrross: > Sending scroll ID from new node to old is covered

@dzane17 I'm specifically thinking of the case where I have a mixed cluster with the nodes behind a load balancer. I get a scroll ID from the REST client then send that ID back in a subsequent request. I have no control which node will receive the REST request. What happens if this scroll ID was generated by a new node on the first request but is received by an old node on the second request?

andrross: I would actually expect the following to replicate this failure:

```
./gradlew ':qa:mixed-cluster:v3.2.0#mixedClusterTest' --tests 'org.opensearch.backwards.MixedClusterClientYamlTestSuiteIT' -Dtests.method='test {p0=scroll/10_basic/Basic scroll}'
```

It wouldn't fail 100% of the time because I think it's random which node in the mixed cluster handles the test requests. However, I've been running it in a loop and it doesn't fail. I must be misunderstanding something here...

plugins/workload-management/src/main/java/org/opensearch/plugin/wlm/AutoTaggingActionFilter.java
```diff
@@ -80,14 +83,39 @@ public <Request extends ActionRequest, Response extends ActionResponse> void app
...
 
         if (!isValidRequest || wlmClusterSettingValuesProvider.getWlmMode() == WlmMode.DISABLED) {
             chain.proceed(task, action, request, listener);
             return;
         }
         List<AttributeExtractor<String>> attributeExtractors = new ArrayList<>();
-        attributeExtractors.add(new IndicesExtractor((IndicesRequest) request));
+        if (isSearchRequest) {
+            attributeExtractors.add(new IndicesExtractor((IndicesRequest) request));
+        } else {
```
kaushalmahi12, line 97: This is also an `IndicesExtractor` use case then we should use that since semantically both are same. I think `ActionRequestMetadata` is a much more generic construct and we should not introduce a specific behavior which is not reusable. Since `ParsedScrollId` already has indices then why not use that via searchScrollRequest.

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -60,7 +61,13 @@ static InternalScrollSearchRequest internalScrollSearchRequest(ShardSearchContex
         return new InternalScrollSearchRequest(request, id);
     }
 
+    public static final Version INDICES_IN_SCROLL_ID_VERSION = Version.V_3_3_0;
+
     static String buildScrollId(AtomicArray<? extends SearchPhaseResult> searchPhaseResults, Version version) {
+        return buildScrollId(searchPhaseResults, null, version);
+    }
+
+    static String buildScrollId(AtomicArray<? extends SearchPhaseResult> searchPhaseResults, String[] originalIndices, Version version) {
```
kaushalmahi12, line 70: We should not require the `Version` param here since we should always perform the version check based on `StreamInput` and `StreamOutput` instances since they already have this information.

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -78,6 +85,14 @@ static String buildScrollId(AtomicArray<? extends SearchPhaseResult> searchPhase
                     out.writeString(searchShardTarget.getNodeId());
                 }
             }
+
+            if (version.onOrAfter(INDICES_IN_SCROLL_ID_VERSION)) {
```
kaushalmahi12, line 89: This is not the right way since we want to ensure that the receiving node is on the right version. But if the idea here is that we want to send it only when the sending node is at least `INDICES_IN_SCROLL_ID_VERSION` then this if clause is redundant since this `INDICES_IN_SCROLL_ID_VERSION` is not introduced at later point in time.

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -78,6 +85,14 @@ static String buildScrollId(AtomicArray<? extends SearchPhaseResult> searchPhase
...
+                // To keep autotagging consistent between the initial SearchRequest
+                // and subsequent SearchScrollRequests, we store exactly the same
+                // index targets that were visible to the indices attribute during
+                // the "search" phase
+                out.writeStringArray(originalIndices == null ? Strings.EMPTY_ARRAY : originalIndices);
```
kaushalmahi12: I think `writeOptionalStringArray` is much more cleaner

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
kaushalmahi12: I agree with @andrross here since the behavior will be flakey specially in a mixed version cluster scenario.

server/src/main/java/org/opensearch/action/search/TransportSearchScrollAction.java
```diff
@@ -114,4 +117,19 @@ protected void doExecute(Task task, SearchScrollRequest request, ActionListener<
             listener.onFailure(e);
         }
     }
+
+    @Override
+    public String[] originalIndices(SearchScrollRequest request) {
```
kaushalmahi12: I feel it would be better to move this behavior into the `SearchScrollRequest` since indices are the property of the request not Action

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
laminelam: I think the confusion comes from what _version_ means here.
_buildScrollId_ is passed the cluster **minNodeVersion**, NOT the local node version:

`
Version minNodeVersion = clusterState.nodes().getMinNodeVersion();`

  
So in a mixed cluster (ex: Node A v= 3.3, Node B v = 3.2), _minNodeVersion_ is 3.2 and this new _originalIndices_ is not written. The scroll ID stays in the old format and can be parsed by any node. Only after all nodes are upgraded does the _originalIndices_ get added.

On the read side we use _in.getPosition() < bytes.length_ because the receiver doesn't know which node originally generated the scroll ID.

That's why this is safe even behind a load balancer:

- The 1st request hits Node A --> _minNodeVersion_ = 3.2  ==> no new _originalIndices_ written.
- The 2nd request hits Node B --> parses old-format scroll ID ==> good.

Hope this makes sense now.

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
laminelam: > We should not require the `Version` param here since we should always perform the version check based on `StreamInput` and `StreamOutput` instances since they already have this information.


StreamOutput versioning works only for **node-to-node** messages where the receiver's version is known.

A scroll ID goes to the **client** and may be parsed by any node later, so there is no negotiated version here. This is why the check should be against the cluster _minNodeVersion_ not the  StreamOutput's version, as explained above.

BTW, _version_ is not a new introduced param, it was already there before this PR.

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
laminelam: the decision is based on the [minNodeVersion](https://github.com/opensearch-project/OpenSearch/pull/20151#discussion_r2739400905)

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
andrross: Thanks @laminelam. I was indeed missing the fact this is keyed off of the minimum node version across the cluster. This is a safe way to version the scroll token given it has a short lifetime and isn't intended to be persisted long-term on the client side.

This also explains why the mixed cluster test does not fail :)

plugins/workload-management/src/main/java/org/opensearch/plugin/wlm/AutoTaggingActionFilter.java
laminelam: Hi @kaushalmahi12 
I have made the change.

jainankitk (commented):
Checks failed due to spotless failure:

```
* What went wrong:
Execution failed for task ':plugins:workload-management:spotlessJavaCheck'.
      src/internalClusterTest/java/org/opensearch/plugin/wlm/WlmAutoTaggingIT.java
          @@ -467,24 +467,16 @@
           
           ············try·{
           ················int·afterInitialSearch·=·getCompletions(workloadGroupId);
          -················assertTrue(
          -····················"Expected·completions·to·increase·after·initial·search·with·scroll",
          -····················afterInitialSearch·>·completionsBefore
          -················);
```

Please Run './gradlew spotlessApply' to fix all violations. Also, can you add entry to the changelog?

kaushalmahi12:
Thanks @laminelam for making this change just fix the the gradle checks, rest of it looks good to me.

laminelam:
> Thanks @laminelam for making this change just fix the the gradle checks, rest of it looks good to me.

Thanks @kaushalmahi12 
I already fixed the spotless issue (thx @jainankitk for pointing it out).
When the PR is ready for merge will update the change log

server/src/main/java/org/opensearch/action/search/TransportSearchHelper.java
```diff
@@ -56,11 +57,17 @@ final class TransportSearchHelper {
 
     private static final String INCLUDE_CONTEXT_UUID = "include_context_uuid";
 
+    public static final Version INDICES_IN_SCROLL_ID_VERSION = Version.V_3_4_0;
```
andrross: This will need to be V_3_6_0

laminelam: Done
Thanks @andrross

PR #20163, opened 2025-12-04
Add support for forward translog reading

### LINKED ISSUE #20094: [Feature Request] read forward in translog

Currently in [translog.MultiSnapshot](https://github.com/opensearch-project/OpenSearch/blob/2e4b517576a6822585b836b7d7274a77ddc89ecd/server/src/main/java/org/opensearch/index/translog/MultiSnapshot.java#L82), the translog is read in reverse order. For example, when replaying, it'll process the translog files as `translog_3.tlog -> translog_2.tlog -> translog_1.tlog`. The reason behind this is to avoid stale operations during recovery, as in reverse order, the higher term operation will be replayed, and the lower term operation will be deduped.

Since we now also do [indexShard.trimOperationOfPreviousPrimaryTerms(trimAboveSeqNo);](https://github.com/opensearch-project/OpenSearch/blob/2e4b517576a6822585b836b7d7274a77ddc89ecd/server/src/main/java/org/opensearch/indices/recovery/RecoveryTarget.java#L257) in the finalizingRecovery phase, the duplicated operations from the previous term will be trimmed, which enables us to read the translog forward in most cases.

There is a TODO in the code: [// TODO: Read translog forward in 9.0+](https://github.com/opensearch-project/OpenSearch/blob/2e4b517576a6822585b836b7d7274a77ddc89ecd/server/src/main/java/org/opensearch/index/translog/MultiSnapshot.java#L82). The author ended up leaving it there because there might be edge cases when read forward, for example, the recovery replica failed before it was able to trim the translog, and it then retries another PEER recovery, when replaying translog locally up to it's global checkpoint, it might still read the stale operations. 
However, this edge case can be avoided if we trim the translog in the early phase, before it starts the translog replay.

- Add settings to allow force read forward in translog
- When it's reading forward, trim the translog before it actually starts the replay.

### Related component

Indexing:Replication

_No response_

### Additional context

_No response_

xuxiong1:
### Description
This commit introduces a feature to read translog operations in forward order
(oldest to newest) instead of the default backward order (newest to oldest).

Changes:
- Add `index.translog.read_forward` setting (default: false) in `IndexSettings`
- Update MultiSnapshot to support bidirectional reading based on setting

Tests:
- testRecoveryTrimsLocalTranslogWithReadForward (RecoveryTests)
- testSeqNoCollisionWithReadForward (IndexLevelReplicationTests)
- testSnapshotReadOperationForward (LocalTranslogTests)

### Related Issues
Resolves #20094

### Check List

## Summary by CodeRabbit

* **New Features**
  * Add a configurable index-level option to enable forward reading of transaction logs for snapshots and recoveries.

* **Improvements**
  * Resync now advances translog generation before trimming and adjusts initial sync boundaries to improve recovery correctness.

* **Tests**
  * Added tests validating forward translog reading across snapshots, replication/recovery flows, and recovery trimming.

* **Documentation**
  * Updated changelog to announce forward translog read support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Support dynamic consumer configuration update in pull-based ingestion ([#19963](https://github.com/opensearch-project/OpenSearch/pull/19963))
 - Cache the `StoredFieldsReader` for scroll query optimization ([#20112](https://github.com/opensearch-project/OpenSearch/pull/20112))
 - Add Hybrid Cardinality collector to prioritize Ordinals Collector ([#19524](https://github.com/opensearch-project/OpenSearch/pull/19524))
-- Add support for forward translog reading ([#20163](https://github.com/opensearch-project/OpenSearch/pull/20163)
 - Add support for forward translog reading ([#20163](https://github.com/opensearch-project/OpenSearch/pull/20163))
 
 ### Changed


server/src/main/java/org/opensearch/index/translog/MultiSnapshot.java
```diff
@@ -79,15 +81,30 @@ public int skippedOperations() {
...
-            Translog.Operation op;
-            while ((op = current.next()) != null) {
-                if (op.seqNo() == SequenceNumbers.UNASSIGNED_SEQ_NO || seenSeqNo.getAndSet(op.seqNo()) == false) {
-                    return op;
-                } else {
-                    overriddenOperations++;
+        if (readForward) {
+            // Read forward: from index 0 to translogs.length - 1
+            for (; index < translogs.length; index++) {
+                final TranslogSnapshot current = translogs[index];
```
yupeng9: nit: the iteration logic is identical and can be extracted as a private method

xuxiong1: thanks, updated

server/src/main/java/org/opensearch/index/translog/MultiSnapshot.java
```diff
@@ -54,17 +54,19 @@ final class MultiSnapshot implements Translog.Snapshot {
...
-    MultiSnapshot(TranslogSnapshot[] translogs, Closeable onClose) {
+    MultiSnapshot(TranslogSnapshot[] translogs, Closeable onClose, boolean readForward) {
         this.translogs = translogs;
         this.totalOperations = Arrays.stream(translogs).mapToInt(TranslogSnapshot::totalOperations).sum();
         this.overriddenOperations = 0;
         this.onClose = onClose;
         this.seenSeqNo = new SeqNoSet();
-        this.index = translogs.length - 1;
+        this.readForward = readForward;
+        this.index = readForward ? 0 : translogs.length - 1;
```
msfroh: If you declare a field `private final PrimitiveIterator.OfInt iterator;`, then you can write this as:

```
        IntStream forwardStream = IntStream.range(0, translogs.length);
        if (readForward) {
            this.iterator = forwardStream.iterator();
        } else {
            this.iterator = forwardStream.map(i -> translogs.length - 1 - i).iterator();
        }
        index = iterator.nextInt();
```

Then the `next()` method can be:

```
    @Override
    public Translog.Operation next() throws IOException {
        while (iterator.hasNext()) {
            final TranslogSnapshot current = translogs[index];
            Translog.Operation op;
            while ((op = current.next()) != null) {
                if (op.seqNo() == SequenceNumbers.UNASSIGNED_SEQ_NO || seenSeqNo.getAndSet(op.seqNo()) == false) {
                    return op;
                } else {
                    overriddenOperations++;
                }
            }
            index = iterator.nextInt();
        }
        return null;
    }
```

msfroh: Oh -- actually, I guess that won't work, because the semantics of `hasMore` goes one step farther. 

Maybe if you do `IntStream forwardStream = IntStream.rangeClosed(0, translogs.length);` it will work, though.

msfroh: Anyway, the basic idea is that you put all of your conditional logic in one place in the constructor -- you either have an ascending stream `0 .. translogs.length` or a descending stream `translogs.length - 1 .. -1`?

xuxiong1: Good idea! I think this should just work:
```
        if (readForward) {
            this.iterator = IntStream.range(0, translogs.length).iterator();
        } else {
            this.iterator = IntStream.range(0, translogs.length).map(i -> translogs.length - 1 - i).iterator();
        }
```

msfroh:
@xuxiong1  -- Looks like the latest test failure is related to this change

```
org.opensearch.indices.recovery.RecoveryTests.testRecoveryTrimsLocalTranslogWithReadForward failed

java.lang.AssertionError: NoOp{seqNo=26, primaryTerm=41, reason='filling gaps'}
Expected: <40L>
     but: was <41L>
```

```
REPRODUCE WITH: 
```

```
./gradlew ':server:test' --tests 'org.opensearch.indices.recovery.RecoveryTests.testRecoveryTrimsLocalTranslogWithReadForward' -Dtests.seed=AA22BA30D3B12559 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=yue-Hans -Dtests.timezone=America/Knox_IN -Druntime.java=25
```

xuxiong1:
> @xuxiong1 -- Looks like the latest test failure is related to this change
> 
> ```
> org.opensearch.indices.recovery.RecoveryTests.testRecoveryTrimsLocalTranslogWithReadForward failed
> 
> java.lang.AssertionError: NoOp{seqNo=26, primaryTerm=41, reason='filling gaps'}
> Expected: <40L>
>      but: was <41L>
> ```
> 
> ```
> REPRODUCE WITH: 
> ```
> 
> ```
> ./gradlew ':server:test' --tests 'org.opensearch.indices.recovery.RecoveryTests.testRecoveryTrimsLocalTranslogWithReadForward' -Dtests.seed=AA22BA30D3B12559 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=yue-Hans -Dtests.timezone=America/Knox_IN -Druntime.java=25
> ```

Was able to reproduce with the seed.

I think the issue is from the primary replica re-sync process, the new primary would re-sync its operations (from global checkpoint to max seqNo) to the old primary (now replica), which would cause the duplicate operations (same seqNo, different term) in the old primary's translog. 

The re-sync process also triggers` trimOperationOfPreviousPrimaryTerms` on the replica, but was passed with the max seqNo as the `trimAboveSeqNo` (honestly, I didn't get why we would pass the max seq instead of the start seq), so in the replica the operations between the global checkpoint to the max seqNo are not actually trimmed. Then the read forward would encounter the stale operation (read backward actually hides the issue).

The fix is to trim on replica during re-sync, starting from the global checkpoint instead of max seqNo, like how it's done in the [finalizeRecovery](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/indices/recovery/RecoveryTarget.java#L257)

PR #20166, opened 2025-12-04
write FipsPlugin to configure all test-clusters

beanuwave:
### Description
Write FipsPlugin to configure all test-clusters

### Check List


buildSrc/src/main/java/org/opensearch/gradle/testclusters/TestClustersPlugin.java
```diff
@@ -111,6 +113,11 @@ public void apply(Project project) {
 
         // register cluster hooks
         project.getRootProject().getPluginManager().apply(TestClustersHookPlugin.class);
+
```
reta, line 116: Hm ... I think this is not the way we should be doing it, it is going to sneak into any test case that uses test clusters, we have two options here:
 - add test project to https://github.com/opensearch-project/OpenSearch/tree/main/qa folder
 - add example plugin to https://github.com/opensearch-project/OpenSearch/tree/main/plugins/examples

Please note that test clusters are used outside of the core and core plugins

PR #20173, opened 2025-12-05
Updated the Composite and Cardinality aggregators

vinaykpud:
1. Made changes in the `partial_agg_optimizer.rs`, to take the function name instead of just name, bcz it can be alias depending on the Physical plan. So `e.fun().name()` will get the correct name 
2. Changes in `CompositeAggregator.java` to handle when we have `sourceConfig.fieldType` is script filedType
3. Changes in `CardinalityAggregator.java` to recognize names like `dc(UserID)[hll_registers]`
4. Changes in `SearchEngineResultConversionUtils.java` to handle aggregator of type `CardinalityAggregator`

### Description

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/search/query/SearchEngineResultConversionUtils.java
+++ b/server/src/main/java/org/opensearch/search/query/SearchEngineResultConversionUtils.java
@@ -83,9 +83,9 @@ public class SearchEngineResultConversionUtils {
                 if (aggregator instanceof ValueCountAggregator) {
                     docCount = ((InternalValueCount) subAgg).getValue();
                 }
-                if (aggregator instanceof CardinalityAggregator) {
-                    docCount = ((InternalCardinality) subAgg).getValue();
-                }
+//                if (aggregator instanceof CardinalityAggregator) {
+//                    docCount = ((InternalCardinality) subAgg).getValue();
+//                }
                 subAggs.add(subAgg);
             }
         }


server/src/main/java/org/opensearch/search/aggregations/bucket/composite/CompositeAggregator.java
```diff
@@ -739,10 +740,13 @@ public List<InternalAggregation> convert(Map<String, Object[]> shardResult, Sear
         List<CompositeKey> compositeKeys = new ArrayList<>(shardResult.size());
         for (int i = 0; i < shardResult.get(shardResult.keySet().stream().findFirst().get()).length; i++) {
             for (CompositeValuesSourceConfig sourceConfig : sourceConfigs) {
-                if (sourceConfig.fieldType() == null) {
-                    throw new UnsupportedOperationException("Composite aggregation does not support script field types");
-                }
-                Object[] values = shardResult.get(sourceConfig.fieldType().name());
+//                if (sourceConfig.fieldType() == null) {
+//                    throw new UnsupportedOperationException("Composite aggregation does not support script field types");
```
mch2, line 744: lets fall back to our logic here instead of commenting out?

```java
                if (sourceConfig.fieldType() == null) {
                Object[] values = shardResult.get(sourceConfig.name());
                currentCompositeKey.add(searchContext.convertToComparable(values[i]));
                 continue;

                }
```

vinaykpud: @mch2 
I tested with fall back logic, but it breaks another query (40)
```
query="source=hits| where CounterID = 62 AND EventDate >= '2013-07-01' AND EventDate <= '2013-07-31' AND IsRefresh = 0| eval Src = case(SearchEngineID = 0 AND AdvEngineID = 0, Referer else '')| eval Dst = URL| stats count() as PageViews by TraficSourceID, SearchEngineID, AdvEngineID, Src, Dst| sort - PageViews| head 10"
```
here:
sourceConfig.name() is Dst and fieldType was not null,
sourceConfig. fieldType().name() is URL
and with fallback logic it will use URL and it cant find any values for URL.

So i think lets keep it this way for now.

server/src/main/java/org/opensearch/search/query/SearchEngineResultConversionUtils.java
```diff
@@ -76,6 +83,9 @@ public static Tuple<List<InternalAggregation>, Long> extractSubAggsAndDocCount(A
                 if (aggregator instanceof ValueCountAggregator) {
                     docCount = ((InternalValueCount) subAgg).getValue();
                 }
+                if (aggregator instanceof CardinalityAggregator) {
+                    docCount = ((InternalCardinality) subAgg).getValue();
```
expani: This seems incorrect. Cardinality is an approximation and can't replace doc count.

vinaykpud: I havent dived deep on this much, just did whats needed to fix the error for the query. 

this was added for queries like this were failing ie where we group by some string
```
query="source=hits | where MobilePhoneModel != '' | stats dc(UserID) as u by MobilePhoneModel | sort - u | head 10"
```

expani: Got it we should figure out why `agg_for_doc_count` column is not coming in the response from datafusion instead of replacing with distinct count of a different column. 

This can also cause issues with merging results at Coordinator

vinaykpud: this is group by for distinct count, in this case we are not adding `agg_for_doc_count` column in the calcite plan

server/src/main/java/org/opensearch/search/aggregations/metrics/CardinalityAggregator.java
```diff
@@ -765,7 +765,10 @@ public long nextValue() throws IOException {
 
     @Override
     public InternalAggregation convertRow(Map<String, Object[]> shardResult, int row, SearchContext searchContext) {
-        Object[] hlls = shardResult.get(name);
+        Object[] hlls = shardResult.get(name + "[hll_registers]");
```
expani, line 768: Can you try queries with Alias on distinct cardinality and filtering on the alias ?

expani: Is this also required without your SQL changes ?

vinaykpud: No, this was required after my SQL changes.

vinaykpud: dint we already cover this type of queries in Integ test?

PR #20176, opened 2025-12-06
Added ability to prefetch docs in fetch phase

### LINKED ISSUE #18780: [Feature Request] AsyncIO and IO concurrency in Fetch phase

While executing the fetch phase for the queries, we iterate overall desired document ids to load and fetch the documents using StoredFieldReader sequentially, this is probably fine on faster storage devices. But on slower storage this latency significantly adds up. 

I tested the impact of this on a "slower" storage device and fetching 10 documents in a very small data set by making sure none of the data is in page cache. 
I executed a simple term query (on this 11Gb) data set and on cold page cache this term query took ~200ms. Out of this ~150ms or so were contributed to fetch phase. 

Lucene supports prefetching in storedfield reader, that takes docId as an input. Which enable use to prefetch the data for required docId asynchronously. Based on how many documents needs to be fetched we can also think of doing this concurrently (for eg if we want to return 100 documents versus 10 or so). 

In some cases fdt files are loaded as niofs files, we can also implement prefetching in niofs using posix fadvices. We probably have to maintain a separate buffer in niofs index input for prefetched data.

### Related component

_No response_

_No response_

### Additional context

_No response_

sruti1312:
### Description
Adding ability to prefetch docs in fetch phase

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/18780

### Check List

Performance result for 

| Size | With Prefetch P50 (ms) | Without Prefetch P50 (ms) | With Prefetch P90 (ms) | Without Prefetch P90 (ms) | With Prefetch P100 (ms) | Without Prefetch P100 (ms) |
|------|------------------------|---------------------------|-------------------------|---------------------------|-------------------------|----------------------------|
| Default | 450.869 | 482.155 | 502.351 | 493.867 | 513.384 | 499.374 |
| 100 | 505.812 | 652.855 | 530.202 | 727.217 | 563.502 | 743.535 |
| 1000 | 1914.27 | 4127.89 | 1966.09 | 4298.37 | 1972.7 | 4676.91 |

Fetch Phase Cost (Arthas) for size as 10:
| Configuration | P50 (ms) | P90 (ms) | P99 (ms) | P100 (ms) |
|---------------|----------|----------|----------|-----------|
| With Prefetch | 14.703235 | 22.755569 | 94.606265 | 106.321523 |
| Without Prefetch | 29.532539 | 36.979982 | 98.615315 | 112.293080 |

For different queries,
| Query | With Prefetch P50 (ms) | Without Prefetch P50 (ms) | With Prefetch P90 (ms) | Without Prefetch P90 (ms) | With Prefetch P100 (ms) | Without Prefetch P100 (ms) |
|-------|------------------------|---------------------------|-------------------------|---------------------------|-------------------------|----------------------------|
| term | 468.926 | 474.929 | 476.166 | 525.064 | 511.542 | 677.766 |
| asc_sort_timestamp | 590.812 | 603.017 | 612.373 | 659.616 | 623.102 | 717.948 |
| desc_sort_timestamp | 578.757 | 643.036 | 592.766 | 701.633 | 605.739 | 747.248 |

## Summary by CodeRabbit

* **New Features**
  * Added an experimental, dynamic index-level setting (enabled by default) to control document prefetching during the fetch phase; can be toggled at runtime per index.

* **Tests**
  * Added unit tests covering default behavior, explicit configuration, dynamic updates, prefetch thresholds, sequencing, nested-doc handling, loop bounds, and reader/doc calculations.

* **Documentation**
  * CHANGELOG updated to note the new fetch-phase prefetch capability.

### DIFF
--- a/server/src/main/java/org/opensearch/index/IndexSettings.java
+++ b/server/src/main/java/org/opensearch/index/IndexSettings.java
@@ -36,6 +36,7 @@ import org.apache.lucene.index.MergePolicy;
 import org.apache.lucene.sandbox.index.MergeOnFlushMergePolicy;
 import org.opensearch.Version;
 import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.common.annotation.ExperimentalApi;
 import org.opensearch.common.annotation.PublicApi;
 import org.opensearch.common.logging.Loggers;
 import org.opensearch.common.settings.IndexScopedSettings;
@@ -886,6 +887,14 @@ public final class IndexSettings {
         Property.Dynamic
     );
 
+    @ExperimentalApi
+    public static final Setting<Boolean> PREFETCH_DOCS_DURING_FETCH_ENABLED = Setting.boolSetting(
+        "index.prefetch_docs.fetch_phase.enabled",
+        true,
+        Property.IndexScope,
+        Property.Dynamic
+    );
+
     private final Index index;
     private final Version version;
     private final Logger logger;
@@ -1034,6 +1043,11 @@ public final class IndexSettings {
      * Denotes whether search via star tree index is enabled for this index
      */
     private volatile boolean isStarTreeIndexEnabled;
+
+    /**
+     * Denotes whether prefetch of docs during fetch phase is enabled for this index
+     */
+    private volatile boolean prefetchDocsEnabled;
 
     /**
      * Returns the default search fields for this index.
@@ -1207,6 +1221,8 @@ public final class IndexSettings {
             TieredMergePolicyProvider.INDEX_COMPOUND_FORMAT_SETTING,
             tieredMergePolicyProvider::setNoCFSRatio
         );
+        prefetchDocsEnabled = scopedSettings.get(PREFETCH_DOCS_DURING_FETCH_ENABLED);
+        scopedSettings.addSettingsUpdateConsumer(PREFETCH_DOCS_DURING_FETCH_ENABLED, this::setPrefetchDocsEnabled);
         scopedSettings.addSettingsUpdateConsumer(
             TieredMergePolicyProvider.INDEX_MERGE_POLICY_DELETES_PCT_ALLOWED_SETTING,
             tieredMergePolicyProvider::setDeletesPctAllowed
@@ -2308,5 +2324,14 @@ public final class IndexSettings {
 
     public boolean isDerivedSourceEnabled() {
         return derivedSourceEnabled;
+    }
+
+    @ExperimentalApi
+    public boolean isPrefetchDocsEnabled() {
+        return this.prefetchDocsEnabled;
+    }
+
+    private void setPrefetchDocsEnabled(boolean prefetchDocsEnabled) {
+        this.prefetchDocsEnabled = prefetchDocsEnabled;
     }
 }
--- a/server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
+++ b/server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
@@ -35,8 +35,11 @@ package org.opensearch.search.fetch;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.lucene.codecs.StoredFieldsReader;
+import org.apache.lucene.index.FilterLeafReader;
+import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.ReaderUtil;
+import org.apache.lucene.index.SegmentReader;
 import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.TotalHits;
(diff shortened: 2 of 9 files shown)


server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
```diff
@@ -172,10 +175,45 @@ public void execute(SearchContext context, String profileDescription) {
...
+                    }
+
+                    if (prefetchReaderContext.reader() instanceof SequentialStoredFieldsLeafReader lf && useSequentialPrefetch) {
+                        lf.getSequentialStoredFieldsReader().prefetch(relativeDocId);
+                    } else {
+                        if (prefetchReaderContext.reader() instanceof FilterLeafReader filterLeafReader) {
+                            LeafReader leafReader = FilterLeafReader.unwrap(filterLeafReader);
+                            if (leafReader instanceof SegmentReader segmentReader) {
+                                segmentReader.getFieldsReader().prefetch(relativeDocId);
+                            }
```
abiesps, line 216: Should we have some upper bound on how many prefetch calls we make. 
We could also batch documents by segments and call prefetch only for  one docId per page size for each segment. Let me know your thoughts on this ?

abiesps: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingStoredFieldsReader.java#L624 - I guess, this code is already taking care of it.

abiesps: I am also wondering, if there is a way to know this 'prefetching distance' (When to call prefetch before actual IO) and what is its relationship with IO latency (given there is enough memory and compute). Like in this case, you were able to see around 50% improvement in fetch phase execution time by iterating over docId twice. 
Do we know why the time hasn't improved further, if there is still IO during the loop of actually fetching doc Ids.

abiesps: Actually the granularity of prefetched blocks cache is at clone level of IndexInput (https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingStoredFieldsReader.java#L624) . Can we validate if this translate to per request scope or per thread. 
If its per request, there is a scope of removing redundant prefetch call by having another cache at higher granular level (than request scope).

bowenlan-amzn:
Seems to me the scenario is mainly for a cold start on a slow storage, where page cache is empty, using prefetch API could warm up quicker because of parallel IO. 

I see the numbers in the description which is nice! Could you also note down the machine, workload and probably some steps to clear page cache before benchmark.

Some metrics could also help to show the impact of this change, and useful for operation readiness.

iostat could show
- io wait should decrease
- read wait should decrease
- read requests merged should increase (prefetch may batch?)

page cache
- cache hits and misses should be higher and lowere for prefetch respectively

server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
```diff
@@ -172,10 +175,45 @@ public void execute(SearchContext context, String profileDescription) {
             }
         }
 
+        int prefetchReaderIndex = -1;
+        LeafReaderContext prefetchReaderContext = null;
+        boolean hasSequentialDocs = hasSequentialDocs(docs);
+        boolean useSequentialPrefetch = hasSequentialDocs && docs.length >= 10;
```
bowenlan-amzn: A thought here on `docs.length`, when it's very large, probably we also shouldn't do this prefetch, or do this in a batched manner. 
I'm little worried early prefetched pages may be evicted from page cache before they are read.

server/src/main/java/org/opensearch/index/IndexSettings.java
```diff
@@ -866,6 +866,13 @@ public static IndexMergePolicy fromString(String text) {
         Property.Dynamic
     );
 
+    public static final Setting<Boolean> PREFETCH_DOCS_DURING_FETCH_ENABLED = Setting.boolSetting(
```
bowenlan-amzn, line 891: Since this file is annotated with public API, please label this as experimental so easy to update or remove in the future. Similar for other public methods and fields, and in SearchContext class also.

server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
abiesps: > Actually the granularity of prefetched blocks cache is at clone level of IndexInput (https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/lucene90/compressing/Lucene90CompressingStoredFieldsReader.java#L624) . Can we validate if this translate to per request scope or per thread. 

Nevermind its at segment reader level.

msfroh:
@sruti1312  -- I'm curious about whether this is for the case where stored fields are all paged into RAM or where we potentially need to page in from disk? (I imagine it would also be pretty good for the case where stored fields are in a remote directory.)

That is, I'm wondering if we do parallel LZ4 block decompression on the stored fields or if it's just parallelizing the I/O part.

server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
```diff
@@ -174,8 +177,55 @@ public void execute(SearchContext context, String profileDescription) {
...
+                        int docId = docs[index].docId;
+                        int readerIndex = ReaderUtil.subIndex(docId, context.searcher().getIndexReader().leaves());
+                        if (currentReaderIndex != readerIndex) {
+                            currentReaderContext = context.searcher().getIndexReader().leaves().get(readerIndex);
+                        }
+                        int relativeDocId = docId - currentReaderContext.docBase;
+                        int rootId = findRootDocumentIfNested(context, currentReaderContext, relativeDocId);
+                        // skipping prefetch for nested docs as it needs additional processing like
+                        // loading the document source and filtering it based on the nested document ID
+                        if (rootId == -1) {
```
kkewwei, line 205: On my Mac (with a 16KB page size), when retrieving documents, nested documents and their parent document are quite likely to be contained within the same page. I’m therefore wondering if this check is really required.

server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
```diff
@@ -174,8 +177,55 @@ public void execute(SearchContext context, String profileDescription) {
...
+
+        // Prefetch blocks containing the first N documents to ensure these blocks are loaded in advance.
+        // For most queries, which fetch up to 1000 docs, this improves cold-start latency
+        // without evicting useful cached blocks.
+        int prefetchDocsThreshold = 1000;
+        int numberOfDocsToPrefetch = Math.min(docs.length, prefetchDocsThreshold);
+
+        if (context.isPrefetchDocsEnabled()) {
+            // skipping prefetch for scroll queries
+            if (context.scrollContext() == null) {
```
kkewwei, line 190: Why do scroll queries not require preloading? Is it because they only run a single time? Logically, scroll queries handle a much more continuous stream of documents, so preloading ought to yield better results.

server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
```diff
@@ -174,8 +177,55 @@ public void execute(SearchContext context, String profileDescription) {
...
+                        if (currentReaderIndex != readerIndex) {
+                            currentReaderContext = context.searcher().getIndexReader().leaves().get(readerIndex);
+                        }
+                        int relativeDocId = docId - currentReaderContext.docBase;
+                        int rootId = findRootDocumentIfNested(context, currentReaderContext, relativeDocId);
+                        // skipping prefetch for nested docs as it needs additional processing like
+                        // loading the document source and filtering it based on the nested document ID
+                        if (rootId == -1) {
+                            // for sequential docs within the same segment we only prefetch the first doc
+                            if (currentReaderContext.reader() instanceof SequentialStoredFieldsLeafReader lf && hasSequentialDocs) {
```
kkewwei, line 207: When consecutive documents stretch beyond the scope of a single page_size, loading only one page will result in truncated output. Maybe we don’t have to worry about repeated loading operations, as @abiesps pointed out, Lucene will efficiently handle the quick reloading of the same data block on our behalf.

server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
kkhatua: @kkewwei  
Default page size is 4KB for Linux
Having the safety check is good for extreme cases anyway.

server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
kkhatua: @sruti1312 It might be good to have a test for this scenario ( I think your current tests might already be covering this implicitly)

server/src/main/java/org/opensearch/search/fetch/FetchPhase.java
```diff
@@ -174,8 +177,55 @@ public void execute(SearchContext context, String profileDescription) {
 
         int currentReaderIndex = -1;
         LeafReaderContext currentReaderContext = null;
+        boolean hasSequentialDocs = hasSequentialDocs(docs) && docs.length >= 10;
+
+        // Prefetch blocks containing the first N documents to ensure these blocks are loaded in advance.
+        // For most queries, which fetch up to 1000 docs, this improves cold-start latency
+        // without evicting useful cached blocks.
+        int prefetchDocsThreshold = 1000;
```
navneet1v, line 185: any reason for keeping this as 1K?

navneet1v:
@sruti1312 thanks for raising this PR. One question I have here is, when derived source is enabled then prefetch will not happen based on the current code. So we might need to build a separate mechanism to ensure that we call prefetch on the individual doc values which constructs the source when derived source is enabled.

kkewwei:
Same feature has been merged. #21285, Should we close the PR?

sruti1312:
Since this feature is already merged, closing this PR.

PR #20178, opened 2025-12-06
Fix peer recovery activity tracking

### LINKED ISSUE #20177: [BUG] Hung peer recovery permanently blocks replica allocation.
### Describe the bug

These were suggested by Codex after a multi hour analysis of pods
  - Environment: OpenSearch 3.3.0, three-node cluster, default cluster.routing.allocation.node_concurrent_{incoming,outgoing} of 2.
  - Steps: kill/restart a data node so its replicas need to recover; simulate a stuck recovery (e.g., block traffic between source and target). Observe that the shard
    **remains INITIALIZING forever,** consuming the only available recovery slot.
  - Expected: OpenSearch eventually marks the recovery failed and retries / lets other replicas proceed.
  - Actual: recovery stays stuck indefinitely, unassigned_shards never clear, operator must manually delete the shard or wipe the node.

Suggested fix:
  1. Auto-fail hung recoveries: Add a timeout to peer recoveries so that if no bytes transfer within (say) X minutes, OpenSearch marks the shard copy as failed and retries
     on another node. This would **free the node_concurrent_*_recoveries slots** and let other replicas progress without manual intervention.
  2. Heuristic for “no valid shard copy”: If a node rejoins with stale data and the cluster sees repeated NODE_LEFT → INITIALIZING loops, automatically wipe that shard
     copy (or at least flag it as unusable) so allocation can proceed. This mirrors what operators do manually by deleting PVCs.

### Related component

_No response_

### To Reproduce

There is no repro steps and it's hard to recover from such cases, would love to see a solution. 

### Expected behavior

The cluster recovers w/o manual intervention.

### Additional Details

N/A

maxlepikhin:
### Description

Prevent peer/segment recovery timeouts from being reset by mere status polling. Previously, every call to ReplicationCollection#get() bumped the recovery’s  lastAccessTime, so a hung recovery could sit in INITIALIZING indefinitely. This change removes that implicit update and explicitly touches the timestamp only when meaningful work happens (start, file transfer, clean‑files, translog, finalize, etc.), ensuring indices.recovery.recovery_activity_timeout actually fails stuck recoveries.

Ensure recovery/replication touch lastAccessTime only when handling real work: centralize the update in `ReplicationTarget.createOrFinishListener` (not on polling) and have explicit touches for the few non-listener entry points.

Also fix routing diff test to supply a fresh blob stream per read to avoid checksum/footer corruption.”

  ### Related Issues

  Resolves #20177

  ### Check List

  - [x] Functionality includes testing. (./gradlew :server:test --tests org.opensearch.recovery.ReplicationCollectionTests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Recovery and replication flows now refresh targets' last-access timestamps when actively used; passive collection lookups no longer update last-access to prevent misleading activity or timeout resets.
  * Fixes hung peer recovery that could permanently block replica allocation.

* **Tests**
  * Added tests verifying last-access update behavior and that polling doesn't reset recovery/replication timeouts.
  * Updated routing-table tests to create serialized inputs lazily during reads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Cleanup HttpServerTransport.Dispatcher in Netty tests ([#20160](https://github.com/opensearch-project/OpenSearch/pull/20160))
 
 ### Fixed
+- Fix hung peer recovery permanently blocks replica allocation ([#20177](https://github.com/opensearch-project/OpenSearch/issues/20177))
 - Fix bug of warm index: FullFileCachedIndexInput was closed error ([#20055](https://github.com/opensearch-project/OpenSearch/pull/20055))
 - Fix flaky test ClusterMaxMergesAtOnceIT.testClusterLevelDefaultUpdatesMergePolicy ([#18056](https://github.com/opensearch-project/OpenSearch/issues/18056))
 - Fix bug in Assertion framework(Yaml Rest test): numeric comparison fails when comparing Integer vs Long (or Float vs Double) ([#19376](https://github.com/opensearch-project/OpenSearch/pull/19376))


andrross:
@mch2 Can you take a look at this?

server/src/main/java/org/opensearch/indices/replication/common/ReplicationCollection.java
```diff
@@ -310,7 +310,6 @@ public static class ReplicationRef<T extends ReplicationTarget> extends AutoClos
          */
         public ReplicationRef(T status) {
             super(status);
-            status.setLastAccessTime();
```
andrross, line 313: In general I like this change. I would argue it's an anti-pattern to mutate the underlying object by simply constructing the wrapper. However, how can we have confidence that we're properly setting the last access time in all cases where necessary and not actually introducing a regression somewhere?

maxlepikhin: +1 to the question of how to guarantee that all cases are covered. I was hoping a person who is familiar with the code base may suggest an alternative, happy to follow up and implement. The change was created by Codex, it analyzed the code for any potential misses.

maxlepikhin: Alright, adjusted the PR to handle timestamp updates centrally in createOrFinishListener.

Most recovery handlers go through createOrFinishListener, which now touches lastAccessTime only when handling a real request. The remaining explicit touches stay because those paths don’t use that helper:
- The initial doRecovery setup,
- The primary-context handoff (simple handoff + empty response)
- And the segment replication kickoff (local orchestration, not a transport handler)

 They’re still real work, so we touch them explicitly; ordinary polling via ReplicationCollection#get doesn’t bump the timestamp.

andrross:
@maxlepikhin You've got a code formatting error:

```

* What went wrong:
      src/test/java/org/opensearch/cluster/routing/remote/RemoteRoutingTableServiceTests.java
          @@ -573,11 +573,13 @@
           ············compressor,
           ············Version.CURRENT
           ········);
          -········when(blobContainer.readBlob(indexName)).thenAnswer(invocation·->·remoteIndexRoutingTable.indexRoutingTableFormat.serialize(
          -············clusterState.getRoutingTable().getIndicesRouting().get(indexName),
          -············uploadedFileName,
          -············compressor
          -········).streamInput());
          +········when(blobContainer.readBlob(indexName)).thenAnswer(
          +············invocation·->·remoteIndexRoutingTable.indexRoutingTableFormat.serialize(
          +················clusterState.getRoutingTable().getIndicesRouting().get(indexName),
          +················uploadedFileName,
          +················compressor
          +············).streamInput()
          +········);
           ········TestCapturingListener<IndexRoutingTable>·listener·=·new·TestCapturingListener<>();
           ········CountDownLatch·latch·=·new·CountDownLatch(1);
           
````

**Comment by maxlepikhin** [comment]
> @maxlepikhin You've got a code formatting error:
> 
> ```
> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> Execution failed for task ':server:spotlessJavaCheck'.
> > The following files had format violations:
>       src/test/java/org/opensearch/cluster/routing/remote/RemoteRoutingTableServiceTests.java
>           @@ -573,11 +573,13 @@
>            ············compressor,
>            ············Version.CURRENT
>            ········);
>           -········when(blobContainer.readBlob(indexName)).thenAnswer(invocation·->·remoteIndexRoutingTable.indexRoutingTableFormat.serialize(
>           -············clusterState.getRoutingTable().getIndicesRouting().get(indexName),
>           -············uploadedFileName,
>           -············compressor
>           -········).streamInput());
>           +········when(blobContainer.readBlob(indexName)).thenAnswer(
>           +············invocation·->·remoteIndexRoutingTable.indexRoutingTableFormat.serialize(
>           +················clusterState.getRoutingTable().getIndicesRouting().get(indexName),
>           +················uploadedFileName,
>           +················compressor
>           +············).streamInput()
>           +········);
>            ········TestCapturingListener<IndexRoutingTable>·listener·=·new·TestCapturingListener<>();
>            ········CountDownLatch·latch·=·new·CountDownLatch(1);
>            
>   Run './gradlew spotlessApply' to fix all violations.
> ```

Thanks, updated.

File: server/src/main/java/org/opensearch/indices/recovery/PeerRecoveryTargetService.java [hunk 2]
```diff
@@ -397,6 +398,7 @@ class PrepareForTranslogOperationsRequestHandler implements TransportRequestHand
         public void messageReceived(RecoveryPrepareForTranslogOperationsRequest request, TransportChannel channel, Task task) {
             try (ReplicationRef<RecoveryTarget> recoveryRef = onGoingRecoveries.getSafe(request.recoveryId(), request.shardId())) {
                 final RecoveryTarget recoveryTarget = recoveryRef.get();
+                recoveryTarget.setLastAccessTime();
```
mch2: from your comment in https://github.com/opensearch-project/OpenSearch/pull/20178/changes#r2621171078, We can remove the explicit call in these handlers as well?

maxlepikhin: I don't believe so, per Codex' analysis the additional calls are needed as the listener is not used by all.

maxlepikhin: Verified with Codex:
```
  - All peer/segment recovery transport handlers route through ReplicationTarget#createOrFinishListener(...), which bumps the timestamp when the listener is non-null (real work).
  - The few non-transport paths still touch it explicitly: initial recovery setup, primary-context handoff, and segment replication orchestration.
  - Plain lookups via ReplicationCollection#get* no longer extend liveness, which is what we wanted to avoid.
```

mch2: line 402 below invokes `recoveryTarget.createOrFinishListener`, RecoveryTarget is not overriding this method?  This isn't initial recovery setup or handoff.

maxlepikhin: Thank you. Removed redundant setLastAccessTime() calls in PeerRecoveryTargetService before createOrFinishListener.

andrross:
I rebased and added a CHANGELOG entry

maxlepikhin:
> I rebased and added a CHANGELOG entry

@andrross looks like the CI failed, ownership = DCO or something else?

andrross:
@maxlepikhin DCO is fine, looks like some sort of connection failure on test setup. I'll retry the tests.

Did you answer @mch2's question above?

> Your logic makes sense, though I'm curious what is polling/resetting the last access time here. The internal monitor is calling get on ConcurrentHashMap [here](https://github.com/opensearch-project/OpenSearch/blob/5ced78c3d11fcfcefb495714bc9ca98c27b63c73/server/src/main/java/org/opensearch/indices/replication/common/ReplicationCollection.java#L336) - so that shouldn't be resetting the timestamp.
> 
> api calls to cat recovery also wouldn't reset this time as its fetching and reading recoverystate.

Did you confirm that this fixes the problem you discovered?

maxlepikhin:
> Your logic makes sense, though I'm curious what is polling/resetting the last access time here. The internal monitor is calling `get` on ConcurrentHashMap [here](https://github.com/opensearch-project/OpenSearch/blob/5ced78c3d11fcfcefb495714bc9ca98c27b63c73/server/src/main/java/org/opensearch/indices/replication/common/ReplicationCollection.java#L336) - so that shouldn't be resetting the timestamp.
> 
> api calls to cat recovery also wouldn't reset this time as its fetching and reading recoverystate.

@mch2 
Here is Codex' reply (I asked it to find concrete examples of the call sites). It makes sense to me.
```shell
• You’re right that the monitor and cat recovery don’t mutate it. The resets were coming from the ReplicationRef constructor calling setLastAccessTime(), so every lookup bumped the clock—
  even if no work followed. Examples:

  - PeerRecoveryTargetService handlers (e.g., file chunk at server/src/main/java/org/opensearch/indices/recovery/PeerRecoveryTargetService.java:578-582, translog ops at :447-459) did
    getSafe(...) before checking for duplicates; that lookup alone refreshed the timestamp.
  - Segment replication runner (server/src/main/java/org/opensearch/indices/replication/SegmentReplicator.java:322-333) bumped it just by fetching the ref.

  We removed that implicit bump and now update lastAccessTime only when we actually handle the request (createOrFinishListener when it returns non-null, plus the few non-transport paths like
  doRecovery). Polling via get/getSafe no longer resets the timer, which fixes the observed timeout issue.

```

maxlepikhin:
> @maxlepikhin DCO is fine, looks like some sort of connection failure on test setup. I'll retry the tests.
> 
> Did you answer @mch2's question above?
> 
> > Your logic makes sense, though I'm curious what is polling/resetting the last access time here. The internal monitor is calling get on ConcurrentHashMap [here](https://github.com/opensearch-project/OpenSearch/blob/5ced78c3d11fcfcefb495714bc9ca98c27b63c73/server/src/main/java/org/opensearch/indices/replication/common/ReplicationCollection.java#L336) - so that shouldn't be resetting the timestamp.
> > api calls to cat recovery also wouldn't reset this time as its fetching and reading recoverystate.
> 
> Did you confirm that this fixes the problem you discovered?

1. Replied to @mch2 , missed that comment, apologies.
2. Re repro, spent some time trying to get a cluster in minikube into that condition with network policy or iptables unsuccessfully. So no repro to test the fix.

andrross:
@mch2 Can you take another look here?

mch2:
I'm hesitant with this change if we don't have a concrete test showing this fixes stuck replications, not simply that the time has/hasn't incremented.  The change to `createOrFinishListener` would theoretically prevent the last access time from being bumped on requests the target has already received.  However, the source would eventually hit a retry limit if that were the case.  Will see if i can try a quick test to repro.  

A few other questions:
1. are you using segrep in your cluster?
2. have you changed any of the retry settings? ie `indices.recovery.retry_delay_network` `indices.recovery.internal_action_timeout`?
3. Any other exceptions in your logs?

andrross:
> I'm hesitant with this change if we don't have a concrete test showing this fixes stuck replications

I tend to agree. This change is a pretty low-level change in behavior where there's a risk of an unintended regression. I would prefer to not to make a change like this until we have a clear reproduction of the problem and verification of this fix.

maxlepikhin:
> I'm hesitant with this change if we don't have a concrete test showing this fixes stuck replications, not simply that the time has/hasn't incremented. The change to `createOrFinishListener` would theoretically prevent the last access time from being bumped on requests the target has already received. However, the source would eventually hit a retry limit if that were the case. Will see if i can try a quick test to repro.
> 
> A few other questions:
> 
> 1. are you using segrep in your cluster?
[ML] No.
> 2. have you changed any of the retry settings? ie `indices.recovery.retry_delay_network` `indices.recovery.internal_action_timeout`?
[ML] No special retry settings.
> 3. Any other exceptions in your logs?
[ML] There was a failure to join, not any other that I can find from internal slack or other sources.

Logically the change makes sense to me - only update the timestamp when there is recovery work progress, if no progress fail the recovery so it can be retried. The instance of stuck recovery in December of 2025 was not the only one.

maxlepikhin:
@mch2 @andrross see [the issue](https://github.com/opensearch-project/OpenSearch/issues/20177), I've added new evidence, let me know if you have any suggestions for what diagnostics to run, I will keep the cluster running for today 02/10/26 only. Now, there are logs from the 3 replicas which can be shared (perhaps in more private way).

maxlepikhin:
> @mch2 @andrross see [the issue](https://github.com/opensearch-project/OpenSearch/issues/20177), I've added new evidence, let me know if you have any suggestions for what diagnostics to run, I will keep the cluster running for today 02/10/26 only. Now, there are logs from the 3 replicas which can be shared (perhaps in more private way).

@mch2 @andrross ping

Issue #19851, opened 2025-10-31
[BUG] LeafReader removes all SubReaderWrappers incase IndexWriter encounters a non aborting Exception


RS146BIJAY:
### Describe the bug

In OpenSearch, during [recovery flow](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/InternalEngine.java#L2763), documents are queried using a DirectoryReader to be replayed as operations during translog replay. Now incase OpenSearch IndexWriter encounters any non aborted exception, underlying segments can contain both hard and soft deleted documents. In this case, OpenSearch [unwraps](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/Lucene.java#L933) a LeafReader and [uses it](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/Lucene.java#L944) for querying documents which are replayed as operation for Translog replay.

Since this unwrapping process will remove all wrappers from LeafReader, it becomes an issue for features like DerivedSource where we rely on a [DerivedSourceDirectoryReader](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/index/DerivedSourceDirectoryReader.java#L56) wrapper over LeafReader to extract source for a document. For Derived Source enabled index, with wrapper removed, [a MissingHistoryOperationsException will be thrown](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/LuceneChangesSnapshot.java#L313) during Translog replay as source field will be null for the documents, failing recovery.

### Related component

Indexing, Search

### Expected behavior

DirectoryReader should not unwrap LeafReader incase IndexWriter encounters a non aborting Exception

RS146BIJAY:
This is an edge case scenario, but can cause recovery failure with derive source

RS146BIJAY:
Tagging folks on this issue: @msfroh , @rishabhmaurya , @jainankitk , @shwetathareja   for feedback.

msfroh:
This sounds to me like it's not a bug in OpenSearch (since the behavior has always been that way), but rather it's a problem with the implementation of derived source.

Derived source could probably have been implemented as a different form of `SourceLookup` instead of as a DirectoryReader wrapper.

RS146BIJAY:
[...]

PR #20193, opened 2025-12-09
LeafReader should not remove SubReaderWrappers incase IndexWriter encounters a non aborting Exception

RS146BIJAY:
### Description
In OpenSearch, during [recovery flow](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/InternalEngine.java#L2763), documents are queried using a DirectoryReader to be replayed as operations during translog replay. Now incase OpenSearch IndexWriter encounters any non aborted exception, underlying segments can contain both hard and soft deleted documents. In this case, OpenSearch [unwraps](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/Lucene.java#L933) a LeafReader and [uses it](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/Lucene.java#L944) for querying documents which are replayed as operation for Translog replay.

Since this unwrapping process will remove all wrappers from LeafReader, it becomes an issue for features like DerivedSource where we rely on a [DerivedSourceDirectoryReader](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/index/DerivedSourceDirectoryReader.java#L56) wrapper over LeafReader to extract source for a document. For Derived Source enabled index, with wrapper removed, [a MissingHistoryOperationsException will be thrown](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/LuceneChangesSnapshot.java#L313) during Translog replay as source field will be null for the documents, failing recovery.

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/19851

### DIFF
--- a/server/src/main/java/org/opensearch/common/lucene/Lucene.java
+++ b/server/src/main/java/org/opensearch/common/lucene/Lucene.java
@@ -84,6 +84,7 @@ import org.apache.lucene.util.Version;
 import org.opensearch.ExceptionsHelper;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.SuppressForbidden;
+import org.opensearch.common.lucene.index.DerivedSourceLeafReader;
 import org.opensearch.common.lucene.search.TopDocsAndMaxScore;
 import org.opensearch.common.util.iterable.Iterables;
 import org.opensearch.core.common.Strings;
@@ -951,11 +952,35 @@ public class Lucene {
                     }
 
                     assert numDocs == popCount(hardLiveDocs) : numDocs + " != " + popCount(hardLiveDocs);
-                    return new LeafReaderWithLiveDocs(segmentReader, hardLiveDocs, numDocs);
+                    if (isDerivedSourceEnabled(leaf)) {
+                        return new LeafReaderWithLiveDocs(leaf, hardLiveDocs, numDocs);
+                    } else {
+                        return new LeafReaderWithLiveDocs(segmentReader, hardLiveDocs, numDocs);
+                    }
                 }
 
                 private boolean isContextAwareEnabled(SegmentReader reader) {
                     return reader.getSegmentInfo().info.getAttribute(CriteriaBasedCodec.BUCKET_NAME) != null;
+                }
+
+                /**
+                 * A FilterCodecReader can never accept a DerivedSourceLeafReader as a delegate as it is IndexReader.
+                 * DerivedSourceLeafReader can be wrapped up by only FilterLeafReader.
+                 * @param reader the underlying leafReader.
+                 *
+                 * @return whether derived source is enabled or not.
+                 */
+                public boolean isDerivedSourceEnabled(LeafReader reader) {
+                    if (reader instanceof SegmentReader) {
+                        return false;
+                    } else if (reader instanceof DerivedSourceLeafReader) {
+                        return true;
+                    } else if (reader instanceof FilterLeafReader) {
+                        FilterLeafReader filterLeafReader = (FilterLeafReader) reader;
+                        return isDerivedSourceEnabled(filterLeafReader.getDelegate());
+                    }
+
+                    return false;
                 }
             });
         }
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/opensearch/common/lucene/Lucene.java
```diff
@@ -951,7 +951,7 @@ public LeafReader wrap(LeafReader leaf) {
                     }
 
                     assert numDocs == popCount(hardLiveDocs) : numDocs + " != " + popCount(hardLiveDocs);
-                    return new LeafReaderWithLiveDocs(segmentReader, hardLiveDocs, numDocs);
+                    return new LeafReaderWithLiveDocs(leaf, hardLiveDocs, numDocs);
```
Bukhtawar: Do we need to do it selectively for derived source enabled cases, to reduce the impact of the change and is there an alternate way to put that check?

RS146BIJAY: We can just check if leaf instance is of type ```DerivedSourceLeafReader```.

RS146BIJAY: Addressed the comments.

server/src/main/java/org/opensearch/common/lucene/Lucene.java
```diff
@@ -951,12 +952,32 @@ public LeafReader wrap(LeafReader leaf) {
...
+                    } else {
+                        return new LeafReaderWithLiveDocs(segmentReader, hardLiveDocs, numDocs);
+                    }
                 }
 
                 private boolean isContextAwareEnabled(SegmentReader reader) {
                     return reader.getSegmentInfo().info.getAttribute(CriteriaBasedCodec.BUCKET_NAME) != null;
                 }
+
+                public boolean isDerivedSourceEnabled(LeafReader reader) {
```
atris, line 973: This implementation introduces a layering violation. This class is a low-level utility class and should not have dependencies on feature-specific implementations like `DerivedSourceLeafReader`. Hardcoding `instanceof` checks here breaks the Open-Closed Principle; if we add another wrapper that needs similar handling, we'd have to modify this core class again.

Can we instead use a marker interface (e.g., `PreserveOnUnwrap`) or a more generic abstraction? We need a way to signal "do not unwrap" without coupling the core engine to specific features.

```
public interface PreserveOnUnwrap {
    // Marker interface for LeafReaders that should not be implicitly stripped
}
```

Then, `DerivedSourceLeafReader` implements this interface and the check becomes:
`if (reader instanceof PreserveOnUnwrap) {`

RS146BIJAY: Thanks @atris for feedback. To give you some context, currently incase hard delete is present, we are removing all reader wrappers. I think the unwrapping done here seems like a bug and may not be caught till now because there is no chaos testing present in OpenSearch as hard deletes happens only when an Exception is thrown while Indexing in lucene: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/DocumentsWriterPerThread.java#L289. This will become an issue in general for any feature which adds a wrapper over default leaf reader like DerivedSourceLeafReader, which is dependent on leaf reader wrapper to recreate source. We restricted this unwrapping prevention to DerivedSourceLeafReader because we saw this feature becoming an issue only with this feature itself.

Another point to note down here is in case there is no hard deletes, which will be in most of the case, we are not unwrapping the LeafReader: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/lucene/Lucene.java#L936. This maybe the case why we were unable to catch the issue earlier as well. For eg, with derived source this becomes an issue when hard Deletes is present at the exact time when we are taking a SegmentInfoSnapshot during translog replay. Incase merges gets triggered before snapshot is taken, hard deletes may be removed and we will no longer see this issue.

Just wondering if we see this part becoming an issue for multiple such features, won't it make more sense to remove this check all together and not to unwrap the reader incase hard deletes are present. 

Let me know if this makes sense. I am ok to make your modifications in the PR.

atris: Agreed. Unwrapping to SegmentReader inherently breaks the FilterLeafReader contract for any wrapper that provides functionality (like stored fields) rather than just filtering.

Your proposal to wrap leaf directly is the correct generic fix. It solves the issue for `DerivedSource`
(and any future wrappers) without adding coupling or special cases to this class.

Please proceed with that change.

Issue #20062, opened 2025-11-20
[Feature Request] Proposal: Move Primary Terms and In-Sync Allocation IDs from Index Metadata to Routing Table


SwethaGuptha:
### Background

OpenSearch cluster state consists of two main components: metadata[[1]](https://github.com/SwethaGuptha/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java#L966-L978) and routing table[[2]](https://github.com/SwethaGuptha/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/routing/RoutingTable.java#L82). Currently, index metadata stores both user-configured data and internal shard management data (primary terms and in-sync allocation IDs) for an index, which creates unnecessary coupling between user configuration and internal routing operations.

```
Class IndexMetadata {

        final Index index,

        final long version,

        final long mappingVersion,

        final long settingsVersion,

        final long aliasesVersion,

        final long[] primaryTerms,

        ...
        ...

        final Map<Integer, Set<String>> inSyncAllocationIds,

        .....
}
```

**Primary Term:** A counter for each shard index which is incremented when the primary shard starts, fails, or when a replica is promoted to primary. It is used to determine shard freshness and prevent split-brain scenarios.

**Allocation ID:**  A UUID generated by the cluster manager when a shard copy is assigned to a node.

**In-Sync Allocation IDs:** Set of allocation IDs for each shard that are in-sync with the primary shard. The allocationIds are added/removed from the set when a shard on a node is started/failed/relocated due to node joins/leaves, and replica management. 

The primary term and in-sync allocation ids of shards are updated by the cluster manager in the reroute routing allocation rounds initiated due to shard shard/shard left/node drop tasks.

### Problem Statement

Primary terms and in-sync allocation IDs are internal routing metadata, not user configuration. Storing them in index metadata causes:

1. Tight  coupling between user configuration changes and internal shard  operations
2. Unnecessary  cluster state updates when only routing information changes
3. Conceptual  misalignment - routing data stored in metadata layer

### Proposed Solution

In-order to make the modelling decision we also need to understand how the cluster state is persisted and recovered currently.

#### Cluster state Persistence

When the cluster recovers, the last cluster state can be recovered by loading cluster state persisted on disk/remote

[...]

PR #20194, opened 2025-12-09
Introduce primaryTerm and inSynAllocationIds in IndexRoutingTable

SwethaGuptha:
### Description
First PR for https://github.com/opensearch-project/OpenSearch/issues/20062: Introduce primaryTerms and inSyncAllocationIds fields in IndexRoutingTable  to make IndexRoutingTable the source of truth for these fields instead of IndexMetadata.

Changes
- Added primaryTerms and inSyncAllocationIds fields
- Added getters, builder methods, serialization (version-gated with Version.V_3_4_0), validation, equals/hashCode/toString
- Validation is lenient (only validates if fields are populated)

**Not Included**
- Remote publication cluster diff serialization changes.
- Fields remain unpopulated (null/empty by default). Future PR will populate these fields from IndexMetadata and migrate read locations to use IndexRoutingTable as the source of truth.

### Related Issues

### Check List
- [Y] Functionality includes testing - Existing test suite.
- ~~[ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.~~
- ~~[ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.~~

## Summary by CodeRabbit

* **Refactor**
  * Cluster routing now tracks and exposes primary terms and in-sync allocation IDs to improve shard-state visibility and management.
  * Validation and serialization updated for compatibility with version 3.4.0 and later, and diagnostic output now includes the new routing metadata.

### DIFF
--- a/server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
+++ b/server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
@@ -33,8 +33,10 @@
 package org.opensearch.cluster.routing;
 
 import org.apache.lucene.util.CollectionUtil;
+import org.opensearch.Version;
 import org.opensearch.cluster.AbstractDiffable;
 import org.opensearch.cluster.Diff;
+import org.opensearch.cluster.DiffableUtils;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.cluster.metadata.Metadata;
 import org.opensearch.cluster.routing.RecoverySource.EmptyStoreRecoverySource;
@@ -43,6 +45,7 @@ import org.opensearch.cluster.routing.RecoverySource.LocalShardsRecoverySource;
 import org.opensearch.cluster.routing.RecoverySource.PeerRecoverySource;
 import org.opensearch.cluster.routing.RecoverySource.RemoteStoreRecoverySource;
 import org.opensearch.cluster.routing.RecoverySource.SnapshotRecoverySource;
+import org.opensearch.cluster.routing.allocation.IndexMetadataUpdater;
 import org.opensearch.common.Randomness;
 import org.opensearch.common.annotation.PublicApi;
 import org.opensearch.core.common.io.stream.BufferedChecksumStreamOutput;
@@ -51,16 +54,20 @@ import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.common.io.stream.VerifiableWriteable;
 import org.opensearch.core.index.Index;
 import org.opensearch.core.index.shard.ShardId;
+import org.opensearch.index.seqno.SequenceNumbers;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.TreeSet;
 import java.util.function.Predicate;
 
 /**
@@ -95,7 +102,20 @@ public class IndexRoutingTable extends AbstractDiffable<IndexRoutingTable>
 
     private final List<ShardRouting> allActiveShards;
 
+    private final long[] primaryTerms;
+
+    private final Map<Integer, Set<String>> inSyncAllocationIds;
+
     IndexRoutingTable(Index index, final Map<Integer, IndexShardRoutingTable> shards) {
+        this(index, shards, null, null);
+    }
+
+    IndexRoutingTable(
+        Index index,
+        final Map<Integer, IndexShardRoutingTable> shards,
+        final long[] primaryTerms,
+        final Map<Integer, Set<String>> inSyncAllocationIds
+    ) {
         this.index = index;
         this.shuffler = new RotationShardShuffler(Randomness.get().nextInt());
         this.shards = Collections.unmodifiableMap(shards);
@@ -108,6 +128,8 @@ public class IndexRoutingTable extends AbstractDiffable<IndexRoutingTable>
             }
         }
         this.allActiveShards = Collections.unmodifiableList(allActiveShards);
+        this.primaryTerms = primaryTerms;
+        this.inSyncAllocationIds = inSyncAllocationIds != null ? Collections.unmodifiableMap(inSyncAllocationIds) : Collections.emptyMap();
     }
 
     /**
@@ -139,6 +161,30 @@ public class IndexRoutingTable extends AbstractDiffable<IndexRoutingTable>
                 expected.remove(indexShardRoutingTable.shardId().id());
             }
             throw new IllegalStateException("Wrong number of shards in routing table, missing: " + expected);
+        }
+
+        // validate primaryTerms - only if populated
+        if (primaryTerms != null && primaryTerms.length > 0 && primaryTerms.length != indexMetadata.getNumberOfShards()) {
+            throw new IllegalStateException(
+                "Primary terms length ["
+                    + primaryTerms.length
+                    + "] does not match number of shards ["
+                    + indexMetadata.getNumberOfShards()
+                    + "]"
+            );
+        }
+
+        // validate inSyncAllocationIds - only if populated
+        if (inSyncAllocationIds != null
+            && !inSyncAllocationIds.isEmpty()
+            && inSyncAllocationIds.size() != indexMetadata.getNumberOfShards()) {
+            throw new IllegalStateException(
+                "InSyncAllocationIds size ["
+                    + inSyncAllocationIds.size()
+                    + "] does not match number of shards ["
+                    + indexMetadata.getNumberOfShards()
+                    + "]"
+            );
         }
 
         boolean isSearchOnlyClusterBlockEnabled = indexMetadata.getSettings()
@@ -261,6 +307,61 @@ public class IndexRoutingTable extends AbstractDiffable<IndexRoutingTable>
     }
 
     /**
+     * The term of the current selected primary. This is a non-negative number incremented when
+     * a primary shard is assigned after a full cluster restart or a replica shard is promoted to a primary.
+     * <p>
+     * Note: since we increment the term every time a shard is assigned, the term for any operational shard (i.e., a shard
+     * that can be indexed into) is larger than 0. See {@link IndexMetadataUpdater#applyChanges}.
+     **/
+    public long getPrimaryTerm(int shardId) {
+        if (primaryTerms == null) {
+            throw new IllegalStateException("Primary term not populated for index {} " + index);
+        }
+        if (primaryTerms.length < shardId) {
+            throw new IllegalArgumentException(
+                "Primary term not found for shard ["
+                    + index
+                    + "]["
+                    + shardId
+                    + "]. Primary term tracked for "
+                    + primaryTerms.length
+                    + "shards."
+            );
+        }
+
+        return this.primaryTerms[shardId];
+    }
+
+    public long[] getPrimaryTerms() {
+        if (primaryTerms == null) {
+            throw new IllegalStateException("Primary term not tracked for index {} " + index);
+        }
+        return this.primaryTerms.clone();
+    }
+
+    public Set<String> getInSyncAllocationIds(int shardId) {
+        if (inSyncAllocationIds.size() < shardId) {
+            throw new IllegalArgumentException(
+                "InSyncAllocationIds not found for shard ["
+                    + index
+                    + "]["
+                    + shardId
+                    + "]. InSyncAllocationIds tracked for "
+                    + inSyncAllocationIds.size()
+                    + "shards."
+            );
+        }
+        return inSyncAllocationIds.get(shardId);
+    }
+
+    public Map<Integer, Set<String>> getInSyncAllocationIds() {
+        if (inSyncAllocationIds == null || inSyncAllocationIds.isEmpty()) {
+            throw new IllegalStateException("InSyncAllocationIds not tracked for index {} " + index);
+        }
+        return inSyncAllocationIds;
+    }
+
+    /**
      * Returns <code>true</code> if all shards are primary and active. Otherwise <code>false</code>.
      */
     public boolean allPrimaryShardsActive() {
@@ -356,7 +457,8 @@ public class IndexRoutingTable extends AbstractDiffable<IndexRoutingTable>
 
         if (!index.equals(that.index)) return false;
         if (!shards.equals(that.shards)) return false;
-
+        if (!Arrays.equals(primaryTerms, that.primaryTerms)) return false;
+        if (!Objects.equals(inSyncAllocationIds, that.inSyncAllocationIds)) return false;
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -95,7 +101,16 @@ public class IndexRoutingTable extends AbstractDiffable<IndexRoutingTable>
 
     private final List<ShardRouting> allActiveShards;
 
-    IndexRoutingTable(Index index, final Map<Integer, IndexShardRoutingTable> shards) {
+    private final long[] primaryTerms;
+
+    private final Map<Integer, Set<String>> inSyncAllocationIds;
+
+    IndexRoutingTable(
```
gargharsh3134, line 113: This class is a Public API, we need to maintain the constructor compatibility.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -108,6 +123,8 @@ public class IndexRoutingTable extends AbstractDiffable<IndexRoutingTable>
             }
         }
         this.allActiveShards = Collections.unmodifiableList(allActiveShards);
+        this.primaryTerms = primaryTerms;
+        this.inSyncAllocationIds = Collections.unmodifiableMap(inSyncAllocationIds);
     }
```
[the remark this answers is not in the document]
gargharsh3134: Please address this.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -260,6 +301,30 @@ public IndexShardRoutingTable shard(int shardId) {
...
 
+    /**
+     * The term of the current selected primary. This is a non-negative number incremented when
+     * a primary shard is assigned after a full cluster restart or a replica shard is promoted to a primary.
+     * <p>
+     * Note: since we increment the term every time a shard is assigned, the term for any operational shard (i.e., a shard
+     * that can be indexed into) is larger than 0. See {@link IndexMetadataUpdater#applyChanges}.
+     **/
+    public long getPrimaryTerm(int shardId) {
```
gargharsh3134, line 316: Should we maintain parity with the existing getter in IndexMetadata?
```
    public long primaryTerm(int shardId) {
        return this.primaryTerms[shardId];
    }
```

Also, since we are allowing null values in validate, can we have null checks?

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -260,6 +301,30 @@ public IndexShardRoutingTable shard(int shardId) {
...
+        return this.primaryTerms[shardId];
+    }
+
+    public long[] getPrimaryTerms() {
+        return this.primaryTerms.clone();
+    }
+
+    public Set<String> getInSyncAllocationIds(int shardId) {
+        assert shardId >= 0 && shardId < inSyncAllocationIds.size();
+        return inSyncAllocationIds.get(shardId);
```
gargharsh3134, line 354: Same, do we need null checks?

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -416,9 +521,61 @@ public static class Builder {
 
         private final Index index;
         private final Map<Integer, IndexShardRoutingTable> shards = new HashMap<>();
+        private long[] primaryTerms = null;;
+        private final Map<Integer, Set<String>> inSyncAllocationIds;
+        private int numberOfShards;
```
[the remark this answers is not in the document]
gargharsh3134: Please address this.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -811,13 +968,41 @@ public Builder addShard(ShardRouting shard) {
...
-            return new IndexRoutingTable(index, shards);
+            if (primaryTerms == null) {
+                initializePrimaryTerms();
+            } else if (primaryTerms.length != numberOfShards) {
+                throw new IllegalStateException(
+                    "primaryTerms length is [" + primaryTerms.length + "] but should be equal to number of shards [" + numberOfShards + "]"
+                );
+            }
+
+            // fill missing slots in inSyncAllocationIds with empty set if needed and make all entries immutable
```
gargharsh3134, line 1017: When can this case arise? Should we log an error here?

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -381,6 +459,16 @@ public static IndexRoutingTable readFrom(StreamInput in) throws IOException {
             builder.addIndexShard(IndexShardRoutingTable.Builder.readFromThin(in, index));
         }
 
+        if (in.getVersion().onOrAfter(Version.V_3_4_0)) {
```
gargharsh3134, line 498: I believe branch for 3.4 has already been cut. Do we need to change checks to 3.5?

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -381,6 +459,16 @@ public static IndexRoutingTable readFrom(StreamInput in) throws IOException {
...
+            builder.setPrimaryTerms(in.readLongArray());
+            int inSyncAllocationIdsSize = in.readVInt();
+            for (int i = 0; i < inSyncAllocationIdsSize; i++) {
+                int shardId = in.readVInt();
+                Set<String> allocationIds = DiffableUtils.StringSetValueSerializer.getInstance().read(in, shardId);
+                builder.setInSyncAllocationIds(shardId, allocationIds);
+            }
+        }
+
         return builder.build();
```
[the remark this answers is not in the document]
gargharsh3134: Please address this, it will lead to failures during deserialization.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -395,11 +483,28 @@ public void writeTo(StreamOutput out) throws IOException {
...
+                DiffableUtils.StringSetValueSerializer.getInstance().write(cursor.getValue(), out);
+            }
+        }
     }
 
     public void writeVerifiableTo(BufferedChecksumStreamOutput out) throws IOException {
         index.writeTo(out);
+        // TODO: checksum calculation
         out.writeMapValues(shards, (stream, value) -> IndexShardRoutingTable.Builder.writeVerifiableTo(value, stream));
+        if (out.getVersion().onOrAfter(Version.V_3_4_0)) {
```
gargharsh3134, line 538: Same, lets verify the version checks.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -395,11 +483,28 @@ public void writeTo(StreamOutput out) throws IOException {
...
+            for (final Map.Entry<Integer, Set<String>> cursor : inSyncAllocationIds.entrySet()) {
+                out.writeVInt(cursor.getKey());
+                DiffableUtils.StringSetValueSerializer.getInstance().write(cursor.getValue(), out);
+            }
+        }
     }
 
     public void writeVerifiableTo(BufferedChecksumStreamOutput out) throws IOException {
         index.writeTo(out);
+        // TODO: checksum calculation
```
gargharsh3134: For my understanding, why do we need checksum calculation while writing here, isn't that auto calculated?

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -416,9 +521,61 @@ public static class Builder {
...
+        }
+
+        private void initializePrimaryTerms() {
+            assert primaryTerms == null;
+            if (numberOfShards < 0) {
+                throw new IllegalStateException("you must set the number of shards before setting/reading primary terms");
+            }
+            primaryTerms = new long[numberOfShards];
+            Arrays.fill(primaryTerms, SequenceNumbers.UNASSIGNED_PRIMARY_TERM);
+        }
```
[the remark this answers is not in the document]
gargharsh3134: Lets update the checks to use:
`if (numberOfShards <= 0)`

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -416,9 +521,61 @@ public static class Builder {
...
+            this.inSyncAllocationIds.putAll(inSyncAllocationIds);
+            return this;
+        }
+
+        public Set<String> getInSyncAllocationIds(int shardId) {
+            return inSyncAllocationIds.get(shardId);
+        }
+
+        public Builder setNumberOfShards(int numberOfShards) {
+            this.numberOfShards = numberOfShards;
```
gargharsh3134, line 615: Do we even require this member variable? Its easy to miss out on setting this in ser/de. Besides, this anyway doesn't get set from IndexMetadata, so it seems like we will end up asserting the same value which we set using array's length.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -356,20 +421,33 @@ public boolean equals(Object o) {
...
         return true;
     }
 
     @Override
     public int hashCode() {
         int result = index.hashCode();
         result = 31 * result + shards.hashCode();
+        result = 31 * result + Arrays.hashCode(primaryTerms);
+        result = 31 * result + inSyncAllocationIds.hashCode();
         return result;
```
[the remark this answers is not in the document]
SwethaGuptha: Handled inSyncAllocationIds in constructor.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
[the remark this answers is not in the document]
SwethaGuptha: Added the null check for now. For context, inSyncAllocationIds can only be null for the case of Remote routing table diff downloads where diff of all indices routing table is shared in one file. As part of this PR we are not handling the remote publication diff downloads and will be addressed separately once the change is validated end to end in transport mode.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -260,6 +301,30 @@ public IndexShardRoutingTable shard(int shardId) {
...
+    }
+
+    public long[] getPrimaryTerms() {
+        return this.primaryTerms.clone();
+    }
+
+    public Set<String> getInSyncAllocationIds(int shardId) {
+        assert shardId >= 0 && shardId < inSyncAllocationIds.size();
+        return inSyncAllocationIds.get(shardId);
+    }
```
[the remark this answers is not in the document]
SwethaGuptha: Addressed.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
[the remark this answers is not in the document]
SwethaGuptha: Addressed.

server/src/main/java/org/opensearch/cluster/routing/RoutingTableIncrementalDiff.java
```diff
@@ -114,7 +115,8 @@ public IndexRoutingTableIncrementalDiff(StreamInput in) throws IOException {
 
         @Override
         public IndexRoutingTable apply(IndexRoutingTable part) {
-            return new IndexRoutingTable(index, indexShardRoutingTables.apply(part.getShards()));
+            // TODO: fix this for remote publication
+            return new IndexRoutingTable(index, indexShardRoutingTables.apply(part.getShards()), null, null);
         }
```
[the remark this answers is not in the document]
SwethaGuptha: Remote publication diff ser/de is not being handled as part of this PR. Added a todo for it and handled NPE while reading these fields.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
SwethaGuptha: Thanks, I have reverted the original constructor and added a new constructor for additional fields.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
SwethaGuptha: It's okay to leave it at 3.4 for now. We will not merge this PR now until the remaining PRs are published. All the PRs will be merged together, we can update the version at the time of merge to main.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
SwethaGuptha: One valid case is initializing the in-sync when creating a new index. There might be some valid cases during snapshot/remote restore, dangling indices restore. Will add a error log later if required.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
SwethaGuptha: Replied in previous comment.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
SwethaGuptha: This will be set from IndexMetadata. Variable numberOfShards is required to configure the primaryTerms and in-sync allocation Ids size. I have handled the ser/de bug.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
SwethaGuptha: The writeVerifiableTo is used by remote checksum calculation and data order needs to be maintained for checksum to match. Since HashMap and sets don't maintain any specific order, a custom logic is required to ensure the order is maintained.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
SwethaGuptha: I don't see a specific reason to maintain this parity. Getters of other IndexRoutingTable class variables are prefixed with get, hence kept the getter naming convention within the class.

>Also, since we are allowing null values in validate, can we have null checks?


In the core the values will be null only for Remote downloads which is not addressed in scope of this CR.  Added check for NPE.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -260,6 +306,61 @@ public IndexShardRoutingTable shard(int shardId) {
...
+        }
+        return inSyncAllocationIds.get(shardId);
+    }
+
+    public Map<Integer, Set<String>> getInSyncAllocationIds() {
+        if (inSyncAllocationIds == null || inSyncAllocationIds.isEmpty()) {
+            throw new IllegalStateException("InSyncAllocationIds not tracked for index {} " + index);
+        }
+        return inSyncAllocationIds;
+    }
```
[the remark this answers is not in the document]
SwethaGuptha: Not relevant.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
gargharsh3134: I don't think that is the case. The underlying method is already sorting map keys for deterministic order and the existing classes too do not have any custom logic, please check if we are missing anything.

server/src/main/java/org/opensearch/cluster/routing/IndexRoutingTable.java
```diff
@@ -416,9 +559,61 @@ public static class Builder {
...
+        public long getPrimaryTerm(int shardId) {
+            if (primaryTerms == null) {
+                initializePrimaryTerms();
+            }
+            return this.primaryTerms[shardId];
+        }
+
+        private void initializePrimaryTerms() {
+            assert primaryTerms == null;
+            if (numberOfShards < 0) {
```
gargharsh3134, line 593: Shouldn't the check be <= 0? Is there any case when numberOfShards can be 0?

Issue #20224, opened 2025-12-12
[BUG] Netty4HttpChannelsReleaseIntegTests.testAcceptedChannelsGetCleanedUpOnTheNodeShutdown test needs fix?


sgup432:
### Describe the bug

I discovered an issue with this test while working on a different PR. It executes N search request and then assumes that there will be N open http channels. While it is working fine, but when I disabled telemetry setting, this test started to fail.

Seems like when telemetry is disabled, it uses the underlying netty http channel which is tied to per TCP connection rather than per request. So while we create N request, it only creates <N http channels, and reuses them. Which leads to this test failing.

When telemetry is enabled, we create `TracebaleHttpChannel` on top of underlying netty channel per request, which is why this test passes then.

### Related component

Other

### To Reproduce

In test `Netty4HttpChannelsReleaseIntegTests`, add below (ie disable telemetry)
```
protected Settings featureFlagSettings() {
        Settings.Builder featureSettings = Settings.builder().put(super.featureFlagSettings());
        featureSettings.put(FeatureFlags.TELEMETRY_SETTING.getKey(), false);
        return featureSettings.build();
    }
```

### Expected behavior

The test should pass with telemetry disabled as well.

### Additional Details

**Plugins**
Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**

**Additional context**

sgup432:
Pinging @reta and @fdesu on this one as they authored this. Let me know if my understanding is correct here.

reta:
> [@fdesu](https://github.com/fdesu)

Thanks for reporting it @sgup432 , I sent a fix https://github.com/opensearch-project/OpenSearch/pull/20225 , thank you

sgup432:
Thanks @reta for taking care of this!

PR #20225, opened 2025-12-12
Relax Netty4HttpChannelsReleaseIntegTests assertions

reta:
### Description
Relax Netty4HttpChannelsReleaseIntegTests assertions.

It is difficult to match the exact number of HTTP channels since:
 - there is 10 connections per route default setting (RestClient)
 - there are at least 2 nodes in the cluster (master + data)
 - if additional plugins are used (like telemetry), the same HTTP channel may be wrapped around, inflating the number of HTTP channels being tracked

We could modify RestCancellableNodeClient to wrap / unwrap channels but it does not justify changing the production code in order to perform such tricks for tests only.

### Related Issues
Closes https://github.com/opensearch-project/OpenSearch/issues/20224

### Check List

## Release Notes

* **Tests**
  * Relaxed HTTP channel count assertions in transport-related integration tests: strict equality checks were replaced with tolerant "greater-or-equal" validations and explanatory comments were added to account for runtime variations (default connections, multi-node setups, and instrumentation).

### DIFF
--- a/plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpChannelsReleaseIntegTests.java
+++ b/plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpChannelsReleaseIntegTests.java
@@ -30,6 +30,7 @@ import java.util.concurrent.TimeUnit;
 import static org.opensearch.test.hamcrest.OpenSearchAssertions.assertAcked;
 import static org.hamcrest.Matchers.anyOf;
 import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
 
 @ClusterScope(scope = OpenSearchIntegTestCase.Scope.TEST, supportsDedicatedMasters = false, numDataNodes = 1)
 public class ReactorNetty4HttpChannelsReleaseIntegTests extends OpenSearchReactorNetty4IntegTestCase {
@@ -66,8 +67,17 @@ public class ReactorNetty4HttpChannelsReleaseIntegTests extends OpenSearchReacto
         }
         countDownLatch.await();
 
-        // no channels get closed in this test, hence we expect as many channels as we created in the map
-        assertEquals("All channels remain open", initialHttpChannels + numChannels, RestCancellableNodeClient.getNumChannels());
+        // no channels get closed in this test, hence we expect as many channels as we created in the map at most
+        // it is difficult to match the exact number of HTTP channels since:
+        // - there is 10 connections per route default setting (RestClient)
+        // - there are at least 2 nodes in the cluster (master + data)
+        // - if additional plugins are used (like telemetry), the same HTTP channel may
+        // be wrapped around, inflating the number of HTTP channels being tracked
+        assertThat(
+            "All channels remain open",
+            RestCancellableNodeClient.getNumChannels(),
+            greaterThanOrEqualTo(initialHttpChannels + 10 /* default connections per route */)
+        );
     }
 
     /**


modules/transport-netty4/src/internalClusterTest/java/org/opensearch/transport/netty4/Netty4HttpChannelsReleaseIntegTests.java
```diff
@@ -66,8 +67,17 @@ public void testAcceptedChannelsGetCleanedUpOnTheNodeShutdown() throws Interrupt
...
+        // no channels get closed in this test, hence we expect as many channels as we created in the map at most
+        // it is difficult to match the exact number of HTTP channels since:
+        // - there is 10 connections per route default setting (RestClient)
+        // - there are at least 2 nodes in the cluster (master + data)
+        // - if additional plugins are used (like telemetry), the same HTTP channel may
+        // be wrapped around, inflating the number of HTTP channels being tracked
+        assertThat(
+            "All channels remain open",
+            RestCancellableNodeClient.getNumChannels(),
+            greaterThanOrEqualTo(initialHttpChannels + 10 /* default connections per route */)
```
reta, line 79: @fdesu wondering what do you think, I believe it is sufficient to make sure that we have more channels than before we started the test, the exact number may vary indeed, thank you.

fdesu: @reta thanks for a very detailed comment it makes total sense! The purpose of this test was, indeed, to make sure all acquired channels get released on the node shutdown. If integration tests acquired-channels-wise are independent (which they should, since the teardown rule kicks in after every test suite), asserting the number of channels at the end is higher than the initial should do the trick.

For the sake of clarity though, maybe we should either:
1. Force the test suite to use telemetry, which should force a new channel for each request and preserve the original assertion OR
2. simply do `greaterThan(initialHttpChannels)` since we want to make sure that some channels got acquired

Wdyt?

Also, any solution for this test suite applies to `ReactorNetty4HttpChannelsReleaseIntegTests`, right?

reta: Thanks @fdesu 

> 1. Force the test suite to use telemetry, which should force a new channel for each request and preserve the original assertion OR

I think we shouldn't dependent on presence of telemetry (or any other plugin), makes tests difficult to troubleshoot

> simply do greaterThan(initialHttpChannels) since we want to make sure that some channels got acquired

We could do that as well, the number 10 is actually very reliable measure here, but `greaterThan` will also do it. If you feel like `greaterThan` is enough, please suggest a change, I will apply it

> Also, any solution for this test suite applies to ReactorNetty4HttpChannelsReleaseIntegTests, right?

That is right, addressing it.

fdesu: I’m good with 10 and the greaterThanOrEqual, thanks!

sgup432: @reta Should we explicitly disable telemetry for this test/suite? Seems like telemetry is explicitly enabled for all the IT tests by default, which is why this test passed in the first place.

Alternatively, we could disable telemetry by default for all ITs, but that doesn’t seem like the right approach either.

andrross (approved):

modules/transport-netty4/src/internalClusterTest/java/org/opensearch/transport/netty4/Netty4HttpChannelsReleaseIntegTests.java
reta: @sgup432 we probably should not disable it but randomize it, like majority of our tests do. Testing with telemetry was / is super valuable to catch any missing instrumentation edge cases (not closed spans etc).

PR #20229, opened 2025-12-14
Fix OpenSearchNode working directory to avoid executing in immutable Gradle cache directory

### LINKED ISSUE #20226: [BUG] TestClusters plugin corrupts Gradle 9.x transform cache when canUseSharedDistribution() is true
### Describe the bug

When `canUseSharedDistribution()` returns true, OpenSearchNode runs directly from the Gradle artifact transform cache. OpenSearch writes gc.log files to its logs/ directory, which corrupts the immutable transform cache.

This was silently ignored in Gradle 8.5 and lower but fails in Gradle 8.6 and higher which validates transform cache immutability.

### Related component

Build

### To Reproduce

1. Clone the [`cross-cluster-replication`](https://github.com/opensearch-project/cross-cluster-replication) repository.
2. Run `bash gradlew integTest --no-daemon` on macOS or Linux
3. See error:

> FAILURE: Build failed with an exception.
> 
> * What went wrong:
> 
> Execution failed for task ':integTestRemote'.
> 
> > A build operation failed.
> 
> > The contents of the immutable workspace '/home/ci-runner/.gradle/caches/9.2.1/transforms/b3198452916980b6cbc53967c3593f45' have been modified. These workspace directories are not supposed to be modified once they are created. The modification might have been caused by an external process, or could be the result of disk corruption.

And the list of files in the validated directory includes
```
 - logs (Directory, 604b1c77ab560ca8636aef5e32829e77)
   - gc.log (RegularFile, 21e068d7b808e0222de187f02e99f0b4)
   - gc.log.00 (RegularFile, c4403ef5fef4d86ed83d5e3916aa057d)
```

### Expected behavior

Tests do not violate Gradle 9.x requirement for immutability of the workspace, and run normally (usually passing).

### Additional Details

**Plugins**

Only CCR plugin (this is a necessary condition to trigger the bug; failure does not occur when security plugin is installed, see below.)

 - OS: Either macOS or Linux.  failure does not occur on Windows (see below)
 - Version 3.4.0 / main, using Gradle 9.2.1 (also fails on 9.2.0)

**Additional context**

Lots and lots of details at https://github.com/gradle/gradle/issues/28475 (not the same issue but good technical discussion)

Gradle 8.6 added a requirement for immutability of the workspace, so this shared workspace has always been used, but as of Gradle 9.2+/JDK25 enforced with 3.4.0, this fails some CI without additional plugins

This is violated when a shared workspace is used and `getExtractedDistributionDir()` is used, as OpenSearch executes in that directory and creates JVM GC logs there.
https://github.com/opensearch-project/OpenSearch/blob/b09dcc9dc82b8c2e7ff8d2a67525422945f5a654/buildSrc/src/main/java/org/opensearch/gradle/testclusters/OpenSearchNode.java#L859-L863

The shared distribution is used when:
 - Not on Windows
 - No other plugins
 - No additional JAR files or modules
https://github.com/opensearch-project/OpenSearch/blob/b09dcc9dc82b8c2e7ff8d2a67525422945f5a654/buildSrc/src/main/java/org/opensearch/gradle/testclusters/OpenSearchNode.java#L590-L594

I was able to bypass the bug and succeed locally by forcing `canUseSharedDistribution()` to always return false:
```diff
   private boolean canUseSharedDistribution() {
-    // using original location can be too long due to MAX_PATH restrictions on windows CI
-    // TODO revisit when moving to shorter paths on CI by using Teamcity
-    return OS.current() != OS.WINDOWS && extraJarFiles.size() == 0 && modules.size() == 0 && plugins.size() == 0;
+    // Never use shared distribution to avoid corrupting Gradle 9.x immutable transform cache
+    // OpenSearch writes log files to the distribution directory, which violates cache immutability
+    return false;
   }
```

dbwiddis:
### Description

Modified OpenSearchNode binary execution directory from `getDistroDir()` to `workingDir.toFile()` so that all OpenSearch bin scripts run from the node's working directory instead of the Gradle cache directory.

Changed the Unix executable to use absolute paths since we changed the working directory.

### Related Issues
- Fixes #20226
- Supercedes #20228 which should be closed

### Check List

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed test cluster startup so tools run from the node working directory and use explicit script paths, preventing failures when executed in immutable or cached build directories.


buildSrc/src/main/java/org/opensearch/gradle/testclusters/OpenSearchNode.java
```diff
@@ -725,12 +725,17 @@ private void runOpenSearchBinScriptWithInput(String input, String tool, CharSequ
         try (InputStream byteArrayInputStream = new ByteArrayInputStream(input.getBytes(StandardCharsets.UTF_8))) {
             LoggedExec.exec(project, spec -> {
                 spec.setEnvironment(getOpenSearchEnvironment());
-                spec.workingDir(getDistroDir());
```
cwperks, line 728: When is `getDistroDir` resolved to the gradle cache? Is that in instances when running integTest against a testcluster that is using the `ARCHIVE` distribution?

I'm wondering if we should start to remove the restrictions on copying only the transport modules in the `INTEG_TEST` distribution and then start a campaign across the plugins to use `INTEG_TEST` which always pulls the latest commit from core without pulling a min distribution though I'm not sure how that would impact test speed. 

Let me try to dig up the comment about the distinction of ARCHIVE vs INTEG_TEST.

cwperks: Here was the comment: https://github.com/opensearch-project/query-insights/pull/441#issuecomment-3373948309

dbwiddis: > When is getDistroDir resolved to the gradle cache? Is that in instances when running integTest against a testcluster that is using the ARCHIVE distribution?

It's reproducible on `main` using ccr repo, so perhaps you can see if this change fixes it?

In any case, I'm not sure we want to execute in the distro dir and not the working dir...

dbwiddis:
Closing as this has been addressed by a different approach.

PR #20239, opened 2025-12-15
Add Rust loggers to vectorised-spi lib

abhita:
### Description
Add Rust Loggers library to `vectorized-exec-spi` shareable by `engine-datafusion` Plugin and `parquet-dataformat` Module

### Related Issues

### Check List

### DIFF
--- a/libs/vectorized-exec-spi/rust/src/logger.rs
+++ b/libs/vectorized-exec-spi/rust/src/logger.rs
@@ -11,28 +11,31 @@ use std::sync::OnceLock;
 
 static JAVA_VM: OnceLock<JavaVM> = OnceLock::new();
 
-// Log level constants (must match Java RustLoggerBridge constants)
-pub const LEVEL_TRACE: i32 = 0;
-pub const LEVEL_DEBUG: i32 = 1;
-pub const LEVEL_INFO: i32 = 2;
-pub const LEVEL_WARN: i32 = 3;
-pub const LEVEL_ERROR: i32 = 4;
+/// Log level enum (must match Java RustLoggerBridge.LogLevel ordinals)
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(i32)]
+pub enum LogLevel {
+    Debug = 0,
+    Info = 1,
+    Error = 2,
+}
 
-/// Initialize the logger with a JVM reference.
-/// This can be called multiple times safely - only the first call takes effect.
-pub fn init_logger(jvm: JavaVM) {
-    JAVA_VM.set(jvm).ok();
+impl LogLevel {
+    fn as_i32(self) -> i32 {
+        self as i32
+    }
 }
 
 /// Initialize the logger from a JNIEnv reference.
 /// This is a convenience function that extracts the JVM from the env.
 pub fn init_logger_from_env(env: &JNIEnv) {
     if let Ok(jvm) = env.get_java_vm() {
-        init_logger(jvm);
+        JAVA_VM.set(jvm).ok();
     }
 }
 
-pub fn log(level: i32, message: &str) {
+// Private log function - only used by macros
+fn log(level: LogLevel, message: &str) {
     if let Some(jvm) = JAVA_VM.get() {
         if let Ok(mut env) = jvm.attach_current_thread() {
             let result = (|| -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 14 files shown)


libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/RustLoggerBridge.java
```diff
@@ -0,0 +1,73 @@
...
+public class RustLoggerBridge {
+
+    private static final Logger logger = LogManager.getLogger(RustLoggerBridge.class);
+
+    /**
+     * Log an info level message from Rust code.
+     *
+     * @param message the message to log
+     */
+    public static void logInfo(String message) {
```
Bukhtawar: Instead use ` public void log(int level, String message)` ?

abhita: Makes sense - a single `log(int level, String message)` would simplify the code.
I initially went with separate methods for readability on the Rust side (`e.g., log_info! vs log(2, ...)`) and to avoid maintaining level constants across both languages.
How about `log(String level, String message)` ? keeping readability `("INFO", "DEBUG")` while still consolidating to a single method ?

abhita: Thinking of it, having `string` type as log-level would introduce more complexity to add extra `env.new_string()` allocation. Having an `int` loglevel would be concise but we would end up maintaining loglevel constants.

Let me make the changes

modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustLoggerBridge.java
```diff
@@ -8,27 +8,49 @@
...
-
+/**
+ * Bridge class that allows Rust code to log messages through Java's logging framework.
+ * This class delegates to the shared RustLoggerBridge in vectorized-exec-spi.
+ *
+ * @deprecated Use {@link org.opensearch.vectorized.execution.jni.RustLoggerBridge} directly.
+ *             This class is kept for backward compatibility with existing Rust JNI code.
+ */
+@Deprecated
 public class RustLoggerBridge {
```
raghuvanshraj: We shouldn't need this class anymore right? We should be able to reuse the RustLoggerBridge from the lib itself?

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -129,21 +133,27 @@ pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_initTokio
         env.get_java_vm().expect("Failed to get JavaVM")
     });
 
+    // Initialize the logger with the JVM for Rust->Java logging bridge
+    // Get a fresh JVM reference for the logger
+    if let Ok(jvm) = env.get_java_vm() {
+        logger::init_logger(jvm);
```
raghuvanshraj: I think this call should be centralized. It should come from OpenSearch when loading the lib.

raghuvanshraj: Should be okay for now, but if we want to make this class centralized across all rust plugins, we would likely want it to be initialized via OpenSearch.

modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustLoggerBridge.java
abhita: Yes. This was a miss. Will be posting the changes with the Log level changes

plugins/engine-datafusion/jni/src/lib.rs
abhita: For this, we would end up establishing dynamic linking instead of static shareable crate like `#[link(name = "opensearch_vectorized_spi")]` ..
Let me quickly guage the complexity involved to get this working

abhita: To establish dynamic linking, the flow would look like below
1. Remove the shareable crate
2. Build `vectprized-spec-spi` as `cdylib` and export C-compatible FFI functions
3. Initialize logger: `RustLoggerBridge.initNativeLogger()` within Opensearch when the lib is loaded
4. Add `#[link(name = "opensearch_vectorized_spi")]` FFI declarations for plugins/modules using the loggers
4. Share the same pointer across plugins and modules.

If we are good with establishing the dynamic linking, we can proceed with this

libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/RustLoggerBridge.java
abhita: Done

modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustLoggerBridge.java
abhita: Done

libs/vectorized-exec-spi/rust/src/logger.rs
```diff
@@ -0,0 +1,124 @@
...
+
+/// Initialize the logger from a JNIEnv reference.
+/// This is a convenience function that extracts the JVM from the env.
+pub fn init_logger_from_env(env: &JNIEnv) {
+    if let Ok(jvm) = env.get_java_vm() {
+        init_logger(jvm);
+    }
+}
+
+pub fn log(level: i32, message: &str) {
```
raghuvanshraj: Do we need this to be public?

libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/RustLoggerBridge.java
```diff
@@ -0,0 +1,68 @@
...
+    /** Log level constant for TRACE */
+    public static final int LEVEL_TRACE = 0;
+    /** Log level constant for DEBUG */
+    public static final int LEVEL_DEBUG = 1;
+    /** Log level constant for INFO */
+    public static final int LEVEL_INFO = 2;
+    /** Log level constant for WARN */
+    public static final int LEVEL_WARN = 3;
+    /** Log level constant for ERROR */
+    public static final int LEVEL_ERROR = 4;
```
raghuvanshraj: Better to create an enum for this?

libs/vectorized-exec-spi/rust/src/logger.rs
```diff
@@ -0,0 +1,124 @@
...
+
+/// Convenience function for logging at TRACE level
+pub fn log_trace(message: &str) {
+    log(LEVEL_TRACE, message);
+}
+
+// Macros for convenient logging with format! syntax
+
+/// Log at INFO level with format! syntax
+#[macro_export]
```
raghuvanshraj: Why macro?

libs/vectorized-exec-spi/rust/src/logger.rs
```diff
@@ -0,0 +1,124 @@
...
+    JAVA_VM.set(jvm).ok();
+}
+
+/// Initialize the logger from a JNIEnv reference.
+/// This is a convenience function that extracts the JVM from the env.
+pub fn init_logger_from_env(env: &JNIEnv) {
+    if let Ok(jvm) = env.get_java_vm() {
+        init_logger(jvm);
+    }
+}
```
raghuvanshraj: I don't think we need both of these?

libs/vectorized-exec-spi/rust/src/logger.rs
abhita: Added this to enable ease of formatting
instead of multiple `log(INFO, &format!("",xyz))` , defined a rule to accept as many arguments and format it accordingly

libs/vectorized-exec-spi/rust/src/logger.rs
abhita: Cleaning this up

libs/vectorized-exec-spi/rust/src/logger.rs
abhita: Right. Erasing the first one

libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/RustLoggerBridge.java
abhita: Noted

libs/vectorized-exec-spi/src/main/java/org/opensearch/vectorized/execution/jni/RustLoggerBridge.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Log a message at the specified level from Rust code.
+     * This method is called by Rust via JNI.
+     *
+     * @param level the log level ordinal (0=DEBUG, 1=INFO, 2=ERROR)
+     * @param message the message to log
+     */
+    public static void log(int level, String message) {
```
raghuvanshraj, line 49: ~Why are we not accepting the enum here?~

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/jni/NativeBridge.java
```diff
@@ -19,6 +19,7 @@ public final class NativeBridge {
 
     static {
         NativeLibraryLoader.load("opensearch_datafusion_jni");
+        initLogger();
```
alchemist51, line 22: Should initLogger() not be present in a vectorised-exec-spi?

abhita: Currently `vectorized-exec-spi` acts as a static dependency (share-able crate through path dependency) for both `datafusion` plugin and `dataformat` module. 
In-order to do this, we would have to establish dynamic linking which would involve lot more changes as documented in the conversation - https://github.com/opensearch-project/OpenSearch/pull/20239#discussion_r2621906401

To make things, simple we went ahead with `SetOnce()` procedure so that first one who sets the logger, the next one re-uses the same.

PR #20250, opened 2025-12-16
Prevent criteria update for context aware indices

RS146BIJAY:
### Description
This modification prevents updates to fields that determine a document's grouping criteria. By blocking such updates, we eliminate the need for complex version synchronization across multiple IndexWriters. This approach simplifies version management, since the latest version of a document will always reside in the group-specific writer marked for refresh, while the previous version may be either in old or parent IndexWriter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent grouping criteria updates for context-aware enabled indices. System now validates and blocks operations that would attempt to modify existing grouping criteria, ensuring consistency across index operations.

### DIFF
--- a/server/src/main/java/org/opensearch/common/lucene/uid/VersionsAndSeqNoResolver.java
+++ b/server/src/main/java/org/opensearch/common/lucene/uid/VersionsAndSeqNoResolver.java
@@ -32,13 +32,16 @@
 
 package org.opensearch.common.lucene.uid;
 
+import org.apache.lucene.index.FilterLeafReader;
 import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.index.LeafReader;
 import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.SegmentReader;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.util.CloseableThreadLocal;
 import org.opensearch.common.annotation.PublicApi;
 import org.opensearch.common.util.concurrent.ConcurrentCollections;
+import org.opensearch.index.codec.CriteriaBasedCodec;
 
 import java.io.IOException;
 import java.util.List;
@@ -155,7 +158,8 @@ public final class VersionsAndSeqNoResolver {
      * <li>a doc ID and a version otherwise
      * </ul>
      */
-    public static DocIdAndVersion loadDocIdAndVersion(IndexReader reader, Term term, boolean loadSeqNo) throws IOException {
+    public static DocIdAndVersion loadDocIdAndVersion(IndexReader reader, Term term, boolean loadSeqNo, String currentCriteria)
+        throws IOException {
         PerThreadIDVersionAndSeqNoLookup[] lookups = getLookupState(reader, term.field());
         List<LeafReaderContext> leaves = reader.leaves();
         // iterate backwards to optimize for the frequently updated documents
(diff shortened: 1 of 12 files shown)


server/src/main/java/org/opensearch/common/lucene/uid/VersionsAndSeqNoResolver.java
```diff
@@ -165,6 +170,18 @@ public static DocIdAndVersion loadDocIdAndVersion(IndexReader reader, Term term,
             PerThreadIDVersionAndSeqNoLookup lookup = lookups[leaf.ord];
             DocIdAndVersion result = lookup.lookupVersion(term.bytes(), loadSeqNo, leaf);
             if (result != null) {
+                if (result.version > 0 && currentCriteria != null) {
+                    SegmentReader unwrappedReader = (SegmentReader) (FilterLeafReader.unwrap(leaf.reader()));
+                    String prevCriteria = unwrappedReader.getSegmentInfo().info.getAttribute(CriteriaBasedCodec.BUCKET_NAME);
+                    assert prevCriteria != null;
+                    if (prevCriteria.equals(currentCriteria) == false) {
```
Bukhtawar, line 176: Can we use `UnsupportedOperationException` instead?

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -648,8 +648,23 @@ public boolean hasNewIndexingOrUpdates() {
...
-    String getCriteriaForDoc(BytesRef uid) {
-        return liveIndexWriterDeletesMap.getCriteriaForDoc(uid);
+    public String getCurrentCriteria(BytesRef uid) {
+        assertKeyedLockHeldByCurrentThread(uid);
+        return liveIndexWriterDeletesMap.current.getCriteriaForDoc(uid);
+    }
+
+    public String getCriteriaUnderLock(BytesRef uid) {
+        return getCriteriaUnderLock(uid, liveIndexWriterDeletesMap);
+    }
```
Bukhtawar: If the criteria is immutable do we need to really go through the lock?

server/src/main/java/org/opensearch/index/engine/DocumentIndexWriter.java
```diff
@@ -91,4 +92,6 @@ void deleteDocument(
     boolean isWriteLockedByCurrentThread();
 
     Releasable obtainWriteLockOnAllMap();
+
+    String getCriteriaUnderLock(BytesRef uid);
```
mgodwan: This method is not too aligned with the interface as criteria is a CompositeIndexWriter specific detail. Do we need the method to be defined here?

server/src/main/java/org/opensearch/common/lucene/uid/VersionsAndSeqNoResolver.java
```diff
@@ -165,6 +170,18 @@ public static DocIdAndVersion loadDocIdAndVersion(IndexReader reader, Term term,
...
+                        throw new ImmutableCriteriaException(
+                            "Updating grouping criteria is not allowed for context aware enabled indices",
+                            null
```
mgodwan: This only throws error if the bucket is changed (e.g. 4xx can't be changed to 5xx but a 404 can be changed to 429 if buckets are defined as 4xx & 5xx) We should document this behaviour.

server/src/main/java/org/opensearch/index/engine/DocumentIndexWriter.java
RS146BIJAY: Addressed the comments.

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
RS146BIJAY: Addressed the comments

Bukhtawar (approved):

Issue #20131, opened 2025-12-01
[RFC] Introduce concurrent translog recovery to accelerate segment replication primary promotion


guojialiang92:
## Background

The purpose of this RFC is to discuss the solution to the **long primary promotion time for segment replication** mentioned in https://github.com/opensearch-project/OpenSearch/issues/20118. Primary promotion may take up to `15` minutes or even longer. During this period, write throughput significantly decreases.

### Primary promotion takes a long time
- The test cluster includes `4 16U64G` data nodes, with segment replication and merged segment warmer enabled. Execute the following command.

```
opensearch-benchmark execute-test --target-hosts http://[fdbd:dc05:b:235::33]:9463 --workload=nyc_taxis --workload-params="search_clients: 4, bulk_indexing_clients: 32, number_of_shards:4, number_of_replicas: 1" --kill-running-processes
```

- After restarting the data-0 node, a write throughput decline lasting more than `15` minutes can be observed.

<img width="2826" height="1246" alt="Image" src="https://github.com/user-attachments/assets/c27201d9-2b65-434d-9179-fc24140f9364" />

- Through the logs, we can see that a total of `3,911,465` operations were recovered from the translog.

```
[2025-11-22T01:29:43,325][TRACE][o.o.i.e.Engine           ] [byte-es-test-native-data-1] [nyc_taxis][1] flushing post recovery from translog: ops recovered [3911465], current translog generation [1693]
```

### Reasons for the long translog recovery time
[In the segment replication scenario, replicas only record the translog](https://github.com/opensearch-project/OpenSearch/blob/1ee30dcc8334f63c51bed276397fa4d50406a8e6/server/src/main/java/org/opensearch/index/engine/NRTReplicationEngine.java#L242). The default refresh interval for the `nyc_taxis` is `30` seconds. In a production environment, a refresh interval of `30` to `60` seconds is also a common configuration. This means that any document that has not been synchronized through segment replication within the refresh interval will be recovered via the translog during the primary promotion process.

### Accelerate translog recovery
1. Add a dedicated thread pool `translog_recovery` for translog recovery.
2. Provides dynamic configuration at the index level to enable concurrent recovery of translog and specify the number of translog operations processed by a single thread.
[...]

PR #20251, opened 2025-12-16
Introduce concurrent translog recovery to accelerate segment replication primary promotion

guojialiang92:
### Description

The purpose of this PR is to introduce the translog concurrent recovery mechanism mentioned in https://github.com/opensearch-project/OpenSearch/issues/20131, which is used to accelerate the primary promotion of segment replication.

#### Default concurrency strategy

- Translog concurrent recovery is restricted to use in the primary promotion scenario of segment replication.
- Introduce a dedicated thread pool `translog_recovery` with a size equal to the number of cores and an unbounded queue.
- Introduce cluster-level dynamic configuration `indices.translog_concurrent_recovery.enable`, which is used to enable translog concurrent recovery, with a default value of `false`.
- Introduce cluster-level dynamic configuration `indices.translog_concurrent_recovery.batch_size`, which is used to represent the number of translog operations processed by a single thread, with a default value of `500,000`.
- When the number of translog recovery operations is less than `indices.translog_concurrent_recovery.batch_size`, concurrency is not enabled, remaining consistent with the current execution logic.
- When the number of translog recovery operations exceeds `indices.translog_concurrent_recovery.batch_size`.
   - If `translog_recovery` is not busy, enable concurrency, with each thread responsible for executing the number of operations specified by `indices.translog_concurrent_recovery.batch_size`.
   - If `translog_recovery` is busy, do not enable concurrency, keeping it consistent with the current execution logic.

### Related Issues
Resolves #[[20131](https://github.com/opensearch-project/OpenSearch/issues/20131)]

### Check List

## Summary by CodeRabbit

* **New Features**
  * Added dynamic cluster settings to enable concurrent translog recovery and to configure batch size.
  * Introduced a dedicated translog-recovery thread pool and batched parallel translog recovery to speed shard recovery when conditions allow, with automatic fallback to the sequential path.

* **Tests**
  * Added end-to-end tests validating concurrent translog recovery, primary promotion, replication, and translog/Lucene state consistency.

### DIFF
--- a/server/src/main/java/org/opensearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/opensearch/index/shard/IndexShard.java
@@ -399,6 +399,8 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
     private final ReferencedSegmentsPublisher referencedSegmentsPublisher;
     private final Set<MergedSegmentCheckpoint> pendingMergedSegmentCheckpoints = Sets.newConcurrentHashSet();
     private final MergedSegmentTransferTracker mergedSegmentTransferTracker;
+
+    // Used to limit the number of concurrent translog tasks. When the semaphore is exhausted, serial recovery is used.
     private static final Semaphore translogConcurrentRecoverySemaphore = new Semaphore(1000);
 
     @InternalApi


server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5274,14 +5276,63 @@ public void close() throws IOException {
...
-            snapshot,
-            Engine.Operation.Origin.LOCAL_RESET,
-            () -> {
-                // TODO: add a dedicate recovery stats for the reset translog
+        final TranslogRecoveryRunner translogRunner = (snapshot) -> {
+            Engine engine = newEngineReference.get();
+            assert null != engine;
+            int totalOperations = snapshot.totalOperations();
+            long batchSize = recoverySettings.getTranslogConcurrentRecoveryBatchSize();
+            boolean threadPoolNotBusy = ((OpenSearchThreadPoolExecutor) threadPool.executor(ThreadPool.Names.TRANSLOG_RECOVERY)).getQueue()
```
atris: This check is racy. `isEmpty() `only means no backlog now. In a recovery storm, multiple shards will pass this check and simultaneously dump tasks onto the shared pool. Suggest using a Semaphore.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5274,14 +5276,63 @@ public void close() throws IOException {
...
+                List<Future<Integer>> translogRecoveryFutureList = new ArrayList<>();
+                try {
+                    // Since the translog does not change at this time, it is safe to re-partition the translog snapshot here.
+                    for (int i = 0; i < batches; i++) {
+                        long start = localCheckpoint + 1 + (long) i * batchSize;
+                        long end = (i == batches - 1) ? Long.MAX_VALUE : start + batchSize - 1;
+                        Translog.Snapshot translogSnapshot = engine.translogManager().newChangesSnapshot(start, end, false);
+                        translogSnapshotList.add(translogSnapshot);
+                        translogRecoveryFutureList.add(
+                            threadPool.executor(ThreadPool.Names.TRANSLOG_RECOVERY)
```
atris: Could this OOM the node if `totalOperations `is high. Do we need to batch in flight requests with a semaphore?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5274,14 +5276,63 @@ public void close() throws IOException {
...
+                                }))
+                        );
+                    }
+                    Exception exception = null;
+                    int totalRecovered = 0;
+                    for (Future<Integer> translogRecoveryFuture : translogRecoveryFutureList) {
+                        try {
+                            int recoveredOps = translogRecoveryFuture.get();
+                            totalRecovered += recoveredOps;
+                        } catch (Exception e) {
```
atris, line 5331: If one batch fails, should we explicitly cancel all pending futures in `translogRecoveryFutureList`?

guojialiang92:
Thank you for your review @atris. I will make improvements. :)

server/src/main/java/org/opensearch/index/shard/IndexShard.java
guojialiang92: Done. 
It should be noted that `Future.cancel(true)` is disabled in OpenSearch.
If the future is already running, `FutureUtils#cancel` will not interrupt the thread. In this case, it is necessary to wait for the future to finish running.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
guojialiang92: Added `IndexShard#translogConcurrentRecoverySemaphore` to address the concurrency issue.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
guojialiang92: Added `IndexShard#translogConcurrentRecoverySemaphore`, with an upper limit of `1000`, to prevent memory OOM.

guojialiang92:
Hi, @atris.
I have already made improvements based on the suggestions. Could you please help review the code again?

server/src/main/java/org/opensearch/threadpool/ThreadPool.java
```diff
@@ -258,6 +260,7 @@ public ThreadPool(
         final int genericThreadPoolMax = boundedBy(4 * allocatedProcessors, 128, 512);
         final int snapshotDeletionPoolMax = boundedBy(4 * allocatedProcessors, 64, 256);
         builders.put(Names.GENERIC, new ScalingExecutorBuilder(Names.GENERIC, 4, genericThreadPoolMax, TimeValue.timeValueSeconds(30)));
+        builders.put(Names.TRANSLOG_RECOVERY, new FixedExecutorBuilder(settings, Names.TRANSLOG_RECOVERY, allocatedProcessors, -1));
```
atris, line 263: IMO, queue_size: -1 is unsafe for a recovery pool. During a node restart, this will absorb high number of tasks and potentially OOM the node. Using a bounded queue ) to force rejection/backpressure when the node is saturated might be prudent.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -394,6 +397,7 @@ Runnable getGlobalCheckpointSyncer() {
     private final ReferencedSegmentsPublisher referencedSegmentsPublisher;
     private final Set<MergedSegmentCheckpoint> pendingMergedSegmentCheckpoints = Sets.newConcurrentHashSet();
     private final MergedSegmentTransferTracker mergedSegmentTransferTracker;
+    private final Semaphore translogConcurrentRecoverySemaphore = new Semaphore(1000);
```
atris: Since this is an instance variable, it limits concurrency per shard, effectively allowing NumShards * 1000 concurrent tasks on the node. For this to protect the node, it must be static, or we should rely on the thread pool queue.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5274,14 +5278,84 @@ public void close() throws IOException {
...
+            if (isConcurrentRecovery) {
+                List<Translog.Snapshot> translogSnapshotList = new ArrayList<>();
+                List<Future<Integer>> translogRecoveryFutureList = new ArrayList<>();
+                try {
+                    // Since the translog does not change at this time, it is safe to re-partition the translog snapshot here.
+                    for (int i = 0; i < batches; i++) {
+                        long start = localCheckpoint + 1 + (long) i * batchSize;
+                        long end = (i == batches - 1) ? Long.MAX_VALUE : start + batchSize - 1;
+                        Translog.Snapshot translogSnapshot = engine.translogManager().newChangesSnapshot(start, end, false);
+                        translogSnapshotList.add(translogSnapshot);
```
atris: We shouldn't open all snapshots upfront. Snapshots hold file channel references and off-heap buffers. Opening multiple of them broadly across multiple shards will exhaust file descriptors. Suggest moving this inside the submitted task so they are opened lazily.

atris: Out-of-order completion forces the `LocalCheckpointTracker`
 to retain bitsets for all future batches until the stalled batch (e.g. Batch 1) completes. If Batch 1 stalls, we hold hundreds of MBs of processed checkpoint data in heap per shard.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5274,14 +5278,84 @@ public void close() throws IOException {
...
+                    }
+                    Exception exception = null;
+                    int totalRecovered = 0;
+                    for (Future<Integer> translogRecoveryFuture : translogRecoveryFutureList) {
+                        try {
+                            if (exception != null) {
+                                FutureUtils.cancel(translogRecoveryFuture);
+                            }
+                            if (false == translogRecoveryFuture.isCancelled()) {
+                                int recoveredOps = translogRecoveryFuture.get();
```
atris: This blocks the coordinating thread sequentially. If the first batch hangs, we don't notice failures in subsequent batches. Maybe use `ExecutorCompletionService` to consume results as they complete and cancel any remaining futures immediately on the first exception.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
guojialiang92: You're right. I limited the maximum value of `RecoverySettings#INDICES_TRANSLOG_CONCURRENT_RECOVERY_BATCH_SIZE` to `1` million to avoid excessive memory usage of the `bitset` in `LocalCheckpointTracker` during out-of-order execution of translog concurrent recovery.
Considering the worst-case scenario, with `1000` concurrent recovery tasks, each task recovering `1` million translogs, the bitset occupancy is approximately `125MB`.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
guojialiang92: Done.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
guojialiang92: Done.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
guojialiang92: Yes. When using a bounded queue, it is necessary to handle rejected cases, such as retrying. I prefer to use a static semaphore, and when the semaphore is exhausted, it can fall back to the serial processing mode.

server/src/main/java/org/opensearch/threadpool/ThreadPool.java
guojialiang92: Thanks @atris. It has been resolved using a static semaphore, with a default value of `1000`.

guojialiang92:
Thank you for your rigorous review @atris.

Based on your review suggestions, I have made the following optimizations.

1. I adjusted `IndexShard#translogConcurrentRecoverySemaphore` to be static to ensure that the queue size is limited to `1000`, and if the semaphore is exhausted, it can fall back to serial processing.
2. The maximum value of `RecoverySettings#INDICES_TRANSLOG_CONCURRENT_RECOVERY_BATCH_SIZE` is limited to ensure that memory usage is controllable under the worst-case scenario, and explanations have also been added to the code.
3. The logic for initializing and closing the translog snapshot has been moved to the task execution process to avoid occupying too many file descriptors.
4. Use `ExecutorCompletionService` to obtain completed tasks. When an exception occurs, cancel other `futures`.

Could you please help review it again?

PR #20263, opened 2025-12-16
Added support for groupby with String term for distinct count

vinaykpud:
Q9 ie where we group by the dc with Numeric field its already working because it uses [`NumericTermsAggregator`](  https://github.com/opensearch-project/OpenSearch/blob/c3a8415c1678f3117e1bd4148af2c225534ae8d0/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/NumericTermsAggregator.java#L526), here `subAggregators` are [`CardinalityAggregator`](https://github.com/opensearch-project/OpenSearch/blob/165df7233bcaeadc2ffab02c9440c5d9744b40f1/server/src/main/java/org/opensearch/search/aggregations/metrics/CardinalityAggregator.java#L767) since its DC.
eg:
```
query='source=hits | stats dc(UserID) as u by RegionID | sort -u | head 10'
```

---

But for queries like 11, 12, 13, 14 where we group by the dc with String field this changes required. 
eg:
```
query="source=hits | where MobilePhoneModel != '' | stats dc(UserID) as u by MobilePhoneModel | sort - u | head 10"
```

Because it uses [`AbstractStringTermsAggregator`](https://github.com/opensearch-project/OpenSearch/blob/9ce258a655417a36a51e2ae630579cd087934364/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/AbstractStringTermsAggregator.java#L128) and fails in the `SearchEngineResultConversionUtils.java` as `CardinalityAggregator` is not handled and added that in this PR

Test result:

```
    PASSING: [1, 2, 3, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 28, 31, 32, 33, 34, 35, 36, 37, 38, 39, 40, 42, 43]
    FAILING WITH 200: [4, 24, 25, 26, 27]
    FAILING: [29, 30]
```

### Description

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/search/query/SearchEngineResultConversionUtils.java
+++ b/server/src/main/java/org/opensearch/search/query/SearchEngineResultConversionUtils.java
@@ -83,9 +83,10 @@ public class SearchEngineResultConversionUtils {
                 if (aggregator instanceof ValueCountAggregator) {
                     docCount = ((InternalValueCount) subAgg).getValue();
                 }
-//                if (aggregator instanceof CardinalityAggregator) {
-//                    docCount = ((InternalCardinality) subAgg).getValue();
-//                }
+                // This is required for the cases when we group by distinct count by any string term
+                if (aggregator instanceof CardinalityAggregator) {
+                    docCount = ((InternalCardinality) subAgg).getValue();
+                }
                 subAggs.add(subAgg);
             }
         }


server/src/main/java/org/opensearch/search/query/SearchEngineResultConversionUtils.java
```diff
@@ -83,9 +83,10 @@ public static Tuple<List<InternalAggregation>, Long> extractSubAggsAndDocCount(A
                 if (aggregator instanceof ValueCountAggregator) {
                     docCount = ((InternalValueCount) subAgg).getValue();
                 }
-//                if (aggregator instanceof CardinalityAggregator) {
-//                    docCount = ((InternalCardinality) subAgg).getValue();
-//                }
+                // This is required for the cases when we group by distinct count by any string term
```
expani, line 86: We should figure out the reason for `agg_for_doc_count` not getting added. 
IMO Using approximate count as `doc_count` is incorrect.

vinaykpud: I've checked two queries where `agg_for_doc_count` is not added to the plan, yet they behave differently. In the first query, `source=hits | where MobilePhoneModel != '' | stats dc(UserID) as u by MobilePhoneModel | sort - u | head 10`, we group by a String field with a distinct count aggregation. In the second query, `source=hits | stats dc(UserID) as u by RegionID | sort -u | head 10`, we group by a Numeric field with the same distinct count aggregation. The numeric query uses `NumericTermsAggregator` and passes, while the string query uses `AbstractStringTermsAggregator` and fails. The failing query doesn't populate `docCount`, which is why I added this change. However, I'm trying to understand why we can't retrieve `docCount` this way and why `agg_for_doc_count` is expected specifically in this scenario.

expani: Looks like we are missing the check in SQL plugin causing `agg_for_doc_count` to be missed. Here's the required diff
```
diff --git a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
index a5a804b16..e7c6b082f 100644
--- a/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
+++ b/opensearch/src/main/java/org/opensearch/sql/opensearch/request/OpenSearchQueryRequest.java
@@ -831,7 +831,7 @@ public class OpenSearchQueryRequest implements OpenSearchRequest {
                     public RelNode visit(LogicalAggregate aggregate) {
                         boolean hasCount =
                                 aggregate.getAggCallList().stream()
-                                        .anyMatch(call -> call.getAggregation().getKind() == SqlKind.COUNT);
+                                        .anyMatch(call -> call.getAggregation().getKind() == SqlKind.COUNT && call.isDistinct() == false);

                         if (hasCount) {
                             return super.visit(aggregate);
```

That's why queries with distinct count agg didn't have `agg_for_doc_count` added. 

Adding this the plan's look like follows 

```
[INFO ][o.o.s.o.r.OpenSearchQueryRequest] [runTask-0] Calcite Logical Plan after Conversion
 LogicalSystemLimit(sort0=[$0], dir0=[DESC-nulls-last], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalSort(sort0=[$0], dir0=[DESC-nulls-last], fetch=[10])
    LogicalProject(u=[$1], RegionID=[$0], agg_for_doc_count=[$2])
      LogicalAggregate(group=[{0}], u=[APPROX_COUNT_DISTINCT($1)], agg_for_doc_count=[COUNT()])
        LogicalProject(RegionID=[$68], UserID=[$84])
          CalciteLogicalIndexScan(table=[[OpenSearch, hits]])
```

>However, I'm trying to understand why we can't retrieve docCount this way and why agg_for_doc_count is expected specifically in this scenario.

doc_count is expected for all scenarios. 

This is required for merging at Coordinator to work properly. See https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/InternalTerms.java#L395 

There are multiple places that rely on doc count for instance https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/InternalTerms.java#L474 relies on the same to prune buckets lower that a certain doc count.

expani:
Closing in favor of https://github.com/vinaykpud/sql/pull/29

PR #20290, opened 2025-12-19
Public getIncludes method in SourceFieldMapper

### LINKED ISSUE #20289: [Feature Request] Make includes publicly accessible in SourceFieldMapper

Creating this issue to add a public getter for the includes list in SourceFieldMapper.java. 

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/mapper/SourceFieldMapper.java

This is required for a knn derived source path:
https://github.com/opensearch-project/k-NN/issues/3006

Currently there is no efficient way to retrieve the includes from a SourceFieldMapper object.

Ask is to add a public getter function for the includes of a SourceFieldMapper class.

The only other alternative possible in the knn derived source path is to get the string source from mapperservice, and find includes through there, but this is not efficient.

Add a public includes getter in SourceFieldMapper.

### Related component

_No response_

_No response_

### Additional context

_No response_

buddharajusahil:
### Description
This change is adding a public getter function for the includes list in SourceFieldMapper.

### Related Issues
Resolves #20289

### Check List

## Summary by CodeRabbit

* **New Features**
  * Added a public accessor to read which source fields are included, complementing the existing exclusion accessor.

* **Refactor**
  * Modernized the include/exclude accessors to return a more flexible collection-style type; external behavior and results remain unchanged.


navneet1v (approved):

server/src/main/java/org/opensearch/index/mapper/SourceFieldMapper.java
```diff
@@ -282,6 +282,10 @@ public boolean isComplete() {
         return complete;
     }
 
+    public String[] getIncludes() {
+        return includes.clone();
+    }
```
andrross: I don't have the specific context on how this is going to be used, but in general using something from the collections API is preferable in modern Java compared to passing around raw arrays. Can you doing something like the following?


```suggestion
    public Collection<String> getIncludes() {
        return List.of(includes);
    }
```

Note we'd want to change the `getExcludes` API as well if we did this, but that should be okay given that it is an unreleased changed.

buddharajusahil: Sure, this makes sense, will change getExcludes as well.

PR #20306, opened 2025-12-22
Implementing docs, commits and segments stats in CompositeEngine

shank9918:
### Description
Implementing DocStats, CommitStats & SegmentStats interfaces in CompositeEngine and integrating it with IndexShard.

### Related Issues

### Check List

### DIFF
--- a/plugins/engine-datafusion/jni/src/lib.rs
+++ b/plugins/engine-datafusion/jni/src/lib.rs
@@ -8,7 +8,7 @@ use std::num::NonZeroUsize;
  * compatible open source license.
  */
 use std::ptr::addr_of_mut;
-use jni::objects::{JByteArray, JClass, JObject};
+use jni::objects::{JByteArray, JClass, JMap, JObject};
 use jni::objects::JLongArray;
 use jni::sys::{jboolean, jbyteArray, jint, jlong, jstring};
 use jni::{JNIEnv, JavaVM};
@@ -51,7 +51,7 @@ pub mod logger;
 use vectorized_exec_spi::{log_info, log_error, log_debug};
 
 use crate::custom_cache_manager::CustomCacheManager;
-use crate::util::{create_file_meta_from_filenames, parse_string_arr, set_action_listener_error, set_action_listener_error_global, set_action_listener_ok, set_action_listener_ok_global};
+use crate::util::{create_file_meta_from_filenames, parse_string_arr, set_action_listener_error, set_action_listener_error_global, set_action_listener_ok, set_action_listener_ok_global, set_action_listener_ok_global_with_map};
 use datafusion::execution::memory_pool::{GreedyMemoryPool, TrackConsumersPool};
 
 use crate::statistics_cache::CustomStatisticsCache;
@@ -482,6 +482,29 @@ impl CustomFileMeta {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct FileStats {
+    /// Total file size in bytes
+    pub size: u64,
+
+    /// Total number of rows in the file
+    pub num_rows: i64,
+}
+
+impl FileStats {
+    pub fn new(size: u64, num_rows: i64) -> Self {
+        Self { size, num_rows }
+    }
+
+    pub fn size(&self) -> u64 {
+        self.size
+    }
+
+    pub fn num_rows(&self) -> i64 {
+        self.num_rows
+    }
+}
+
 #[no_mangle]
 pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_executeQueryPhaseAsync(
     mut env: JNIEnv,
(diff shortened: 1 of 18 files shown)


plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -574,6 +597,55 @@ pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_executeQu
...
+                                    &DataFusionError::Execution(format!("Failed to create global ref: {}", e)));
+            return;
+        }
+    };
+    let io_runtime = manager.io_runtime.clone();
+
+    let shard_view = unsafe { &*(shard_view_ptr as *const ShardView) };
+    let files_meta = shard_view.files_metadata();
+
+    io_runtime.block_on(async move {
```
bharath-techie, line 632: Lets create and use a different runtime ? 

IO runtime if busy on search threads might delay stats API [ which might be okay as well - but thinking out aloud ]

shank9918: This will be only called once per DatafusionReader creation. Also, we're lazily calling this rust function when required. I think having a different runtime will be an overkill just for stats function.

server/src/main/java/org/opensearch/index/engine/exec/coord/CompositeEngine.java
```diff
@@ -844,6 +854,68 @@ public void flush(boolean force, boolean waitIfOngoing) throws EngineException {
...
+        try {
+            Map<String, String> userData = compositeEngineCommitter.getLastCommittedData();
+            long generation = Long.parseLong(userData.get(LAST_COMPOSITE_WRITER_GEN_KEY));
+            String id = userData.get(CATALOG_SNAPSHOT_ID);
+            // TODO - Implement numDocs
+            return new CommitStats(userData, generation, id, 0);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
```
Bukhtawar, line 970: Can we not maintain this in-memory in the commit path and load from disk on fresh starts?

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/search/DatafusionReader.java
```diff
@@ -56,6 +65,23 @@ public DatafusionReader(String directoryPath, CompositeEngine.ReleasableRef<Cata
         System.out.println("File names: " + Arrays.toString(fileNames));
         System.out.println("Directory path: " + directoryPath);
         this.readerHandle = new ReaderHandle(directoryPath, fileNames, this::releaseCatalogSnapshot);
+        this.segmentStatsFuture = new CompletableFuture<>();
+        setupSegmentStatsCompletableFuture(segmentStatsFuture);
```
mgodwan: Can we do this lazily? Not all readers may warrant a stats call, and we can possibly avoid the call if some reader view doesn't get a stats call.

server/src/main/java/org/opensearch/index/engine/EngineSearcher.java
```diff
@@ -13,7 +13,6 @@
 import org.opensearch.search.aggregations.SearchResultsCollector;
 
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
```
mgodwan, line 16: [nit] can be removed

raghuvanshraj:
@shank9918 remote store ITs are failing on this PR.
Can you check once?
```
./gradlew :modules:parquet-data-format:internalClusterTest
```
```
./gradlew :modules:parquet-data-format:integTest
```

server/src/main/java/org/opensearch/index/engine/exec/coord/CompositeEngine.java
shank9918: Yes, sounds good. I've made this change and LuceneCommitEngine now maintains last committed segment infos in memory and refreshes that either during fresh starts or during commit as it happens today in InternalEngine.

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/search/DatafusionReader.java
shank9918: Yes, makes sense. updated reader to lazily initialize the stats

server/src/main/java/org/opensearch/index/engine/EngineSearcher.java
shank9918: done

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
```diff
@@ -490,4 +510,17 @@ public void executeFetchPhase(DatafusionContext context) throws IOException {
...
+    @Override
+    public Map<String, FileStats> fetchSegmentStats() throws IOException {
+        DatafusionReader datafusionReader = null;
+        try {
+            datafusionReader = datafusionReaderManager.acquire();
+            return datafusionReader.fetchSegmentStats();
+        } finally {
+            if (datafusionReader != null) {
+                datafusionReaderManager.release(datafusionReader);
+            }
```
Bukhtawar: You could replace it with try-with-resource, datafusionReader implements AutoCloseable

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/search/DatafusionReader.java
```diff
@@ -88,6 +99,39 @@ public int getRefCount() {
...
+        if (currentSegmentStats == null) {
+            synchronized (this) {
+                currentSegmentStats = segmentStats;
+                if (currentSegmentStats == null) {
+                    segmentStats = currentSegmentStats = startSegmentStatsFetch();
+                }
+            }
+        }
+        return currentSegmentStats;
+    }
```
Bukhtawar: Do you really need currentSegmentStats and segmentStats

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/search/DatafusionReader.java
```diff
@@ -88,6 +99,39 @@ public int getRefCount() {
...
+                segmentStatsFuture.complete(map);
+            }
+
+            @Override
+            public void onFailure(Exception e) {
+                segmentStatsFuture.completeExceptionally(e);
+            }
+        });
+        return segmentStatsFuture.join();
+    }
```
Bukhtawar: You can use `LatchedActionListener` instead

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
shank9918: Yep done now.

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/search/DatafusionReader.java
```diff
@@ -8,18 +8,34 @@
...
+import java.util.concurrent.TimeUnit;
+
 /**
  * DataFusion reader for JNI operations.
  */
 public class DatafusionReader implements Closeable {
+
+    private static final Logger logger = LogManager.getLogger(DatafusionReader.class);
+
+    private static final TimeValue FETCH_TIMEOUT = TimeValue.timeValueMillis(15);
```
Bukhtawar: Seems pretty low, given usual garbage collection pauses are around 20-30ms, can we do 500ms?

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -574,6 +597,55 @@ pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_executeQu
...
+        Err(e) => {
+            log_error!("Failed to create global ref: {}", e);
+            set_action_listener_error(&mut env, listener,
+                                    &DataFusionError::Execution(format!("Failed to create global ref: {}", e)));
+            return;
+        }
+    };
+    let io_runtime = manager.io_runtime.clone();
+
+    let shard_view = unsafe { &*(shard_view_ptr as *const ShardView) };
```
Bukhtawar, line 629: Do we need safety checks like
```
let shard_view = unsafe {
    if shard_view_ptr == 0 {
        set_action_listener_error(&mut env, listener, 
            &DataFusionError::Execution("Null pointer".into()));
        return;
    }
    &*(shard_view_ptr as *const ShardView)
};
```

Issue #13197, opened 2024-04-15
[RFC] Refactor Metadata management in Cluster Manager code as separate lib


shwetathareja:
### Is your feature request related to a problem? Please describe

Today, Cluster Manager metadata management like create/update/delete of an index, alias, template, data stream etc. are tightly coupled with cluster state or cluster management. Hence, any mutable action is tied to enqueuing a task in CM pending_tasks queue and then ClusterStateExecutor updates the ClusterState object. This make it difficult to evolve Cluster Manager to be more cloud native provider friendly, And, also eventually run the CM in stateless mode. 

### Describe the solution you'd like

Refactor the Cluster Manager code to separate Metadata management as a lib for core entities and there are already custom data  types which are pluggable from plugins perspective. This will ensure both cluster and cloud native implementation can build on top of it.

### Related component

Cluster Manager

PR #20316, opened 2025-12-24
Created a new lib metadata-common with model classes for AliasMetadata and MappingMetadata

gargmanik13:
### Description
- Created a new library metadata-common under libs folder in the OpenSearch repo.
- This will help to separate out the metadata related dependencies which are currently coupled with cluster management dependencies in the server package. It'll make the Cluster Manager more cloud native provider friendly and help to eventually run the CM in stateless mode.
- As part of this change, added model classes for AliasMetadata and MappingMetadata which can be used independently and with existing OpenSearch dependencies to store relevant metadata.
- This will be followed by UTs and IndexMetadataModel class

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/13197

### Check List


libs/metadata-common/src/main/java/org/opensearch/cluster/metadata/AliasMetadataModel.java
```diff
@@ -0,0 +1,357 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.cluster.metadata;
```
rajiv-kv, line 9: the package name can start with `org.opensearch.metadata.common` to be consistent with library name `metadata-common`.

The model classes can be in `org.opensearch.metadata.common.index.model`

PR #20321, opened 2025-12-24
transport-netty4 module set TLS SNI when server_name is provided

### LINKED ISSUE #17316: [BUG] Cross-cluster search in proxy mode not using server_name in SNI
### Describe the bug

Hello,
during setup of cross cluster search in proxy mode I found out issue, that opensearch is not using server_name as part of SNI in TLS Client Hello. Can you please help?

### Related component

Search:Remote Search

### To Reproduce

1. Create opensearch remote cluster via API
```
PUT /_cluster/settings
{
  "persistent": {
    "cluster": {
      "remote": {
        "opensearch-test": {
          "mode": "proxy",
          "proxy_address": "opensearch.org:443",
          "server_name": "opensearch.org"
        }
      }
    }
  }
}
```

2. Start tcpdump capture to file with
```
tcpdump -i eth0 "dst port 443 and dst opensearch.org and (tcp[((tcp[12] & 0xf0) >> 2)] = 0x16)" -w /tmp/nosni.pcap
```

3. Try cross cluster search via API
```
GET opensearch-test:test_index/_search
```

4. Stop tcpdump
5. Analyse /tmp/nosni.pcap in Wireshark, there won't be any SNI information presented

### Expected behavior

There should be SNI information presented in captured packets same way as during following procedure:
1. Start tcpdump capture to file with
```
tcpdump -i eth0 "dst port 443 and dst opensearch.org and (tcp[((tcp[12] & 0xf0) >> 2)] = 0x16)" -w /tmp/sni.pcap
```
2. Test SSL connection with 
```
echo "Q" | openssl s_client -connect opensearch.org:443
```
4. Stop tcpdump
5. Analyse /tmp/sni.pcap in Wireshark, there will be SNI information presented

### Additional Details

**Plugins**
opensearch-alerting                  2.17.1.0
opensearch-anomaly-detection         2.17.1.0
opensearch-asynchronous-search       2.17.1.0
opensearch-cross-cluster-replication 2.17.1.0
opensearch-custom-codecs             2.17.1.0
opensearch-flow-framework            2.17.1.0
opensearch-geospatial                2.17.1.0
opensearch-index-management          2.17.1.0
opensearch-job-scheduler             2.17.1.0
opensearch-knn                       2.17.1.0
opensearch-ml                        2.17.1.0
opensearch-neural-search             2.17.1.0
opensearch-notifications             2.17.1.0
opensearch-notifications-core        2.17.1.0
opensearch-observability             2.17.1.0
opensearch-performance-analyzer      2.17.1.0
opensearch-reports-scheduler         2.17.1.0
opensearch-security                  2.17.1.0
opensearch-security-analytics        2.17.1.0
opensearch-skills                    2.17.1.0
opensearch-sql                       2.17.1.0
opensearch-system-templates          2.17.1.0
prometheus-exporter                  2.17.1.0
query-insights                       2.17.1.0
repository-s3                        2.17.1

**Screenshots**
Wireshark example of /tmp/nosni.pcap 

Image

Wireshark example of /tmp/sni.pcap 

Image

- opensearchproject/opensearch:2.17.1 image running inside EKS

**Additional context**
N/A

orages:
### Description
Populate the "ServerNames" SSL parameter when `cluster.remote.<cluster_alias>.server_name` is set (according to the documentation https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/cluster-settings/).

### Related Issues
Resolves #17316

### Check List

## Summary by CodeRabbit

* **New Features**
  * Client-side Server Name Indication (SNI) for secure transport: clients can now provide a configured server name when establishing SSL/TLS connections.

* **Documentation**
  * Changelog updated to note that cluster.remote.<cluster_alias>.server_name now populates SNI.

### DIFF
--- a/server/src/main/java/org/opensearch/plugins/SecureTransportSettingsProvider.java
+++ b/server/src/main/java/org/opensearch/plugins/SecureTransportSettingsProvider.java
@@ -132,7 +132,8 @@ public interface SecureTransportSettingsProvider {
      * @return if supported, builds the {@link SSLEngine} instance
      * @throws SSLException throws SSLException if the {@link SSLEngine} instance cannot be built
      */
-    default Optional<SSLEngine> buildSecureClientTransportEngine(Settings settings, String serverName, String hostname, int port) throws SSLException {
+    default Optional<SSLEngine> buildSecureClientTransportEngine(Settings settings, String serverName, String hostname, int port)
+        throws SSLException {
         return buildSecureClientTransportEngine(settings, hostname, port);
     }
 }


modules/transport-netty4/src/main/java/org/opensearch/transport/netty4/ssl/SecureNetty4Transport.java
```diff
@@ -192,6 +198,21 @@ private ClientSSLHandler(
             this.secureTransportSettingsProvider = secureTransportSettingsProvider;
             this.hostnameVerificationEnabled = hostnameVerificationEnabled;
             this.hostnameVerificationResovleHostName = hostnameVerificationResovleHostName;
+            this.serverName = null;
+        }
+
+        private ClientSSLHandler(
```
reta, line 195: Thanks for the pull request @orages , the SSL-related configuration is provided by `SecureTransportSettingsProvider` which is (among other things) responsible for building client / server `SSLEngine`s. I think this would be an appropriate place to introduce SNI names support, which also become available to all transports.

orages: Thanks for your reply.

As I am not a professional Java developer and not familiar with the OpenSearch codebase, I will need assistance/guidance to make the changes you want.

From my current understanding, the `SecureTransportSettingsProvider` is not implemented anywhere (other than in tests) in the OpenSearch repo.  
I've found an implementation in the `opensearch-security` plugin in [/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java](https://github.com/opensearch-project/security/blob/7f168b194979d21b751cd0ae4c0c1cb0508db248/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java). Which seems to be the one used in typical OpenSearch installations.  
Is that correct (and does it mean changes have to be done in sync in the security plugin)?  

If I'm correct, I guess the expected changes would be:

1. In OpenSearch: Updating the `SecureTransportSettingsProvider` interface to add a `buildSecureClientTransportEngine` method taking a `serverName` parameter (in addition to the `settings`, `hostname` and `port`).
2. In security: Updating the `buildSecureClientTransportEngine` implementation of the `OpenSearchSecureSettingsFactory` to add a version with the `serverName` parameter and pass it to the `createSSLEngine` method of the underlying `sslContextHandler`.
3. In security: Update the `SslContextHandler` to add a `createSSLEngine` method that will create and set the SNI in the SSL parameters (calling the version without serverName for engine creation and reusing the code currently added in the `ClientSSLHandler`).
4. Update the tests accordingly.

Does it make sense to you or am I missing something?

Thanks again,

reta: Thank you @orages 

> I've found an implementation in the `opensearch-security` plugin in [/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java](https://github.com/opensearch-project/security/blob/7f168b194979d21b751cd0ae4c0c1cb0508db248/src/main/java/org/opensearch/security/ssl/OpenSearchSecureSettingsFactory.java). Which seems to be the one used in typical OpenSearch installations.
> Is that correct (and does it mean changes have to be done in sync in the security plugin)?

That is correct, the `security` plugin is the one providing security related settings to the OpenSearch service

> In OpenSearch: Updating the SecureTransportSettingsProvider interface to add a buildSecureClientTransportEngine method taking a serverName parameter (in addition to the settings, hostname and port).

I would assume the `settings` parameters should already hold the `cluster.remote.<cluster_alias>.server_name ` , it should be sufficient, right?

orages: Thank you @reta.

I tried to inspect the content of `settings` (using loggers) but I don't see the content if the `cluster.remote.<...>` namespace in it.

Moreover, since multiple remote clusters can be configured at the same time, even won't we need the server name or the cluster name to know for which one we are building the SSL engine ?

reta: > Moreover, since multiple remote clusters can be configured at the same time, even won't we need the server name or the cluster name to know for which one we are building the SSL engine ?

Yeah, it is unexpected to have `settings` incomplete ( no `cluster.remote.<...>` ones), my initial plan was pretty straightforward:
 - collect cluster aliases from settings (similarly to what `RemoteConnectionStrategy.getRemoteClusters(settings)` does)
 - since we have `hostname` and `port` (they are passed to `buildSecureClientTransportEngine`), we should be able to find out which cluster alias corresponds to it (that would be `cluster.remote.<alias>.proxy_address` setting)
 - having alias, we should be able to figure out server name to use

orages: >  - since we have hostname and port (they are passed to buildSecureClientTransportEngine), we should be able to find out which cluster alias corresponds to it (that would be cluster.remote.<alias>.proxy_address setting)

It seems to be the case only when hostname verification is enabled. Otherwise, `hostname` and `port` are set to `null` and `-1` respectively.  
Besides the solution you suggest will not work in situations where multiple remote clusters are connected through the same proxy and routing is done using the SNI. Is this on purpose ?

reta: > Besides the solution you suggest will not work in situations where multiple remote clusters are connected through the same proxy and routing is done using the SNI. Is this on purpose ?

Yeah, I don't know how practical that would be but in general - it is an issue indeed. I am thinking about 3 options here:

- add synthetic `Setting` (fe SERVER_NAME) to `settings` that is populated from `DiscoveryNode` attribute `server_name` (this setting could be declared in `SecureTransportSettingsProvider` and documented)

- add `buildSecureClientTransportEngine` hook with `serverName` 

```
default Optional<SSLEngine> buildSecureClientTransportEngine(Settings settings, String serverName, String hostname, int port) throws SSLException {
    return buildSecureClientTransportEngine(settings, hostname, port);
}
```

- or, add `buildSecureClientTransportEngine` with `DiscoveryNode` 

```
Optional<SSLEngine> buildSecureClientTransportEngine(Settings settings, DiscoveryNode node) throws SSLException;
```

The latter looks preferable to me, but it does not cleanly apply to the Initialization sequence `SSLClientChannelInitializer` -> `ClientSSLHandler` -> `connect` where we only deal with `SocketAddress`. The first option has a benefit of essentially having zero changes but we don't have a good established mechanism / practice to deal with such synthetic `Setting`s, so probably the 2nd option would be the best fit, wdyt? Thanks @orages

orages: Thanks @reta

As I am not well versed in Java and the codebase I don't know if my opinion is relevant here...  
Per what you said the hook with `serverName` looks indeed like the best solution for now.

I'll update this PR to add your suggestion and move the SSLParameters stuff in a PR for the security plugin.

reta:
Thanks @orages , the change is looking good, could you please resolve the conflicts and take all checks to passing state? thank you

PR #20325, opened 2025-12-26
Fixes for remote store integration

mgodwan:
Fixes in upload flow
1. Close acquired snapshot
2. Clone should not acquire
3. Translog interaction should be on indexer, rather than engine.
4. Seq No management for remote store verification

Pending: Tests


server/src/main/java/org/opensearch/index/engine/exec/coord/LastRefreshedCheckpointListener.java
```diff
@@ -0,0 +1,48 @@
...
+package org.opensearch.index.engine.exec.coord;
+
+import org.apache.lucene.search.ReferenceManager;
+import org.opensearch.index.seqno.LocalCheckpointTracker;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+final class LastRefreshedCheckpointListener implements ReferenceManager.RefreshListener {
+    final AtomicLong refreshedCheckpoint;
+    volatile AtomicLong pendingCheckpoint;
```
Bukhtawar: AtomicLong is implicitly volatile in terms of memory visibility

server/src/main/java/org/opensearch/index/engine/exec/coord/LastRefreshedCheckpointListener.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package org.opensearch.index.engine.exec.coord;
+
+import org.apache.lucene.search.ReferenceManager;
+import org.opensearch.index.seqno.LocalCheckpointTracker;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+final class LastRefreshedCheckpointListener implements ReferenceManager.RefreshListener {
```
Bukhtawar: Maybe move this as an inner class for CompositeEngine?

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshot.java
```diff
@@ -239,10 +239,9 @@ public String toString() {
...
     }
 
-    @Override
-    public CatalogSnapshot clone() {
-        // TODO this doesn't call super.clone right now
-        return new CatalogSnapshot(getId(), getVersion(), getSegments(), catalogSnapshotMap, indexFileDeleterSupplier);
+    public CatalogSnapshot cloneNoAcquire() {
+        // Still using the clone call since Lucene call requires clone. This will allow a SegmentsInfos backed CatalogSnapshot to use the same method in calls.
+        return this;
     }
```
Bukhtawar, line 245: Technically this isn't clone right?

server/src/main/java/org/opensearch/index/shard/ReleasableRetryableRefreshListener.java
```diff
@@ -64,7 +64,13 @@ public ReleasableRetryableRefreshListener(ThreadPool threadPool) {
     @Override
     public final void afterRefresh(boolean didRefresh, CompositeEngine.ReleasableRef<CatalogSnapshot> catalogSnapshot) throws IOException {
         // TODO CompositeEngine filters CatalogSnapshotAwareListeners, keeping this for now
+
         afterRefresh(didRefresh);
+        try {
+            catalogSnapshot.close();
+        } catch (Exception ex) {
+
+        }
```
Bukhtawar: Assuming this is the older version of the snapshot that would be dec referenced?

mgodwan: This is the latest one, but the RefreshListener relies on fetching the latest catalog snapshot from engine later. I've modified this to be passed a supplier so that it can acquire and release both in the same context.

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshot.java
mgodwan: Yes.
We need a clone for SegmentsInfo but it is not required for default CatalogSnaphot. Once we have a SegmentInfos backed CatalogSnapshot for Lucene changes to work with same code, this would make more sense.
Hence, kept it like this for now.

server/src/main/java/org/opensearch/index/engine/exec/coord/LastRefreshedCheckpointListener.java
mgodwan: Kept this outside so that is can shared between InternalEngine and CompositeEngine. Made changes in InternalEngine as well.

server/src/main/java/org/opensearch/index/engine/exec/coord/LastRefreshedCheckpointListener.java
mgodwan: Ack.

Issue #19921, opened 2025-11-07
[Context Aware Segments] Remove child level directory during refresh


RS146BIJAY:
### Is your feature request related to a problem? Please describe

Remove child level directory during refresh once child level IndexWriter is synced with parent IndexWriter.

### Describe the solution you'd like

TODO

### Related component

Indexing

cwperks:
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/cwperks) [3](https://github.com/imdhemy)

PR #20326, opened 2025-12-28
Remove child level directory on refresh for CompositeIndexWriter

RS146BIJAY:
### Description
Remove child level directory during refresh once child level IndexWriter is synced with parent IndexWriter.

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/19921

## Summary by CodeRabbit

* **Bug Fixes**
  * Child index directories are now removed during refresh, preventing accumulation of orphaned on-disk directories.

* **Tests**
  * Added/updated tests to verify child directories are deleted after refresh and to validate concurrent indexing during refresh cycles.

### DIFF
--- a/server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
+++ b/server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
@@ -19,6 +19,7 @@ import org.apache.lucene.index.LiveIndexWriterConfig;
 import org.apache.lucene.index.Term;
 import org.apache.lucene.store.AlreadyClosedException;
 import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FSDirectory;
 import org.apache.lucene.util.BytesRef;
 import org.opensearch.OpenSearchException;
 import org.opensearch.common.CheckedBiFunction;
@@ -39,6 +40,7 @@ import org.opensearch.index.store.Store;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
```diff
@@ -540,12 +543,28 @@ private void refreshDocumentsForParentDirectory(CriteriaBasedIndexWriterLookup o
...
     }
 
+    private FSDirectory getLocalFSDirectory(Directory localDirectory) {
+        FSDirectory localFSDirectory;
+        if (localDirectory instanceof FSDirectory) {
+            localFSDirectory = (FSDirectory) localDirectory;
+        } else {
+            // In this case it should be a FilterDirectory wrapped over FSDirectory as per above validation.
+            localFSDirectory = (FSDirectory) (((FilterDirectory) localDirectory).getDelegate());
+        }
```
mgodwan: We should assert that it is always an FSDirectory since we use createFSDirectory method from store

server/src/test/java/org/opensearch/index/engine/CompositeIndexWriterForAppendTests.java
```diff
@@ -187,8 +189,49 @@ public void testUnableToObtainLockOnActiveLookupWhenWriteLockDuringIndexing() th
...
+                compositeIndexWriter.addDocuments(operation.docs(), operation.uid());
+            }
+
+            compositeIndexWriter.beforeRefresh();
+            compositeIndexWriter.afterRefresh(true);
+
+            long directoryCount = Files.find(
+                dataPath,
+                1,
+                (path, attributes) -> attributes.isDirectory() == true && path.endsWith("extra0") == false
```
mgodwan, line 199: [nit] Add a comment explaining why extra0 needs to be removed [ExtrasFS]

RS146BIJAY: Addressed the comments.

server/src/main/java/org/opensearch/index/engine/CompositeIndexWriter.java
RS146BIJAY: Addressed the comments

Issue #20349, opened 2026-01-01
[BUG] Create Index Fails for simple_pattern


mgodwan:
### Describe the bug

For `simple_pattern` tokenizer, if the automaton generated during initialization is an a non-deterministic automaton, the index creation fails. This has regressed recently as the same regex pattern earlier used to work in older OpenSearch versions.

### Related component

Other

### To Reproduce

* Create an index
```
{
  "settings": {
    "index": {
      "analysis": {
        "tokenizer": {
          "some_tokenizer": {
            "type": "simple_pattern",
            "pattern": "[A-Za-z0-9]+|[^\n - ]"
          }
        },
        "analyzer": {
          "some_analyzer": {
            "type": "custom",
            "tokenizer": "some_tokenizer",
            "filter": ["lowercase"]
          }
        }
      },
      "number_of_shards": "1",
      "number_of_replicas": "0"
    }
  },
  "mappings": {
    "dynamic": "false",
    "properties": {
      "field1": {
        "type": "text",
        "analyzer": "some_analyzer"
      }
    }
  }
}
```

* See the below error
```
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"please determinize the incoming automaton first"}],"type":"illegal_argument_exception","reason":"please determinize the incoming automaton first"},"status":400}
```

### Expected behavior

Index creation should go through. 

### Additional Details

**Plugins**
analysis-commons (module)

**Screenshots**
NA

**Additional context**
This is working till OpenSearch 2.x and has broken in OpenSearch 3.x. This is related to github.com/apache/lucene/pull/485 which does not determinize the Automaton generated by default anymore.

PR #20350, opened 2026-01-01
Determinize Automaton for simple_pattern and simple_pattern_split tokenizer

mgodwan:
Determinize Automaton for simple_pattern and simple_pattern_split tokenizer.

### Description
Instead of passing the regex pattern directly to Lucene tokenizer implementation, OpenSearch now passes a determinized automaton to Lucene.

### Related Issues
Fixes https://github.com/opensearch-project/OpenSearch/issues/20349

### Check List

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Tokenizers now handle complex regular-expression patterns more reliably by using a deterministic processing approach, improving accuracy and stability of token extraction.

* **Tests**
  * Added tests that validate tokenizer behavior with complex regex patterns to ensure consistent tokenization outcomes.

### DIFF
--- a/modules/analysis-common/src/main/java/org/opensearch/analysis/common/SimplePatternSplitTokenizerFactory.java
+++ b/modules/analysis-common/src/main/java/org/opensearch/analysis/common/SimplePatternSplitTokenizerFactory.java
@@ -34,6 +34,9 @@ package org.opensearch.analysis.common;
 
 import org.apache.lucene.analysis.Tokenizer;
 import org.apache.lucene.analysis.pattern.SimplePatternSplitTokenizer;
+import org.apache.lucene.util.automaton.Automaton;
+import org.apache.lucene.util.automaton.Operations;
+import org.apache.lucene.util.automaton.RegExp;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.env.Environment;
 import org.opensearch.index.IndexSettings;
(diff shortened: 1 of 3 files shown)


modules/analysis-common/src/main/java/org/opensearch/analysis/common/SimplePatternSplitTokenizerFactory.java
```diff
@@ -51,6 +53,8 @@ public SimplePatternSplitTokenizerFactory(IndexSettings indexSettings, Environme
 
     @Override
     public Tokenizer create() {
-        return new SimplePatternSplitTokenizer(pattern);
+        return new SimplePatternSplitTokenizer(
+            Operations.determinize(new RegExp(pattern).toAutomaton(), Operations.DEFAULT_DETERMINIZE_WORK_LIMIT)
```
Bukhtawar: Can we have ctor create `new RegExp(pattern).toAutomaton()` upfront, unless `create` is being invoked once per instance?

mgodwan: `create` is invoked for every document. 

I could see from the code that Tokenizer does not call any mutating methods on Automaton object, but Automaton object itself has some methods which can mutate its state. Hence, I was initializing it every time a tokenizer is created for a field to not rely on the implementation detail there. 
It seems safe to initialize and determinize automaton once, and the reuse it based on how Lucene uses it today in the indexing analyzer path. Let me know your thoughts as well. Will make the change it this sounds ok.

Bukhtawar: Thanks for the change, looks good

PR #20359, opened 2026-01-04
Stream transport refactor

rishabhmaurya:
* Make `FlightClientChannel` `sendMessage` async
* *  make use of virtual threads
* fix for headers when no batches are sent; This can fail today, thus always propagating headers.
* set version from headers for incoming response deserialization
* Make serialization to VectorStreamOutput more efficient by writing to heap based buffer of 8192 bytes before flushing it to Arrow buffers. 

Minor
* Remove unnecessary use of `Optional` in `FlightServerChannel`.
* Customized thread factory for NIOELG.

### Description

### Related Issues

### Check List

## Summary by CodeRabbit

* **New Features**
  * Added Flight thread pool minimum size configuration option.
  * Added Flight publish port configuration option.

* **Bug Fixes**
  * Enhanced Arrow Flight stream transport stability.

* **Chores**
  * Refined event loop thread allocation and management.
  * Implemented thread naming for improved observability.
  * Updated test suite for stream transport validation.

### DIFF
--- a/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
+++ b/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
@@ -272,8 +272,7 @@ class FlightClientChannel implements TcpChannel {
             Runnable task = () -> {
                 try (var ignored = threadContext.stashContext()) {
                     if (header == null) {
-                        cleanupStreamResponse(streamResponse);
-                        throw new StreamException(StreamErrorCode.INTERNAL, "Header is null");
+                        handleStreamException(streamResponse, new StreamException(StreamErrorCode.INTERNAL, "Header is null"));
                     }
                     threadContext.setHeaders(header.getHeaders());
                     handler.handleStreamResponse(streamResponse);
--- a/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
+++ b/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
@@ -55,7 +55,7 @@ class FlightServerChannel implements TcpChannel {
     private volatile boolean cancelled = false;
     private final ExecutorService executor;
     private final long correlationId;
-    private AtomicInteger batchNumber = new AtomicInteger(0);
+    private final AtomicInteger batchNumber = new AtomicInteger(0);
 
     public FlightServerChannel(
         ServerStreamListener serverStreamListener,
@@ -128,21 +128,14 @@ class FlightServerChannel implements TcpChannel {
             long rootSize = FlightUtils.calculateVectorSchemaRootSize(root);
             callTracker.recordBatchSent(rootSize, System.nanoTime() - batchStartTime);
             logger.debug(
-                "Batch #{} sent for correlation ID: {} in {}ms, size: {} bytes, putNext: {}ms",
+                "Batch #{} sent for correlation ID: {}, size: {} bytes, putNext: {}ms",
                 batchNumber,
                 correlationId,
-                putNextTime,
                 rootSize,
                 putNextTime
             );
         } else {
-            logger.debug(
-                "Batch #{} sent for correlation ID: {} in {}ms, bytes, putNext: {}ms",
-                batchNumber,
-                correlationId,
-                putNextTime,
-                putNextTime
-            );
+            logger.debug("Batch #{} sent for correlation ID: {}, putNext: {}ms", batchNumber, correlationId, putNextTime);
         }
     }
 


plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/VectorStreamOutput.java
```diff
@@ -18,45 +18,60 @@
...
-            vector = (VarBinaryVector) root.get().getVector(0);
+    public VectorStreamOutput(BufferAllocator allocator, VectorSchemaRoot root) {
+        if (root != null) {
+            vector = (VarBinaryVector) root.getVector(0);
             this.root = root;
         } else {
             Field field = new Field("0", new FieldType(true, new ArrowType.Binary(), null, null), null);
             vector = (VarBinaryVector) field.createVector(allocator);
+            // Pre-allocate with reasonable capacity to avoid repeated allocations
+            vector.setInitialCapacity(16);
```
harshavamsi, line 38: Is this 16 bytes? why not 8192?

rishabhmaurya: we can experiment with 8192 if its helping

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
```diff
@@ -69,183 +63,112 @@ public FlightTransportResponse(
...
+            Thread.ofVirtual().start(() -> {
+                try {
+                    long start = System.nanoTime();
+                    flightStream = flightClient.getStream(ticket, new HeaderCallOption(callHeaders));
+                    if ((System.nanoTime() - start) / 1_000_000 > 10) {
+                        logger.debug(
+                            "FlightClient.getStream() for correlationId: {} took {}ms",
+                            correlationId,
+                            (System.nanoTime() - start) / 1_000_000
+                        );
```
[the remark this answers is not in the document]
bowenlan-amzn: Prob need to remove the if condition in the end. Fine to keep a debug or trace log of the timing here.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
```diff
@@ -69,183 +63,112 @@ public FlightTransportResponse(
...
-     */
-    public Header getHeader() {
-        ensureOpen();
-        initializeStreamIfNeeded();
-        return currentHeader;
+    void openAndPrefetchAsync(CompletableFuture<Header> future) {
+        if (prefetchStarted) return;
+
+        synchronized (this) {
+            if (prefetchStarted) return;
```
bowenlan-amzn, line 81: Just curious is this prefetch thing a must-do, or just an optimization

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransport.java
```diff
@@ -409,7 +410,9 @@ protected InboundHandler createInboundHandler(
     }
 
     private EventLoopGroup createEventLoopGroup(String name, int threads) {
-        return new NioEventLoopGroup(threads);
+        AtomicInteger threadCounter = new AtomicInteger(0);
+        ThreadFactory threadFactory = r -> new Thread(r, name + "-" + threadCounter.incrementAndGet());
+        return new NioEventLoopGroup(threads, threadFactory);
```
bowenlan-amzn: Notice NioEventLoopGroup is deprecated, should we update to MultiThreadIoEventLoopGroup as doc suggest

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
```diff
@@ -112,6 +112,11 @@ public FlightClientChannel(
         this.closeListeners = new CopyOnWriteArrayList<>();
         this.stats = new ChannelStats();
         this.isClosed = false;
+        // Initialize with timestamp + global counter to ensure uniqueness with multiple channels
```
bowenlan-amzn, line 115: Should we pass in different initialValue per clientchannel?

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
```diff
@@ -69,183 +63,112 @@ public FlightTransportResponse(
...
-            if (flightStream.next()) {
-                currentRoot = flightStream.getRoot();
-                currentHeader = headerContext.getHeader(correlationId);
-                // Capture the batch size before deserialization
-                currentBatchSize = FlightUtils.calculateVectorSchemaRootSize(currentRoot);
-                return deserializeResponse();
-            } else {
-                streamExhausted = true;
-                return null;
+            boolean hasNext = firstBatchConsumed ? flightStream.next() : (firstBatchConsumed = true);
```
bowenlan-amzn, line 121: Why is this not using virtual thread?

rishabhmaurya: once first response is available, its synchronous for now. Maybe this can be changed in future, but there are other good ways to make it fully async by modifying the FlightStream API itself.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
rishabhmaurya: this is to make things async and calling response handler when response is available instead of making thread wait

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/VectorStreamOutput.java
rishabhmaurya: I will probably benchmark it but 8192 makes more sense. Keeping the comment open

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
rishabhmaurya: sure, what do you have in your mind?
Also, currently its a single channel profile, but wanted to fix this in case we make it multi-channel, so that it works OOB.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransport.java
rishabhmaurya: good point, updated in latest version.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/VectorStreamOutput.java
rishabhmaurya: this is actually number of entries and not bytes, so that's probably okay.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
```diff
@@ -106,35 +108,60 @@ public void sendBatch(ByteBuffer header, VectorStreamOutput output) {
...
-            long rootSize = FlightUtils.calculateVectorSchemaRootSize(root.get());
+            long rootSize = FlightUtils.calculateVectorSchemaRootSize(root);
             callTracker.recordBatchSent(rootSize, System.nanoTime() - batchStartTime);
+            logger.debug(
+                "Batch #{} sent for correlation ID: {} in {}ms, size: {} bytes, putNext: {}ms",
+                batchNumber,
+                correlationId,
+                putNextTime / 1_000_000,
+                rootSize,
+                putNextTime
```
bowenlan-amzn: `putNextTime / 1_000_000` here seems double dividing, it's already divided above? 
And do we need 2 debug logger here?

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
```diff
@@ -106,35 +108,60 @@ public void sendBatch(ByteBuffer header, VectorStreamOutput output) {
...
+            if (root == null) {
+                // Set header if no batches were sent
+                middleware.setHeader(header);
+                logger.debug("Completing empty stream for correlation ID: {}", correlationId);
+            } else {
+                logger.debug("Completing stream for correlation ID: {} after {} batches", correlationId, batchNumber);
+            }
             serverStreamListener.completed();
         } finally {
             callTracker.recordCallEnd(StreamErrorCode.OK.name());
```
bowenlan-amzn, line 160: Can callTracker be null? See you check if it's null above.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
```diff
@@ -244,39 +250,45 @@ public void sendMessage(BytesReference reference, ActionListener<Void> listener)
...
-        } catch (Exception e) {
-            cleanupStreamResponse(streamResponse);
-            throw e;
-        }
+
+            Runnable task = () -> {
+                try (var ignored = threadContext.stashContext()) {
+                    if (header == null) {
+                        cleanupStreamResponse(streamResponse);
+                        throw new StreamException(StreamErrorCode.INTERNAL, "Header is null");
```
bowenlan-amzn: Wondering should handleStreamException be used instead of just throw an exception

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
```diff
@@ -244,39 +250,45 @@ public void sendMessage(BytesReference reference, ActionListener<Void> listener)
...
+                    }
+                    threadContext.setHeaders(header.getHeaders());
+                    handler.handleStreamResponse(streamResponse);
+                } catch (Exception e) {
+                    cleanupStreamResponse(streamResponse);
+                    throw e;
```
bowenlan-amzn, line 281: Same as above

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
```diff
@@ -244,39 +250,45 @@ public void sendMessage(BytesReference reference, ActionListener<Void> listener)
...
-        final String executor = streamResponse.getHandler().executor();
+    private void openStreamAndInvokeHandler(FlightTransportResponse<?> streamResponse) {
+        TransportResponseHandler handler = streamResponse.getHandler();
+        String executor = handler.executor();
+
         if (ThreadPool.Names.SAME.equals(executor)) {
-            executeHandler(threadContext, streamResponse);
-        } else {
-            threadPool.executor(executor).execute(() -> executeHandler(threadContext, streamResponse));
+            logger.warn("Stream transport handler using SAME executor, which may cause blocking behavior");
```
bowenlan-amzn: Worried this log out for every request.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
bowenlan-amzn: I guess the prefetch in the method name is a bit misleading. Maybe just openStream is fine.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
rishabhmaurya: its Nullable, so handled that way throughout the codebase

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
rishabhmaurya: if headers are null, it should be the transport to deal with it and not the handler of API as handler won't know what to do in this case. Its an unexpected transport level error and shouldn't happen.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
rishabhmaurya: again, i don't think handler can do much here. This will happen either when header is null or there was an error handling the streamResponse, which handler would have known anyways.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
rishabhmaurya: good catch!

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
bowenlan-amzn: It's same situation for the above handleStreamException, the error there happens during flightStream initialization and first `next`

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
```diff
@@ -49,10 +50,12 @@ class FlightServerChannel implements TcpChannel {
     private final InetSocketAddress remoteAddress;
     private final List<ActionListener<Void>> closeListeners = Collections.synchronizedList(new ArrayList<>());
     private final ServerHeaderMiddleware middleware;
-    private volatile Optional<VectorSchemaRoot> root = Optional.empty();
+    private volatile VectorSchemaRoot root = null;
     private final FlightCallTracker callTracker;
     private volatile boolean cancelled = false;
     private final ExecutorService executor;
+    private final long correlationId;
+    private AtomicInteger batchNumber = new AtomicInteger(0);
```
bowenlan-amzn: final

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
```diff
@@ -106,35 +108,60 @@ public void sendBatch(ByteBuffer header, VectorStreamOutput output) {
...
-            long rootSize = FlightUtils.calculateVectorSchemaRootSize(root.get());
+            long rootSize = FlightUtils.calculateVectorSchemaRootSize(root);
             callTracker.recordBatchSent(rootSize, System.nanoTime() - batchStartTime);
+            logger.debug(
+                "Batch #{} sent for correlation ID: {} in {}ms, size: {} bytes, putNext: {}ms",
+                batchNumber,
+                correlationId,
+                putNextTime,
+                rootSize,
+                putNextTime
```
bowenlan-amzn, line 135: showing putNextTime twice

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
bowenlan-amzn: Better to just remove this then if it's for multi-channel but we don't have multi-channel for now.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
rishabhmaurya: we check/get the headers only once while opening stream and getting first response.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
rishabhmaurya: as discussed offline, this could be an improvement in future, so keeping it untouched for 3.5

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightClientChannel.java
rishabhmaurya: as discussed offline, have changed in the latest version as this provides a safeguard against APIs whose response handlers aren't protected by timeouts.
There are other concerns here though - like missing headers may result in thread context not be set appropriately and calling response handlers could be risky.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
rishabhmaurya: it prefetches the first batch though and not just open stream

jainankitk (approved):

PR #20361, opened 2026-01-05
Add tracing support for StreamingRestChannel

Hailong-am:
### Description

Add tracing support for StreamingRestChannel, streaming operation start with `prepareResponse` and end with `sendChunk(lastChunk)`, the combination will be in same span

### Related Issues

https://github.com/opensearch-project/OpenSearch-Dashboards/issues/11090

### Check List

## Bug Fixes
* Fixed thread context restoration for streaming HTTP requests in Reactor Netty 4 transport, ensuring proper context handling and cleanup during streaming operations.

## Tests
* Added integration tests to validate thread context restoration behavior for streaming responses.


server/src/main/java/org/opensearch/rest/RestController.java
```diff
@@ -345,6 +346,10 @@ private void dispatchRequest(RestRequest request, RestChannel channel, RestHandl
             }
 
             if (handler.supportsStreaming()) {
+                // for the case TraceableRestChannel wrapped steamingRestChannel, we need to unwrap it
+                if (channel instanceof TraceableRestChannel traceableRestChannel) {
```
reta: Good catch with `TraceableRestChannel`, to address this issue, we would need:
 - introduce `TraceableStreamingRestChannel` that implements `StreamingRestChannel`
 - change `TraceableRestChannel ` to return either `TraceableRestChannel` or `TraceableStreamingRestChannel`

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpServerTransport.java
```diff
@@ -381,6 +382,8 @@ protected Publisher<Void> incomingRequest(HttpServerRequest request, HttpServerR
             request.params()
         );
         if (dispatchHandlerOpt.map(RestHandler::supportsStreaming).orElse(false)) {
+            final ThreadContext.StoredContext storedContext = threadPool.getThreadContext().newStoredContext(false);
```
reta: Let's find out the cause first before doing that, the polluted thread context may pop up in other places and we should understand first why it happens.

Hailong-am: yeah the test cass could reproduce it and you could help to check whether other place  could potentially have same issue

plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4ThreadContextRestorationIT.java
```diff
@@ -0,0 +1,241 @@
...
+        }
+
+        @Override
+        protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) throws IOException {
+            return channel -> {
+                if (channel instanceof StreamingRestChannel) {
+                    StreamingRestChannel streamingChannel = (StreamingRestChannel) channel;
+
+                    // Mimic ML streaming behavior - store thread context
+                    final ThreadContext.StoredContext storedContext = client.threadPool().getThreadContext().newStoredContext(true);
```
reta: @Hailong-am OK, preliminary, I think the issue here is coming from improper `ThreadContext` manipulation within streaming handler implementation (in this case, `ml-common`): the `client.threadPool().getThreadContext().newStoredContext(true)` , `context.restore();` and `storedContext.close()` are very likely executed in different threads, hence not restored properly. I need a bit more time to confirm that and suggest a fix if that is the issue

reta: It seems the proper way to manage the context in this case would be;

```
final Supplier<ThreadContext.StoredContext> restorable = client
    .threadPool()
    .getThreadContext()
    .newRestorableContext(true);
```

plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4ThreadContextRestorationIT.java
```diff
@@ -0,0 +1,241 @@
...
+                        "Cache-Control",
+                        List.of("no-cache"),
+                        "Connection",
+                        List.of("keep-alive")
+                    );
+                    streamingChannel.prepareResponse(RestStatus.OK, headers);
+
+                    // Use Flux pattern like ML streaming action
+                    Flux.from(streamingChannel).ofType(HttpChunk.class).collectList().flatMap(chunks -> {
+                        try (ThreadContext.StoredContext context = storedContext) {
```
reta: And than just use it like this:

```
try (ThreadContext.StoredContext ctx = restorable.get()) {
   .... // no need to call `context.restore();` 
}
```

plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4ThreadContextRestorationIT.java
Hailong-am: Nice finding. It turns out the way we used to managing thread context is not the right way and make the thread context not well restored.

Hailong-am: @reta with the correct way to manage context i don't have duplicate header issue anymore. Do you think this PR is still valid and worked as a fallback to clean up thread context?

reta: @Hailong-am we definitely should not do any fallbacks to clean the context - it should be fixed, not workarounded. But tracing related finding (https://github.com/opensearch-project/OpenSearch/pull/20361/changes#r2672199308) would certainly worth  addressing. Thank you.

Hailong-am: @reta I have addressed the tracing related comments and remove the thread context handling, would you mind to help to take a look?

server/src/main/java/org/opensearch/http/AbstractHttpServerTransport.java
```diff
@@ -366,7 +368,11 @@ void dispatchRequest(final RestRequest restRequest, final RestChannel channel, f
             final Span span = tracer.startSpan(SpanBuilder.from(restRequest));
             try (final SpanScope spanScope = tracer.withSpanInScope(span)) {
                 if (channel != null) {
-                    traceableRestChannel = TraceableRestChannel.create(channel, span, tracer);
+                    if (channel instanceof StreamingRestChannel streamingRestChannel) {
```
reta: Let's hide this class check inside `TraceableRestChannel.create` please

server/src/main/java/org/opensearch/telemetry/tracing/channels/TraceableStreamingRestChannel.java
```diff
@@ -0,0 +1,166 @@
...
+ * </ul>
+ *
+ * <p><b>Thread Safety:</b> This class is designed to handle concurrent chunk sending from
+ * multiple threads (common in reactive streaming scenarios). Each operation establishes
+ * its own span scope, and span completion is guarded by an atomic boolean to ensure
+ * the span is ended exactly once.
+ *
+ * @opensearch.internal
+ */
+public class TraceableStreamingRestChannel implements StreamingRestChannel {
```
reta: I think we should be using `TraceableRestChannel` here, something along this lines (TraceableRestChannel  may need to open a few things to be reusable, or we could extract `TraceableRestChannel<C extends RestChannel>` base class):

```
class TraceableStreamingRestChannel extends TraceableRestChannel implements StreamingRestChannel {
}
``` 

You will also get the right implementation of `sendResponse` out of the box

server/src/main/java/org/opensearch/telemetry/tracing/channels/TraceableStreamingRestChannel.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public void sendChunk(HttpChunk chunk) {
+        // Each chunk operation executes within the span scope for proper trace context propagation
+        try (SpanScope ignored = tracer.withSpanInScope(span)) {
+            delegate.sendChunk(chunk);
+        }
+
+        // End the span when the last chunk is sent (thread-safe with atomic boolean)
+        if (chunk.isLast() && spanEnded.compareAndSet(false, true)) {
```
reta: Wondering why do we have to use AtomicBoolean  here? There is only one last chunk to be sent, we should be able to end the span once

reta:
@Hailong-am thanks a lot for working on it, I think we are mostly done!

server/src/main/java/org/opensearch/telemetry/tracing/channels/TraceableStreamingRestChannel.java
Hailong-am: Just in case there could have multiple chunk been marked as last and that will cause we end the span for multiple times

reta: It should never happen, it would cause much more serious issues

Hailong-am: I'm OK to remove this kind of check but i have tested with multiple chunk marked as last and didn't notice any issue. could you clarify what kind issue would have?

reta: I mean the last chuck triggers response completion on the HTTP transport side, having multiple last chunks to be received would cause issues trying to send the data over the connection that has been closed already.

Hailong-am: That make sense. I have already remove the check, can you help me review it again?

reta:
@Hailong-am sorry, conflicts popped up, could you please resolve them? thank you

Issue #15284, opened 2024-08-16
[Feature Request] A field to store histogram data in Opensearch


deshsidd:
### Is your feature request related to a problem? Please describe

 Did some research and does not look like we have a clear way of storing histogram data in OpenSearch.

### Describe the solution you'd like

Need a field to store histogram data in the following format in OpenSearch:
"values" : [0.1, 0.2, 0.3, 0.4, 0.5], 
 "counts" : [5, 8, 3, 1, 2] 
 

### Related component

Indexing

### Describe alternatives you've considered

Can try to store it as an object with an array of bucket values and counts but wondering if there is a better way we can store this to make visualizations and queries easier.

dhwanilpatel:
Thanks for filing the request @deshsidd , 

Ask is to create the mapping type to store the histogram data only, or we would want to optimize the queries on it as well? For that we might need to think of how we store this histogram in which underlying data structures based on different type of histogram.

deshsidd:
@dhwanilpatel We want to optimize on the queries on it as well. We should efficiently be able to compute percentile and other statistics.

deshsidd:
@dhwanilpatel Let me try to take this up. Might need to collaborate with you on this.

msfroh:
Technically, we probably don't need to worry as much about querying these fields. Computing aggregations over histograms though would allow us to combine all histograms within a given time period (i.e. treat it as a metric aggregation under a date histogram aggregation) to get percentile metrics for each time, so one could graph p99 over time (for example). I think we could reasonably get arbitrary percentiles, overstat/understat (i.e. what proportion of values are over/under a given target).

That means histograms probably need to be stored as binary doc values that efficiently encode the values and counts. I don't think we need to initially support any other Lucene structures (terms/postings, point trees).

deshsidd:
@dhwanilpatel Will you be prioritizing this task?

PR #20373, opened 2026-01-05
Draft: Add support for histogram field and aggregations

deshsidd:
Draft PR to support for Histogram field and aggregations.

### Description

### Related Issues
Addresses: #15284

### Check List

## Release Notes

* **New Features**
  * Added a new Histogram field type to support indexing and querying histogram-structured data with values and counts
  * Extended aggregations and queries to work with histogram fields, including min, max, sum, average, stats, percentiles, range, histogram aggregations, and exists queries

### DIFF
--- a/server/src/main/java/org/opensearch/index/fielddata/HistogramLeafFieldData.java
+++ b/server/src/main/java/org/opensearch/index/fielddata/HistogramLeafFieldData.java
@@ -1,0 +1,76 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.fielddata;
+
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.DocValues;
+import org.apache.lucene.index.LeafReader;
+import org.apache.lucene.util.BytesRef;
+import org.opensearch.indices.fielddata.Histogram;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+
+public class HistogramLeafFieldData implements LeafFieldData {
+    private final LeafReader reader;
+    private final String fieldName;
+
+    public HistogramLeafFieldData(LeafReader reader, String fieldName) {
+        this.reader = reader;
+        this.fieldName = fieldName;
+    }
+
+    public HistogramValues getHistogramValues() throws IOException {
+        final BinaryDocValues values = DocValues.getBinary(reader, fieldName);
+        return new HistogramValues() {
+            @Override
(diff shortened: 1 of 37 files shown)


server/src/main/java/org/opensearch/index/fielddata/HistogramLeafFieldData.java
```diff
@@ -0,0 +1,76 @@
...
+            public Histogram histogram() throws IOException {
+                BytesRef bytesRef = values.binaryValue();
+                return decodeHistogram(bytesRef);
+            }
+        };
+    }
+
+    private Histogram decodeHistogram(BytesRef bytesRef) {
+        ByteBuffer buffer = ByteBuffer.wrap(bytesRef.bytes, bytesRef.offset, bytesRef.length);
+        int size = buffer.getInt();
```
atris, line 47: This allocates three objects (two arrays + the Histogram wrapper) for every single document we visit. In an aggregation running over millions of docs, this will cause massive heap pressure. We really need a flyweight accessor here that reuses the underlying arrays or buffers instead of allocating fresh ones per doc.

server/src/main/java/org/opensearch/index/fielddata/HistogramValuesSource.java
```diff
@@ -0,0 +1,134 @@
...
+                    currentValues = histogram.getValues();
+                    currentValueIndex = 0;
+                    return currentValues.length > 0;
+                }
+                currentValues = null;
+                return false;
+            }
+
+            @Override
+            public double nextValue() throws IOException {
```
atris, line 68: This implementation is dropping data. doubleValues() returns the bucket keys, but we are completely ignoring the associated counts here. Aggregators consuming this (like NumericHistogramAggregator) will count each entry as 1, meaning a histogram bucket with count: 50 is weighted exactly the same as one with count: 1. This makes the aggregation result incorrect. We need NumericHistogramAggregator to be aware of the counts.

server/src/test/java/org/opensearch/index/mapper/HistogramFieldMapperTests.java
```diff
@@ -0,0 +1,413 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.mapper;
```
atris, line 9: We seem to be missing aggregation tests entirely. These tests cover parsing and serialization, but nothing verifies that terms or histogram aggregations actually yield the correct results (verifying the counts are summed up). Given the issue in HistogramValuesSource, we strictly need an AggregatorTestCase to catch this.

atris (changes requested):

PR #20400, opened 2026-01-09
Add bcFips dependencies as server dependencies for single distribution that is FIPS capable

cwperks:
### Description

This PR adds the following dependencies as api-level dependencies on `server/`

- bc-fips-2.1.2.jar
- bcpg-fips-2.1.11.jar
- bcpkix-fips-2.1.9.jar
- bctls-fips-2.1.20.jar
- bcutil-fips-2.1.4.jar

This is done so that these dependencies are always available on the runtime classpath under `lib/`. They are currently only available under lib when OpenSearch is built with the `-Pcrypto.standard=FIPS-140-3` flag.

The problem with conditionally including these `deps/` under lib is that it complicates adapting the plugins for FIPS mode and would be simpler to always include these deps.

Note: The security plugin will have adapt to this change and a couple other plugins in the default distribution may also need to adapt to the change. 

### Related Issues

See discussion on https://github.com/opensearch-project/technical-steering/issues/77

### Check List

## Summary by CodeRabbit

* **New Features**
  * FIPS mode can be enabled via an environment flag; startup scripts and compose files updated so the FIPS installer runs in containers. Java REST tests can run in FIPS-approved-only mode.

* **Chores**
  * Centralized BouncyCastle handling and license mapping across the build.
  * Removed some FIPS-specific dependency declarations where consolidated.
  * Tightened third‑party audits so previously-ignored crypto classes are now audited.
  * Updated changelog entry for Unreleased 3.x.

### DIFF
--- a/qa/fips-compliance/build.gradle
+++ b/qa/fips-compliance/build.gradle
@@ -49,3 +49,7 @@ afterEvaluate {
 }
 
 test.enabled = false
+
+javaRestTest {
+  systemProperty 'org.bouncycastle.fips.approved_only', 'true'
+}


server/build.gradle
```diff
@@ -78,6 +78,7 @@ dependencies {
   testRuntimeOnly project(':libs:opensearch-plugin-classloader')
 
   api libs.bundles.lucene
+  api libs.bundles.bouncycastle
```
reta: I believe we definitely should not include those: `server` is used by mostly (if not all) any single plugin out there, no matter if it has security related features or dependencies. Making it deployment only (as it is now) only complicates a few (like `security`, and probably that's about it). It introduces compile time dependency to BC where there is no such need at all (at least for now).

```
    <dependency>
      <groupId>org.bouncycastle</groupId>
      <artifactId>bc-fips</artifactId>
      <version>2.1.2</version>
      <scope>compile</scope>
      <exclusions>
        <exclusion>
          <groupId>*</groupId>
          <artifactId>*</artifactId>
        </exclusion>
      </exclusions>
    </dependency>

```

cwperks: Won't that necessarily mean 2 separate distributions? One of the problems is that the security plugin cannot be installed in a distro built with `-Pcrypto.standard=FIPS-140-3`. 

If we plan for a single distro, then what would the script from https://github.com/opensearch-project/technical-steering/issues/77#issuecomment-3603671829 do? Would that move jars around?

beanuwave:
If BC libraries are always present on the classpath, this logic https://github.com/opensearch-project/OpenSearch/blob/375ad8715054ceeee05c1fec73bdb9a95b2795a9/distribution/docker/src/docker/bin/docker-entrypoint.sh#L76-L78 should rely on a different condition indicator, and redundant BC deps inside plugins and modules could be cleaned up.

cwperks:
> If BC libraries are always present on the classpath, this logic
> 
> https://github.com/opensearch-project/OpenSearch/blob/375ad8715054ceeee05c1fec73bdb9a95b2795a9/distribution/docker/src/docker/bin/docker-entrypoint.sh#L76-L78
> 
> should rely on a different condition indicator, and redundant BC deps inside plugins and modules could be cleaned up.

Is there an env var we can rely on there?

server/build.gradle
reta: I believe we could have one distribution (if that is the goal) but without introducing hard dependencies

beanuwave:
> Is there an env var we can rely on there?

No, there isn’t one currently. However, an environment variable could be introduced, similar to `FIPS_GENERATE_TRUSTSTORE`

server/build.gradle
cwperks: That is the goal. A single distribution that is FIPS capable. Simplest solution would be to always include the BCFIPS jars. 

We do have a [FIPS installer script](https://docs.opensearch.org/latest/security/configuration/fips/) to help prepare a cluster for running in FIPS approved mode.

reta: > That is the goal. A single distribution that is FIPS capable. Simplest solution would be to always include the BCFIPS jars.

So let do that while building distribution, we don't need to introduce hard dependencies on component(s), those do have much larger impact on downstreams

cwperks: I don't see how we can avoid not having a separate distribution that we need to publish as a project then.

Currently we have 2 distros:

- min distro
- default distro (includes default package of plugins)

We should have FIPS variant of both:

- min distro FIPS (built with -Pcrypto.standard=FIPS-140-3)
- default distro FIPS (where all plugins are also built with -Pcrypto.standard=FIPS-140-3)

reta: > I don't see how we can avoid not having a separate distribution that we need to publish as a project then.

Apologies if I am missing something, but I think we are mixing two separate issues here:
 1. Dependencies: to unify distribution(s), we could always include FIPS related jars when building min or default distro by just copying them during distribution builds (I suspect this is what this pull request tried to do by elevating FIPS to become mandatory dependency to one of the modules). 
 2. Or, alternatively, if we target only FIPS-140-3 as the end goal, we could make -Pcrypto.standard=FIPS-140-3 a default - that would always produce a single flavor of the distribution as well, including all the JVM related flags

cwperks: for #2, how will dependency management work? Currently, it copies BCFips jars under lib/ when creating the distribution when running with `-Pcrypto.standard=FIPS-140-3`. How do downstream plugins know that those are available upstream? I suppose it is only the bcfips deps so plugins can update build logic accordingly, but that being said I don't see a big difference between 1 and 2, esp if the suggestion for 2) is to run with Pcrypto.standard=FIPS-140-3 by default.

reta: >  How do downstream plugins know that those are available upstream? 

This will come with the distribution - and would require plugins to use the proper scope (runtime), which is fine. I suspect `-Pcrypto.standard=FIPS-140-3` would also require JVM arguments to use approved only algorithms (if I am not mistaken).

cwperks: it just seems like inverse of the current situation. i.e. we switch to default to building with `-Pcrypto.standard=FIPS-140-3` so if you don't want the BCFIPS deps then you can build from src/ w/o passing `-Pcrypto.standard=FIPS-140-3` and we still maintain the 2 distros: 1) min and 2) default.

I think it makes sense to have these deps available by default, but also happy to instruct anyone making changes in plugins to take `-Pcrypto.standard=FIPS-140-3` into account and ensure to treat bcfips deps as runtime deps when that flag is passed.

nibix: @reta @cwperks Chiming in here late, I read through quite a lot of stuff related to FIPS, but as this is a quite complex topic there is a big chance that I am missing something. So, apologies in advance if I get something wrong.

I have a question about this:

>>  How do downstream plugins know that those are available upstream?
> This will come with the distribution - and would require plugins to use the proper scope (runtime), which is fine. I suspect -Pcrypto.standard=FIPS-140-3 would also require JVM arguments to use approved only algorithms (if I am not mistaken).

@reta Could you elaborate what you mean with "use the proper scope (runtime)" in the context of plugins?

Do you mean something like this in the dependencies of a plugin?

```
dependencies {
    runtimeOnly "org.bouncycastle:bc-fips:..."
    runtimeOnly "org.bouncycastle:bcpkix-fips:$..."
    ...
```

reta: Yes, correct @beanuwave , so the behavior will be shaped by distribution at runtime (`bc-fips` presence / absence)

nibix: To be honest, I kind of struggle to understand it. I guess I get something wrong.

My understanding of such a runtimeOnly dependency for a plugin would be the following:

1. The plugin cannot reference the classes directly in the source code
2. Rather they will be only available in the runtime class path to have indirect effects or be accessed by reflection.
3. This is achieved by adding the dependency to the plugin zip bundle

I guess 1 and 2 are fine, but won't 3 cause a jar hell exception?

cwperks: When bcFips is provided from the core (i.e. when core built with `-Pcrypto.standard=FIPS-140-3`), then security would treat bcFips as a `compileOnly` dependency. In the absence of that build param then security would treat as an `implementation` dependency.

nibix: but that effectively means that such plugins need to be built two times, once including the jar, once excluding it, right?

cwperks: > but that effectively means that such plugins need to be built two times, once including the jar, once excluding it, right?

it does mean that there are different ways of assembling yes.

That being said, I suppose we would have a choice with the min and default distributions to use `-Pcrypto.standard=FIPS-140-3` always which essentially makes it a default.

As I was saying above, its essentially the inverse of the current situation where FIPS is non-default and needs to be built separately. If `-Pcrypto.standard=FIPS-140-3` is the default for the build (and we still only support 2 distros min and default), then you would need to specifically build `-Pcrypto.standard=None` from source.

andrross: I believe the crux of @reta's argument here is that adding an `api` dependency on the BC libraries at essentially the root of the dependency tree is a very heavy-handed approach and introduces tight coupling (a compile-time dependency) on BC libraries where it is not otherwise necessary. This goes against architectural best practices and is likely to cause problems long-term even if it works in the short term. @reta is that right? Am I missing anything?

However, I do like the goal of having a single distribution as opposed to adding another dimension on which we'd have to build artifacts for. I think we're all in agreement on this? @reta @cwperks @beanuwave please correct me if I'm wrong. Are there any cases were we foresee a problem by including the BC libraries by default (other than a one time update to the security plugin to not bring in duplicate libraries once they are included by default)?

As for _how_ we go about making this change, I would lean towards making `-Pcrypto.standard=FIPS-140-3` a default option, and then change the security plugin to take a `compileOnly` dependency on bouncycastle. I think it must be compileOnly because the libraries will be provided at runtime, but the security plugin code still needs the code on the classpath at compile time.

reta: Thanks @andrross 

> This goes against architectural best practices and is likely to cause problems long-term even if it works in the short term. @reta is that right? Am I missing anything?

💯 That is correct.

> However, I do like the goal of having a single distribution as opposed to adding another dimension on which we'd have to build artifacts for. I think we're all in agreement on this?

This is also correct, I think we aligned on that

> As for how we go about making this change, I would lean towards making -Pcrypto.standard=FIPS-140-3 a default option, and then change the security plugin to take a compileOnly dependency on bouncycastle. I think it must be compileOnly because the libraries will be provided at runtime, but the security plugin code still needs the code on the classpath at compile time.

That would also make it I believe, in this case the distribution will bundle BC fips libraries whereas `security` plugin (and may be other plugins) just assumes those are there. The dependencies also stay clear, I think this is good option. Thank you @andrross 

@cwperks @beanuwave wdyt folks?

cwperks: yes, sounds like a plan to me and its good to have clarity on the way forward. I will close this PR and open an issue on the opensearch-build repo to start building min and default distro with `-Pcrypto.standard=FIPS-140-3`

beanuwave: Sorry for joining the discussion late.

> so the behavior will be shaped by distribution at runtime (bc-fips presence / absence)

This is absolutely right — `opensearch-env` will log _"BouncyCastle FIPS library found, setting FIPS JVM options"_ based on whether BC libraries are present on the classpath. Also, the current implementation does not support switching FIPS mode on/off at runtime. The original requirements did not require this capability, which is why it was considered initially but later dropped.

> >As for how we go about making this change, I would lean towards making -Pcrypto.standard=FIPS-140-3 a default option, and then change the security plugin to take a compileOnly dependency on bouncycastle. I think it must be compileOnly because the libraries will be provided at runtime, but the security plugin code still needs the code on the classpath at compile time.

> I will close this PR and open an issue on the opensearch-build repo to start building min and default distro with -Pcrypto.standard=FIPS-140-3

Bundling BC libraries would tremendously simplify the FIPS setup for core and might be the answer to avoid extra distributions on top. However, we would be putting the cart before the horse by making FIPS the new default. The restrictions, trade-offs, and cut-offs required to make FIPS work are significant and continue to evolve with the standard. That said, this could become an interesting topic once the security-plugin receives its FIPS-related changes, along with the other ~25 plugins included in the default distribution.
@nibix @andrross @reta @cwperks

Issue #6777, opened 2023-03-21
Fuzziness support for `bool_prefix`-type `multi_match` queries


haversnail:
**Is your feature request related to a problem? Please describe.**
Yes, see this [open issue](https://github.com/elastic/elasticsearch/issues/56229) from Elasticsearch. TL;DR, querying `search_as_you_type` fields with [the recommended](https://opensearch.org/docs/2.6/field-types/search-as-you-type/) `multi_match` `bool_prefix` query does not take into account any "fuzziness" value.

**Steps to reproduce:**

`PUT /my_index`
```jsonc
{
  "mappings": {
    "properties": {
      "field": {
        "my_field": "search_as_you_type"
      }
    }
  }
}
```

`PUT /my_index/_doc/1`
```jsonc
{
  "my_field": "license"
}
```

`GET /my_index/_search`
```jsonc
{
  "query": {
    "multi_match": {
      "query": "lisence", // misspelled word
      "type": "bool_prefix",
      "fuzziness": 1,
      "fields": [
        "my_field",
        "my_field._2gram",
        "my_field._3gram"
      ]
    }
  }
}
```

Currently, the steps above yield the following result:
```jsonc
{
  // ...
  "hits": {
    "total": {
      "value": 0,
      "relation": "eq"
    },
    "max_score": null,
    "hits": []
  }
}
```

**Describe the solution you'd like**
It would be beneficial and intuitive to have fuzziness applied to the final search term, so that results are still returned as they would be for other common datatypes/query types with a `fuzziness` property specified.

**Describe alternatives you've considered**
- *[Adding a space to the end of the query value](https://github.com/elastic/elasticsearch/issues/56229#issuecomment-1302282050)*
	- While this works for complete word queries, this results in edge n-grams no longer being included in the search results. For example, `lic` would no longer return the document with the term `license` as it would normally.
- *Including a `suggestion` block to the query*
	- Although this solution "works," this requires an additional lookup with the suggested/proper term, as well as time/effort spent on the client side redesigning a workflow around this.

epiphanatic:
This issue has been known about for about 3 years - Elastic search, I hope you guys realize that this exact issue is your competitors greatest strength

dblock:
@epiphanatic care to contribute, maybe start by adding (failing) [YAML REST tests](https://github.com/opensearch-project/OpenSearch/blob/main/TESTING.md#testing-the-rest-layer) that express the behavior you're looking for?

[...]

PR #20401, opened 2026-01-09
Fuzziness support for bool prefix type and fuzzy match queries

deshsidd:
Querying search_as_you_type fields with [the recommended](https://opensearch.org/docs/2.6/field-types/search-as-you-type/) multi_match bool_prefix query does not take into account any "fuzziness" value. This PR aims to fix issue.

### Description

### Related Issues
Addresses: #6777 

### Check List

### DIFF
--- a/server/src/internalClusterTest/java/org/opensearch/search/query/BoolPrefixFuzzinessIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/search/query/BoolPrefixFuzzinessIT.java
@@ -1,0 +1,118 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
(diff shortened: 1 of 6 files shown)


atris:
Thanks for this, @deshsidd

I noticed a couple of things that might bite us here, mostly around graph queries/synonyms.

First, in MatchQuery.java (L790), the new loop for analyzeMultiBoolean seems to force an AND on synonyms when fuzziness is on. If I search for wifi (synonym wi-fi) with operator: AND, it adds +wifi +wi-fi. Since a doc rarely has both terms within the fuzzy edit distance of each other, this will kill matches. We probably need a nested disjunction (SHOULD clause) for those synonyms.

Second, the `while (source.incrementToken())` loop in createBooleanPrefixQuery ignores position increments. This flattens the graph, which breaks things for synonyms and shingles (pretty common for bool_prefix / search_as_you_type). We should probably respect the graph structure there too, similar to standard match queries.

Thoughts on how we want to handle that?

PR #20355, opened 2026-01-02
Adding Index Setting for Data compression

darjisagar7:
<!--  Thanks for sending a pull request, here are some tips:

1. If this is a fix for an undisclosed security vulnerability, please STOP. All security vulnerability reporting and fixes should be done as per our security policy https://github.com/opensearch-project/OpenSearch/security/policy
2. If this is your first time, please read our contributor guidelines: https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md and developer guide https://github.com/opensearch-project/OpenSearch/blob/main/DEVELOPER_GUIDE.md
3. Ensure you have added or ran the appropriate tests for your PR: https://github.com/opensearch-project/OpenSearch/blob/main/TESTING.md
-->

### Description
[Describe what this change achieves]

```
>....                                                                                                                                                                                                              
        "compression.enabled": "true",
        "optimized.enabled": "true",
        "index.parquet.page_row_limit": 50000,
        "index.parquet.compression_type": "ZSTD"
    },
    "mappings": {
        "properties": {
            "id": {
                "type": "keyword"
            },
            "name": {
                "type": "keyword"
            },
            "age": {
                "type": "integer"
            },
            "salary": {
                "type": "long"
            },
            "score": {
                "type": "double"
            },
            "active": {
                "type": "boolean"
            },
            "created_date": {
                "type": "date"
            }
        }
    }
}'

{"acknowledged":true,"shards_acknowledged":true,"index":"index-7"}%                                                                                                                                                                                                                                                                                         
```

```
parquet-tools inspect OpenSearch/build/testclusters/runTask-0/data/nodes/0/indices/jhYkYQ8PQNGg-UAsLQEq8g/0/parquet/_parquet_file_generation_16.parquet

############ file meta data ############
created_by: parquet-rs version 53.4.1
num_columns: 15
num_rows: 6
num_row_groups: 1
format_version: 1.0
serialized_size: 2793


############ Columns ############
_routing
_doc_count
active
salary
_ignored
_seq_no
score
name
_id
id
created_date
_version
age
___row_id
_primary_term

############ Column(_routing) ############
name: _routing
path: _routing
max_definition_level: 1
max_repetition_level: 0
physical_type: BYTE_ARRAY
logical_type: String
converted_type (legacy): UTF8
compression: ZSTD (space_saved: -47%)

############ Column(_doc_count) ############
name: _doc_count
path: _doc_count
max_definition_level: 1
max_repetition_level: 0
physical_type: INT64
[...]

PR #11027, opened 2023-10-31
Enable Fuzzy codec for doc id fields using a bloom filter

mgodwan:
### Description
Enable Fuzzy codec for doc id fields using a bloom filter. This provides added performance advantage for update and get operations.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/4489

### Check List
- [x] New functionality includes testing.
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
- [x] Failing checks are inspected and point to the corresponding known issue(s) (See: [Troubleshooting Failing Builds](../blob/main/CONTRIBUTING.md#troubleshooting-failing-builds))
- [x] Commits are signed per the DCO using --signoff
- [x] Commit changes are listed out in CHANGELOG.md file (See: [Changelog](../blob/main/CONTRIBUTING.md#changelog))
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #20404, opened 2026-01-12
Adding bloom_filter_enable field mapping parameter to allow field level bloom filter for Parquet

cocosz:
### Description

This PR adds a new `bloom_filter_enable` field mapping parameter that allows users to enable Parquet bloom filters at the field level. The implementation provides:

- Works with all major field types (text, keyword, numeric, date, boolean, binary, ip)
- Extends the base `ParametrizedFieldMapper` class without modifying individual field mappers
- Uses `MapperService` to collect field configurations and pass them to the Rust Parquet writer
- Bloom filter configurations are properly inherited during Parquet file merge operations
- Defaults to `false` (disabled) to maintain existing behavior

**Usage Example:**
```json
PUT /my_index
{
  "mappings": {
    "properties": {
      "user_id": {
        "type": "keyword",
        "bloom_filter_enable": true
      },
      "timestamp": {
        "type": "date",
        "bloom_filter_enable": true
      }
    }
  },
  "settings": {
    "index.optimized.enabled": true
  }
}

### DIFF
--- a/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustBridge.java
+++ b/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustBridge.java
@@ -3,6 +3,7 @@ package com.parquet.parquetdataformat.bridge;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Map;
 
 /**
  * JNI bridge to the native Rust Parquet writer implementation.
@@ -29,8 +30,7 @@ public class RustBridge {
     public static native void initLogger();
 
     // Enhanced native methods that handle validation and provide better error reporting
-    public static native void createWriter(String file, long schemaAddress) throws IOException;
-    public static native void setBloomFilterConfig(String file, String fieldName, boolean enabled, double fpp, long ndv) throws IOException;
+    public static native void createWriter(String file, long schemaAddress, Map<String, Boolean> bloomFilterFields) throws IOException;
     public static native void write(String file, long arrayAddress, long schemaAddress) throws IOException;
     public static native ParquetFileMetadata closeWriter(String file) throws IOException;
     public static native void flushToDisk(String file) throws IOException;
@@ -40,5 +40,5 @@ public class RustBridge {
 
 
     // Native method declarations - these will be implemented in the JNI library
-    public static native void mergeParquetFilesInRust(List<Path> inputFiles, String outputFile);
+    public static native void mergeParquetFilesInRust(List<Path> inputFiles, String outputFile, Map<String, Boolean> bloomFilterFields);
 }
--- a/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustBridge.java
+++ b/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustBridge.java
@@ -30,6 +30,7 @@ public class RustBridge {
 
     // Enhanced native methods that handle validation and provide better error reporting
     public static native void createWriter(String file, long schemaAddress) throws IOException;
+    public static native void setBloomFilterConfig(String file, String fieldName, boolean enabled, double fpp, long ndv) throws IOException;
     public static native void write(String file, long arrayAddress, long schemaAddress) throws IOException;
     public static native ParquetFileMetadata closeWriter(String file) throws IOException;
     public static native void flushToDisk(String file) throws IOException;
--- a/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/vsr/VSRManager.java
+++ b/modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/vsr/VSRManager.java
@@ -11,7 +11,6 @@ package com.parquet.parquetdataformat.vsr;
 import com.parquet.parquetdataformat.bridge.ArrowExport;
 import com.parquet.parquetdataformat.bridge.NativeParquetWriter;
 import com.parquet.parquetdataformat.bridge.ParquetFileMetadata;
-import com.parquet.parquetdataformat.bridge.RustBridge;
 import com.parquet.parquetdataformat.memory.ArrowBufferPool;
 import com.parquet.parquetdataformat.writer.ParquetDocumentInput;
 import org.apache.arrow.vector.types.pojo.Schema;
@@ -39,6 +38,8 @@ import java.util.concurrent.atomic.AtomicReference;
 public class VSRManager implements AutoCloseable {
 
     private static final Logger logger = LogManager.getLogger(VSRManager.class);
+
+    private final String BLOOM_FILTER_ENABLE_PARAM = "bloom_filter_enable";
 
     private final AtomicReference<ManagedVSR> managedVSR = new AtomicReference<>();
     private final Schema schema;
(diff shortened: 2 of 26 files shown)

### CODE REVIEW

File: modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustBridge.java [hunk 1]
```diff
@@ -30,6 +30,7 @@ public class RustBridge {
 
     // Enhanced native methods that handle validation and provide better error reporting
     public static native void createWriter(String file, long schemaAddress) throws IOException;
+    public static native void setBloomFilterConfig(String file, String fieldName, boolean enabled, double fpp, long ndv) throws IOException;
```
darjisagar7: Instead of updating field by field, can we not pass the exact map to the rust in one go?

modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/vsr/VSRManager.java
```diff
@@ -61,13 +70,56 @@ public VSRManager(String fileName, Schema schema, ArrowBufferPool arrowBufferPoo
...
             try (ArrowExport export = managedVSR.get().exportSchema()) {
+                int totalConfigs = fieldConfigs.values().stream().mapToInt(Map::size).sum();
+                logger.info("Initializing writer for file: {} with {} field configurations", fileName, totalConfigs);
+
+                applyFieldConfigurations();
+
+                logger.info("Creating native Parquet writer for file: {}", fileName);
                 writer = new NativeParquetWriter(fileName, export.getSchemaAddress());
+                logger.info("Successfully initialized Parquet writer for file: {}", fileName);
             }
```
darjisagar7: Do we need all this logging here?

server/src/main/java/org/opensearch/index/mapper/ParametrizedFieldMapper.java
```diff
@@ -79,6 +79,9 @@ public abstract class ParametrizedFieldMapper extends FieldMapper {
 
     private static final DeprecationLogger deprecationLogger = DeprecationLogger.getLogger(ParametrizedFieldMapper.class);
 
+
+    public static final String BLOOM_FILTER_ENABLE_PARAM = "bloom_filter_enable";
```
darjisagar7, line 83: Do we need this setting at index level?

server/src/main/java/org/opensearch/index/mapper/MappedFieldType.java
```diff
@@ -101,6 +102,18 @@ public MappedFieldType(
...
+    ) {
+        this(name, isIndexed, isStored, hasDocValues, false, textSearchInfo, meta);
+    }
+
+    public MappedFieldType(
+        String name,
+        boolean isIndexed,
+        boolean isStored,
+        boolean hasDocValues,
+        boolean bloomFilterEnabled,
```
mgodwan: Instead of changing the constructor, can we use a setter? Similar to how it is done for boost, indexAnalyser, etc?

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -719,6 +722,63 @@ public ObjectMapper getObjectMapper(String name) {
...
+
+        T defaultValue = null;
+        boolean defaultValueFound = false;
+
+        for (Mapper mapper : this.mapper.mapping().root()) {
+            logger.debug("Processing mapper: {} (type: {})", mapper.simpleName(), mapper.getClass().getSimpleName());
+
+            if (mapper instanceof ParametrizedFieldMapper) {
+                ParametrizedFieldMapper fieldMapper = (ParametrizedFieldMapper) mapper;
+                ParametrizedFieldMapper.Builder builder = fieldMapper.getMergeBuilder();
```
mgodwan: Why do we need to create this builder? Can we expose something on MappedFieldType which is generic and also prevents creating this builder instance for every writer?

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -719,6 +722,63 @@ public ObjectMapper getObjectMapper(String name) {
...
+
+                if (builder != null) {
+                    if (!defaultValueFound) {
```
mgodwan: This can be removed?

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -719,6 +722,63 @@ public ObjectMapper getObjectMapper(String name) {
...
+                        defaultValue = builder.getParameterDefaultValue(parameterName, parameterType);
+                        defaultValueFound = true;
+                        logger.debug("Found default value for parameter '{}': {}", parameterName, defaultValue);
+                    }
+
+                    T parameterValue = builder.getParameterValue(parameterName, parameterType);
+                    logger.debug("Field '{}': parameter '{}' = {} (default: {})",
+                        fieldMapper.simpleName(), parameterName, parameterValue, defaultValue);
+
+                    if (parameterValue != null && !parameterValue.equals(defaultValue)) {
```
mgodwan: This seems buggy as 
1. If default and explicit are same, we won't enter the condition.
2. defaultValue is only set for first encountered field. This check won't be correctly evaluate for any subsequented fields.

modules/parquet-data-format/src/main/rust/src/lib.rs
```diff
@@ -26,12 +26,31 @@ pub use vectorized_exec_spi::{log_info, log_error, log_debug};
 lazy_static! {
     static ref WRITER_MANAGER: DashMap<String, Arc<Mutex<ArrowWriter<File>>>> = DashMap::new();
     static ref FILE_MANAGER: DashMap<String, File> = DashMap::new();
+    static ref BLOOM_FILTER_CONFIGS: DashMap<String, DashMap<String, BloomFilterConfig>> = DashMap::new();
```
mgodwan: Why do we need this map? These conflgs should be a part of writer creation args only

mgodwan: This BLOOM_FILTER_CONFIGS maps will grow with each file. That will cause memory issues.

mgodwan (changes requested):

server/src/main/java/org/opensearch/index/mapper/ParametrizedFieldMapper.java
cocosz: Yes we need a field level setting

kkewwei:
@cocosz It this Bloom Filter sames with the one from #11027? I also plan to add BloomFilter support for fields to improve the efficiency of term lookups on high-cardinality data (haven’t started the work yet). It seems there’s a configuration conflict.

modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/bridge/RustBridge.java
```diff
@@ -29,7 +30,7 @@ public class RustBridge {
     public static native void initLogger();
 
     // Enhanced native methods that handle validation and provide better error reporting
-    public static native void createWriter(String file, long schemaAddress) throws IOException;
+    public static native void createWriter(String file, long schemaAddress, Map<String, Boolean> bloomFilterFields) throws IOException;
```
darjisagar7, line 33: You can refer to this PR as passing individual writer config as a parameter will need more number of inputs rather we can pass one object as a json string

https://github.com/opensearch-project/OpenSearch/pull/20355

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -719,6 +722,56 @@ public ObjectMapper getObjectMapper(String name) {
...
+     * Parameters are accessed directly from MappedFieldType to avoid creating builder instances.
+     *
+     * @param parameterName the name of the parameter to extract (e.g., "bloom_filter_enable")
+     * @param parameterType the type of the parameter (e.g., Boolean.class)
+     * @param <T> the type of the parameter value
+     * @return map of field names to parameter values (only non-default values)
+     * @throws IllegalArgumentException if the parameter is not supported
+     */
+    @SuppressWarnings("unchecked")
+    public <T> Map<String, T> getFieldConfigurations(String parameterName, Class<T> parameterType) {
```
darjisagar7, line 738: Can we make this function more generic so that it can be used for other configs also.

Issue #20050, opened 2025-11-19
[RFC] Proposal for Moving KNN interfaces to Opensearch core


akash-shankaran:
### Is your feature request related to a problem? Please describe

This RFC proposes moving the vector interfaces from KNN plugin to core. This enhancement will provide extensibility to new vector engines in OpenSearch with the ability to adhere to a common set of interfaces, and extend them to other plugins such as neural-search. This is presently high-level proposal, and detailed approaches may follow as needed. 

### Challenge
Vector Search is a growing area, and OpenSearch as a project is well positioned as a project to tackle the challenges of this workload. Currently in OpenSearch, there are 2 plugins which deal with vector search: 1) K-nn 2) JVector plugin. 
Both have their own strengths and uses.
K-nn plugin is designed to support multiple vector engines such as: 

1. Lucene 
2. FAISS 
3. NMSLib (deprecated). 

The native engines are supported using a JNI mechanism within k-nn. 
 
Jvector plugin, which got added in 2025, is designed to support the Jvector vector engine[2], which may be extended in the future to other engines. 
There are 2 challenges that we see here: 
1. **Extensibility (or lack of) of KnnVector interfaces**: There are no standardized interfaces across the vector plugins dealing with the vector engines, and core OpenSearch. 
2. **Decoupling other plugins from relying on the vector plugins**: A second problem that this proposal aims to alleviate is the dependency of neural-search plugin (used for Hybrid search) directly on the K-nn plugin, which prevents adding the JVector engine.[1]

### Motivation
All the plugins today adhere to Lucene’s interface constructs for KnnVectors.
This is true for the JVector and Lucene engines, which are implemented in Java.
FAISS/Nmslib are native engines within the k-nn plugin, which also adhere to the Lucene interfaces with additional layers within themselves. 
This shows the common dependency on Lucene Knn interfaces amongst the various vector engines. A previous proposal [3] tried addressing this problem within the K-nn plugin. However, given how vectors evolved since then within the OpenSearch project (i.e. new plugin development via opensearch-jvector [4]), it might be best to uplevel this approach and move the relevant code and interfaces to Opensearch-core. 

### Scope
This proposal will cover the following: 

- Generate discussion around the vector interfaces, that’ll benefit from moving to core. 
- Improve the extensibility around vectors, to simplify working with new vector engines by improving the interfaces.
- Improve the use of the new vector engines (e.g. jvector via jvector-plugin) in other parts of OpenSearch, such as neural-search. 
[...]

PR #20411, opened 2026-01-13
Introduce AdditionalCodecs and EnginePlugin::getAdditionalCodecs hook to allow additional Codec registration

reta:
### Description
Introduce `AdditionalCodecs` and `EnginePlugin::getAdditionalCodecs` hook to allow additional `Codec` registration. A while ago we introduced  `CodecServiceFactory` to provide an ability for the plugins to supply own custom codecs. It does the job and is actively used by:
 - `custom-codecs`
 - `k-nn`
 - `neural-search`
 - `security-analytics`
 - `jvector`

One of the major limitations that `CodecServiceFactory` has is composability - for particular index, only one  `CodecServiceFactory` could be provided, otherwise the exception is going to be raise (see please https://forum.opensearch.org/t/knn-doesnt-coexist-with-seismic-sparse/27709 as one of the recent examples). 

Currently, there are two major uses of  `CodecServiceFactory`, mutually exclusive in the scope of the same index:
 - register additional codec with non-trivial instantiation (`custom-codecs` fe)
 - wrap every single codec (`k-nn`, `neural-search`, `security-analytics`, `jvector`)

This pull request is the first step towards replacing `CodecServiceFactory` with composable alternatives: introduce new `AdditionalCodecs` and `EnginePlugin::getAdditionalCodecs` hook to allow additional `Codec` registration, without the need to provide `CodecServiceFactory` / own `CodecService` instance.

The change is non-breaking and is fully compatible with all existing plugins (although it deprecates the `CodecService` constructor). The next steps are:
- replacing `CodecService` constructor usage (`k-nn`, `neural-search`, `security-analytics`, `jvector`)
- replacing  `getCustomCodecServiceFactory` with `getAdditionalCodecs` (`custom-codecs`)

### Related Issues
Part of https://github.com/opensearch-project/OpenSearch/issues/20050

### Check List

## Summary by CodeRabbit

* **New Features**
  * Plugins can contribute codec registries via a new extension point to supply additional codecs.

* **Refactor**
  * Codec initialization now composes codecs from registries with conflict-resolution behavior.

* **Public API**
  * Codec service construction extended to accept registries; the older constructor is deprecated and callers updated.

* **Tests**
  * Tests updated to use the new construction path.

* **Documentation**
  * CHANGELOG updated to document the codec registry extension.

### DIFF
--- a/server/src/main/java/org/opensearch/index/codec/CodecServiceConfig.java
+++ b/server/src/main/java/org/opensearch/index/codec/CodecServiceConfig.java
@@ -25,18 +25,18 @@ public final class CodecServiceConfig {
     private final IndexSettings indexSettings;
     private final MapperService mapperService;
     private final Logger logger;
-    private final List<AdditionalCodecs> registries;
+    private final List<AdditionalCodecs> additionalCodecs;
 
     public CodecServiceConfig(
         IndexSettings indexSettings,
         @Nullable MapperService mapperService,
         @Nullable Logger logger,
-        List<AdditionalCodecs> registries
+        List<AdditionalCodecs> additionalCodecs
     ) {
         this.indexSettings = Objects.requireNonNull(indexSettings);
         this.mapperService = mapperService;
         this.logger = logger;
-        this.registries = Objects.requireNonNull(registries);
+        this.additionalCodecs = Objects.requireNonNull(additionalCodecs);
     }
 
     public IndexSettings getIndexSettings() {
@@ -53,7 +53,7 @@ public final class CodecServiceConfig {
         return logger;
     }
 
-    public List<AdditionalCodecs> getRegistries() {
-        return registries;
+    public List<AdditionalCodecs> getAdditionalCodecs() {
+        return additionalCodecs;
     }
 }
--- a/server/src/main/java/org/opensearch/index/engine/EngineConfigFactory.java
+++ b/server/src/main/java/org/opensearch/index/engine/EngineConfigFactory.java
@@ -58,7 +58,7 @@ import java.util.function.Supplier;
 public class EngineConfigFactory {
     private final CodecServiceFactory codecServiceFactory;
     private final TranslogDeletionPolicyFactory translogDeletionPolicyFactory;
-    private final List<AdditionalCodecs> registries;
+    private final List<AdditionalCodecs> additionalCodecs;
 
     /** default ctor primarily used for tests without plugins */
     public EngineConfigFactory(IndexSettings idxSettings) {
@@ -134,7 +134,7 @@ public class EngineConfigFactory {
         final CodecService instance = codecService.orElse(null);
         this.codecServiceFactory = (instance != null) ? (config) -> instance : codecServiceFactory.orElse(null);
         this.translogDeletionPolicyFactory = translogDeletionPolicyFactory.orElse((idxs, rtls) -> null);
-        this.registries = Collections.unmodifiableList(codecRegistries);
+        this.additionalCodecs = Collections.unmodifiableList(codecRegistries);
     }
 
     /**
@@ -212,7 +212,7 @@ public class EngineConfigFactory {
     }
 
     public CodecService newDefaultCodecService(IndexSettings indexSettings, @Nullable MapperService mapperService, Logger logger) {
-        return new CodecService(mapperService, indexSettings, logger, registries);
+        return new CodecService(mapperService, indexSettings, logger, additionalCodecs);
     }
 
     public CodecService newCodecServiceOrDefault(
@@ -222,7 +222,7 @@ public class EngineConfigFactory {
         CodecService defaultCodecService
     ) {
         return this.codecServiceFactory != null
-            ? this.codecServiceFactory.createCodecService(new CodecServiceConfig(indexSettings, mapperService, logger, registries))
+            ? this.codecServiceFactory.createCodecService(new CodecServiceConfig(indexSettings, mapperService, logger, additionalCodecs))
             : defaultCodecService;
     }
 }


server/src/main/java/org/opensearch/index/codec/CodecService.java
```diff
@@ -94,6 +109,29 @@ public CodecService(@Nullable MapperService mapperService, IndexSettings indexSe
...
+                        // replace registered codec with the new one
+                        codecs.put(name, resolved);
+                    } else if (resolved == null) /* nothing to register */ {
+                        // remove registered codec
+                        codecs.remove(name);
+                    }
+                }
+
+            }
+        }
```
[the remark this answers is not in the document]
reta: We cannot do that but document: the default codec could be replaced (in general), so should rely on default codec supplier (as it is now) and hit it may not be used during initialization

msfroh:
Oooh! I need to review this one carefully.

I've put some thought into composable codecs and had more or less concluded that it can't be done elegantly (or at least, I can't think of a way to do it elegantly). 

In particular, order and/or priority matters, unless you can ensure that only one codec is used for each field (where by "field", I'm using it in the Lucene sense, where `SortedSetDocValues` for field `foo` are distinct from the terms + postings for field `foo`.). I don't know of a good way to detect whether the Codecs operate on different fields without modifying the Codec interface. 

The SegmentInfos record the codec used for each segment as strings, and you need to be able to map those back to the SPI for a specific codec. So, for order, `codec1+codec2` is not the same as `codec2+codec1` -- even if the codecs touch different fields (so they're functionally equivalent), I'm not sure how to get the SPI working unless you register all orders or canonicalize the orders.

I look forward to reading through this to understand how you've tackled that problem! (Probably next week, as this week is **really** busy.)

reta:
> I look forward to reading through this to understand how you've tackled that problem! (Probably next week, as this week is **really** busy.)

Thanks a lot @msfroh ! You are absolutely right, this is the second case (I documented them in the description) which is not covered by this particular change (but we would need to look into that soon), thank you! 

PS: With this new hook, we should be able to use custom codecs with any plugin combination.

reta:
@msfroh would really appreciate if you could take a look (taking into account the risk clearance here https://github.com/opensearch-project/OpenSearch/pull/20411#issuecomment-3746431005), it would be great to have it out 3.5.0 but `custom-codecs`  plugin need a small change to benefit from that, wondering if we could make it in time for code freeze. Thanks a mill!

server/src/main/java/org/opensearch/index/codec/CodecRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * In case there is already registered {@code Codec} with the same name, allows to resolve
+     * the conflict by:
+     *   - returning a {@code newCodec}, replaces the {@code oldCodec}
+     *   - returning a {@code oldCodec}, effectively ignores the {@code newCodec}
+     *   - returning a {@code null}, removes the {@code Codec} altogether
+     *
+     * By default, the implementation returns a {@code newCodec} that replaces
+     * the {@code oldCodec}.
```
msfroh: Is this the existing behavior? Is there existing behavior?

I'm thinking maybe just throwing an exception would be a good default.

server/src/main/java/org/opensearch/index/codec/CodecService.java
```diff
@@ -94,6 +110,36 @@ public CodecService(@Nullable MapperService mapperService, IndexSettings indexSe
...
+        final Supplier<Codec> defaultCodec = () -> codecs.get(DEFAULT_CODEC);
+        for (CodecRegistry registry : registries) {
+            final Map<String, Codec> additionalCodecs = registry.getCodecs(mapperService, indexSettings, defaultCodec);
+            for (Map.Entry<String, Codec> additionalCodec : additionalCodecs.entrySet()) {
+                final String name = additionalCodec.getKey();
+
+                // Default codec could not be changed
+                if (name.equalsIgnoreCase(LUCENE_DEFAULT_CODEC) == true) {
+                    continue;
+                }
```
msfroh, line 124: Similar to the name conflict question, I would throw an exception here.

reta:
Thanks @msfroh !

> @reta -- I'm just trying to understand how one would go about implementing plugins that take care of codec composition. Can you provide an example?

This case is not covered in this pull request (there 2 basic flows, I touched on that in description). The composition will be addressed (working on design) but it would be more complex I suspect (due to all the problems that you have described). The case that I am trying to solve here is this: register additional codec with non-trivial instantiation (`custom-codecs` fe) but not composition just yet.

Just quick follow up, the way plugins do implement composition right now is overriding `CodecService::getCodec` blindly, the areas I am looking at are (those are just ideas):
 - does it have to be this way? (may be for some, but not for all)
 - some custom codecs could veto composition 
 - some custom codecs could explicitly state the order the composition may/should happen

Does it make sense? 
Thanks a lot for looking! 

[1] https://forum.opensearch.org/t/codec-error-with-knn/16358

server/src/main/java/org/opensearch/index/codec/CodecRegistry.java
reta: > Is this the existing behavior? Is there existing behavior?

As of today - there is no default in this case, the one who implements CodecService could do whatever it wants to (return existing codec or new one, do not use existing codecs at all, ...)

> I'm thinking maybe just throwing an exception would be a good default.

Certainly, thank you!

msfroh:
> Just quick follow up, the way plugins do implement composition right now is overriding `CodecService::getCodec` blindly, the areas I am looking at are (those are just ideas):

IMO, there are a bunch of really interesting things we could do there. 

I would lean toward the idea of OpenSearch having **one** Codec that's linked to the underlying Lucene codec. Then we can get into mapping-based per-field formats. So, instead of tying a pluggable Codec to an index, we would connect it to individual field. We have something in that direction with PerFieldMappingPostingFormatCodec, but I would cover **all** the per-field formats (which for now include vectors and doc values). Essentially, we would use the mappings to keep track of which codec to use for which field.

At that point, I would potentially lean towards plugins being able to return `PostingFormat`, `DocValuesFormat`, and `KnnVectorsFormat`, instead of returning a `Codec`. Each field would be able to customize one or more of those, but you wouldn't be able to do composition at the field level.

The big problem (IMO) is that we can't easily migrate per-field formats across versions, because we don't have per-field information in the SegmentInfos. If a format delegates to Lucene, it's fine -- we'll still track the built-in Lucene codec versions. I don't have an easy solution for the case when you have old segments written in MyCustomVectorFormatV3 and you want to write new segments written in MyCustomVectorFormatV4. The best I can suggest is that you just have MyCustomVectorFormat and it manages its own versions in per-field metadata. (So, the newer version that writes V4 still knows how to read V3.)

What do you think?

Edit: Oh... also all of the above is roughly "how I think it should be", mostly ignoring how things are today. I *think* we could offer a new per-field way that would be incompatible with the old per-index way, but support both in read/write mode for OpenSearch 3, then disable creating indexes the old way in OpenSearch 4, finally removing read support for the old way in OpenSearch 5.

reta:
> I would lean toward the idea of OpenSearch having **one** Codec that's linked to the underlying Lucene codec. Then we can get into mapping-based per-field formats. So, instead of tying a pluggable Codec to an index, we would connect it to individual field.

I would very much agree on this vision, logically it shouldn't be per index per per field indeed 

> At that point, I would potentially lean towards plugins being able to return PostingFormat, DocValuesFormat, and KnnVectorsFormat, instead of returning a Codec. Each field would be able to customize one or more of those, but you wouldn't be able to do composition at the field level.

Let me explore this route, I haven't thought about it this way

> Edit: Oh... also all of the above is roughly "how I think it should be", mostly ignoring how things are today. I think we could offer a new per-field way that would be incompatible with the old per-index way, but support both in read/write mode for OpenSearch 3, then disable creating indexes the old way in OpenSearch 4, finally removing read support for the old way in OpenSearch 5.

I am all in to make it better. Let me move this pull request to draft state and spend more time to explore the design space that covers composition and additional codecs registration. 

Thanks a lot @msfroh !

navneet1v:
This is a great change would love to see this merged.

navneet1v:
> At that point, I would potentially lean towards plugins being able to return PostingFormat, DocValuesFormat, and KnnVectorsFormat, instead of returning a Codec. Each field would be able to customize one or more of those, but you wouldn't be able to do composition at the field level.

+1 on this thought. I have been thinking on this but never got time to actually move forward with this. One challenge I see is, some formats are really per field formats and plugins returning them is pretty straight forward like KNNVectors, PostingsFormat etc. The difficulty we are going to face is on the formats which are 1 per index like StoredFieldFormat. Would love to know what we think over that? 

One of the best thing to do would be able to have a k-NN/jVector index with `zstd` compression stored field.

navneet1v:
@reta let me know when the PR is ready for review. I can help in reviewing the code, even though I am not a maintainer of core, but I can help in review if needed. :)

reta:
> +1 on this thought. I have been thinking on this but never got time to actually move forward with this. One challenge I see is, some formats are really per field formats and plugins returning them is pretty straight forward like KNNVectors, PostingsFormat etc. The difficulty we are going to face is on the formats which are 1 per index like StoredFieldFormat. Would love to know what we think over that?

Thanks @navneet1v , I have spent a last few days on prototyping different options, `StoredFieldFormat` (and `CompoundFormat`, just to mention) does indeed complicates the design. I think I found the possible way to implement that for the case when there is not ambiguity between plugins chasing same fields). I am thinking towards opening RFC shortly (a few more edge cases to cover) with the suggested extension points for per field formats (DocValues / KnnVectors), including StoredFields / Compound ones.

> One of the best thing to do would be able to have a k-NN/jVector index with zstd compression stored field.

Yes, this change unblocks that, thank you.

reta:
> @reta let me know when the PR is ready for review. I can help in reviewing the code, even though I am not a maintainer of core, but I can help in review if needed. :)

Thanks @navneet1v , I think we should be good to go with this change. @msfroh RFC for per field extension points to follow (https://github.com/opensearch-project/OpenSearch/pull/20411#issuecomment-3805458840), I expect it to have much larger scope (and complexity). 

With respect to this pull request, I did a few things:
 - renamed `CodeRegistry` to `AdditionalCodecs` to reflect the fact it will be appicable more broadly
 - completely removed the `Codec` names conflict resolution in favor of throwing an exception immediately, that would eliminate any ambiguity on plugins load ordering / codec wrapping 

I hope it make sense, thank you!

reta:
@msfroh wondering if we could move forward with this one (taking into account rough edges removed and we have an RFC to work on, this change should unblock `zstd` & co as default codecs), thanks!

msfroh:
> @msfroh wondering if we could move forward with this one (taking into account rough edges removed and we have an RFC to work on, this change should unblock `zstd` & co as default codecs), thanks!

Sounds good to me. Thanks @reta!

PR #20417, opened 2026-01-14
Allow removing plugin that's optionally extended

### LINKED ISSUE #20387: [BUG] Workload management plugin cannot be upgraded in-situ
### Describe the bug

As per documentation, in order to upgrade a cluster one needs to upgrade the plugins first, so that during the rolling restart the nodes pick up the new upgraded plugin version.

In a cluster that has [workload-management](https://docs.opensearch.org/latest/tuning-your-cluster/availability-and-recovery/workload-management/wlm-feature-overview/) plugin installed, the upgrade procedure (i.e., remove and install) does not work, because:

```bash
./bin/opensearch-plugin remove workload-management
ERROR: plugin [workload-management] cannot be removed because it is extended by other plugins: [opensearch-security]
```

Should this safety check be dropped so that the upgrade procedure is unblocked?

As workaround, one can manually move/delete the plugin folder (typically at `/usr/share/opensearch/plugins/workload-management`). 

### Related component

Plugins

### To Reproduce

1. Bootstrap a cluster on any version
2. Install `workload-management` plugin (`./bin/opensearch-plugin install workload-management`)
3. Try to remove it with `./bin/opensearch-plugin remove workload-management`

### Expected behavior

`./bin/opensearch-plugin remove workload-management` should be allowed and not blocked by 
```
ERROR: plugin [workload-management] cannot be removed because it is extended by other plugins: [opensearch-security]
```

### Additional Details

**Plugins**
all bundled ones + workload-management 

 - OS: AlmaLinux
 - Version 9.7

cwperks:
### Description

This is a small PR to fix an issue calling `./bin/opensearch-plugin remove <plugin-name>` on a plugin that is optionally extended from another installed plugin. Given that the plugin is optionally extensible, it means the extending plugin does not require for it to be installed.

Tested by installing workload-management, then security and then removing workload-management.

Before the change it fails with the message from the linked issue. After the change workload-management can be removed as expected.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/20387

### Check List


server/src/main/java/org/opensearch/plugins/PluginsService.java
```diff
@@ -410,7 +410,7 @@ public static List<String> findPluginsByDependency(Path pluginsDir, String plugi
         Set<Bundle> bundles = getPluginBundles(pluginsDir);
         for (Bundle bundle : bundles) {
             for (String extendedPlugin : bundle.plugin.getExtendedPlugins()) {
-                if (extendedPlugin.equals(pluginName)) {
+                if (!bundle.plugin.isExtendedPluginOptional(extendedPlugin) && extendedPlugin.equals(pluginName)) {
```
reta: I think the removal of optional plugin is legit. do you think we could still print a warning (not error) since we don't really know if that may or may not cause issues?

cwperks: Added a warning:

```
./bin/opensearch-plugin remove workload-management
2026-01-14T02:05:53.731208Z main WARN Some features of this plugin may not function without the dependencies being installed.
```

reta: Thanks a lot @cwperks , I would probably consider changing a bit the signature and return the list of required and optional plugins, since we printing a warning prematurely here - we may found non-optional dependency and fail later. The caller of the `findPluginsByDependency` could print the warning.

cwperks: I moved the printing of the warning after collecting the list of plugins that optionally extend the plugin being removed

```
./bin/opensearch-plugin remove workload-management
2026-01-14T15:00:16.171553Z main WARN Some features of [opensearch-security] may not function without the dependency [workload-management] being installed.
-> removing [workload-management]...
```

server/src/main/java/org/opensearch/plugins/PluginsService.java
```diff
@@ -407,14 +407,22 @@ public static void checkJarHellForPlugin(
...
-                    usedBy.add(bundle.plugin.getName());
+                    if (bundle.plugin.isExtendedPluginOptional(extendedPlugin)) {
+                        optionalDeps.add(bundle.plugin.getName());
+                    } else {
+                        usedBy.add(bundle.plugin.getName());
+                    }
                 }
             }
         }
+        if (!optionalDeps.isEmpty()) {
```
reta: If `useBy` is not empty - we could not remove plugin anyway, right? Let us `return (usedBy,  optionalDeps);` and allow the callers to decide what to do (error or log), thanks

cwperks: Pushed a new commit to address this. This is the output now:

```
./bin/opensearch-plugin remove workload-management
WARNING: Some features of [opensearch-security] may not function without the dependency [workload-management].
-> removing [workload-management]...
```

reta (approved):

kmk142789420-stack (approved):

Issue #20433, opened 2026-01-18
[FEATURE] Add `Alt-Svc` header support to advertise HTTP/3 availability


reta:
### Is your feature request related to a problem? Please describe

It is impossible to determine, in advance, whether a target server supports HTTP/3. It is also impossible to upgrade an existing HTTP/1.1 or HTTP/2 connection to an HTTP/3 connection, since HTTP/1.1 and HTTP/2 are built on top of TCP streams while HTTP/3's QUIC is built on top of UDP datagrams. 

### Describe the solution you'd like

By default, if enabled, HTTP/3 transport is available on the same port as HTTP/1.1 and HTTP/2 transports. The `Alt-Svc` header (see please [1]) is the standard way to advertise alternative services (HTTP/2 or HTTP/3).

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Alt-Svc

### Related component

Other

### Describe alternatives you've considered

None

### Additional context

Part of https://github.com/opensearch-project/OpenSearch/issues/4451

PR #20434, opened 2026-01-18
Add Alt-Svc header support to advertise HTTP/3 availability

reta:
### Description
It is impossible to determine, in advance, whether a target server supports HTTP/3. It is also impossible to upgrade an existing HTTP/1.1 or HTTP/2 connection to an HTTP/3 connection, since HTTP/1.1 and HTTP/2 are built on top of TCP streams while HTTP/3's QUIC is built on top of UDP datagrams.

By default, if enabled, HTTP/3 transport is available on the same port as HTTP/1.1 and HTTP/2 transports. The `Alt-Svc` header (see please [1]) is the standard way to advertise alternative services (HTTP/2 or HTTP/3).

Examples:

 - when HTTP/2 is used
```
< HTTP/2 200
< alt-svc: h3=":9200"; ma=3600
< x-opensearch-version: OpenSearch/3.5.0-SNAPSHOT (opensearch)
< content-type: application/json; charset=UTF-8
< content-length: 572
```

 - when HTTP/3 is used
```
< HTTP/3 200
< alt-svc: h2=":9200"; ma=3600
< x-opensearch-version: OpenSearch/3.5.0-SNAPSHOT (opensearch)
< content-type: application/json; charset=UTF-8
< content-length: 572
```

[1] https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Alt-Svc

### Related Issues
Closes https://github.com/opensearch-project/OpenSearch/issues/20433

### Check List

### DIFF
--- a/modules/transport-netty4/src/internalClusterTest/java/org/opensearch/http/netty4/Netty4Http3IT.java
+++ b/modules/transport-netty4/src/internalClusterTest/java/org/opensearch/http/netty4/Netty4Http3IT.java
@@ -22,27 +22,18 @@ import org.opensearch.http.HttpServerTransport;
 import org.opensearch.http.HttpTransportSettings;
 import org.opensearch.http.netty4.http3.Http3Utils;
 import org.opensearch.plugins.Plugin;
-import org.opensearch.plugins.SecureAuxTransportSettingsProvider;
-import org.opensearch.plugins.SecureHttpTransportSettingsProvider;
-import org.opensearch.plugins.SecureSettingsFactory;
-import org.opensearch.plugins.SecureTransportSettingsProvider;
-import org.opensearch.plugins.TransportExceptionHandler;
-import org.opensearch.test.KeyStoreUtils;
+import org.opensearch.test.AbstractSecureSettingsPlugin;
 import org.opensearch.test.OpenSearchIntegTestCase.ClusterScope;
 import org.opensearch.test.OpenSearchIntegTestCase.Scope;
 import org.opensearch.transport.Netty4ModulePlugin;
 import org.opensearch.transport.NettyAllocator;
-import org.opensearch.transport.netty4.ssl.SslUtils;
 
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLException;
-import javax.net.ssl.TrustManagerFactory;
 
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
-import java.util.Optional;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
@@ -52,10 +43,8 @@ import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http2.Http2SecurityUtil;
 import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
-import io.netty.pkitesting.CertificateBuilder.Algorithm;
 import io.netty.util.ReferenceCounted;
 
-import static org.opensearch.test.KeyStoreUtils.KEYSTORE_PASSWORD;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.equalTo;
 import static org.hamcrest.Matchers.containsInAnyOrder;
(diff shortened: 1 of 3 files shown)


reta:
@cwperks @andrross would appreciate if you folks find the time to take a look, closes the gap on HTTP/3 experimental support, thank you!

plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpIT.java
```diff
@@ -0,0 +1,235 @@
...
+
+import static org.opensearch.test.KeyStoreUtils.KEYSTORE_PASSWORD;
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.hasSize;
+
+@ClusterScope(scope = Scope.TEST, supportsDedicatedMasters = false, numDataNodes = 1)
+public class ReactorNetty4HttpIT extends OpenSearchReactorNetty4IntegTestCase {
+    public static final class SecureSettingsPlugin extends Plugin {
```
cwperks: Is there any way this can be abstracted to the test framework if its shared in tests across multiple different subprojects?

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/HttpResponseHeadersFactories.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import static org.opensearch.http.HttpTransportSettings.SETTING_HTTP_HTTP3_ENABLED;
+
+/**
+ * Default implementations for {@link HttpResponseHeadersFactory} for all supported HTTP
+ * protocol versions.
+ */
+final class HttpResponseHeadersFactories {
```
cwperks, line 32: Since this isn't specific to netty, should we consider moving this to a lib and re-using across all the netty plugins?

reta: I was thinking about that to be fair, but is it worth the effort at this moment - not sure, the   `HttpResponseHeadersFactories` are implemented differently for both plugins (since we have to deal with different APIs), the reuse benefits are not very clear, but I could have a spike for it for sure.

plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpIT.java
reta: I think we could move it to the `test-framework`,  let me try it, thanks @cwperks

reta: Done, introduced `AbstractSecureSettingsPlugin` (`test:framework` module) that contains most of the repeatable code

modules/transport-netty4/src/internalClusterTest/java/org/opensearch/http/netty4/Netty4Http3IT.java
```diff
@@ -0,0 +1,159 @@
...
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assume.assumeThat;
+
+@ClusterScope(scope = Scope.TEST, supportsDedicatedMasters = false, numDataNodes = 1)
+public class Netty4Http3IT extends OpenSearchNetty4IntegTestCase {
+    public static final class SecureSettingsPlugin extends AbstractSecureSettingsPlugin {
+        public SecureSettingsPlugin() {
+            super(InsecureTrustManagerFactory.INSTANCE, Http2SecurityUtil.CIPHERS);
```
reta, line 59: `AbstractSecureSettingsPlugin` has no dependencies to any `netty` modules (and so does `test:framework` module), providing trust manager and `SSLEngine` by subclassing `AbstractSecureSettingsPlugin`

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/HttpResponseHeadersFactories.java
reta: Here we are https://github.com/opensearch-project/OpenSearch/pull/20447

PR #20442, opened 2026-01-20
Fix CriteriaBasedCodec to work with delegate codec

RS146BIJAY:
### Description
Fix CriteriaBasedCodec to work with delegate codec

### Related Issues
Fix CriteriaBasedCodec to work with delegate codec


server/src/main/resources/META-INF/services/org.apache.lucene.codecs.Codec
```diff
@@ -1,4 +1,3 @@
 org.opensearch.index.codec.composite.composite912.Composite912Codec
 org.opensearch.index.codec.composite.composite103.Composite103Codec
 org.opensearch.index.codec.composite.backward_codecs.composite101.Composite101Codec
-org.opensearch.index.codec.CriteriaBasedCodec
```
mgodwan, line 4: Is this needed to be registered in SPI?

RS146BIJAY: We won't need register name for Codec as SPI will no longer instantiate new Codec as we are passing delegate codec name in parameter. SPI will instantiate only new PostingsFormat: https://github.com/opensearch-project/OpenSearch/pull/20442/changes#diff-577d44ce3ec453c1808090ac68139e531a6d7ba4a0c65350553967c3e159289b

Bukhtawar: Curious how does this work in tandem with the zstd codec?

reta: I think it won't work as expected for posting formats: the codec should be loaded just fine (CriteriaBasedCodec does use delegate codec name), the posting format however will be the one that the codec supplies (I think that would be the case even for default codec, Lucene103), not the one CriteriaBasedCodec overrides

RS146BIJAY: For [PerFieldPostingsFormat](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldPostingsFormat.java#L64), the Posting format for each field can be registered separately independent of codec. The idea here is to override the behaviour of posting format for only _id field, while all other fields will continue to use delegate codec postings format. The new CriteriaBasedPostingsFormat primarly attaches few attributes for grouping criteria and delegate codec name, but still delegates actual write and read operations to the underlying delegate codec.

So during write time, postings format of _id field will be handled by ```CriteriaBasedPostingsFormat```.  Therefore [fieldInfo attribute that stores the format name](https://github.com/apache/lucene/blob/5bf2835026e896a8ed40e1922fa70892d1eaa873/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldPostingsFormat.java#L302) (```PerFieldPostingsFormat.format```) will be set to CriteriaBasedPostingsFormat for ```_id``` field.  Within this format, the [actual write is still delegated to posting format of delegate codec](https://github.com/opensearch-project/OpenSearch/pull/20442/changes#diff-55238fb5d6bfee98b5294b5c920c6ff36d07446df0827eb8f1224d41465b85b0R95). Additionally, the [delegate postings format's name is saved](https://github.com/opensearch-project/OpenSearch/pull/20442/changes#diff-55238fb5d6bfee98b5294b5c920c6ff36d07446df0827eb8f1224d41465b85b0R115) in a segmentInfos attribute for use during reading.

During read flow, when the per field posting format of _id field is instantiated, [SPI will instantiate a instance of CriteriaBasedPostingsFormat as this is the format name saved in field info attribute](https://github.com/apache/lucene/blob/5bf2835026e896a8ed40e1922fa70892d1eaa873/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldPostingsFormat.java#L302) during write time. Inside CriteriaBasedPostingsFormat, [actual read](https://github.com/opensearch-project/OpenSearch/pull/20442/changes#diff-55238fb5d6bfee98b5294b5c920c6ff36d07446df0827eb8f1224d41465b85b0R80) is still delegated to original posting format, whose name is retrieved from the saved segmentInfo attribute.

Let me know if this makes sense.

reta: > During read flow, when the per field posting format of _id field is instantiated, [SPI will instantiate a instance of CriteriaBasedPostingsFormat as this is the format name saved in field info attribute](https://github.com/apache/lucene/blob/5bf2835026e896a8ed40e1922fa70892d1eaa873/lucene/core/src/java/org/apache/lucene/codecs/perfield/PerFieldPostingsFormat.java#L302) during write time. Inside CriteriaBasedPostingsFormat, [actual read](https://github.com/opensearch-project/OpenSearch/pull/20442/changes#diff-55238fb5d6bfee98b5294b5c920c6ff36d07446df0827eb8f1224d41465b85b0R80) is still delegated to original posting format, whose name is retrieved from the saved segmentInfo attribute.

Thanks @RS146BIJAY , I think it should work (I didn't know about PostingFormat SPIs, thank you). And for non per-field posting, we just use the one that delegate provides, should be fine.

Bukhtawar:
https://build.ci.opensearch.org/job/gradle-check/71107/console
> REPRODUCE WITH: ./gradlew ':server:internalClusterTest' --tests 'org.opensearch.versioning.ConcurrentSeqNoVersioningIT.testSeqNoCASLinearizability' -Dtests.seed=EB27B7DAED0BDF34 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=fr-CH -Dtests.timezone=Asia/Chita -Druntime.java=25
  2> java.lang.AssertionError: ClusterHealthResponse has timed out - returned: [{"cluster_name":"TEST-TEST_WORKER_VM=[835]-CLUSTER_SEED=[5945460847737489031]-HASH=[33719779F34]-cluster","status":"yellow","timed_out":true,"number_of_nodes":5,"number_of_data_nodes":3,"discovered_master":true,"discovered_cluster_manager":true,"active_primary_shards":2,"active_shards":6,"relocating_shards":0,"initializing_shards":0,"unassigned_shards":2,"delayed_unassigned_shards":0,"number_of_pending_tasks":0,"number_of_in_flight_fetch":0,"task_max_waiting_in_queue_millis":0,"active_shards_percent_as_number":75.0}]
    Expected: is <false>
         but: was <true>
        at __randomizedtesting.SeedInfo.seed([EB27B7DAED0BDF34:30E999B0F719FD80]:0)
        at org.opensearch.test.hamcrest.OpenSearchAssertions.assertNoTimeout(OpenSearchAssertions.java:121)
        at org.opensearch.test.OpenSearchIntegTestCase.ensureClusterSizeConsistency(OpenSearchIntegTestCase.java:1112)
        at org.opensearch.test.OpenSearchTestClusterRule.afterInternal(OpenSearchTestClusterRule.java:319)
        at org.opensearch.test.OpenSearchTestClusterRule.after(OpenSearchTestClusterRule.java:188)
        at org.opensearch.test.OpenSearchTestClusterRule$1.evaluate(OpenSearchTestClusterRule.java:374)

PR #20444, opened 2026-01-20
Fix IndexShardTests: Add explicit null-check for Engine reference and fix FileMetadata serialization handling

ask-kamal-nayan:
### Description
This PR fixes failing `IndexShardTests` by addressing two main issues:

1. Adding explicit null-checking when accessing the underlying Engine reference
2. Fixing FileMetadata serialization/deserialization for proper handling of Lucene file names in non-optimized indices

### Problem 1: Engine Null Reference

Several `IndexShardTests` were failing with `NullPointerException` when tests accessed shard methods before the engine was fully initialized or after it was closed.

### Problem 2: FileMetadata Serialization

For non-optimized (standard Lucene) indices, file names were incorrectly being used in serialized format (e.g., `segment_1.si:::lucene`) instead of plain file names (e.g., `segment_1.si`), causing file lookup failures and test issues.

### Problem 3: RemoteStoreRefreshListenerTests Failing:
Because of the changes made to RemoteStoreRefreshListener class to support mulit-format because of which all tests in RemoteStoreRefreshListenerTests were failing.

### Tests Now Passing (4 tests fixed)

1. `testCloseShardWhileEngineIsWarming`
2. `testEstimateTotalDocSize`
3. `testIndexingOperationsListeners`
4. `testSyncSegmentsFromGivenRemoteSegmentStore`

### All RemoteStoreRefreshListenerTests are fixed now:
image

Before this fix 81 tests were passing:
image
After this fix  85 tests are passing. Most of the remaining failed tests are related to translogOps compression  during shardClose call:
image

### Tests:
-  :modules:parquet-data-format:internalClusterTest are passing.
- ParquetRemoteStoreRecoveryIt tests are passing except testReplicaPromotionWithTranslogReplay, also failing in datafusion branch.
- Manual tests going on.

### DIFF
--- a/server/src/main/java/org/opensearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/opensearch/index/shard/IndexShard.java
@@ -5956,15 +5956,14 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
             }
 
             if (file.startsWith(IndexFileNames.SEGMENTS)) {
-                    assert segmentNFile == null : "There should be only one SegmentInfosSnapshot file";
-                    if(isOptimizedIndex()){
-                        segmentNFile = file;
-                    }
-                    else{
-                        segmentNFile = fileMetadata.file();
-                    }
+                assert segmentNFile == null : "There should be only one SegmentInfosSnapshot file";
+                if (isOptimizedIndex()) {
+                    segmentNFile = file;
+                } else {
+                    segmentNFile = fileMetadata.file();
                 }
             }
+        }
 
             if (toDownloadSegments.isEmpty() == false) {
                 try {
--- a/server/src/main/java/org/opensearch/index/store/RemoteSegmentStoreDirectory.java
+++ b/server/src/main/java/org/opensearch/index/store/RemoteSegmentStoreDirectory.java
@@ -331,7 +331,7 @@ public class RemoteSegmentStoreDirectory extends FilterDirectory implements Remo
      */
     @Override
     public void deleteFile(String name) throws IOException {
-        String fileName = new FileMetadata(name).toString();
+        String fileName = new FileMetadata(name).serialize();
         String remoteFilename = getExistingRemoteFilename(fileName);
         if (remoteFilename != null) {
             remoteDataDirectory.deleteFile(remoteFilename);


server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -2541,9 +2541,12 @@ public boolean isRemoteSegmentStoreInSync() {
...
             if (directory.readLatestMetadataFile() != null) {
                 Collection<String> uploadFiles = directory.getSegmentsUploadedToRemoteStore().keySet();
-                try (GatedCloseable<SegmentInfos> segmentInfosGatedCloseable = getSegmentInfosSnapshot()) {
-                    Collection<String> localSegmentInfosFiles = segmentInfosGatedCloseable.get().files(true);
-                    Set<String> localFiles = new HashSet<>(localSegmentInfosFiles);
+                try (CompositeEngine.ReleasableRef<CatalogSnapshot> catalogSnapshotRef = getCatalogSnapshotFromEngine()) {
+                    Collection<FileMetadata> localFileMetadataList = catalogSnapshotRef.getRef().getFileMetadataList();
+                    Set<String> localFiles = localFileMetadataList.stream()
+                        .filter(fm -> !RemoteStoreRefreshListener.EXCLUDE_FILES.contains(fm.file()))
+                        .map(FileMetadata::serialize)
```
raghuvanshraj, line 2548: If these are local file references, why are we using serialize? We should be using `FileMetadata::file` here.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5941,7 +5958,14 @@ private String copySegmentFiles(
...
                     assert segmentNFile == null : "There should be only one SegmentInfosSnapshot file";
-                    segmentNFile = file;
+                    if(isOptimizedIndex())
+                    {
+                        segmentNFile = file;
+                    }
+                    else
+                    {
+                        segmentNFile = fileMetadata.file();
+                    }
```
raghuvanshraj: If we are doing a `new FileMetadata(file)` call, file should already be delimited with data format name, why is the segmentsNFile reference okay to be updated as file here?

raghuvanshraj (changes requested):

server/src/main/java/org/opensearch/index/shard/IndexShard.java
ask-kamal-nayan: There is a check below which checks if the uploadedFiles contains these local files and uploadedFile names are serialized. I have updated the name to serializedLocalFiles for better readability of the code.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
ask-kamal-nayan: These segmentNFiles are used by caller `syncSegmentsFromGivenRemoteSegmentStore`. syncSegmentsFromGivenRemoteSegmentStore API calls OpenInput on this segmentNFile so for normal storeDirectory we need the name in plain file format but for compositeStoreDirectory we need the serialized name so that it can be routed to correct directory.

server/src/main/java/org/opensearch/index/store/RemoteSegmentStoreDirectory.java
```diff
@@ -331,10 +331,11 @@ public String[] listAll() throws IOException {
      */
     @Override
     public void deleteFile(String name) throws IOException {
-        String remoteFilename = getExistingRemoteFilename(name);
+        String fileName = new FileMetadata(name).toString();
```
raghuvanshraj: nit: Can we use serialize directly here?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5958,12 +5957,10 @@ private String copySegmentFiles(
...
-                    {
+                    if(isOptimizedIndex()){
                         segmentNFile = file;
                     }
-                    else
-                    {
+                    else{
                         segmentNFile = fileMetadata.file();
                     }
                 }
```
raghuvanshraj: nit: formatting

Issue #18833, opened 2025-07-25
[Feature Request] IO optimizations in Search


abiesps:
### Is your feature request related to a problem? Please describe

We are proposing a roadmap item for opensearch query performance focused on IO characterstics of queries. 
Most deployments of opensearch uses SSD based storage devices where the IO latencies are really good. But there are use cases where we want to deploy opensearch on non-SSD based storage devices. And we have found that opensearch query performance really suffers in those cases. 

We ran an experiment by running Big5 workload on opensearch deployment with an underlying storage device with performance 3x of a regular SSD. And we found that on opensearch query performance on cold page cache shows a regression of 5x[?] verify these numbers with Sruti (even for a simple term query). 

We believe that there are ways to hide the IO latency of the underlying storage device so that it is not directly translated to query latency. To prove this, we performed a number of experiments to validate our hypothesis and tried to improve the performance of term query on big5 workload. 
And finally with all the optimizations, we were able to bring the query latencies of term query on slower storage device (with the optimization to be comparable) to latencies on SSDs (without the optimization).

Our high level proposal is to make sure that opensearch process is able to saturate the underlying resources(hopefully the CPU) when a query arrives or until we hit the critical chain limit (in cases of slower storage devices) .

In order to achieve this goal of improving cold start query latency, first and foremost, we need benchmarks that exercise the IO path for queries.

Next up, we will use the profiles on these benchmarks to identify opportunities of parallel and async IO.  Lucene already has some form of this implementation through prefetch and inter and intra segment concurrency. However,  these optimizations either don’t have complete coverage in terms of maximum IO parallelism or the defaults are tuned assuming that the data is available in page cache already.  

Third, we are also looking into implementing our own buffer pool with direct IO and io_uring to have more control on IO and memory management. We will instrument the page access path and optimize query path depending on access patterns and data locality.

### Describe the solution you'd like

In principle, we are proposing to make following changes in the opensearch. 

1. Benchmarking enhancement 
2. Non blocking IO and Buffer pool 
3. IO concurrency
4. Prefetching and per query optimizations
5. Fetch phase optimization

**Benchmarking enhancement**

For benchmarking enhancements, we have following proposals. 
1. Build cold start testing in opensearch-benchmarks.  
2. Opensearch benchmarks don’t have enough randomization to simulate an “average” case.  Basically we want to reproduce a scenario where there is a variation in query inputs (field name and values) along with the query types. This will relate more closely to an actual full text search customer workload.
[...]

PR #20446, opened 2026-01-21
[Draft] Optionally add virtual threads for search and index_searcher threadpools 

### LINKED ISSUE #18840: [Feature Request] Virtual thread support in OpenSearch

In continuation to discussion here (https://github.com/opensearch-project/OpenSearch/issues/18833). We have found from our experiments that virtual threads are imperative to achieve maximum io concurrency in opensearch. Virtual threads can also make the problem of statically combining segments into partition a bit simpler, we can also assign a single segment (or part of a segment) to an independent virtual thread. 

We are proposing to use virtual thread as an alternative to  searcher threadpool (and possibly search threadpool) to execute the queries. Virtual threads can also simplify the work assignment algorithms in intra and inter-segment concurrency.

### Related component

_No response_

_No response_

### Additional context

_No response_

peteralfonsi:
### Description
This PR adds an ExecutorService implementation which shares the logic of `ResizableExecutorBuilder`, but uses a virtual thread factory instead. If a feature flag is enabled, the search and index_searcher threadpools will use this implementation. When paired with io_uring, this lets us yield threads which are waiting for IO to complete, improving concurrency. 

The PR is still in a draft state. In particular, before merging I want to make more things dynamic (at least the threadpool max size count, and perhaps even whether we use virtual threads at all if this is possible), and most likely change the logic of the multiplier setting to instead set the default number of search and index_searcher threads to something like num_cpu + io_concurrency, and reduce the queue size from 1000 accordingly such that the sum of the max thread count and the queue size remains unchanged. 

We will also want to expose virtual thread scheduling stats from VirtualThreadSchedulerMXBean in some API, but this is currently in another branch (https://github.com/opensearch-project/OpenSearch/compare/main...peteralfonsi:OpenSearch:vthreads-api) as it may require some nasty reflection to get it to work and I want to see if we can do it more cleanly. 

I also have some test results (both functional and performance) which I will attach here before marking this as ready. 

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/18840, part of larger issue https://github.com/opensearch-project/OpenSearch/issues/18833

### Check List


reta:
@peteralfonsi I think we had many conversation regarding VTs (in OpenSearch and Lucene as well) in the past and general conclusion is that search threads use file I/O (reading segments) and JDK fileystem access is mostly blocking (JEP [1] also explicitly mentions that, VT are not helpful there). To your point, as far I know, OpenJDK does not support io_uring at the moment, so what are the benefits and where they are coming from? Thank you.

[1] https://openjdk.org/jeps/444

peteralfonsi:
@reta This is intended to go alongside an io_uring PR from @kaushalmahi12 (he's on vacation this week so he likely won't post it until next Monday, but his branch does have a functioning implementation). With both changes combined, we see pretty significant speedups of ~50% or so on various big5 OSB operations. 

I'm not sure about OpenJDK not supporting io_uring - maybe Kaushal or @abiesps could speak more to that?

abiesps:
Yes, this is correct @reta . In our past experiments we were able to validate that virtual threads alone doesn't provide a lot of benefits with regular file I/O in Java. Thats why we will use this along with iouring based File IO. 
We are developing IOuring based File I/O in a separate plugin in OpenSearch. Plugin will provide 'FileChannel' like apis which can be easily integrated into different flavours of IndexInputs that are used in lucene and OpenSearch.

msfroh:
This is really neat, but I'm on the same page as @reta about this going against advice from the Lucene community. Of course, that advice is based on Lucene's use of MMapDirectory or NIOFSDirectory. I'm assuming the new io_uring-based implementation must supply a different Directory implementation that returns a different flavor of IndexInput.

server/src/main/java/org/opensearch/threadpool/ThreadPool.java
```diff
@@ -173,17 +176,26 @@ public static ThreadPoolType fromType(String type) {
         }
     }
 
-    public static final Map<String, ThreadPoolType> THREAD_POOL_TYPES;
+    public static Map<String, ThreadPoolType> THREAD_POOL_TYPES;
```
andrross, line 179: We should find a way to solve the problem of deciding if virtual threads are allowed at runtime without making this a mutable static field.

andrross (commented):
> most likely change the logic of the multiplier setting to instead set the default number of search and index_searcher threads to something like num_cpu + io_concurrency

As I understand it, the general idea of virtual threads is that you don't want to set any fixed limit to how many can be created. They are very cheap to create. If they do minimal work and mostly block, then you generally don't want to set a cap for how many can be created. This is how `ExecutorService.newVirtualThreadPerTaskExecutor()` works.

PR #20449, opened 2026-01-21
Updating composite engine reference before translog recovery

raghuvanshraj:
### Description

### Related Issues

### Check List


server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5545,6 +5546,7 @@ public void close() throws IOException {
                 LocalCheckpointTracker::new,
                 TranslogEventListener.NOOP_TRANSLOG_EVENT_LISTENER
             );
+            oldCompositeEngine = currentCompositeEngineReference.getAndSet(newCompositeEngine);
```
ask-kamal-nayan, line 5549: Shall we also add it in try block so that even if some exception comes after setting the newCompositeEngine while translog Recovery and all, still we can close the old engine?

mgodwan (approved):

PR #20461, opened 2026-01-22
Fix CPU stall in geotile_grid aggregation (#20413)

### LINKED ISSUE #20413: [BUG] geotile_grid aggregation on LineString maxes out CPU and stalls cluster
### Describe the bug

When running a simple geotile_grid aggregation on a geoshape field that has been indexed with only a single document with a geojson LineString, we see that the query times out and the data node CPU jumps to a high value and stays that way indefinitely until the node is manually restarted.

### Related component

Search:Aggregations

### To Reproduce

Note 1: I have done this with an Amazon OpenSearch managed cluster but I have not tried to reproduce it locally.
Note 2: I originally saw the issue on a larger cluster with r7g.4xlarge data nodes even though the below steps mention a very small single-node cluster.
1. Create a new Amazon OpenSearch managed cluster with the following settings: 
    - Domain creation method - Standard Create
    - Templates - Dev/test
    - Availability Zones - 1-AZ without standby
    - Engine Version - 2.19
    - Number of Data Nodes - 1 r8g.large.search data node
2. Create an index with the following definition:
```py
opensearch.indices.create(
    index_name,
    {
        "settings": {
            "index.number_of_shards": 1,
            "index.number_of_replicas": 0
        },
        "mappings": {
            "properties": {
                "geolocation": {"type": "geo_shape"}
            }
        }
    },
)
```
3. Index a single document with the following data:
```py
document = {
    "geolocation": {
        "type": "LineString",
        "coordinates": [
            [120.69105000000002, -2.1092199999999366],
            [120.75767000000008, -2.159189999999967],
            [120.82192000000009, -2.1877399999999625],
        ]
    }
}
opensearch.index(index_name, document)
```
4. Run the following search query to verify the document was successfully indexed:
```py
query = {
    "size": 1,
    "timeout": "60s",
    "query": {
        "match_all": {}
    }
}

try:
    result = opensearch.search(index=index_name, body=query)
except RequestError as e:
    print(e)
    raise

print(json.dumps(result, indent=2))
```
5. Run the following query to reproduce the issue:
```py
query = {
    "size": 1,
    "timeout": "60s",
    "query": {
        "match_all": {}
    },
    "aggs": {
        "locations": {
            "geotile_grid": {
                "field": "geolocation",
                "precision": 29
            }
        }
    }
}

try:
    result = opensearch.search(index=index_name, body=query)
except RequestError as e:
    print(e)
    raise

print(json.dumps(result, indent=2))
```
6. Note that a TimeoutError occurs. Go to Amazon CloudWatch and plot the Data Node CPUUtilization metric. Note that the CPU has jumped to ~50% and stays there (this instance type has 2 vCPUs, so this is essentially one vCPU handling the single shard and getting maxed out).
7. monitor for a while and notice that the CPU never goes down. In addition running `POST /_tasks/_cancel` does not work. The only option we have found is to restart the node.

### Expected behavior

If this type of query is supported it should not bring down the cluster with only a single document. if it's not it should throw an error.

### Additional Details

**Plugins**
Default Amazon OpenSearch configuration as mentioned above.

 - OS: Linux - AWS Managed cluster

CodeItAlone:
PR Description
Issue Reference
Fixes #20413

Description
This PR resolves a significant performance issue where the geotile_grid aggregation causes a 100% CPU stall (unresponsive cluster) when processing complex geo_shape data (specifically LineString) at high precision levels (e.g., precision 29).

The root cause was identified as an unbounded computational loop in the aggregation logic. When a LineString crosses a large number of tiles at high zoom levels, the engine attempts to calculate and collect every single tile intersection on the main execution thread without any safety limits.

Changes
Implemented a Guard Clause: Added a safety threshold in GeoGridAggregator.java to check the valuesCount of a document before processing.

Early Exit Strategy: If a single document's tile count exceeds the defined limit of 10,000, the aggregator now skips that document and moves to the next, preventing the CPU from entering a long-running stall.

Circuit Breaker Logic: This approach acts as a localized circuit breaker to maintain cluster stability during "heavy" geographic queries.

Testing Performed
Unit Tests: Successfully ran modules:geo:test to ensure no regressions in geographic aggregation logic.

Targeted Testing: Verified the fix specifically using GeoTileGridAggregatorTests.

Manual Verification: Confirmed that queries with complex LineStrings that previously stalled the cluster now return successfully without high CPU overhead.

### DIFF
--- a/modules/geo/src/main/java/org/opensearch/geo/search/aggregations/bucket/geogrid/GeoGridAggregator.java
+++ b/modules/geo/src/main/java/org/opensearch/geo/search/aggregations/bucket/geogrid/GeoGridAggregator.java
@@ -31,6 +31,8 @@
 
 package org.opensearch.geo.search.aggregations.bucket.geogrid;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.SortedNumericDocValues;
 import org.apache.lucene.search.ScoreMode;
@@ -57,7 +59,9 @@ import java.util.Map;
  *
  * @opensearch.api
  */
+
 public abstract class GeoGridAggregator<T extends BaseGeoGrid> extends BucketsAggregator {
+    private static final Logger logger = LogManager.getLogger(GeoGridAggregator.class);
 
     protected final int requiredSize;
     protected final int shardSize;
@@ -94,13 +98,21 @@ public abstract class GeoGridAggregator<T extends BaseGeoGrid> extends BucketsAg
     public LeafBucketCollector getLeafCollector(LeafReaderContext ctx, final LeafBucketCollector sub) throws IOException {
         SortedNumericDocValues values = valuesSource.longValues(ctx);
         return new LeafBucketCollectorBase(sub, null) {
+            final int MAX_TILES_PER_DOCUMENT = 10000;
+
             @Override
             public void collect(int doc, long owningBucketOrd) throws IOException {
                 if (values.advanceExact(doc)) {
                     final int valuesCount = values.docValueCount();
 
+                    if (valuesCount > MAX_TILES_PER_DOCUMENT) {
+                        // Log a warning so the user knows why data is missing
+                        logger.warn("Skipping doc [{}] in aggregation [{}] due to excessive tiles: [{}]", doc, name, valuesCount);
+                        return;
+                    }
                     long previous = Long.MAX_VALUE;
                     for (int i = 0; i < valuesCount; ++i) {
+
                         final long val = values.nextValue();
                         if (previous != val || i == 0) {
                             long bucketOrdinal = bucketOrds.add(owningBucketOrd, val);


modules/geo/src/main/java/org/opensearch/geo/search/aggregations/bucket/geogrid/GeoGridAggregator.java
```diff
@@ -94,13 +98,21 @@ public ScoreMode scoreMode() {
...
+            final int MAX_TILES_PER_DOCUMENT = 10000;
+
             @Override
             public void collect(int doc, long owningBucketOrd) throws IOException {
                 if (values.advanceExact(doc)) {
                     final int valuesCount = values.docValueCount();
 
+                    if (valuesCount > MAX_TILES_PER_DOCUMENT) {
+                        // Log a warning so the user knows why data is missing
+                        logger.warn("Skipping doc [{}] in aggregation [{}] due to excessive tiles: [{}]", doc, name, valuesCount);
```
andrross, line 110: A user is unlikely to see the server-side logs. Instead they will see their query complete with no indication that they are getting incomplete/incorrect results.

andrross (commented):
The original issue (#20413) shows this issue can happen with a single document with a 3 coordinate line string. What about that case makes this so complex? (I'll note that I'm not an expert on the geo functionality)

The original issue also shows that the stuck query is not respecting the timeout/cancellation. That definitely appears to be a bug. If these queries can be long running, then they need to be cancel-able. Have you looked into fixing that?

andrross:
Closing due to lack of response. @CodeItAlone please reopen if you'd like to continue working on this.

PR #20471, opened 2026-01-23
Refactor streaming agg query phase planning

bowenlan-amzn:
### Description

Streaming aggregation feature supports an agg defined with
- Top level should be string terms agg and numeric terms agg
- Second level/sub aggregation could be numeric terms, cardinality, max, min, sum (we should add avg and value_count soon)
If multiple aggs fit above constraints, it should also be good.

Segment streaming topN heuristic set to `shard_size`
Also user has a index level setting to provide the segment topN size. 
The final topN is the larger one of these 2

The topN is compared with `search.aggregations.streaming.max_estimated_bucket_count` to decide whether streaming should be used or not. If exceed, streaming is considered to stream back too many buckets per segment so we will fall back to default aggregation path which send response per shard.

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/search/streaming/FlushModeResolver.java
+++ b/server/src/main/java/org/opensearch/search/streaming/FlushModeResolver.java
@@ -13,11 +13,10 @@ import org.opensearch.common.settings.Setting;
 import org.opensearch.search.aggregations.AggregationBuilder;
 import org.opensearch.search.aggregations.AggregatorFactories;
 import org.opensearch.search.aggregations.bucket.terms.TermsAggregationBuilder;
-import org.opensearch.search.aggregations.metrics.AvgAggregationBuilder;
 import org.opensearch.search.aggregations.metrics.CardinalityAggregationBuilder;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/opensearch/search/streaming/FlushModeResolver.java
```diff
@@ -79,120 +88,83 @@ public final class FlushModeResolver {
...
-
-        return FlushMode.PER_SEGMENT;
+        return isMetricAggregation(agg);
     }
 
+    private static boolean isMetricAggregation(AggregationBuilder agg) {
+        return agg instanceof CardinalityAggregationBuilder
+            || agg instanceof MaxAggregationBuilder
+            || agg instanceof MinAggregationBuilder
+            || agg instanceof AvgAggregationBuilder
```
rishabhmaurya: do we support avg? 
also, don't we support sum now?

bowenlan-amzn: updated

Issue #20485, opened 2026-01-27
[BUG] Workload group(Query Group) creation fails when request hits data node due to updated_at validation against CM clock


hyanggeun:
### Describe the bug

**Summary**
Creating a WLM workload group (PUT /_wlm/workload_group) intermittently fails with WorkloadGroup.updatedAtInMillis is not a valid epoch when the REST request is served by a data node. The failure rate increases when traffic is routed to data nodes, but disappears when sent directly to the cluster-manager node.

**Environment**
* OpenSearch: 3.4.0
* Feature: Workload Management (WLM) workload groups
* TLS enabled
* Cluster with separate data and cluster-manager nodes
* NTP appears healthy (sub‑50ms jitter, no obvious clock skew in node stats)

**Current Behavior** 
Request sometimes returns:

```json
{
  "error": {
    "root_cause": [
      { "type": "illegal_argument_exception", "reason": "WorkloadGroup.updatedAtInMillis is not a valid epoch" }
    ],
    "type": "illegal_argument_exception",
    "reason": "WorkloadGroup.updatedAtInMillis is not a valid epoch"
  },
  "status": 400
}
```

It happens more often when the REST request is served by a data node; sending the same request directly to the cluster-manager node avoids the error.

### Related component

Plugins

### To Reproduce

1. check node timestamps

```
GET _nodes/stats?filter_path=nodes.*.name,nodes.*.timestamp
 "nodes": {
    "RIYgIe82SJKB6_v0tFhrHw":{
       "timestamp":1769418930198,
       "name":"cm-0"
    },
    "uRuwSYQYQOCVumlqnvp9AQ":{
       "timestamp":1769418930193,
       "name":"data-0"
    },
    "lbuwkeheQg6e8CYkUrmhBQ":{
       "timestamp":1769418930202,
       "name":"cm-1"
    },
    "QnvaRRBlRv6kg2DGoXKmiw":{
       "timestamp":1769418930215,
       "name":"cm-2"
    },
    "88Hk51bYRailzRngI33wmg":{
       "timestamp":1769418930211,
       "name":"data-1"
}
```

2. call request directly in `data-1` node(1769418930211) -> `cm-0` node (1769418930198)

3. api call failed in data-1 node
```
curl -XPUT https://localhost:9200/_wlm/workload_group
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
[...]

PR #20486, opened 2026-01-27
Bugfix: Relax updatedAt validation for WLM workload group creation

hyanggeun:
#### Summary
This change fixes intermittent failures when creating WLM workload groups on data nodes. The current validation rejects updatedAtInMillis if it is slightly ahead of the cluster-manager time, which can happen with minor clock skew. I believe a strict updatedAt upper-bound check is unnecessary for creation, so this patch only verifies that the value is a valid epoch (>= 0).
#### Changes
Remove the “not in the future” check from WorkloadGroup.isValid(...)
Keep only the lower-bound epoch validation
#### Why
Requests routed to data nodes can fail with WorkloadGroup.updatedAtInMillis is not a valid epoch due to small time differences across nodes. Since updatedAt is not critical for creation integrity, strict validation doesn’t seem necessary.

### Related Issues
Fixes https://github.com/opensearch-project/OpenSearch/issues/20485

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/cluster/metadata/WorkloadGroup.java
+++ b/server/src/main/java/org/opensearch/cluster/metadata/WorkloadGroup.java
@@ -79,6 +79,16 @@ public class WorkloadGroup extends AbstractDiffable<WorkloadGroup> implements To
         this.updatedAtInMillis = updatedAt;
     }
 
+    public static boolean isValid(long updatedAt) {
+        long minValidTimestamp = Instant.ofEpochMilli(0L).getMillis();
+
+        // Use Instant.now() to get the current time in seconds since epoch
+        long currentSeconds = Instant.now().getMillis();
+
+        // Check if the timestamp is within a reasonable range
+        return minValidTimestamp <= updatedAt && updatedAt <= currentSeconds;
+    }
+
     public WorkloadGroup(StreamInput in) throws IOException {
         this(in.readString(), in.readString(), new MutableWorkloadGroupFragment(in), in.readLong());
     }
--- a/modules/autotagging-commons/common/src/main/java/org/opensearch/rule/autotagging/RuleValidator.java
+++ b/modules/autotagging-commons/common/src/main/java/org/opensearch/rule/autotagging/RuleValidator.java
@@ -17,6 +17,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 
+import static org.opensearch.cluster.metadata.WorkloadGroup.isValid;
+
 /**
  * This is the validator for rule. It ensures that the rule has a valid description, feature value,
  * update time, attribute map, and the rule adheres to the feature type's constraints.
@@ -121,7 +123,7 @@ public class RuleValidator {
     }
 
     private List<String> validateUpdatedAtEpoch() {
-        if (updatedAt != null && Instant.parse(updatedAt).getMillis() <= 0L) {
+        if (updatedAt != null && !isValid(Instant.parse(updatedAt).getMillis())) {
             return List.of("Rule update time is not a valid epoch");
         }
         return new ArrayList<>();


kaushalmahi12:
I think it should be fine to remove this check since it will be very rare for user to supply this value since it is generated on the node which recieves the create/update request.

cwperks:
@kaushalmahi12 I don't want to block this PR or anything, but I think we need a more standard way of tracking changes to config indices like the security index, ISM, alerting and WLM. There is technically a solution with the security auditlog, but we've been discussing with @nagarajg17 something separate for the audit log that would be geared towards operators and cluster admins who specifically want to keep track of config-type changes. 

Essentially, a more abstracted feature than https://github.com/opensearch-project/security/issues/5407 and https://github.com/opensearch-project/security/issues/5093#issuecomment-2745345181 that any plugin can leverage out of the box.

jainankitk:
> it would be good to also add UT around it.

@hyanggeun - While you are at it, can you also add changelog entry and resolve precommit failures?

hyanggeun:
@jainankitk @kaushalmahi12 
Thanks for the feedback. I'll add the necessary unit tests, update the changelog entry, and resolve the pre-commit issues.

server/src/main/java/org/opensearch/cluster/metadata/WorkloadGroup.java
```diff
@@ -82,11 +82,8 @@ public WorkloadGroup(String name, String _id, MutableWorkloadGroupFragment mutab
     public static boolean isValid(long updatedAt) {
         long minValidTimestamp = Instant.ofEpochMilli(0L).getMillis();
 
-        // Use Instant.now() to get the current time in seconds since epoch
-        long currentSeconds = Instant.now().getMillis();
-
         // Check if the timestamp is within a reasonable range
-        return minValidTimestamp <= updatedAt && updatedAt <= currentSeconds;
+        return minValidTimestamp <= updatedAt;
```
dzane17: nit: 
```
return updatedAt >= 0;
```

jainankitk: I guess we can rid of this method `isValid` and add the check `updatedAt >= 0` inline

jainankitk:
Unrelated failure, retried gradle check:

```
[Test Result](https://build.ci.opensearch.org/job/gradle-check/71021/testReport/) (1 failure / -12)

    [org.opensearch.example.systemingestprocessor.ExampleSystemIngestProcessorClientYamlTestSuiteIT.test {yaml=example-system-ingest-processor/20_system_ingest_processor/Processor injects field when index is created from matching template where trigger_setting is true}](https://build.ci.opensearch.org/job/gradle-check/71021/testReport/junit/org.opensearch.example.systemingestprocessor/ExampleSystemIngestProcessorClientYamlTestSuiteIT/test__yaml_example_system_ingest_processor_20_system_ingest_processor_Processor_injects_field_when_index_is_created_from_matching_template_where_trigger_setting_is_true_/)
```

server/src/main/java/org/opensearch/cluster/metadata/WorkloadGroup.java
```diff
@@ -79,16 +79,6 @@ public WorkloadGroup(String name, String _id, MutableWorkloadGroupFragment mutab
         this.updatedAtInMillis = updatedAt;
     }
 
-    public static boolean isValid(long updatedAt) {
```
andrross, line 82: This class is annotated with `@PublicApi` so removing a publicly visible static method is a breaking change that is not allowed.

hyanggeun: @andrross 
Thanks for the review. I restored the isValid method to its original behavior, and kept the constructor validation as intended with the updatedAt <= 0 check.

PR #20493, opened 2026-01-27
Add hook to provide GrpcInterceptorProvider Settings

finnegancarroll:
### Description
Plugins implementing a GrpcInterceptorProvider may want their interceptor behavior to depend on settings.
This change adds an `initSettings` hook called on `GrpcPlugin.createComponents` which provides these interceptors visibility into settings. 

### Related Issues
N/A

### Check List

### DIFF
--- a/modules/transport-grpc/src/test/java/org/opensearch/transport/grpc/GrpcPluginTests.java
+++ b/modules/transport-grpc/src/test/java/org/opensearch/transport/grpc/GrpcPluginTests.java
@@ -1050,7 +1050,7 @@ public class GrpcPluginTests extends OpenSearchTestCase {
 
     public void testGrpcInterceptorProviderEmpty() {
         GrpcInterceptorProvider prov = threadContext -> List.of();
-        assertDoesNotThrow(() ->prov.initNodeSettings(Settings.EMPTY));
+        assertDoesNotThrow(() -> prov.initNodeSettings(Settings.EMPTY));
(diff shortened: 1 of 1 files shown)


modules/transport-grpc/spi/src/main/java/org/opensearch/transport/grpc/spi/GrpcInterceptorProvider.java
```diff
@@ -20,6 +21,15 @@
  */
 public interface GrpcInterceptorProvider {
 
+    /**
+     * Provide visibility into node settings.
+     * @param settings for use in interceptors.
+     * @return chaining.
+     */
+    default GrpcInterceptorProvider initSettings(Settings settings) {
```
karenyrx: Is there a use case of having to `return this;` for chaining? 
(Intuitively, I feel like init*() methods usually return `void`)

finnegancarroll: edit: moved this to return void

prudhvigodithi:
At high level LGTM, @finnegancarroll can you fix the codecov ?

finnegancarroll:
Thanks for taking a look @prudhvigodithi , adding a small unit test to bolster code cov here. This line seems to functionally used all over the tests but with objects that are provided by Mockito which confuses code cov.

PR #20494, opened 2026-01-28
fix: Separate internal ignore settings from user ignore settings in restore

piyushk6130:
Description

This PR fixes an issue where internal ignore settings were being blocked by user-unremovable settings protection during snapshot restore operations.

Problem

When restoring snapshots across clusters (e.g., from a remote-store enabled cluster to a non-remote-store cluster), internal settings like index.remote_store.* need to be filtered out. However, the current implementation combines internal ignore settings with user ignore settings, causing internal settings to be blocked by the USER_UNREMOVABLE_SETTINGS protection check.

Solution

Separate internal ignore settings into their own filter lists (internalKeyFilters and internalSimpleMatchPatterns) that are evaluated first and bypass the user-unremovable settings protection. This ensures:

Internal ignore settings (like index.remote_store.*) are always filtered regardless of protection status
User ignore settings continue to respect the USER_UNREMOVABLE_SETTINGS protection
Testing

Manually tested cross-cluster restore scenarios:

Remote store → Non-remote store: Remote store settings correctly stripped ✓
Non-remote store → Remote store: Settings correctly applied by target cluster ✓
Remote store → Remote store: Settings preserved ✓
Non-remote store → Non-remote store: No changes ✓

### DIFF
--- a/server/src/main/java/org/opensearch/snapshots/RestoreService.java
+++ b/server/src/main/java/org/opensearch/snapshots/RestoreService.java
@@ -820,6 +820,9 @@ public class RestoreService implements ClusterStateApplier {
                         Settings settings = indexMetadata.getSettings();
                         Set<String> keyFilters = new HashSet<>();
                         List<String> simpleMatchPatterns = new ArrayList<>();
+                        Set<String> internalKeyFilters = new HashSet<>();
+                        List<String> internalSimpleMatchPatterns = new ArrayList<>();
+
                         for (String ignoredSetting : ignoreSettings) {
                             if (!Regex.isSimpleMatchPattern(ignoredSetting)) {
                                 if (USER_UNREMOVABLE_SETTINGS.contains(ignoredSetting)) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/opensearch/snapshots/RestoreService.java
```diff
@@ -840,16 +843,29 @@ private IndexMetadata updateIndexSettings(
...
+                                internalKeyFilters.add(ignoredSetting);
                             } else {
-                                simpleMatchPatterns.add(ignoredSetting);
+                                internalSimpleMatchPatterns.add(ignoredSetting);
                             }
                         }
 
                         Predicate<String> settingsFilter = k -> {
+                            // Check internal ignore patterns first (they override protection)
+                            for (String filterKey : internalKeyFilters) {
```
linuxpi: can we replace the for loop with a exists check in the `Set`

server/src/main/java/org/opensearch/snapshots/RestoreService.java
```diff
@@ -840,16 +843,29 @@ private IndexMetadata updateIndexSettings(
...
+                                if (k.equals(filterKey)) {
+                                    return false;
+                                }
+                            }
```
linuxpi: we need to cover this change in some UT or IT

ajaymovva: +1

server/src/main/java/org/opensearch/snapshots/RestoreService.java
piyushk6130: done

server/src/main/java/org/opensearch/snapshots/RestoreService.java
piyushk6130: added UT.

server/src/test/java/org/opensearch/snapshots/RestoreServiceTests.java
```diff
@@ -159,4 +159,36 @@ public void testValidateReplicationTypeRestoreSettings_WhenSnapshotIsSegment_Res
             () -> RestoreService.validateReplicationTypeRestoreSettings(snapshot, ReplicationType.SEGMENT.toString(), indexMetadata)
         );
     }
+
+    // Tests for internal vs user ignore settings filter separation (PR #20494)
+
```
linuxpi, line 164: we also need to cover `updateIndexSettings` in our test. Also we should write a full fledged IT

linuxpi (approved):

PR #20481, opened 2026-01-26
TopN selection for streaming terms aggregations

rishabhmaurya:
### Description
**Problem**
Streaming terms aggregators flush results per-segment, sending all buckets to the coordinator. This causes:

1. Performance degradation: Excessive data transfer and coordinator memory usage when dealing with high-cardinality fields

2. Memory pressure: Coordinator must process and reduce all buckets from all segments

**Solution**
Implemented TopN selection at segment level using quickselect algorithm, handling subAggs an supported metric Aggs, to send only the most relevant buckets to the coordinator. 

**Benefits:**

Better Performance: 
Reduces data transfer and coordinator memory/CPU usage by sending only TOP N buckets per segment

Controlled Accuracy:
TopN selection introduces potential accuracy trade-offs, which can be controlled using:

`size` parameter: Number of final buckets to return

`shard_size` parameter: Number of buckets to collect per segment (not per shard in streaming aggregations)

New `index.aggregation.streaming.min_shard_size` setting (default: 1000): Minimum buckets per segment to ensure accuracy. Ideally, we want to set its default as a function of shard_size and number of segments, @bowenlan-amzn is currently analyzing this part. 

**Testing**
A. SubAggregationIT: 
1. Validates TopN selection with max/cardinality sub-aggregation ordering (ASC/DESC)
2. Verifies correct bucket selection and `otherDocCount` calculation.

B. Unit Tests:
`StreamStringTermsAggregatorTests:` tests for TopN selection with various sub-aggregation orderings + minDocCount filtering

`StreamNumericTermsAggregatorTests:` 13 tests covering LongTerms, DoubleTerms, UnsignedLongTerms with max/min/avg/sum/cardinality ordering + minDocCount. Verifies correct bucket selection and `otherDocCount` calculation.

I will be opening a public doc issue for entire streaming aggregations.

**Note: these changes are built on top of open PR https://github.com/opensearch-project/OpenSearch/pull/20359, so contains extra changes.** 

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [x] Functionality includes testing.
~~- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.~~
~~- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.~~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
[...]

PR #20510, opened 2026-01-30
Streaming agg planning fallback enhancement

bowenlan-amzn:
### Description

Streaming agg planning fallback when

- low cardinality string term
- match all + not match segment with deleted docs

query in the search context seems won't be null and if so it will be parsed to match all
https://github.com/opensearch-project/OpenSearch/blob/2d7e8d9b5bdaa004de24dd0262de9e82599454b2/server/src/main/java/org/opensearch/search/DefaultSearchContext.java#L448

### Related Issues

### Check List

### DIFF
--- a/server/src/test/java/org/opensearch/search/aggregations/FactoryStreamingCostEstimationTests.java
+++ b/server/src/test/java/org/opensearch/search/aggregations/FactoryStreamingCostEstimationTests.java
@@ -15,8 +15,13 @@ import org.apache.lucene.index.DirectoryReader;
 import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.BooleanClause;
+import org.apache.lucene.search.BooleanQuery;
 import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.MatchAllDocsQuery;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.util.BytesRef;
 import org.opensearch.core.common.breaker.CircuitBreaker;
@@ -652,9 +657,7 @@ public class FactoryStreamingCostEstimationTests extends AggregatorTestCase {
                     TermsAggregationBuilder termsBuilder = new TermsAggregationBuilder("terms").field("category").size(10);
 
                     // Non-match-all query (using a term query) - should be streamable even with clean segments
-                    org.apache.lucene.search.Query termQuery = new org.apache.lucene.search.TermQuery(
-                        new org.apache.lucene.index.Term("category", "cat_100")
-                    );
+                    Query termQuery = new TermQuery(new Term("category", "cat_100"));
                     FactoryAndContext result = createAggregatorFactoryWithQuery(termsBuilder, searcher, termQuery, fieldType);
                     StreamingCostMetrics metrics = ((StreamingCostEstimable) result.factory).estimateStreamingCost(result.searchContext);
 
@@ -687,10 +690,7 @@ public class FactoryStreamingCostEstimationTests extends AggregatorTestCase {
     ) throws IOException {
         // Use a BooleanQuery wrapper to avoid being detected as match-all
         // This simulates a filtered query scenario
-        org.apache.lucene.search.Query nonMatchAllQuery = new org.apache.lucene.search.BooleanQuery.Builder().add(
-            new MatchAllDocsQuery(),
-            org.apache.lucene.search.BooleanClause.Occur.MUST
-        ).build();
+        Query nonMatchAllQuery = new BooleanQuery.Builder().add(new MatchAllDocsQuery(), BooleanClause.Occur.MUST).build();
         SearchContext searchContext = createSearchContext(
             searcher,
             createIndexSettings(),
@@ -709,7 +709,7 @@ public class FactoryStreamingCostEstimationTests extends AggregatorTestCase {
     private FactoryAndContext createAggregatorFactoryWithQuery(
         AggregationBuilder aggregationBuilder,
         IndexSearcher searcher,
-        org.apache.lucene.search.Query query,
+        Query query,
         MappedFieldType... fieldTypes
     ) throws IOException {
         SearchContext searchContext = createSearchContext(searcher, createIndexSettings(), query, createBucketConsumer(), fieldTypes);


server/src/test/java/org/opensearch/search/aggregations/FactoryStreamingCostEstimationTests.java
```diff
@@ -533,13 +678,58 @@ private MultiBucketConsumerService.MultiBucketConsumer createBucketConsumer() {
...
+     * Uses a non-match-all query to avoid triggering the match-all optimization fallback.
      */
     private FactoryAndContext createAggregatorFactory(
         AggregationBuilder aggregationBuilder,
         IndexSearcher searcher,
         MappedFieldType... fieldTypes
+    ) throws IOException {
+        // Use a BooleanQuery wrapper to avoid being detected as match-all
+        // This simulates a filtered query scenario
+        org.apache.lucene.search.Query nonMatchAllQuery = new org.apache.lucene.search.BooleanQuery.Builder().add(
```
rishabhmaurya: nit: use import instead of full package name

rishabhmaurya:
when explicit `execution_hint` is passed as map for term aggs, we can skip streaming agg.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
```diff
@@ -709,13 +714,84 @@ public StreamingCostMetrics estimateStreamingCost(SearchContext searchContext) {
...
+        for (LeafReaderContext leaf : leaves) {
+            try {
+                SortedSetDocValues docValues = valuesSource.ordinalsValues(leaf);
+                if (docValues != null) {
+                    maxCardinality = Math.max(maxCardinality, docValues.getValueCount());
+                }
+            } catch (IOException e) {
+                // If we can't read doc values, skip this check
+                return false;
+            }
```
Bukhtawar, line 763: Is there value in caching this for subsequent queries?

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
```diff
@@ -709,13 +714,84 @@ public StreamingCostMetrics estimateStreamingCost(SearchContext searchContext) {
             return StreamingCostMetrics.nonStreamable();
         }
 
-        if (valuesSource instanceof WithOrdinals || valuesSource instanceof ValuesSource.Numeric) {
+        if (valuesSource instanceof WithOrdinals ordinalsValuesSource) {
+            if (shouldDisableStreamingForOrdinals(searchContext, ordinalsValuesSource)) {
+                return StreamingCostMetrics.nonStreamable();
+            }
+            return new StreamingCostMetrics(true, segmentTopN);
```
atris, line 721: How would this work for high-cardinality global ordinals? AFAIU It unconditionally approves streaming even if the underlying field has millions of unique terms. Combined with `StreamStringTermsAggregator`
's dense allocation strategy, this can potentially lead to OOM/GC-thrashing vector. We must restrict this to low-cardinality overrides or ensure the consumer uses sparse structures.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
rishabhmaurya: Index stats (and probably query stats sometime in future 🤞🏽) would be a goldmine for such planning tasks and caching makes a total sense as it doesn't have to real time. 
Me and @bowenlan-amzn had a discussion on caching it specially when we were thinking about finding distribution of point values, using BKDs, for its use in query planning and caching them as it could be a bit expensive. 
We will take it as an enhancement for future versions as streaming aggregation becomes more mainstream in future.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
rishabhmaurya: @atris we should be good with controlling memory overhead for high cardinalities aggregation on both data nodes and coordinator with https://github.com/opensearch-project/OpenSearch/pull/20481.
Just like `shard_size` setting in traditional aggs, which is set per request, https://github.com/opensearch-project/OpenSearch/pull/20481 introduces `index.aggregation.streaming.min_segment_size` to control #buckets per segment, with default as 1000. If someone overrides to a very high value, to improve on accuracy,  and query planner sees max total buckets could exceed `search.aggregations.streaming.max_estimated_bucket_count`, then it will fallback to traditional aggregation. And on top of it all, we have max bucket limit of 65K at coordinator and if it breaches, request will fail.  I hope that addresses your concern.

atris: Sounds good.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
bowenlan-amzn: `getValueCount` at shard level should be pretty lightweight compared to the aggregation execution that reads all the matching doc values, so doesn't seem need to cache.

But it's valuable to centralize such stats to do query planning on coordinator node in the future like Risahbh mentioned.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
bowenlan-amzn: Just to add on Rishabh's comment here. The OOM/GC protection we have here is mainly to protect the coordinator. We limit the size of the partial aggregation buckets that are going to be sent to coordinator. And these buckets are built after collection phase.
During the collection, when terms aggregation retrieves doc value and bucketizes that, the same logic is used to protect memory.

https://github.com/opensearch-project/OpenSearch/blob/48b08fb794071a7e951fe4a3af18c37494d3bf68/server/src/main/java/org/opensearch/search/aggregations/bucket/BucketsAggregator.java#L121-L130

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregatorFactory.java
rishabhmaurya: @Bukhtawar feel free to share your further opinions here and we can take it as a follow up. For now resolving this comment.

atris:
I see it is already merged. Nevertheless, approved from my side.

PR #20514, opened 2026-01-31
Fix listBlobsByPrefixInSortedOrder in EncryptedBlobContainer to adhere limit

### LINKED ISSUE #20543: [BUG] EncryptedBlobContainer ignores limit for listing blobs in sorted order by default and performs sorting and trimming after loading all blobs from store
### Describe the bug

**Context**

In Remote Cluster State enabled clusters, we restore the latest remote cluster state on cluster-manager nodes before we start joining an existing cluster or election. To restore the latest cluster state, we get the latest manifest present in remote by getting the first item in Lexicographic order. BlobContainer implementations such as S3 already support fetching the latest item in Lexicographic without the need to list all objects and sort them internally.

**Problem**

When Remote Cluster state repositories are setup with S3 as well as encryptions, we wrap the S3BlobContainer under a EncryptedBlobContainer class. The EncryptedBlobContainer class is missing an implementation of the `listBlobsByPrefixInSortedOrder(blobNamePrefix, limit, blobNameSortOrder)` method, causing it to fall back to the default implementation in BlobContainer (`listBlobsByPrefix(blobNamePrefix)`) which loads all blobs matching the prefix into memory, sorts them and returns the required number of items which were passed as limit. 

On bootstrap of cluster managers with Remote Cluster State being enabled, we use the referenced method to get the latest manifest. In scenarios were the manifest files are grown to a large extent, the restore would take time proportional to the number of manifests and also cause memory exhaustion as we were fetching all the manifests in remote.

### Related component

Cluster Manager

### To Reproduce

1. Disable cleanup of manifests and allow the manifests to grow to large extent
2. Bootstrap a new cluster-manager for the domain and let it restore from S3

The restoration would take a large amount of time and inflate the JVM in the process as it would load all the files in memory instead of using S3BlobContainers's implementation of the `listBlobsByPrefixInSortedOrder` method which would just return the latest blob.

Reference wall profiling - 
Image

### Expected behavior

As S3BlobContainer already provides an implementation to directly get the latest blob after sorting in Lexicographic order, we do not need to load all the files to memory, sort and return the latest. This will help in not only improving bootstrapping time in cases when there is a manifest pile-up but also prevent JVM exhaustion

### Additional Details

**Plugins**

**Screenshots**

**Additional context**

Pranshu-S:
### Description
As S3BlobContainer already provides an implementation to directly get the latest blob after sorting in Lexicographic order, we do not need to load all the files to memory, sort and return the latest. This will help in not only improving bootstrapping time in cases when there is a manifest pile-up but also prevent JVM exhaustion

Before the change (~500k files in S3)
image

After the change
image

### Related Issues
Resolves #20543 20543

### Check List


server/src/main/java/org/opensearch/common/blobstore/EncryptedBlobContainer.java
```diff
@@ -247,4 +247,23 @@ public void listBlobsByPrefixInSortedOrder(
         );
         blobContainer.listBlobsByPrefixInSortedOrder(blobNamePrefix, limit, blobNameSortOrder, encryptedMetadataListener);
     }
+
+    @Override
+    public List<BlobMetadata> listBlobsByPrefixInSortedOrder(String blobNamePrefix, int limit, BlobNameSortOrder blobNameSortOrder)
+        throws IOException {
+        List<BlobMetadata> blobsList = blobContainer.listBlobsByPrefixInSortedOrder(blobNamePrefix, limit, blobNameSortOrder);
```
rajiv-kv, line 254: Looks like the default implementation of `listBlobsByPrefixInSortedOrder` is the unoptimized one for large lists. It might be good to add WARN log if the retrieved result (for in-memory sorting) is greater than say 5K

server/src/main/java/org/opensearch/common/blobstore/EncryptedBlobContainer.java
```diff
@@ -247,4 +247,23 @@ public void listBlobsByPrefixInSortedOrder(
...
+        if (blobsList == null) {
```
rajiv-kv, line 255: is it expected to return null ?

server/src/test/java/org/opensearch/gateway/remote/RemoteManifestManagerTests.java
```diff
@@ -116,6 +119,34 @@ public void testReadLatestMetadataManifestFailedIOException() throws IOException
...
+        EncryptedBlobContainer<Object, Object> encryptedBlobContainer = new EncryptedBlobContainer<>(mockBlobContainer, cryptoHandler);
+
+        final BlobPath blobPath = mock(BlobPath.class);
+        when((blobStoreRepository.basePath())).thenReturn(blobPath);
+        when(blobPath.add(anyString())).thenReturn(blobPath);
+        when(blobPath.buildAsString()).thenReturn("/blob/path/");
+        when(encryptedBlobContainer.path()).thenReturn(blobPath);
+        when(blobStore.blobContainer(any())).thenReturn(encryptedBlobContainer);
+
+        when(mockBlobContainer.listBlobsByPrefixInSortedOrder("manifest" + DELIMITER, 1, BlobContainer.BlobNameSortOrder.LEXICOGRAPHIC))
```
rajiv-kv, line 139: Is it possible to verify if the limit parameter is passed to s3 request (when the enclosing container is s3 based) ? How do we verify that the default implementation is not being used ?

server/src/main/java/org/opensearch/common/blobstore/EncryptedBlobContainer.java
Pranshu-S: It depends on the per blobContainer implementation. It ideally shouldn't but there is no parent validation to not return a null.

server/src/test/java/org/opensearch/gateway/remote/RemoteManifestManagerTests.java
Pranshu-S: The tests in core-package do not work specifically with a plugins like S3 so it will not be possible to directly setup a validation here. I can look around in S3RepositoryPlugin if this can be validated there.

Pranshu-S:
Result - 
[Test Result](https://build.ci.opensearch.org/job/gradle-check/70871/testReport/) (2 failures / +2)
[org.opensearch.index.engine.CompositeIndexWriterForAppendTests.testConcurrentIndexAndDeleteDuringRefresh](https://build.ci.opensearch.org/job/gradle-check/70871/testReport/junit/org.opensearch.index.engine/CompositeIndexWriterForAppendTests/testConcurrentIndexAndDeleteDuringRefresh/)


Flaky - https://github.com/opensearch-project/OpenSearch/issues/20005

server/src/main/java/org/opensearch/common/blobstore/EncryptedBlobContainer.java
shwetathareja: where is bug actually in the listener impl?

server/src/test/java/org/opensearch/gateway/remote/RemoteManifestManagerTests.java
```diff
@@ -116,6 +119,34 @@ public void testReadLatestMetadataManifestFailedIOException() throws IOException
...
+        BlobContainer mockBlobContainer = mock(BlobContainer.class);
+        CryptoHandler<Object, Object> cryptoHandler = mock(CryptoHandler.class);
+        EncryptedBlobContainer<Object, Object> encryptedBlobContainer = new EncryptedBlobContainer<>(mockBlobContainer, cryptoHandler);
+
+        final BlobPath blobPath = mock(BlobPath.class);
+        when((blobStoreRepository.basePath())).thenReturn(blobPath);
+        when(blobPath.add(anyString())).thenReturn(blobPath);
+        when(blobPath.buildAsString()).thenReturn("/blob/path/");
+        when(encryptedBlobContainer.path()).thenReturn(blobPath);
+        when(blobStore.blobContainer(any())).thenReturn(encryptedBlobContainer);
```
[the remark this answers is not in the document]
shwetathareja: you can mock encryptedBlobContainer as well

Pranshu-S:
Appears to be stuck
```
time passed: 4680
```

PR #20515, opened 2026-01-31
Implementing batched deletions of stale ClusterMetadataManifests in RemoteClusterStateCleanupManager and adding configurable deletions timeouts

### LINKED ISSUE #20564: [BUG] Remote Cluster State cleanup failures due to deletions timeouts resulting in stale-metadata pile-ups
### Describe the bug

**Context**

The RemoteClusterStateCleanupManager runs a scheduled cleanup task every 5 minutes (configurable) to remove stale cluster metadata files from S3. The cleanup process follows a three-phase approach:

1. Listing all manifest files
2. Identifying which files are stale by comparing active vs outdated manifests
3. Deleting the stale files in sequential batches.

The deletions are in a sync path - that is we attempt to first delete all the stale global metadata files, wait for its completion and similarly delete IndexMetadata and finally Manfiest files

**Problem**

A change was done to [Add timeout handling for S3 blob container async operations](https://github.com/opensearch-project/OpenSearch/pull/18583/changes) in the sync operation used for s3 deletions where-in the caller would throw a IOException when the deletions would not complete within 30 seconds. Due to this change, if there are large number of entities to delete (such as global metadata or IndexMetadata), we end up timing out in the wait.

The deletion phase in RemoteClusterStateCleanupManager executes four sequential deleteStalePaths() calls for different file types (global metadata, index metadata, ephemeral attributes, and manifests), all wrapped in a single try-catch block. If we perform large number of operations in between a cleanup run, the large number of files could lead to deletions to time-out for say. For example, if the global-metadata deletions time out, an IOException is thrown which causes the entire cleanup process to abort immediately. This means that subsequent deletion calls are never executed, leaving stale index metadata, ephemeral files, and manifest files undeleted in remote storage. The next attempt it picked up after another 300s (5 minutes) which further increases the number of files to delete.

### Related component

Cluster Manager

### To Reproduce

1. Setup a remote publication enabled cluster
2. Run a large throughput of operations such as create, delete, update on entities stored in cluster state.
3. The deletions would start timing out and stale files would keep on piling

```
[o.o.g.r.RemoteClusterStateCleanupManager] Error while deleting stale Remote Cluster Metadata files
java.io.IOException: Delete operation timed out after 30 seconds
        at org.opensearch.repositories.s3.S3BlobContainer.getFutureValue(S3BlobContainer.java:425)
        at org.opensearch.repositories.s3.S3BlobContainer.deleteBlobsIgnoringIfNotExists(S3BlobContainer.java:409)
        at org.opensearch.common.blobstore.EncryptedBlobContainer.deleteBlobsIgnoringIfNotExists(EncryptedBlobContainer.java:152)
        at org.opensearch.index.translog.transfer.BlobStoreTransferService.deleteBlobs(BlobStoreTransferService.java:270)
        at org.opensearch.gateway.remote.RemoteClusterStateCleanupManager.deleteStalePaths(RemoteClusterStateCleanupManager.java:447)
        at org.opensearch.gateway.remote.RemoteClusterStateCleanupManager.deleteClusterMetadata(RemoteClusterStateCleanupManager.java:323)
```

### Expected behavior

We should either default to older behaviour where we have the S3 deletions in a sync manner or delete the entities in batches.

### Additional Details

**Plugins**
S3 Repository Plugin

**Screenshots**

**Additional context**

Pranshu-S:
### Description

Fixes remote cluster state cleanup failures that were causing **stale metadata pile-ups** in remote storage when deletions time out along with fixes mention in tagged GitHub Issue

**Issue context (#20564):** `RemoteClusterStateCleanupManager` runs every ~5 minutes (configurable) and performs *sequential* deletions (global metadata → index metadata → ephemeral attrs → manifests). A recent change added a **30s timeout** to the S3 “sync” delete path; with large delete sets this can throw `IOException`, **abort the whole cleanup run** (single try/catch), and leave later phases undeleted—making the next run even larger and more likely to fail. 

### What this PR does

* Batch deletes stale `manifest*` blobs to reduce per-call delete size and avoid timeout/payload issues. 
  * Controlled by:
    * `cluster.remote_store.state.cleanup.batch_size`
    * `cluster.remote_store.state.cleanup.max_batches`
* Makes delete timeout configurable for remote-state cleanup
* Fixing issues where the update-interval call was not being honoured when the deletion was disabled (set to -1)
* Fixing mis-match in deletions  workflow where we were deleting manifests before the the IndexRouting paths which could lead to IndexRouting paths being dangled 
* Fixing issue where we were moving the `lastCleanupAttemptStateVersion` even though the cleanup were failing which resulted in next deletes to become no-op if there were no cluster-state changes (or less than 10) and previous deletions failed.
* Reducing logging overhead of DEBUG as it was printing every single file to delete. Moved it to TRACE.

### Related Issues
Resolves #20564 

### Check List

### DIFF
--- a/server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
+++ b/server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
@@ -755,6 +755,36 @@ public class RemoteClusterStateCleanupManagerTests extends OpenSearchTestCase {
         assertFalse(remoteClusterStateCleanupManager.getStaleFileDeletionTask().isClosed());
     }
 
+    public void testUpdateCleanupIntervalReschedulesWhenTaskNotScheduled() {
+        remoteClusterStateCleanupManager.start();
+
+        TimeValue disabledInterval = TimeValue.MINUS_ONE;
+        Settings disableSettings = Settings.builder().put("cluster.remote_store.state.cleanup_interval", disabledInterval).build();
+        clusterSettings.applySettings(disableSettings);
+
+        assertFalse(remoteClusterStateCleanupManager.getStaleFileDeletionTask().isScheduled());
+
+        TimeValue enabledInterval = TimeValue.timeValueMinutes(10);
+        Settings enableSettings = Settings.builder().put("cluster.remote_store.state.cleanup_interval", enabledInterval).build();
+        clusterSettings.applySettings(enableSettings);
+
+        assertTrue(remoteClusterStateCleanupManager.getStaleFileDeletionTask().isScheduled());
+        assertEquals(enabledInterval, remoteClusterStateCleanupManager.getStaleFileDeletionTask().getInterval());
+    }
+
+    public void testUpdateCleanupIntervalDoesNotRescheduleWhenTaskScheduled() {
+        remoteClusterStateCleanupManager.start();
+
+        assertTrue(remoteClusterStateCleanupManager.getStaleFileDeletionTask().isScheduled());
+
+        TimeValue newInterval = TimeValue.timeValueMinutes(10);
+        Settings newSettings = Settings.builder().put("cluster.remote_store.state.cleanup_interval", newInterval).build();
+        clusterSettings.applySettings(newSettings);
+
+        assertTrue(remoteClusterStateCleanupManager.getStaleFileDeletionTask().isScheduled());
+        assertEquals(newInterval, remoteClusterStateCleanupManager.getStaleFileDeletionTask().getInterval());
+    }
+
     public void testRemoteCleanupSkipsOnOnlyElectedClusterManager() {
         DiscoveryNodes nodes = mock(DiscoveryNodes.class);
         when(nodes.isLocalNodeElectedClusterManager()).thenReturn(false);
@@ -1171,12 +1201,25 @@ public class RemoteClusterStateCleanupManagerTests extends OpenSearchTestCase {
             anyString(),
             anyString(),
             eq(batch2.subList(0, manifestsToRetain)),
-            staleCaptor.capture()
+            eq(batch2.subList(manifestsToRetain, batch2.size()))
         );
-        List<List<BlobMetadata>> staleInvocations = staleCaptor.getAllValues();
-        assertTrue(
-            "Second invocation should have manifest-91.dat in stale list",
-            staleInvocations.get(0).stream().anyMatch(blob -> "manifest-91.dat".equals(blob.name()))
+    }
+
+    public void testDeleteClusterMetadataIllegalStateException() throws IOException {
+        String clusterName = "test-cluster";
+        String clusterUUID = "test-uuid";
+        List<BlobMetadata> activeBlobs = Arrays.asList(new PlainBlobMetadata("manifest1.dat", 1L));
+        List<BlobMetadata> staleBlobs = Arrays.asList(new PlainBlobMetadata("manifest2.dat", 1L));
+
+        when(remoteManifestManager.fetchRemoteClusterMetadataManifest(eq(clusterName), eq(clusterUUID), any())).thenThrow(
+            new IllegalStateException("test illegal state")
         );
+
+        remoteClusterStateCleanupManager.start();
+
+        IllegalStateException thrown = assertThrows(IllegalStateException.class, () -> {
+            remoteClusterStateCleanupManager.deleteClusterMetadata(clusterName, clusterUUID, activeBlobs, staleBlobs);
+        });
+        assertEquals("test illegal state", thrown.getMessage());
     }
 }
--- a/server/src/test/java/org/opensearch/cluster/routing/remote/RemoteRoutingTableServiceTests.java
+++ b/server/src/test/java/org/opensearch/cluster/routing/remote/RemoteRoutingTableServiceTests.java
@@ -28,6 +28,7 @@ import org.opensearch.common.blobstore.stream.write.WritePriority;
 import org.opensearch.common.compress.DeflateCompressor;
 import org.opensearch.common.settings.ClusterSettings;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.util.TestCapturingListener;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.compress.Compressor;
@@ -834,5 +835,55 @@ public class RemoteRoutingTableServiceTests extends OpenSearchTestCase {
         );
         assertEquals("test exception", thrown.getMessage());
         verify(blobContainer).deleteBlobsIgnoringIfNotExists(stalePaths);
+    }
+
+    public void testDeleteStaleIndexRoutingPathsWithTimeValue() throws IOException {
+        doNothing().when(blobContainer).deleteBlobsIgnoringIfNotExists(any(), any(TimeValue.class));
+        when(blobStore.blobContainer(any())).thenReturn(blobContainer);
+        List<String> stalePaths = Arrays.asList("path1", "path2");
+        TimeValue timeValue = TimeValue.timeValueSeconds(30);
+        remoteRoutingTableService.doStart();
+        remoteRoutingTableService.deleteStaleIndexRoutingPaths(stalePaths, timeValue);
+        verify(blobContainer).deleteBlobsIgnoringIfNotExists(stalePaths, timeValue);
+    }
+
+    public void testDeleteStaleIndexRoutingPathsWithTimeValueThrowsIOException() throws IOException {
+        when(blobStore.blobContainer(any())).thenReturn(blobContainer);
+        List<String> stalePaths = Arrays.asList("path1", "path2");
+        TimeValue timeValue = TimeValue.timeValueSeconds(30);
+        doThrow(new IOException("test exception")).when(blobContainer)
+            .deleteBlobsIgnoringIfNotExists(Mockito.anyList(), any(TimeValue.class));
+
+        remoteRoutingTableService.doStart();
+        IOException thrown = assertThrows(IOException.class, () -> {
+            remoteRoutingTableService.deleteStaleIndexRoutingPaths(stalePaths, timeValue);
+        });
+        assertEquals("test exception", thrown.getMessage());
+        verify(blobContainer).deleteBlobsIgnoringIfNotExists(stalePaths, timeValue);
+    }
+
+    public void testDeleteStaleIndexRoutingDiffPathsWithTimeValue() throws IOException {
+        doNothing().when(blobContainer).deleteBlobsIgnoringIfNotExists(any(), any(TimeValue.class));
+        when(blobStore.blobContainer(any())).thenReturn(blobContainer);
+        List<String> stalePaths = Arrays.asList("path1", "path2");
+        TimeValue timeValue = TimeValue.timeValueSeconds(30);
+        remoteRoutingTableService.doStart();
+        remoteRoutingTableService.deleteStaleIndexRoutingDiffPaths(stalePaths, timeValue);
+        verify(blobContainer).deleteBlobsIgnoringIfNotExists(stalePaths, timeValue);
+    }
+
+    public void testDeleteStaleIndexRoutingDiffPathsWithTimeValueThrowsIOException() throws IOException {
+        when(blobStore.blobContainer(any())).thenReturn(blobContainer);
+        List<String> stalePaths = Arrays.asList("path1", "path2");
+        TimeValue timeValue = TimeValue.timeValueSeconds(30);
+        doThrow(new IOException("test exception")).when(blobContainer)
+            .deleteBlobsIgnoringIfNotExists(Mockito.anyList(), any(TimeValue.class));
+
+        remoteRoutingTableService.doStart();
+        IOException thrown = assertThrows(IOException.class, () -> {
+            remoteRoutingTableService.deleteStaleIndexRoutingDiffPaths(stalePaths, timeValue);
+        });
+        assertEquals("test exception", thrown.getMessage());
+        verify(blobContainer).deleteBlobsIgnoringIfNotExists(stalePaths, timeValue);
     }
 }


Pranshu-S:
Appears to be stuck -

```
WORKFLOW_URL
time passed: 3960
```

Pranshu-S:
Looks like the execution has not started.

```
time passed: 5340
WORKFLOW_URL
```

Pranshu-S:
Error - 
[Test Result](https://build.ci.opensearch.org/job/gradle-check/70869/testReport/) (1 failure / ±0)
[org.opensearch.gradle.internal.InternalDistributionBwcSetupPluginFuncTest.builds distribution from branches via archives assemble](https://build.ci.opensearch.org/job/gradle-check/70869/testReport/junit/org.opensearch.gradle.internal/InternalDistributionBwcSetupPluginFuncTest/builds_distribution_from_branches_via_archives_assemble/)


Flaky - https://github.com/opensearch-project/OpenSearch/issues/19808

server/src/main/java/org/opensearch/index/translog/transfer/TransferService.java
```diff
@@ -101,6 +102,10 @@ void uploadBlob(
 
     void deleteBlobs(Iterable<String> path, List<String> fileNames) throws IOException;
 
+    default void deleteBlobs(Iterable<String> path, List<String> fileNames, TimeValue timeout) throws IOException {
+        deleteBlobs(path, fileNames);
+    }
```
rajiv-kv, line 107: this seems to be ignoring the timeout passed by caller, do we need a default implementation ?

server/src/main/java/org/opensearch/common/blobstore/BlobContainer.java
```diff
@@ -281,6 +282,17 @@ default void writeBlobAtomicWithMetadata(
...
 
+    /**
+     * Deletes the blobs with given names. This method will not throw an exception
+     * when one or multiple of the given blobs don't exist and simply ignore this case.
+     *
+     * @param   blobNames  The names of the blob to delete.
+     * @throws  IOException if a subset of blob exists but could not be deleted.
+     */
+    default void deleteBlobsIgnoringIfNotExists(List<String> blobNames, TimeValue timeout) throws IOException {
+        deleteBlobsIgnoringIfNotExists(blobNames);
```
rajiv-kv, line 295: if it is to avoid compilation errors, can we throw exception that it is not supported ?

plugins/repository-s3/src/test/java/org/opensearch/repositories/s3/S3BlobContainerTimeoutTests.java
```diff
@@ -0,0 +1,109 @@
...
+            return null;
+        }).when(s3BlobContainer).getFutureValue(any(), eq(timeout));
+
+        s3BlobContainer.deleteBlobsIgnoringIfNotExists(blobNames, timeout);
+
+        verify(s3BlobContainer).deleteBlobsAsyncIgnoringIfNotExists(eq(blobNames), any());
+        verify(s3BlobContainer).getFutureValue(any(), eq(timeout));
+    }
+
+    public void testGetFutureValueWithTimeout() throws Exception {
```
rajiv-kv, line 61: not sure if testGetFutureValueWithTimeout and  testGetFutureValueWithMinusOneTimeout are adding much value

plugins/repository-s3/src/test/java/org/opensearch/repositories/s3/S3BlobContainerTimeoutTests.java
```diff
@@ -0,0 +1,109 @@
...
+
+        future.onResponse(expectedResult);
+
+        S3BlobContainer realContainer = new S3BlobContainer(BlobPath.cleanPath(), mockBlobStore);
+
+        String result = realContainer.getFutureValue(future, timeout);
+        assertEquals(expectedResult, result);
+    }
+
+    public void testDeleteBlobsWithDefaultTimeout() throws IOException {
```
rajiv-kv, line 74: looks some duplicate code with `testDeleteBlobsWithCustomTimeout`

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -48,8 +48,15 @@ public class RemoteClusterStateCleanupManager implements Closeable {
...
     public static final int SKIP_CLEANUP_STATE_CHANGES = 10;
+    public static final int MANIFEST_CLEANUP_BATCH_SIZE_DEFAULT = 1000;
+    public static final int MANIFEST_CLEANUP_MAX_BATCHES_DEFAULT = 100;
     public static final TimeValue CLUSTER_STATE_CLEANUP_INTERVAL_DEFAULT = TimeValue.timeValueMinutes(5);
     public static final TimeValue CLUSTER_STATE_CLEANUP_INTERVAL_MINIMUM = TimeValue.MINUS_ONE;
+    public static final TimeValue CLUSTER_STATE_CLEANUP_TIMEOUT_DEFAULT = TimeValue.MINUS_ONE;
+
+    public static final String REMOTE_CLUSTER_STATE_CLEANUP_BATCH_SIZE_SETTING_NAME = "cluster.remote_store.state.cleanup_batch_size";
+    public static final String REMOTE_CLUSTER_STATE_CLEANUP_TIMEOUT_SETTING_NAME = "cluster.remote_store.state.cleanup_timeout";
+    public static final String REMOTE_CLUSTER_STATE_CLEANUP_MAX_BATCHES_SETTING_NAME = "cluster.remote_store.state.cleanup_max_batches";
```
rajiv-kv: nit : should we use dot instead of underscore `cluster.remote_store.state.cleanup.batch_size`

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -62,11 +69,47 @@ public class RemoteClusterStateCleanupManager implements Closeable {
...
         Setting.Property.Dynamic
     );
+
+    /**
+     * Setting to specify the batch size for manifest cleanup operations
+     */
+    public static final Setting<Integer> REMOTE_CLUSTER_STATE_CLEANUP_BATCH_SIZE_SETTING = Setting.intSetting(
+        REMOTE_CLUSTER_STATE_CLEANUP_BATCH_SIZE_SETTING_NAME,
+        MANIFEST_CLEANUP_BATCH_SIZE_DEFAULT,
+        RETAINED_MANIFESTS + SKIP_CLEANUP_STATE_CHANGES,
```
rajiv-kv, line 80: we can move minimum size retention to the implementation (consuming the min batch size)  and not couple with minimum value of the setting. The min value for this setting can be a constant.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -332,7 +419,9 @@ void deleteClusterMetadata(
                     e
                 );
                 remoteStateStats.indexRoutingFilesCleanupAttemptFailed();
+                throw e;
```
rajiv-kv, line 428: why do we need to throw error ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -366,42 +457,53 @@ void deleteStaleClusterMetadata(String clusterName, String clusterUUID, int mani
...
+                        remoteManifestManager.getManifestFolderPath(clusterName, clusterUUID),
+                        MANIFEST,
+                        cleanupBatchSize
+                    );
+
+                    if (batchManifests.size() > manifestsToRetain) {
+                        List<BlobMetadata> manifestsToDeletes = batchManifests.subList(manifestsToRetain, batchManifests.size());
+                        logger.debug("[Batch {}] Deleting [{}] stale manifests", batchesProcessed, manifestsToDeletes.size());
+
+                        deleteClusterMetadata(clusterName, clusterUUID, batchManifests.subList(0, manifestsToRetain), manifestsToDeletes);
```
rajiv-kv, line 504: should we cache the `ClusterMetadataManifest` of `manifestsToRetain` across the batches  (to avoid repeated fetches + deser) ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -366,42 +457,53 @@ void deleteStaleClusterMetadata(String clusterName, String clusterUUID, int mani
...
-        } catch (Exception e) {
-            deleteStaleMetadataRunning.set(false);
-            throw e;
-        }
+
+                if (batchesProcessed == cleanupMaxBatches) {
+                    logger.warn("Exhausted batch limit for deleting entities. Attempted [{}] batches", cleanupMaxBatches);
+                }
+            } catch (Exception e) {
+                logger.error("Error during batch cleanup of manifests for cluster [{}]. Exception: {}", clusterName, e);
```
rajiv-kv: we can retain the old message "Exception occurred while deleting Remote Cluster Metadata for clusterUUIDs {}",

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
```diff
@@ -287,18 +295,22 @@ public void testDeleteClusterMetadata() throws IOException {
...
                 new BlobPath().add(GLOBAL_METADATA_PATH_TOKEN).buildAsString() + "global_metadata.dat"
-            )
+            ),
+            CLUSTER_STATE_CLEANUP_TIMEOUT_DEFAULT
+        );
+        verify(container).deleteBlobsIgnoringIfNotExists(
+            List.of(getFormattedIndexFileName(index1Metadata.getUploadedFilePath())),
+            CLUSTER_STATE_CLEANUP_TIMEOUT_DEFAULT
         );
-        verify(container).deleteBlobsIgnoringIfNotExists(List.of(getFormattedIndexFileName(index1Metadata.getUploadedFilePath())));
```
rajiv-kv, line 292: should we add verify `never()` for this method interaction in the code ? Also should we add `verifyNoMoreInteractions ` for container mock.

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
```diff
@@ -790,4 +797,302 @@ public void testRemoteCleanupSchedulesEvenAfterFailure() {
         assertTrue(task.isScheduled());
         assertEquals(2, callCount.get());
     }
+
+    public void testRemoteClusterStateCleanupBatchSizeSetting() {
```
rajiv-kv: do we need setting verification related tests ? Not seeing much value.

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
```diff
@@ -790,4 +797,302 @@ public void testRemoteCleanupSchedulesEvenAfterFailure() {
...
+        List<BlobMetadata> manifests = new ArrayList<>();
+        for (int i = 0; i < 15; i++) {
+            manifests.add(new PlainBlobMetadata("manifest" + i + ".dat", 1L));
+        }
+
+        when(remoteManifestManager.getManifestFolderPath(eq(clusterName), eq(clusterUUID))).thenReturn(
+            new BlobPath().add(encodeString(clusterName)).add(CLUSTER_STATE_PATH_TOKEN).add(clusterUUID).add(MANIFEST)
+        );
+
+        RemoteClusterStateCleanupManager spyManager = spy(remoteClusterStateCleanupManager);
```
rajiv-kv: spyManager -> cleanUpManager

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
```diff
@@ -790,4 +797,302 @@ public void testRemoteCleanupSchedulesEvenAfterFailure() {
...
+
+        for (int i = 0; i < batchSize; i++) {
+            batch1.add(new PlainBlobMetadata("manifest" + i + ".dat", 1L));
+            batch2.add(new PlainBlobMetadata("manifest" + (i + batchSize) + ".dat", 1L));
+            batch3.add(new PlainBlobMetadata("manifest" + (i + 2 * batchSize) + ".dat", 1L));
+        }
+
+        when(mockTransferService.listAllInSortedOrder(any(), eq(MANIFEST), eq(batchSize))).thenReturn(batch1, batch2, batch3);
+
+        CountDownLatch latch = new CountDownLatch(maxBatches);
```
rajiv-kv, line 1016: is it possible to verify the list of manifest files that are being deleted in each batch ?

Should we spy on the other delete method
```
  deleteClusterMetadata(
        String clusterName,
        String clusterUUID,
        List<BlobMetadata> activeManifestBlobMetadata,
        List<BlobMetadata> staleManifestBlobMetadata
```

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
```diff
@@ -790,4 +797,302 @@ public void testRemoteCleanupSchedulesEvenAfterFailure() {
...
+        BlobStoreTransferService mockTransferService = mock(BlobStoreTransferService.class);
+
+        List<BlobMetadata> batch1 = new ArrayList<>();
+        List<BlobMetadata> batch2 = new ArrayList<>();
+        List<BlobMetadata> batch3 = new ArrayList<>();
+
+        for (int i = 0; i < batchSize; i++) {
+            batch1.add(new PlainBlobMetadata("manifest" + i + ".dat", 1L));
+            batch2.add(new PlainBlobMetadata("manifest" + (i + batchSize) + ".dat", 1L));
+            batch3.add(new PlainBlobMetadata("manifest" + (i + 2 * batchSize) + ".dat", 1L));
```
rajiv-kv, line 1011: Should we add cases 
- no new file created b/w batch list (batches having top-n files repeated)
- one additional manifest file created b/w batch list 
- RETAINED_MANIFESTS manifest files created b/w batch list

server/src/main/java/org/opensearch/index/translog/transfer/TransferService.java
Pranshu-S: I can either throw an UnsupportedOperationException or add a WARN log incase this is the fallback for not implemented. I had done it in this format for testing but I suppose throwing an UnsupportedOperationException would be better. 

Will update

server/src/main/java/org/opensearch/common/blobstore/BlobContainer.java
Pranshu-S: Yes, will update.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Both `RETAINED_MANIFESTS` and `SKIP_CLEANUP_STATE_CHANGES` are constants.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Was for testing, will revert.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: We would need to implement a listener on the ListObjectsV2 paginated response if we want to keep the manifests and iterate through each page of ListObjects however such an implementation is not present today. In order to over-complicate this fix - I have avoided that change.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Sure, got it.

plugins/repository-s3/src/test/java/org/opensearch/repositories/s3/S3BlobContainerTimeoutTests.java
Pranshu-S: True, will remove.

plugins/repository-s3/src/test/java/org/opensearch/repositories/s3/S3BlobContainerTimeoutTests.java
Pranshu-S: It validates the timeout value is provided when calling `future.get(timeout, unit)`

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Ack, will update.

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
Pranshu-S: Yes, will update.

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
Pranshu-S: Fair, will remove

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
Pranshu-S: Yes, that will improve the validations. Will add.

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
Pranshu-S: Will update.

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
Pranshu-S: Will update.

server/src/test/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerTests.java
Pranshu-S: Ack, will update.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
rajiv-kv: I was thinking on the lines of introducing a hashmap which is created outside the while loop and passing it along with `deleteClusterMetadata`.

Pranshu-S:
Test result - 
```
[Test Result](https://build.ci.opensearch.org/job/gradle-check/71031/testReport/) (2 failures / -3)
[org.opensearch.cluster.shards.ClusterShardLimitIT.testOpenIndexOverLimit {p0={"opensearch.experimental.feature.writable_warm_index.enabled":"true"}}](https://build.ci.opensearch.org/job/gradle-check/71031/testReport/junit/org.opensearch.cluster.shards/ClusterShardLimitIT/testOpenIndexOverLimit__p0___opensearch_experimental_feature_writable_warm_index_enabled___true___/)
[org.opensearch.remotestore.RestoreShallowSnapshotV2IT.testRestoreOperationsShallowCopyEnabled {p0={"opensearch.experimental.feature.writable_warm_index.enabled":"true"}}](https://build.ci.opensearch.org/job/gradle-check/71031/testReport/junit/org.opensearch.remotestore/RestoreShallowSnapshotV2IT/testRestoreOperationsShallowCopyEnabled__p0___opensearch_experimental_feature_writable_warm_index_enabled___true___/)
```

Flaky - https://github.com/opensearch-project/OpenSearch/issues/19916
Flaky - https://github.com/opensearch-project/OpenSearch/issues/16658

Pranshu-S:
Closing this PR in virtue of https://github.com/opensearch-project/OpenSearch/pull/20566

PR #20526, opened 2026-02-02
Add warmup phase for pull-based ingestion

### LINKED ISSUE #20506: [Feature Request] Wait for lag to catch up in pull-based ingestion

In pull-based ingestion, currently whenever a new node is added or a shard relocation happens, we start consuming from the previous checkpoint offset from Kafka/Kinesis. During this time until the lag is not consumed, the node is surveying using stale data. Ideally, we should wait for the lag to get caught up or go below certain threshold before we go into serving state. 

When a node is added or is being restarted, or there is a shard relocation during boot up, we will wait for the lag to get caught up for a certain timeout period. Once the lag is caught up, only then we will go into serving state. 

### Related component

Indexing

_No response_

### Additional context

_No response_

kaustubhbutte17:
## Description
In pull-based ingestion, when a new node is added or shard relocation happens, consumption starts from the previous checkpoint offset. During this time, the shard serves stale data until the lag is consumed.

This PR introduces a "warmup" phase that prevents shards from serving queries until they have caught up with the streaming source - analogous to how push-based replication waits for replicas to sync before serving.

Resolves #20506

## Solution
When warmup is enabled, a shard entering the `postRecovery` phase will block until:
1. The pointer-based lag drops to or below the configured threshold, OR
2. The timeout is reached (proceeds with warning)

This ensures data consistency for users who prefer correctness over immediate availability.

## New Settings
| Setting | Type | Default | Description |
|---------|------|---------|-------------|
| `index.ingestion_source.warmup.timeout` | TimeValue | **-1** (disabled) | Warmup wait time. `-1` = disabled, `>=0` = enabled with that timeout |
| `index.ingestion_source.warmup.lag_threshold` | long | **100** | Acceptable pointer-based message lag for warmup completion |

All settings are **Final** - set at index creation time, cannot be changed later.

## State Machine

```
Index Created -> WARMING_UP -> (lag <= threshold OR timeout) -> POLLING -> STARTED
```

During WARMING_UP:
- Poller is actively ingesting from the source (internally POLLING/PROCESSING)
- But externally reports state as `WARMING_UP` via `GET /{index}/ingestion/_state`
- Shard remains in `INITIALIZING` - does NOT serve search queries
- Search returns `503 all shards failed`
- If poller is paused, warmup is skipped immediately
- On timeout, warmup completes with a warning log (shard proceeds to serve)

### Enable warmup with 5 minute timeout:
```json
{
  "settings": {
    "index.ingestion_source.warmup.timeout": "5m"
  }
}
```

### Custom warmup with lag threshold:
```json
{
  "settings": {
    "index.ingestion_source.warmup.timeout": "10m",
    "index.ingestion_source.warmup.lag_threshold": 50
  }
}
```

### Warmup disabled (default):
No settings needed. Default `timeout=-1` means warmup is disabled.

## Implementation Details
- Added `WarmupConfig` record class in `IngestionSource` with two fields: `timeout` and `lagThreshold`
- `timeout = -1` means warmup disabled; `timeout >= 0` means enabled
- Warmup blocking occurs in `IndexShard.postRecovery()` before transitioning to `POST_RECOVERY` state
- Timeout and error handling is encapsulated in `IngestionEngine.awaitWarmupComplete()`
- Uses `CountDownLatch` for thread-safe blocking/signaling
- `cachedPointerBasedLag` initialized to `-1` to prevent premature warmup completion before lag data is available
- Warmup state visible via `GET /{index}/ingestion/_state` API (`poller_state: WARMING_UP`)
- When poller is paused during warmup, warmup completes immediately (skipped)
- On timeout, warmup always proceeds with a warning log (no shard failure to avoid livelock)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,13 +23,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add intra segment support for single-value metric aggregations ([#20503](https://github.com/opensearch-project/OpenSearch/pull/20503))
 - Add ref_path support for package-based hunspell dictionary loading ([#20840](https://github.com/opensearch-project/OpenSearch/pull/20840))
 - Add support for enabling pluggable data formats, starting with phase-1 of decoupling shard from engine, and introducing basic abstractions ([#20675](https://github.com/opensearch-project/OpenSearch/pull/20675))
-- Add getWrappedScorer method to ProfileScorer for plugin access to wrapped scorers ([#20548](https://github.com/opensearch-project/OpenSearch/issues/20548))
-- Support expected cluster name with validation in CCS Sniff mode ([#20532](https://github.com/opensearch-project/OpenSearch/pull/20532))
-- Add security policy to allow `accessUnixDomainSocket` in `transport-grpc` module ([#20463](https://github.com/opensearch-project/OpenSearch/pull/20463))
-- Add range validations in query builder and field mapper ([#20497](https://github.com/opensearch-project/OpenSearch/issues/20497))
-- [Workload Management] Enhance Scroll API support for autotagging ([#20151](https://github.com/opensearch-project/OpenSearch/pull/20151))
-- Add indices to search request slowlog ([#20588](https://github.com/opensearch-project/OpenSearch/pull/20588))
-- Add warmup phase to wait for lag to catch up in pull-based ingestion before serving ([#20526](https://github.com/opensearch-project/OpenSearch/pull/20526))
 
 - Add warmup phase to wait for lag to catch up in pull-based ingestion before serving ([#20526](https://github.com/opensearch-project/OpenSearch/pull/20526))
 ### Changed
--- a/plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
+++ b/plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
@@ -107,7 +107,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.topic", "test")
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.param.auto.offset.reset", "latest")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -136,7 +135,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.topic", "test")
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.param.auto.offset.reset", "latest")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -272,7 +270,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.pointer.init.reset", "earliest")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -372,7 +369,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.pointer.init.reset", "earliest")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -425,7 +421,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.pointer.init.reset", "earliest")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -507,7 +502,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.pointer.init.reset", "earliest")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -548,7 +542,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.pointer.init.reset", "earliest")
                 .put("ingestion_source.pointer_based_lag_update_interval", "3s")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -620,7 +613,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.pointer.init.reset", "earliest")
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -715,7 +707,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.pointer.init.reset", "earliest")
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .put("index.periodic_flush_interval", "5s")
                 .build(),
@@ -751,7 +742,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.pointer.init.reset", "earliest")
                 .put("ingestion_source.mapper_type", "raw_payload")
                 .put("ingestion_source.error_strategy", "drop")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             "{\"properties\":{\"name\":{\"type\": \"text\"},\"age\":{\"type\": \"integer\"}}}}"
@@ -869,7 +859,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.param.auto.offset.reset", "latest")
                 .put("ingestion_source.param.max.poll.records", "100")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             mapping
@@ -980,7 +969,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.param.auto.offset.reset", "none")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             mapping
@@ -1061,7 +1049,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
                 .put("ingestion_source.param.auto.offset.reset", "earliest")
-                .put("ingestion_source.warmup.enabled", false)
                 .put("ingestion_source.all_active", true)
                 .build(),
             mapping
@@ -1372,7 +1359,6 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
                 .put("ingestion_source.pointer.init.reset", "earliest")
                 .put("ingestion_source.param.topic", topicName)
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
-                .put("ingestion_source.warmup.enabled", true)
                 .put("ingestion_source.warmup.lag_threshold", 0)
                 .put("ingestion_source.warmup.timeout", "10m")
                 .put("ingestion_source.all_active", true)
--- a/server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
+++ b/server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
@@ -1032,24 +1032,14 @@ public class IndexMetadata implements Diffable<IndexMetadata>, ToXContentFragmen
     );
 
     /**
-     * Defines if warmup phase is enabled for pull-based ingestion. When enabled, shards will wait for
-     * lag to catch up before serving queries after node restart or shard relocation.
-     */
-    public static final String SETTING_INGESTION_SOURCE_WARMUP_ENABLED = "index.ingestion_source.warmup.enabled";
-    public static final Setting<Boolean> INGESTION_SOURCE_WARMUP_ENABLED_SETTING = Setting.boolSetting(
-        SETTING_INGESTION_SOURCE_WARMUP_ENABLED,
-        false,
-        Property.IndexScope,
-        Property.Final
-    );
-
-    /**
      * Defines the maximum time to wait for lag to catch up during warmup phase.
+     * A value of -1 means warmup is disabled (the default). A value >= 0 enables warmup with that timeout.
      */
     public static final String SETTING_INGESTION_SOURCE_WARMUP_TIMEOUT = "index.ingestion_source.warmup.timeout";
-    public static final Setting<TimeValue> INGESTION_SOURCE_WARMUP_TIMEOUT_SETTING = Setting.positiveTimeSetting(
+    public static final Setting<TimeValue> INGESTION_SOURCE_WARMUP_TIMEOUT_SETTING = Setting.timeSetting(
         SETTING_INGESTION_SOURCE_WARMUP_TIMEOUT,
-        new TimeValue(5, TimeUnit.MINUTES),
+        TimeValue.timeValueMillis(-1),
+        TimeValue.timeValueMillis(-1),
         Property.IndexScope,
         Property.Final
     );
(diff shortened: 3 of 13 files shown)


server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
```diff
@@ -992,6 +992,79 @@ public Iterator<Setting<?>> settings() {
         }, Property.IndexScope, Property.Dynamic)
     );
 
+    /**
+     * Defines if warmup phase is enabled for pull-based ingestion. When enabled, shards will wait for
+     * lag to catch up before serving queries after node restart or shard relocation.
+     */
+    public static final String SETTING_INGESTION_SOURCE_WARMUP_ENABLED = "index.ingestion_source.warmup.enabled";
```
varunbharadwaj: We could create a `WarmupConfig` class to hold enabled, warmupTimeout, offsetLagThreshold, timeLagThreshold, failOnTimeout attributes. This can be added in `IngestionSource`, similar to IngestionSource.PointerInitReset today.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -2473,7 +2473,31 @@ public void postRecovery(String reason) throws IndexShardStartedException, Index
             // we may not expose operations that were indexed with a refresh listener that was immediately
             // responded to in addRefreshListener. The refresh must happen under the same mutex used in addRefreshListener
             // and before moving this shard to POST_RECOVERY state (i.e., allow to read from this shard).
-            getEngine().refresh("post_recovery");
+            Engine engine = getEngine();
+            engine.refresh("post_recovery");
+
+            // Wait for ingestion warmup if enabled (pull-based ingestion only)
+            if (engine instanceof IngestionEngine) {
```
varunbharadwaj: This seems like a good place for warmup, as we are still in recovery mode, before updating to POST_RECOVERY state. I'm wondering if we can just call `ingestionEngine.awaitWarmupComplete` here and move timeout and error handling inside the IngestionEngine.awaitWarmupComplete. We just throw the exceptions from awaitWarmupComplete in case of timeout etc. That way, most of the pull-based logic is contained within the engine. What do you think?

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
```diff
@@ -992,6 +992,79 @@ public Iterator<Setting<?>> settings() {
...
+
+    /**
+     * Defines the maximum time to wait for lag to catch up during warmup phase.
+     */
+    public static final String SETTING_INGESTION_SOURCE_WARMUP_TIMEOUT = "index.ingestion_source.warmup.timeout";
+    public static final Setting<TimeValue> INGESTION_SOURCE_WARMUP_TIMEOUT_SETTING = Setting.positiveTimeSetting(
+        SETTING_INGESTION_SOURCE_WARMUP_TIMEOUT,
+        new TimeValue(5, TimeUnit.MINUTES),
+        Property.IndexScope,
+        Property.Final
```
varunbharadwaj, line 1044: In a subsequent PR, we could make these settings dynamic to avoid recreating the index when we need to change / mitigate in case of an incident.

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -192,6 +233,13 @@ protected void startPoll() {
         }
         logger.info("Starting poller for shard {}", shardId);
 
+        // Initialize warmup if enabled
+        if (warmupEnabled && !warmupComplete) {
+            warmupStartTime = System.currentTimeMillis();
+            state = State.WARMING_UP;
+            logger.info("Starting warmup phase for shard {}, waiting for lag to catch up", shardId);
```
varunbharadwaj: Let's also include indexName in the log

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -203,12 +251,21 @@ protected void startPoll() {
                 // Initialize/reinitialization consumer
                 if (this.consumer == null || reinitializeConsumer) {
                     handleConsumerInitialization();
+                    // Still check warmup status even if consumer init fails, so timeout can trigger
+                    if (!warmupComplete && warmupEnabled) {
```
varunbharadwaj: I think we could skip this checkWarmupStatus() and only check after updatePointerBasedLagIfNeeded is called, to let the offset-based lag value to be computed for the first time

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -352,6 +409,77 @@ public boolean isClosed() {
...
+    /**
+     * Returns true if shard initialization should fail when warmup times out.
+     */
+    public boolean isWarmupFailOnTimeout() {
+        return warmupFailOnTimeout;
+    }
+
+    @Override
+    public boolean awaitWarmupComplete(long timeoutMs) throws InterruptedException {
+        if (!warmupEnabled || warmupComplete) {
```
varunbharadwaj: We could call isWarmupComplete()

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -352,6 +409,77 @@ public boolean isClosed() {
...
+    /**
+     * Check if warmup conditions are met and mark warmup as complete if so.
+     */
+    private void checkWarmupStatus() {
+        long currentLag;
+        long threshold;
+
+        // Determine which lag metric to use
+        if (warmupUseTimeLag || cachedPointerBasedLag < 0) {
+            // Use time-based lag for Kinesis or when pointer lag unavailable
```
varunbharadwaj: cachedPointerBasedLag is 0 by default, when pointer based lag calculation is disabled or in case of Kinesis when it is not supported. We could highlight this and let the user decide if they want to use offset based of time based lag. Depending on which one configured, check the lag or skip the warmup. For offset based lag, we will also need pointerBasedLagUpdateIntervalMs to be set (enabled by default) - we can highlight this point.

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -352,6 +409,77 @@ public boolean isClosed() {
...
+            currentLag = computeTimeBasedLag();
```
varunbharadwaj: I think if we try to retrieve the time based lag before the first message is read, time-based lag from lastPolledMessageTimestamp will be 0. Maybe we also have to take this into consideration 🤔

CHANGELOG.md
```diff
@@ -5,6 +5,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.x]
 ### Added
+- Add support for fields containing dots in their name as literals ([#19958](https://github.com/opensearch-project/OpenSearch/pull/19958))
```
varunbharadwaj: These entries were removed once we moved from 3.4 to 3.5. We can manually update this file based on latest [CHANGELOG](https://github.com/opensearch-project/OpenSearch/blob/main/CHANGELOG.md?plain=1)

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
```diff
@@ -992,6 +992,79 @@ public Iterator<Setting<?>> settings() {
...
+    );
+
+    /**
+     * Defines the acceptable pointer-based lag threshold. Warmup completes when lag is at or below this value.
+     * A value of 0 means fully caught up (no lag).
+     */
+    public static final String SETTING_INGESTION_SOURCE_WARMUP_LAG_THRESHOLD = "index.ingestion_source.warmup.lag_threshold";
+    public static final Setting<Long> INGESTION_SOURCE_WARMUP_LAG_THRESHOLD_SETTING = Setting.longSetting(
```
varunbharadwaj, line 1052: If we only want to allow either offset or time based warmup, we could add a validation here to ensure only one of them is set. Reference: [link](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java#L942)

msfroh:
I feel like a default `lag_threshold` of zero may be dangerous, since you could end up with a scenario where ingestion is fast enough that we're always a little behind (but not falling farther behind), so the shard never really comes up.

---

In the long run, I feel like the best option is to leave the lag decision up to the coordinator. (At least that's the approach I've seen/implemented in the past.) 

Essentially, if you have (say) three possible shard copies to search and one of them is lagging behind (whether because of recovery or because the node itself had a temporary slowdown), then the coordinator can choose to ignore the lagging shard. On the other hand, if all shard copies are lagging by similar amounts (because of an indexing surge), then they're all equally bad.

That's a bigger architectural change, though, since it involves reporting freshness metrics from shards to coordinators.

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
kaustubhbutte17: good point, will take it up in a following PR

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
kaustubhbutte17: removed time based lag check option for now

kaustubhbutte17:
> I feel like a default `lag_threshold` of zero may be dangerous, since you could end up with a scenario where ingestion is fast enough that we're always a little behind (but not falling farther behind), so the shard never really comes up.
> 
> In the long run, I feel like the best option is to leave the lag decision up to the coordinator. (At least that's the approach I've seen/implemented in the past.)
> 
> Essentially, if you have (say) three possible shard copies to search and one of them is lagging behind (whether because of recovery or because the node itself had a temporary slowdown), then the coordinator can choose to ignore the lagging shard. On the other hand, if all shard copies are lagging by similar amounts (because of an indexing surge), then they're all equally bad.
> 
> That's a bigger architectural change, though, since it involves reporting freshness metrics from shards to coordinators.

Yeah thats a good point. For this reason, I have kept a timeout safeguard so even in case the lag isnt caught up, the node will still go into serving state after reaching a 5 min timeout. There is a config to optionally fail the shard if the lag is not caught up. For now do you recommend keeping a little higher non-zero threshold for lag catchup?

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -654,4 +658,53 @@ public ShardIngestionState getIngestionState() {
...
+    /**
+     * Block until warmup is complete or timeout occurs.
+     * This method handles all warmup logic internally including timeout handling and error throwing.
+     *
+     * @throws OpenSearchException if warmup times out and failOnTimeout is configured
+     * @throws InterruptedException if the thread is interrupted while waiting
+     */
+    public void awaitWarmupComplete() throws InterruptedException {
+        IngestionSource ingestionSource = engineConfig.getIndexSettings().getIndexMetadata().getIngestionSource();
+        if (ingestionSource == null || !ingestionSource.isWarmupEnabled()) {
```
varunbharadwaj: Should we also check poller status, and skip warmup if poller is paused?
```
if (ingestionSource == null || !ingestionSource.isWarmupEnabled() || streamPoller.isPaused()) {
}
```

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -209,6 +247,11 @@ protected void startPoll() {
                 // Update lag periodically. Lag is updated even if the poller is paused.
                 updatePointerBasedLagIfNeeded();
 
+                // Check warmup status if not yet complete
+                if (!warmupComplete && warmupEnabled) {
```
varunbharadwaj: We can simplify this as
```
if (isWarmupComplete() == false) {
    checkWarmupStatus();
}
```

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
varunbharadwaj: We will be resetting the State.WARMING_UP status at this point to POLLING. Should we skip updating to polling if status is warmup? 🤔 
This might help users identify warming up shard. Also it makes it easier in ITs to check when warmup is done. What do you think?

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
```diff
@@ -107,6 +107,7 @@ public void testKafkaIngestion_RewindByTimeStamp() {
                 .put("ingestion_source.param.topic", "test")
                 .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
```
varunbharadwaj, line 108: Let's add an IT for warmup in IngestFromKafkaIT.java. We can keep it simple for basic functionality. Something like - 
1. Publish 10 messages
2. Set lag threshold as 0, with a long enough timeout (ex: 10min)
3. Wait for poller to enter polling status (assuming we keep warmup status till warm up is done)
4. We can validate doc count = 10.

server/src/main/java/org/opensearch/cluster/metadata/IngestionSource.java
```diff
@@ -221,6 +252,77 @@ public String toString() {
...
+     * Class encapsulating the warmup configuration for pull-based ingestion.
+     * When warmup is enabled, shards will wait for lag to catch up before serving queries
+     * after node restart or shard relocation.
+     */
+    @ExperimentalApi
+    public static class WarmupConfig {
+        /**
+         * Default warmup configuration matching IndexMetadata setting defaults.
+         */
+        public static final WarmupConfig DEFAULT = new WarmupConfig(true, TimeValue.timeValueMinutes(5), 100L, false);
```
msfroh: This feels dangerous, since we're repeating the defaults from the settings. It's very easy for them to get out of sync.

Given that we should always load the `WarmupConfig` based on the index settings (which will populate the default values), this constant feels unnecessary.

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
```diff
@@ -1091,4 +1104,52 @@ public void testDynamicConfigUpdateOnNoMessages() throws Exception {
...
+
+        // Step 4: Wait for poller to enter POLLING state (warmup complete)
+        waitForState(() -> {
+            GetIngestionStateResponse ingestionState = getIngestionState(indexName);
+            return ingestionState.getShardStates().length == 1
+                && ingestionState.getShardStates()[0].getPollerState().equalsIgnoreCase("polling");
+        });
+
+        // Step 5: Validate all 10 documents are searchable
+        waitForSearchableDocs(10, List.of(nodeA));
```
varunbharadwaj: As the expectation is to index all 10 documents in warmup state, all of them must be visible as soon as we enter the polling state. I think we can replace the `waitForSearchableDocs` with a single query and assert for 10 documents. This will catch any issues in warmup logic better if in case we don't have 10 docs available at this time. We could call a refresh() before executing the query to make all existing docs visible.

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
```diff
@@ -1003,6 +1003,54 @@ public Iterator<Setting<?>> settings() {
...
+        Property.Final
+    );
+
+    /**
+     * Defines the maximum time to wait for lag to catch up during warmup phase.
+     */
+    public static final String SETTING_INGESTION_SOURCE_WARMUP_TIMEOUT = "index.ingestion_source.warmup.timeout";
+    public static final Setting<TimeValue> INGESTION_SOURCE_WARMUP_TIMEOUT_SETTING = Setting.positiveTimeSetting(
+        SETTING_INGESTION_SOURCE_WARMUP_TIMEOUT,
+        new TimeValue(5, TimeUnit.MINUTES),
```
varunbharadwaj: Shall we use a smaller default for now till we make it a dynamic config?

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -352,6 +380,108 @@ public boolean isClosed() {
...
+    /**
+     * Check if warmup conditions are met and mark warmup as complete if so.
+     *
+     * Warmup uses offset-based lag (cachedPointerBasedLag) which tracks the difference between
+     * the current consumer position and the end of the stream. This is the preferred mode for
+     * Kafka and other sources that support offset-based lag calculation.
+     * Note: cachedPointerBasedLag is 0 by default and is only updated after updatePointerBasedLagIfNeeded()
+     * is called.
+     */
+    private void checkWarmupStatus() {
```
varunbharadwaj: While we are making the other changes, could maybe rename this to updateWarmupStatus as it updates the state?

kaustubhbutte17:
CI failure in `:qa:rolling-upgrade:v2.19.5-remote#twoThirdsUpgradedTest` is a flaky test (process died during rolling upgrade test). Not related to this PR. Retriggering.

run gradle check

server/src/main/java/org/opensearch/cluster/metadata/IngestionSource.java
```diff
@@ -233,6 +265,72 @@ public String toString() {
         }
     }
 
+    /**
+     * Class encapsulating the warmup configuration for pull-based ingestion.
+     * When warmup is enabled, shards will wait for lag to catch up before serving queries
+     * after node restart or shard relocation.
+     */
+    @ExperimentalApi
```
varunbharadwaj: We have removed `@ExperimentalApi` recently for pull-based ingestion. Do you want to remove it for this one as well?

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -662,4 +663,53 @@ public ShardIngestionState getIngestionState() {
             shardPointer != null ? shardPointer.toString() : ""
         );
     }
+
+    /**
+     * Returns true if the warmup phase is complete and the shard is ready to serve.
+     */
+    public boolean isWarmupComplete() {
```
varunbharadwaj: Are we using this anywhere?

varunbharadwaj:
Looks good to me! Thanks for the changes! Minor comments.

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
```diff
@@ -1015,6 +1015,54 @@ public Iterator<Setting<?>> settings() {
         }, Property.IndexScope, Property.Final)
     );
 
+    /**
+     * Defines if warmup phase is enabled for pull-based ingestion. When enabled, shards will wait for
+     * lag to catch up before serving queries after node restart or shard relocation.
+     */
+    public static final String SETTING_INGESTION_SOURCE_WARMUP_ENABLED = "index.ingestion_source.warmup.enabled";
+    public static final Setting<Boolean> INGESTION_SOURCE_WARMUP_ENABLED_SETTING = Setting.boolSetting(
```
andrross: Do we need this setting? Could the "warmup disabled" behavior be implemented by having a warmup timeout of zero?

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
```diff
@@ -1015,6 +1015,54 @@ public Iterator<Setting<?>> settings() {
...
+        Property.IndexScope,
+        Property.Final
+    );
+
+    /**
+     * Defines if shard initialization should fail when warmup times out.
+     * If false, shard proceeds with a warning. If true, shard initialization fails.
+     */
+    public static final String SETTING_INGESTION_SOURCE_WARMUP_FAIL_ON_TIMEOUT = "index.ingestion_source.warmup.fail_on_timeout";
+    public static final Setting<Boolean> INGESTION_SOURCE_WARMUP_FAIL_ON_TIMEOUT_SETTING = Setting.boolSetting(
```
andrross: Is failing on timeout ever really the right answer? If you need your system to never serve data that is past some sort of staleness threshold, then you'll need a solution that can detect when an already-running shard falls past a certain staleness threshold. That same mechanism would detect newly initialized but stale shards.

The biggest concern here is whether the system would make incremental progress in the failure case. If not, then you could get stuck in a cycle where a shard never initializes in time, fails, then tries to get re-initialized somewhere else but gets stuck in a failure loop that never makes progress.

server/src/main/java/org/opensearch/cluster/metadata/IngestionSource.java
```diff
@@ -233,6 +264,72 @@ public String toString() {
         }
     }
 
+    /**
+     * Class encapsulating the warmup configuration for pull-based ingestion.
+     * When warmup is enabled, shards will wait for lag to catch up before serving queries
+     * after node restart or shard relocation.
+     */
+    @PublicApi(since = "3.6.0")
+    public static class WarmupConfig {
```
andrross: This can be a `record` class

kaustubhbutte17: Makes sense, updated

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
kaustubhbutte17: Yes I observed this same issue during my tests too, the lag catchup took time -> failed shard init -> shard kept moving to new nodes but failed -> cluster went red. I have now removed the fail on timeout option. We now log a warning and continue into serving state

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
kaustubhbutte17: Removed this setting, user can now set lag threshold as -1 to indicate that warmup is not needed

server/src/main/java/org/opensearch/cluster/metadata/IngestionSource.java
kaustubhbutte17: Thanks, updated

PR #20530, opened 2026-02-02
Adding stream request flag to SearchRequestContext

harshavamsi:
### Description
Adding stream request flag to SearchRequestContext so that other plugins can consume it. Query insights can use this information to emit a dimension on whether search request was powered by streaming

### Related Issues

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 - Move Randomness from server to libs/common ([#20570](https://github.com/opensearch-project/OpenSearch/pull/20570))
 - Use env variable (OPENSEARCH_FIPS_MODE) to enable opensearch to run in FIPS enforced mode instead of checking for existence of bcFIPS jars ([#20625](https://github.com/opensearch-project/OpenSearch/pull/20625))
+- Update streaming flag to use search request context ([#20530](https://github.com/opensearch-project/OpenSearch/pull/20530))
 
 ### Fixed
 - Fix flaky test failures in ShardsLimitAllocationDeciderIT ([#20375](https://github.com/opensearch-project/OpenSearch/pull/20375))


rishabhmaurya:
can't we add it to top level agg builder instead of term agg builder to make it more generic?

harshavamsi:
> can't we add it to top level agg builder instead of term agg builder to make it more generic?

makes sense to me, will do

server/src/main/java/org/opensearch/search/aggregations/AggregationBuilder.java
```diff
@@ -185,6 +185,25 @@ public enum BucketCardinality {
      */
     public abstract BucketCardinality bucketCardinality();
 
+    /**
+     * Enables or disables streaming mode for this aggregation.
+     *
+     * @param enabled whether streaming should be enabled
+     * @return this builder for chaining
+     */
+    public AggregationBuilder streamingEnabled(boolean enabled) {
```
bowenlan-amzn: To add in a class with public api notation, have an experimental api annotation would be safer.

harshavamsi: added

harshavamsi:
> On a high level, streaming could be another dimension to label search request, without combining with term aggregation here. Example: A search request labeled with agg: terms streaming: true query: range ...

I feel that will pollute the way we are able to categorize requests. Is it possible for request to be terms aggregation and streaming enabled but not use stream search because of flush mode thresholds?

bowenlan-amzn:
> > On a high level, streaming could be another dimension to label search request, without combining with term aggregation here. Example: A search request labeled with agg: terms streaming: true query: range ...
> 
> I feel that will pollute the way we are able to categorize requests. Is it possible for request to be terms aggregation and streaming enabled but not use stream search because of flush mode thresholds?

Yes, streaming maybe disabled on data node before shard execution.
You make me realize that an accurate way to distinguish whether an aggregation is being executed using streaming or not is in the coordinator node reduce logic.

https://github.com/opensearch-project/OpenSearch/blob/80c92e95818504b10f58c85efebd5679e3009dcd/server/src/main/java/org/opensearch/action/search/StreamSearchQueryThenFetchAsyncAction.java#L108

Here would be a good point to add the metric for streaming

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregationBuilder.java
```diff
@@ -183,6 +186,9 @@ public TermsAggregationBuilder(StreamInput in) throws IOException {
         includeExclude = in.readOptionalWriteable(IncludeExclude::new);
         order = InternalOrder.Streams.readOrder(in);
         showTermDocCountError = in.readBoolean();
+        if (in.getVersion().onOrAfter(Version.V_3_5_0)) {
```
bowenlan-amzn: If we only going to use this field on coordinator, we don't need to put this in stream input or output

server/src/test/java/org/opensearch/search/streaming/FlushModeResolverTests.java
```diff
@@ -63,4 +66,93 @@ public void testDecideFlushModeSmallTopN() {
         FlushMode result = FlushModeResolver.decideFlushMode(smallTopN, FlushMode.PER_SHARD, 100_000);
         assertEquals(FlushMode.PER_SEGMENT, result);
     }
+
+    public void testIsStreamableDoesNotModifyAggregation() {
```
bowenlan-amzn: Seems no point to test this?

server/src/main/java/org/opensearch/search/streaming/FlushModeResolver.java
```diff
@@ -131,6 +131,21 @@ public static boolean isStreamable(AggregatorFactories.Builder aggregations) {
         return true;
     }
 
+    /**
+     * Enables streaming mode on all top-level aggregations that support it.
```
bowenlan-amzn: This is not enable streaming, just set a flag. Can we remove or improve this doc?

server/src/test/java/org/opensearch/search/streaming/FlushModeResolverTests.java
```diff
@@ -63,4 +66,93 @@ public void testDecideFlushModeSmallTopN() {
...
+    public void testStreamingEnabledDefaultBehaviorForNonStreamingAggregations() {
+        MaxAggregationBuilder maxAgg = new MaxAggregationBuilder("max_value");
+        maxAgg.field("value");
+
+        assertFalse(maxAgg.isStreamingEnabled());
+
+        assertSame(maxAgg, maxAgg.streamingEnabled(true));
+
+        assertFalse(maxAgg.isStreamingEnabled());
+    }
```
bowenlan-amzn: Not sure the point of these tests, can remove.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/TermsAggregationBuilder.java
harshavamsi: Even though we set the flag at the coordinator node, I assume that the builders are passed around on the data nodes?

cc: @rishabhmaurya could you verify?

bowenlan-amzn: Yeah they will be passed around but if data node doesn't rely on this field, we don't need to send through the node-to-node transport.

harshavamsi:
@rishabhmaurya @bowenlan-amzn I updated the PR to be simpler. We don't make a change in the agg builder, instead we add it in the search request context, this way we are ensuring the integrity of the change without polluting the agg builder. Downstream plugins that have access to request context should get the streaming flag

PR #20536, opened 2026-02-04
WLM group custom search settings framework

dzane17:
### Description

This PR adds the foundational infrastructure for defining custom search settings within a WLM group, along with the first setting implementation (`timeout`).

#### What's included

- `WorkloadGroupSearchSettings` enum with validation framework for search settings
- `search_settings` field in `MutableWorkloadGroupFragment` and `WorkloadGroup`
- `WorkloadGroupRequestOperationListener` integration to apply settings during request start
- `timeout` setting implementation

#### Supported setting (this PR)

**`timeout`** - Enforces a hard upper bound on how long a search is allowed to execute, helping keep latency predictable for a workload group and avoiding situations where slow or stalled queries tie up search threads. The WLM timeout is only applied when the request does not already have an explicit timeout set.

### Usage

Create WLM group with `search_settings`:

```bash
curl -XPUT "http://localhost:9200/_wlm/workload_group?pretty" -H 'Content-Type: application/json' -d'
{
  "name": "analytics",
  "resiliency_mode": "enforced",
  "resource_limits": {
    "cpu": 0.1,
    "memory": 0.1
  },
  "search_settings": {
    "timeout": "30s"
  }
}'
```

Update WLM group:
```
curl -XPUT "http://localhost:9200/_wlm/workload_group/analytics?pretty" -H 'Content-Type: application/json' -d'
{
  "search_settings": {
    "timeout": "1m"
  }
}'
```

### Related Issues
Part of https://github.com/opensearch-project/OpenSearch/issues/20555

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Introduced strategy planner interfaces for indexing and deletion ([#20585](https://github.com/opensearch-project/OpenSearch/pull/20585))
 - Implement FieldMappingIngestionMessageMapper for pull-based ingestion ([#20729](https://github.com/opensearch-project/OpenSearch/pull/20729))
 - Added support of WarmerRefreshListener in NRTReplicationEngine to trigger warmer after replication on replica shards ([#20650](https://github.com/opensearch-project/OpenSearch/pull/20650))
+- WLM group custom search settings - groundwork and timeout ([#20536](https://github.com/opensearch-project/OpenSearch/issues/20536))
 
 ### Changed
 - Make telemetry `Tags` immutable ([#20788](https://github.com/opensearch-project/OpenSearch/pull/20788))
--- a/server/src/main/java/org/opensearch/wlm/MutableWorkloadGroupFragment.java
+++ b/server/src/main/java/org/opensearch/wlm/MutableWorkloadGroupFragment.java
@@ -8,6 +8,7 @@
 
 package org.opensearch.wlm;
 
+import org.opensearch.Version;
 import org.opensearch.cluster.AbstractDiffable;
 import org.opensearch.common.annotation.ExperimentalApi;
 import org.opensearch.core.common.io.stream.StreamInput;
@@ -18,9 +19,11 @@ import org.opensearch.core.xcontent.XContentParser;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.TreeMap;
 import java.util.function.Function;
 
 /**
@@ -31,17 +34,32 @@ public class MutableWorkloadGroupFragment extends AbstractDiffable<MutableWorklo
 
     public static final String RESILIENCY_MODE_STRING = "resiliency_mode";
     public static final String RESOURCE_LIMITS_STRING = "resource_limits";
+    public static final String SEARCH_SETTINGS_STRING = "search_settings";
     private ResiliencyMode resiliencyMode;
     private Map<ResourceType, Double> resourceLimits;
+    private Map<String, String> searchSettings;
 
-    public static final List<String> acceptedFieldNames = List.of(RESILIENCY_MODE_STRING, RESOURCE_LIMITS_STRING);
+    public static final List<String> acceptedFieldNames = List.of(RESILIENCY_MODE_STRING, RESOURCE_LIMITS_STRING, SEARCH_SETTINGS_STRING);
 
     public MutableWorkloadGroupFragment() {}
 
+    /**
+     * Constructor for tests only. Production code should use the full constructor below.
+     */
     public MutableWorkloadGroupFragment(ResiliencyMode resiliencyMode, Map<ResourceType, Double> resourceLimits) {
+        this(resiliencyMode, resourceLimits, new HashMap<>());
+    }
+
+    public MutableWorkloadGroupFragment(
+        ResiliencyMode resiliencyMode,
+        Map<ResourceType, Double> resourceLimits,
+        Map<String, String> searchSettings
+    ) {
         validateResourceLimits(resourceLimits);
+        WorkloadGroupSearchSettings.validateSearchSettings(searchSettings);
         this.resiliencyMode = resiliencyMode;
         this.resourceLimits = resourceLimits;
+        this.searchSettings = searchSettings;
     }
 
     public MutableWorkloadGroupFragment(StreamInput in) throws IOException {
@@ -52,6 +70,13 @@ public class MutableWorkloadGroupFragment extends AbstractDiffable<MutableWorklo
         }
         String updatedResiliencyMode = in.readOptionalString();
         resiliencyMode = updatedResiliencyMode == null ? null : ResiliencyMode.fromName(updatedResiliencyMode);
+        if (in.getVersion().onOrAfter(Version.V_3_6_0)) {
+            // Read null marker: true means searchSettings is null (not specified)
+            boolean isNull = in.readBoolean();
+            searchSettings = isNull ? null : in.readMap(StreamInput::readString, StreamInput::readString);
+        } else {
+            searchSettings = new HashMap<>();
+        }
     }
 
     interface FieldParser<T> {
@@ -80,14 +105,20 @@ public class MutableWorkloadGroupFragment extends AbstractDiffable<MutableWorklo
         }
     }
 
+    static class SearchSettingsParser implements FieldParser<Map<String, String>> {
+        public Map<String, String> parseField(XContentParser parser) throws IOException {
+            return parser.mapStrings();
+        }
+    }
+
     static class FieldParserFactory {
         static Optional<FieldParser<?>> fieldParserFor(String fieldName) {
-            if (fieldName.equals(RESOURCE_LIMITS_STRING)) {
-                return Optional.of(new ResourceLimitsParser());
-            } else if (fieldName.equals(RESILIENCY_MODE_STRING)) {
-                return Optional.of(new ResiliencyModeParser());
-            }
-            return Optional.empty();
+            return switch (fieldName) {
+                case RESILIENCY_MODE_STRING -> Optional.of(new ResiliencyModeParser());
+                case RESOURCE_LIMITS_STRING -> Optional.of(new ResourceLimitsParser());
+                case SEARCH_SETTINGS_STRING -> Optional.of(new SearchSettingsParser());
+                default -> Optional.empty();
+            };
         }
     }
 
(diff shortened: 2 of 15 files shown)


server/src/main/java/org/opensearch/wlm/listeners/WorkloadGroupRequestOperationListener.java
```diff
@@ -36,11 +56,112 @@ public WorkloadGroupRequestOperationListener(WorkloadGroupService workloadGroupS
...
+                        if (requestTimeout == null) {
+                            searchRequest.source().timeout(wlmTimeout);
+                        } else {
+                            searchRequest.source().timeout(requestTimeout.millis() <= wlmTimeout.millis() ? requestTimeout : wlmTimeout);
+                        }
+                        break;
+                    case MAX_BUCKET:
+                        // Store max_bucket as a request header so it propagates to data nodes
+                        threadPool.getThreadContext().putHeader(MAX_BUCKET.getSettingName(), entry.getValue());
+                        break;
```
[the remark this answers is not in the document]
dzane17: There is no request-provided `max_buckets`. `max_buckets` is only a cluster setting

kaushalmahi12:
@dzane17 ! thanks for raising this PR it would be super helpful to have the control over these settings at much granular level.
But can you please try to break this PR into 2 or more ?

dzane17:
@kaushalmahi12 Done, this 1st PR is ready now

server/src/main/java/org/opensearch/wlm/listeners/WorkloadGroupRequestOperationListener.java
```diff
@@ -36,11 +52,56 @@ public WorkloadGroupRequestOperationListener(WorkloadGroupService workloadGroupS
...
+        for (Map.Entry<String, String> entry : workloadGroup.getSearchSettings().entrySet()) {
+            try {
+                WorkloadGroupSearchSettings.WlmSearchSetting settingKey = WorkloadGroupSearchSettings.WlmSearchSetting.fromKey(
+                    entry.getKey()
+                );
+                if (settingKey == null) continue;
+
+                switch (settingKey) {
+                    case PHASE_TOOK:
+                        searchRequest.setPhaseTook(Boolean.parseBoolean(entry.getValue()));
```
jainankitk: What if the search request overrides the phase_took setting? That should take precedence over workload group right?

dzane17: Based on discussions with @kaushalmahi12 and @kkhatua, we decided that WLM settings should take precedence over per-request settings. This is because WLM groups are configured by cluster administrators, whereas per-request parameters are set by individual users. Allowing per-request values to override WLM settings would effectively turn this feature into mere “recommendations” rather than authoritative controls.

While this approach means that WLM settings cannot be overridden, it aligns with the intended behavior of a strict resiliency mechanism. If an administrator wants individual users to have full control over certain search parameters, they can simply omit those settings when configuring their WLM group.

server/src/main/java/org/opensearch/action/search/SearchRequest.java
```diff
@@ -649,15 +649,20 @@ public int getMaxConcurrentShardRequests() {
         return maxConcurrentShardRequests == 0 ? 5 : maxConcurrentShardRequests;
     }
 
+    /**
+     * Returns the raw internal value of maxConcurrentShardRequests without applying the default.
```
kaushalmahi12: Is this value extracted and set from workload group into search request ?

server/src/main/java/org/opensearch/wlm/listeners/WorkloadGroupRequestOperationListener.java
```diff
@@ -8,24 +8,40 @@
...
 
-    public WorkloadGroupRequestOperationListener(WorkloadGroupService workloadGroupService, ThreadPool threadPool) {
+    public WorkloadGroupRequestOperationListener(
+        WorkloadGroupService workloadGroupService,
+        ThreadPool threadPool,
+        ClusterService clusterService
+    ) {
         this.workloadGroupService = workloadGroupService;
         this.threadPool = threadPool;
+        this.clusterService = clusterService;
```
kaushalmahi12: WorkloadGroupService also keeps the active workload groups. So this additional dependency on ClusterService might be redundant

server/src/main/java/org/opensearch/wlm/MutableWorkloadGroupFragment.java
```diff
@@ -111,23 +142,39 @@ static Optional<FieldParser<?>> fieldParserFor(String fieldName) {
...
-                } else if (field.equals(RESOURCE_LIMITS_STRING)) {
-                    setResourceLimits((Map<ResourceType, Double>) value);
+                switch (field) {
+                    case RESILIENCY_MODE_STRING -> setResiliencyMode((ResiliencyMode) value);
+                    case RESOURCE_LIMITS_STRING -> setResourceLimits((Map<ResourceType, Double>) value);
+                    case SEARCH_SETTINGS_STRING -> setSearchSettings((Map<String, String>) value);
                 }
             } catch (IOException e) {
-                throw new IllegalArgumentException("parsing error encountered for the field " + field);
+                throw new IllegalArgumentException(String.format(Locale.ROOT, "parsing error encountered for the field '%s'", field));
```
kaushalmahi12, line 175: Thanks for fixing this.

server/src/main/java/org/opensearch/wlm/WorkloadGroupSearchSettings.java
```diff
@@ -0,0 +1,107 @@
...
+
+        /**
+         * Validates the given value for this setting.
+         * @param value the value to validate
+         * @throws IllegalArgumentException if the value is invalid
+         */
+        void validate(String value) {
+            String error = validator.apply(value);
+            if (error != null) {
+                throw new IllegalArgumentException("Invalid value for " + settingName + ": " + error);
```
kaushalmahi12: lets also add the invalid value in the exception msg

server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
```diff
@@ -476,6 +473,11 @@ void executeRequest(
             );
             searchRequestContext.getSearchRequestOperationsListener().onRequestStart(searchRequestContext);
 
+            // Check phase_took cluster setting AFTER wlm search setting is optionally added in onRequestStart
+            if (originalSearchRequest.isPhaseTook() == null) {
+                originalSearchRequest.setPhaseTook(clusterService.getClusterSettings().get(SEARCH_PHASE_TOOK_ENABLED));
+            }
+
```
kaushalmahi12: Why not move this to `SearchRequestOperationsListener#onRequestStart` hook, that seems relevant since this value setting logic is in one place

server/src/main/java/org/opensearch/wlm/listeners/WorkloadGroupRequestOperationListener.java
dzane17: Good point. Fixed it

server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
dzane17: These lines set `phase_took` when the corresponding cluster setting is enabled. I’m just moving the block down so that precedence is correct.

Listeners that override `onRequestStart` have their own business logic, and they might not be enabled at all. It is safer to keep this logic here to ensure `phase_took` is always set correctly.

server/src/main/java/org/opensearch/action/search/SearchRequest.java
dzane17: This method is not used in this PR. I'll remove it now and add it in the next PR when max concurrent shards is onboarded.

server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
kaushalmahi12: Is the WLM OPerational listener conditionally added ?

kaushalmahi12 (approved):
Lets resolve the merge conflict, rest LGTM!

server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
dzane17: The WLM listener is always enabled right now. It would convenient to resolve the setting completely in `WorkloadGroupRequestOperationListener.onRequestStart` but I feel that would improperly introduce a dependency of core search functionality on WLM.

server/src/main/java/org/opensearch/wlm/listeners/WorkloadGroupRequestOperationListener.java
jainankitk: This seems like a concern to me, especially since we are changing the default behavior as of today. Not sure what is the right experience, but silently ignoring request level override looks incorrect to me

server/src/main/java/org/opensearch/wlm/MutableWorkloadGroupFragment.java
```diff
@@ -52,6 +70,13 @@ public MutableWorkloadGroupFragment(StreamInput in) throws IOException {
         }
         String updatedResiliencyMode = in.readOptionalString();
         resiliencyMode = updatedResiliencyMode == null ? null : ResiliencyMode.fromName(updatedResiliencyMode);
+        if (in.getVersion().onOrAfter(Version.V_3_6_0)) {
+            // Read null marker: true means searchSettings is null (not specified)
+            boolean isNull = in.readBoolean();
+            searchSettings = isNull ? null : in.readMap(StreamInput::readString, StreamInput::readString);
```
jainankitk, line 76: Okay, now we are defining 3 stats of this `searchSettings` map - null, empty, filled. I am wondering if there are any scenarios where null is getting exercised? `searchSettings` is always non-null as the constructor initializes to empty map right?

```
builder.startObject(SEARCH_SETTINGS_STRING);
            Map<String, String> settings = searchSettings != null ? searchSettings : Map.of();
            Map<String, String> sortedSettingsMap = new TreeMap<>(settings);
            for (Map.Entry<String, ?> e : sortedSettingsMap.entrySet()) {
                builder.field(e.getKey(), e.getValue());
            }
```

WLM_SEARCH_SETTINGS_LLD.md
```diff
@@ -0,0 +1,463 @@
...
+## Supported Settings
+
+| Setting | Type | Description | Conflict Resolution |
+|---------|------|-------------|---------------------|
+| `phase_took` | boolean | Include phase timing in response | WLM always overrides |
+| `cancel_after_time_interval` | TimeValue | Cancel search after interval | min(WLM, request) |
+| `timeout` | TimeValue | Search request timeout | min(WLM, request) |
+| `max_concurrent_shard_requests` | int | Max concurrent shard requests | min(WLM, request) or WLM if unset |
+| `batched_reduce_size` | int (≥2) | Batched reduce size | min(WLM, request) |
+| `max_buckets` | int | Max aggregation buckets | WLM overrides via header |
```
jainankitk: Should we be checking in the spec for LLM?

search_stats_negative_counter_analysis.md
```diff
@@ -0,0 +1,152 @@
+# SearchStats Negative Counter Bug — Root Cause Analysis
+
+## Summary
+
+The `current` counter in `SearchStats.PhaseStatsLongHolder` goes negative due to
+double-decrements in the request-level phase lifecycle. The workaround on lines 91-92
+of `SearchStats.java` clamps negative values to 0 during serialization, but the
+underlying bugs remain.
+
+## Architecture Context
```
jainankitk: you probably need to add rule for skipping the markdown specification files in your global .gitignore configuration

CHANGELOG.md
```diff
@@ -38,6 +38,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Introduced strategy planner interfaces for indexing and deletion ([#20585](https://github.com/opensearch-project/OpenSearch/pull/20585))
 - Implement FieldMappingIngestionMessageMapper for pull-based ingestion ([#20729](https://github.com/opensearch-project/OpenSearch/pull/20729))
 - Added support of WarmerRefreshListener in NRTReplicationEngine to trigger warmer after replication on replica shards ([#20650](https://github.com/opensearch-project/OpenSearch/pull/20650))
+- WLM group custom search settings - groundwork and timeout ([#20536](https://github.com/opensearch-project/OpenSearch/issues/20536))
+- Add security policy to allow `accessUnixDomainSocket` in `transport-grpc` module ([#20463](https://github.com/opensearch-project/OpenSearch/pull/20463))
```
jainankitk: Seems like the merge conflict was not resolved properly

server/src/main/java/org/opensearch/wlm/MutableWorkloadGroupFragment.java
dzane17: There is one case where `searchSettings` can be null. That is during a WLM group update request. User may want to update only `resource_limits` in an existing group like: 
```
% curl -XPUT "http://localhost:9200/_wlm/workload_group/analytics?pretty" -H 'Content-Type: application/json' -d'
{               
  "resource_limits": {
    "cpu": 0.2,
    "memory": 0.4
  }
}'
```

There needs to be a way to distinguish the above request body (null searchSettings) vs below (explicit empty searchSettings)
```
{               
  "resource_limits": {
    "cpu": 0.2,
    "memory": 0.4
  },
  "search_settings": {}
}'
```

WLM_SEARCH_SETTINGS_LLD.md
dzane17: This is an unnecessary design doc. Removed it

search_stats_negative_counter_analysis.md
dzane17: Deleted this file

CHANGELOG.md
dzane17: Resolved conflict

jainankitk (approved):
Can you add few more tests to increase the code coverage?

PR #20542, opened 2026-02-04
Fix copy_to functionality for geo_point fields with object/array values

### LINKED ISSUE #20540: [BUG] CopyTo does not work with geo_point field in a specific document.
### Describe the bug

When "CopyTo" is set for a geo_point field, an error occurs if the input document is in a specific format. This is a similar issue to [knn issue #2636](https://github.com/opensearch-project/k-NN/issues/2636) and is caused by values being defined across multiple tokens.

### Related component

Indexing

### To Reproduce

Reproducing the issue 
1. Define an index
```
PUT testindex1
{
  "mappings": {
    "properties": {
      "point": {
        "type": "geo_point",
        "copy_to": ["point_copied"]
      },
      "point_copied": {
        "type": "geo_point"
      }
    }
  }
}
```
2. Put a document with object format.
```
PUT testindex2/_doc/1
{
  "point": { 
    "lat": 40.71,
    "lon": 74.00
  }
}
```
An error occurs.
```
{
  "error": {
    "root_cause": [
      {
        "type": "parse_exception",
        "reason": "geo_point expected"
      }
    ],
    "type": "mapper_parsing_exception",
    "reason": "failed to parse field [point_copied] of type [geo_point]",
    "caused_by": {
      "type": "parse_exception",
      "reason": "geo_point expected"
    }
  },
  "status": 400
}
```

### Expected behavior

Reproducing the issue 
1. Define an index
```
PUT testindex1
{
  "mappings": {
    "properties": {
      "point": {
        "type": "geo_point",
        "copy_to": ["point_copied"]
      },
      "point_copied": {
        "type": "geo_point"
      }
    }
  }
}
```
2. Put a document with a format with single token.
```
PUT testindex1/_doc/1
{
  "point": "40.71,74.00" 
}
```
3. No errors occur.
```
{
  "_index": "testindex1",
  "_id": "1",
  "_version": 1,
  "result": "created",
  "_shards": {
    "total": 2,
    "successful": 2,
    "failed": 0
  },
  "_seq_no": 0,
  "_primary_term": 1
}
```

krocky-cooky:
### Description

This PR adds support for the `copy_to` attribute on `geo_point` fields when they are represented as complex structures (objects or arrays).

#### Problem

Previously, when a `geo_point` field with a `copy_to` attribute was indexed using object notation (e.g., `{"lat": 40.71, "lon": 74.00}`) or array notation (e.g., `[74.00, 40.71]`), the copy operation would fail. This occurred because the XContent parser consumed all tokens during the initial field parsing, leaving no data available for the subsequent `copy_to` operation. This is a similar issue to[ k-NN Plugin issue #2536](https://github.com/opensearch-project/k-NN/issues/2636) and is caused by values being defined across multiple tokens.

#### Solution

This PR introduces a mechanism to preserve the field value for reuse:

1. Serialize complex values to bytes: When a field value is an object or array, the parser state is serialized to a byte array before parsing begins.
2. Reparse for copy_to fields: The byte array is used to create fresh parser instances for each `copy_to` target field, allowing the same data to be parsed multiple times.

#### Changes

- Added `parseFieldWithCopyTo()` method to handle field parsing with `copy_to` support for both simple and complex value types
- Added `parseChildToBytes()` helper method to serialize the current parser state to a byte array
- Extended `parseCopyFields()` to accept an optional byte array parameter for reparsing complex structures
- Added test cases for `geo_point` fields with `copy_to` in both object and array formats

#### Testing

- Added `testGeoPointWithCopyTo()` to verify object notation (`{"lat": ..., "lon": ...}`)
- Added `testGeoPointArrayWithCopyTo()` to verify array notation (`[lon, lat]`)
- All existing tests continue to pass (124 tests total in `DocumentParserTests`)

### Related Issues
Resolves #20540

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
+++ b/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
@@ -37,17 +37,17 @@ import org.apache.lucene.index.IndexableField;
 import org.opensearch.OpenSearchParseException;
 import org.opensearch.Version;
 import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.common.CheckedBiConsumer;
 import org.opensearch.common.collect.Tuple;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.time.DateFormatter;
 import org.opensearch.common.xcontent.LoggingDeprecationHandler;
 import org.opensearch.common.xcontent.XContentHelper;
-import org.opensearch.common.CheckedBiConsumer;
 import org.opensearch.core.common.Strings;
-import org.opensearch.core.xcontent.MediaType;
-import org.opensearch.core.xcontent.XContentParser;
-import org.opensearch.core.xcontent.XContentBuilder;
 import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.xcontent.MediaType;
+import org.opensearch.core.xcontent.XContentBuilder;
+import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.index.IndexSettings;
 import org.opensearch.index.mapper.DynamicTemplate.XContentFieldType;
 import org.opensearch.script.ContextAwareGroupingScript;
@@ -1556,7 +1556,7 @@ final class DocumentParser {
 
     private static byte[] parseChildToBytes(ParseContext context) throws IOException {
         XContentParser parser = context.parser();
-        try(XContentBuilder builder = XContentBuilder.builder(parser.contentType().xContent())) {
+        try (XContentBuilder builder = XContentBuilder.builder(parser.contentType().xContent())) {
             builder.copyCurrentStructure(parser);
             return BytesReference.toBytes(BytesReference.bytes(builder));
         }
@@ -1564,9 +1564,8 @@ final class DocumentParser {
 
     private static void parseFieldWithCopyTo(ParseContext context, FieldMapper fieldMapper) throws IOException {
         XContentParser.Token token = context.parser().currentToken();
-        if ((token == XContentParser.Token.START_ARRAY || 
-            token == XContentParser.Token.START_OBJECT) && 
-            !fieldMapper.copyTo().copyToFields().isEmpty()) {
+        if ((token == XContentParser.Token.START_ARRAY || token == XContentParser.Token.START_OBJECT)
+            && !fieldMapper.copyTo().copyToFields().isEmpty()) {
             byte[] childBytes = parseChildToBytes(context);
             // After parseChildToBytes, the original parser has consumed the full structure.
             // Parse the field using a fresh parser over the captured bytes.
@@ -1574,9 +1573,7 @@ final class DocumentParser {
             try (
                 XContentParser innerParser = parser.contentType()
                     .xContent()
-                    .createParser(parser.getXContentRegistry(),
-                        parser.getDeprecationHandler(),
-                        childBytes)
+                    .createParser(parser.getXContentRegistry(), parser.getDeprecationHandler(), childBytes)
             ) {
                 innerParser.nextToken();
                 ParseContext innerContext = context.switchParser(innerParser);
--- a/server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
+++ b/server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
@@ -3517,16 +3517,11 @@ public class DocumentParserTests extends MapperServiceTestCase {
             b.startObject("location_copy").field("type", "geo_point").endObject();
         }));
 
-        ParsedDocument doc = mapper.parse(source(b -> {
-            b.startObject("point")
-                .field("lat", 40.71)
-                .field("lon", 74.00)
-            .endObject();
-        }));
+        ParsedDocument doc = mapper.parse(source(b -> { b.startObject("point").field("lat", 40.71).field("lon", 74.00).endObject(); }));
 
         // Verify that the geo_point field itself exists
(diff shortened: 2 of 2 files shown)


hasnain2808:
Hi @krocky-cooky 
I was trying to debug https://github.com/opensearch-project/k-NN/issues/2636 when I found this PR
are you actively working on this PR/ issue?

krocky-cooky:
Thanks for the comment.
This PR doesn't include test cases for vector_field, but merging it should also resolve issue opensearch-project/k-NN#2636. It's currently awaiting review from the maintainers. 
Did you have a different implementation plan in mind?

It looks like the issue was recently triaged. 
@andrross Could you take a look at this when you get a chance? Happy to make any changes as needed.

hasnain2808:
> Thanks for the comment.
> 
> This PR doesn't include test cases for vector_field, but merging it should also resolve issue opensearch-project/k-NN#2636. It's currently awaiting review from the maintainers. 
> 
> Did you have a different implementation plan in mind?
> 
> 
> 
> It looks like the issue was recently triaged. 
> 
> @andrross Could you take a look at this when you get a chance? Happy to make any changes as needed.

My solution was going in a  similar direction and imo the same solution will work for knn too. I was using geo point with arrays as test

We need a maintainer to enable ci on this pr too

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -1339,6 +1339,44 @@ private static void parseDynamicValue(
     }
 
     /** Creates instances of the fields that the current field should be copied to */
+    private static void parseCopyFields(ParseContext context, List<String> copyToFields, byte[] sourceBytes) throws IOException {
```
andrross, line 1343: Can you avoid some duplication here with a helper like the following?

```java
private static void parseCopyFields(ParseContext context, List<String> copyToFields) throws IOException {
    parseCopyFieldsInternal(context, copyToFields, (copyToContext, field) -> parseCopy(field, copyToContext));
}

private static void parseCopyFields(ParseContext context, List<String> copyToFields, byte[] sourceBytes) throws IOException {
    parseCopyFieldsInternal(context, copyToFields, (copyToContext, field) -> {
        XContentParser parser = copyToContext.parser();
        try (
            XContentParser innerParser = parser.contentType()
                .xContent()
                .createParser(parser.getXContentRegistry(), parser.getDeprecationHandler(), sourceBytes)
        ) {
            innerParser.nextToken();
            parseCopy(field, copyToContext.switchParser(innerParser));
        }
    });
}

private static void parseCopyFieldsInternal(
    ParseContext context,
    List<String> copyToFields,
    CheckedBiConsumer<ParseContext, String, IOException> copyAction
) throws IOException {
    if (!context.isWithinCopyTo() && copyToFields.isEmpty() == false) {
        context = context.createCopyToContext();
        for (String field : copyToFields) {
            ParseContext.Document targetDoc = null;
            for (ParseContext.Document doc = context.doc(); doc != null; doc = doc.getParent()) {
                if (field.startsWith(doc.getPrefix())) {
                    targetDoc = doc;
                    break;
                }
            }
            assert targetDoc != null;
            ParseContext copyToContext;
            if (targetDoc == context.doc()) {
                copyToContext = context;
            } else {
                copyToContext = context.switchDoc(targetDoc);
            }
            copyAction.accept(copyToContext, field);
        }
    }
}
```

server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
```diff
@@ -3505,4 +3505,82 @@ public void testParseNonDynamicArrayWithNullValuesDisableObjects() throws Except
...
+        // Verify that the geo_point field itself exists
+        assertNotNull(doc.rootDoc().getField("point"));
+        
+        // Verify that the field was copied to the copy_to target
+        IndexableField[] copiedFields = doc.rootDoc().getFields("location_copy");
+        assertNotNull(copiedFields);
+        assertTrue(copiedFields.length > 0);
+    }
+
+    public void testGetPointArrayWithMultipleCopyTo() throws Exception {
```
andrross:
```suggestion
    public void testGeoPointArrayWithMultipleCopyTo() throws Exception {
```

krocky-cooky:
@andrross 
Thank you so much for the review. I've addressed all the feedback.

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -1339,7 +1340,30 @@ private static void parseDynamicValue(
...
+        parseCopyFieldsInternal(context, copyToFields, (copyToContext, field) -> {
+            XContentParser parser = copyToContext.parser();
+            try (
+                XContentParser innerParser = parser.contentType()
+                    .xContent()
+                    .createParser(parser.getXContentRegistry(), parser.getDeprecationHandler(), sourceBytes)
+            ) {
+                innerParser.nextToken();
+                copyToContext = copyToContext.switchParser(innerParser);
+                parseCopy(field, copyToContext);
```
andrross: Parameter re-assignment is generally not a good idea. Better to create a separate local variable.

```suggestion
                ParseContext switched = copyToContext.switchParser(innerParser);
                parseCopy(field, switched);
```

andrross (commented):
Looks like you have a spotless formatting error. Can you fix that and rebase from main?

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
krocky-cooky: fixed!

andrross (approved):

Issue #20550, opened 2026-02-05
[BUG] Segment replication continuously fails due to stale metadata checkpoint


guojialiang92:
### Describe the bug

I have observed that in a cluster with **local segment replication** enabled, there may be a phenomenon where the primary shard is never able to synchronize segments to replicas. After a period of time, since the translog in the replica can never be garbage collected, it will cause the disk space of the replica node to be exhausted.

After troubleshooting, I found that before the first occurrence of the "[`Rejecting stale metadata checkpoint`](https://github.com/opensearch-project/OpenSearch/blob/ea01bb5fc3b7f54b42c6eb4dc6354a764c06978b/server/src/main/java/org/opensearch/indices/replication/AbstractSegmentReplicationTarget.java#L174)" log, the segment replication process of the replica shard had failed due to the `CircuitBreakingException` of the node where the primary shard was located.

1. `CircuitBreakingException` stack

```
[2026-01-29T12:57:49,641][ERROR][o.o.i.r.SegmentReplicationTargetService] [byte-es-siem-edr-prd-data-34] [shardId 2] [replication id 526476] Replication failed, timing data
: {INIT=0, REPLICATING=0}
org.opensearch.indices.replication.common.ReplicationFailedException: Segment Replication failed
        at org.opensearch.indices.replication.SegmentReplicator$2.onFailure(SegmentReplicator.java:206) 
        at org.opensearch.action.ActionListener$1.onFailure(ActionListener.java:88)
        at org.opensearch.action.ActionRunnable.onFailure(ActionRunnable.java:103) 
        at org.opensearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:54) 
        at org.opensearch.common.util.concurrent.OpenSearchExecutors$DirectExecutorService.execute(OpenSearchExecutors.java:348) 
        at org.opensearch.common.util.concurrent.ListenableFuture.notifyListener(ListenableFuture.java:120) 
        at org.opensearch.common.util.concurrent.ListenableFuture.lambda$done$0(ListenableFuture.java:112) 
        at java.util.ArrayList.forEach(ArrayList.java:1511) 
        at org.opensearch.common.util.concurrent.ListenableFuture.done(ListenableFuture.java:112) 
        at org.opensearch.common.util.concurrent.BaseFuture.setException(BaseFuture.java:178) 
        at org.opensearch.common.util.concurrent.ListenableFuture.onFailure(ListenableFuture.java:149) 
[...]

PR #20551, opened 2026-02-05
[segment replication] Fix segment replication infinite retry due to stale metadata checkpoint

guojialiang92:
### Description
This PR is to address the issues in #[[20550](https://github.com/opensearch-project/OpenSearch/issues/20550)].

In the case of [failure retry](https://github.com/opensearch-project/OpenSearch/blob/ea01bb5fc3b7f54b42c6eb4dc6354a764c06978b/server/src/main/java/org/opensearch/indices/replication/SegmentReplicationTargetService.java#L369), no [ckp verification](https://github.com/opensearch-project/OpenSearch/blob/ea01bb5fc3b7f54b42c6eb4dc6354a764c06978b/server/src/main/java/org/opensearch/indices/replication/AbstractSegmentReplicationTarget.java#L171) is performed. As long as the process enters Phase `GET_SEGMENT_FILES` during the retry, [the primary shard will clear the residual segment replication information](https://github.com/opensearch-project/OpenSearch/blob/ea01bb5fc3b7f54b42c6eb4dc6354a764c06978b/server/src/main/java/org/opensearch/indices/replication/OngoingSegmentReplications.java#L78) due to [IndexShardClosedException caused by the replica](https://github.com/opensearch-project/OpenSearch/blob/ea01bb5fc3b7f54b42c6eb4dc6354a764c06978b/server/src/main/java/org/opensearch/indices/replication/common/ReplicationCollection.java#L183).

### Related Issues
Resolves #[[20550](https://github.com/opensearch-project/OpenSearch/issues/20550)]

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Prevent criteria update for context aware indices ([#20250](https://github.com/opensearch-project/OpenSearch/pull/20250))
 - Update EncryptedBlobContainer to adhere limits while listing blobs in specific sort order if wrapped blob container supports ([#20514](https://github.com/opensearch-project/OpenSearch/pull/20514))
 - [segment replication] Fix segment replication infinite retry due to stale metadata checkpoint ([#20551](https://github.com/opensearch-project/OpenSearch/pull/20551))
+- Changing opensearch.cgroups.hierarchy.override causes java.lang.SecurityException exception ([#20565](https://github.com/opensearch-project/OpenSearch/pull/20565))
 
 ### Dependencies
 - Bump `ch.qos.logback:logback-core` and `ch.qos.logback:logback-classic` from 1.5.24 to 1.5.27 ([#20525](https://github.com/opensearch-project/OpenSearch/pull/20525))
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/opensearch/indices/replication/SegmentReplicationTarget.java
```diff
@@ -128,6 +138,6 @@ protected void finalizeReplication(CheckpointInfoResponse checkpointInfoResponse
 
     @Override
     public SegmentReplicationTarget retryCopy() {
-        return new SegmentReplicationTarget(indexShard, checkpoint, source, listener);
+        return new SegmentReplicationTarget(indexShard, checkpoint, source, true, listener);
```
atris: This misses the isRetry flag. Any flow using retryCopy (like `ReplicationCollection#reset`) will drop the state and could regress back into the infinite loop.

Please update to:

`return new SegmentReplicationTarget(indexShard, checkpoint, source, isRetry, listener);`

server/src/internalClusterTest/java/org/opensearch/indices/replication/SegmentReplicationIT.java
```diff
@@ -140,6 +143,81 @@ private static String indexOrAlias() {
         return randomBoolean() ? INDEX_NAME : "alias";
     }
 
+    public void testSegmentReplicationWithException() throws Exception {
```
atris: This test asserts that retries happen, but because the checkpoints match, it passes even without the fix. Please add a case that mocks a Stale Checkpoint response to verify the bypass logic is actually engaged.

guojialiang92: Thanks @atris. I modified it to execute `latch2.countDown();` after `handler.messageReceived(request, channel, task);` to ensure that the replica has received the new ckp, so that the issue can be reproduced stably.

server/src/main/java/org/opensearch/indices/replication/SegmentReplicationTarget.java
guojialiang92: done

Issue #20522, opened 2026-02-02
[BUG] (v3.2.0) changing opensearch.cgroups.hierarchy.override causes java.lang.SecurityException exception


huvalk:
### Describe the bug

By running the OS in the container, creating a sub сgroup and changing opensearch.cgroups.hierarchy.override to its path results in an exception
```
[2026-02-02T09:08:53,379][WARN ][o.o.m.o.OsProbe          ] [test-cluster-z502-1.es-test-cluster.svc.stg-clickhouse01-z502.k8s.o3.ru] exception retrieving free physical memory
java.lang.reflect.InvocationTargetException: null
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:119) ~[?:?]
        at java.base/java.lang.reflect.Method.invoke(Method.java:565) ~[?:?]
        at org.opensearch.monitor.os.OsProbe.getFreePhysicalMemorySize(OsProbe.java:109) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.monitor.os.OsProbe.osStats(OsProbe.java:693) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.monitor.os.OsService.<init>(OsService.java:71) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.monitor.MonitorService.<init>(MonitorService.java:62) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.node.Node.<init>(Node.java:1028) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.node.Node.<init>(Node.java:482) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.bootstrap.Bootstrap$5.<init>(Bootstrap.java:249) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.bootstrap.Bootstrap.setup(Bootstrap.java:249) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.bootstrap.Bootstrap.init(Bootstrap.java:411) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.bootstrap.OpenSearch.init(OpenSearch.java:168) [opensearch-3.2.0.jar:3.2.0]
        at org.opensearch.bootstrap.OpenSearch.execute(OpenSearch.java:159) [opensearch-3.2.0.jar:3.2.0]
[...]

PR #20565, opened 2026-02-06
Changing opensearch.cgroups.hierarchy.override causes java.lang.SecurityException exception

reta:
### Description
At the moment, the permissions we grant by default apply to root of `cgroup` the hierarchy:

```
  permission java.io.FilePermission "/sys/fs/cgroup/memory.swap.max", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/memory.swap.current", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/memory.max", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/memory.current", "read";
```

The `opensearch.cgroups.hierarchy.override` may change that and in general, the permissions become dynamic, based on the `<cgroup name>` name (whatever `opensearch.cgroups.hierarchy.override` points to):

```
  permission java.io.FilePermission "/sys/fs/cgroup/<cgroup name>/memory.swap.max", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/<cgroup name>/memory.swap.current", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/<cgroup name>/memory.max", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/<cgroup name>/memory.current", "read";
```

The suggested fix it to support `opensearch.cgroups.hierarchy.override` expansion in policy files, so for example `opensearch.cgroups.hierarchy.override=/test` would translate to:

```
  permission java.io.FilePermission "/sys/fs/cgroup/test/memory.swap.max", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/test/memory.swap.current", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/test/memory.max", "read";
  permission java.io.FilePermission "/sys/fs/cgroup/test/memory.current", "read";
```

How to test / reproduce:
1.  `sudo mkdir /sys/fs/cgroup/test`
2. `sudo chmod --recursive o+rw /sys/fs/cgroup/test`
3. `OPENSEARCH_JAVA_OPTS="-Dopensearch.cgroups.hierarchy.override=/test" cgexec -g memory:test ./bin/opensearch`

### Related Issues
Resolves  https://github.com/opensearch-project/OpenSearch/issues/20522

### Check List

### DIFF
--- a/libs/agent-sm/agent-policy/src/main/java/org/opensearch/secure_sm/policy/PolicyFile.java
+++ b/libs/agent-sm/agent-policy/src/main/java/org/opensearch/secure_sm/policy/PolicyFile.java
@@ -143,7 +143,7 @@ public class PolicyFile extends java.security.Policy {
      * Expands additional system property placeholders with arbitrary position and value semantics (like
      * cgroup fe)
      */
-    private static String expandSystemProperty(final String property, String separator, final String value) {
+    private static String expandKnownSystemProperty(final String property, String separator, final String value) {
         final int index = value.indexOf(separator + "${" + property + "}" + separator);
         String path = System.getProperty(property);
 
@@ -175,7 +175,7 @@ public class PolicyFile extends java.security.Policy {
         } else if (pe.name().contains("${opensearch.cgroups.hierarchy.override}")) {
             return new PermissionEntry(
                 pe.permission(),
-                expandSystemProperty("opensearch.cgroups.hierarchy.override", "/", pe.name()),
+                expandKnownSystemProperty("opensearch.cgroups.hierarchy.override", "/", pe.name()),
                 pe.action()
             );
         }


libs/agent-sm/agent-policy/src/test/java/org/opensearch/secure_sm/policy/PolicyFileTests.java
```diff
@@ -0,0 +1,37 @@
...
+        final PermissionCollection permissions = policy.getPermissions(getClass().getProtectionDomain());
+
+        final List<FilePermission> filePermissions = permissions.elementsAsStream()
+            .filter(FilePermission.class::isInstance)
+            .map(FilePermission.class::cast)
+            .toList();
+        assertThat(filePermissions, hasSize(3));
+
+        assertThat(filePermissions.stream().filter(p -> p.getName().equalsIgnoreCase("/sys/fs/cgroup/memory")).count(), equalTo(1L));
+        assertThat(filePermissions.stream().filter(p -> p.getName().contains("/lib/security/cacerts")).count(), equalTo(1L));
```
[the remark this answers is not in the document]
reta: No, reevaluate within test context

libs/agent-sm/agent-policy/src/main/java/org/opensearch/secure_sm/policy/PolicyFile.java
```diff
@@ -148,6 +172,12 @@ private static PermissionEntry expandPermissionName(PermissionEntry pe) {
             return new PermissionEntry(pe.permission(), expandKnownSystemProperty("java.home", pe.name()), pe.action());
         } else if (pe.name().startsWith("${user.home}")) {
             return new PermissionEntry(pe.permission(), expandKnownSystemProperty("user.home", pe.name()), pe.action());
+        } else if (pe.name().contains("${opensearch.cgroups.hierarchy.override}")) {
```
cwperks, line 175: This is fine adding this here, but in the future I'd like a clear delineation when to use each syntax:

1. ${...} - _known_ system prop
2. ${{...}} - unknown? system prop

Its confusing what is considered known vs. not known IMO.

reta: Thanks @cwperks , looking for better options here for sure: `opensearch.cgroups.hierarchy.override` is somewhat special since it kind of maps to the path segment within the filesystem `/sys/fs/cgroup/` and is handled respectively (not an arbitrary placeholder to say)

peterzhuamazon:
Hi @reta do we need to update anything in https://github.com/peterzhuamazon/OpenSearch/blob/main/distribution/packages/src/common/systemd/opensearch.service ?

Thanks.

reta:
> Hi @reta do we need to update anything in https://github.com/peterzhuamazon/OpenSearch/blob/main/distribution/packages/src/common/systemd/opensearch.service ?
> 
> Thanks.

Thanks @peterzhuamazon , I believe manual changes will be needed there, since this is really a JVM property that we cannot use for systemd hardening, what do you think? The reporter on the issue mentioned it:

> The paths for systemd cgroups are specified, but systemd cannot be used in a container.

libs/agent-sm/agent-policy/src/main/java/org/opensearch/secure_sm/policy/PolicyFile.java
reta: Made it (hopefully) a bit cleared, renamed the method to `expandKnownSystemProperty`

peterzhuamazon:
Hi @reta do we have any defaults for `opensearch.cgroups.hierarchy.override`? I can definitely update that in opensearch.service to start with. Else I can keep it for now and check back next release to see if anything blocked.

For deb/rpm it is not as flexible as tar, so I think we can hardcode the default into systemd file for now.

reta:
> Hi @reta do we have any defaults for `opensearch.cgroups.hierarchy.override`? I can definitely update that in opensearch.service to start with. Else I can keep it for now and check back next release to see if anything blocked.

@peterzhuamazon Yes, default is `/` which is what we need and fits default settings perfectly (no issues). Basically, we don't need any changes for default settings.

PR #18583, opened 2025-06-23
Add timeout handling for S3 blob container async operations

ashking94:
### Description
This PR adds timeout handling for async operations in the S3 blob container to prevent snapshot deletion threads from getting stuck indefinitely. Previously, when S3 operations would hang or never complete, the snapshot deletion threadpool could get stuck with non-zero active thread count.

The change can be summarised as below -
- Added a 30-second timeout for async operations in `getFutureValue()`
- Added proper cancellation of futures on timeout
- Added new test case `testDeleteTimeoutWithNeverCompletingAsyncDeletionFuture` to verify timeout behavior

### Related Issues
Resolves #18314

### Check List
- [x] Functionality includes testing.
- ~[ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.~
- ~[ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #20566, opened 2026-02-06
Implementing batched deletions of stale ClusterMetadataManifests in R…

### LINKED ISSUE #20564: [BUG] Remote Cluster State cleanup failures due to deletions timeouts resulting in stale-metadata pile-ups

Pranshu-S:
### Description

Fixes remote cluster state cleanup failures that were causing **stale metadata pile-ups** in remote storage when deletions time out along with fixes mention in tagged GitHub Issue

**Issue context (#20564):** `RemoteClusterStateCleanupManager` runs every ~5 minutes (configurable) and performs *sequential* deletions (global metadata → index metadata → ephemeral attrs → manifests). A recent change added a **30s timeout** to the S3 “sync” delete path; with large delete sets this can throw `IOException`, **abort the whole cleanup run** (single try/catch), and leave later phases undeleted—making the next run even larger and more likely to fail. 

### What this PR does

* Batch deletes stale `manifest*` blobs to reduce per-call delete size and avoid timeout/payload issues. 
  * Controlled by:
    * `cluster.remote_store.state.cleanup.batch_size`
    * `cluster.remote_store.state.cleanup.max_batches`
* Makes delete timeout configurable for remote-state cleanup
* Fixing issues where the update-interval call was not being honoured when the deletion was disabled (set to -1)
* Fixing mis-match in deletions  workflow where we were deleting manifests before the the IndexRouting paths which could lead to IndexRouting paths being dangled 
* Fixing issue where we were moving the `lastCleanupAttemptStateVersion` even though the cleanup were failing which resulted in next deletes to become no-op if there were no cluster-state changes (or less than 10) and previous deletions failed.
* Reducing logging overhead of DEBUG as it was printing every single file to delete. Moved it to TRACE.

### Related Issues
Resolves #20564 

### Check List


Pranshu-S:
Flaky test - 
```
[org.opensearch.example.systemingestprocessor.ExampleSystemIngestProcessorClientYamlTestSuiteIT.test {yaml=example-system-ingest-processor/20_system_ingest_processor/Processor injects field when index is created from matching template where trigger_setting is true}](https://build.ci.opensearch.org/job/gradle-check/71055/testReport/junit/org.opensearch.example.systemingestprocessor/ExampleSystemIngestProcessorClientYamlTestSuiteIT/test__yaml_example_system_ingest_processor_20_system_ingest_processor_Processor_injects_field_when_index_is_created_from_matching_template_where_trigger_setting_is_true_/)
```

Pranshu-S:
Looks like some issues with BwC tests
```
» [2026-02-07T03:11:30,001][INFO ][o.o.c.s.ClusterManagerService] [v2.19.5-remote-0] Tasks batched with key: org.opensearch.cluster.coordination.JoinHelper, count:4 and sample tasks: elected-as-cluster-manager ([2] nodes joined)[{v2.19.5-remote-0}{QCAW9Wz-S9eqbEMAwegJKg}{ioB2VK4DRL-C-xSgEFmPeQ}{127.0.0.1}{127.0.0.1:35119}{dimr}{upgraded=true, testattr=test, shard_indexing_pressure_enabled=true} elect leader, {v2.19.5-remote-2}{i4ANL5k_TJCv-W2I-WjMeQ}{nsQZoJMFR3WL0pwdo-RlJw}{127.0.0.1}{127.0.0.1:41821}{dimr}{testattr=test, shard_indexing_pressure_enabled=true} elect leader, _BECOME_CLUSTER_MANAGER_TASK_, _FINISH_ELECTION_], term: 53, version: 345, delta: cluster-manager node changed {previous [], current [{v2.19.5-remote-0}{QCAW9Wz-S9eqbEMAwegJKg}{ioB2VK4DRL-C-xSgEFmPeQ}{127.0.0.1}{127.0.0.1:35119}{dimr}{upgraded=true, testattr=test, shard_indexing_pressure_enabled=true}]}
» [2026-02-07T03:11:30,021][INFO ][o.o.c.c.FollowersChecker ] [v2.19.5-remote-0] FollowerChecker{discoveryNode={v2.19.5-remote-1}{t0pClBcpTbqa1UYnnxSLow}{Vdf3ZRR-QZOC6KkSSKFmiw}{127.0.0.1}{127.0.0.1:43807}{dimr}{testattr=test, shard_indexing_pressure_enabled=true}, failureCountSinceLastSuccess=1, [cluster.fault_detection.follower_check.retry_count]=3} disconnected
»  org.opensearch.transport.NodeNotConnectedException: [v2.19.5-remote-1][127.0.0.1:43807] Node not connected
»  	at org.opensearch.transport.ClusterConnectionManager.getConnection(ClusterConnectionManager.java:223)
»  	at org.opensearch.transport.TransportService.getConnection(TransportService.java:990)
»  	at org.opensearch.transport.TransportService.sendRequest(TransportService.java:949)

File: server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java [hunk 1]
```diff
@@ -62,11 +70,47 @@ public class RemoteClusterStateCleanupManager implements Closeable {
...
         Setting.Property.Dynamic
     );
+
+    /**
+     * Setting to specify the batch size for manifest cleanup operations
+     */
+    public static final Setting<Integer> REMOTE_CLUSTER_STATE_CLEANUP_BATCH_SIZE_SETTING = Setting.intSetting(
+        REMOTE_CLUSTER_STATE_CLEANUP_BATCH_SIZE_SETTING_NAME,
+        MANIFEST_CLEANUP_BATCH_SIZE_DEFAULT,
+        RETAINED_MANIFESTS + SKIP_CLEANUP_STATE_CHANGES,
```
rajiv-kv: can we address this :- https://github.com/opensearch-project/OpenSearch/pull/20515 (had some follow-up comment)

I think the better option is to have validation based on `RETAINED_MANIFESTS + SKIP_CLEANUP_STATE_CHANGES` instead of changing the min value for the setitng.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -62,11 +70,47 @@ public class RemoteClusterStateCleanupManager implements Closeable {
...
+        Setting.Property.Dynamic
+    );
     private static final Logger logger = LogManager.getLogger(RemoteClusterStateCleanupManager.class);
     private final RemoteClusterStateService remoteClusterStateService;
     private final RemotePersistenceStats remoteStateStats;
     private BlobStoreTransferService blobStoreTransferService;
     private TimeValue staleFileCleanupInterval;
+    private int cleanupBatchSize;
+    private int cleanupMaxBatches;
+    private TimeValue cleanupTimeout;
```
rajiv-kv: these needs to be volatile references to handle concurrency.

server/src/main/java/org/opensearch/cluster/routing/remote/InternalRemoteRoutingTableService.java
```diff
@@ -286,4 +287,26 @@ public void deleteStaleIndexRoutingDiffPaths(List<String> stalePaths) throws IOE
             throw e;
         }
     }
+
+    @Override
+    public void deleteStaleIndexRoutingPaths(List<String> stalePaths, TimeValue timeValue) throws IOException {
```
rajiv-kv: Can we remove the existing methods ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -320,36 +400,64 @@ void deleteClusterMetadata(
...
                     () -> new ParameterizedMessage("Error while deleting stale index routing diff files {}", staleIndexRoutingDiffPaths),
                     e
                 );
                 remoteStateStats.indicesRoutingDiffFileCleanupAttemptFailed();
+                // throw exception as we do not want to fail repeatedly on all batches
+                throw e;
             }
+
+            // Delete Manifests in the very end to avoid dangling routing files in-case deletion of stale index routing
+            // files after deleting manifests
```
rajiv-kv, line 422: Can we add a unit test for this coverage ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -117,9 +168,38 @@ private void updateCleanupInterval(TimeValue updatedInterval) {
...
+    private void rescheduleAsyncDeletionTask() {
+        if (staleFileDeletionTask != null) {
+            staleFileDeletionTask.rescheduleIfNecessary();
+        }
+    }
+
+    private void updateCleanupBatchSize(Integer updatedBatchSize) {
+        this.cleanupBatchSize = updatedBatchSize;
+        logger.info("updated remote state cleanup batch size to {}", updatedBatchSize);
+        rescheduleAsyncDeletionTask();
```
rajiv-kv: why do we need to reschedule for settings other than timeout ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -117,9 +168,38 @@ private void updateCleanupInterval(TimeValue updatedInterval) {
         // After updating the interval, we need to close the current task and create a new one which will run with updated interval
         if (staleFileDeletionTask != null && !staleFileDeletionTask.getInterval().equals(updatedInterval)) {
             staleFileDeletionTask.setInterval(updatedInterval);
+            if (staleFileDeletionTask.isScheduled() == false) {
```
rajiv-kv: how will the schedule get updated without rescheduling ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -359,49 +467,68 @@ void deleteClusterMetadata(
...
+                }
+
+                // Update version only after successful completion
+                lastCleanupAttemptStateVersion = cleanUpAttemptStateVersion;
+            } catch (Exception e) {
+                logger.error(
+                    "Exception occurred while deleting Remote Cluster Metadata for clusterUUIDs [{}]. Exception: {}",
+                    clusterUUID,
+                    e
+                );
```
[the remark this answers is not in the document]
rajiv-kv: can we fix this ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -359,49 +467,68 @@ void deleteClusterMetadata(
...
+
+                if (batchesProcessed == cleanupMaxBatches) {
+                    logger.warn("Exhausted batch limit for deleting entities. Attempted [{}] batches", cleanupMaxBatches);
+                }
+
+                // Update version only after successful completion
+                lastCleanupAttemptStateVersion = cleanUpAttemptStateVersion;
+            } catch (Exception e) {
+                logger.error(
+                    "Exception occurred while deleting Remote Cluster Metadata for clusterUUIDs [{}]. Exception: {}",
```
rajiv-kv: should we also include `cleanUpAttemptStateVersion`

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -359,49 +467,68 @@ void deleteClusterMetadata(
...
+                    batchesProcessed++;
+
+                    List<BlobMetadata> batchManifests = getBlobStoreTransferService().listAllInSortedOrder(
+                        remoteManifestManager.getManifestFolderPath(clusterName, clusterUUID),
+                        MANIFEST,
+                        cleanupBatchSize
+                    );
+
+                    if (Objects.nonNull(batchManifests) && batchManifests.size() > manifestsToRetain) {
+                        List<BlobMetadata> manifestsToDeletes = batchManifests.subList(manifestsToRetain, batchManifests.size());
```
rajiv-kv, line 480: lets add a TODO to avoid repetitive fetch of manifestsToRetain across batches.

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
```diff
@@ -294,4 +295,70 @@ private BlobPath getIndexRoutingPath(BlobPath baseMetadataPath, String indexUUID
...
+            .cluster()
+            .prepareUpdateSettings()
+            .setPersistentSettings(
+                Settings.builder()
+                    .put("cluster.remote_store.state.cleanup.batch_size", 25)
+                    .put("cluster.remote_store.state.cleanup.max_batches", 3)
+                    .put(REMOTE_CLUSTER_STATE_CLEANUP_INTERVAL_SETTING.getKey(), "-1") // Disable cleanup
+            )
+            .get();
+        assertTrue(response.isAcknowledged());
```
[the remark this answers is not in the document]
Pranshu-S: Cleanup is enabled by default. Not required.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: If the value of the "REMOTE_CLUSTER_STATE_CLEANUP_BATCH_SIZE_SETTING" is less than RETAINED_MANIFESTS + SKIP_CLEANUP_STATE_CHANGES, then the deletion would effectively never work since we would continue skipping. Hence adding the minimum validation for it.

With respect to the valdiation.

```
curl -XPUT "http://localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d'
{
  "persistent" : {
    "cluster.remote_store.state.cleanup.batch_size" : 10
  }
}'
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Failed to parse value [10] for setting [cluster.remote_store.state.cleanup.batch_size] must be >= 20"}],"type":"illegal_argument_exception","reason":"Failed to parse value [10] for setting [cluster.remote_store.state.cleanup.batch_size] must be >= 20"},"status":400}%          
```

Also RETAINED_MANIFESTS and SKIP_CLEANUP_STATE_CHANGES are individually constants and cannot be changed by settings.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Right, will update.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: You are right, this would not be required. If we are making them volatile - it should get picked or applied immediately

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: When we do not disable the task, we also reschedule as part of updating the interval
https://github.com/opensearch-project/OpenSearch/blob/8134f12ab452e73487f30135a139302a85e4be76/server/src/main/java/org/opensearch/common/util/concurrent/AbstractAsyncTask.java#L88-L93

Here "cancellable" is the task itself. When we set the interval to "-1", the task gets disabled after ongoing run and its set to null

https://github.com/opensearch-project/OpenSearch/blob/8134f12ab452e73487f30135a139302a85e4be76/server/src/main/java/org/opensearch/common/util/concurrent/AbstractAsyncTask.java#L128-L131

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
```diff
@@ -294,4 +295,70 @@ private BlobPath getIndexRoutingPath(BlobPath baseMetadataPath, String indexUUID
...
+        List<String> finalManifests = repository.blobStore()
+            .blobContainer(manifestContainerPath)
+            .listBlobsByPrefix("manifest")
+            .keySet()
+            .stream()
+            .sorted()
+            .collect(Collectors.toList());
+        List<String> last10Final = finalManifests.subList(Math.max(0, finalManifests.size() - 10), finalManifests.size());
+
+        assertEquals("Last 10 manifest files should remain the same after cleanup", last10Initial, last10Final);
```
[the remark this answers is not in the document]
shwetathareja: can you please check this comment

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
rajiv-kv: >RETAINED_MANIFESTS and SKIP_CLEANUP_STATE_CHANGES are individually constants and cannot be changed by settings

Yeah, this is my concern. the setting's minimum value is based on the logic in code, and will be difficult for user to understand. Can we instead add a validation handler for the check and give message on lines of "should be greater than number of manifests to retain" ?

> then the deletion would effectively never work since we would continue skipping.
I assume the code will correct correctly even if this is the case.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
rajiv-kv: Can you please explain as to why we need this additional check and the check in `rescheduleIfNecessary` will not suffice ?

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -48,8 +49,15 @@ public class RemoteClusterStateCleanupManager implements Closeable {
 
     public static final int RETAINED_MANIFESTS = 10;
     public static final int SKIP_CLEANUP_STATE_CHANGES = 10;
+    public static final int MANIFEST_CLEANUP_BATCH_SIZE_DEFAULT = 1000;
+    public static final int MANIFEST_CLEANUP_MAX_BATCHES_DEFAULT = 100;
```
rajiv-kv: this seems to be a high number, should we make it 3 by default ?

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
```diff
@@ -294,4 +295,124 @@ private BlobPath getIndexRoutingPath(BlobPath baseMetadataPath, String indexUUID
             RemoteStoreEnums.PathHashAlgorithm.FNV_1A_BASE64
         );
     }
+
+    public void testRemoteCleanupMultipleBatchDeletion() throws Exception {
```
rajiv-kv: can we parametrize this test and run for various batch sizes ?
1, 5, 50, 1000

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
```diff
@@ -294,4 +295,124 @@ private BlobPath getIndexRoutingPath(BlobPath baseMetadataPath, String indexUUID
...
+            .get();
+        assertTrue(response.isAcknowledged());
+
+        updateClusterStateNTimes(50);
+
+        RepositoriesService repositoriesService = internalCluster().getClusterManagerNodeInstance(RepositoriesService.class);
+        BlobStoreRepository repository = (BlobStoreRepository) repositoriesService.repository(REPOSITORY_NAME);
+        BlobPath manifestContainerPath = getBaseMetadataPath(repository).add("manifest");
+
+        List<String> initialManifests = repository.blobStore()
```
rajiv-kv, line 343: should we assert (or) wait until the initialManifests size is >=50

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
```diff
@@ -294,4 +295,124 @@ private BlobPath getIndexRoutingPath(BlobPath baseMetadataPath, String indexUUID
...
+
+        assertBusy(() -> {
+            List<String> currentManifests = repository.blobStore()
+                .blobContainer(manifestContainerPath)
+                .listBlobsByPrefix("manifest")
+                .keySet()
+                .stream()
+                .sorted()
+                .collect(Collectors.toList());
+            assertTrue("Manifests should be cleaned up in batches", currentManifests.size() < initialManifests.size());
```
rajiv-kv: Shouldn't it be exactly equal to last10Initial.size ?

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
```diff
@@ -294,4 +295,124 @@ private BlobPath getIndexRoutingPath(BlobPath baseMetadataPath, String indexUUID
...
+        updateClusterStateNTimes(50);
+
+        RepositoriesService repositoriesService = internalCluster().getClusterManagerNodeInstance(RepositoriesService.class);
+        BlobStoreRepository repository = (BlobStoreRepository) repositoriesService.repository(REPOSITORY_NAME);
+        BlobPath manifestContainerPath = getBaseMetadataPath(repository).add("manifest");
+
+        response = client().admin()
+            .cluster()
+            .prepareUpdateSettings()
+            .setPersistentSettings(Settings.builder().put(REMOTE_CLUSTER_STATE_CLEANUP_INTERVAL_SETTING.getKey(), "10s"))
```
rajiv-kv: Are we setting it to high value  so that we can validate before the second run ?

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
```diff
@@ -294,4 +295,124 @@ private BlobPath getIndexRoutingPath(BlobPath baseMetadataPath, String indexUUID
...
+        assertTrue(response.isAcknowledged());
+
+        List<String> initialManifests = repository.blobStore()
+            .blobContainer(manifestContainerPath)
+            .listBlobsByPrefix("manifest")
+            .keySet()
+            .stream()
+            .toList();
+
+        assertBusy(() -> {
```
rajiv-kv: Can we get reference of `RemoteClusterStateCleanupManager` and execute exactly once, instead of loop waiting ?

server/src/main/java/org/opensearch/common/blobstore/BlobContainer.java
```diff
@@ -281,6 +282,19 @@ default void writeBlobAtomicWithMetadata(
...
+    /**
+     * Deletes the blobs with given names. This method will not throw an exception
+     * when one or multiple of the given blobs don't exist and simply ignore this case.
+     *
+     * @param   blobNames  The names of the blob to delete.
+     * @param   timeout    The timeout for the delete operation.
+     * @throws  IOException if a subset of blob exists but could not be deleted.
+     */
+    default void deleteBlobsIgnoringIfNotExists(List<String> blobNames, TimeValue timeout) throws IOException {
+        LogManager.getLogger(this.getClass()).warn("No implementation exists for using timeout, skipping timeout input");
```
Bukhtawar: Doesn't seem to be right behaviour to ignore parameters.

Pranshu-S: The original implementation done in https://github.com/opensearch-project/OpenSearch/pull/18583/changes has two issues:
1. The timeout was introduced by default in only S3Container and was not a behaviour for other container implementation. That is, it became specific to nature to only s3 plugin. Ideally it should have been made in a way that other containers can use it as well
2. Even with the timeouts being applied in S3Container, we still continue to run the deletions in background as this is async which is also a wrong expectation since the timeout should ideally implement a closure of background delete operations as well. If i were to implement a base timeout the way it is done in S3Plugin today at the interface - we will end up still performing the deletions in background and give clients the timeout exception which is also wrong.

Considering these, I have for now defaulted to using the base implementation without timeouts when containers do not define them. In my opinion we will need to restructure the async and sync calls and implementations such that they honour timeouts before we enforce it.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: I think 100 should be fine, we are fetching a low number of manifests and with each batch run, the list is getting refreshed so we are not piling up any information from previous batches.

This will also help in clearing any previous manifest as well.

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
Pranshu-S: We disable the cleanup in this test and run 
```
        updateClusterStateNTimes(50);
```

Which should create 50 manifest files

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
Pranshu-S: I updated 50 times and ran 1 batch of 20 deletions, hence we will have around 30 files and not last10Initial

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
Pranshu-S: Updated, intended to add 1s.

Pranshu-S:
There seems to be failures while setting up credentials in the gradle check workflow - 
```
Run aws-actions/configure-aws-credentials@v5
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Error: Could not assume role with OIDC: Not authorized to perform sts:AssumeRoleWithWebIdentity
```

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
```diff
@@ -359,49 +458,72 @@ void deleteClusterMetadata(
...
-            deleteStaleMetadataRunning.set(false);
-            throw e;
-        }
+
+                if (batchesProcessed == cleanupMaxBatches) {
+                    logger.warn("Exhausted batch limit for deleting entities. Attempted [{}] batches", cleanupMaxBatches);
+                }
+
+                // Update version only after successful completion
+                lastCleanupAttemptStateVersion = cleanUpAttemptStateVersion;
```
rajiv-kv: needs to be skipped if we exhaust batch

peterzhuamazon:
Hi @Pranshu-S sorry for the confusion that has been fixed now.
If you push a new commit it should take effect.

Thanks.

Pranshu-S:
Looks like gradle check didn't run
```
WORKFLOW_URL
time passed: 3840
```

Pranshu-S:
```
[Test Result](https://build.ci.opensearch.org/job/gradle-check/71204/testReport/) (1 failure / ±0)
[org.opensearch.indices.pollingingest.DefaultStreamPollerTests.testConsumerReinitializationWithNoInitialMessages](https://build.ci.opensearch.org/job/gradle-check/71204/testReport/junit/org.opensearch.indices.pollingingest/DefaultStreamPollerTests/testConsumerReinitializationWithNoInitialMessages/)
```

Flaky

Pranshu-S:
[Test Result](https://build.ci.opensearch.org/job/gradle-check/71207/testReport/) (2 failures / +2)
[org.opensearch.indices.replication.SegmentReplicationSuiteIT.testFullRestartDuringReplication](https://build.ci.opensearch.org/job/gradle-check/71207/testReport/junit/org.opensearch.indices.replication/SegmentReplicationSuiteIT/testFullRestartDuringReplication/)
[org.opensearch.indices.replication.SegmentReplicationTargetServiceTests.testProcessLatestCheckpointIfCheckpointAhead](https://build.ci.opensearch.org/job/gradle-check/71207/testReport/junit/org.opensearch.indices.replication/SegmentReplicationTargetServiceTests/testProcessLatestCheckpointIfCheckpointAhead/)

Flaky

Pranshu-S:
[Test Result](https://build.ci.opensearch.org/job/gradle-check/71209/testReport/) (2 failures / +1)
[org.opensearch.cluster.coordination.NodeJoinLeftIT.testClusterStabilityWhenClusterStatePublicationLagsWithLongRunningListenerOnApplierThread](https://build.ci.opensearch.org/job/gradle-check/71209/testReport/junit/org.opensearch.cluster.coordination/NodeJoinLeftIT/testClusterStabilityWhenClusterStatePublicationLagsWithLongRunningListenerOnApplierThread/)
[org.opensearch.versioning.ConcurrentSeqNoVersioningIT.testSeqNoCASLinearizability](https://build.ci.opensearch.org/job/gradle-check/71209/testReport/junit/org.opensearch.versioning/ConcurrentSeqNoVersioningIT/testSeqNoCASLinearizability/)

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Updated.

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
Pranshu-S: Updated.

rajiv-kv (approved):
LGTM !

server/src/main/java/org/opensearch/common/blobstore/BlobContainer.java
shwetathareja: @Pranshu-S please add javadocs that default implementation doesnt handle timeout and each container implementation has to handle explicitly.

server/src/main/java/org/opensearch/cluster/routing/remote/InternalRemoteRoutingTableService.java
```diff
@@ -280,10 +289,37 @@ public void deleteStaleIndexRoutingPaths(List<String> stalePaths) throws IOExcep
     public void deleteStaleIndexRoutingDiffPaths(List<String> stalePaths) throws IOException {
         try {
             logger.debug(() -> "Deleting stale index routing diff files from remote - " + stalePaths);
-            blobStoreRepository.blobStore().blobContainer(BlobPath.cleanPath()).deleteBlobsIgnoringIfNotExists(stalePaths);
+            if (blobStoreRepository.blobStore().blobContainer(BlobPath.cleanPath()) instanceof AsyncMultiStreamBlobContainer) {
```
rajiv-kv, line 300: can we avoid fetching this twice `blobStoreRepository.blobStore().blobContainer(BlobPath.cleanPath())` and move to a local variable ?

server/src/main/java/org/opensearch/cluster/routing/remote/InternalRemoteRoutingTableService.java
```diff
@@ -280,10 +289,37 @@ public void deleteStaleIndexRoutingPaths(List<String> stalePaths) throws IOExcep
...
             logger.error(() -> new ParameterizedMessage("Failed to delete some stale index routing diff paths from {}", stalePaths), e);
             throw e;
         }
     }
+
+    private void deleteAsyncInternal(List<String> fileNames) throws IOException {
+        PlainActionFuture<Void> future = new PlainActionFuture<>();
+        try {
+            ((AsyncMultiStreamBlobContainer) blobStoreRepository.blobStore().blobContainer(BlobPath.cleanPath()))
+                .deleteBlobsAsyncIgnoringIfNotExists(fileNames, future);
```
rajiv-kv: - can we pass the parent blobcontainer and timeout as parameter ?
- nit : can be named as `deleteWithTimeout`

Pranshu-S: Updated.

server/src/main/java/org/opensearch/cluster/routing/remote/InternalRemoteRoutingTableService.java
Pranshu-S: Updated.

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
Pranshu-S: Updated.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Updated

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Updated.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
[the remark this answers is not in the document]
Pranshu-S: Updated.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Updated.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Updated.

server/src/main/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManager.java
Pranshu-S: Removed in this PR, will address in another one.

server/src/main/java/org/opensearch/common/blobstore/BlobContainer.java
Pranshu-S: We have removed this implementation in this PR, the changes are only constraint to RemoteClusterStateCleanupManager and InternalRoutingTableService

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
Pranshu-S: Updated. Resolving.

server/src/internalClusterTest/java/org/opensearch/gateway/remote/RemoteClusterStateCleanupManagerIT.java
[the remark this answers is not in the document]
Pranshu-S: Updated.

PR #20575, opened 2026-02-09
Fix content dropping in html_strip character filter (#20554)

### LINKED ISSUE #20554: [BUG] html_strip character filter issue
### Describe the bug

Whenever html_strip character filter sees an HTML tag with an attribute which value ends in = it drop everything between it and the next quote.

If html_strip is used in a field analyzer it will drop perfectly valid tokens instead of the HTML.

### Related component

Search

### To Reproduce

POST /_analyze
```json
{
    "char_filter": [
        {
            "type": "html_strip"
        }
    ],
    "tokenizer": "standard",
    "text": "<a href=\"https://www.example.com/?test=\">example</a> this gets discarded<a href=\"https://www.example.com/\">example2</a> continues here"
}
```

### Expected behavior

Tokens should be produced for the "this gets discarded" string between the two <a> tags.

### Additional Details

**Host/Environment:**
 - OS: 
 - Version 3.2.0

Seungmin123:
### Description
This PR fixes a bug in the `html_strip` character filter (and ingest processor) where content was being incorrectly dropped when an HTML attribute value ended with an equals sign (`=`).

The root cause is a parsing issue in Lucene's `HTMLStripCharFilter`. To resolve this without forking the Lucene class, this change introduces a pre-processing step using `PatternReplaceCharFilter`. It identifies the problematic pattern (an equals sign followed by a double quote or whitespace) and escapes it to the HTML entity `&#61;`. Since `HTMLStripCharFilter` decodes entities in its output text, this workaround is transparent to users while ensuring correct HTML stripping.

### Related Issues
Resolves #20554

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Move Randomness from server to libs/common ([#20570](https://github.com/opensearch-project/OpenSearch/pull/20570))
 
 ### Fixed
+- Fix content dropping in html_strip character filter when attribute values end with = ([#20554](https://github.com/opensearch-project/OpenSearch/issues/20554))
 - Fix flaky test failures in ShardsLimitAllocationDeciderIT ([#20375](https://github.com/opensearch-project/OpenSearch/pull/20375))
 - Prevent criteria update for context aware indices ([#20250](https://github.com/opensearch-project/OpenSearch/pull/20250))
 - Update EncryptedBlobContainer to adhere limits while listing blobs in specific sort order if wrapped blob container supports ([#20514](https://github.com/opensearch-project/OpenSearch/pull/20514))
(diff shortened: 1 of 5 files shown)


sandeshkr419:
[Search Triage] @Seungmin123 Thanks for raising a PR. Did we try testing the HTML Strip Char filter in Lucene in isolation? Just to validate if this is a bug in Lucene and needs a fix in upstream.

msfroh:
@Seungmin123  -- can you please try creating a unit test for the Lucene `HTMLStripCharFilter` without any OpenSearch code? 

In particular, if you clone the Lucene repo and add a failing test case to https://github.com/apache/lucene/blob/main/lucene/analysis/common/src/test/org/apache/lucene/analysis/charfilter/TestHTMLStripCharFilter.java, then you can open an issue against Lucene and potentially supply the fix there. If the underlying Lucene class has a bug, we should fix it there so that everyone using Lucene gets the benefit.

Seungmin123:
Thanks for the feedback. I have tested the HTMLStripCharFilter in isolation using only Lucene 10.3.2 (without any OpenSearch code), and I can confirm that this is indeed a bug in Lucene upstream.

Reproduction Results
The bug is triggered when an attribute value is double-quoted and ends with an equals sign immediately followed by the tag closer (e.g., href="...?test=">).

It appears that the Lucene parser's state machine fails to recognize the closing quote in this specific ="> sequence. As a result, it continues to consume all subsequent text and the next tag's content as part of the attribute value until it encounters the next double quote in the document.

Seungmin123:
https://github.com/apache/lucene/issues/15754

Issue #15754 has been submitted to track this bug. I'm keeping an eye on the updates.

PR #20578, opened 2026-02-09
Relax field-level meta validation constraints

### LINKED ISSUE #19884: [Feature Request] Relax `meta` block on field level

I'm always frustrated when I cannot add comments and explanation to my code :)
The index structure allows us to have `_meta` block in `mappings`. 
Why not on (each) field level in `properties`?

Allow to add a `_meta` or a `description` block to the fields

```javascript
{
    "settings": {  ... },
    "mappings": {
        "properties": {
            "address_coordinates": {
                "type": "geo_point",
                "_meta": {
                       "description": "Geo Point in form lat,lan"
                }
            }, ........
        },
        "_meta": {
            "application": "My Beautiful Index",
            ...
        }
    }
}
```

### Update
Many fields (not all) already support the `meta` block. However, this block is under heavy restrictions, which prohibit users from adding descriptive information.
Based on discussions, it was decided to relax the restrictions. However, since not all the fields support the `meta` block, **AND** refactoring from `meta` to `_meta` will result in _Breaking Changes_ - the field will keep its name `meta` (without underscore)

urmichm:
### Description
Restrictions on the field-level `meta` have been relaxed:
- meta accepts any number of entries as long as the following restrictions are met
- `meta`'s values must be Strings of any length
- `meta`'s values can **not** be nulls
- `meta`'s keys must be Strings of any length

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/19884

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/mapper/TypeParsers.java
+++ b/server/src/main/java/org/opensearch/index/mapper/TypeParsers.java
@@ -64,6 +64,9 @@ public class TypeParsers {
     public static final String INDEX_OPTIONS_FREQS = "freqs";
     public static final String INDEX_OPTIONS_POSITIONS = "positions";
     public static final String INDEX_OPTIONS_OFFSETS = "offsets";
+    public static final int META_MAX_ENTRIES = 100;
+    public static final int META_MAX_KEY_LENGTH = 255;
+    public static final int META_MAX_VALUE_LENGTH = 10_000;
 
     public static void checkNull(String propName, Object propNode) {
         if (false == propName.equals("null_value") && propNode == null) {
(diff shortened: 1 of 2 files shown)


urmichm:
I shall add some more explanation on why relaxing the restrictions won't impact on the cluster stage.
Currently, we have `_meta` on the mappings level and `meta` on the field level. Since only the field level has the restriction, the data is simply migrated to the other place with extra characters.

We **keep** the `meta` strictly as `Map<String,String>`, but drop the restriction on _how long_ the `keys` and `values` are and how many entries are in the map. 

Example:
```json
{
    "settings": {  ... },
    "mappings": {
        "properties": {
            "secret_text": {
                "type": "text",
                "meta": {
                       // this 'meta' block will be rejected due to field below
                       "description": "This text has already too many chars and will not be allowed."
                }
            }, ........
        },
        "_meta": {
            "fields_description": {
               // the whole '_meta' block will be allowed with this nested parameter
               "secret_text" : "This text has already too many chars but will be allowed anyway."
            },
            ...
        }
    }
}
```

 @sandeshkr419 please share your feedback :)

urmichm:
Hello @ShawnQiang1
Would you mind to have a look? :)

sandeshkr419:
[Search Triage] @msfroh Can you help with the review here when you get the chance.

msfroh (commented):
@urmichm -- I've checked the commit message that added these restrictions, and I'm inclined to agree with removing them or at least relaxing them. As @epugh called out, adding more verbose values would be really helpful for agents to understand and interpret the schema of an index.

Instead of wholesale removing the restrictions, may I suggest that we relax them? Going fully unbounded still opens the opportunity for users to shoot themselves in the foot. I would still prefer that a field not have (say) 1MB of metadata.

Maybe we could relax to the following:
1. Max number of keys raised to 100 (from 5). 
2. Increase the max key length to 255 (from 20).
3. Increase the max length of any string value to 10000 (from 50).

I would also potentially be open to allowing nested objects, arrays, and numeric fields in field metadata, but I think that's a more complicated change.

What do you think?

Edit: Incidentally, it looks like you also need to fix some Spotless errors with `./gradlew spotlessApply`.

urmichm:
@msfroh thank you for your feedback 
I agree that nested objects and arrays would be useful, but we would need to think how to calculate the size of such non-string values.

Naïve approach would be to convert the values into the string and check it's length. However, this approach has lots of limitations.

I can update the current restrictions to the suggested values and open a new issue to allow non-string types with proper validation

urmichm:
Given that the `values` could be 10,000 chars long, naïve approach is not as bad as it might seem.

Alternatively we can restrict the total size of meta, rather then number of keys. 
Meta is sort of a json object, we can assume:
 - any string (key or value) is measured by length
 - numbers - 4 "bytes"
 - boolean - 1 "byte"
 - complex types (arrays / objects) - sum of all internal elements

Examples:
- `{"a": 45}` = 5 "bytes"
  - 1 from `a`
  - 4 from `45`
- `{"a": true}` = 2 "bytes"
  - 1 from `a`
  - 1 from `true`
- `{"abc": 123, "obj":{"a":12, "b":"str"}}` = 19 "bytes"
  - 3 from `abc`
  - 4 from `123`
  - 3 from 'obj'
  - 9 from `{"a":12, "b":"str"}`  (1+4+1+3)

We can split numbers into `int` and `float`. 

cc: @msfroh

urmichm:
@msfroh the branch is re-based and pipeline is green
please have a look on this MR once again :)

msfroh (approved):

PR #20582, opened 2026-02-10
Add cluster.blocks.read.auto_release setting (#20539)

### LINKED ISSUE #20539: [BUG] `DiskThresholdMonitor` unconditionally auto-releases user-set `index.blocks.read` setting
### Describe the bug

### Related component

Search

### To Reproduce

### Expected behavior

### Additional Details

```
        .collect(Collectors.toSet());
    
    }
}
```

Seungmin123:
### Description
This PR fixes a bug in the `DiskThresholdMonitor` where it was unconditionally auto-releasing `index.blocks.read` on all indices, even when manually set by users.

This fix introduces a new dynamic cluster setting `cluster.blocks.read.auto_release` (defaulting to `true` for backward compatibility) that controls whether the cluster should automatically release read blocks. By setting this to `false`, users can ensure that manually-set read blocks persist until explicitly removed.

### Related Issues
Resolves #20539

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/cluster/routing/allocation/DiskThresholdMonitor.java
+++ b/server/src/main/java/org/opensearch/cluster/routing/allocation/DiskThresholdMonitor.java
@@ -111,6 +111,15 @@ public class DiskThresholdMonitor {
      */
     private final Set<String> nodesOverHighThresholdAndRelocating = Sets.newConcurrentHashSet();
 
+    /**
+     * The names of the indices whose {@code index.blocks.read} block was applied by this monitor rather than by an operator. Because the
+     * monitor reuses the user-facing {@code index.blocks.read} setting, cluster state carries no record of which blocks it owns, so the
+     * distinction is tracked here. It is only consulted when {@link DiskThresholdSettings#INDEX_READ_BLOCK_AUTO_RELEASE} is disabled; on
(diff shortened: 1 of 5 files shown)


Seungmin123:
Re: the suggestion to claim block ownership only after the apply succeeds — the concern is real, but I'm keeping the current ordering deliberately, so recording the reasoning here.

**The mis-claim is self-healing.** The reconciliation at the top of `handleReadBlocks` drops any index that does not currently carry a read block, and it runs before the release set is computed. So a failed apply clears its own bogus claim on the next `onNewInfo` — one `cluster.info.update.interval`, 30s by default. For a stale claim to actually cause a wrong release, a failed apply, the file cache search threshold clearing, and an operator setting `index.blocks.read` on that same index would all have to land inside that one window. While the threshold is still breached the index stays in `indicesToBlockRead` and is excluded from the release set regardless.

**Claiming on success trades that for a permanent failure.** If the settings update is applied but the response is lost — request timeout, or a cluster-manager step-down mid-request — the block exists in cluster state and was never claimed. On the next run the index already carries the block, so it is filtered out of `indicesToApplyReadBlock` and can never be claimed afterwards. With `auto_release=false` that block is then stranded for good, which is precisely the failure mode this PR exists to remove.

So the current ordering accepts a bogus claim that expires in one cycle in order to avoid a block that strands forever. Happy to revisit if a reviewer prefers the other trade-off.

PR #20549, opened 2026-02-04
Expose wrapped scorer in ProfileScorer via getWrappedScorer method

martin-gaievski:
### Description
This change exposes the wrapped scorer in `ProfileScorer` via a new `getWrappedScorer()` method.

**Background**: When profiling is enabled, OpenSearch wraps scorers with `ProfileScorer` to collect timing metrics. However, plugin queries (like neural-search's `HybridQuery`) may extend the standard Lucene `Scorer` API with custom methods. Currently, plugins cannot access these custom methods when profiling is enabled because the wrapped scorer is stored in a private field with no accessor.

**Solution**: Add a `getWrappedScorer()` method that returns the wrapped scorer, following the same pattern already established in the codebase:
- `ProfileCollector.getDelegate()` (same profiling package)
- `ProfilingAggregator.unwrapAggregator()` (aggregation profiling)

**Use case**: neural-search's `HybridQuery` needs to call `docIdStream()` on its `HybridBulkScorer`, which is not part of the standard Lucene `Scorer` API. 

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/20548

### Check List
- [X] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #20607, opened 2026-02-12
Fix ProfileScorer.getChildren() to expose wrapped scorer in scorer tree

### LINKED ISSUE #20548: [Feature Request] Add getWrappedScorer() method to ProfileScorer for consistency with ProfileCollector

ProfileCollector has a `getDelegate()` method to access the wrapped collector, and ProfilingAggregator has `unwrapAggregator()`. However, ProfileScorer lacks an equivalent method to access the wrapped scorer.

__Use Case__: Plugin queries (like neural-search's HybridQuery) may expose custom methods on their Scorer implementations that are not part of the standard Lucene Scorer API. When profiling is enabled, these queries cannot access their custom methods because ProfileScorer wraps the original scorer.

Add getWrappedScorer() method to ProfileScorer, consistent with the existing ProfileCollector.getDelegate() pattern. Risk are same as existing unwrap patterns - callers bypassing the wrapper lose profiling instrumentation for those calls, but functional correctness is preserved.

### Related component

Search:Query Capabilities

We don't have profiler for hybrid query hosted in neural-search plugin. We can access the wrapped scorers via reflection, but this solution is quite fragile (fail if package or class name changes in core, access to private class variable can be blocked due to security concerns), and not really performant comparing to direct method calls

### Additional context

_No response_

martin-gaievski:
### Description
`ProfileScorer.getChildren()` currently delegates directly to the wrapped scorer's `getChildren()`, 
which skips the `ProfileScorer` level in the scorer tree. This makes the wrapped scorer invisible to 
any code traversing the scorer hierarchy via the standard Lucene `Scorable.getChildren()` API.

This change makes `ProfileScorer` expose the wrapped scorer as a child with the `"PROFILED"` 
relationship label, allowing plugins and custom collectors to discover wrapped scorers through 
profiling wrappers using standard tree traversal — without requiring reflection.

### Changes

- **`ProfileScorer.java`**: Changed `getChildren()` from `return scorer.getChildren()` to 
  `return Collections.singletonList(new ChildScorable(scorer, "PROFILED"))`
- **`ProfileScorerTests.java`**: Added `testGetChildren_exposesWrappedScorerAsChild()` to verify 
  the new behavior

### Related Issues
Resolves #20548

PR #20549 added `getWrappedScorer()` — a public method on `ProfileScorer` that returns the wrapped scorer. This was intended to let neural-search unwrap `ProfileScorer` to find `HybridQueryScorer`.

When implementing the neural-search side, we discovered that `getWrappedScorer()` alone doesn't solve the problem because `ProfileScorer` is a package-private `final class`. Even though `getWrappedScorer()` is a `public` method, Java's access control system blocks `Method.invoke()` from external modules (plugins) on methods of non-public classes. The plugin code gets:

```javascript
IllegalAccessException: class LeafCollector cannot access a member of class 
org.opensearch.search.profile.query.ProfileScorer with modifiers "public"
```

This means the neural-search plugin had to use `method.setAccessible(true)` + `@SuppressForbidden` to call `getWrappedScorer()` via reflection — defeating the purpose of having a clean public API.

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/search/profile/query/ProfileScorer.java
+++ b/server/src/main/java/org/opensearch/search/profile/query/ProfileScorer.java
@@ -36,6 +36,7 @@ import org.apache.lucene.search.DocIdSetIterator;
 import org.apache.lucene.search.Scorer;
 import org.apache.lucene.search.TwoPhaseIterator;
 import org.opensearch.search.profile.AbstractProfileBreakdown;
+import org.opensearch.search.profile.ProfilingWrapper;
 import org.opensearch.search.profile.Timer;
 
 import java.io.IOException;
--- a/server/src/test/java/org/opensearch/search/profile/query/ProfileScorerTests.java
+++ b/server/src/test/java/org/opensearch/search/profile/query/ProfileScorerTests.java
@@ -42,6 +42,7 @@ import org.apache.lucene.search.ScoreMode;
 import org.apache.lucene.search.Scorer;
 import org.apache.lucene.search.Weight;
 import org.opensearch.search.profile.ProfileMetricUtil;
+import org.opensearch.search.profile.ProfilingWrapper;
 import org.opensearch.test.OpenSearchTestCase;
 
 import java.io.IOException;
--- a/server/src/main/java/org/opensearch/search/profile/ProfilingWrapper.java
+++ b/server/src/main/java/org/opensearch/search/profile/ProfilingWrapper.java
@@ -6,7 +6,7 @@
  * compatible open source license.
  */
 
-package org.opensearch.search.profile.query;
+package org.opensearch.search.profile;
 
 import org.opensearch.common.annotation.PublicApi;
 
@@ -14,8 +14,8 @@ import org.opensearch.common.annotation.PublicApi;
  * Generic interface for profiling wrappers that decorate search components with timing instrumentation
  * while preserving access to the original (delegate) component.
  * <p>
- * When search profiling is enabled, OpenSearch wraps scorers, collectors, and other components in
- * profiling decorators that are package-private. This interface provides a public contract for plugins
+ * When search profiling is enabled, OpenSearch wraps scorers, collectors, aggregators, and other components
+ * in profiling decorators that are package-private. This interface provides a public contract for plugins
  * to detect and unwrap such wrappers, accessing the underlying component without requiring reflection
  * or knowledge of package-private implementation classes.
  * <p>
@@ -28,7 +28,8 @@ import org.opensearch.common.annotation.PublicApi;
  * }</pre>
  *
  * @param <T> the type of the wrapped component (e.g., {@link org.apache.lucene.search.Scorer},
- *            {@link org.apache.lucene.search.Collector})
+ *            {@link org.apache.lucene.search.Collector},
+ *            {@link org.opensearch.search.aggregations.Aggregator})
  * @opensearch.api
  */
 @PublicApi(since = "3.6.0")
--- a/server/src/main/java/org/opensearch/search/profile/aggregation/ProfilingAggregator.java
+++ b/server/src/main/java/org/opensearch/search/profile/aggregation/ProfilingAggregator.java
@@ -39,6 +39,7 @@ import org.opensearch.search.aggregations.InternalAggregation;
 import org.opensearch.search.aggregations.LeafBucketCollector;
 import org.opensearch.search.aggregations.support.AggregationPath.PathElement;
 import org.opensearch.search.internal.SearchContext;
+import org.opensearch.search.profile.ProfilingWrapper;
 import org.opensearch.search.profile.Timer;
 import org.opensearch.search.sort.SortOrder;
 
@@ -48,7 +49,7 @@ import java.util.Iterator;
 /**
  * An aggregator that aggregates the performance profiling of other aggregations
  */
-public class ProfilingAggregator extends Aggregator {
+public class ProfilingAggregator extends Aggregator implements ProfilingWrapper<Aggregator> {
 
     private final Aggregator delegate;
     private final AggregationProfiler profiler;
@@ -160,6 +161,11 @@ public class ProfilingAggregator extends Aggregator {
     @Override
     public String toString() {
         return delegate.toString();
+    }
+
+    @Override
+    public Aggregator getDelegate() {
+        return delegate;
     }
 
     @Override
--- a/server/src/main/java/org/opensearch/search/profile/aggregation/ProfilingLeafBucketCollector.java
+++ b/server/src/main/java/org/opensearch/search/profile/aggregation/ProfilingLeafBucketCollector.java
@@ -34,6 +34,7 @@ package org.opensearch.search.profile.aggregation;
 
 import org.apache.lucene.search.Scorable;
 import org.opensearch.search.aggregations.LeafBucketCollector;
+import org.opensearch.search.profile.ProfilingWrapper;
 import org.opensearch.search.profile.Timer;
 
 import java.io.IOException;
@@ -43,9 +44,9 @@ import java.io.IOException;
  *
  * @opensearch.internal
  */
-public class ProfilingLeafBucketCollector extends LeafBucketCollector {
+public class ProfilingLeafBucketCollector extends LeafBucketCollector implements ProfilingWrapper<LeafBucketCollector> {
 
-    private LeafBucketCollector delegate;
+    private final LeafBucketCollector delegate;
     private Timer collectTimer;
 
     public ProfilingLeafBucketCollector(LeafBucketCollector delegate, AggregationProfileBreakdown profileBreakdown) {
@@ -61,6 +62,11 @@ public class ProfilingLeafBucketCollector extends LeafBucketCollector {
         } finally {
             collectTimer.stop();
         }
+    }
+
+    @Override
+    public LeafBucketCollector getDelegate() {
+        return delegate;
     }
 
     @Override
--- a/server/src/main/java/org/opensearch/search/profile/query/ProfileCollector.java
+++ b/server/src/main/java/org/opensearch/search/profile/query/ProfileCollector.java
@@ -39,6 +39,7 @@ import org.apache.lucene.search.FilterLeafCollector;
 import org.apache.lucene.search.LeafCollector;
 import org.apache.lucene.search.Scorable;
 import org.apache.lucene.search.ScoreMode;
+import org.opensearch.search.profile.ProfilingWrapper;
 
 import java.io.IOException;
 
@@ -47,7 +48,7 @@ import java.io.IOException;
  *
  * @opensearch.internal
  */
-final class ProfileCollector extends FilterCollector {
+final class ProfileCollector extends FilterCollector implements ProfilingWrapper<Collector> {
 
     private long time;
     private long sliceStartTime;
--- a/server/src/test/java/org/opensearch/search/profile/aggregation/ProfilingWrapperAggregationTests.java
+++ b/server/src/test/java/org/opensearch/search/profile/aggregation/ProfilingWrapperAggregationTests.java
@@ -1,0 +1,51 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.search.profile.aggregation;
+
+import org.opensearch.search.aggregations.Aggregator;
+import org.opensearch.search.aggregations.LeafBucketCollector;
+import org.opensearch.search.profile.ProfilingWrapper;
+import org.opensearch.test.OpenSearchTestCase;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests that ProfilingAggregator and ProfilingLeafBucketCollector
+ * correctly implement the ProfilingWrapper interface.
+ */
+public class ProfilingWrapperAggregationTests extends OpenSearchTestCase {
+
+    public void testProfilingAggregatorImplementsProfilingWrapper() {
+        Aggregator mockAggregator = mock(Aggregator.class);
+        when(mockAggregator.name()).thenReturn("test_agg");
+        AggregationProfiler profiler = new AggregationProfiler();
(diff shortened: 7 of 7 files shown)


martin-gaievski:
Looks like rarely failing test, not related to this change, passes in my local environment 

```
Task :server:test

REPRODUCE WITH: ./gradlew ':server:test' --tests 'org.opensearch.index.shard.IndexShardTests.testRestoreSearchOnlyShardFromStoreOnNewNode' -Dtests.seed=B3BDACD5418087E4 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=ha-NE -Dtests.timezone=Asia/Makassar -Druntime.java=25

IndexShardTests > testRestoreSearchOnlyShardFromStoreOnNewNode FAILED
    [index][[index][0]] IndexShardRecoveryException[Failed to recover from gateway]; nested: CorruptIndexException[misplaced codec footer (file truncated?): length=0 but footerLength==16 (resource=metadata__9223372036854775805__9223372036854775804__9223372036854775802__9223372036854775805__-1039764442__9223370265996427162__2)];
        at __randomizedtesting.SeedInfo.seed([B3BDACD5418087E4:7369770EE6AE81DC]:0)
        at app//org.opensearch.index.shard.StoreRecovery.internalRecoverFromStoreForSearchReplica(StoreRecovery.java:815)
        at app//org.opensearch.index.shard.StoreRecovery.lambda$recoverFromStore$0(StoreRecovery.java:127)
        at app//org.opensearch.core.action.ActionListener.completeWith(ActionListener.java:344)
        at app//org.opensearch.index.shard.StoreRecovery.recoverFromStore(StoreRecovery.java:124)
        at app//org.opensearch.index.shard.IndexShard.recoverFromStore(IndexShard.java:3198)
        at app//org.opensearch.index.shard.IndexShardTestCase.recoverFromStore(IndexShardTestCase.java:1479)
        at app//org.opensearch.index.shard.IndexShardTestCase.recoverShardFromStore(IndexShardTestCase.java:1085)
        at app//org.opensearch.index.shard.IndexShardTests.testRestoreSearchOnlyShardFromStoreOnNewNode(IndexShardTests.java:3092)

        Caused by:
        org.apache.lucene.index.CorruptIndexException: misplaced codec footer (file truncated?): length=0 but footerLength==16 (resource=metadata__9223372036854775805__9223372036854775804__9223372036854775802__9223372036854775805__-1039764442__9223370265996427162__2)
            at app//org.apache.lucene.codecs.CodecUtil.checksumEntireFile(CodecUtil.java:616)
            at app//org.opensearch.common.io.VersionedCodecStreamWrapper.readStream(VersionedCodecStreamWrapper.java:63)
            at app//org.opensearch.index.store.RemoteSegmentStoreDirectory.readMetadataFile(RemoteSegmentStoreDirectory.java:282)
            at app//org.opensearch.index.store.RemoteSegmentStoreDirectory.readLatestMetadataFile(RemoteSegmentStoreDirectory.java:271)
            at app//org.opensearch.index.store.RemoteSegmentStoreDirectory.init(RemoteSegmentStoreDirectory.java:178)
            at app//org.opensearch.index.shard.IndexShard.syncSegmentsFromRemoteSegmentStore(IndexShard.java:5497)
            at app//org.opensearch.index.shard.IndexShard.syncSegmentsFromRemoteSegmentStore(IndexShard.java:5480)
            at app//org.opensearch.index.shard.IndexShard.innerOpenEngineAndTranslog(IndexShard.java:2871)
            at app//org.opensearch.index.shard.IndexShard.openEngineAndSkipTranslogRecovery(IndexShard.java:2838)
            at app//org.opensearch.index.shard.IndexShard.openEngineAndSkipTranslogRecoveryFromSnapshot(IndexShard.java:2832)
            at app//org.opensearch.index.shard.StoreRecovery.completeRecovery(StoreRecovery.java:843)
            at app//org.opensearch.index.shard.StoreRecovery.internalRecoverFromStoreForSearchReplica(StoreRecovery.java:813)
            ... 7 more

> Task :plugins:mapper-murmur3:bundlePlugin
```

server/src/main/java/org/opensearch/search/profile/query/ProfileScorer.java
```diff
@@ -102,7 +103,11 @@ public float score() throws IOException {
 
     @Override
     public Collection<ChildScorable> getChildren() throws IOException {
-        return scorer.getChildren();
+        // Expose the wrapped scorer as a child so that plugins and collectors can traverse
+        // through profiling wrappers using the standard Lucene Scorable.getChildren() API.
+        // Without this, ProfileScorer is invisible in the scorer tree, preventing custom
+        // collectors from finding wrapped scorers.
+        return Collections.singletonList(new ChildScorable(scorer, "PROFILED"));
```
jainankitk: This seems hacky and inconsistent to me. I am wondering if there is a way to expose `getWrapperScorer` otherwise

martin-gaievski: Thank you for the feedback @jainankitk, that's a fair point. I agree that changing `getChildren()` semantics is not the cleanest approach — it does mix the "wrapper" concept with the "query tree children" concept that `getChildren()` is designed for.

Root problem is that `ProfileScorer` is package-private, so even though we added a `public getWrappedScorer()` method in #20549, plugins cannot call it without `Method.setAccessible(true)` because Java's access control blocks reflection calls to public methods on non-public classes from external modules.

The simplest solution would be to make `ProfileScorer` public:

```java
public final class ProfileScorer extends Scorer {
```
 
This would allow plugins to use `instanceof ProfileScorer` directly (no class name string comparison) and call `getWrappedScorer()` directly (no reflection, no `setAccessible`)
 
ProfileScorer class is already `final` so it can't be extended, and `ProfileWeight` in the same package is already `public`. The visibility change wouldn't affect any existing behavior.

Alternatively, we could introduce a small public interface like `ScorerWrapper` that `ProfileScorer` implements, keeping `ProfileScorer` itself package-private. But that feels over-engineered for this use case.

Would making `ProfileScorer` public be acceptable? I can update this PR accordingly.
I'll revert the `getChildren()` change as it's not needed if `ProfileScorer` is made public.

martin-gaievski: I pushed change with a new ScoreWrapper interface, with such approach we don't need to make `ProfileScorer` public and we don't need to touch children scorers.

@jainankitk let me know if this approach works for you

server/src/main/java/org/opensearch/search/profile/query/ScorerWrapper.java
```diff
@@ -0,0 +1,42 @@
...
+ * if (scorer instanceof ScorerWrapper) {
+ *     Scorer wrappedScorer = ((ScorerWrapper) scorer).getWrappedScorer();
+ *     // access custom scorer methods
+ * }
+ * }</pre>
+ *
+ * @opensearch.api
+ */
+@PublicApi(since = "3.6.0")
+public interface ScorerWrapper {
```
owaiskazi19: How about?
```suggestion
public interface WrappedScorerAccessor {
```

server/src/test/java/org/opensearch/search/profile/query/ProfileScorerTests.java
```diff
@@ -112,4 +113,30 @@ public void testGetWrappedScorer() throws IOException {
...
+        Query query = new MatchAllDocsQuery();
+        Weight weight = query.createWeight(new IndexSearcher(new MultiReader()), ScoreMode.TOP_SCORES, 1f);
+        FakeScorer fakeScorer = new FakeScorer(weight);
+        QueryProfileBreakdown profile = new QueryProfileBreakdown(ProfileMetricUtil.getDefaultQueryProfileMetrics());
+        ProfileScorer profileScorer = new ProfileScorer(fakeScorer, profile);
+
+        // getChildren() should delegate to the wrapped scorer's getChildren()
+        Collection<Scorer.ChildScorable> children = profileScorer.getChildren();
+        assertEquals("FakeScorer has no children, so ProfileScorer should return empty collection", 0, children.size());
+    }
```
owaiskazi19, line 174: Can we also add a test to simulate an actual plugin call where a ProfileScorer wraps another ProfileScorer (nested profiling)? Basically how a Scorable reference unwraps the ProfileScorer from ScorerWrapper

```java
  // Create a chain: outer ProfileScorer -> inner ProfileScorer -> FakeScorer
    ProfileScorer innerProfileScorer = new ProfileScorer(fakeScorer, profile);
    ProfileScorer outerProfileScorer = new ProfileScorer(innerProfileScorer, profile);

    Scorable scorable = outerProfileScorer;

    // First unwrap: outer -> inner
    assertTrue(scorable instanceof ScorerWrapper);
    Scorer firstUnwrap = ((ScorerWrapper) scorable).getWrappedScorer();
    
    and then second one...
 ```

File: server/src/main/java/org/opensearch/search/profile/query/ScorerWrapper.java (hunk 2 above)
Comment (martin-gaievski) [reply server/src/main/java/org/opensearch/search/profile/query/ScorerWrapper.java]: Agree, this name sounds more descriptive

File: server/src/test/java/org/opensearch/search/profile/query/ProfileScorerTests.java (hunk 3 above)
Comment (martin-gaievski) [reply server/src/test/java/org/opensearch/search/profile/query/ProfileScorerTests.java:174]: Added testUnwrapFromScorableReference() test method that does this

File: CHANGELOG.md [hunk 4]
```diff
@@ -5,7 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased 3.x]
 ### Added
-- Add getWrappedScorer method to ProfileScorer for plugin access to wrapped scorers ([#20548](https://github.com/opensearch-project/OpenSearch/issues/20548))
+- Add WrappedScorerAccessor interface for plugin access to wrapped scorers in profiling decorators ([#20548](https://github.com/opensearch-project/OpenSearch/issues/20548))
```
owaiskazi19: We link the PR here and below on L31 as well

server/src/test/java/org/opensearch/search/profile/query/ProfileScorerTests.java
```diff
@@ -112,4 +114,48 @@ public void testGetWrappedScorer() throws IOException {
...
+        ProfileScorer profileScorer = new ProfileScorer(fakeScorer, profile);
+
+        // ProfileScorer implements WrappedScorerAccessor, allowing plugins to detect and unwrap
+        // profiling wrappers using instanceof without reflection
+        assertTrue("ProfileScorer should implement WrappedScorerAccessor", profileScorer instanceof WrappedScorerAccessor);
+        WrappedScorerAccessor accessor = (WrappedScorerAccessor) profileScorer;
+        assertSame("WrappedScorerAccessor.getWrappedScorer() should return the original scorer", fakeScorer, accessor.getWrappedScorer());
+    }
+
+    public void testUnwrapFromScorableReference() throws IOException {
```
owaiskazi19: Tests the same like `testImplementsWrappedScorerAccessor`. Can add nested profiling here to differentiate it

martin-gaievski: updated test class, added nested profiling unwrap test:
- replaced `testUnwrapFromScorableReference` with`testUnwrapNestedProfilingScorerFromScorableReference`
- creates `FakeScorer` → `innerProfileScorer` → `outerProfileScorer` (two profiling layers)
- plugin receives `Scorable` reference, iteratively unwraps through both layers
- verifies intermediate unwrap is still `WrappedScorerAccessor`, final unwrap is the original `FakeScorer`

owaiskazi19 (approved):
Looks good. Thanks @martin-gaievski . Will wait for @jainankitk to take a look before merging in

server/src/main/java/org/opensearch/search/profile/query/WrappedScorerAccessor.java
```diff
@@ -0,0 +1,42 @@
...
+ * if (scorer instanceof WrappedScorerAccessor) {
+ *     Scorer wrappedScorer = ((WrappedScorerAccessor) scorer).getWrappedScorer();
+ *     // access custom scorer methods
+ * }
+ * }</pre>
+ *
+ * @opensearch.api
+ */
+@PublicApi(since = "3.6.0")
+public interface WrappedScorerAccessor {
```
jainankitk: I am wondering if we should add generic interface that can be implemented by any ProfilingWrapper to return the delegate?

server/src/main/java/org/opensearch/search/profile/query/ProfileScorer.java
jainankitk: While making `ProfileScorer` public is the simplest way and better than changing `getChildren()` imo, it takes away the flexibility provided by internal class. Hence, I am good with the interface approach.

server/src/main/java/org/opensearch/search/profile/query/WrappedScorerAccessor.java
martin-gaievski: That's a great idea and I agree it would be more future-proof. We could introduce a generic interface like:

```java
@PublicApi(since = "3.6.0")
public interface ProfilingWrapper<T> {
    T getDelegate();
}
```

`ProfileScorer` would implement `ProfilingWrapper<Scorer>`, and `ProfileCollector` (which already has a `getDelegate()` method) could also implement it for consistency.

I can rename `WrappedScorerAccessor` to `ProfilingWrapper<T>`, update `ProfileScorer` to `implements ProfilingWrapper<Scorer>`, rename method to `getDelegate()`. 

@jainankitk will this address your concern/suggestion?

martin-gaievski: pushed commit that added more generic ProfilingWrapper interface

jainankitk (approved):
Thanks @martin-gaievski for addressing the review comment. Can we implement this interface for `ProfilingAggregator`, `ProfileCollector` and any other class providing access to the delegate?

martin-gaievski:
> Thanks @martin-gaievski for addressing the review comment. Can we implement this interface for `ProfilingAggregator`, `ProfileCollector` and any other class providing access to the delegate?

We can add the interface to ProfilingAggregator and ProfilingLeafBucketCollector, just pushed new commit with corresponding change

PR #20608, opened 2026-02-12
Adding support for Row Id mapping for composite merge

darjisagar7:
### Description

This PR will create the RowIdMapping which will be used for the composite merges

image

image

### Related Issues

### Check List

### DIFF
--- a/modules/parquet-data-format/src/main/rust/src/parquet_merge.rs
+++ b/modules/parquet-data-format/src/main/rust/src/parquet_merge.rs
@@ -1,6 +1,5 @@
 use jni::JNIEnv;
-use jni::objects::{JClass, JObject, JString};
-use jni::sys::jint;
+use jni::objects::{JClass, JObject, JString, JValue};
 use std::fs::File;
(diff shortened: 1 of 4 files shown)


modules/parquet-data-format/src/main/rust/src/parquet_merge.rs
```diff
@@ -209,14 +239,23 @@ fn process_files(
             file_batches += 1;
         }
 
+        // Create mappings for this file
+        for old_row_id in 0..file_rows as i64 {
```
Shailesh-Kumar-Singh, line 243: Here we are creating `RowIdMappingData` struct for every row, for a merge of files totaling N rows, this will be 3N Java heap allocations triggered from Rust through JNI +  4N JNI calls from `create_row_id_mapping_object`
Instead, since the mapping is just `new_row_id = file_start_offset + old_row_id`, we can pass only per-file segment metadata (file ID, start offset, row count) across JNI and materialize the full mapping in Java.

PR #20609, opened 2026-02-12
[segment replication] Simplify the ckp verification for the primary shard relocation

guojialiang92:
### Description
This PR is to address the issues in #[[20610](https://github.com/opensearch-project/OpenSearch/issues/20610)].

If we can specifically identify the scenario of primary shard relocation, we can avoid continuously patching the ckp verification logic.

Fortunately, we set state [handoffInProgress](https://github.com/opensearch-project/OpenSearch/blob/4ea835964381be2b008a9161bd8935b9886d0309/server/src/main/java/org/opensearch/index/seqno/ReplicationTracker.java#L1776) during the hand-off phase of peer recovery. We can leverage this state to require that when the primary shard receives request `GET_CHECKPOINT_INFO`, **it must be a started primary shard that is not in the hand-off process**.

### Related Issues
Resolves #[[20610](https://github.com/opensearch-project/OpenSearch/issues/20610)]

### Check List


server/src/main/java/org/opensearch/indices/replication/AbstractSegmentReplicationTarget.java
```diff
@@ -165,28 +162,6 @@ public void startReplication(ActionListener<Void> listener, BiConsumer<Replicati
...
 
         checkpointInfoListener.whenComplete(checkpointInfo -> {
-            ReplicationCheckpoint getMetadataCheckpoint = checkpointInfo.getCheckpoint();
-            // Only enforce strict checkpoint validation during normal replication, not during recovery.
-            // During recovery (shard is INITIALIZING or RELOCATING), the replica may have a stale checkpoint
-            // from before a restart, and should accept the primary's current state even if it appears older.
-            // See: https://github.com/opensearch-project/OpenSearch/issues/19234
-            boolean isRecovering = indexShard.routingEntry().initializing() || indexShard.routingEntry().relocating();
-            if (indexShard.indexSettings().isSegRepLocalEnabled()
-                && checkpoint.isAheadOf(getMetadataCheckpoint)
```
atris, line 175: We cannot remove `checkpoint.isAheadOf(getMetadataCheckpoint)`. `isHandoffInProgress` handles the relocation race but fails against split-brain or Zombie Primaries returning outdated metadata. Without this check, the replica will accept older metadata causing state rollback (deletion of committed segments). The safety check must stay.

atris (changes requested):

server/src/main/java/org/opensearch/indices/replication/AbstractSegmentReplicationTarget.java
Bukhtawar: I'm wondering whether an existing test would catch this?

PR #20625, opened 2026-02-13
Use env variable (OPENSEARCH_FIPS_MODE) to enable opensearch to run in FIPS enforced mode instead of checking for existence of bcFIPS jars

cwperks:
### Description

This PR updates logic in `./bin/opensearch-env` (and `/bin/opensearch-env.bat`) to change the condition for running in FIPS approved mode. Currently, opensearch-env checks for the existence of `bc-fips*` jars under the `lib/` folder and accordingly adds system props to ensure opensearch enforces FIPS-approved security policy. bc-fips folders are not currently included in the default distribution, and need to be built in from source by creating a distribution with the FIPS build param (`-Pcrypto.standard=FIPS-140-3`).

When the min distribution and default distribution start building with `-Pcrypto.standard=FIPS-140-3` by default, then it would mean that this block is always hit and FIPS approved mode would be enforced. We do not (yet) want that to be default and instead rely on a cluster administrator to explicitly enable that. 

This PR introduces an env var (`OPENSEARCH_FIPS_MODE`) that a cluster admin can use to enable it.

For example

```
OPENSEARCH_FIPS_MODE=true ./bin/opensearch
```

### Related Issues
Resolves https://github.com/opensearch-project/opensearch-build/issues/5979

### Check List

### DIFF
--- a/distribution/src/bin/opensearch-env.bat
+++ b/distribution/src/bin/opensearch-env.bat
@@ -33,10 +33,19 @@ for %%I in ("%OPENSEARCH_PATH_CONF%..") do set OPENSEARCH_PATH_CONF=%%~dpfI
 REM FIPS mode is runtime-configured via env var (default: false)
 if "%OPENSEARCH_FIPS_MODE%"=="" set "OPENSEARCH_FIPS_MODE=false"
 
-REM Enable only if value equals "true" (case-insensitive)
+rem Check if any bc-fips jar exists on classpath
+rem run in FIPS JVM if jar is found
+set "FOUND_BC_FIPS="
+if exist "%OPENSEARCH_HOME%\lib\bc-fips*.jar" (
+    set "FOUND_BC_FIPS=true"
+)
+
+REM Enable only if value equals "true" (case-insensitive) AND BC-FIPS JAR is present
 if /I "%OPENSEARCH_FIPS_MODE%"=="true" (
-    echo FIPS mode enabled (OPENSEARCH_FIPS_MODE=%OPENSEARCH_FIPS_MODE%), setting JVM options.
-    set "OPENSEARCH_JAVA_OPTS=-Dorg.bouncycastle.fips.approved_only=true -Djava.security.properties=""%OPENSEARCH_PATH_CONF%\fips_java.security"" %OPENSEARCH_JAVA_OPTS%"
+    if "%FOUND_BC_FIPS%"=="true" (
+        echo FIPS mode enabled, setting JVM options.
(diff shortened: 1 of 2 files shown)


cwperks:
@beanuwave let me know what you think of this change. I think we should change the check to something different and not have it be checking for the existence of bcFips jars under the `lib/` folder.

If the change in this PR looks ok, then I will raise a docs PR to accompany this.

beanuwave:
@cwperks I like this approach - it's one step closer to having `bcfips.jar` included on the class-path. However, I'm missing the trigger part. `OPENSEARCH_FIPS_MODE` would likly be set by gradle build script or docker's `entrypoint.sh`. All tests would need to pass when building with the `-Pcrypto.standard=FIPS-140-3` parameter.

cwperks:
@beanuwave agreed on running the entire gradle check against a distro built with `-Pcrypto.standard=FIPS-140-3`, but the scope of this PR is more focused on the logic within `opensearch-env` which then adds the system props to enforce that the OpenSearch process is run with FIPS approved algorithms.

distribution/src/bin/opensearch-env.bat
```diff
@@ -30,12 +30,13 @@ if not defined OPENSEARCH_PATH_CONF (
...
 
-rem Check if any bc-fips jar exists on classpath
-rem run in FIPS JVM if jar is found
-set "FOUND_BC_FIPS="
-if exist "%OPENSEARCH_HOME%\lib\bc-fips*.jar" (
-    echo BouncyCastle FIPS library found, setting FIPS JVM options.
-    set OPENSEARCH_JAVA_OPTS=-Dorg.bouncycastle.fips.approved_only=true -Djava.security.properties="%OPENSEARCH_PATH_CONF%\fips_java.security" %OPENSEARCH_JAVA_OPTS%
+REM FIPS mode is runtime-configured via env var (default: false)
+if "%OPENSEARCH_FIPS_MODE%"=="" set "OPENSEARCH_FIPS_MODE=false"
+
```
reta, line 35: @cwperks I would keep the `if exist "%OPENSEARCH_HOME%\lib\bc-fips*.jar" (` check, people do tailor distributions (for whatever reasons), if the `bc-fips` libraries are not present, the FIPS related settings are noop essentially, we should fail to start

cwperks: ok, to make sure we are on the same page check for both 1) jar existence and 2) `OPENSEARCH_FIPS_MODE=true`. Currently its just checking for 1) and if we start building the min distro with `-Pcrypto.standard=FIPS-140-3` by default then it will cause issues because these jars will always be present.

reta: Correct, I think we could make this check into Bootstrap later on, but the simplest way for now is to check both

cwperks: Added the additional checks back into this file and the batch file version.

reta (approved):

Issue #19638, opened 2025-10-14
[Feature Request][gRPC] Move primitive array types out of source


msfroh:
### Is your feature request related to a problem? Please describe

Adding gRPC support has been great for performance and interoperability on OpenSearch.

That said, from @karenyrx's presentation and OpenSearchCon NA, it looks like a lot of the performance gain comes from combining gRPC/Protobuf with SMILE encoding for document source, whether as part of an `IndexRequest` within a `_bulk` call or as part of the `SearchHit`s returned from a search request. That works, but SMILE is kind of a weird format that's not widely used.

The performance benefit of SMILE largely comes from the fact that the JSON representation of numeric arrays is absolutely terrible. Using 1 byte per base-10 digit, plus a comma between values generally makes JSON at least 2x less efficient compared to a typical binary representation.

### Describe the solution you'd like

I propose that we add an optional `Map<String, Object> fieldValues` field to the `IndexRequest` class to accompany the document source on the way into OpenSearch. These values would generally be primitive arrays and/or other binary data types. (You *could* pass any field, I suppose, but you wouldn't see much benefit for text/keyword/single-valued numeric fields where the JSON representation isn't that bad.) The idea is that these fields are not encoded as any kind of XContent. **Note that this also means that regular REST APIs won't see any benefit, since they send their whole request as XContent.**

On the way in, the `fieldValues` field should get copied from the `IndexRequest` into the `SourceToParse`, which gets passed to the `DocumentMapper#parse` method. Somewhere in there, we should copy these fields into the `ParsedDocument`.

This `fieldValues` field would then be part of the `Protobuf` representation of `IndexRequest`, which means we can pass the values as raw Protobuf, without any kind of additional XContent encoding.

Since these fields are not recorded as part of the document source, you'd need to use a mechanism like derived source to read them out if you want them in your `SearchHit`s. I don't think we need to modify `SearchHit`, since we already have the `documentFields` object, which returns fields outside of source and can use any arbitrary `Object` as values.

### Related component

Clients

### Describe alternatives you've considered

[...]

PR #20635, opened 2026-02-15
[Indexing] Add support of "extra fields" outside _source indexing

### LINKED ISSUE #20765: [Feature Request] Add support of "extra fields" outside _source indexing

Currently, document fields are ingested from the request source payload via the _XContent_ parsing pipeline before being consumed by field mappers.

For high-performance ingestion scenarios (e.g., protobuf/gRPC), we need the ability to pass additional typed field values alongside the document source during indexing.

This issue proposes extending _IndexRequest_ / _SourceToParse_ to carry optional _ExtraFieldValues_ so selected mappers can consume them directly, bypassing JSON parsing for those fields and reducing CPU/allocation overhead. This improves indexing throughput for large structured fields such as dense vectors.

Additional field values would come from _ExtraFieldValues_

### Related component

Indexing

_No response_

### Additional context

[[Feature Request][gRPC] Move primitive array types out of source](https://github.com/opensearch-project/OpenSearch/issues/19638)](https://github.com/opensearch-project/OpenSearch/issues/19638)

laminelam:
Resolves #20765  

This PR introduces _ExtraFieldValues_, a side-channel mechanism for supplying field values outside of __source_ during indexing.

This change is part of a broader effort to enhance gRPC indexing, enabling primitive vector values to be indexed as non-_source data for improved ingestion performance.

**Key points:**

- Adds ExtraFieldValues support to IndexRequest and UpdateRequest (doc and upsert paths).

- Enables mappers to consume values directly, bypassing JSON parsing.

- Currently supports float arrays only (primitive and packed little-endian representations).

- Transport serialization is version-gated for backward compatibility.

- For now, only doc-based updates and upserts are supported (script not supported).

 
## Benchmark

Benchmarked ingest throughput across increasing vector dimensions using four representations:

- JSON (float array)
- JSON (binary/base64)
- SMILE  
- ExtraFieldValues (non __source_ side-channel)

Across all tested dimensions:

JSON float arrays show the lowest throughput and degrade rapidly as vector size increases.

JSON (binary/base64) performs significantly better than float arrays.

SMILE and JSON(base64) are generally comparable, with SMILE often providing significant improvements at higher dimensions.

ExtraFieldValues consistently provides the highest throughput and scales best with increasing vector size.

At higher dimensions (e.g., 1024+), ExtraFieldValues shows a clear and consistent advantage over the _source-based approaches, **achieving up to 60× higher throughput compared to JSON arrays**.

These results indicate that bypassing __source_ parsing for large vector payloads can substantially improve ingest throughput.

Note: ExtraFieldValues was tested with the _PackedFloatArray_ option (packed little-endian floats)

extra_fields

### Ingest Throughput (baseline = JSON float array)
 
docs/sec (X× faster with json array as base)

| Dim  | JSON (array) | JSON (base64/binary) | SMILE (binary) | ExtraFieldValues |
|------|--------------|----------------------|----------------|------------------|
| 8    | 19,619 (1.0×) | 30,817 (1.57×) | 27,405 (1.40×) | 26,793 (1.37×) |
| 64   | 6,572 (1.0×)  | 23,637 (3.60×) | 28,010 (4.26×) | 26,284 (4.00×) |
| 128  | 3,649 (1.0×)  | 19,428 (5.32×) | 14,375 (3.94×) | 25,899 (7.10×) |
| 512  | 1,132 (1.0×)  | 9,723 (8.59×)  | 12,532 (11.07×) | 20,830 (18.40×) |
| 768  | 778 (1.0×)    | 6,883 (8.85×)  | 10,424 (13.40×) | 18,443 (23.70×) |
| 1024 | 591 (1.0×)    | 5,630 (9.53×)  | 8,453 (14.31×)  | 17,663 (29.90×) |
| 1536 | 399 (1.0×)    | 3,951 (9.89×)  | 6,061 (15.18×)  | 15,398 (38.56×) |
| 2048 | 299 (1.0×)    | 3,034 (10.14×) | 4,659 (15.56×)  | 14,887 (49.74×) |
| 3072 | 196 (1.0×)    | 2,060 (10.50×) | 3,058 (15.59×)  | 11,840 (60.34×) |

 

### Related Issues

https://github.com/opensearch-project/OpenSearch/issues/19638

### Check List


laminelam:
Hi @msfroh 

As [per our conversation](https://github.com/opensearch-project/OpenSearch/issues/19638#issuecomment-3774722009), this is the non grpc part of the contribution. 

The contribution is split across multiple PRs to simplify review and maintain separation of concerns.

server/src/main/java/org/opensearch/action/index/IndexRequest.java
```diff
@@ -153,6 +155,9 @@ public IndexRequest(@Nullable ShardId shardId, StreamInput in) throws IOExceptio
         id = in.readOptionalString();
         routing = in.readOptionalString();
         source = in.readBytesReference();
+        extraFieldValues = in.getVersion().onOrAfter(Version.V_3_6_0)
+            ? Objects.requireNonNullElse(in.readOptionalWriteable(ExtraFieldValues::new), ExtraFieldValues.EMPTY)
+            : ExtraFieldValues.EMPTY;
```
msfroh: Can you replace the ternary operator with an `if/else`? It's mostly so that when we remove the version check in a future version (either 4.0 or 5.0), the resulting diff is a little cleaner.

server/src/main/java/org/opensearch/index/mapper/extrasource/FloatArrayValue.java
```diff
@@ -0,0 +1,67 @@
...
+
+package org.opensearch.index.mapper.extrasource;
+
+import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+
+import java.io.IOException;
+
+public non-sealed interface FloatArrayValue extends ExtraFieldValue {
```
msfroh, line 22: I'm wondering if it makes sense to add both `PackedFloatArray` and `PrimitiveFloatArray`?

Is the idea that clients can choose which kind of encoding they want to use when they send a float array? If that's the case, I'm wondering if we can convert to whichever one is better. That conversion can probably happen at the gRPC layer. I'm just thinking that we can remove some code if we only support one float array representation. Also, we could pare down some of the interface methods like `isPackedLE`.

msfroh (commented):
Thanks @laminelam ! The change overall looks good to me. I left a couple of questions/suggestions.

msfroh:
Looks like we're getting some errors for missing JavaDoc on public classes:

```
/home/runner/work/OpenSearch/OpenSearch/server/src/main/java/org/opensearch/index/mapper/extrasource/BytesValue.java:17: error: BytesValue (record): javadocs are missing
public record BytesValue(BytesReference bytes) implements ExtraFieldValue {
       ^
/home/runner/work/OpenSearch/OpenSearch/server/src/main/java/org/opensearch/index/mapper/extrasource/ExtraFieldValue.java:18: error: ExtraFieldValue (interface): javadocs are missing
public sealed interface ExtraFieldValue permits FloatArrayValue, BytesValue {
              ^
/home/runner/work/OpenSearch/OpenSearch/server/src/main/java/org/opensearch/index/mapper/extrasource/ExtraFieldValue.java:21: error: Type (enum): javadocs are missing
    enum Type {
    ^
/home/runner/work/OpenSearch/OpenSearch/server/src/main/java/org/opensearch/index/mapper/extrasource/FloatArrayValue.java:17: error: FloatArrayValue (interface): javadocs are missing
public non-sealed interface FloatArrayValue extends ExtraFieldValue {
                  ^
/home/runner/work/OpenSearch/OpenSearch/server/src/main/java/org/opensearch/index/mapper/extrasource/ExtraFieldValues.java:21: error: ExtraFieldValues (class): javadocs are missing
public final class ExtraFieldValues implements Writeable {
```

Do these all need to be `public`? I guess we the ones above probably do, since `ExtraFieldValues` and `ExtraFieldValue` are both referenced from `DocumentParser`. Then the field mappers that use them will need `BytesValue` or `FloatArrayValue` to be able to extract the actual values. We might be able to make the `FloatArrayValue` implementations package-private, though (if we want to keep both of them).

server/src/main/java/org/opensearch/index/mapper/extrasource/FloatArrayValue.java
laminelam: Thanks @msfroh for the review.

The two representations move the cost to different parts of the indexing pipeline: client --> coordinator -->  shard  -->  mapper.

With **PrimitiveFloatArray**, floats are serialized one by one during transport (client to coordinator and coordinator to shard). The mapper then receives an already decoded float array and does no additional work. Still, the elements are serialized twice during the full cycle.

With **PackedFloatArray**, the vector is transported as a packed byte sequence. This avoids the float serialization loops during transport, but the mapper needs to decode the floats when they are accessed.

So the trade off is where the CPU cost happens: _PrimitiveFloatArray_ makes transport (including serialization/deserialization) heavier but the mapper cheaper, while _PackedFloatArray_ makes transport cheaper but adds a small decode cost at the mapper. 

For large vectors the transport serialization tends to dominate, which is why the packed representation performs much better in the benchmarks.

So, we want to allow clients to choose the trade off that fits their needs: simplicity (float[]) vs optimized transport (packed)

See the benchmark [here](https://github.com/opensearch-project/OpenSearch/issues/19638#issuecomment-3774529669) 
(vector = PackedFloatArray  VS float = PrimitiveFloatArray)

laminelam:
> Do these all need to be `public`? I guess we the ones above probably do, since `ExtraFieldValues` and `ExtraFieldValue` are both referenced from `DocumentParser`. Then the field mappers that use them will need `BytesValue` or `FloatArrayValue` to be able to extract the actual values. We might be able to make the `FloatArrayValue` implementations package-private, though (if we want to keep both of them).

Yes, was planning to add some javadoc before the PR is merged. 
You're right `FloatArrayValue` implementations could be made private (with some public factory methods). Will do.

msfroh:
@laminelam -- Can you rebase this onto the latest `main`? 

@andrross has done some amazing work recently on fixing flaky tests, so I suspect that the test failures from your earlier commits should pass now.

laminelam:
> ❌ Gradle check result for [ece40b5](https://github.com/opensearch-project/OpenSearch/commit/ece40b58bad2477045a5dcf63d409006901ff322): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/74216/)
> 



@msfroh

TimeoutException in Kafka plugin

server/src/main/java/org/opensearch/action/index/IndexRequest.java
```diff
@@ -153,6 +155,11 @@ public IndexRequest(@Nullable ShardId shardId, StreamInput in) throws IOExceptio
         id = in.readOptionalString();
         routing = in.readOptionalString();
         source = in.readBytesReference();
+        if (in.getVersion().onOrAfter(Version.V_3_6_0)) {
```
msfroh: Ahh... I noticed that the recent failures were because we were getting `No version type match [-3]`. That's coming from https://github.com/opensearch-project/OpenSearch/blob/0d54c1600b16081c15902a095111a9a530767099/server/src/main/java/org/opensearch/index/VersionType.java#L323-L332

And the reason is because it's trying to read `extraFieldValues` during a backwards-compatibility test with OpenSearch 3.6.

So, the three places in your diff that say `V_3_6_0` need to be updated to `V_3_7_0`.

laminelam: Hi @msfroh 

Updated the version number.
Any chance you can merge the PR at your earliest convenience as I have other PRs to push on top of this one?

msfroh (approved):

msfroh:
@laminelam -- there was a check failure related to the presence of the CHANGELOG.md file. Since this PR started, we've removed the need for manual CHANGELOG updates (which is nice, since they were a frequent cause of merge conflicts). I pushed a commit to remove the CHANGELOG from this PR: https://github.com/opensearch-project/OpenSearch/pull/20635/commits/478183b1fcbe666a0cb859ae594e8e47dbb9ca81

Hopefully the Gradle check step will pass and I'll go ahead and merge this.

msfroh:
Hmm... way too many flaky test failures, which is weird, since they're doing a lot better lately.

I rebased onto the latest `main` and force-pushed your branch. Hopefully that should help.

laminelam:
Thank you @msfroh 
Really appreciate your time on this.

navneet1v:
Hi @laminelam,
This is a clever way to avoid the large float arrays being getting stored in `_source` and then use LE format to encode to vectors. I have few questions here:
1. How we are solving the re-indexing problem here? since vectors are not present in the _source. 
2. When node to node recovery happens since all the data is not in the _source the recovery will never ingest vectors back.
3. How we are handling the nested fields here where some of the child documents doesn't have vectors in it.
4. The gains which we are observing as 60x for 3072D, I am thinking if we added the vectors back to `_source` what degradation it will lead? We might need vectors in _source because of above points. 

I am working on feature in KNN plugin where vectors can be retrieved as doc_values rather than getting them in _source, where I am adding the similar LE format for the vectors to be returned in JSON. https://github.com/opensearch-project/k-NN/issues/3315#issuecomment-4435580126

and as a next step I was thinking to implement the binary format in the index for JSON with rest(prototype already present in my local :P ). and in your benchmarks I can see you got 8x improvements for 768D, I will see if I can repro it.

cc: @msfroh

Issue #20642, opened 2026-02-17
[BUG] No index warmer on replica shards in segment replication leading to cold start issues


Kartik031299:
### Describe the bug

We are using Opensearch version 2.15 with segment replication enabled. We have seen some cold start issues happening with replica shards where first search request after each refresh seems to be taking bit longer and subsequent requests were faster as compared to first request after each refresh. This behavior was observed especially for replica shards and was not seen on primary shard.

Just to mentioned: We have different indices configured in our clusters and some of them also have parent-child join field in mapping with global_eager_ordinals set to true.
We found out via local debugging that global ordinal map is being built and caches in fielddata cache on primary shard after every refresh , however on replica shard during replication checkpoint processing , no global ordinal map is being built

As a result during first query after refresh on replica nodes, there is an additional overhead of building global ordinal map and storing it in fielddata cache which can be further reused by subsequent requests however first request after refresh taking a hit  

Even after taking a look at latest documentation: https://docs.opensearch.org/latest/mappings/mapping-parameters/eager_global_ordinals/, it does not indicate above behavior to be expected with segment replicated clusters

When we investigated further, we found out that NRTReplicationEngine is being used for replica shards as compared to InternalEngine being used for primary shards
Further we saw that NRTReplicationEngine is using NRTReaderManager where refreshIfNeeded method is being invoked and we don't see any external refresh listeners triggering index warmer. 

On checking internal listeners which were being added within NRTReplicationEngine constructor for reader manager, we didn't see any index warmer refresh listener being registered 

```
            this.completionStatsCache = new CompletionStatsCache(() -> acquireSearcher("completion_stats"));
            this.readerManager = readerManager;
            this.readerManager.addListener(completionStatsCache);
            // NRT Replicas do not have a concept of Internal vs External reader managers.
            // We also do not want to wire up refresh listeners for waitFor & pending refresh location.
            // which are the current external listeners set from IndexShard.
            // Only wire up the internal listeners.
            for (ReferenceManager.RefreshListener listener : engineConfig.getInternalRefreshListener()) {
                this.readerManager.addListener(listener);
            }
```

We were not able to find any related open Github issues related to above problem and was not able to understand reason of not adding IndexWarmer to replica shard listener. We will be also happy to raise a patch to add support but first wanted to understand whether this is something which was done intentional due to relevant  reasons or not 

### Related component

Search:Performance

### To Reproduce

[...]

PR #20650, opened 2026-02-17
Add support of IndexWarmer for replica shards with segment replication enabled

Kartik031299:
### Description
Added WarmerRefreshListener to trigger Engine.Warmer on every refresh invoked for NRTReplicationReaderManager whenever segments are updates on replica shards after processing replication checkpoint
1. This gives advantage of pre-loading global ordinals in field data cache to avoid cold start penalty with replica shards after new refresh

### Related Issues
Bug: #20642 

### Check List


Kartik031299:
> ❌ Gradle check result for [d3bc827](https://github.com/opensearch-project/OpenSearch/commit/d3bc827f19161b9a33ea1934648fc9c42cf93048): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/71457/)
> 

Flaky tests for ClusterShardLimitIT found
image

There is known expected flakiness for this test class: https://metrics.opensearch.org/_dashboards/app/dashboards#/view/e5e64d40-ed31-11ee-be99-69d1dbc75083?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-90d,to:now))&_a=(description:'',filters:!(),fullScreenMode:!f,options:(hidePanelTitles:!f,useMargins:!t),query:(language:kuery,query:ClusterShardLimitIT),timeRestore:!t,title:'OpenSearch%20Gradle%20Check%20Report',viewMode:view)

Kartik031299:
> ❌ Gradle check result for [a632e02](https://github.com/opensearch-project/OpenSearch/commit/a632e02fd166579519deb75e003c1030dbbb5f68): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/71466/)
> 

CI again failed due to flaky test: org.opensearch.cluster.coordination.NodeJoinLeftIT.testClusterStabilityWhenClusterStatePublicationLagsWithLongRunningListenerOnApplierThread
image

server/src/main/java/org/opensearch/index/engine/NRTReplicationEngine.java
```diff
@@ -560,4 +565,50 @@ private DirectoryReader getDirectoryReader() throws IOException {
...
+                    warmer.warm(reader);
+                } catch (Exception e) {
+                    if (isEngineClosed.get() == false) {
+                        logger.warn("failed to warm reader replica", e);
+                    }
+                } finally {
+                    readerManager.release(reader);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Kartik031299: Addressed issue

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
```diff
@@ -0,0 +1,316 @@
...
+
+        // Capture warmer stats before force merge
+        long replicaWarmerTotalBeforeMerge = getWarmerTotalForShardType(
+            client().admin().indices().prepareStats(INDEX_NAME).clear().setWarmer(true).get().getShards(),
+            false
+        );
+        long primaryWarmerTotalBeforeMerge = getWarmerTotalForShardType(
+            client().admin().indices().prepareStats(INDEX_NAME).clear().setWarmer(true).get().getShards(),
+            true
+        );
```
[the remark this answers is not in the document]
Kartik031299: Addressed

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
```diff
@@ -0,0 +1,316 @@
...
+            .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 1)
+            .put(IndexModule.INDEX_QUERY_CACHE_ENABLED_SETTING.getKey(), false)
+            .put(IndexMetadata.SETTING_REPLICATION_TYPE, ReplicationType.SEGMENT)
+            .build();
+    }
+
+    @Before
+    private void setup() {
+        internalCluster().startClusterManagerOnlyNode();
+    }
```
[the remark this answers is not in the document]
Kartik031299: Addressed

server/src/main/java/org/opensearch/index/engine/NRTReplicationEngine.java
```diff
@@ -560,4 +565,56 @@ private DirectoryReader getDirectoryReader() throws IOException {
...
+                        }
+                    } finally {
+                        readerManager.release(reader);
+                    }
+                } catch (IOException e) {
+                    if (isEngineClosed.get() == false) {
+                        logger.warn("failed to acquire reader for warming on replica", e);
+                    }
+                }
+            }
```
Bukhtawar, line 617: Do we know from benchmarks how long does the warmer take? I am worried it might end up blocking the refresh thread for long

Kartik031299: @Bukhtawar I am not sure where can I get this data. Maybe we can use API: https://docs.opensearch.org/latest/api-reference/index-apis/stats/ to get time taken by index warmer. Can you let me know how can we get this data ?

Bukhtawar: Maybe just add a timer to time the warm operation, and emit logs if stats aren't helpful

Kartik031299: @Bukhtawar I see there are some metrics already being published within ShardIndexWarmerService like total time in millis and total number of executions for warmer that can get us an idea of how long it is taking to run on average. Can we use those metrics as they are already being published and exposed via nodes stats API ? Can we run benchmark testing on this PR and get this data ?
I don't know whether publishing any new metric here will help

test/framework/src/main/java/org/opensearch/index/engine/EngineTestCase.java
```diff
@@ -864,20 +864,23 @@ public EngineConfig config(
             refreshListener,
             indexSort,
             globalCheckpointSupplier,
-            globalCheckpointSupplier == null ? null : () -> RetentionLeases.EMPTY
+            (Engine.Warmer) null
```
ashking94: nit - why do we need to type cast null?

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * Integration tests that verify index warming (e.g., eager global ordinals loading) works correctly
+ * on NRT replica shards during segment replication. This validates the end-to-end flow where
+ * index warmer gets triggered when new segments arrive on replicas via segment replication.
+ *
+ * <p>Uses a join field ({@code parentToDocs}) with {@code eager_global_ordinals: true}
+ * to exercise the warming path during segment replication.</p>
+ */
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.TEST, numDataNodes = 0)
+public class SegmentReplicationReplicaIndexWarmerIT extends OpenSearchIntegTestCase {
```
ashking94: This does not appear to be the correct package for the segment replication related ITs. Can you move this to same location where other segrep ITs exists?

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
```diff
@@ -0,0 +1,263 @@
...
+
+            long primaryWarmerTotal = 0;
+            long replicaWarmerTotal = 0;
+
+            for (ShardStats shardStats : shardStatsArray) {
+                WarmerStats warmerStats = shardStats.getStats().getWarmer();
+                assertNotNull("Warmer stats should not be null", warmerStats);
+
+                if (shardStats.getShardRouting().primary()) {
+                    primaryWarmerTotal += warmerStats.total();
```
ashking94: Should there be an equal check for primary and replica warmer invocation count?

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
```diff
@@ -0,0 +1,263 @@
...
+                } else {
+                    replicaWarmerTotal += warmerStats.total();
+                }
+            }
+
+            assertThat("Warmer should have been invoked on primary shard", primaryWarmerTotal, greaterThan(0L));
+            assertThat("Warmer should have been invoked on replica shard via WarmerRefreshListener", replicaWarmerTotal, greaterThan(0L));
```
ashking94: there is some duplication which can be reduced here.

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
```diff
@@ -0,0 +1,263 @@
...
+                "Warmer should have been invoked on primary shard after force merge",
+                primaryTotal,
+                greaterThan(primaryWarmerTotalBeforeMerge)
+            );
+            assertThat(
+                "Warmer invocation count on replica should increase after force merge replication",
+                replicaTotal,
+                greaterThan(replicaWarmerTotalBeforeMerge)
+            );
+        });
```
ashking94: see a scope to reduce duplication here.

server/src/internalClusterTest/java/org/opensearch/indices/replication/SegmentReplicationReplicaIndexWarmerIT.java
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * Integration tests that verify index warming (e.g., eager global ordinals loading) works correctly
+ * on NRT replica shards during segment replication. This validates the end-to-end flow where
+ * index warmer gets triggered when new segments arrive on replicas via segment replication.
+ *
+ * <p>Uses a keyword field with {@code eager_global_ordinals: true}
+ * to exercise the warming path during segment replication.</p>
+ */
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.TEST, numDataNodes = 0)
+public class SegmentReplicationReplicaIndexWarmerIT extends SegmentReplicationBaseIT {
```
ashking94, line 32: Why do we need 2 similar ITs for checking the same behaviour? Since the flows for segment replication is same regardless whether parent join exists or not, I would probably prefer to have 1 test in here - which tests both that you intend to.

server/src/main/java/org/opensearch/index/engine/NRTReplicationEngine.java
ashking94: I believe we should have this change done regardless as currenly the replica shows cold start problem when a request lands on it in comparison to primary. Moreover, this probably appears to be a miss than intentional since primary already does call IndexWarmer to pre-load eager global ordinals.

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
Kartik031299: I think it got duplicated here by mistake, will remove it from here

test/framework/src/main/java/org/opensearch/index/engine/EngineTestCase.java
Kartik031299: yeah don't need it, will remove it from here

server/src/internalClusterTest/java/org/opensearch/indices/replication/SegmentReplicationReplicaIndexWarmerIT.java
Kartik031299: @ashking94 intention was to only have 1 IT (this one). Deletion of other one was missed, will do it

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
Kartik031299: will use utility method: `getWarmerTotalPerShardType` here

modules/parent-join/src/internalClusterTest/java/org/opensearch/join/query/SegmentReplicationReplicaIndexWarmerIT.java
Kartik031299: @ashking94 I think there can be cases where segments are also merged on primary (leading to triggering refresh) and those segments have not been yet transferred to replica node , so there is a chance of mismatch happening
Also there is a warmer being triggered on primary just after creation as well, which is currently not case for replica shard
So I think equal count might not be true across primary and replica shard

server/src/main/java/org/opensearch/index/engine/NRTReplicationEngine.java
Kartik031299: We will be doing benchmark testing on this change internally and will share results later

Kartik031299: We did perform benchmark testing for big5 and so workloads. No regression seen in indexing and search latencies. We also captured index warmer stats for so workload and we can see that now index warmer is being triggered on replica nodes and average time per warmer trigger is around 0.02 ms (which should be fine)

  Warmer Stats

  [PRIMARY]
    Total warmer triggers : 76572
    Total time            : 3324 ms
    Avg time per trigger  : 0.04 ms
    Currently running     : 0

  [REPLICA]
    Total warmer triggers : 75844
    Total time            : 1643 ms
    Avg time per trigger  : 0.02 ms

  [OVERALL]

    Total warmer triggers : 152416
    Total time            : 4967 ms
    Avg time per trigger  : 0.03 ms
    Currently running     : 0

PR #20667, opened 2026-02-18
Change result type to accept multiple batches

alchemist51:
### Description

### Related Issues

### Check List

### DIFF
--- a/plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
+++ b/plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
@@ -68,6 +68,7 @@ import org.opensearch.search.fetch.FetchSubPhase;
 import org.opensearch.search.fetch.subphase.FetchSourceContext;
 import org.opensearch.search.internal.ReaderContext;
 import org.opensearch.search.internal.SearchContext;
+import org.opensearch.datafusion.search.DfResult;
 import org.opensearch.search.internal.ShardSearchRequest;
 import org.opensearch.search.lookup.SourceLookup;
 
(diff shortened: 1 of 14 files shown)


plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
```diff
@@ -244,7 +244,11 @@ public void executeQueryPhase(DatafusionContext context) {
                                 fieldValues[i] = fieldVector.getObject(i);
                             }
                         }
-                        finalRes.put(fieldName, fieldValues);
+                        if(finalRes.containsKey(fieldName)) {
+                            finalRes.get(fieldName).addAll(Arrays.asList(fieldValues));
+                        } else {
+                            finalRes.put(fieldName, new ArrayList<>(Arrays.asList(fieldValues)));
```
mgodwan, line 251: Can we avoid this copy? Use a decorating list maybe? We can measure the perf impact as well once to see if overhead of copy is high (at least when only 1 batch is fetched) enough to justify the following:

```
public class FlattenedList<T> extends AbstractList<T> {
    private final Object[][] array;
    private final int totalSize;

    public FlattenedList(Object[][] array) {
        this.array = array;
        int count = 0;
        for (Object[] row : array) {
            count += row.length;
        }
        this.totalSize = count;
    }

    public T get(int index) {
       // Compute the index in the row-major 2d array using binary search
    }

    public int size() {
        return totalSize;
    }
}
```

alchemist51: I was able to remove this copy by changing the type of fieldValues to ArrayList. Simplified the code as well now.

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DatafusionEngine.java
```diff
@@ -303,20 +307,23 @@ public void collect(RecordBatchStream value) {
                         for (Field field : root.getSchema().getFields()) {
                             String fieldName = field.getName();
                             FieldVector fieldVector = root.getVector(fieldName);
-                            Object[] fieldValues = new Object[fieldVector.getValueCount()];
+                            List<Object> fieldValues = new ArrayList<>();
```
alchemist51: Wanted to make it fixed size from getValueCount, it is throwing exception in setting it in few cases. Will update on this.

alchemist51: Updated.

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/search/DatafusionContext.java
```diff
@@ -86,7 +86,7 @@ public class DatafusionContext extends SearchContext {
     private final IndexService indexService;
     private final QueryShardContext queryShardContext;
     private DatafusionQuery datafusionQuery;
-    private Map<String, Object[]> dfResults;
+    private Map<String, List<Object>> dfResults;
```
expani: Should we consider putting this inside it's own class, something like 

```
interface DataFusionResult {
   void addResult(Object[] values);
   Object getValueAt(String key, int index);
}

class InMemoryDataFusionResult implements DataFusionResult { 
  Map<String, List<Object>> dfResults;
} 
```

If the initial implementation did this ( partially on me ), the change of `Object[]` -> `List<Object[]>` would be contained within this class and not require changes across all aggregators. 

In future, if we use a different data structure it can be easily replaced as well.

mgodwan (approved):

Issue #20644, opened 2026-02-17
[RFC] OpenSearch Engine: Pluggable Component Bundling


mgodwan:
## Summary

OpenSearch today relies exclusively on Apache Lucene for data storage and query execution. While Lucene excels at full-text search and near real-time indexing, this tight coupling limits our ability to support workloads optimized for different storage formats (covered here in more details: https://github.com/opensearch-project/OpenSearch/issues/18416 ). Users running analytical queries on time-series data, for example, would benefit from columnar formats like Parquet that offer superior compression and scan performance. Similarly, vector similarity search and graph analytics require specialized data structures which may require modeling outside the Lucene codecs.

This document proposes the details around OpenSearch Execution Engine—a pluggable architecture that enables multiple storage formats and execution engines to coexist within OpenSearch. The design maintains backward compatibility with existing Lucene-based indices while providing clear extension points for new formats like Parquet, ORC, Lance, etc. We focus on three key areas: (1) the plugin interfaces that enable format extensibility and their bundling, (2) how these plugins integrate with OpenSearch's indexing and data availability flows, and (3) how the architecture composes with existing engines like InternalEngine, IngestionEngine, NRTReplicationEngine, etc..

## Goals

1. Enable pluggable storage formats beyond Lucene (Parquet, ORC, Arrow, custom formats) and any auxiliary data structures to facilitate faster search, without degrading on the memory overhead during document indexing.
2. Enable ingesting documents to the choice of index (multi format or Lucene) on the same distribution to facilitate migration. We can choose to apply artificial blocks to restrict the type of indexing supported on a cluster. 
3. Cross Version Compatibility: The design should take care of ensuring any changes in data-format are supported to be read across next N OpenSearch versions. Also, during a rolling/blue-green deployment, transport communication between nodes should continue to work without any downtime [transport is not in scope of below discussion and would be covered later]. For existing indices, we may require re-indexing to use newer proposed implementation in this document during the migration.
4. Provide clear plugin interfaces with minimal implementation overhead. Reuse existing OpenSearch constructs and integration touch-points wherever possible to ensure any development happening around the peripheral/common features is applicable to the new and existing engines.

## Current State: Lucene Integration Layers

OpenSearch's architecture is deeply integrated with Lucene across three primary layers. Understanding these integration points is essential for designing the abstraction layer.

### Layer 1: Engine Integration

[...]

PR #20675, opened 2026-02-19
Add indexer interface for shard interaction with underlying engines

mgodwan:
## Summary

This PR introduces an `Indexer` interface that abstracts the interaction between `IndexShard` and the underlying storage engine. Today, `IndexShard` holds a direct reference to `Engine` and calls engine methods throughout its lifecycle — tightly coupling shard-level coordination logic to a single engine implementation. This change decouples that relationship by introducing `Indexer` as the primary contract through which `IndexShard` performs document operations, manages sequence numbers, controls lifecycle events (flush, refresh, force merge), and queries statistics. A concrete implementation, `EngineBackedIndexer`, wraps the existing `Engine` to preserve full backward compatibility. This is a foundational step toward enabling pluggable and composite engine architectures in OpenSearch, as described in RFC [#20644](https://github.com/opensearch-project/OpenSearch/issues/20644).

---

## Motivation

OpenSearch's engine layer is responsible for indexing, searching, and managing the lifecycle of data on a shard. As the project evolves toward supporting composite engines (e.g., a primary Lucene engine paired with a secondary vector or columnar engine), the current tight coupling between `IndexShard` and `Engine` becomes a bottleneck. There is no clean abstraction boundary that allows a shard to interact with multiple engine implementations or swap engine behavior without modifying `IndexShard` directly.

This PR establishes that abstraction boundary.

---

## Architecture: Before vs. After

**Before (Tightly Coupled)**

```
┌─────────────────────────────────────────────┐
│                  IndexShard                 │
│                                             │
│  index()  delete()  flush()  refresh()  ... │
│       │        │       │        │           │
│       └────────┴───────┴────────┘           │
│                    │                        │
│             Direct Engine ref               │
└────────────────────┼────────────────────────┘
                     │
                     ▼
           ┌─────────────────┐
           │     Engine      │  (InternalEngine / NRTReplicationEngine / etc.)
           └─────────────────┘
```

**After (Interface-Abstracted)**

```
┌─────────────────────────────────────────────┐
│                  IndexShard                 │
│                                             │
│  index()  delete()  flush()  refresh()  ... │
│       │        │       │        │           │
│       └────────┴───────┴────────┘           │
│                    │                        │
│             Indexer interface ref           │
└────────────────────┼────────────────────────┘
                     │
                     ▼
           ┌─────────────────────┐
           │      Indexer        │  <<interface>>
           │  (LifecycleAware,   │
           │   Closeable)        │
           └──────────┬──────────┘
                      │
          ┌───────────┴────────────┐
          │                        │
          ▼                        ▼
┌──────────────────────┐   ┌──────────────────────────┐
│  EngineBackedIndexer │   │  (Future: Composite-      │
│  (wraps Engine)      │   │   IndexerImpl, etc.)      │
└──────────┬───────────┘   └──────────────────────────┘
           │
           ▼
  ┌─────────────────┐
  │     Engine      │
  └─────────────────┘
```

---

## New Class Structure

```
org.opensearch.index.engine
├── EngineBackedIndexer.java          ← Concrete Indexer wrapping Engine
├── Engine.java                       ← Minor cleanup (unintended changes removed)
│
└── exec/
    ├── Indexer.java                  ← Core interface (@ExperimentalApi)
    ├── Segment.java                  ← Segment metadata model
    ├── MergeResult.java              ← Result of a merge operation
    ├── WriterFileSet.java            ← File set produced by a writer
    └── DataFormat.java               ← Data format identifier abstraction

org.opensearch.index.shard
├── IndexShard.java                   ← Migrated to use Indexer interface
└── RemoteStoreRefreshListener.java   ← Migrated to use Indexer interface
```

---

## Indexer Interface Responsibilities

```
┌──────────────────────────────────────────────────────┐
│                    Indexer Interface                 │
├──────────────────────────────────────────────────────┤
│  Document Operations                                 │
│    index(Engine.Index)   →  Engine.IndexResult       │
│    delete(Engine.Delete) →  Engine.DeleteResult      │
│    noOp(Engine.NoOp)     →  Engine.NoOpResult        │
│    prepareIndex(...)     →  Engine.Index             │
│    prepareDelete(...)    →  Engine.Delete            │
├──────────────────────────────────────────────────────┤
│  Sequence Number & Checkpoint Management             │
│    getLocalCheckpoint()                              │
│    getMaxSeqNo()                                     │
│    advanceMaxSeqNoOfUpdatesOrDeletes(long)           │
│    currentOngoingRefreshCheckpoint()                 │
├──────────────────────────────────────────────────────┤
│  Lifecycle Operations                                │
│    flush(FlushRequest)                               │
│    refresh(String)                                   │
│    forceMerge(ForceMergeRequest)                     │
│    close()                                           │
├──────────────────────────────────────────────────────┤
│  Statistics & Monitoring                             │
│    docStats()                                        │
│    segmentStats(...)                                 │
│    mergeStats()                                      │
│    getTranslogStats()                                │
└──────────────────────────────────────────────────────┘
```

---

## EngineBackedIndexer: Delegation Model

```
IndexShard.getIndexer()
        │
        ▼
EngineBackedIndexer
        │
        │  delegates all calls to:
        ▼
    Engine (volatile ref)
        │
        ├── InternalEngine
        └── NRTReplicationEngine
```

The `engine` reference inside `EngineBackedIndexer` is volatile and mirrors the existing engine swap semantics in `IndexShard`. The `getEngine()` accessor is intentionally marked for future removal once all callers migrate away from direct engine access.

---

## Migration Path

```
Phase 1 (This PR)
─────────────────
IndexShard  ──uses──►  Indexer (via EngineBackedIndexer)
RemoteStoreRefreshListener  ──uses──►  Indexer

Phase 2 (Future)
────────────────
Deprecate & remove getEngine() from EngineBackedIndexer
Introduce SearcherInterface (tracked via TODO comments)
Deprecate direct acquireSearcher / acquireLastIndexCommit paths

Phase 3 (Future)
────────────────
Introduce CompositeIndexer for multi-engine shards
Register DataFormats via DataFormatRegistry at plugin bootstrap
```

---

## Related Issues & PRs

- RFC: [#20644 – OpenSearch Engine: Pluggable Component Bundling](https://github.com/opensearch-project/OpenSearch/issues/20644)
- Related: `mgodwan/OpenSearch#201` – Changes for supporting `DataFormatRegistry`

---

## Check List

- [x] API changes companion pull request created (pending interface stabilization)
- [ ] Public documentation issue/PR created (pending)
- [x] Changelog entry added

---

## Notes for Reviewers

- The `Indexer` interface is annotated `@ExperimentalApi` — the contract is expected to evolve as composite engine support matures.
- `EngineBackedIndexer.getEngine()` is a temporary escape hatch for callers not yet migrated; it is documented for removal.
- Several `IndexShard` methods are marked `@Deprecated` with TODO comments pending a `SearcherInterface` companion PR.
- The `applyOnEngine` utility method currently throws `IllegalStateException` for non-`EngineBackedIndexer` instances; this will be addressed as the interface matures.


server/src/main/java/org/opensearch/index/engine/EngineBackedIndexer.java
```diff
@@ -0,0 +1,362 @@
...
+import org.opensearch.indices.pollingingest.PollingIngestStats;
+import org.opensearch.search.suggest.completion.CompletionStats;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+@ExperimentalApi
+public class EngineBackedIndexer implements Indexer {
```
Bukhtawar, line 38: nit: Good to add extensive Java docs, will help to generate tests and improve efficiency through GenAI tooling

server/src/main/java/org/opensearch/index/engine/exec/Indexer.java
```diff
@@ -0,0 +1,421 @@
...
+import org.opensearch.search.suggest.completion.CompletionStats;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+
+import static org.opensearch.index.engine.Engine.HISTORY_UUID_KEY;
+
+@PublicApi(since = "1.0.0")
```
Bukhtawar: Should we keep newer interfaces Experimental we can bump them up once we freeze the interface changes

server/src/main/java/org/opensearch/index/engine/EngineBackedIndexer.java
mgodwan: Ack. adding here. Will add more with flow details as well to ensure tests can be easily simulated

server/src/main/java/org/opensearch/index/engine/exec/Indexer.java
mgodwan: Fixing it

server/src/main/java/org/opensearch/index/engine/exec/Indexer.java
```diff
@@ -0,0 +1,421 @@
...
+public interface Indexer extends LifecycleAware, Closeable {
```
Bukhtawar: Over time this interface and its implementations have bloated up I was thinking of breaking it down as below, also trying to see how we can do this without breaking compatibility
 1. IndexerEngineOperations
   Core document operations: index(), delete(), noOp()
  Operation preparation: prepareIndex(), prepareDelete()
  History operations
2. IndexerStateManager 
Sequence numbers and checkpoints
Timestamps (auto-id, write timestamps)
State queries and updates
3. IndexerLifecycleOperations 
Top-level operations: flush(), refresh(), forceMerge()
Buffer management
Throttling control
Lifecycle events
4. IndexerStatistics
Document, segment, and merge statistics
Memory usage metrics
Monitoring data

server/src/main/java/org/opensearch/index/engine/exec/Segment.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    /**
+     * Adds searchable files for a specific data format to this segment.
+     *
+     * @param dataFormat the data format identifier
+     * @param writerFileSetGroup the set of files for this data format
+     */
+    public void addSearchableFiles(String dataFormat, WriterFileSet writerFileSetGroup) {
```
Bukhtawar: Instead of Strings wondering how we can add stringer validations like 
```
public void addSearchableFiles(String dataFormat, WriterFileSet writerFileSetGroup) {
    // Validate the format is registered
    DataFormatRegistry.validate(dataFormat);
    
    dfGroupedSearchableFiles.put(dataFormat, writerFileSetGroup);
}
```

This is based off a `DataFormatRegistry` that is used as plugins load the data format plugin and perform a register at bootstrap, which can be later used for validation and enforcing constraints on mapping and else where. `DataFormatDescriptor` is where we keep properties what a data format can support(eventually can be used by mappers)
```
public static void register(DataFormatDescriptor descriptor) {
        if (descriptor == null) {
            throw new IllegalArgumentException("Data format descriptor cannot be null");
        }

        String identifier = descriptor.identifier();
        DataFormatDescriptor existing = REGISTERED_FORMATS.putIfAbsent(identifier, descriptor);

        if (existing != null) {
            throw new IllegalArgumentException(
                String.format(
                    "Data format '%s' is already registered. Existing: %s, Attempted: %s",
                    identifier,
                    existing,
                    descriptor
                )
            );
        }

        logger.info("Registered data format: {} - {}", identifier, descriptor.description());
    }
```

server/src/main/java/org/opensearch/index/engine/dataformat/MergeResult.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Result of a merge operation containing merged writer file sets.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class MergeResult {
+
+    private final Map<DataFormat, WriterFileSet> mergedWriterFileSet;
```
darjisagar7, line 25: MergeResult should also have RowIdMapping right? As the primary engine result will have that Mappings for secondary engines?

server/src/main/java/org/opensearch/index/search/MultiMatchQuery.java
```diff
@@ -340,6 +340,7 @@ static Query blendTerms(
             // best effort: add clauses that are not term queries so that they have an opportunity to match
             // however their score contribution will be different
             // TODO: can we improve this?
+
```
Bukhtawar, line 343: nit: unintented?

server/src/main/java/org/opensearch/index/shard/RemoteStoreRefreshListener.java
```diff
@@ -449,14 +451,15 @@ private boolean isRefreshAfterCommitSafe() {
 
     void uploadMetadata(Collection<String> localSegmentsPostRefresh, SegmentInfos segmentInfos, ReplicationCheckpoint replicationCheckpoint)
         throws IOException {
-        final long maxSeqNo = ((InternalEngine) indexShard.getEngine()).currentOngoingRefreshCheckpoint();
+        final long maxSeqNo = ((InternalEngine) ((EngineBackedIndexer) indexShard.getIndexer()).getEngine())
+            .currentOngoingRefreshCheckpoint();
```
Bukhtawar:
```suggestion
        final long maxSeqNo = indexShard.getIndexer().currentOngoingRefreshCheckpoint();
```

server/src/main/java/org/opensearch/index/engine/EngineBackedIndexer.java
```diff
@@ -0,0 +1,413 @@
...
+    public GatedCloseable<CatalogSnapshot> acquireSnapshot() {
+        GatedCloseable<SegmentInfos> segmentInfos = engine.getSegmentInfosSnapshot();
+        SegmentInfosCatalogSnapshot catalogSnapshot = new SegmentInfosCatalogSnapshot(segmentInfos.get());
+        return new GatedCloseable<>(catalogSnapshot, segmentInfos::close);
+    }
+
+    public Engine getEngine() {
+        return engine;
+    }
+}
```
Bukhtawar, line 393: Eventually we should get rid of this public method to avoid leaking the mutable engine state

server/src/main/java/org/opensearch/index/engine/Engine.java
```diff
@@ -2234,5 +2234,4 @@ public long getMaxSeenAutoIdTimestamp() {
      * to advance this marker to at least the given sequence number.
      */
     public abstract void advanceMaxSeqNoOfUpdatesOrDeletes(long maxSeqNoOfUpdatesOnPrimary);
-
```
Bukhtawar, line 2237: nit: unintended

server/src/main/java/org/opensearch/index/shard/RemoteStoreRefreshListener.java
Bukhtawar: I believe this we can get to eventually

mgodwan: Yes, handled this as well. Moved the case to EngineBackedIndexer for this.

server/src/main/java/org/opensearch/index/engine/EngineBackedIndexer.java
mgodwan: Yes, once classes relying on IndexShard migrate away, we can get rid of this. 
Let me add the same to javadoc for this method.

server/src/main/java/org/opensearch/index/engine/dataformat/RefreshInput.java
```diff
@@ -0,0 +1,71 @@
...
+        this.existingSegments = new ArrayList<>();
+    }
+
+    /**
+     * Sets the existing segments.
+     *
+     * @param existingSegments the list of existing segments
+     */
+    public void setExistingSegments(List<org.opensearch.index.engine.exec.coord.Segment> existingSegments) {
+        this.existingSegments = existingSegments;
```
alchemist51: nit: Should this be unmodifiable list?

server/src/main/java/org/opensearch/index/engine/dataformat/DocumentInput.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Represents a document input for adding fields and metadata to a writer.
+ *
+ * @param <T> the type of the final input representation
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public interface DocumentInput<T> extends AutoCloseable {
```
alchemist51: Should T be bound to Dataformats?

server/src/main/java/org/opensearch/index/engine/exec/Segment.java
```diff
@@ -0,0 +1,104 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.engine.exec.coord;
```
bharath-techie: nit: should this be `org.opensearch.index.engine.exec`

server/src/main/java/org/opensearch/index/engine/dataformat/DocumentInput.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Sets the primary term for this document.
+     *
+     * @param fieldName the field name
+     * @param seqNo the sequence number
+     */
+    default void setPrimaryTerm(String fieldName, long seqNo) {
+        // Default no-op implementations, override as needed
+    }
+}
```
Bukhtawar: Should we instead have a `PrimaryDocumentInput` that extends `DocumentInput` and enforces document metadata fields.

server/src/main/java/org/opensearch/index/engine/dataformat/Writer.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Interface for writing documents to a data format.
+ *
+ * @param <P> the type of document input
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public interface Writer<P extends DocumentInput<?>> {
```
Bukhtawar:
```suggestion
public interface Writer<P extends DocumentInput<?>> extends Closeable {
```

server/src/main/java/org/opensearch/index/engine/dataformat/Writer.java
```diff
@@ -0,0 +1,53 @@
...
+     * Synchronizes the writer to ensure data is persisted.
+     *
+     * @throws IOException if an I/O error occurs
+     */
+    void sync() throws IOException;
+
+    /**
+     * Closes the writer and releases resources.
+     */
+    void close();
```
Bukhtawar: Should this also throw IOException in case the implementation decides the release resources

server/src/main/java/org/opensearch/index/engine/dataformat/WriteResult.java
```diff
@@ -0,0 +1,26 @@
...
+ * @param success whether the write was successful
+ * @param e the exception if the write failed, null otherwise
+ * @param version the document version
+ * @param term the primary term
+ * @param seqNo the sequence number
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public record WriteResult(boolean success, Exception e, long version, long term, long seqNo) {
```
Bukhtawar:
```suggestion
public sealed interface WriteResult {

    record Success(long version, long term, long seqNo) implements WriteResult {}

    record Failure(Exception cause) implements WriteResult {}
}
```

server/src/main/java/org/opensearch/index/engine/dataformat/IndexingExecutionEngine.java
```diff
@@ -0,0 +1,82 @@
...
+     * @throws IOException if an I/O error occurs during refresh
+     */
+    RefreshResult refresh(RefreshInput refreshInput) throws IOException;
+
+    /**
+     * Returns the data format handled by this engine.
+     *
+     * @return the data format
+     */
+    DataFormat getDataFormat();
```
Bukhtawar:
```suggestion
    T getDataFormat();
```

server/src/main/java/org/opensearch/index/engine/dataformat/Merger.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Merges files in secondary data formats using an existing row ID mapping.
+     *
+     * @param fileMetadataList list of writer file sets to merge
+     * @param rowIdMapping the row ID mapping from a prior merge
+     * @param newWriterGeneration the writer generation number
+     * @return merge result containing merged file metadata
+     */
+    MergeResult merge(List<WriterFileSet> fileMetadataList, RowIdMapping rowIdMapping, long newWriterGeneration);
```
Bukhtawar: Do we need to make primary engine/secondary engine interfaces more explicit?

server/src/main/java/org/opensearch/index/engine/dataformat/RefreshInput.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Input data for a refresh operation, containing existing segments and writer files.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class RefreshInput {
```
Bukhtawar: Should we make this Immutable with a Builder pattern?

server/src/main/java/org/opensearch/index/engine/exec/IndexFileReferenceManager.java
```diff
@@ -0,0 +1,37 @@
...
+    void addFileReferences(CatalogSnapshot snapshot);
+
+    /**
+     * Removes file references for all files in the given catalog snapshot.
+     * Decrements reference counts and may trigger deletion of unreferenced files.
+     *
+     * @param snapshot the catalog snapshot whose file references should be removed
+     */
+    void removeFileReferences(CatalogSnapshot snapshot);
+}
```
Bukhtawar: The interface maybe lacks atomicity for instance what happens if some files can ref count incremented while others have been closed. Also what happens if the add and remove happens concurrently across different files?

server/src/main/java/org/opensearch/index/engine/exec/WriterFileSet.java
```diff
@@ -0,0 +1,145 @@
...
+    public WriterFileSet(Path directory, long writerGeneration, long numRows) {
+        this.numRows = numRows;
+        this.files = new HashSet<>();
+        this.writerGeneration = writerGeneration;
+        this.directory = directory.toString();
+    }
+
+    public void add(String file) {
+        this.files.add(file);
+    }
```
Bukhtawar: Do we need this? Should this be a part of the Builder itself?

mgodwan: Removed

server/src/main/java/org/opensearch/index/engine/dataformat/RefreshInput.java
mgodwan: Ack. Changed it.

server/src/main/java/org/opensearch/index/engine/exec/Segment.java
mgodwan: Planning to handle this in the follow up PR. I've ensured field type capabilities are added. I will add DataFormatRegistry and related wiring from node to indexer in the following PR.

server/src/main/java/org/opensearch/index/engine/dataformat/WriteResult.java
mgodwan: I think this change may require instaceof check for classes consuming WriteResult (vended by addDoc methods).

server/src/main/java/org/opensearch/index/engine/dataformat/Merger.java
mgodwan: I've rolled this into a single method by creating a MergeInput now. We should not need to expose primary/secondary details in core directly with that

server/src/main/java/org/opensearch/index/engine/dataformat/DocumentInput.java
mgodwan: Gotten rid of the metadata fields for now. Re-looked at the way we would write indexing code, and we should be able to skip needing separate methods for these, as the underlying engine/dataformat won't need to understand whether fields are meta fields or not. Core can provide capabilities needed to be supported by engine for these fields, and we should be able to use dataformat based on priority order.

andrross:
@mgodwan I just rebased the latest version increment that is needed to get the bwc tests to pass

server/src/main/java/org/opensearch/index/engine/exec/WriterFileSet.java
```diff
@@ -0,0 +1,141 @@
...
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.file.Path;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Represents a set of files produced by a writer during indexing operations.
+ * Groups files by directory and writer generation, tracking metadata such as row count and total size.
+ * This class is serializable and can be transmitted across nodes.
```
andrross: StreamInput/StreamOutput is the convention for transmitting across nodes. Are you actually using Java serialization here?

server/src/main/java/org/opensearch/index/engine/exec/WriterFileSet.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "WriterFileSet{" + "directory=" + directory + ", writerGeneration=" + writerGeneration + ", files=" + files + '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        WriterFileSet other = (WriterFileSet) o;
```
andrross: This can throw a ClassCastException

server/src/main/java/org/opensearch/index/engine/exec/WriterFileSet.java
```diff
@@ -0,0 +1,141 @@
...
+        return this.directory.equals(other.directory)
+            && this.files.equals(other.files)
+            && this.getWriterGeneration() == other.getWriterGeneration();
+    }
+
+    @Override
+    public int hashCode() {
+        return this.directory.hashCode() + this.files.hashCode();
```
andrross: Missing writer generation

andrross: Assuming this class is meant to be immutable, you can make it a `record` and avoid needing to worry about this stuff

server/src/main/java/org/opensearch/index/engine/exec/WriterFileSet.java
```diff
@@ -0,0 +1,141 @@
...
+            if (directory == null) {
+                throw new IllegalStateException("directory must be set");
+            }
+
+            if (writerGeneration == null) {
+                throw new IllegalStateException("writerGeneration must be set");
+            }
+
+            WriterFileSet fileSet = new WriterFileSet(directory, writerGeneration, numRows);
+            fileSet.files.addAll(this.files);
```
andrross: I'd recommend passing an immutable `Set.copyOf` to WriterFileSet instead of adding to a mutable instance

server/src/main/java/org/opensearch/index/engine/dataformat/FileInfos.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        /**
+         * Builds the FileInfos instance.
+         *
+         * @return a new FileInfos instance
+         */
+        public FileInfos build() {
+            FileInfos fileInfos = new FileInfos();
+            writerFilesMap.forEach(fileInfos::putWriterFileSet);
```
andrross: Here also...`Map.copyOf()` is preferable to pass an immutable map. Can you make FileInfos a record?

server/src/main/java/org/opensearch/index/engine/exec/WriterFileSet.java
mgodwan: Removed the serializable interface. We intend to use StreamInput/StreamOutput only for these.

mgodwan: Will add the Writable interface when we actually implement logic to use it

server/src/main/java/org/opensearch/index/engine/exec/WriterFileSet.java
mgodwan: Ah! Fixing this.

PR #20685, opened 2026-02-20
Check for -Dorg.bouncycastle.fips.approved_only in testclusters to run with FIPS enforcement

cwperks:
### Description

Currently, testclusters will always assume that it should run with FIPS enforcement if `-Pcrypto.standard=FIPS-140-3` is supplied. The problem is, this is a build param needed for assembly, but at runtime the JVM expects the system prop `org.bouncycastle.fips.approved_only` to enforce that a cluster is running in FIPS mode.

Without this change, testclusters fail with

```

* What went wrong:
Execution failed for task ':integ-test:integTest'.
> Can not start node{:integ-test:integTest-0} in FIPS JVM, missing keystore password
```

and the workaround is to add `keystorePassword 'notarealpasswordphrase'` to the testclusters declaration.

### Check List

### DIFF
--- a/buildSrc/src/main/java/org/opensearch/gradle/info/FipsBuildParams.java
+++ b/buildSrc/src/main/java/org/opensearch/gradle/info/FipsBuildParams.java
@@ -37,7 +37,7 @@ public class FipsBuildParams {
     }
 
     public static boolean isInFipsApprovedOnlyMode() {
-return isInFipsMode() && "true".equals(System.getProperty("org.bouncycastle.fips.approved_only"));
+        return isInFipsMode() && "true".equals(System.getProperty("org.bouncycastle.fips.approved_only"));
     }
 
     public static String getFipsMode() {


cwperks:
@reta when you have a moment, this is a quick fix to fix an issue encountered with testclusters.

buildSrc/src/main/java/org/opensearch/gradle/testclusters/OpenSearchNode.java
```diff
@@ -549,7 +549,8 @@ public synchronized void start() {
             logToProcessStdout("installed plugins");
         }
 
-        if (FipsBuildParams.isInFipsMode() && keystorePassword.isEmpty()) {
+        if ((FipsBuildParams.isInFipsMode() && "true".equals(System.getProperty("org.bouncycastle.fips.approved_only")))
```
reta: May be we could introduce `isInFipsApprovedOnlyMode` to `FipsBuildParams`?
```suggestion
        if (FipsBuildParams.isInFipsApprovedOnlyMode()  && keystorePassword.isEmpty()) {
        }
```

cwperks: Should I take the opportunity and rename `isInFipsMode` to `FipsBuildParams.isPresent()` and have that return true if `-Pcrpyto.standard=FIPS-140-3`?

cwperks: I'm going to leave the other method as in, but IMO `isInFipsMode()` is not a great name for the param being present.

cwperks: Added `FipsBuildParams.isInFipsApprovedOnlyMode` which returns true when `-Dorg.bouncycastle.fips.approved_only=true` is present.

reta: Sorry, a bit late, `isInFipsApprovedOnlyMode` sounds good

buildSrc/src/main/java/org/opensearch/gradle/info/FipsBuildParams.java
```diff
@@ -36,6 +36,10 @@ public static boolean isInFipsMode() {
         return DEFAULT_FIPS_MODE.equals(fipsMode);
     }
 
+    public static boolean isInFipsApprovedOnlyMode() {
+        return "true".equals(System.getProperty("org.bouncycastle.fips.approved_only"));
```
reta: Super nit, I think `approved_only` only makes sense when FIPS mode is activated:

```suggestion
return isInFipsMode() && "true".equals(System.getProperty("org.bouncycastle.fips.approved_only"));
```

reta (approved):

PR #3119, opened 2022-04-30
Set term vector flags to false for ._index_prefix field (#1901).

vpehkone:
Signed-off-by: Vesa Pehkonen <vesa.pehkonen@intel.com>

### Description
This fix the issue #1901
 
### Issues Resolved
This fix the issue #1901
 
### Check List
- [x ] New functionality includes testing.
  - [x ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [x ] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #20691, opened 2026-02-20
added yaml rest integration test & document-indexing test

### LINKED ISSUE #1901: [BUG] `term_vector` does not work on `search_as_you_type` fields
**Describe the bug**
TL;DR: same issue reported [here](https://github.com/elastic/elasticsearch/issues/66407).

Per ElasticSearch's 7.10 [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/search-as-you-type.html#general-params) (which AFAICT should be compatible with OpenSearch 1.x), setting `term_vector` on a `search_as_you_type` field should only affect the shingle subfields, but not the index prefix subfield. However, when attempting to index documents with this mapping, an `illegal_argument_exception` is thrown.

**To Reproduce**
1. Create a new index with the following mapping:
```jsonc
{
  // ...
  "my_field":
    "type": "search_as_you_type",
    "term_vector": "with_positions_offsets"
  // ...
}
```
2. Attempt to index a document containing a `my_field` field
3. Observe the following error:
```json
{
  "reason": "cannot index term vector offsets when term vectors are not indexed (field=\"my_field._index_prefix\")"
}
```

**Expected behavior**
Documents should be indexed without issue. Only the `._2gram` and `._3gram` shingle subfields should have the `term_vector` set.

**Plugins**
N/A

**Screenshots**
N/A

**Host/Environment**
 - Version: 1.2.3 (opensearchproject/opensearch:latest)

ThyTran1402:
### Description
- The bugs for this issue are search_as_you_type creates three subfields:                                                                                                                                        - ._2gram, ._3gram (ShingleFieldMapper): should inherit term_vector                                                                                                                            - ._index_prefix (PrefixFieldMapper): must NOT get term_vector; the prefix index structure is incompatible with term vector storage
- When term_vector offsets/positions flags were set on the root field, the _index_prefix subfield's Lucene FieldType could end up with offsets/positions flags set but storeTermVectors = false, which is an illegal combination that Lucene rejects at indexing time with an exception.
- The fix for this bug is adding YAML REST integration test and document-indexing test (regression risk). And explicitly call prefixft.setStoreTermVectors(false) when constructing the prefix field's FieldType SearchAsYouTypeFieldMapper.Builder.build(). Although new FieldType() defaults to false, the explicit call guards against any future refactoring that might copy or propagate the root field's FieldType settings.
### Related Issues
Resolves #1901 

### Check List

### DIFF
--- a/modules/mapper-extras/src/yamlRestTest/resources/rest-api-spec/test/search-as-you-type/10_basic.yml
+++ b/modules/mapper-extras/src/yamlRestTest/resources/rest-api-spec/test/search-as-you-type/10_basic.yml
@@ -1240,3 +1240,27 @@ setup:
 
   - match: { hits.total: 1 }
   - match: { hits.hits.0._source.a_field: "quick brown fox jump lazy dog" }
+
+---
+"index document with term_vector on search_as_you_type field":
+
+  - do:
+      indices.create:
+        index: test-term-vector-index
+        body:
+          settings:
+            number_of_replicas: 0
+          mappings:
+            properties:
+              my_field:
+                type: search_as_you_type
+                term_vector: with_positions_offsets
+
+  - do:
+      index:
+        index: test-term-vector-index
+        id: 1
+        body:
+          my_field: "hello world"
+
+  - match: { result: created }
--- a/modules/mapper-extras/src/yamlRestTest/resources/rest-api-spec/test/search-as-you-type/10_basic.yml
+++ b/modules/mapper-extras/src/yamlRestTest/resources/rest-api-spec/test/search-as-you-type/10_basic.yml
@@ -1245,6 +1245,11 @@ setup:
 "index document with term_vector on search_as_you_type field":
 
   - do:
+      indices.delete:
+        index: test-term-vector-index
+        ignore_unavailable: true
+
+  - do:
       indices.create:
         index: test-term-vector-index
         body:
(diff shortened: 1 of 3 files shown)


modules/mapper-extras/src/yamlRestTest/resources/rest-api-spec/test/search-as-you-type/10_basic.yml
```diff
@@ -1240,3 +1240,32 @@ setup:
 
   - match: { hits.total: 1 }
   - match: { hits.hits.0._source.a_field: "quick brown fox jump lazy dog" }
+
```
gaobinlong, line 1243: Was the issue resolved by https://github.com/opensearch-project/OpenSearch/pull/3119? So this PR aims to add more tests?

ThyTran1402: Yup. The issue fixed by PR #3119. This PR only added test to verifies actual document indexing doesn't throw, and not just that FieldType flags are set correctly. They close a gap in the existing test coverage so we could regress the production fix without the existing tests catching the actual exception.

Let me know what your thoughts about it?

Thank you!

ThyTran1402:
Hi @gaobinlong 

I think the PR is ready for review.

Thank you!

ThyTran1402:
Hi @gaobinlong 

I just checked again and realized that the bug was not fully fixed it yet. And the bug is _index_prefix subfield does not call setTermVectorParams, but there may be a code path possibly through Lucene's segment merging or the test framework's document writer where term vector metadata from the sibling shingle fields bleeds into the prefix field's FieldInfo, causing a conflict. So, I fixed it recently, let me know if that works.

Can you review it again please?

Thank you!

Zhiyilang074811:
Nice test addition! Ready for review.

PR #20695, opened 2026-02-20
Fix flaky test in RankFeatureMetaFieldMapperTests by removing shared index creation

### LINKED ISSUE #20661: [AUTOCUT] Gradle Check Flaky Test Report for Gradle Test Run :modules:mapper-extras:test
## Flaky Test Report for `Gradle Test Run :modules:mapper-extras:test`

Noticed the `Gradle Test Run :modules:mapper-extras:test` has some flaky, failing tests that failed during **post-merge actions** and **timer trigerred run on main branch**.

### Details

| Git Reference | Merged Pull Request | Build Details | Test Name |
|---------------|----------------------|---------------|-----------|
| main | [null](https://github.com/opensearch-project/OpenSearch/pull/null) | [70693](https://build.ci.opensearch.org/job/gradle-check/70693/testReport/) | `Gradle Test Run :modules:mapper-extras:test.failed to execute tests` |

The other pull requests, besides those involved in post-merge actions, that contain failing tests with the `Gradle Test Run :modules:mapper-extras:test` class are:

- [19008](https://github.com/opensearch-project/OpenSearch/pull/19008)
- [19352](https://github.com/opensearch-project/OpenSearch/pull/19352)
- [19629](https://github.com/opensearch-project/OpenSearch/pull/19629)
- [19684](https://github.com/opensearch-project/OpenSearch/pull/19684)
- [19782](https://github.com/opensearch-project/OpenSearch/pull/19782)
- [19878](https://github.com/opensearch-project/OpenSearch/pull/19878)
- [19944](https://github.com/opensearch-project/OpenSearch/pull/19944)
- [19958](https://github.com/opensearch-project/OpenSearch/pull/19958)
- [20057](https://github.com/opensearch-project/OpenSearch/pull/20057)
- [20160](https://github.com/opensearch-project/OpenSearch/pull/20160)
- [20163](https://github.com/opensearch-project/OpenSearch/pull/20163)
- [20171](https://github.com/opensearch-project/OpenSearch/pull/20171)
- [20227](https://github.com/opensearch-project/OpenSearch/pull/20227)
- [20308](https://github.com/opensearch-project/OpenSearch/pull/20308)
- [20326](https://github.com/opensearch-project/OpenSearch/pull/20326)
- [20335](https://github.com/opensearch-project/OpenSearch/pull/20335)
- [20345](https://github.com/opensearch-project/OpenSearch/pull/20345)
- [20372](https://github.com/opensearch-project/OpenSearch/pull/20372)
- [20383](https://github.com/opensearch-project/OpenSearch/pull/20383)
- [20480](https://github.com/opensearch-project/OpenSearch/pull/20480)

For more details on the failed tests refer to [OpenSearch Gradle Check Metrics](https://metrics.opensearch.org/_dashboards/app/dashboards#/view/e5e64d40-ed31-11ee-be99-69d1dbc75083) dashboard.

getsaurabh02:
Move index creation from @Before setup() into individual test methods to prevent resource conflicts when tests run in parallel. This fixes the 'failed to execute tests' error in CI environments.

Fixes #20661 

Ran the test 100 times to ensure no flakiness locally
```
 for i in {1..100}; do echo "=== Run $i/100 ==="; ./gradlew :modules:mapper-extras:test --tests "RankFeatureMetaFieldMapperTests" --rerun-tasks 2>&1 | grep -E "BUILD|FAILED"; done | tee test_results.txt
```

This follows a pattern I found at other places:

- [IndexServiceTests.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/test/java/org/opensearch/index/IndexServiceTests.java#L106)
- [RetentionLeaseActionsTests.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/test/java/org/opensearch/index/seqno/RetentionLeaseActionsTests.java#L69)
- [IndexFieldDataServiceTests.java](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/test/java/org/opensearch/index/fielddata/IndexFieldDataServiceTests.java#L105) 
- [ParentJoinFieldMapperTests.java](https://github.com/opensearch-project/OpenSearch/blob/main/modules/parent-join/src/test/java/org/opensearch/join/mapper/ParentJoinFieldMapperTests.java#L74)


cwperks:
>  Run './gradlew spotlessApply' to fix all violations.

@getsaurabh02 please run spotless apply

modules/mapper-extras/src/test/java/org/opensearch/index/mapper/RankFeatureMetaFieldMapperTests.java
```diff
@@ -44,13 +44,9 @@
 
 public class RankFeatureMetaFieldMapperTests extends OpenSearchSingleNodeTestCase {
 
-    IndexService indexService;
-    DocumentMapperParser parser;
-
     @Before
```
cwperks: Can we get rid of the `setup()` block entirely if its empty?

Does it make sense for all tests in this suite to create an index with the name `test`. Should we consider randomizing the index name creation at all?

getsaurabh02: Thanks @cwperks. makes sense to remove ``setup()``. Had kept it with comment for reference. 

Also randomized index names helps avoid potential conflicts. Will update

cwperks:
```
> Task :modules:lang-painless:compileTestJava
Forbidden method invocation: java.lang.String#toLowerCase() [Uses default locale]
  in org.opensearch.index.mapper.RankFeatureMetaFieldMapperTests (RankFeatureMetaFieldMapperTests.java:52)
Forbidden method invocation: java.lang.String#toLowerCase() [Uses default locale]
  in org.opensearch.index.mapper.RankFeatureMetaFieldMapperTests (RankFeatureMetaFieldMapperTests.java:78)
```

Need to specify locale, i.e. `.toLowerCase(Locale.ROOT)`

andrross:
@getsaurabh02 JUnit creates a new instance of the class for each test method invocation, and runs `@Before` annotated setup methods before each test method. Any parallelism that happens will be done by forking off an entirely separate JVM instance to run other test classes in parallel. There should never be multiple test methods running in parallel inside any given VM. My understanding is that the code implemented here should not be functionally different from what existed previously.

getsaurabh02:
> @getsaurabh02 JUnit creates a new instance of the class for each test method invocation, and runs `@Before` annotated setup methods before each test method. Any parallelism that happens will be done by forking off an entirely separate JVM instance to run other test classes in parallel. There should never be multiple test methods running in parallel inside any given VM. My understanding is that the code implemented here should not be functionally different from what existed previously.

@andrross makes sense. Is is possible actual fix that likely resolved the flakiness was adding randomized index names ``(randomAlphaOfLength(10).toLowerCase(Locale.ROOT))``? It does prevent conflicts with indices from previous test runs or other tests.

andrross:
> ...randomized index names ... It does prevent conflicts with indices from previous test runs or other tests.

@getsaurabh02 It should not matter. This test does leave state in between test runs because it is an `OpenSearchSingleNodeTestCase` which populates a [static `NODE` field](https://github.com/opensearch-project/OpenSearch/blob/f3632f1a09993521697ec905a40d66091d8664f3/test/framework/src/main/java/org/opensearch/test/OpenSearchSingleNodeTestCase.java#L105) that will survive across test instances and test runs. However, the tearDown() logic deletes all indices and fails if any continue to exist. I think we would see a clear failure if there was a conflict.

andrross:
@getsaurabh02  The failure I see in the logs is:

```
* What went wrong:
Execution failed for task ':modules:mapper-extras:test'.
> Test process encountered an unexpected problem.
   > Process 'Gradle Test Executor 325' finished with non-zero exit value 1
```

and also

```
org.gradle.internal.remote.internal.ConnectException: Could not connect to server [1ca6ae1b-de7e-4d53-997d-f9cb4febc59a port:39599, addresses:[/127.0.0.1]]. Tried addresses: [/127.0.0.1].
	at org.gradle.internal.remote.internal.inet.TcpOutgoingConnector.connect(TcpOutgoingConnector.java:73)
	at org.gradle.internal.remote.internal.hub.MessageHubBackedClient.getConnection(MessageHubBackedClient.java:36)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:112)
	at org.gradle.process.internal.worker.child.SystemApplicationClassLoaderWorker.call(SystemApplicationClassLoaderWorker.java:72)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.run(GradleWorkerMain.java:69)
	at worker.org.gradle.process.internal.worker.GradleWorkerMain.main(GradleWorkerMain.java:74)
Caused by: java.net.ConnectException: Connection refused
	at java.****/sun.nio.ch.Net.pollConnect(Native Method)
	at java.****/sun.nio.ch.Net.pollConnectNow(Net.java:639)
	at java.****/sun.nio.ch.SocketChannelImpl.finishTimedConnect(SocketChannelImpl.java:1267)
	at java.****/sun.nio.ch.SocketChannelImpl.blockingConnect(SocketChannelImpl.java:1309)
	at java.****/sun.nio.ch.SocketAdaptor.connect(SocketAdaptor.java:106)
	at org.gradle.internal.remote.internal.inet.TcpOutgoingConnector.tryConnect(TcpOutgoingConnector.java:86)
	at org.gradle.internal.remote.internal.inet.TcpOutgoingConnector.connect(TcpOutgoingConnector.java:60)
	... 5 more
```

I think the Gradle daemon lost connection to the test worker because the test worker process died. The cause would likely be something like the JVM hard-crashed with some an error or the operating system OOM killer killed the process. I don't think this is a problem with the test logic, unless the test is filling up the heap leading to an OOM and a hard crash of the JVM.

Were you able to replicate this failure? Did your for loop test fail before this change?

Issue #19902, opened 2025-11-05
[Meta] OpenSearch Composable Query Engine


mch2:
### Please describe the end goal of this project

This is a meta to start tracking changes related to [OpenSearch Composable Query Engine Architecture](https://github.com/opensearch-project/OpenSearch/issues/19653), where our goal is to introduce plugin points for query execution outside of lucene based implementations. I think we can organize this by splitting into two sub meta issues one for indexing side changes outlined in [OpenSearch Execution Engine ](https://github.com/opensearch-project/OpenSearch/issues/18416)https://github.com/opensearch-project/OpenSearch/issues/18416 and one for [Query planning and execution](https://github.com/opensearch-project/OpenSearch/issues/18847)

There are folks working on a POC on the feature/datafusion branch - https://github.com/opensearch-project/OpenSearch/pulls?q=is%3Apr+is%3Aopen+base%3Afeature%2Fdatafusion with a growing diff of changes.  I'd like to cut this meta now and split out tasks so we can slowly start pulling changes back to main.

We have been working on an initial goal to introduce a flow where indices are backed by a single parquet file and queryable via PPL.  We will cut sub issues that go into detail on the extension points and components proposed, but the high level query/fetch flow would look like the following:

<img width="2571" height="1102" alt="Image" src="https://github.com/user-attachments/assets/738a8e6d-5e53-4d1d-b323-479ac57c6ad5" />

Update - the above design is inside of /feature/datafsion. - below is what I'm thinking for the next version to bring into sandbox plugins.

<img width="980" height="1042" alt="Image" src="https://github.com/user-attachments/assets/b4cfb7cf-57cd-4b66-aefc-16349da62d05" />

### Issues

Search:

- [ ] Introduce new Search engine plugin point for alternate query execution.
- [ ] DataFusion engine plugin implementation with parquet reader support

more coming....

Indexing:
- [ ] Introduce Composite Engine abstraction

more coming...

### Related component

Search

Issue #18416, opened 2025-06-02
[RFC] OpenSearch Execution Engine


mgodwan:
### Is your feature request related to a problem? Please describe

# RFC: OpenSearch Execution Engine

This RFC proposes extending OpenSearch to support new storage formats along with corresponding indexing and/or query engines in addition to Lucene. This enhancement will provide users with flexibility to choose data format/engines based on their specific use cases, performance requirements and data characteristics. This RFC currently focuses on a high level overview of the proposal, detailed approaches will follow.

## Motivation

Users leverage OpenSearch for a wide variety of use-cases. These use cases spread across multiple data sources (catalog, transactions, etc.) and data (logs, metrics, traces, etc.) from different components in their pipeline. OpenSearch users face various pain points categorized along the following themes:

#### Storage Cost

* Different data types and access patterns require different storage optimizations
* Need for more cost-effective storage solutions for specific workloads

#### Data Tiering

* Limited flexibility in how data is stored across different tiers
* Need for better control over data placement and storage formats

#### New Use Cases

* First-class support for features like cross-index joins
* Aggregation push down to the data storage layer

### Current Data Storage Limitations

Today, OpenSearch uses Lucene’s codec interface to store the data. This allows OpenSearch to leverage different formats that are exposed through Lucene and extend them as necessary (e.g. doc values, points, postings, vectors, etc.) to build required implementations suited to the use case. But this has certain drawbacks:

1. This limits addition of any new fields / field types which doesn’t fit in with existing types supported by Lucene codecs (e.g. multi-field data structures).
2. This also requires constructs like roll-ups, aggregation, etc to be built on application layer, rather than being built on the existing data directly.

There have been similar discussions in the OpenSearch community which this RFC aims to provide a high level solution for:

1. https://github.com/opensearch-project/OpenSearch/issues/13668
2. https://github.com/opensearch-project/OpenSearch/issues/17501

## Goals

* Enable pluggable storage formats and corresponding indexing, and query engines.
* Maintain backward compatibility with data stored using Lucene engine.
* Consistency of API interface.
* Usage of mixed engines within the same index (provided the data format is compatible with the engines).
* Data migration from existing sources

### Data Storage Format

In order to extend OpenSearch to support more use-cases and increase the efficiency (better compression) of data storage and retrieval, OpenSearch can make the data storage format configurable. This will allow OpenSearch to expose multiple other index types with different field combinations, encoding, etc. 

### Projections

[...]

PR #20697, opened 2026-02-20
initial commit of analytics engine plugin to sandbox

mch2:
### Description
This PR introduces a set of experimental plugins that we are building out related to [this](https://github.com/opensearch-project/OpenSearch/issues/19902) issue, that starts moving feature/engine-datafusion into main.

This introduces:
analytics-backend-datafusion - responsible for executing queries against datafusion back-end.  No-Op right now.
analytics-engine - main "hub" for wiring up front-ends, back-ends, and a plan executor (component that is responsible for executing a query end-to-end).

For now its largely no-op, providing wiring for back-ends and the engine using ExtensiblePlugin.  There is a test front-end inside of the engine that uses the UnifiedQueryAPI from sql plugin that will invoke our executor.

back-ends will implement `AnalyticsBackEndPlugin` and provide to the engine via spi. that for now defines its capabilities in terms of supported calcite operators/functions, and a path to execution via an `EngineBridge`.  These common types are kept in a sandbox lib.

### Related Issues
Related issues:
https://github.com/opensearch-project/OpenSearch/issues/19902
https://github.com/opensearch-project/OpenSearch/issues/18416

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/README.md
+++ b/sandbox/plugins/analytics-engine/README.md
@@ -15,5 +15,5 @@ This is the stable plugin name that all extending plugins declare in `extendedPl
 
 ## Key classes
 
-- **`BaseAnalyticsPlugin`** — The `ExtensiblePlugin` hub. Handles discovery, wiring, and Guice binding.
+- **`AnalyticsPlugin`** — The `ExtensiblePlugin` hub. Handles discovery, wiring, and Guice binding.
 - **`OpenSearchSchemaBuilder`** — Converts `ClusterState` index mappings into Calcite tables with typed columns.


Bukhtawar (commented):
Thanks for the initial draft Marc, looks like a good start
The one high-level concern I have is that one single plugin is responsible for planning, orchestration, substrait conversion etc would become unwieldy overtime, espl when we bring in plan optimisers and orchestrators(stage-based). 
Would it make sense to tease out planner/optimisers into a plugin and the maybe orchestrators as yet another while letting server have the stitching logic across these plugins. Also I am not sure if we need the `ExecutionEngine` to be invoked from the extensible plugin(avoiding the SPI complexity for native execution engine)

mch2:
> Would it make sense to tease out planner/optimisers into a plugin and the maybe orchestrators as yet another while letting server have the stitching logic across these plugins.

Still trying this out, my thinking with this is to keep it self contained (at least while we're in sandbox) & opt in so we aren't bloating server for users who don't use this plugin.  Perhaps we can move planner/optimizer to /libs but it will also carry the calcite dependency with it.

> Also I am not sure if we need the ExecutionEngine to be invoked from the extensible plugin(avoiding the SPI complexity for native execution engine)

The query engine is going to have fragments we'll need to send to the native engine for both data nodes and coordinators.  So we can wire it via SPI, or through pluggable interface in core but I think the driver should remain in the plugin itself.

mch2:
Ok gave this a shot and I think it may work out.  Ultimately we have a query-planner plugin, the engine plugin (implementing ExecutionEnginePlugin), and a module/ for the rest (what used to be extensible-engine) - thinking about pulling what i can out to a lib there as well, but this should remove the direct call from the engine into the native executor.

Bukhtawar:
> Ok gave this a shot and I think it may work out. Ultimately we have a query-planner plugin, the engine plugin (implementing ExecutionEnginePlugin), and a module/ for the rest (what used to be extensible-engine) - thinking about pulling what i can out to a lib there as well, but this should remove the direct call from the engine into the native executor.

Thanks Marc for the change, looks awesome. The plugin segregations across planner and executor look pretty neat(thanks for moving the executor to modules to simplify plugin installation for various FE lang). Overall love the simplification while balancing extensibility.

mch2:
needs some clean up still - but updated the interfaces here @Bukhtawar.

```
server/src/main/java/org/opensearch/plugins/          ← Server-layer interfaces
├── QueryPlanExecutor.java                             │ execute(Object plan, Object context) → Iterable<Object[]>
├── QueryEnginePlugin.java                             │ createComponents(..., backEnds), getExecutor() → QueryPlanExecutor
├── AnalyticsFrontEndPlugin.java                       │ createComponents(..., QueryPlanExecutor)
└── AnalyticsBackEndPlugin.java                        │ marker interface

^ if we used SPI we could clean this up a bit - remove Object types, have front-end and back-end wire directly in.
Or... we could wire up by passing pluginsservice down to AnalyticsEnginePlugin and have it configure the fe/be plugins.

sandbox/plugins/analytics-engine/                      ← Engine plugin (Gradle: opensearchplugin)
├── build.gradle                                       │ calcite-core, calcite-linq4j, avatica-core, guava, + runtime deps
└── src/main/java/org/opensearch/fe/ppl/
    ├── AnalyticsEnginePlugin.java                     │ implements QueryEnginePlugin, ActionPlugin
    └── PlanExecutor.java                              │ implements QueryPlanExecutor (stub: logs fragment, returns empty)

sandbox/plugins/analytics-frontend-ppl/                ← PPL front-end plugin (Gradle: opensearchplugin)
```

and in future:
```
sandbox/plugins/analytics-frontend-dsl
sandbox/plugins/engine-datafusion
sandbox/plugins/engine-lucene
```

sandbox/plugins/analytics-frontend-ppl/src/main/java/org/opensearch/fe/ppl/planner/DefaultEngineExecutor.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.calcite.linq4j.Enumerable;
+import org.apache.calcite.linq4j.Linq4j;
+import org.apache.calcite.rel.RelNode;
+import org.opensearch.plugins.QueryPlanExecutor;
+
+/**
+ * Default {@link EngineExecutor} implementation that wraps a {@link QueryPlanExecutor} provided
+ * by server.
+ */
+public class DefaultEngineExecutor implements EngineExecutor {
```
Bukhtawar: Assuming this isn't PPL specific? Is this the part to be moved to libs to be shared across front-ends?

mch2: it is, this is simply a wrapper for the provided QueryPlanExecutor from the analytics-plugin to convert results to enumerable.  Will remove the interface.

This is used in PushDownPlanner, that when we execute a scan will invoke our QueryPlanExecutor with the plan fragment that our engine can handle.  PushDownPlanner is only used in the sql/ppl plugin to use unified query api.

andrross:
@mch2 Annoying high level comment, but how about some README.md files in each new component's root directory to give some context about what each thing does and how it fits in?

mch2:
@Bukhtawar For now i've separated this into front-end/back-end/executor, with a shared lib and plugin "hub" to wire in shared components.  As for query planning, everything planning related in the ppl plugin is there to allow the UQP to build plans for our executor.  The Executor itself will do its own planning rbo/cbo,scheduling that will be reused across front-ends

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsBackEndPlugin.java
```diff
@@ -0,0 +1,32 @@
...
+ * SPI extension point for back-end execution engines (DataFusion, Lucene, etc.).
+ *
+ * <p>Implementations are discovered by the analytics engine hub via
+ * {@code ExtensiblePlugin.loadExtensions} and passed to the
+ * {@link QueryPlanExecutorPlugin} during initialization.
+ *
+ * @opensearch.internal
+ */
+public interface AnalyticsBackEndPlugin {
+    /** Unique engine name (e.g., "lucene", "datafusion"). */
```
Bukhtawar, line 19: Assuming this is purely execution engines and unrelated to data formats like Parquet/Vortex etc would still be datafusion

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/BaseAnalyticsPlugin.java
```diff
@@ -0,0 +1,95 @@
...
+        SchemaProvider schemaProvider = clusterState ->
+            OpenSearchSchemaBuilder.buildSchema((ClusterState) clusterState);
+        schemaProviderRef.set(schemaProvider);
+
+        if (executorPlugins.isEmpty() == false) {
+            QueryPlanExecutorPlugin executorPlugin = executorPlugins.get(0);
+            QueryPlanExecutor executor = executorPlugin.createExecutor(backEnds);
+            executorRef.set(executor);
+        }
+    }
```
Bukhtawar: The only plugin that cares about the native execution engine is likely the query executor plugin? Do we really need another plugin to wire it up?
I would have been fine with this plugin, if front-end plugins needed the backend info, in which case the hub makes sense. But if the backends are relevant only to the query executor why not have the executor directly wire it up or have backend passed from the /server

Maybe I am not able to appreciate the true value of `analytics-engine` plugin . Unlike `jobs-scheduler` plugin I don't see this plugin doing any undifferentiated heavy lifting(acting as a hub isn't one of those maybe?). I am also not sure if we need to have a Calcite dependency too in here(good to avoid, if tomorrow we want to move away from Calcite). I assume the place where we need them would still be the front end and the executor plugins which should be fine for now.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineBridge.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Interface for engine execution.
+ * Actual implementations live in engine-specific plugins (e.g., engine-datafusion).
+ */
+public interface EngineBridge<T> {
+
+    T convertFragment(RelNode fragment);
+
+    Iterator<VectorSchemaRoot> execute(T fragment);
```
bharath-techie: Should we have a hard dependency on arrow for responses ? Can we keep it generic

sandbox/plugins/analytics-frontend-ppl/src/main/java/org/opensearch/fe/action/RestUnifiedPPLAction.java
```diff
@@ -0,0 +1,152 @@
...
+ * {@link TransportUnifiedPPLAction}. Parses JSON body with a {@code query}
+ * field, validates the query text, and converts the response to JSON with
+ * {@code columns}, {@code datarows}, {@code total}, {@code size}, and
+ * {@code status} fields.
+ *
+ * <p>Route: {@code POST /_plugins/_query_engine/_unified/ppl}
+ */
+public class RestUnifiedPPLAction extends BaseRestHandler {
+
+    public static final String ROUTE_PATH = "/_plugins/_analytics_engine/ppl";
```
Bukhtawar: We might need a deprecated route for maintaining backward compatibility. I am assuming we want to support both analytics and the current ppl implementations into this new plugin.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsFrontEndPlugin.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * <p>Marker interface for discovery by the analytics engine hub.
+ * Front-ends receive {@code QueryPlanExecutor} and {@code SchemaProvider}
+ * via Guice injection.
+ *
+ * @opensearch.internal
+ */
+public interface AnalyticsFrontEndPlugin {
+    // Marker interface for SPI discovery
+}
```
Bukhtawar: Why do front end plugins need SPI discovery given they don't depend on other plugins

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineCapabilities.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Declares what the custom engine supports using Calcite's own types.
+ */
+public class EngineCapabilities {
```
Bukhtawar: Discussed with @bharath-techie, we need to make this independent of query as this will dictate ingestion formats too. This needs to be controlled through a registry that integrates closely with mapper

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsBackEndPlugin.java
mch2: Correct execution only.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/BaseAnalyticsPlugin.java
mch2: Its main purpose right now is allowing the executor to be pluggable. Without this the front and back ends would extend the executor impl directly.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsFrontEndPlugin.java
mch2: They don't. Will yank this.

sandbox/plugins/analytics-frontend-ppl/src/main/java/org/opensearch/fe/action/RestUnifiedPPLAction.java
mch2: Are u suggesting this plugin also supports existing sql plugin push down to dsl? I'm not sure we want to go that route until we've built up support here with our engine.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineBridge.java
mch2: Do we have that use case for non arrow? I don't want to over generalize here.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineCapabilities.java
mch2: There's two separate things here. What you're describing building the schema which absolutely should connect to mappings/state. I tried putting that behind the schema provider. The other which is what this class is, is communicating what our back end can support in terms of functions/operators. That we then use to build an api definition at the executor. That's what this is. I was hesitant to include this to start. We can simply maintain a static list for now.

sandbox/plugins/analytics-frontend-ppl/src/main/java/org/opensearch/fe/action/RestUnifiedPPLAction.java
Bukhtawar: With optimised index mode we could reuse the same endpoint(backward compatible) without having to support the existing dsl conversion.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineCapabilities.java
```diff
@@ -0,0 +1,90 @@
...
+                Boolean childResult = operand.accept(this);
+                if (childResult != null && !childResult) {
+                    return false;
+                }
+            }
+            return true;
+        }
+    }
+
+    public boolean supportsAllAggFunctions(List<AggregateCall> aggCalls) {
```
vinaykpud: For the DSL frontend, I was thinking to validate the given SqlOperator/AggFunction first before converting them to RexNode/AggregateCall so that we can fail fast without converting fully.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/BaseAnalyticsPlugin.java
andrross: So what exactly does the alternative look like without a pluggable QueryPlanExecutor? Do the front end plugins take a dependency on the thing that implements the QueryPlanExecutor and instantiate it directly? If so, I guess the question is how many front end plugins and how many QueryPlanExecutors do we expect to have?

Bukhtawar: The alternative I was thinking we have a default implementation of analytics-engine but that resides in /server similar to the ClusterPlugin and have extension points as plugins. So pretty much the hub logic can sit in core with plugin service doing all this wiring?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineCapabilities.java
mch2: we can add a method here for your use case

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/BaseAnalyticsPlugin.java
mch2: i removed this "hub" concept for now to get going.  I think once we are ready to promote these out of sandbox that makes sense wdyt?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineBridge.java
mch2: updated this, this will be a ptr anyway to a stream.

sandbox/plugins/analytics-frontend-ppl/src/main/java/org/opensearch/fe/action/RestUnifiedPPLAction.java
mch2: yeah agree - was discussing this with @penghuo, and we are thinking to simply make our existing sql plugin the front-end at this time, that invokes our executor, and avoid all the re-routing.

sandbox/libs/build.gradle
```diff
@@ -20,13 +20,14 @@ subprojects {
   project.afterEvaluate {
     configurations.all { Configuration conf ->
       dependencies.matching { it instanceof ProjectDependency }.all { ProjectDependency dep ->
-        Project depProject = dep.dependencyProject
+        Project depProject = project.project(dep.path)
         if (depProject != null
-          && false == depProject.path.equals(':libs:opensearch-core')
+          && (false == depProject.path.equals(':libs:opensearch-core') &&
+              false == depProject.path.equals(':libs:opensearch-common'))
```
andrross, line 26: I think this change is fine, because core depends on common. I'm guessing we never previously had a case where a lib needed to depend on common but not core. Do you have a lib that wants to depend on common but not core?

sandbox/libs/analytics-framework/build.gradle
```diff
@@ -0,0 +1,419 @@
...
+ * compatible open source license.
+ */
+
+/*
+ * Shared Calcite/Arrow types for the analytics engine plugins.
+ * Contains EngineBridge, EngineCapabilities, QueryPlanExecutor, SchemaProvider.
+ * Plugins depend on this; the /modules SPI layer does NOT.
+ */
+
+def calciteVersion = '1.41.0'
```
andrross, line 15: This is only used in one place. Inline it unless you expect that to change

sandbox/libs/analytics-framework/build.gradle
```diff
@@ -0,0 +1,419 @@
...
+    'com.fasterxml.jackson.core.type.TypeReference',
+    'com.fasterxml.jackson.core.util.DefaultIndenter',
+    'com.fasterxml.jackson.core.util.DefaultPrettyPrinter',
+    'com.fasterxml.jackson.core.util.Separators',
+    'com.fasterxml.jackson.core.util.Separators$Spacing',
+    'com.fasterxml.jackson.databind.DeserializationFeature',
+    'com.fasterxml.jackson.databind.ObjectMapper',
+    'com.fasterxml.jackson.databind.ObjectWriter',
+
+    // --- Guava (excluded from compile classpath by OpenSearch policy) ---
```
andrross: Is this going to work? I'm assuming guava is not an optional dependency

sandbox/plugins/analytics-engine/src/internalClusterTest/java/org/opensearch/fe/planner/unified/ClickBenchUnifiedPipelineIT.java
```diff
@@ -0,0 +1,395 @@
...
+    private static void addDateField(XContentBuilder builder, String name) throws Exception {
+        builder.startObject(name)
+            .field("type", "date")
+            .field("format", "yyyy-MM-dd HH:mm:ss||strict_date_optional_time||epoch_millis")
+            .endObject();
+    }
+
+    // --- ClickBench query tests ---
+
+    public void testClickBenchQ1() throws Exception {
```
andrross: Nitpick, but `@ParametersFactory` would probably be a better fit, something like:

https://github.com/opensearch-project/OpenSearch/blob/64684649d3095041a5200a3e441e8d0ba6c03e6b/server/src/test/java/org/opensearch/search/approximate/ApproximatePointRangeQueryTests.java#L65-L76

sandbox/plugins/analytics-engine/src/test/java/fe/ppl/action/TestPPLTransportActionTests.java
```diff
@@ -0,0 +1,172 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package fe.ppl.action;
```
andrross: Package name doesn't follow the org.opensearch convention

sandbox/libs/build.gradle
mch2: The framework lib doesn't actually depend on either common/core right now. This build.gradle is stale as written and the lib build fails with `Could not get unknown property 'dependencyProject'`.  This is only to bring sandbox/libs/build.gradle in sync with regular libs/build.gradle.

sandbox/libs/analytics-framework/build.gradle
mch2: i expect it will change but inlining it until we need to.

sandbox/libs/analytics-framework/build.gradle
mch2: Yeah, you're right - fixing this.  We'll also need to explicitly bring in transitive calcite deps - fixing it.

sandbox/libs/build.gradle
andrross: Ah, I missed that libs/build.grade had this change

sandbox/plugins/analytics-engine/README.md
```diff
@@ -0,0 +1,19 @@
...
+3. Calls `QueryPlanExecutorPlugin.createExecutor(backEnds)` to build the `QueryPlanExecutor`.
+4. Binds `QueryPlanExecutor` and `SchemaProvider` into Guice via lazy providers, so front-end transport actions can `@Inject` them.
+
+## How it fits in
+
+This is the stable plugin name that all extending plugins declare in `extendedPlugins`. It owns no query language logic and no execution logic — it just discovers, wires, and exposes the components that other plugins provide.
+
+## Key classes
+
+- **`BaseAnalyticsPlugin`** — The `ExtensiblePlugin` hub. Handles discovery, wiring, and Guice binding.
```
andrross:
```suggestion
- **`AnalyticsPlugin`** — The `ExtensiblePlugin` hub. Handles discovery, wiring, and Guice binding.
```

PR #20702, opened 2026-02-21
Fix/collision of indexpatterns

### LINKED ISSUE #837: [BUG] Validation-Error: collision of indexpatterns with differences after a wildcard
(initially posted as https://github.com/opendistro-for-elasticsearch/index-management/issues/446)

**Describe the bug**
We have indexes named `app-test-*-some-*` and `app-test-*-some_other-*` 
Therefore we create two indexpatterns which should be clearly distinguishable.

Unfortunately this does not work as the patterns are supposed to not be unique

But the patterns are clearly distinguisable in my eyes because of the character after 'some' (underscore vs. dash)

It seems that this is not only when a difference is between underscore and dash but that only the part before the first wildcard is checked:

**Other plugins installed**
Opendistro 1.13.2 without only the standard plugins

**To Reproduce**
1. Go to Dev-Tools and send the request 1:
```
PUT /_index_template/app-test-some-
{
  "index_patterns": ["app-test-*-some-*" ]
}
```
2. Go to Dev-Tools and send the request 2:
```
PUT /_index_template/app-test-some_other-
{
  "index_patterns": ["app-test-*-some_other-*" ]
}
```
3. This gives error:
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "index template [app-test-some_other-] has index patterns [app-test-*-some_other-*] matching patterns from existing templates [app-test-some-] with patterns (app-test-some- => [app-test-*-some-*]) that have the same priority [0], multiple index templates may not match during index creation, please use a different priority"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "index template [app-test-some_other-] has index patterns [app-test-*-some_other-*] matching patterns from existing templates [app-test-some-] with patterns (app-test-some- => [app-test-*-some-*]) that have the same priority [0], multiple index templates may not match during index creation, please use a different priority"
  },
  "status" : 400
}
```

also:

'''
PUT /_index_template/app-test-some_other-
{
"index_patterns": ["app-test--completedifferent-" ]
}
'''
gives this error

**Expected behavior**
The patterns should be both accepted as a dash and an underscore are two different characters. Parts after a wildcard should be respected

**Screenshots**

**Desktop (please complete the following information):**
Browser, curl and Python-Http-call (problem not client related)

**Additional context**
This worked using the old api (_template) in previous version (1.12 ?)

ThyTran1402:
### Description
- Root cause: the previous implementation used a full Lucene automaton intersection (Operations.intersection) to decide whether two pattern sets conflict. This is correct since the string app-test--some_other--some- does technically satisfy both patterns simultaneously, but such a string would never appear as a real index name. Because * absorbs any characters (including - and _), wildcards in one pattern can "swap roles" to cover literal segments of the other, producing false conflicts for any two multi-wildcard patterns that share a common literal prefix before the first *

- Fix: Replace the automaton intersection with a minimum-string heuristic. For each pattern, every * is substituted with "" to produce its minimum matching string. If that string matches any pattern in the other set (using the existing Regex.simpleMatch), the two sets are considered conflicting. This restricts conflict detection to cases where naturally named index names would match both templates.
- The heuristic still catches genuine conflicts like baz vs baz*, or logs-* vs logs-prod-* while correctly accepting patterns that are distinguishable by literal segments placed after wildcards.

### Related Issues
Resolves  #837 

### Check List


msfroh:
@ThyTran1402, can you please run `./gradlew :server:spotlessApply`, add the changed files, commit them, and push them to your PR branch?

We're getting the following build failure:

```
* What went wrong:
      src/test/java/org/opensearch/cluster/metadata/MetadataIndexTemplateServiceTests.java
          @@ -1318,9 +1318,7 @@
           
           ····public·void·testPatternsActuallyOverlap()·{
           ········//·Patterns·that·share·a·genuine·conflict:·baz·matches·baz*
          -········assertTrue(
          ...
          -············0L,
          -············null,
      ... (7 more lines that didn't fit)

File: rest-api-spec/src/main/resources/rest-api-spec/test/indices.put_index_template/10_basic.yml [hunk 1]
```diff
@@ -102,6 +102,49 @@
...
               number_of_replicas: 0
 
+---
+"Put distinguishable multi-wildcard templates at the same priority":
+  - skip:
+      features: allowed_warnings
+
+  - do:
+      allowed_warnings:
+        - "index template [app-test-some] has index patterns [app-test-*-some-*] matching patterns from existing older templates [global] with patterns (global => [*]); this template [app-test-some] will take precedence during new index creation"
```
msfroh, line 114: I'm a little confused about this one, since `app-test-*-some-*` doesn't match `app-test-some`. 

The shortest possible match for `app-test-*-some-*` would be be `app-test--some-`.

server/src/main/java/org/opensearch/cluster/metadata/MetadataIndexTemplateService.java
```diff
@@ -873,6 +867,36 @@ static Map<String, List<String>> findConflictingV2Templates(
...
+    /**
+     * Checks whether two sets of index patterns practically overlap, using a minimum-string heuristic.
+     * <p>
+     * For each pattern in either set, the wildcard character {@code *} is replaced with the empty string
+     * to produce its "minimum matching string". If that minimum string matches any pattern in the other
+     * set, the two pattern sets are considered to overlap.
+     * <p>
+     * This approach is intentionally more lenient than a full automaton intersection. Full automaton
+     * intersection correctly identifies theoretical overlaps that arise only from creative wildcard usage
+     * (e.g. {@code app-test-*-some-*} and {@code app-test-*-some_other-*} share the synthetic string
```
msfroh, line 879: These wouldn't actually overlap. If the second string was `app-test-*-some-other-*`, then there would be overlap. 

Is that the kind of case that you're looking to address?

rest-api-spec/src/main/resources/rest-api-spec/test/indices.put_index_template/10_basic.yml
```diff
@@ -102,6 +102,49 @@
...
+
+  - do:
+      indices.get_index_template:
+        name: app-test-some
+  - match: {index_templates.0.index_template.index_patterns: ["app-test-*-some-*"]}
+
+  - do:
+      indices.get_index_template:
+        name: app-test-some-other
+  - match: {index_templates.0.index_template.index_patterns: ["app-test-*-some_other-*"]}
```
msfroh, line 141: I don't understand how this one matches. How does `app-test-some-other` match `app-test-*-some_other-*`? We shouldn't match on the `_` character, and we're missing the `-` after `other`.

rest-api-spec/src/main/resources/rest-api-spec/test/indices.put_index_template/10_basic.yml
ThyTran1402: Yup. So, index template `app-test-some` is the name of the template being registered and index patterns `app-test-*-some-*` its actual index pattern. 
The warning: `app-test-*-some-*` overlaps with global's `*`, not that `app-test-some` (the name) matches the pattern. The overlap is between `app-test-*-some-*` and `*` (the global template). Since `*` matches everything, it overlaps with any pattern, including `app-test-*-some-*`.

Hope this will make sense to you.

server/src/main/java/org/opensearch/cluster/metadata/MetadataIndexTemplateService.java
ThyTran1402: Yes. So, for the underscore case: 
- `app-test-*-some-*` vs `app-test-*-some_other-*`                                                                                                                                                      Full automaton intersection finds the synthetic string `app-test--some_other--some-` which matches both. But that string would never appear as a real index name (double dashes, contrived structure).                                                                                                                                                                                        The minimum-string heuristic: strip `*` from each → `app-test--some-` and `app-test--some_other-`. Neither matches the other's pattern → correctly reports no overlap.       
- for the hyphen case:
- `app-test-*-some-*` vs `app-test-*-some-other-*`
Strip `*` from `app-test-*-some-other-*` → `app-test--some-other-`
- `* = ""`, then `-some-`, then `* = other-`
The heuristic correctly flags this as a real overlap because a real index name like `app-test-x-some-other-y` would genuinely match both patterns simultaneously.

Let me know what your thoughts on this.

rest-api-spec/src/main/resources/rest-api-spec/test/indices.put_index_template/10_basic.yml
ThyTran1402: hmmm interesting. I think app-test-some-other never gets matched against `app-test-*-some_other-*`. The template name and the index pattern it contains are two independent things with the GET just retrieves the template by name, and the match verifies what was stored inside it.

- `indices.get_index_template: name: app-test-some-other `, this is a GET request that fetches the template by its registered name (`app-test-some-other`). It's a direct key lookup, not a wildcard match.  
- `match: {index_templates.0.index_template.index_patterns: ["app-test-*-some_other-*"]}`. This is just asserting that the response body's index_patterns field equals `["app-test-*-some_other-*"]` like verifying the template was stored correctly.   

Let me know what you think about this?

ThyTran1402:
> @ThyTran1402, can you please run `./gradlew :server:spotlessApply`, add the changed files, commit them, and push them to your PR branch?
> 
> We're getting the following build failure:
> 
> ```
> * What went wrong:
> Execution failed for task ':server:spotlessJavaCheck'.
> > The following files had format violations:
>       src/test/java/org/opensearch/cluster/metadata/MetadataIndexTemplateServiceTests.java
>           @@ -1318,9 +1318,7 @@
>            
>            ····public·void·testPatternsActuallyOverlap()·{
>            ········//·Patterns·that·share·a·genuine·conflict:·baz·matches·baz*
>           -········assertTrue(
>           ...
>           -············0L,
>           -············null,
>       ... (7 more lines that didn't fit)
>   Run './gradlew spotlessApply' to fix all violations.
> ```

Hi @msfroh 

I just pushed the changes. Can you review it again please?

Thank you.

msfroh:
Thanks, @ThyTran1402  -- I think I'm starting to understand.

The build is still failing, but that's just because it tries to check backward compatibility by running every test in a mixed-version cluster. In this case, because the change is going into OpenSearch 3.6.0, it doesn't work when mixed with OpenSearch 3.5 nodes or OpenSearch 2.19 nodes.

In your YamlRestTest, you just need to add a `skip` entry for versions less than `3.6.0` like:

```
---
"Put distinguishable multi-wildcard templates at the same priority":
  - skip:
      features: allowed_warnings
      version: " - 3.5.99"
      reason: "template overlap check was relaxed in 3.6"
```

ThyTran1402:
> Thanks, @ThyTran1402 -- I think I'm starting to understand.
> 
> The build is still failing, but that's just because it tries to check backward compatibility by running every test in a mixed-version cluster. In this case, because the change is going into OpenSearch 3.6.0, it doesn't work when mixed with OpenSearch 3.5 nodes or OpenSearch 2.19 nodes.
> 
> In your YamlRestTest, you just need to add a `skip` entry for versions less than `3.6.0` like:
> 
> ```
> ---
> "Put distinguishable multi-wildcard templates at the same priority":
>   - skip:
>       features: allowed_warnings
>       version: " - 3.5.99"
>       reason: "template overlap check was relaxed in 3.6"
> ```

Yup. Thank you for guiding me through this test checks. 
I just pushed the fix.

ThyTran1402:
Hi @msfroh 

Can you please review it?

Thank you.

ThyTran1402:
Hi @msfroh 

Do you know why it still failed some of the gradle check for jenkins? 😭

msfroh:
> Hi @msfroh 
> 
> 
> 
> Do you know why it still failed some of the gradle check for jenkins? 😭 

Let me look into this today. I got working on some other stuff and somehow convinced myself that this had already been merged.

ThyTran1402:
> > Hi @msfroh
> > Do you know why it still failed some of the gradle check for jenkins? 😭
> 
> Let me look into this today. I got working on some other stuff and somehow convinced myself that this had already been merged.

Please let me know if there're other test files or docs that I can look at to fix the tests 😭 .

ThyTran1402:
It still did not pass jenkins. Is there anything that I should do from my side to pass the gradle jenkins tests? @msfroh 
Since it seems like flaky test 😢

msfroh:
@ThyTran1402 Nothing needed from your part, I think. We do have a lot of flaky tests. I'll keep retrying after confirming that the test failures are not related to your PR. (They probably aren't, but it helps to check.)

ThyTran1402:
> @ThyTran1402 Nothing needed from your part, I think. We do have a lot of flaky tests. I'll keep retrying after confirming that the test failures are not related to your PR. (They probably aren't, but it helps to check.)

Is there a way that I can take a look or we can reduce the flaky tests? @msfroh

msfroh:
> Is there a way that I can take a look or we can reduce the flaky tests?

So, @andrross has actually just recently created a new dashboard for the flakiest of tests: https://metrics.opensearch.org/_dashboards/app/dashboards#/view/87737cf0-1409-11f1-a611-eb8fce42240d

He's been picking them off one-by-one, starting with the worst offenders.

ThyTran1402:
> > Is there a way that I can take a look or we can reduce the flaky tests?
> 
> So, @andrross has actually just recently created a new dashboard for the flakiest of tests: https://metrics.opensearch.org/_dashboards/app/dashboards#/view/87737cf0-1409-11f1-a611-eb8fce42240d
> 
> He's been picking them off one-by-one, starting with the worst offenders.

Oh this is awesome 💯 . I have not checked or seen this new dashboard before. I think it helped a lot for catching flaky tests. Thank you for your suggestion.

msfroh:
Oh -- looks like this finally passed on Friday evening. I spent the weekend sleeping off a nasty cold and didn't notice it.

Thank you @ThyTran1402 for fixing this issue! And a big thank you to @andrross for cleaning up so many of the flaky tests recently.

Issue #20721, opened 2026-02-24
[Feature Request] Support configurable document mapping in pull-based ingestion


imRishN:
### Is your feature request related to a problem? Please describe

Pull-based ingestion currently requires messages to be in a specific envelope format:

  ```json
  {"_id": "doc1", "_source": {"field": "value"}, "_op_type": "index"}
```

  Real-world streaming sources produce raw data that doesn't conform to this format. Users must run an external preprocessing layer (e.g., Flink, custom consumers) to transform raw messages into the expected envelope before publishing to an intermediate topic.

  This issue tracks adding configurable document mapping so users can tell OpenSearch which raw message fields map to _id, _version, and _op_type:

```json
  PUT /my-index
  {
    "settings": {
      "ingestion_source": {
        "type": "kafka",
        "mapper_type": "field_mapping",
        "mapper_settings.id_field": "user_id",
        "mapper_settings.version_field": "timestamp",
        "mapper_settings.op_type_field": "is_deleted"
      }
    }
  }
```

### Describe the solution you'd like

  Scope

  - mapper_settings.* prefix setting on IngestionSource for mapper-specific options
  - field_mapping mapper type that extracts _id, _version, _op_type from configurable top-level fields in the raw message
  - Extracted fields are removed from _source; remaining fields become the document source
  - Backward compatible: existing default and raw_payload mappers unchanged
  - Version compatibility check to prevent usage in mixed clusters

### Additional context

  Related

  - #16495 — [RFC] Streaming ingestion (pull based)
  - #16930 — [META] Pull-based Ingestion

PR #20722, opened 2026-02-24
Add mapper_settings support and field_mapping mapper type for pull-based ingestion

imRishN:
### Description
Adds a new `mapper_settings.*` prefix setting to `IngestionSource` that allows mapper-specific configuration to be passed to message mappers in pull-based ingestion. Also adds `field_mapping` as a new `MapperType` enum value with a version compatibility check to prevent usage in mixed clusters.

This is a foundational change — the settings are stored but not yet consumed. The actual `FieldMappingIngestionMessageMapper` implementation will follow in a subsequent PR.

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/20721

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/indices/pollingingest/mappers/IngestionMessageMapper.java
+++ b/server/src/main/java/org/opensearch/indices/pollingingest/mappers/IngestionMessageMapper.java
@@ -14,7 +14,6 @@ import org.opensearch.index.Message;
 import org.opensearch.indices.pollingingest.ShardUpdateMessage;
 
 import java.util.Locale;
-import java.util.Set;
 
 /**
  * Interface for mapping ingestion messages to ShardUpdateMessage format.
@@ -24,16 +23,6 @@ import java.util.Set;
  * validate and drop messages. Validations will be done as part of message processing in the {@link org.opensearch.indices.pollingingest.MessageProcessorRunnable}</p>
  */
 public interface IngestionMessageMapper {
-
-    /** Mapper setting key: source field to use as document _id */
-    String ID_FIELD = "id_field";
-    /** Mapper setting key: source field to use as document _version */
-    String VERSION_FIELD = "version_field";
-    /** Mapper setting key: source field to determine operation type (index vs delete) */
-    String OP_TYPE_FIELD = "op_type_field";
-
-    /** Valid mapper_settings keys for the field_mapping mapper type */
-    Set<String> FIELD_MAPPING_VALID_SETTINGS = Set.of(ID_FIELD, VERSION_FIELD, OP_TYPE_FIELD);
 
     /**
      * Maps and processes an ingestion message to a shard update message.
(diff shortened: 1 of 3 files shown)


server/src/test/java/org/opensearch/cluster/metadata/MetadataCreateIndexServiceTests.java
```diff
@@ -3793,4 +3793,89 @@ public void testFieldPropertyOverrideInParseV1Mappings() throws Exception {
...
+        ClusterState state = ClusterState.builder(ClusterName.CLUSTER_NAME_SETTING.getDefault(Settings.EMPTY)).nodes(nodes).build();
+
+        Settings settings = Settings.builder().put(IndexMetadata.SETTING_INGESTION_SOURCE_MAPPER_TYPE, "field_mapping").build();
+
+        IllegalArgumentException e = expectThrows(
+            IllegalArgumentException.class,
+            () -> MetadataCreateIndexService.validateIngestionSourceSettings(settings, state)
+        );
+        assertTrue(e.getMessage().contains("mapper_type [field_mapping] requires all nodes"));
+        assertTrue(e.getMessage().contains(Version.V_3_5_0.toString()));
```
varunbharadwaj, line 3828: Should we test a mix of 3.5 and 3.6 instead of old version as that is the most common case of failure for this use case?

server/src/main/java/org/opensearch/indices/pollingingest/mappers/IngestionMessageMapper.java
```diff
@@ -24,6 +25,16 @@
  */
 public interface IngestionMessageMapper {
 
+    /** Mapper setting key: source field to use as document _id */
+    String ID_FIELD = "id_field";
```
varunbharadwaj: ID_FIELD, VERSION_FIELD and OP_TYPE_FIELD are `field_mapping` specific constants. Should we move them into the field mapper class instead of the common interface? In `MetadataCreateIndexService.validateIngestionSourceSettings mapperType`, we already check the mapper type, so we could just do something like FieldMapperMessageMapper.getValidSettings(). What do you think?

server/src/test/java/org/opensearch/cluster/metadata/MetadataCreateIndexServiceTests.java
imRishN: The test already does this - 

```
DiscoveryNodes nodes = DiscoveryNodes.builder()                                                                                                                                                                                   
      .add(newNode("node1"))        // V_3_6_0 (CURRENT = V_3_6_0)                                                                                                                                                                  
      .add(oldNode)                 // V_3_5_0                                                                                                                                                                                      
      .build();
```

  newNode("node1") creates a node with Version.CURRENT which is V_3_6_0, and oldNode is V_3_5_0. So it's already testing the mixed 3.5 + 3.6 cluster scenario. The minNodeVersion resolves to V_3_5_0 which triggers the validation error.

server/src/main/java/org/opensearch/indices/pollingingest/mappers/IngestionMessageMapper.java
imRishN: Updated

Tarun-kishore (approved):

PR #20727, opened 2026-02-25
Prevent request corruption by avoiding failing entire bulk request on single shard rejection from shard backpressure

### LINKED ISSUE #20724: [BUG] Shard indexing pressure results in request corruption
### Describe the bug

OpenSearch uses unsafe buffers for bulk indexing requests by default. This was added for performance reasons, and overrides default netty logic to only release the buffers from the buffer pool when the HTTP response is ready. This prevents reusing the buffer while the request is in-progress, protecting the request payloads.

However, shard indexing pressure request rejection can corrupt the indexing payload in the following scenario:
1. Assume the bulk request contains shards local to the coordinating node, along with other shards which are overloaded.
2. Local shard is processed first, and submitted for processing.
3. Later when we reach processing the overloaded shard, it is rejected due to shard indexing pressure. This results in an exception, which will trigger the failure callback.
4. Bulk request will be failed, and the buffers will be released.
5. At this point, the local shard request is still in-progress, but the released buffer is recycled and overwritten.
6. This results in corrupted indexing payloads at the engine layer. It could also be possible for the engine to successfully index, and corruption only visible in the translog entry, failing on recovery.

The same issue also happens on the replica, but rarely compared to the primary depending on the timing.

One potential solution to verify this is to move the index shard pressure checks before any shard request is submitted. Alternatively, we can wait for the already submitted shards to finish before failing the bulk request.

We try the first option to confirm this solves the payload corruption. We can see below that parsing errors stop after making this change.

Image

### Related component

Indexing

### To Reproduce

The issue can be seen on OS 3.x versions (3.4 is where it was observed, and issue also exists on latest main branch - unreleased 3.6). Issue depends on race conditions, when bulk request fails partially, with new requests overwriting in-flight payloads.

### Expected behavior

There should be no request payload corruption

### Additional Details

_No response_

varunbharadwaj:
### Description
Shard backpressure rejections result in failing entire bulk request, releasing unsafe buffer which can cause request corruption. More details can be found in #20724. This PR updates the logic to treat shard level pressure check failures as regular shard failures, without failing the entire bulk request. This way, we continue to process other shards, and return shard level errors in the bulk response.

This change also solves the request corruption problem when shard level pressure check is enabled.

### Related Issues
Resolves #20724

### DIFF
--- a/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java
+++ b/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java
@@ -120,6 +120,7 @@ import java.util.SortedMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicIntegerArray;
+import java.util.function.Consumer;
 import java.util.function.LongSupplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -723,6 +724,8 @@ public class TransportBulkAction extends HandledTransportAction<BulkRequest, Bul
             final DocStatusStats docStatusStats = new DocStatusStats();
             String nodeId = clusterService.localNode().getId();
 
+            final boolean isOnlySystem = isOnlySystem(bulkRequest, clusterService.state().metadata().getIndicesLookup(), systemIndices);
+
             for (Map.Entry<ShardId, List<BulkItemRequest>> entry : requestsByShard.entrySet()) {
                 final ShardId shardId = entry.getKey();
                 final List<BulkItemRequest> requests = entry.getValue();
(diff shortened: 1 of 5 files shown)


andrross:
For batch style APIs in general, the best practice is that the overall request only fails with a non-200 response for things like failing authentication or invalid request syntax. Shard backpressure rejections are clearly shard-level failures and as such only apply to part of the batch _bulk request, so it seems that it is clearly a bug to fail the entire request as opposed to sending a 200 with individual document-level failures for the documents that were rejected due to backpressure. I mean, obviously request corruption is a bug but the behavior is wrong even if corruption wasn't possible. Do I understand that correctly?

However, tuning clients to handle backpressure properly is often a delicate balance so I do worry about the impact if today the service sends 429s and after this fix it starts sending 200s with the backpressure details in the response. @msfroh @Bukhtawar what do you think?

varunbharadwaj:
> I mean, obviously request corruption is a bug but the behavior is wrong even if corruption wasn't possible. Do I understand that correctly?

Yeah, that was the idea behind this approach.

>However, tuning clients to handle backpressure properly is often a delicate balance so I do worry about the impact if today the service sends 429s and after this fix it starts sending 200s with the backpressure details in the response.

This is a good point, agree. I was thinking if someone is relying on the 429s from shard level back-pressure mechanism, they are very likely impacted by the request corruption issue. Another alternative tested was to first apply the shard level checks on all shards (phase 1) and later send the shard requests (phase 2). So we fail the bulk request if any shard is overloaded. This will maintain better compatibility with current behavior. Open to suggestions on which approach might be preferred here.

andrross:
> Another alternative tested was to first apply the shard level checks on all shards (phase 1) and later send the shard requests (phase 2). So we fail the bulk request if any shard is overloaded.

Honestly this seems like a less risky approach, but I'd like to get some opinions from others familiar with the indexing path, e.g. @Bukhtawar @mgodwan @gbbafna 

It seems like there might be other aspects to this behavior change as well. With this change, in case of a single slow shard then indexing traffic will proceed a full throughput to the other shards. This seems good but could lead to data skew across shards if the slowness persists. That could be a big change to a system relying on the current behavior.

server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java
```diff
@@ -737,20 +740,50 @@ protected void doRun() {
...
+
+                        docStatusStats.inc(bulkItemResponse.status());
+                        responses.set(request.id(), bulkItemResponse);
+                    }
+
+                    if (counter.decrementAndGet() == 0) {
+                        indicesService.addDocStatusStats(docStatusStats);
+                        listener.onResponse(
+                            new BulkResponse(responses.toArray(new BulkItemResponse[responses.length()]), buildTookInMillis(startTimeNanos))
+                        );
```
Bukhtawar, line 770: Can we avoid duplicating finishHim()?

server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java
```diff
@@ -842,7 +859,8 @@ private void finishHim() {
                     }
                     span.setError(e);
                     span.endSpan();
-                    throw e;
+                    // Treat execution failures as shard-level failures rather than failing the entire bulk request
+                    onShardFailure.accept(e);
```
Bukhtawar, line 863: Looks like we need a test for this, unless an existing test already covers this?

Bukhtawar:
> > Another alternative tested was to first apply the shard level checks on all shards (phase 1) and later send the shard requests (phase 2). So we fail the bulk request if any shard is overloaded.
> 
> Honestly this seems like a less risky approach, but I'd like to get some opinions from others familiar with the indexing path, e.g. @Bukhtawar @mgodwan @gbbafna
> 
> It seems like there might be other aspects to this behavior change as well. With this change, in case of a single slow shard then indexing traffic will proceed a full throughput to the other shards. This seems good but could lead to data skew across shards if the slowness persists. That could be a big change to a system relying on the current behavior.

Do you think this can be enabled behind a bulk query param while still keeping the existing behavior as default?

PR #20729, opened 2026-02-25
Implement FieldMappingIngestionMessageMapper for pull-based ingestion

### LINKED ISSUE #20728: [Feature Request] Implement field_mapping mapper type for pull-based ingestion

Implements the `field_mapping` mapper type introduced in #20722, enabling pull-based ingestion to extract document metadata from configurable fields in raw stream messages.

### Mapper settings

  | Setting | Description |
  |---------|-------------|
  | `id_field` | Source field to use as document `_id`. Auto-generated if absent. |
  | `version_field` | Source field to use as document `_version` (external versioning). |
  | `op_type_field` | Source field to determine operation type (index vs delete). |
  | `op_type_field.delete_value` | Optional. The value that indicates a delete. If not set, field is parsed as boolean. 

### Example

  ```json
  PUT /my-index
  {
    "settings": {
      "ingestion_source": {
        "type": "kafka",
        "mapper_type": "field_mapping",
        "mapper_settings.id_field": "user_id",
        "mapper_settings.version_field": "timestamp",
        "mapper_settings.op_type_field": "is_deleted"
      }
    }
  }
```

  Raw message `{"user_id": "abc", "timestamp": 123, "is_deleted": false, "name": "alice"}` gets mapped to `{_id=abc, _version=123, _op_type=index, _source={"name": "alice"}`.

### Related component

_No response_

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by imRishN**
### Description
Implements the field_mapping mapper type that extracts `_id`, `_version`, and `_op_type` from configurable top-level fields in raw stream messages. Extracted fields are removed from `_source`. Supports boolean op_type detection by default, with an optional `delete_value` setting for custom value matching (e.g., "Y"/"N", "expired", integer flags).

Also updates the mapper factory to accept mapper settings and wires mapper settings from `IngestionSource` through `DefaultStreamPoller`.

### Related Issues
Resolves #20728

### Check List

### CODE REVIEW

File: server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java [hunk 1]
```diff
@@ -35,13 +47,104 @@ public class FieldMappingIngestionMessageMapper implements IngestionMessageMappe
...
     /** Mapper setting key: source field to determine operation type (index vs delete) */
     public static final String OP_TYPE_FIELD = "op_type_field";
+    /** Mapper setting key: the value of op_type_field that indicates a delete operation */
+    public static final String DELETE_VALUE = "op_type_field.delete_value";
 
     /** Valid mapper_settings keys for this mapper type */
-    public static final Set<String> VALID_SETTINGS = Set.of(ID_FIELD, VERSION_FIELD, OP_TYPE_FIELD);
+    public static final Set<String> VALID_SETTINGS = Set.of(ID_FIELD, VERSION_FIELD, OP_TYPE_FIELD, DELETE_VALUE);
+
+    private static final String OP_TYPE_INDEX = "index";
```
varunbharadwaj: Is it possible to extract these constants out? RawPayloadIngestionMessageMapper also seems to define OP_TYPE_INDEX.

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
```diff
@@ -35,13 +47,104 @@ public class FieldMappingIngestionMessageMapper implements IngestionMessageMappe
...
+        payloadMap.put(SOURCE, rawPayload);
+
+        return new ShardUpdateMessage(pointer, message, payloadMap, autoGeneratedIdTimestamp);
+    }
+
+    /**
+     * Validates that a field value is a scalar type (string, number, boolean) and not a nested object or array.
+     */
+    private static void validateScalar(String fieldName, Object value) {
+        if (value instanceof Map || value instanceof Iterable) {
```
varunbharadwaj: Does this check work for other types like arrays? Do you think we can just validate the type is a String or a Number? Would that work?

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
```diff
@@ -35,13 +47,104 @@ public class FieldMappingIngestionMessageMapper implements IngestionMessageMappe
...
+        Map<String, Object> rawPayload = IngestionUtils.getParsedPayloadMap((byte[]) message.getPayload());
+        Map<String, Object> payloadMap = new HashMap<>();
+
+        // Extract _id
+        long autoGeneratedIdTimestamp = UNSET_AUTO_GENERATED_TIMESTAMP;
+        if (idField != null && rawPayload.containsKey(idField)) {
+            Object idValue = rawPayload.remove(idField);
+            validateScalar(idField, idValue);
+            String idStr = String.valueOf(idValue);
+            if (idStr.isEmpty() || "null".equals(idStr)) {
```
varunbharadwaj: Why do we want to ignore "null" values? I think it would be difficult to guess what value user topic might be using - null, none, etc. I think it is better to just ensure the value is not empty and avoid additional checks?
Should we also call trim() and then validate it is not empty?

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
```diff
@@ -35,13 +47,104 @@ public class FieldMappingIngestionMessageMapper implements IngestionMessageMappe
...
+                payloadMap.put(ID, idStr);
+            }
+        } else {
+            String id = RequestUtils.generateID();
+            payloadMap.put(ID, id);
+            autoGeneratedIdTimestamp = System.currentTimeMillis();
+        }
+
+        // Extract _version
+        if (versionField != null && rawPayload.containsKey(versionField)) {
```
varunbharadwaj: For version field, I think we should always receive a version or never receive it. External versioning check is only applied when `_version` field is present. Should we fail if user specified a version field and we don't find it? It would prevent accidentally overwriting newer version document if version is missing.

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
```diff
@@ -35,13 +47,104 @@ public class FieldMappingIngestionMessageMapper implements IngestionMessageMappe
...
+            Object val = rawPayload.remove(opTypeField);
+            validateScalar(opTypeField, val);
+            String stringVal = String.valueOf(val);
+            boolean isDelete;
+            if (deleteValue != null) {
+                // Compare against configured delete value
+                isDelete = deleteValue.equals(stringVal);
+            } else {
+                // Default: parse as boolean
+                isDelete = Boolean.parseBoolean(stringVal);
```
varunbharadwaj: I'm wondering if we could just ask the user to specify the delete value if they want to support deletes. By default, we could use `index`. Interpreting it as a boolean only when user doesn't specify a delete value seems confusing to me 🤔

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
imRishN: Good point. Updating it

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
imRishN: Makes sense. Updating it

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
imRishN: This is a good point actually. If the user explicitly configured version_field, they want every document to be version-checked. A missing version field can be a bad data. 

I will make change to see that if version_field is configured but missing from the message, it should throw exception  and then the error strategy (BLOCK/DROP) can handle

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
imRishN: Published this change

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
imRishN: Updated

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
imRishN: Updated

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
imRishN: Made the change. Let me know if this looks good?

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
imRishN: Hmm I understand your point. The current behavior is:                                                                                                                                                                                            
```                                                                                                                                                                                                                                    
  - delete_value not set → parse as boolean (true = delete)                                                                                                                                                                         
  - delete_value set → string comparison  
 ```

In above case the user has to know that without delete_value, their field is treated as boolean, but with it, it's treated as a string match. 

I am thinking to simplify this - `op_type_field` + `delete_value` go together. If the user wants deletes, they provide both. If they don't care about deletes, they don't configure op_type_field at all. We could even validate that `delete_value` requires op_type_field to be set. Wdyt?

Comment (varunbharadwaj) [reply server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java]: Yeah, this sounds good to me. While we are at this, we could also add support for `create` to support all existing types.

Comment (varunbharadwaj) [reply server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java]: We can just resolve the ordering something like create first, then delete, if user provides same value for both - or add a validation to make sure they are not the same.

Comment (imRishN) [reply server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java]: Made the changes and added support for create value

File: server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java [hunk 6]
```diff
@@ -22,26 +32,149 @@
...
+        }
+
+        // Extract _version
+        if (versionField != null) {
+            if (rawPayload.containsKey(versionField) == false) {
+                throw new IllegalArgumentException("configured version_field [" + versionField + "] is missing from the message");
+            }
+            Object versionValue = rawPayload.remove(versionField);
+            validateScalar(versionField, versionValue);
+            payloadMap.put("_version", String.valueOf(versionValue).trim());
```
varunbharadwaj: We can use the constant `VersionFieldMapper.NAME` for `_version`

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -1309,6 +1309,28 @@ static void validateIngestionSourceSettings(Settings settings, ClusterState stat
                         );
                     }
                 }
+                // Validate cross-key dependencies
```
varunbharadwaj: Would it make sense to update the IngestionMessageMapper interface with a `validate(mapperType, mapperSettings)` method, and move the mapper_settings keys validations and the delete/create value checks into the FieldMappingIngestionMessageMapper?

server/src/main/java/org/opensearch/indices/pollingingest/mappers/FieldMappingIngestionMessageMapper.java
```diff
@@ -22,26 +33,175 @@
...
+            autoGeneratedIdTimestamp = System.currentTimeMillis();
+        }
+
+        // Extract _version
+        if (versionField != null) {
+            if (rawPayload.containsKey(versionField) == false) {
+                throw new IllegalArgumentException("configured version_field [" + versionField + "] is missing from the message");
+            }
+            Object versionValue = rawPayload.remove(versionField);
+            validateScalar(versionField, versionValue);
```
varunbharadwaj, line 153: Should version always be numeric?

varunbharadwaj: Let's revisit this in the next PR along with the ITs. Resolving it for now.

msfroh:
As a suggestion, @imRishN and @varunbharadwaj -- don't merge from main into a PR branch. Instead, rebase the PR branch on top of main.

The squashed commit has a very noisy commit message as it picked up a bunch of other commits that were merged into the PR branch: https://github.com/opensearch-project/OpenSearch/commit/4ecd7206a7b27c072f2bc664498d293f323dd8d8

varunbharadwaj:
@msfroh yeah, I realized this after merging it. Discussed offline with Peter to see if it is possible to amend the commit to fix it - but looks like it is not possible. Will double check the commits next time.

PR #20733, opened 2026-02-26
[Backport 3.5] Fix stream transport TLS cert hot-reload by using live SSLContext fro…

rishabhmaurya:
…m SecureTransportSettingsProvider

### Description
backport of https://github.com/opensearch-project/OpenSearch/pull/20734
Stream transport failed after reload certs with "cert file read errors". The original `DefaultSslContextProvider` built `JdkSslContext` using `KeyManagerFactory/TrustManagerFactory` obtained from SslConfiguration, which reads cert files from disk.It can happen that some implementation calls reloadcerts which rebuild the in-memory `SSLContext` and then delete the cert files. Any new connection after deletion triggers a fresh `JdkSslContext` build, which fails because the files no longer exist.

**Why SecureTransportSettingsProvider needed a new method**
`transport-netty4` avoids this by calling `buildSecureServerTransportEngine() / buildSecureClientTransportEngine()` per connection — these return engines from the live in-memory `SSLContext` with no file reads. However, Arrow Flight uses gRPC-netty's `ServerTlsHandler / ClientTlsHandler`, which require a `SslContext` object (not a raw `SSLEngine`). The correct pattern (used by `transport-grpc`) is to wrap the live `javax.SSLContext` once in a `JdkSslContext` — `JdkSslContext` holds a reference to the `SSLContext` and calls `createSSLEngine()` on it per connection, so re-initializing the `SSLContext` with new key material takes effect on the next handshake automatically.

`transport-grpc` gets the live `SSLContext` via `SecureAuxTransportSettingsProvider.buildSecureAuxServerTransportContext()`, but Arrow Flight is a `Transport` (not `AuxTransport`) and receives `SecureTransportSettingsProvider`, which had no equivalent method. Adding `buildSecureTransportContext()` (default returns `Optional.empty()`) exposes the live `SSLContext` through the same interface Arrow Flight already has.

**Security plugin change**
The security plugin implements `buildSecureTransportContext()` to return the live `SSLContext` from `SslContextHandler` — one line, identical to how `buildSecureAuxServerTransportContext()` works. See: https://github.com/opensearch-project/security/pull/5971

**Hostname verification and ALPN**
With the client using `JdkSslContext` directly, `gRPC-netty` `ClientTlsHandler` behaviors required additional handling:

1. **ALPN**: `ClientTlsHandler` reads `getSSLParameters()`, adds `endpointIdentificationAlgorithm`, then calls `setSSLParameters()`. `ALPN` must already be set on the engine before this round-trip, so `AlpnPresettingClientSslContext` pre-sets it in `newEngine()`.

2. **Hostname verification**: `ClientTlsHandler` unconditionally `sets endpointIdentificationAlgorithm="HTTPS"` after newEngine() returns, regardless of the `SslContext` configuration. When `enforce_hostname_verification=false`, this causes handshake failures if the server cert lacks the peer IP as a SAN. `AlpnAwareSSLEngineWrapper` intercepts `setSSLParameters()` and strips the algorithm. It must live in the `io.netty.handler.ssl `package to implement `ApplicationProtocolAccessor` (package-private in Netty), which is required for `SslHandler.applicationProtocol()` to return the negotiated protocol post-handshake

### Related Issues

### Check List


rishabhmaurya:
PR Code Analyzer:
Finding 1 (High): 
Intentional — the wrapper is only instantiated when `enforceHostnameVerification=false` (operator-configured), and stripping is necessary to prevent `gRPC-netty's` `ClientTlsHandler` from unconditionally re-enabling hostname verification after `newEngine()` returns.

Finding 2 (Medium): 
Intentional — `ApplicationProtocolAccessor` is package-private in Netty and cannot be implemented from outside `io.netty.handler.ssl;` there is no other way which i could find to expose the negotiated ALPN protocol to `SslHandler.applicationProtocol()` without forking Netty or using reflection.

jainankitk (approved):

PR #20734, opened 2026-02-26
Fix stream transport TLS cert hot-reload by using live SSLContext fro…

rishabhmaurya:
…m SecureTransportSettingsProvider

### Description
Stream transport failed after reloadcerts with "cert file read errors". The original `DefaultSslContextProvider` built `JdkSslContext` using `KeyManagerFactory/TrustManagerFactory` obtained from SslConfiguration, which reads cert files from disk.It can happen that some implementation calls reloadcerts which rebuild the in-memory `SSLContext` and then delete the cert files. Any new connection after deletion triggers a fresh `JdkSslContext` build, which fails because the files no longer exist.

**Why SecureTransportSettingsProvider needed a new method**
`transport-netty4` avoids this by calling `buildSecureServerTransportEngine() / buildSecureClientTransportEngine()` per connection — these return engines from the live in-memory `SSLContext` with no file reads. However, Arrow Flight uses gRPC-netty's `ServerTlsHandler / ClientTlsHandler`, which require a `SslContext` object (not a raw `SSLEngine`). The correct pattern (used by `transport-grpc`) is to wrap the live `javax.SSLContext` once in a `JdkSslContext` — `JdkSslContext` holds a reference to the `SSLContext` and calls `createSSLEngine()` on it per connection, so re-initializing the `SSLContext` with new key material takes effect on the next handshake automatically.

`transport-grpc` gets the live `SSLContext` via `SecureAuxTransportSettingsProvider.buildSecureAuxServerTransportContext()`, but Arrow Flight is a `Transport` (not `AuxTransport`) and receives `SecureTransportSettingsProvider`, which had no equivalent method. Adding `buildSecureTransportContext()` (default returns `Optional.empty()`) exposes the live `SSLContext` through the same interface Arrow Flight already has.

**Security plugin change**
The security plugin implements `buildSecureTransportContext()` to return the live `SSLContext` from `SslContextHandler` — one line, identical to how `buildSecureAuxServerTransportContext()` works. See: https://github.com/opensearch-project/security/pull/5971

**Hostname verification and ALPN**
With the client using `JdkSslContext` directly, `gRPC-netty` `ClientTlsHandler` behaviors required additional handling:

1. **ALPN**: `ClientTlsHandler` reads `getSSLParameters()`, adds `endpointIdentificationAlgorithm`, then calls `setSSLParameters()`. `ALPN` must already be set on the engine before this round-trip, so `AlpnPresettingClientSslContext` pre-sets it in `newEngine()`.

2. **Hostname verification**: `ClientTlsHandler` unconditionally `sets endpointIdentificationAlgorithm="HTTPS"` after newEngine() returns, regardless of the `SslContext` configuration. When `enforce_hostname_verification=false`, this causes handshake failures if the server cert lacks the peer IP as a SAN. `AlpnAwareSSLEngineWrapper` intercepts `setSSLParameters()` and strips the algorithm. It must live in the `io.netty.handler.ssl `package to implement `ApplicationProtocolAccessor` (package-private in Netty), which is required for `SslHandler.applicationProtocol()` to return the negotiated protocol post-handshake

### Related Issues

### Check List


CHANGELOG.md
```diff
@@ -15,6 +14,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- Introduce new libs/netty4 module to share common implementation between netty-based plugins and modules (transport-netty4, transport-reactor-netty4) ([#20447](https://github.com/opensearch-project/OpenSearch/pull/20447))
+- Add validation to make crypto store settings immutable ([#20123](https://github.com/opensearch-project/OpenSearch/pull/20123))
+- Introduce concurrent translog recovery to accelerate segment replication primary promotion ([#20251](https://github.com/opensearch-project/OpenSearch/pull/20251))
+- Update to `almalinux:10` ([#20482](https://github.com/opensearch-project/OpenSearch/pull/20482))
+- Add X-Request-Id to uniquely identify a search request ([#19798](https://github.com/opensearch-project/OpenSearch/pull/19798))
+- Added TopN selection logic for streaming terms aggregations ([#20481](https://github.com/opensearch-project/OpenSearch/pull/20481))
+- Added support for Intra Segment Search ([#19704](https://github.com/opensearch-project/OpenSearch/pull/19704))
+- Introduce AdditionalCodecs and EnginePlugin::getAdditionalCodecs hook to allow additional Codec registration ([#20411](https://github.com/opensearch-project/OpenSearch/pull/20411))
```
harshavamsi, line 36: @rishabhmaurya looks like we added to the changelog

Issue #20756, opened 2026-03-02
[Feature Request] Wire IngestService into pull-based ingestion path


imRishN:
### Is your feature request related to a problem? Please describe

Wires `IngestService` from `Node` to `IngestionEngine` to enable ingest pipeline support in pull-based ingestion. This will be wiring-only change with no behavioral impact — `IngestService` is stored but not used yet.

`IngestService` is constructed after `IndicesService` in `Node.java`, and `IngestService` takes `IndicesService` as a constructor arg. This creates a dependency that prevents passing `IngestService` directly to `IndicesService`.

### Describe the solution you'd like

We can use `SetOnce` + `Supplier` late-binding pattern to break the dependency:

  Node (SetOnce)
    → IndicesService (Supplier)
      → IngestionEngineFactory (Supplier)
        → IngestionEngine (@Nullable IngestService)

The supplier can only be resolved when the first ingestion engine starts, by which time `IngestService` is fully constructed.

PR #20757, opened 2026-03-02
Wire IngestService from Node to IngestionEngine for pull-based ingestion

imRishN:
### Description
Wires `IngestService` into the pull-based ingestion path using the `SetOnce` supplier pattern to resolve the dependency between `IndicesService` and `IngestService`. `IngestService` is now accessible in IngestionEngine but not yet used. Future PRs will wire it to MessageProcessorRunnable and implement pipeline execution.

Wiring chain: 
Node (SetOnce<IngestService>) → IndicesService (Supplier<IngestService>) → IngestionEngineFactory → IngestionEngine.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/20756
### Check List


server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -60,11 +61,22 @@ public class IngestionEngine extends InternalEngine {
...
     private final DocumentMapperForType documentMapperForType;
+    @Nullable
+    private final IngestService ingestService;
     private volatile IngestionShardPointer lastCommittedBatchStartPointer;
 
     public IngestionEngine(EngineConfig engineConfig, IngestionConsumerFactory ingestionConsumerFactory) {
+        this(engineConfig, ingestionConsumerFactory, null);
+    }
+
+    public IngestionEngine(
```
monika48: Please add a test case to cover this constructor with both null and non-null IngestService

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -1001,6 +1002,7 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
                 searchRequestStats,
                 remoteStoreStatsTrackerFactory,
                 ingestionConsumerFactories,
+                ingestServiceReference::get,
```
monika48, line 1005: +1 to this review comment


> [SetOnce Race Condition](https://github.com/opensearch-project/OpenSearch/pull/20757/files#diff-f839350640f8bd27f602b3249215252bf07afef5d7e480b23271788af6cda331R1005-R1027)
> ingestServiceReference is a SetOnce that is passed as a supplier to IndicesService before ingestService is created and set. If any index shard is created or the engine factory is invoked before ingestServiceReference.set(ingestService) is called, ingestServiceSupplier.get() will return null. This timing dependency should be validated to ensure no engine creation can happen before the reference is set.

imRishN: This is safe in this case as the construction is all single-threaded in `Node.java` in the order - IndicesService -> then IngestService and then set(). And no shards will be created during this because node hasn't joined the cluster yet. The supplier (ingestServiceReference::get) is only resolved when `IngestionEngineFactory.newReadWriteEngine()` is invoked during shard creation, which is well after all services are constructed. Also we are using the same pattern for `repositoriesServiceReference`. 

Also, `IngestionEngineFactory.newReadWriteEngine()` already handles a null return from the supplier: `IngestService ingestService = ingestServiceSupplier != null ? ingestServiceSupplier.get() : null;`. The engine will simply run without pipeline support if IngestService is somehow null.

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
imRishN: Added

server/src/main/java/org/opensearch/node/Node.java
varunbharadwaj: >The engine will simply run without pipeline support if IngestService is somehow null. 

If the service is null and we skip document transformation, we can index incorrect payloads. As you mentioned, shard initialization is done later and this might never be the case, but lets just double check, and add an assertion maybe to validate.

server/src/main/java/org/opensearch/indices/pollingingest/IngestionEngineFactory.java
```diff
@@ -9,24 +9,34 @@
...
 import org.opensearch.index.engine.NRTReplicationEngine;
+import org.opensearch.ingest.IngestService;
 
 import java.util.Objects;
+import java.util.function.Supplier;
 
 /**
  * Engine Factory implementation used with streaming ingestion.
  */
 public class IngestionEngineFactory implements EngineFactory {
```
varunbharadwaj, line 27: The SetOnce pattern seems fine to me. I'm just thinking if it'll be better to do something as follows.

We create something like a `MessageTransformer` interface. All pull-based ingestion components starting from the IngestionFactory, IngestionEngine, etc accept the MessageTransformer. Then in the `IndicesService.getEngineFactory()`, when IngestionEngineFactory is initialized, we pass in a MessageTransformer instance (IngestPipelineTransformer) wrapping the IngestService (maybe using the SetOnce pattern like we do currently or directly). 

That way, pull-based ingestion flow need not depend directly on anything big above the engine layer. We can also restrict exposing only the pipeline resolving and document transformation functionality in the PBI flow this way.

@imRishN @monika48 What do you both think?

On a side note, PBI can probably support ingest pipelines without INGEST role nodes. We can check and highlight this later on (maybe in IngestPipelineTransformer if we decide that route).

imRishN: The message transformer is a good idea and this abstraction might also give cleaner separation. However from what I could think, the transformer would need to handle pipeline resolution, threadpool management (async to sync) and additional guardrails and so the interface would end up being more than a simple transform. And the interface would just be having only one implementation so in a way its adding a layer without reducing complexity

server/src/main/java/org/opensearch/node/Node.java
imRishN: Added an assertion

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -60,11 +61,22 @@ public class IngestionEngine extends InternalEngine {
     private StreamPoller streamPoller;
     private final IngestionConsumerFactory ingestionConsumerFactory;
     private final DocumentMapperForType documentMapperForType;
+    @Nullable
```
varunbharadwaj: Should we remove the @Nullable as we never expect a null value here? We can do this in the next PR as well, don't want to block on this one.

server/src/main/java/org/opensearch/indices/pollingingest/IngestionEngineFactory.java
```diff
@@ -9,24 +9,34 @@
...
 
     private final IngestionConsumerFactory ingestionConsumerFactory;
+    @Nullable
+    private final Supplier<IngestService> ingestServiceSupplier;
 
     public IngestionEngineFactory(IngestionConsumerFactory ingestionConsumerFactory) {
```
varunbharadwaj: Is this constructor used? If the only usage is in the newly added test in this PR, we can remove it.

imRishN: Yeah, good point. Removed it

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
imRishN: Yup removed

server/src/main/java/org/opensearch/indices/pollingingest/IngestionEngineFactory.java
varunbharadwaj: Discussed offline. We will get back to this topic and refactor once we have the changes for the IngestService integration added in the next PR.

server/src/main/java/org/opensearch/indices/pollingingest/IngestionEngineFactory.java
```diff
@@ -9,24 +9,30 @@
...
 import java.util.Objects;
+import java.util.function.Supplier;
 
 /**
  * Engine Factory implementation used with streaming ingestion.
  */
 public class IngestionEngineFactory implements EngineFactory {
 
     private final IngestionConsumerFactory ingestionConsumerFactory;
+    @Nullable
```
varunbharadwaj, line 30: Do we need the supplier here to be nullable?

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -60,11 +61,17 @@ public class IngestionEngine extends InternalEngine {
...
     private final IngestionConsumerFactory ingestionConsumerFactory;
     private final DocumentMapperForType documentMapperForType;
+    private final IngestService ingestService;
     private volatile IngestionShardPointer lastCommittedBatchStartPointer;
 
     public IngestionEngine(EngineConfig engineConfig, IngestionConsumerFactory ingestionConsumerFactory) {
+        this(engineConfig, ingestionConsumerFactory, null);
+    }
+
+    public IngestionEngine(EngineConfig engineConfig, IngestionConsumerFactory ingestionConsumerFactory, IngestService ingestService) {
```
varunbharadwaj, line 71: Will we have a scenario where we would have a null ingestService? It seems to be only called from the IngestionFactory and tests. We can just keep 1 constructor. We can cleanup in the follow up PR.

server/src/main/java/org/opensearch/indices/IndicesService.java
```diff
@@ -666,6 +670,7 @@ public IndicesService(
             searchRequestStats,
             remoteStoreStatsTrackerFactory,
             ingestionConsumerFactories,
+            () -> null,
```
varunbharadwaj, line 673: Seems like this one is only used in the snapshot resiliency tests. This path would fail for the pull-based ingestion flow. We could leave a comment highlighting this is only for test.

Issue #20766, opened 2026-03-02
[Feature Request] Support profile API in gRPC


karenyrx:
### Is your feature request related to a problem? Please describe

Adding `"profile": true` in the gRPC search request is working already, and the profile is being collected, just not displayed in the grpc response due to lack of protobuf conversion in the response: https://github.com/opensearch-project/OpenSearch/blob/main/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/search/SearchResponseSectionsProtoUtils.java#L96, so throws: 
```
  "message": "\"code:internal message:java.lang.UnsupportedOperationException: profile results are not supported yet\\n\\tat org.opensearch.transport.grpc.proto.response.search.SearchResponseSectionsProtoUtils.checkUnsupportedFeatures(SearchResponseSectionsProtoUtils.java:65)\\n\\tat org.opensearch.transport.grpc.proto.response.search.SearchResponseSectionsProtoUtils.toProto(SearchResponseSectionsProtoUtils.java:43)\\n\\tat org.opensearch.transport.grpc.proto.response.search.SearchResponseProtoUtils.toProto(SearchResponseProtoUtils.java:96)\\n\\tat org.opensearch.transport.grpc.proto.response.search.SearchResponseProtoUtils.toProto(SearchResponseProtoUtils.java:40)\\n\\tat org.opensearch.transport.grpc.listeners.SearchRequestActionListener.onResponse(SearchRequestActionListener.java:45)\\n\\tat org.opensearch.transport.grpc.listeners.SearchRequestActionListener.onResponse(SearchRequestActionListener.java:26)\\n\\tat org.opensearch.action.support.TransportAction$1.onResponse(TransportAction.java:115)\\n\\tat org.opensearch.action.support.TransportAction$1.onResponse(TransportAction.java:109)\\n\\tat 
```

### Describe the solution you'd like

Support profile in SearchResponseSectionsProtoUtils

karenyrx:
Search:Query Capabilities

karenyrx:
Un-self-assigning as I won't be working on this in the short term. If anyone is interested, feel free to take this up

PR #20773, opened 2026-03-04
BUG: Fix NullPointerException in gRPC search response marshalling for…

allan918:
Description
This PR addresses a NullPointerException that occurs during gRPC search response marshalling when the totalHits field in the SearchResponse is null. This scenario happens when a search request is made with the track_total_hits parameter set to false.

The fix introduces a null check at the beginning of the processTotalHits method in [SearchHitsProtoUtils.java](code-assist-path:c:\Users\admin\OpenSearch\modules\transport-grpc\src\main\java\org\opensearch\transport\grpc\proto\response\search\SearchHitsProtoUtils.java). If hits.getTotalHits() is null, the method now correctly handles it by setting an empty total field in the protobuf message and returning early, thus preventing the NullPointerException.

Related Issues
Related to #20766

### DIFF
--- a/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/search/SearchHitsProtoUtils.java
+++ b/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/search/SearchHitsProtoUtils.java
@@ -69,14 +69,18 @@ public class SearchHitsProtoUtils {
     private static void processTotalHits(SearchHits hits, org.opensearch.protobufs.HitsMetadata.Builder hitsMetaData) {
         org.opensearch.protobufs.HitsMetadataTotal.Builder totalBuilder = org.opensearch.protobufs.HitsMetadataTotal.newBuilder();
 
+        if (hits.getTotalHits() == null) {
+            hitsMetaData.setTotal(totalBuilder.build());
+            return;
+        }
+
         // TODO need to pass parameters
         // boolean totalHitAsInt = params.paramAsBoolean(RestSearchAction.TOTAL_HITS_AS_INT_PARAM, false);
         boolean totalHitAsInt = false;
 
         if (totalHitAsInt) {
-            long total = hits.getTotalHits() == null ? -1 : hits.getTotalHits().value();
-            totalBuilder.setInt64(total);
-        } else if (hits.getTotalHits() != null) {
+            totalBuilder.setInt64(hits.getTotalHits().value());
+        } else {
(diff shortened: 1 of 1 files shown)


cwperks:
Thank you for the bugfix @allan918 . Would we be able to add a test to verify the fix as well?

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/search/SearchHitsProtoUtils.java
```diff
@@ -69,14 +69,18 @@ protected static void toProto(SearchHits hits, org.opensearch.protobufs.HitsMeta
...
+            hitsMetaData.setTotal(totalBuilder.build());
+            return;
+        }
+
         // TODO need to pass parameters
         // boolean totalHitAsInt = params.paramAsBoolean(RestSearchAction.TOTAL_HITS_AS_INT_PARAM, false);
         boolean totalHitAsInt = false;
 
         if (totalHitAsInt) {
-            long total = hits.getTotalHits() == null ? -1 : hits.getTotalHits().value();
```
karenyrx, line 77: Currently, the code matches the REST-side exactly 
https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/SearchHits.java#L238-L245

Does that mean this bug is present in both REST and gRPC side?
* If so, could we fix both sides? 
* If not, could you give an example REST and GRPC query and response, so we can better understand this problem?

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/search/SearchHitsProtoUtils.java
```diff
@@ -69,14 +69,18 @@ protected static void toProto(SearchHits hits, org.opensearch.protobufs.HitsMeta
     private static void processTotalHits(SearchHits hits, org.opensearch.protobufs.HitsMetadata.Builder hitsMetaData) {
         org.opensearch.protobufs.HitsMetadataTotal.Builder totalBuilder = org.opensearch.protobufs.HitsMetadataTotal.newBuilder();
 
+        if (hits.getTotalHits() == null) {
+            hitsMetaData.setTotal(totalBuilder.build());
+            return;
+        }
+
         // TODO need to pass parameters
         // boolean totalHitAsInt = params.paramAsBoolean(RestSearchAction.TOTAL_HITS_AS_INT_PARAM, false);
```
karenyrx, line 78: Currently params are not supported on the gRPC side. Is this the real root cause of the problem, instead of the fix proposed in this PR?

PR #20776, opened 2026-03-04
[2.19] Force resolution to plexus-utils:3.5.1

andrross:
The newer versions of plexus-utils are not compatible with Java 11.

### Check List

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -111,6 +111,12 @@ dependencies {
   api "commons-io:commons-io:${props.getProperty('commonsio')}"
   api "net.java.dev.jna:jna:5.14.0"
   api 'com.gradleup.shadow:shadow-gradle-plugin:8.3.10'
+  constraints {
+    api('org.codehaus.plexus:plexus-utils') {
+      version { strictly '3.5.1' }
+      because 'plexus-utils 4.x is incompatible with Java 11'
+    }
+  }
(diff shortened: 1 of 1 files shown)


buildSrc/build.gradle
```diff
@@ -111,6 +111,16 @@ dependencies {
   api "commons-io:commons-io:${props.getProperty('commonsio')}"
   api "net.java.dev.jna:jna:5.14.0"
   api 'com.gradleup.shadow:shadow-gradle-plugin:8.3.10'
+  constraints {
```
AWSHurneyt, line 114: @andrross I tested this change locally using the notifications plugin, but found that I still needed to force the use of `org.codehaus.plexus:plexus-utils:3.5.1` in the plugin build.gradle as well.

I tried using the below changes instead to force consumers of the buildSrc to use the 3.5.1 dependency as well, and that seems to have resolved the notifications build failure for me locally.
```
  api 'com.gradleup.shadow:shadow-gradle-plugin:8.3.10'
  api('org.codehaus.plexus:plexus-utils') {
    version {
      strictly '3.5.1'
    }
    because 'plexus-utils 4.x is incompatible with Java 11'
  }
  api 'org.jdom:jdom2:2.0.6.1'
```

andrross: @AWSHurneyt I tested the current change locally at it seems to work. Adding the `strictly` tag to the constraint seems to be the key.

PR #20780, opened 2026-03-04
Add AGENTS.md for AI coding assistants

andrross:
Initial commit of guidelines specifically for AI agents. I wrote a basic outline by hand, then asked an AI agent to complete the rest by reading DEVELOPER_GUIDE.md and TESTING.md.

There's obviously duplication here, but AGENTS.md has only factual information with no "why" that is useful to humans but takes up the context window for agents. We can evolve this going forward, but for now the duplication seems acceptible to me.

See https://agents.md/

### Check List

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,108 @@
+# AGENTS.md
+
+## Repository Structure
+
+Key components:
+
+- `server` - Core of the OpenSearch server. Includes the distributed system framework as well as most of the search and indexing functionality.
+- `plugins/*` - Optional plugins that extend interfaces defined in `server`. Plugins run in their own classloader that is a child of the `server` classloader.
+- `modules/*` - Architecturally the same as plugins, but are included by default and are not uninstallable.
+- `libs/*` - Libraries that can be used by `server` or any plugin or module.
+- `buildSrc` - The build framework, used by this repository and all external plugin repositories. This is published as the `build-tools` artifact.
+- `sandbox` - Contains `libs`, `modules`, and `plugins` that are under development. These are not included in non-snapshot builds.
+- `distribution` - Builds tar, zip, rpm, and deb packages.
+- `qa` - Integration tests requiring multiple modules/plugins, multi-version cluster tests, and tests in unusual configurations.
+- `test` - Test framework and test fixtures used across the project. Published for external plugin testing.
+
+## Build
+
+JDK 21 is the minimum supported. `JAVA_HOME` must be set.
+
+```
+./gradlew assemble          # build all distributions
+./gradlew localDistro       # build for local platform only
+./gradlew run               # run OpenSearch from source
+./gradlew generateProto     # regenerate protobuf code (if compilation errors)
+```
+
+## Testing
+
+### Unit Testing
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Writing Good Tests
+
+- Prefer unit tests over integration tests. Use `OpenSearchSingleNodeTestCase` only when components are hard to instantiate in isolation.
+- Use randomization for parameters not expected to affect behavior (e.g., shard count for aggregation tests), not for coverage. If different code paths exist (e.g., 1 shard vs 2+ shards), write separate tests for each.
+- Never use `Thread.sleep` directly. Use `assertBusy` or `waitUntil` for polling, or instrument code with concurrency primitives like `CountDownLatch` for deterministic waiting.
+- Do not depend on specific segment topology. If needed, disable background refreshes and force merge after indexing.
+- Clean up all resources at the end of each test. The base test class checks for open file handles and running threads after tear down.
+- Do not abuse randomization in multi-threaded tests — it makes failures non-reproducible.
+- When testing functionality behind a feature flag, use `FeatureFlags.TestUtils` and the `@LockFeatureFlag` annotation.
```
owaiskazi19, line 72: Can we also add to repeat the test execution to avoid flakiness caused by the written test?
```
./gradlew test -Dtests.iters=N -Dtests.class=*.ClassName
```

AGENTS.md
```diff
@@ -0,0 +1,104 @@
...
+./gradlew server:test --tests "*.ReplicaShardBatchAllocatorTests.testNoAsyncFetchData"
+./gradlew :server:internalClusterTest --tests "org.opensearch.action.admin.ClientTimeoutIT.testNodesInfoTimeout"
+
+# run with a specific seed for reproducibility
+./gradlew test -Dtests.seed=DEADBEEF
+```
+
+### Writing Good Tests
+
+- Prefer unit tests over integration tests. Use `OpenSearchSingleNodeTestCase` only when components are hard to instantiate in isolation.
```
owaiskazi19: > Prefer unit tests over integration tests

Do we want to have a preference for the tests?

andrross: I think so. Our track record over the last several years shows that we're not particularly good at writing non-flaky integration tests :) The anti-pattern is that you don't write software in a way that is amenable to unit testing and then you do all functional testing in integration tests, which are slower and more error-prone than single-threaded unit tests. What's your specific concern here? I could see the callout to `OpenSearchSingleNodeTestCase` as unnecessarily specific and might lead to agents avoiding that one specific case.

andrross: How about the following?

> Prefer unit tests over multi-threaded integration tests when unit tests can provide the same test coverage

owaiskazi19: If we want to restrict agent to avoid writing ITs initially, it's fine. Overall, we should let agent add UTs, ITs, Rest test case based on the feature.
Agreed, `OpenSearchSingleNodeTestCase` can remove such specific

owaiskazi19: > Prefer unit tests over multi-threaded integration tests when unit tests can provide the same test coverage

Looks good

PR #20769, opened 2026-03-03
[Backport 2.19] Harden the circuit breaker and failure handle logic in query result consumer

jainankitk:
### Description
Backports https://github.com/opensearch-project/OpenSearch/pull/19396 for preventing estimated circuit breaker limits from becoming negative. This issue was reported on one of the cluster for Amazon Opensearch Service

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #20782, opened 2026-03-05
Clamp negative byte size value stats to prevent API failures

peteralfonsi:
### Description
In production we have seen issues where negative values occur unexpectedly/incorrectly and break serialization in various _nodes APIs, causing the whole API to fail. Clamping these values prevents such incorrect stats from bringing down the API. 

### Check List
- [N/A] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [N/A] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.


libs/core/src/main/java/org/opensearch/core/common/unit/ByteSizeValue.java
```diff
@@ -59,7 +59,7 @@ public class ByteSizeValue implements Writeable, Comparable<ByteSizeValue>, ToXC
     private final ByteSizeUnit unit;
 
     public ByteSizeValue(StreamInput in) throws IOException {
-        size = in.readZLong();
+        size = Math.max(-1, in.readZLong());
```
andrross, line 62: This seems like a significant behavior change to a class that is widely used in this repo (and probably many other repos as well):

```
$ git grep ByteSizeValue | wc -l
    1518
```

peteralfonsi:
@jainankitk I'm happy to reduce the scope of the PR to only what's needed, but that would be clamping ByteSizeValue at its ctor, not at serializing time. The errors originally seen were "Values less than -1 bytes are not supported: -11220b", which comes from the [constructor](https://github.com/opensearch-project/OpenSearch/blob/521cef17225014e717e16a414aa50b5753fe043b/libs/core/src/main/java/org/opensearch/core/common/unit/ByteSizeValue.java#L78). During CircuitBreakerStats serialization, it instantiates a new ByteSizeValue with whatever value is on the local CircuitBreakerStats, and if that number is negative we get this error from the ByteSizeValue ctor.

Just reverting the first 2 of the commits should achieve that, I'll do that now

libs/core/src/main/java/org/opensearch/core/common/unit/ByteSizeValue.java
```diff
@@ -75,7 +75,7 @@ public ByteSizeValue(long bytes) {
 
     public ByteSizeValue(long size, ByteSizeUnit unit) {
         if (size < -1 || (size == -1 && unit != ByteSizeUnit.BYTES)) {
-            throw new IllegalArgumentException("Values less than -1 bytes are not supported: " + size + unit.getSuffix());
+            size = 0;
```
kkhatua, line 78: Can we add this as a warning, preferably with a counter ?

jainankitk:
> The errors originally seen were "Values less than -1 bytes are not supported: -11220b", which comes from the [constructor](https://github.com/opensearch-project/OpenSearch/blob/521cef17225014e717e16a414aa50b5753fe043b/libs/core/src/main/java/org/opensearch/core/common/unit/ByteSizeValue.java#L78). During CircuitBreakerStats serialization, it instantiates a new ByteSizeValue with whatever value is on the local CircuitBreakerStats, and if that number is negative we get this error from the ByteSizeValue ctor.

@peteralfonsi - I am wondering if that issue is already addressed by https://github.com/opensearch-project/OpenSearch/pull/20769 that ensures `CircuitBreakerStats` are not negative? Then, we don't really need this fix right?

peteralfonsi:
@jainankitk Yeah, like I mentioned to Kunal, this fix seems not necessary if we have already fixed CircuitBreakerStats behavior such that it never *tries* to create a negative-valued ByteSizeValue.

jainankitk:
> this fix seems not necessary if we have already fixed CircuitBreakerStats behavior such that it never tries to create a negative-valued ByteSizeValue.

Thanks @peteralfonsi. We should close this PR in that case!

Issue #9464, opened 2023-08-21
[BUG] QueryPhaseTests.testQueryTimeoutChecker is flaky


mch2:
https://build.ci.opensearch.org/job/gradle-check/22250/
https://github.com/opensearch-project/OpenSearch/issues/9204

jed326:
Moving into concurrent search project for now, will take this up in a few days.

jed326:
```
dengjay@88665a3f24d4 OpenSearch %  ./gradlew ':server:test' --tests "org.opensearch.search.query.QueryPhaseTests" -Dtests.seed=13B8EC24EAB06BB4                                         

> Configure project :
========================= WARNING =========================
         Backwards compatibility tests are disabled!
See https://github.com/opensearch-project/OpenSearch/issues/4173
===========================================================
=======================================
OpenSearch Build Hamster says Hello!
  Gradle Version        : 8.3
  OS Info               : Mac OS X 13.5.1 (x86_64)
  JDK Version           : 17 (Amazon Corretto JDK)
  JAVA_HOME             : /Library/Java/JavaVirtualMachines/amazon-corretto-17.jdk/Contents/Home
  Random Testing Seed   : 13B8EC24EAB06BB4
  In FIPS 140 mode      : false
=======================================

BUILD SUCCESSFUL in 7s
52 actionable tasks: 1 from cache, 51 up-to-date
```

Having trouble recreating this failure locally, probably means the flakiness is not related to the random test seed so will need to take a closer look at the test code.

jed326:
It looks like the parts of QueryPhase that are tested here are common to both concurrent and non-concurrent path. Moreover, it looks like this flakiness hasn't been reported in a few weeks, so going to deprioritize this one for now.

jed326:
This test is not flaky just for concurrent search:
- https://build.ci.opensearch.org/job/gradle-check/28561/
- https://build.ci.opensearch.org/job/gradle-check/30447/

The method being tested here is `QueryPhase::createQueryTimeoutChecker` which does not have any implementation differences between concurrent vs non-concurrent search paths

PR #20788, opened 2026-03-05
Make Telemetry Tags Immutable

sakrah:
## Problem

The `Tags` class in `libs/telemetry` is the dimension container for every metric emission in OpenSearch. On high-throughput clusters, tag construction and lookup become a significant CPU overhead on the metrics hot path -- before any actual metric work (atomic increment, gauge read) occurs.

The root cause is structural:

- **Mutable `HashMap` backend** -- every `Tags.create()` allocates a new `HashMap`. Every metric call that passes tags pays this cost, even when the exact same tag combination fires millions of times per second.
- **Unsafe as map keys** -- because `Tags` is mutable and has no `hashCode()`/`equals()` overrides, metric caches require defensive copies on every lookup, and every consumer must build its own caching layer.
- **No precomputed hash** -- every map lookup involving Tags must iterate all entries to compute a hash.

This means the existing API structurally prevents zero-allocation metric hot paths. Caching-only approaches reduce overhead but cannot eliminate it -- the cache key itself is an allocation, and every new metric consumer rediscovers the same problem.

## Solution

Replace the mutable `HashMap`-backed `Tags` with an **immutable sorted-array implementation** with a precomputed hash, inspired by [Micrometer's `Tags`](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/Tags.java) (sorted-array merge) and [OpenTelemetry's `ArrayBackedAttributes`](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/common/ArrayBackedAttributes.java) (parallel arrays, precomputed hash, last-wins dedup).

### Internal storage

```java
public final class Tags {
    private final String[] keys;      // sorted by key
    private final Object[] values;    // parallel to keys
    private final int hashCode;       // precomputed at construction -- O(1) return
}
```

### New API

| Method | Description |
|---|---|
| `Tags.of(key, value)` | 1-tag factory, no sort needed |
| `Tags.of(k1, v1, k2, v2)` | 2-tag factory, conditional swap |
| `Tags.of(k1, v1, k2, v2, k3, v3)` | 3-tag factory |
| `Tags.of(k1, v1, ..., k4, v4)` | 4-tag factory |
| `Tags.of(String... keyValues)` | N-pair varargs factory |
| `Tags.concat(a, b)` | Merge-sort two Tags; `b` wins on collision |
| `Tags.fromMap(Map)` | Bridge from existing map-based callers |
| `size()`, `getKey(i)`, `getValue(i)` | Direct array access |
| `equals()`, `hashCode()` | Content-based, safe as map keys |

### Backward compatibility

`Tags.create()` and all `addTag()` overloads are **deprecated, not removed**. `create()` returns `Tags.EMPTY`. Each `addTag()` returns a new immutable instance via `Tags.concat(this, Tags.of(key, value))`. Existing fluent chains like `Tags.create().addTag("a", "1").addTag("b", "2")` compile and produce correct results -- they just allocate more than the equivalent `Tags.of("a", "1", "b", "2")`.

### Why this matters for downstream consumers

With immutable Tags and a precomputed hash, metric consumers can:
- **Store Tags in fields** and reuse them across calls without re-allocating
- **Use Tags directly as map keys** without defensive copies
- **Share Tags across threads** without synchronization
- **Merge tag sets efficiently** via `concat()` on pre-sorted arrays

## Test plan

- 40 test cases in new `TagsTests.java` covering:
  - All factory methods (`of` overloads, `fromMap`, varargs)
  - Sorting invariants (keys always sorted regardless of input order)
  - Deduplication semantics (last value wins on duplicate keys)
  - `concat()` -- partial overlap, full overlap, interleaved, large remainder, hash consistency, null/empty inputs
  - `equals()`/`hashCode()` contract
  - `getTagsMap()` (String conversion, type preservation)
  - Backward compatibility (`create()`, `addTag()` chaining, all value types)
  - Map-key usability (content-based lookup across different instances)
- `./gradlew :libs:opensearch-telemetry:precommit` passes

### Related Issues

### Check List
- [ x] Functionality includes testing.

### DIFF
--- a/libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
+++ b/libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
@@ -45,100 +45,56 @@ public final class Tags {
     // -----------------------------------------------------------------------
 
     /**
-     * Creates an immutable Tags with one key-value pair.
+     * Creates an immutable Tags with one String-valued pair.
      * @param key   tag key
      * @param value tag value
      * @return new Tags instance
      */
-    public static Tags of(String key, Object value) {
+    public static Tags of(String key, String value) {
         Objects.requireNonNull(key, "key must not be null");
         Objects.requireNonNull(value, "value must not be null");
-        validateValue(value);
         String[] k = { key };
         Object[] v = { value };
         return new Tags(k, v, computeHash(k, v));
     }
 
     /**
-     * Creates an immutable Tags with two key-value pairs.
-     * @param k1 first key
-     * @param v1 first value
-     * @param k2 second key
-     * @param v2 second value
+     * Creates an immutable Tags with one long-valued pair.
+     * @param key   tag key
+     * @param value tag value
      * @return new Tags instance
      */
-    public static Tags of(String k1, Object v1, String k2, Object v2) {
-        Objects.requireNonNull(k1, "k1 must not be null");
-        Objects.requireNonNull(v1, "v1 must not be null");
-        Objects.requireNonNull(k2, "k2 must not be null");
-        Objects.requireNonNull(v2, "v2 must not be null");
-        validateValue(v1);
-        validateValue(v2);
-        int cmp = k1.compareTo(k2);
-        String[] keys;
-        Object[] values;
-        if (cmp < 0) {
-            keys = new String[] { k1, k2 };
-            values = new Object[] { v1, v2 };
-        } else if (cmp > 0) {
-            keys = new String[] { k2, k1 };
-            values = new Object[] { v2, v1 };
-        } else {
-            keys = new String[] { k2 };
-            values = new Object[] { v2 };
-        }
-        return new Tags(keys, values, computeHash(keys, values));
+    public static Tags of(String key, long value) {
+        Objects.requireNonNull(key, "key must not be null");
+        String[] k = { key };
+        Object[] v = { value };
+        return new Tags(k, v, computeHash(k, v));
     }
 
     /**
-     * Creates an immutable Tags with three key-value pairs.
-     * @param k1 first key
-     * @param v1 first value
-     * @param k2 second key
-     * @param v2 second value
-     * @param k3 third key
-     * @param v3 third value
+     * Creates an immutable Tags with one double-valued pair.
+     * @param key   tag key
+     * @param value tag value
      * @return new Tags instance
      */
-    public static Tags of(String k1, Object v1, String k2, Object v2, String k3, Object v3) {
-        Objects.requireNonNull(k1, "k1 must not be null");
-        Objects.requireNonNull(v1, "v1 must not be null");
-        Objects.requireNonNull(k2, "k2 must not be null");
-        Objects.requireNonNull(v2, "v2 must not be null");
-        Objects.requireNonNull(k3, "k3 must not be null");
-        Objects.requireNonNull(v3, "v3 must not be null");
-        validateValue(v1);
-        validateValue(v2);
-        validateValue(v3);
-        return fromPairs(new String[] { k1, k2, k3 }, new Object[] { v1, v2, v3 }, 3);
+    public static Tags of(String key, double value) {
+        Objects.requireNonNull(key, "key must not be null");
+        String[] k = { key };
+        Object[] v = { value };
+        return new Tags(k, v, computeHash(k, v));
     }
 
     /**
-     * Creates an immutable Tags with four key-value pairs.
-     * @param k1 first key
-     * @param v1 first value
-     * @param k2 second key
-     * @param v2 second value
-     * @param k3 third key
-     * @param v3 third value
-     * @param k4 fourth key
-     * @param v4 fourth value
+     * Creates an immutable Tags with one boolean-valued pair.
+     * @param key   tag key
+     * @param value tag value
      * @return new Tags instance
      */
-    public static Tags of(String k1, Object v1, String k2, Object v2, String k3, Object v3, String k4, Object v4) {
-        Objects.requireNonNull(k1, "k1 must not be null");
-        Objects.requireNonNull(v1, "v1 must not be null");
-        Objects.requireNonNull(k2, "k2 must not be null");
-        Objects.requireNonNull(v2, "v2 must not be null");
-        Objects.requireNonNull(k3, "k3 must not be null");
-        Objects.requireNonNull(v3, "v3 must not be null");
-        Objects.requireNonNull(k4, "k4 must not be null");
-        Objects.requireNonNull(v4, "v4 must not be null");
-        validateValue(v1);
-        validateValue(v2);
-        validateValue(v3);
-        validateValue(v4);
-        return fromPairs(new String[] { k1, k2, k3, k4 }, new Object[] { v1, v2, v3, v4 }, 4);
+    public static Tags of(String key, boolean value) {
+        Objects.requireNonNull(key, "key must not be null");
+        String[] k = { key };
+        Object[] v = { value };
+        return new Tags(k, v, computeHash(k, v));
     }
 
     /**
@@ -225,7 +181,6 @@ public final class Tags {
         Object[] values = new Object[keys.length];
         for (int i = 0; i < keys.length; i++) {
             values[i] = Objects.requireNonNull(map.get(keys[i]), "value for key '" + keys[i] + "' must not be null");
-            validateValue(values[i]);
         }
         return new Tags(keys, values, computeHash(keys, values));
     }
@@ -319,7 +274,7 @@ public final class Tags {
     public Tags addTag(String key, String value) {
         Objects.requireNonNull(key, "key cannot be null");
         Objects.requireNonNull(value, "value cannot be null");
-        return Tags.concat(this, Tags.of(key, (Object) value));
+        return Tags.concat(this, Tags.of(key, value));
     }
 
     /**
@@ -330,7 +285,7 @@ public final class Tags {
      */
     public Tags addTag(String key, long value) {
         Objects.requireNonNull(key, "key cannot be null");
-        return Tags.concat(this, Tags.of(key, (Object) value));
+        return Tags.concat(this, Tags.of(key, value));
     }
 
     /**
@@ -341,7 +296,7 @@ public final class Tags {
      */
     public Tags addTag(String key, double value) {
         Objects.requireNonNull(key, "key cannot be null");
-        return Tags.concat(this, Tags.of(key, (Object) value));
+        return Tags.concat(this, Tags.of(key, value));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
```diff
@@ -10,90 +10,412 @@
...
+        Objects.requireNonNull(v4, "v4 must not be null");
+        return fromPairs(new String[] { k1, k2, k3, k4 }, new Object[] { v1, v2, v3, v4 }, 4);
+    }
+
+    /**
+     * Creates Tags from interleaved key-value pairs; must be even length.
+     * @param keyValues alternating keys and values
+     * @return new Tags instance
+     */
+    public static Tags of(String... keyValues) {
```
msfroh: I'm inclined to agree with the AI reviewer on this method, though I guess my main beef is that it *only* accepts `String` values, when the other `of` overloads accept `Object` values.

I agree with the AI reviewer's suggestion of calling this `ofStrings` or maybe even `ofStringPairs`. I still think the case where the tag values are `String` is common enough that a dedicated varargs method for it makes sense.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
```diff
@@ -10,90 +10,412 @@
...
+        Object[] values = (sortedIndex == mergedValues.length) ? mergedValues : Arrays.copyOf(mergedValues, sortedIndex);
+        return new Tags(keys, values, computeHash(keys, values));
+    }
+
+    /**
+     * Creates Tags from a map.
+     * @param map key-value pairs
+     * @return new Tags instance
+     */
+    public static Tags fromMap(Map<String, String> map) {
```
msfroh: Couldn't the value type be `Object` here?

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
```diff
@@ -10,90 +10,412 @@
...
+     * @return value
+     */
+    public Object getValue(int i) {
+        return values[i];
+    }
+
+    /**
+     * Converts to a String-valued map. Intended for flush-time, not the hot path.
+     */
+    public Map<String, String> toMap() {
```
msfroh: Similar to `fromMap`, do we need to convert the values to `String`? If the consumer needs strings, that feels like its responsibility.

By that token, I don't feel like this should be the responsibility of `Tags` -- the `getTagsMap` method below is sufficient.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
```diff
@@ -10,90 +10,412 @@
...
+
+    /** Returns an unmodifiable map preserving original value types. */
+    public Map<String, ?> getTagsMap() {
+        if (keys.length == 0) return Collections.emptyMap();
+        Map<String, Object> map = new HashMap<>(keys.length);
+        for (int i = 0; i < keys.length; i++) {
+            map.put(keys[i], values[i]);
+        }
+        return Collections.unmodifiableMap(map);
+    }
```
msfroh: This could be:

```
Map.Entry<String, ?>[] entries = new Map.Entry<String, ?>[keys.length];
for (int i = 0; i < keys.length; i++) {
  entries[i] = Map.entry(keys[i], values[i]);
}
return Map.ofEntries(entries);
```

It still creates copies of the data, but it's at least more array-focused. It also preserves the sorted order of keys.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
```diff
@@ -10,90 +10,412 @@
...
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof Tags that)) return false;
+        if (this.hashCode != that.hashCode) return false;
+        if (this.keys.length != that.keys.length) return false;
+        for (int i = 0; i < keys.length; i++) {
+            if (!keys[i].equals(that.keys[i])) return false;
+            if (!values[i].equals(that.values[i])) return false;
+        }
```
msfroh: You can just use `return Arrays.equals(keys, that.keys) && Arrays.equals(values, that.values)`.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
```diff
@@ -10,90 +10,412 @@
...
+        return fromPairs(keys, values, count);
+    }
+
+    /**
+     * Merges two Tags. On key collision, {@code b} wins. Either argument may be null.
+     * @param a first tags
+     * @param b second tags
+     * @return merged Tags instance
+     */
+    public static Tags concat(Tags a, Tags b) {
```
msfroh, line 126: Something that occurs to me is that (in practice), a lot of metrics are tagged with a bunch of repeated "stuff". In fact, if that wasn't true, metric cardinality would be unbounded and whoever is responsible for making your metrics searchable would hate you.

Broadly, I think of metric tags falling into three categories that get concatenated together:
1. Node-level tags. (Node id, host name, zone, region, etc.)
2. Operation-level tags (what operation is being measured)
3. Target-level tags (what is the target of the operation -- an index, a shard, etc.)

Essentially, each of these sets of tags could be long-lived (with the third category tied to the lifetime of the target).

I'm wondering if it would make sense in the long run to implement a `Tags concat(Tags... allTags)` method that holds onto references to the underlying `Tags` objects (instead of copying them). It would cut down on the number and size of array allocations, at least.

I don't think it makes sense to do that right now, but it's worth considering as we take metrics more seriously in OpenSearch.

sakrah: Thanks for this feedback. Will keep this in mind

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
sakrah: Will do, thanks

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
sakrah: Thanks for this. will update

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
sakrah: Agreed. will remove it

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
sakrah: Agreed. changes in the new update

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
sakrah: Done

reta:
Thanks for the change, @sakrah , I would be inclined to not replicate the existing implementations, but:
 - convert Tags to interface (API is experimental, we do have head room here)
 - use service loader to bootstrap backing implementation (since it is going to be otel for now, `Attributes`)

The reasoning here is to use the tags implementation already available by the metrics providers, the `libs/opensearch-telemetry` by itself should be mostly about API, not implementation. Hope it make sense, thank you.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
```diff
@@ -10,90 +10,397 @@
...
     /**
      * Add String attribute.
-     * @param key key
+     * @param key   key
      * @param value value
-     * @return Same instance.
+     * @return new Tags instance with the added tag
+     * @deprecated Use {@link #of} or {@link #concat} instead.
      */
+    @Deprecated
```
reta: No need for deprecation I think, this is experimental API that by definition could be changed pretty much freely.

sakrah:
gradle-check failure is unrelated to this PR. Failing tests are known flaky:

- `QueryPhaseTests.testQueryTimeoutChecker` → #9464, #14551
- `CompositeIndexWriterForAppendTests.classMethod` → #20005
- `transport-netty4:internalClusterTest` → infrastructure (zombie Docker container on Jenkins node)

Retrying CI.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
sakrah: Agreed, since the API is experimental I'll remove the @Deprecated annotations.

sakrah: On converting Tags to an interface with SPI -- I see it is an extablished pattern in the codebase  and agree it's a sound direction. The sorted-array implementation here would become the default provider.
I'd prefer to land this as-is and follow up with the interface + SPI extraction in a separate PR. The immutability fix is the urgent part; the interface extraction is orthogonal and would be easier to review separately. 

Happy to open a follow-up issue for it right after this lands.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
```diff
@@ -10,90 +10,383 @@
...
 /**
- * Class to create tags for a meter.
+ * Immutable tags for a meter.
  *
  * @opensearch.experimental
  */
 @ExperimentalApi
-public class Tags {
-    private final Map<String, Object> tagsMap;
+public final class Tags {
```
reta, line 23: General note on the redesign of this API, we initially targeted the following data types (for tag values):
 - string
 - double
 - boolean
 - long

The current implementation generalizes over `Object` and accepts anything, this is not going to work with Otel (or even Micrometer / ... ). I believe we should validate that the type of the value is supported (one from the above list), that could be extended later if needed, thanks.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
reta: I am fine keeping Tags, but we need to harden the implementation a bit (https://github.com/opensearch-project/OpenSearch/pull/20788#discussion_r2901956561), Otel would have done it for us.

libs/telemetry/src/main/java/org/opensearch/telemetry/metrics/tags/Tags.java
msfroh: Ahh... thanks @reta! I wasn't aware of that specific constraint.

@sakrah -- can you rework the constructors and/or the `add` methods to impose those constraints?

The `ofStringPairs` should still be fine, and is likely the most useful "create all the tags in one shot" factory method.

sakrah: Will do, thanks

reta:
> @reta -- It looks like @sakrah made the type validation a runtime thing w/ `instanceof` checks. What are your thoughts on that?

Sorry a bit late, but I think runtime is fine, thank you @msfroh and @sakrah

msfroh:
> > @reta -- It looks like @sakrah made the type validation a runtime thing w/ `instanceof` checks. What are your thoughts on that?
> 
> Sorry a bit late, but I think runtime is fine, thank you @msfroh and @sakrah

Surprise! The checks moved back to compile-time. 😁

PR #20791, opened 2026-03-06
Adding lockable concurrent queue utils to be used for managing composite writers

shank9918:
### Description
- Adding lockable concurrent queue utilities to be used in CompositeEngine for managing pool of composite writers. 

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/common/pool/ConcurrentQueue.java
+++ b/server/src/main/java/org/opensearch/common/pool/ConcurrentQueue.java
@@ -1,0 +1,173 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.common.pool;
+
+import java.util.Iterator;
+import java.util.Queue;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
+ * A thread-safe queue implementation that distributes entries across multiple internal queues
+ * to reduce lock contention.
+ * <p>
+ * This queue uses a striping pattern where entries are distributed across a configurable number
+ * of internal queues, each protected by its own lock. This design allows multiple threads to
+ * operate on the queue concurrently with minimal contention.
+ * </p>
+ * <p>
+ * Thread affinity is achieved by using the current thread's hash code to determine which
+ * internal queue to access first, improving cache locality and reducing contention.
+ * </p>
+ *
+ * @param <T> the type of elements held in this queue
+ * @opensearch.internal
+ */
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/opensearch/common/pool/AlreadyClosedException.java
```diff
@@ -0,0 +1,19 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.common.pool;
+
+public class AlreadyClosedException extends IllegalStateException {
```
mgodwan: Do we need a separate class here? IllegalStateException should be ok to return

server/src/main/java/org/opensearch/common/pool/ConcurrentQueue.java
```diff
@@ -0,0 +1,173 @@
...
+ * </p>
+ * <p>
+ * Thread affinity is achieved by using the current thread's hash code to determine which
+ * internal queue to access first, improving cache locality and reducing contention.
+ * </p>
+ *
+ * @param <T> the type of elements held in this queue
+ * @opensearch.internal
+ */
+public final class ConcurrentQueue<T> {
```
mgodwan: I see all methods are package protected. Can we follow the same for this class?

server/src/main/java/org/opensearch/common/pool/ConcurrentQueue.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+
+    /**
+     * Adds an entry to the queue.
+     *
+     * @param entry the entry to add to the queue
+     */
+    void add(T entry) {
+        final int threadHash = Thread.currentThread().hashCode() & 0xFFFF;
```
mgodwan, line 79: Do we have any benchmarks with thread affinity and without it?

server/src/main/java/org/opensearch/common/pool/ConcurrentQueue.java
```diff
@@ -0,0 +1,173 @@
...
+        return null;
+    }
+
+    /**
+     * Removes the specified entry from the queue.
+     *
+     * @param entry the entry to remove
+     * @return {@code true} if the entry was found and removed, {@code false} otherwise
+     */
+    boolean remove(T entry) {
```
mgodwan, line 158: How often is the remove called? Would it help to have an auxiliary data structure to avoid going through all queues?

server/src/main/java/org/opensearch/common/pool/AlreadyClosedException.java
shank9918: Right sounds good, will update.

server/src/main/java/org/opensearch/common/pool/ConcurrentQueue.java
shank9918: Yes will update.

Issue #1580, opened 2021-11-18
Metricbeat Connection issue with Opensearch - Version mismatch


choudharydivyanshu777:
I am trying to ship Metrics using Metricbeat to opensearch in an openshift platform  but facing issue in connection of metricbeat with opensearch.
I deployed metricbeat using following manifest:
````---
apiVersion: v1
kind: ConfigMap
metadata:
  name: metricbeat-daemonset-config
  namespace: keycloak
  labels:
    k8s-app: metricbeat
data:
  metricbeat.yml: |-
    metricbeat.config.modules:
      # Mounted `metricbeat-daemonset-modules` configmap:
      path: ${path.config}/modules.d/*.yml
      # Reload module configs as they change:
      reload.enabled: false
      monitoring.enabled: false
      

    metricbeat.autodiscover:
      providers:
        - type: kubernetes
          scope: cluster
          node: ${NODE_NAME}
          unique: true
          templates:
            - config:
                - module: kubernetes
                  hosts: ["kube-state-metrics:8080"]
                  period: 10s
                  add_metadata: true
                  metricsets:
                    - state_node
                    - state_deployment
                    - state_daemonset
                    - state_replicaset
                    - state_pod
                    - state_container
                    - state_job
                    - state_cronjob
                    - state_resourcequota
                    - state_statefulset
                    - state_service
                - module: kubernetes
                  metricsets:
                    - apiserver
                  hosts: ["https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}"]
                  bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
                  ssl.certificate_authorities:
                    - /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
                  period: 30s
                # Uncomment this to get k8s events:
                #- module: kubernetes
                #  metricsets:
                #    - event
        # To enable hints based autodiscover uncomment this:
        #- type: kubernetes
        #  node: ${NODE_NAME}
        #  hints.enabled: true

    es.version: 7.10.2  
    monitoring.enabled: false  

    processors:
      - add_cloud_metadata:

    cloud.id: ${ELASTIC_CLOUD_ID}
    cloud.auth: ${ELASTIC_CLOUD_AUTH}

    output.elasticsearch:
      hosts: ['${ELASTICSEARCH_HOST:elasticsearch}:${ELASTICSEARCH_PORT:9200}']
[...]

PR #20796, opened 2026-03-08
Support top_metrics aggregation requests

### LINKED ISSUE #15025: [BUG] Top Metrics Aggregation Exists in Java Client But Not Supported
### What is the bug?
Top Metrics Aggregation Exists in the Java Client, but is not supported by OpenSearch

### How can one reproduce the bug?
Create an aggregation query with top metrics aggregation

### What is the expected behavior?
Preferably support for this aggregation would be added and work the same as elasticsearch implementation. https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-top-metrics.html

### Do you have any screenshots?
none

### Do you have any additional context?
Opensearch-java version: 2.11 
Opensearch cluster version: 2.11

Error: [parsing_exception] Unknown aggregation type [top_metrics] did you mean [top_hits]?

related: https://github.com/opensearch-project/opensearch-java/issues/420

### Workaround
Use top_hits aggregation for now

mrutunjay-kinagi:
## Summary
Add server-side support for `top_metrics` aggregation so requests no longer fail with `Unknown aggregation type [top_metrics]`.

This PR adds:
- `top_metrics` aggregation builder/parser and registration
- aggregator factory + execution path
- internal aggregation result rendering (`top` with `sort` and `metrics`)
- tests for parser registration and builder behavior

## Testing
- `./gradlew :server:test --tests "org.opensearch.search.aggregations.metrics.TopMetricsTests" --tests "org.opensearch.search.aggregations.AggregatorFactoriesTests.testTopMetricsTypeParses"`
- `./gradlew :server:precommit`

## Related
- Fixes #15025


mrutunjay-kinagi:
Follow-up: CI  is green and DCO passes. The remaining blocker is  (currently ~39% patch coverage). I am preparing additional aggregation execution tests to raise coverage for  and  paths.

mrutunjay-kinagi:
Follow-up: CI `gradle-check` is green and DCO passes. The remaining blocker is `codecov/patch` (currently ~39% patch coverage). I am preparing additional aggregation execution tests to raise coverage for `TopMetricsAggregator` and `InternalTopMetrics` paths.

server/src/main/java/org/opensearch/search/aggregations/metrics/TopMetricsAggregator.java
```diff
@@ -0,0 +1,74 @@
...
+        SubSearchContext subSearchContext,
+        String name,
+        List<String> metricFields,
+        SearchContext context,
+        Aggregator parent,
+        Map<String, Object> metadata
+    ) throws IOException {
+        super(name, context, parent, metadata);
+        this.metricFields = List.copyOf(metricFields);
+        this.delegate = new TopHitsAggregator(fetchPhase, subSearchContext, name, context, parent, metadata);
```
jainankitk: The delegate `TopHitsAggregator` is constructed with the same `parent` and `name` as the wrapper. Since both `TopMetricsAggregator` (via `MetricsAggregator` → `Aggregator`) and the delegate register themselves as children of `parent`, this could cause double-registration in the aggregation tree. This may surface as issues with profiling, `doClose()` being called unexpectedly, or incorrect tree traversal.

server/src/main/java/org/opensearch/search/aggregations/metrics/TopMetricsAggregatorFactory.java
```diff
@@ -0,0 +1,94 @@
...
+        subSearchContext.from(0);
+        subSearchContext.size(size);
+        subSearchContext.fetchSourceContext(FetchSourceContext.DO_NOT_FETCH_SOURCE);
+        subSearchContext.explain(false);
+        subSearchContext.version(false);
+        subSearchContext.seqNoAndPrimaryTerm(false);
+        subSearchContext.trackScores(false);
+        if (sort.isPresent()) {
+            subSearchContext.sort(sort.get());
+        }
```
jainankitk, line 82: If `SortBuilder.buildSort()` returns `Optional.empty()` (which can happen when all sort fields are unmapped), the aggregation silently proceeds without any sort applied. Since `top_metrics` is semantically meaningless without a sort, this should throw rather than silently produce non-deterministic results:

```java
if (sort.isPresent()) {
   subSearchContext.sort(sort.get());
} else {
   throw new IllegalStateException("[top_metrics] requires a valid sort but none could be resolved");
}
```

server/src/main/java/org/opensearch/search/aggregations/metrics/TopMetricsAggregatorFactory.java
```diff
@@ -0,0 +1,94 @@
...
+    ) throws IOException {
+        SubSearchContext subSearchContext = new SubSearchContext(searchContext);
+        subSearchContext.parsedQuery(searchContext.parsedQuery());
+        subSearchContext.from(0);
+        subSearchContext.size(size);
+        subSearchContext.fetchSourceContext(FetchSourceContext.DO_NOT_FETCH_SOURCE);
+        subSearchContext.explain(false);
+        subSearchContext.version(false);
+        subSearchContext.seqNoAndPrimaryTerm(false);
+        subSearchContext.trackScores(false);
```
jainankitk: `trackScores` is hardcoded to `false`, but if a user specifies `_score` as the sort field, scores won't be computed and the sort will be meaningless. Should we check if any of the sort builders reference `_score` and enabling score tracking accordingly — similar to how `TopHitsAggregationBuilder` handles this?

server/src/main/java/org/opensearch/search/aggregations/metrics/InternalTopMetrics.java
```diff
@@ -0,0 +1,129 @@
...
+
+            builder.startObject(METRICS_FIELD);
+            for (String metricField : metricFields) {
+                DocumentField field = hit.field(metricField);
+                if (field == null || field.getValues().isEmpty()) {
+                    builder.nullField(metricField);
+                } else if (field.getValues().size() == 1) {
+                    builder.field(metricField);
+                    Object metricValue = field.getValues().get(0);
+                    builder.value(metricValue);
```
jainankitk, line 110: Minor: `builder.field(metricField)` followed by `builder.value(metricValue)` can be simplified to `builder.field(metricField, metricValue)`.

jainankitk (commented):
Thanks @mrutunjay-kinagi for raising this PR. The current tests cover parsing and builder serialization, but there are no integration tests that actually execute the aggregation against indexed documents.

Would be good to add an `InternalTopMetricsTests` extending `InternalAggregationTestCase` and a `TopMetricsAggregatorTests` that indexes documents and verifies end-to-end behavior.

mrutunjay-kinagi:
> Thanks @mrutunjay-kinagi for raising this PR. The current tests cover parsing and builder serialization, but there are no integration tests that actually execute the aggregation against indexed documents.
> 
> Would be good to add an `InternalTopMetricsTests` extending `InternalAggregationTestCase` and a `TopMetricsAggregatorTests` that indexes documents and verifies end-to-end behavior.

I can work on this further. Thanks for your input.

mrutunjay-kinagi:
@jainankitk Thanks for the review. I pushed follow-up changes in commit `2aed977c8af` to address the comments and add stronger end-to-end coverage.

Updated items:
- `TopMetricsAggregator`: delegate `TopHitsAggregator` now uses `this` as parent to avoid sibling registration/lifecycle issues.
- `TopMetricsAggregatorFactory` + `TopMetricsAggregationBuilder`:
  - fail fast when sort cannot be resolved,
  - handle `_score` sort fallback correctly,
  - enable `trackScores` when required.
- `InternalTopMetrics`: added defensive reduce validation and small cleanup in xcontent writing.

Test coverage updates:
- Added `InternalTopMetricsTests` (extends `InternalAggregationTestCase`).
- Added `TopMetricsAggregatorTests` with indexed-doc execution assertions.
- Added new internal-cluster E2E test: `TopMetricsIT` (executes `top_metrics` against indexed docs for field sort and score sort).

Verified locally with:
- `./gradlew :server:test --tests "org.opensearch.search.aggregations.metrics.TopMetricsAggregatorTests" --tests "org.opensearch.search.aggregations.metrics.InternalTopMetricsTests" --tests "org.opensearch.search.aggregations.metrics.TopMetricsTests" --tests "org.opensearch.search.aggregations.AggregatorFactoriesTests.testTopMetricsTypeParses"`
- `./gradlew :server:internalClusterTest --tests "org.opensearch.search.aggregations.metrics.TopMetricsIT"`

mrutunjay-kinagi:
Addressed the gradle-check failure from build #73773. Root cause was :server:spotlessJavaCheck formatting violations in top_metrics files.

Pushed formatting-only fix in commit ecbb26ea4d9 to feature/top-metrics-agg-support.

Local verification: ./gradlew :server:spotlessJavaCheck passes.

mrutunjay-kinagi:
Addressed the  failure from build #73773. Root cause was  formatting violations in top_metrics files.
Pushed formatting-only fix in commit  to .
Local verification: > Task :buildSrc:reaper:compileJava UP-TO-DATE
> Task :buildSrc:reaper:processResources NO-SOURCE
> Task :buildSrc:reaper:classes UP-TO-DATE
> Task :buildSrc:reaper:jar UP-TO-DATE
> Task :buildSrc:compileJava UP-TO-DATE
> Task :buildSrc:compileGroovy UP-TO-DATE
> Task :buildSrc:generateVersionProperties UP-TO-DATE
> Task :buildSrc:pluginDescriptors UP-TO-DATE
> Task :buildSrc:processResources UP-TO-DATE
> Task :buildSrc:classes UP-TO-DATE
> Task :buildSrc:jar UP-TO-DATE
=======================================
OpenSearch Build Hamster says Hello!
  Gradle Version        : 9.4.0
  OS Info               : Mac OS X 26.3.1 (aarch64)
  JDK Version           : 23 (OpenJDK JDK 23 (23.0.1+11-39))
  JAVA_HOME             : /Users/mj/Library/Java/JavaVirtualMachines/openjdk-23.0.1/Contents/Home
  Random Testing Seed   : 70F7FA3109361E6A
  Crypto Standard       : any-supported
=======================================
> Task :server:spotlessJava UP-TO-DATE
> Task :server:spotlessJavaCheck UP-TO-DATE

BUILD SUCCESSFUL in 40s
10 actionable tasks: 10 up-to-date passes.

mrutunjay-kinagi:
@jainankitk  I've updated the PR. What do you think ?

mrutunjay-kinagi:
@jainankitk Its ready for review.

vishdivs:
Hi @mrutunjay-kinagi — thanks for this! We need top_metrics in the index-management plugin (https://github.com/opensearch-project/index-management) for rollup first_with_time/last_with_time metrics (#1580 (https://github.com/opensearch-project/index-management/issues/1580)), where all sub-aggregations run inside CompositeAggregationBuilder.
  
**Concern**: The current implementation delegates to TopHitsAggregator with SubSearchContext, but composite uses deferred collection (doc values only) and doesn't support per-bucket search contexts, the same reason top_hits throws "not supported as a sub-aggregation of composite" today.

While the TopHitsAggregator approach works well for `_score` sort etc., I wanted to suggest a dual-path factory where we can support both composite compatibility and the full-featured TopHits path:
  
```
  @Override
  protected Aggregator createInternal(SearchContext ctx, Aggregator
  parent, ...) {
      if (isInsideComposite(parent)) {
          // Lightweight: per-bucket BigArrays + doc values (like 
  MinAggregator)
          return new TopMetricsDocValuesAggregator(name, sortField,
  metricFields, sortOrder, ...);
      } else {
          // Full-featured: SubSearchContext (supports _score sort 
  etc.)
          return new TopMetricsTopHitsAggregator(name,
  subSearchContext, ...);
      }
  }
```
Happy to contribute the doc-values path if this direction makes sense.

@jainankitk, would love to know your thoughts on the above approach?

mrutunjay-kinagi:
> Hi @mrutunjay-kinagi — thanks for this! We need top_metrics in the index-management plugin (https://github.com/opensearch-project/index-management) for rollup first_with_time/last_with_time metrics (#1580 ([opensearch-project/index-management#1580](https://github.com/opensearch-project/index-management/issues/1580))), where all sub-aggregations run inside CompositeAggregationBuilder.
> 
> **Concern**: The current implementation delegates to TopHitsAggregator with SubSearchContext, but composite uses deferred collection (doc values only) and doesn't support per-bucket search contexts, the same reason top_hits throws "not supported as a sub-aggregation of composite" today.
> 
> While the TopHitsAggregator approach works well for `_score` sort etc., I wanted to suggest a dual-path factory where we can support both composite compatibility and the full-featured TopHits path:
> 
> ```
>   @Override
>   protected Aggregator createInternal(SearchContext ctx, Aggregator
>   parent, ...) {
>       if (isInsideComposite(parent)) {
>           // Lightweight: per-bucket BigArrays + doc values (like 
>   MinAggregator)
>           return new TopMetricsDocValuesAggregator(name, sortField,
>   metricFields, sortOrder, ...);
>       } else {
>           // Full-featured: SubSearchContext (supports _score sort 
>   etc.)
>           return new TopMetricsTopHitsAggregator(name,
>   subSearchContext, ...);
>       }
>   }
> ```
> 
> Happy to contribute the doc-values path if this direction makes sense.
> 
> @jainankitk, would love to know your thoughts on the above approach?

@vishdivs Happy to collaborate and take this to closure. 
/cc @jainankitk

PR #20798, opened 2026-03-08
Add sorting support

Shailesh-Kumar-Singh:
### Description
Adds sort support for the Composite Engine. During indexing, segments are sorted by a configurable column (derived from index sort settings) before being written to Parquet. During merge, a new streaming k-way merge implementation sorts across multiple input files using a min-heap with pipelined parallel encoding using rayon and IO writes using Tokio.


server/src/main/java/org/opensearch/index/LogByteSizeMergePolicyProvider.java
```diff
@@ -57,7 +57,7 @@ public class LogByteSizeMergePolicyProvider implements MergePolicyProvider {
     public static final ByteSizeValue DEFAULT_MIN_MERGE = new ByteSizeValue(16, ByteSizeUnit.MB);
     public static final int DEFAULT_MERGE_FACTOR = 10;
 
-    public static final ByteSizeValue DEFAULT_MAX_MERGED_SEGMENT = new ByteSizeValue(5, ByteSizeUnit.GB);
+    public static final ByteSizeValue DEFAULT_MAX_MERGED_SEGMENT = new ByteSizeValue(512, ByteSizeUnit.MB);
```
bharath-techie, line 60: Is this for both parquet + text combined ? how does this get applied for Lucene + parquet scenarios

Shailesh-Kumar-Singh: We check during the eligible merge selection whether the combined size of the Lucene and Parquet segments is below the configured maximum merged segment size.
Once support for both Lucene and Parquet is enabled, the plan is to increase this limit.

modules/parquet-data-format/benchmarks/src/main/java/com/parquet/parquetdataformat/benchmark/ParquetWriterWriteBenchmark.java
```diff
@@ -39,7 +39,7 @@ public void setup() throws IOException {
         writerCreationBenchmarkData = generator.generate("simple", fieldCount, 0);
         writerWriteBenchmarkData = generator.generate("simple", fieldCount, recordCount);
         filePath = generateTempFilePath();
-        RustBridge.createWriter(filePath, "benchmark-index", writerCreationBenchmarkData.getArrowSchema().memoryAddress());
+        RustBridge.createWriter(filePath, writerCreationBenchmarkData.getArrowSchema().memoryAddress(), "sortColumn", false);
```
mgodwan: is null not allowed for the column name?

server/src/main/java/org/opensearch/index/engine/exec/MergeInput.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.engine.exec;
```
mgodwan: move this to merge pkg

server/src/main/java/org/opensearch/index/engine/exec/MergeInput.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.List;
+
+public class MergeInput {
+    private List<WriterFileSet> fileMetadataList;
+    private long writerGeneration;
+    private String sortingField;
+    private boolean reverseSort;
+
```
mgodwan: mark as final

modules/parquet-data-format/src/main/java/com/parquet/parquetdataformat/vsr/VSRManager.java
```diff
@@ -38,17 +38,17 @@ public class VSRManager implements AutoCloseable {
...
     private final String fileName;
-    private final String indexName;
     private final VSRPool vsrPool;
+    private final String sortColumn;
+    private final boolean reverseSort;
     private NativeParquetWriter writer;
 
 
-    public VSRManager(String fileName, String indexName, Schema schema, ArrowBufferPool arrowBufferPool) {
+    public VSRManager(String fileName, String indexName, Schema schema, ArrowBufferPool arrowBufferPool, String sortColumn, boolean reverseSort) {
```
mgodwan, line 49: Fine for now, but can we just pass indexsettings here?

modules/parquet-data-format/src/main/rust/src/lib.rs
```diff
@@ -4,62 +4,84 @@ use arrow::ffi::{FFI_ArrowArray, FFI_ArrowSchema};
...
 pub use native_settings::NativeSettings;
 pub use field_config::FieldConfig;
 pub use writer_properties_builder::WriterPropertiesBuilder;
-
 // Re-export macros from the shared crate for logging
 pub use vectorized_exec_spi::{log_info, log_error, log_debug};
 
+/// Per-writer sort configuration stored at create time, consumed at close time.
+struct SortConfig {
+    sort_column: Option<String>,
```
mgodwan, line 36: while we are supporting only a single column today, this may evolve to use multiple columns. For main, lets ensure interface can take care of that

Shailesh-Kumar-Singh:
> Can you add profiler as well which was used for microbenchmarking, along with steps to generate the data?

Should we have profiler code in PR?
I have not included the profiler code in the PR, as it was used solely for testing and benchmarking purposes.
Please let me know if you would prefer that it be added.

modules/parquet-data-format/src/main/rust/src/lib.rs
Shailesh-Kumar-Singh: Yes. I am already working on to support multi column sorting - will push in a separate PR.

PR #20805, opened 2026-03-09
Scale test suite timeout with test iterations

### LINKED ISSUE #5630: Update gradle test setup to consider test iterations for suite timeout
Coming from https://github.com/opensearch-project/OpenSearch/issues/1561#issuecomment-1364454897, existing gradle test runs does not consider the iteration count and uses existing 20 mins as the [test suite timeout](https://github.com/opensearch-project/OpenSearch/blob/main/test/framework/src/main/java/org/opensearch/test/OpenSearchTestCase.java#L184). This is problematic as it often leads to test suite timeout and user need to manually update timeout to a higher value.

zheliu2:
## Summary
- Fixes #5630
- When running tests with `-Dtests.iters=N`, the suite timeout (default 20 minutes) now scales by multiplying the base timeout by the number of iterations
- The fix is in the Gradle test setup plugin (`OpenSearchTestBasePlugin`), using the `tests.timeoutSuite` system property that randomizedtesting supports for overriding `@TimeoutSuite` annotations
- Preserves existing Windows behavior (30 min base timeout) and also scales it when iterations are specified
- If the user explicitly sets `-Dtests.timeoutSuite`, it is respected and no automatic scaling is applied

## Test plan
- [ ] Run `./gradlew test -Dtests.iters=10 -Dtests.class=*.SomeTest` and verify the suite does not time out prematurely
- [ ] Verify that without `-Dtests.iters`, behavior is unchanged (20 min default from annotation)
- [ ] Verify that explicit `-Dtests.timeoutSuite` still overrides the automatic scaling

### DIFF
--- a/buildSrc/src/main/java/org/opensearch/gradle/OpenSearchTestBasePlugin.java
+++ b/buildSrc/src/main/java/org/opensearch/gradle/OpenSearchTestBasePlugin.java
@@ -200,9 +200,26 @@ public class OpenSearchTestBasePlugin implements Plugin<Project> {
                 logging.setShowStandardStreams(Util.getBooleanProperty("tests.output", false));
             });
 
-            if (OS.current().equals(OS.WINDOWS) && System.getProperty("tests.timeoutSuite") == null) {
-                // override the suite timeout to 30 mins for windows, because it has the most inefficient filesystem known to man
-                test.systemProperty("tests.timeoutSuite", "1800000!");
+            if (System.getProperty("tests.timeoutSuite") == null) {
+                // Default suite timeout is 20 minutes (1200000ms) as set by @TimeoutSuite annotation.
+                long baseTimeout = 20L * 60 * 1000;
+
+                if (OS.current().equals(OS.WINDOWS)) {
+                    // Override the suite timeout to 30 mins for windows, because it has the most
+                    // inefficient filesystem known to man.
+                    baseTimeout = 30L * 60 * 1000;
+                }
+
+                // Scale suite timeout based on test iterations to prevent false timeouts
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/org/opensearch/gradle/OpenSearchTestBasePlugin.java
```diff
@@ -200,9 +200,26 @@ public void execute(Task t) {
                 logging.setShowStandardStreams(Util.getBooleanProperty("tests.output", false));
             });
 
-            if (OS.current().equals(OS.WINDOWS) && System.getProperty("tests.timeoutSuite") == null) {
-                // override the suite timeout to 30 mins for windows, because it has the most inefficient filesystem known to man
-                test.systemProperty("tests.timeoutSuite", "1800000!");
```
andrross, line 205: What does the `!` do here? I hope it's not 1800000 factorial...

zheliu2: Thanks for the review! Just deleted my previous response since it might be confusing, `!` suffix is already used on [the existing line this PR refactored](https://github.com/opensearch-project/OpenSearch/blob/main/buildSrc/src/main/java/org/opensearch/gradle/OpenSearchTestBasePlugin.java#L205) (`"1800000!"`), and is documented in [TESTING.md#L190-L192](https://github.com/opensearch-project/OpenSearch/blob/main/TESTING.md#L190-L192):

> Change the default suite timeout to 5 seconds for all tests (note the exclamation mark).
> `./gradlew test -Dtests.timeoutSuite=5000! ...`

It tells the randomized testing framework to override the `@TimeoutSuite` annotation value. Happy to add an inline comment if that would help clarify.

PR #20807, opened 2026-03-09
Fix DiskThresholdMonitor auto-releasing user-set read blocks

### LINKED ISSUE #20539: [BUG] `DiskThresholdMonitor` unconditionally auto-releases user-set `index.blocks.read` setting

zheliu2:
## Summary
- Fixes #20539: `DiskThresholdMonitor.handleReadBlocks` unconditionally auto-released `index.blocks.read` on all indices not currently needing a read block, ignoring whether the node was still above a high disk watermark.
- The fix passes `indicesNotToAutoRelease` into `handleReadBlocks` and filters those indices from the auto-release candidates, consistent with how `handleReadOnlyBlocks` already handles `index.blocks.read_only_allow_delete`.
- Added test `testReadBlockNotAutoReleasedWhenNodeAboveHighWatermark` to verify indices on nodes above the high watermark are not auto-released.

## Test plan
- [x] New unit test `testReadBlockNotAutoReleasedWhenNodeAboveHighWatermark` passes
- [x] All existing `DiskThresholdMonitorTests` pass


zheliu2:
Closing — CI gradle-check is consistently failing due to Jenkins infrastructure issues. Will reopen or recreate if needed.

PR #20808, opened 2026-03-09
Fix empty _source includes array returning all fields

### LINKED ISSUE #20612: [Feature Request] [BUG] `_source` include / exclude validation

Example:
```json
GET /people-index/_search
{
  "_source" : {
    "excludes": ["name"],
    "includes": ["name", "age"]
  },
  "query" : {
    // my wonderful query
  },
  "size" : 20,
  "sort" : [ {
  } ]
}
```

## Empty arrays behavior

Example:
```json
GET /people-index/_search
{
  "_source": {
    "includes": [ "" ]
  }, 
}
```
Results in following hits:
```
"hits": [
      {
        "_index": "people-index",
        "_id": "111111",
        "_score": 1.7178972,
        "_source": {}
      },
      {
        "_index": "people-index",
        "_id": "222222",
        "_score": 1.7136133,
        "_source": {}
      },
```

### Related component

Search:Query Insights

_No response_

### Additional context

_No response_

zheliu2:
## Summary

Fixes #20612

- When `_source` includes is set to an empty array (e.g., `"_source": []` or `"_source": {"includes": []}`), the expected behavior is to return an empty `_source: {}`. Previously, an empty includes array was treated the same as not setting includes at all, causing all source fields to be returned.
- Added an `includesExplicit` flag to `FetchSourceContext` to distinguish between "no includes specified" (include everything) and "empty includes explicitly specified" (include nothing).
- The flag is set during XContent parsing (`fromXContent`) and REST request parsing (`parseFromRestRequest`), and is propagated through wire serialization (version-gated to V_3_6_0) for cross-node compatibility.

## Changes

- **`FetchSourceContext.java`**: Added `includesExplicit` boolean field, private 4-arg constructor, version-gated serialization, and filter logic that returns an empty map when includes is explicitly empty.
- **`FetchSourcePhase.java`**: Updated `containsFilters()` to recognize explicitly empty includes as a filter.
- **`FetchSourcePhaseTests.java`**: Added `testEmptyIncludesArray` and `testEmptyIncludesWithExcludes` test cases.

## Test plan

- [x] `FetchSourcePhaseTests.testEmptyIncludesArray` - verifies `_source: []` and `_source: {"includes": []}` both return empty source
- [x] `FetchSourcePhaseTests.testEmptyIncludesWithExcludes` - verifies `_source: {"includes": [], "excludes": ["field1"]}` returns empty source
- [x] All existing `FetchSourcePhaseTests` continue to pass


zheliu2:
Closing — CI gradle-check is consistently failing due to Jenkins infrastructure issues. Will reopen or recreate if needed.

urmichm:
Hi @zheliu2 
This is a duplicate of an existing issue with a prepared PR

PR #20814, opened 2026-03-09
Fix field_caps returning empty results for disable_objects mappings

### LINKED ISSUE #20811: [BUG] `_field_caps` returns empty response for indexes with `disable_objects: true` mappings
### Describe the bug

3.5.0 added a great new feature:

- https://github.com/opensearch-project/OpenSearch/pull/19958

However, I noticed that using such fields completely breaks Opensearch Dashboards index patterns. This seems to be because `_field_caps` returns an empty response:

```
{
  "indices": [],
  "fields": {}
}
```

### Related component

Indexing

### To Reproduce

1. Create an index with a disable_objects field

```
curl -X PUT "localhost:9200/test" -H "Content-Type: application/json" -d '{
  "mappings": {
    "properties": {
      "attributes": { "type": "object", "disable_objects": true }
    }
  }
}'
```

2. Index a document with a nested value under the disable_objects field

```
curl -X POST "localhost:9200/test/_doc/1" -H "Content-Type: application/json" -d '{
  "attributes": { "foo.bar": "value" }
}'
```

3. Call _field_caps API

```
curl "localhost:9200/test/_field_caps?fields=*"
```
 
It returns:

```
{
  "indices": [],
  "fields": {}
}
```

### Expected behavior

The `_field_caps` API should return the `attributes.foo.bar` field amongst other internal fields

### Additional Details

**Plugins**
N/A

**Screenshots**
N/A

 - OS: Ubuntu 24.04
 - Opensearch 3.5.0

**Additional context**
N/A

newtonne:
### Description

`_field_caps` returns an empty response for indexes where a `disable_objects: true` object field has been populated with a child field. A second bug causes such child fields to be corrupted after any subsequent document is indexed into the same index - for example `attributes.foo.bar` becomes `attributes.foo.foo.bar`.

#### Root causes

When `_field_caps` walks the parent chain of a flattened leaf field (e.g. `attributes.foo.bar`), it looks up an `ObjectMapper` for each intermediate path. Under `disable_objects: true`, intermediate paths like `attributes.foo` have no `ObjectMapper` by design - the fix adds a null check to skip them and continue up the chain.

The field name corruption is caused by `ParametrizedFieldMapper.merge()` using `name().lastIndexOf('.')` to reconstruct the parent `ContentPath` when rebuilding a mapper. For a field with `simpleName` `foo.bar` and full name `attributes.foo.bar`, this returns the position of the dot before `bar` rather than the dot before `foo.bar`, so the parent path is computed as `attributes.foo` instead of `attributes`. The fix computes the boundary from `simpleName.length()` instead.

### Related Issues
Resolves #20811

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fix terms lookup subquery fetch limit reading from non-existent index setting instead of cluster `max_clause_count` ([#20823](https://github.com/opensearch-project/OpenSearch/pull/20823))
 - Fix array_index_out_of_bounds_exception with wildcard and aggregations ([#20842](https://github.com/opensearch-project/OpenSearch/pull/20842))
 - Handle dependencies between analyzers ([#19248](https://github.com/opensearch-project/OpenSearch/pull/19248))
+- Fix `_field_caps` returning empty results and corrupted field names for `disable_objects: true` mappings ([#20800](https://github.com/opensearch-project/OpenSearch/pull/20800))
 
 ### Dependencies
 - Bump shadow-gradle-plugin from 8.3.9 to 9.3.1 ([#20569](https://github.com/opensearch-project/OpenSearch/pull/20569))
--- a/rest-api-spec/src/main/resources/rest-api-spec/test/field_caps/40_disable_objects.yml
+++ b/rest-api-spec/src/main/resources/rest-api-spec/test/field_caps/40_disable_objects.yml
@@ -1,0 +1,139 @@
+---
+"Field caps with disable_objects returns leaf fields":
+
+  - skip:
+      version: " - 3.4.99"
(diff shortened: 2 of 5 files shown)


gaobinlong (commented):
The DCO check failed, please amend the commit message with your sign-off, add also add some change log for this PR. In addition, this bug fix is better to have some e2e yaml rest tests, see https://github.com/opensearch-project/OpenSearch/blob/main/TESTING.md#testing-the-rest-layer, could you add it, thanks! @newtonne

newtonne:
Thanks @gaobinlong. Sure, I can take a look at that. Would the rest tests be in addition to the current tests?

gaobinlong:
> Thanks @gaobinlong. Sure, I can take a look at that. Would the rest tests be in addition to the current tests?

Yeah, It's better to have yaml rest tests for API changes.

andrross (approved):

Issue #20397, opened 2026-01-09
[Feature Request] Dynamic fields / dynamic properties


msfroh:
### Is your feature request related to a problem? Please describe

OpenSearch currently supports "dynamic mappings". That is, you can include fields in a document not covered by the current index mapping and we'll go ahead and modify the mapping. 

The way I like to think of it is kind of like if a relational database were sold with a pitch of "If a user tries to INSERT a row into a table that references a column that doesn't exist, we'll pause that INSERT, issue an ALTER TABLE ADD COLUMN operation, then go ahead and process the INSERT." (Obviously, any relational database vendor pitching that would get laughed out of the room, but for some reason people love it on OpenSearch and similar systems, because NoSQL or something.)

One problem is that these ALTER TABLE-like operations are pretty heavy. In the case of OpenSearch, we need to stop processing the index request on the primary shard, send a mapping update to the cluster manager, then wait for the new cluster state to be published. Also, if you add enough fields, the cluster state can get really big. While some cluster state updates can be sent as deltas, I'm pretty sure we only send full mappings. So, these can be large updates once the mappings get really big.

### Describe the solution you'd like

As an alternative, I'd like to pitch something akin to Solr [dynamic fields](https://solr.apache.org/guide/solr/latest/indexing-guide/dynamic-fields.html). Since we currently have `properties` in mappings, I guess for consistency we should call them `dynamic_properties`.

Unlike dynamic mappings, a dynamic field says "Fields whose name match this pattern are mapped like *this*". That is, one entry in the mapping covers any field whose name matches the pattern. I like to think of it as "Static mapping, but dynamic fields". We add more Lucene fields (which are pretty cheap these days), but avoid adding to the OpenSearch-level index mapping.

Essentially, in an index's mappings, alongside the `properties` entry, I would add a `dynamic_properties` entry that looks more or less the same (with the same field type parameters), but where each dynamic property's name is expected to contain a wildcard character.

From a validation standpoint, I would add the following rules (not exhaustive):
1. You cannot add a dynamic field whose pattern would yield an ambiguous match. (So, e.g. you couldn't have both `*_i` and `i_*` if you want to use the Solr convention of using `*_i` for `int` fields.)
2. You cannot add a regular field to the mapping if the field name is matched by a dynamic field, since there may be existing data for a Lucene field with that name covered by the dynamic field.
3. Rules that apply to regular fields: can't change type, can't remove a field, etc.

### Related component

Indexing

[...]

PR #20816, opened 2026-03-10
Add Dynamic Properties Support

aparajita31pandey:
<!--  Thanks for sending a pull request, here are some tips:

1. If this is a fix for an undisclosed security vulnerability, please STOP. All security vulnerability reporting and fixes should be done as per our security policy https://github.com/opensearch-project/OpenSearch/security/policy
2. If this is your first time, please read our contributor guidelines: https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md and developer guide https://github.com/opensearch-project/OpenSearch/blob/main/DEVELOPER_GUIDE.md
3. Ensure you have added or ran the appropriate tests for your PR: https://github.com/opensearch-project/OpenSearch/blob/main/TESTING.md
-->
### Description
Implements `dynamic_properties` on the root mapping: pattern-based field definitions ([Solr-style dynamic fields](https://solr.apache.org/guide/solr/latest/indexing-guide/dynamic-fields.html)) so matching field names are typed from the pattern without adding each name to the cluster state mapping, reducing cluster mapping-update for high-cardinality field names.

Behavior at mapping time :
  - dynamic_properties is an object at the root level, keys are glob patterns, values are field-mapping objects.                          
  - Patterns must contain * (bare * is rejected; use a prefix or suffix).                                                                 
  - Overlapping patterns -  two patterns that could match the same field name are rejected at mapping-update time.  Example  dynamic_properties patterns [*_ms] and [*ms] overlap and hence is rejected                                                                                                                       
  - Explicit mapped fields cannot overlap with any dynamic pattern.                                                                       
  
                                                                                                                                         
 #### Two guardrails prevent uncontrolled growth:
                                                                                                                                          
| Guard | Default | Setting |
| :--- | :---: | :--- |
| Max patterns per index | 50 | Hard-coded (MAX_DYNAMIC_PROPERTIES)   |
|  Lucene field count limit  | 10000 | index.mapping.dynamic_properties.lucene_field.limit |

                                           
### Related Issues
Resolves [#20397]
<!-- List any other related issues here -->

## Test

1. **Created index with dynamic_properties**

``` 
curl -s -X PUT "http://localhost:9200/dp_custom_test" \
  -H "Content-Type: application/json" -d '{
  "mappings": {
    "dynamic_properties": {
      "*_i": { "type": "long" },
      "*_s": { "type": "keyword" }
    },
    "properties": {}
  }
}'
```

**Response:**
> {"acknowledged":true,"shards_acknowledged":true,"index":"dp_custom_test"}
> HTTP 200

2. **Indexed a document**

```
curl -s -X POST "http://localhost:9200/dp_custom_test/_doc/1" \
  -H "Content-Type: application/json" -d '{
  "count_i": 42,
  "name_s": "foo"
}'
```

**Response:**

> {"_index":"dp_custom_test","_id":"1","_version":1,"result":"created","_shards":{"total":2,"successful":2,"failed":0},"_seq_no":0,"_primary_term":1}
> HTTP 201

3.  **Inspect mapping**
```
curl -s "http://localhost:9200/dp_custom_test/_mapping?pretty"
```
Response
```
{
  "dp_custom_test" : {
    "mappings" : {
      "dynamic_properties" : {
        "*_i" : {
          "type" : "long"
        },
        "*_s" : {
          "type" : "keyword"
        }
      }
    }
  }
}
```
Confirms no new mapping update

### Check List


msfroh:
@aparajita31pandey -- you'll need to address the issue from the "Detect Breaking Changes" action. Since `IndexModule` is (unfortunately) public for plugins, we have to assume that plugins could  be calling any `public` methods. Of course, a plugin should **never** call `newIndexService`, but hypothetically they could. So, to maintain backwards compatibility, we can't change the signature of `newIndexService`. Instead, we need to add another overloaded implementation, delegate from the old version, and mark the old version as deprecated. I'm a little shocked that this: https://github.com/opensearch-project/OpenSearch/blob/c6aa3266217059b800d1f191f66a11eba6b9ee49/server/src/main/java/org/opensearch/index/IndexModule.java#L680-L682 got merged, since it violates the spirit of the backwards-compatibility checks. It just guarantees that a plugin that hypothetically calls that (non-deprecated) method will get a `RuntimeException`. So, their compilation will silently work and things will blow up at runtime. IMO, that should **never** have been merged. (Tagging @ashking94, who merged it.) @aparajita31pandey, if you can fix that as part of this PR, we can correct that past mistake.

msfroh (commented):
Can we please add some field count limits as safeguards?

1. As called out by the AI review bot, the overlap detection is `O(n^2)` in the number of dynamic properties. As a result, we should probably put a cap on the number of dynamic properties. Maybe 50 to start? I'd prefer to start with a low limit, since it's easier to raise it than to lower it.
2. As discussed w/ @synhershko on #20397, while the additional Lucene fields are a lot cheaper than OpenSearch mappings, they're still not free (especially when doc values are enabled). I'm not sure how exactly to pull it off, but if we can get access to the current `FieldInfos` in dynamic properties to see if the given field is new (i.e. does `fieldInfos.fieldInfo(name)` return `null`), then we could check the current field count (i.e `fieldInfos.size()`) and fail the indexing operation if we cross some threshold. This one should probably have an associated index setting. This would be similar to `index.mapping.total_fields.limit` ([here](https://github.com/opensearch-project/OpenSearch/blob/83161ce8d515c5540c351b52fc1cd7ed7792abf6/server/src/main/java/org/opensearch/index/mapper/MapperService.java#L143)). Where that defaults to 1000 fields, maybe we can make the (per-shard) Lucene field limit something like 10k? I think a 10x increase in the default limit is appropriate given the lower cost of a Lucene field versus an OpenSearch mapping. We can still make it a dynamic setting, so people can increase it if they have the heap to spare.

aparajita31pandey:
## Gradle check build 74720 failure analysis

**TL;DR:** The failure is `BlockTransferManagerTests > classMethod` — a **flaky upstream test** introduced by #21178 (WritableWarm tiered storage), not caused by this PR.

### Evidence

| | Detail |
|---|---|
| **Test that failed** | `org.opensearch.storage.common.BlockTransferManagerTests.classMethod` |
| **Failure type** | Thread-leak detection by `@ThreadLeakFilters` — the randomized testing framework's cleanup hook, NOT a test assertion |
| **JUnit XML** | `failures="0" errors="0"` for all 9 test methods — every actual test case passes |
| **In isolation** | `./gradlew :server:test --tests "org.opensearch.storage.common.BlockTransferManagerTests"` → `BUILD SUCCESSFUL` (confirmed 3×) |
| **Introduced by** | Upstream commit `ee6052b649d` (#21178) — WritableWarm tiered storage. Not in this PR. |
| **Reproduces without our changes** | Yes — thread leak exists on base branch |

The test creates a full `ThreadPool` in `@Before` but doesn't completely drain all worker threads in `@After`. Under parallel full-suite execution the leftover threads trip the framework's `ThreadLeakControl$StatementRunner` check. Running the class alone gives its threads enough time to wind down naturally.

### Our PR test status (3 consecutive local full-suite runs)

- `IndexShardTests.testCacheWrapperReader` — ✅ PASS (this was the real failure in 74717/74719, fixed in commit `48b274d3059`)
- `DynamicMappingTests`, `RootObjectMapperTests`, `DocumentParserTests`, `FieldTypeLookupTests`, `IndexShardTests` — ✅ all PASS
- `BlockTransferManagerTests.classMethod` — ❌ flaky upstream thread-leak (unrelated to this PR)

Requesting a re-run of gradle-check.

msfroh (approved):
Wow! Thanks @aparajita31pandey -- this is really impressive work!

I'm excited to have this feature in OpenSearch. I think it will make it a lot easier for folks who legitimately need a lot of fields without taking the hit from blocking on the cluster manager (and blowing up cluster state with mapped fields).

PR #20817, opened 2026-03-10
Make IndexShard#getEngine method public for plugin access

venketep-prasad:
## Summary
- Makes `IndexShard#getEngine()` public (was package-private) so plugins can access the engine without resorting to reflection.
- `getEngine()` is preferred over `getEngineOrNull()` because it throws `AlreadyClosedException` when the engine is closed, giving callers a clear signal rather than a null.

## Context
The [time-series-db plugin](https://github.com/opensearch-project/time-series-db) has a requirement to load a block in a shard, which requires the engine access for the shard, `getEngine` on `IndexShard` (see [time-series-db#39](https://github.com/opensearch-project/time-series-db/pull/39#discussion_r2785705308)). As noted by @msfroh in that review:

> Doing this via reflection definitely shouldn't be the way to go... I think I like the idea of exposing `getEngine` better than `getEngineOrNull`. In particular, the former will throw an exception if the engine is closed, which I think probably makes sense.

This PR exposes `getEngine` as public so plugins can use it directly.

## Changes
- `server/src/main/java/org/opensearch/index/shard/IndexShard.java`: Changed `getEngine()` from package-private to `public` and added Javadoc.
- `CHANGELOG.md`: Added entry under `[Unreleased 3.x] > Changed`.

## Test plan
- [ ] Existing unit tests pass (no behavioral change — only visibility widened)
- [ ] Verify no new test failures in `IndexShardTests`

Signed-off-by: Venketep Prasad <venketep.prasad@uber.com>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/src/main/java/org/opensearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/opensearch/index/shard/IndexShard.java
@@ -4017,7 +4017,12 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
         recoveryState.getVerifyIndex().checkIndexTime(Math.max(0, TimeValue.nsecToMSec(System.nanoTime() - timeNS)));
     }
 
-    Engine getEngine() {
+    /**
+     * Returns the current engine of this shard.
+     *
+     * @throws AlreadyClosedException if the engine is closed
+     */
+    public Engine getEngine() {
         Engine engine = getEngineOrNull();
         if (engine == null) {
             throw new AlreadyClosedException("engine is closed");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Added support of WarmerRefreshListener in NRTReplicationEngine to trigger warmer after replication on replica shards ([#20650](https://github.com/opensearch-project/OpenSearch/pull/20650))
 
 ### Changed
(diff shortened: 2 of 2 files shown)


server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -4017,7 +4017,12 @@ private void doCheckIndex() throws IOException {
         recoveryState.getVerifyIndex().checkIndexTime(Math.max(0, TimeValue.nsecToMSec(System.nanoTime() - timeNS)));
     }
 
-    Engine getEngine() {
+    /**
+     * Returns the current engine of this shard.
+     *
+     * @throws AlreadyClosedException if the engine is closed
+     */
+    public Engine getEngine() {
```
mgodwan, line 4025: There is ongoing effort to decouple indexshard from engine, and adding this method to the public API may prevent us from fully achieving it.
Is there any specific use case with tsdb plugin you're looking to solve by exposing getEngine method as a public?

venketep-prasad: The use case is to load a block into a new shard ([PR](https://github.com/opensearch-project/time-series-db/pull/39#discussion_r2785705308)). Specifically, for these 2 functionalities:
1) Moving blocks to a glacier storage cluster
2) Download a block from prod and load it locally to test performance

It requires the engine access from the shard, as the engine has the index manager, which will be used to open the block (once the block is placed in the right path/directory)

PR #20821, opened 2026-03-10
Native engine abstractions

bharath-techie:
### Description
- This PR wires the Index shard with backend plugins via Reader managers.
- Reader managers in backend plugins is not refcounted , readers can only can be retrieved with a catalog snapshot and a catalog snapshot when it gets acquired / released gets ref counted, so we will fully rely on catalog snapshot for reader / files management
- Has changes for CatalogSnapshot, DataFormatRegistry , DataFormatAwareEngine, IndexFileDeleter etc which might be part of other indexing PRs such # 20675 - but added here to give clarity to flow
- Search exec engine is created per query which then creates a context that is associated with the query action which maintains the state throughout its lifecycle.

- WIP :
   - Data format class 
   - Naming conventions for index / source provider and interfaces - this will be consumed by delegates , what this PR intends to do is associate context and establish contract

### Related Issues

### Check List


server/src/main/java/org/opensearch/index/engine/CompositeEngine.java
```diff
@@ -0,0 +1,82 @@
...
+public class CompositeEngine implements Closeable {
+
+    private static final Logger logger = LogManager.getLogger(CompositeEngine.class);
+
+    private final Map<String, List<SearchExecEngine<?, ?>>> readEngines;
+
+    public CompositeEngine(List<SearchEnginePlugin> plugins, ShardPath shardPath) throws IOException {
+        Map<String, List<SearchExecEngine<?, ?>>> engines = new HashMap<>();
+        for (SearchEnginePlugin plugin : plugins) {
+            SearchExecEngine<?, ?> engine = plugin.createSearchExecEngine(shardPath);
```
bharath-techie: Reworking to avoid this in composite engine - we will do it in index shard.  Here we'll just tie SPIs of backend plugins that listens to deletes and refreshes.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsBackEndPlugin.java
```diff
@@ -10,18 +10,23 @@
...
+
+import java.io.IOException;
+import java.util.List;
 
 /**
  * SPI extension point for back-end query engines (DataFusion, Lucene, etc.).
  * @opensearch.internal
  */
-public interface AnalyticsBackEndPlugin {
+public interface AnalyticsBackEndPlugin extends SearchAnalyticsBackEndPlugin {
```
mch2: I think we need to keep these two things separate - the api in SearchAnalyticsBackEndPlugin isn't going to be used by the analytics engine and are largely used for registration/indexing with composite.

server/src/main/java/org/opensearch/index/engine/exec/SearchExecEngine.java
```diff
@@ -0,0 +1,64 @@
...
+        } catch (Exception e) {
+            listener.onFailure(e);
+        }
+    }
+
+    /**
+     * Create a search context. The reader is provided by {@link org.opensearch.index.engine.CompositeEngine}
+     * which owns all reader managers.
+     */
+    C createContext(
```
mch2: i don't know that we need this exposed - this was my motive for the bridge interface as is.  The bridge can manage the lifecycle of the engine specific context vs exposing it outside and is built from the given snapshot.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsBackEndPlugin.java
Bukhtawar: Maybe we could have the same class extend and implement?

server/src/main/java/org/opensearch/index/engine/exec/SearchExecEngine.java
bharath-techie: We need to have a common view of catalog snapshot or reader of the common view exposed.  So that + plan is what we always send to this method.

Mainly this enables filter delegates and in general all search actions to maintain their own contexts through the query lifecycle. 

I'm okay with not exposing this as well as long as we are able to provide the context when we initialize the delegates.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsBackEndPlugin.java
mch2: makes sense

server/src/main/java/org/opensearch/index/engine/CompositeEngine.java
```diff
@@ -0,0 +1,169 @@
...
+        this.sourceProviderSuppliers = sourceProviderSuppliers;
+    }
+
+    // ---- Public getters ----
+
+    public EngineReaderManager<?> getReaderManager(DataFormat format) {
+        return readerManagers.get(format);
+    }
+
+    public SearchExecEngine<?, ?> getSearchExecEngine(DataFormat format) throws IOException {
```
mch2: when we search from the analytics-engine, we're going to end up with a fragment that is assigned to a particular engine from query planning.  We'll need to fetch by engine name, not dataformat - probably just as simple as maintaining a reader by name map.

server/src/main/java/org/opensearch/index/engine/exec/SearchExecEngine.java
mch2: Here's roughly what i'm thinking - where snapshot below is replaced with the shared reader context.

```java
        try (CompositeEngine.ReleasableRef<CatalogSnapshot> snapshot = engine.acquireSnapshot()) {
            EngineBridge<byte[], ? extends EngineResultStream, RelNode> bridge =
                (EngineBridge<byte[], ? extends EngineResultStream, RelNode>) plugin.bridge(engine, snapshot.getRef());

            byte[] converted = bridge.convertFragment(logicalFragment);

            List<Object[]> rows = new ArrayList<>();
            try (EngineResultStream resultStream = bridge.execute(converted)) {
...
                    }
                }
            }
```
The bridge interface (poorly named) was intended to be the point in time searcher specific to the back-end, built from the shared snapshot/reader.  Which is basically this one, so it makes to replace it but thinking it should be the responsibility of AnalyticsBackEndPlugin to vend it on demand from the point in time state/reader managed by the composite engine. Sort of like CompositeEngine's `getSearchExecEngine`, but rather than requiring the CompositeEngine to maintain that its only worrying about refreshing readers behind the scenes and vending the composite reader.

server/src/main/java/org/opensearch/index/engine/CompositeEngine.java
mch2: but i think the back-ends can build the searcher / searchExecEngine on the fly, vs having to maintain this state.  We'd just need to fetch the reader to build searcher from?

sandbox/libs/analytics-framework/build.gradle
```diff
@@ -35,7 +36,7 @@ dependencies {
 
 testingConventions.enabled = false
 
-// analytics-framework does not depend on server
+// analytics-framework depends on server for SearchAnalyticsBackEndPlugin SPI
```
bharath-techie, line 39: Do we let the bridge talk to the engine ? if we need to remove this dependency. 

We can come back to this after we figure out if context is required in analytics plugin for delegates.

server/src/main/java/org/opensearch/index/engine/exec/DataFormatAwareEngineFactory.java
```diff
@@ -0,0 +1,108 @@
...
+        return new DataFormatAwareEngine(readerManagers, engineSuppliers, indexFilterProviderSuppliers, sourceProviderSuppliers);
+    }
+
+    /**
+     * Creates a {@link CatalogSnapshotLifecycleListener} that routes events
+     * through the {@link IndexFileDeleter} and fans out to the given reader managers.
+     *
+     * @param readerManagers the per-format reader managers that receive notifications
+     */
+    public CatalogSnapshotLifecycleListener createCatalogSnapshotListener(Map<DataFormat, EngineReaderManager<?>> readerManagers) {
```
mch2, line 67: where will this be called? from the composite engine?

bharath-techie:
https://github.com/bharath-techie/OpenSearch/pull/53 - Thanks @mch2 for helping with wiring the engine interfaces to analytics core plugin which I borrowed back to this PR. :)

server/src/main/java/org/opensearch/index/engine/exec/DataFormatAwareEngineFactory.java
bharath-techie: Yes from DataFormatAwareEngine that ties to the refresh / delete lifecycle.

libs/common/src/main/java/org/opensearch/common/CheckedTriFunction.java
```diff
@@ -8,11 +8,14 @@
...
 package org.opensearch.common;
 
+import org.opensearch.common.annotation.ExperimentalApi;
+
 /**
  * A {@link TriFunction}-like interface which allows throwing checked exceptions.
  *
  * @opensearch.internal
  */
+@ExperimentalApi
```
Bukhtawar: nit: why does this need to be experimental?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineResultStream.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @opensearch.internal
+ */
+public interface EngineResultStream extends AutoCloseable {
+
+    /**
+     * Returns an iterator over the record batches in this stream.
+     * Each call returns the same iterator instance — the stream is single-pass.
+     */
+    EngineResultBatchIterator iterator();
```
Bukhtawar: Simply use Iterator<EngineResultBatch>?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/ExecutionContext.java
```diff
@@ -0,0 +1,52 @@
...
+ * Execution context carrying reader and delegation state through
+ * the query execution lifecycle.
+ *
+ * @opensearch.internal
+ */
+public class ExecutionContext {
+
+    private final String tableName;
+    private final DataFormatAwareEngine.DataFormatAwareReader reader;
+    SearchShardTask task;
```
Bukhtawar: nit: private final

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/SearchExecEngine.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Executes the context and returns a result stream.
+     * @param context the execution context
+     */
+    EngineResultStream execute(ExecutionContext context) throws IOException;
+
+    @Override
+    default void close() throws IOException {}
+}
```
Bukhtawar:
```suggestion
public interface SearchExecEngine<T, V> extends Closeable {

    /**
     * Creates an execution context from a resolved plan.
     *
     * @param context               ExecutionContext
     */
    void prepare(T context);

    /**
     * Executes the context and returns a result stream.
     * @param context the execution context
     */
    V execute(T context) throws IOException;

    @Override
    default void close() throws IOException {}
}
```

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsSearchBackendPlugin.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Creates a searcher bound to the given reader snapshot.
+     * @param ctx the execution context
+     */
+    SearchExecEngine searcher(ExecutionContext ctx);
+
+    /** Returns the set of RelNode operator classes this backend supports. */
+    default Set<Class<? extends RelNode>> supportedOperators() {
+        return Set.of(LogicalTableScan.class, LogicalFilter.class, LogicalAggregate.class, LogicalProject.class);
+    }
```
Bukhtawar: Are we forcing a Calcite dependency on search backend plugins? I think we should avoid this

bharath-techie: Okay since this is not used in this PR, removing it for now. @mch2 just a heads up for your next PR.

libs/common/src/main/java/org/opensearch/common/CheckedTriFunction.java
bharath-techie: Hey this is added a while back as part of this https://github.com/opensearch-project/OpenSearch/pull/7983 but just getting consumed by a public api.
So experimental annotation i think is wrong.
But for public api , am wondering what's the practice, should we add a older 2.x version or the current version ?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/SearchExecEngine.java
bharath-techie: Today ExecutionContext and EngineResultStream is used as it in the plugins , but still keeping the generics if we need to extend tomorrow.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/EngineResultStream.java
bharath-techie: This iterator has custom methods on the record batches. And stream provides close method to cleanup. 
Do you mean something else ?

bharath-techie: Addressed this.

libs/common/src/main/java/org/opensearch/common/CheckedTriFunction.java
bharath-techie: Added 2.9.0 since it was introduced in 2.9.0.

server/src/main/java/org/opensearch/index/engine/CompositeEngine.java
bharath-techie: We discussed more on this where catalog snapshot is format based and one engine/plugin can have multiple formats. For now going with dataformat, we can simplify it we see a need by having engine/plugin name to format mapping.

server/src/main/java/org/opensearch/index/engine/exec/SearchExecEngine.java
bharath-techie: Yeah this is resolved thanks to the execution context and per query search exec engine from https://github.com/bharath-techie/OpenSearch/pull/53 :)

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AnalyticsBackEndPlugin.java
bharath-techie: This is separated out. Engine Plugin like DF Plugin implements both the interfaces now.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionService.java
```diff
@@ -0,0 +1,107 @@
...
+        // TODO: initialize Tokio runtime and memory pool via NativeBridge
+        // long ptr = NativeBridge.createGlobalRuntime(memoryPoolLimit, spillDirectory, spillMemoryLimit);
+        long ptr = 1L; // placeholder until NativeBridge is wired
+        this.runtimeHandle = new NativeRuntimeHandle(ptr);
+        logger.info("DataFusion service started");
+    }
+
+    @Override
+    protected void doStop() {
+        logger.info("Stopping DataFusion service");
```
Bukhtawar, line 68: nit: debug

bharath-techie: taking this up in next PR

PR #20822, opened 2026-03-10
5990: Fix security plugin issue

### LINKED ISSUE #5990: [BUG] OS 3.1 not showing node level traces for DBQ query
### Describe the bug

Setup: Created a local setup of Open Search 3.1 with jaegar to view traces. I am able to see http level traces but not detailed traces at node/shard level to debug slow DBQ queries. Configs used:

telemetry.feature.tracer.enabled - true
opensearch.experimental.feature.telemetry.enabled - true
telemetry.tracer.enabled - true
telemetry.tracer.sampler.probability - 1
telemetry.otel.tracer.span.exporter.class - io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporter

Plugin: Open Telemetry plugin

Attaching screenshot for the same

Image

### Related component

Plugins

### To Reproduce

1. Run OS version 3.1 in local and enable configs as mentioned above. Also start jaegar so that you are able to visualise traces.
2. Run DBQ query at URL: https://localhost:8443/test-index-dbq-tracing/_delete_by_query?wait_for_completion=true&requests_per_second=500&conflicts=proceed
3. Check jaegar UI with relevant trace id. We will not see scroll or node level traces. Just http level traces.

### Expected behavior

We should see traces at node and scroll level.

### Additional Details

**Plugins**
telemetry-otel

**Screenshots**

Os version 3.1

**Additional context**

Deepti24:
### Description
We are trying to fix: https://github.com/opensearch-project/security/issues/5990
We need to pass transients that should be copied. This change addresses that

### Related Issues
https://github.com/opensearch-project/security/issues/5990
Testing was done as follows:
* Open search was built with these changes using ./gradlew :distribution:archives:linux-arm64-tar:assemble -x test
* Security plugin was build such that is uses changes published by above ./gradlew assemble -Dopensearch.version=3.5.0-SNAPSHOT -Dbuild.snapshot=true -x test
* Then it was run and traces were validated. 

A couple of points in PR:
* A new method getTransients is added to ActionPlugin class. We could have reused headers but felt like it would be unclear which fields are transients to recover in security plugin within headers map. Also, responsibility wise both looked different. Do let me know if my understanding is incorrect
* Ideally OTelTelemetryPlugin should implements getTransients method but it does not extend ActionPlugin currently. Do let me know if modification should be made there appropriately
* Change to ThreadContextBasedTracerContextStorage.java was made as a bug was identified in it where SpanReference was not null but span within it was null. Due to this, security plugin would propagate incorrect empty span reference. Hence, added this check to avoid that. Will add screenshot of same in this PR to highlight where we observed this

### Check List

image

### DIFF
--- a/server/src/main/java/org/opensearch/action/ActionModule.java
+++ b/server/src/main/java/org/opensearch/action/ActionModule.java
@@ -595,13 +595,10 @@ public class ActionModule extends AbstractModule {
                 new RestHeaderDefinition(WorkloadGroupTask.WORKLOAD_GROUP_ID_HEADER, false)
             )
         ).collect(Collectors.toSet());
-        Set<String> transients =
-            Stream.concat(
-                actionPlugins.stream().flatMap(p -> p.getTransients().stream()),
-                Stream.of(
-                    TracerContextStorage.CURRENT_SPAN
-                )
-            ).collect(Collectors.toSet());
+        Set<String> transients = Stream.concat(
+            actionPlugins.stream().flatMap(p -> p.getTransients().stream()),
+            Stream.of(TracerContextStorage.CURRENT_SPAN)
+        ).collect(Collectors.toSet());
         UnaryOperator<RestHandler> restWrapper = null;
         for (ActionPlugin plugin : actionPlugins) {
             UnaryOperator<RestHandler> newRestWrapper = plugin.getRestHandlerWrapper(threadPool.getThreadContext(), headers, transients);
--- a/server/src/main/java/org/opensearch/action/ActionModule.java
+++ b/server/src/main/java/org/opensearch/action/ActionModule.java
@@ -500,6 +500,7 @@ import org.opensearch.rest.action.search.RestPutSearchPipelineAction;
 import org.opensearch.rest.action.search.RestSearchAction;
 import org.opensearch.rest.action.search.RestSearchScrollAction;
 import org.opensearch.tasks.Task;
+import org.opensearch.telemetry.tracing.TracerContextStorage;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.client.node.NodeClient;
 import org.opensearch.usage.UsageService;
@@ -594,9 +595,16 @@ public class ActionModule extends AbstractModule {
                 new RestHeaderDefinition(WorkloadGroupTask.WORKLOAD_GROUP_ID_HEADER, false)
             )
         ).collect(Collectors.toSet());
+        Set<String> transients =
+            Stream.concat(
+                actionPlugins.stream().flatMap(p -> p.getTransients().stream()),
+                Stream.of(
+                    TracerContextStorage.CURRENT_SPAN
+                )
+            ).collect(Collectors.toSet());
         UnaryOperator<RestHandler> restWrapper = null;
         for (ActionPlugin plugin : actionPlugins) {
-            UnaryOperator<RestHandler> newRestWrapper = plugin.getRestHandlerWrapper(threadPool.getThreadContext(), headers);
+            UnaryOperator<RestHandler> newRestWrapper = plugin.getRestHandlerWrapper(threadPool.getThreadContext(), headers, transients);
             if (newRestWrapper != null) {
                 logger.debug("Using REST wrapper from plugin " + plugin.getClass().getName());
                 if (restWrapper != null) {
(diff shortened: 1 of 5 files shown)


cwperks:
@Deepti24 I think we can simplify the changes. I believe https://github.com/cwperks/OpenSearch/pull/345 would solve the issue.

If I understand correctly, the issue is specifically on transient headers that have registered propagators. The code correctly handles the case where these headers are propagated from parent context -> stashed context but there is an issue on the restore workflow. In the PR I linked to above it modifies the restore logic to check for the transients with propagators from within the stashed context and ensures they are propagated back to the parent context. 

I think that would be all the needed changes too because CURRENT_SPAN already had a thread context propagator defined in [ThreadContextBasedTracerContextStorage](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/telemetry/tracing/ThreadContextBasedTracerContextStorage.java)

Deepti24:
@cwperks Yes this should work but just curious we could have done same for rest headers as well ? Just trying to understand why did we go with headersToCopy approach there ?
class SecurityRestFilter-> headersToCopy

cwperks:
> @cwperks Yes this should work but just curious we could have done same for rest headers as well ? Just trying to understand why did we go with headersToCopy approach there ? class SecurityRestFilter-> headersToCopy

There's a few different types of ThreadContext headers:

1. (Request) headers - These are plain String values
2. Transient headers - These are Java Object values
3. Persistent headers - These are like 1), but cannot be stashed

`headersToCopy` predates persistent headers and was done specifically to carry `X-Opaque-Id` from parent to child context. This is a special request id to identify requests coming from Kibana/OSD as I understand it.

Its since moved on to allowing more headers and actually allows plugins to also define headers (1) to copy when stashing the context.

btw stashing the context is done in cases where the system needs to change from user-context -> system context for performing system operations like reading/writing to a system index. 

I definitely think there could be cleanup in this area and would even start to think about removal of `headersToCopy` and the extension point `ActionPlugin.getRestHeaders()` in favor of using persistent headers.

Deepti24:
Thanks for the detailed response @cwperks 
Just one more observation, I guess there is some other issue due to which a span reference is being put with null span as seen in the screenshot. 
Will try to debug where is such span coming from. Basically value of current_span in thread local and original context are different somehow. Tried testing with your code changes (but anyways this is independent of code changes, have seen this before as well)

Screenshot 2026-03-10 at 11 07 20 PM

server/src/main/java/org/opensearch/action/ActionModule.java
```diff
@@ -594,9 +595,13 @@ public ActionModule(
...
             )
         ).collect(Collectors.toSet());
+        Set<String> transients = Stream.concat(
+            actionPlugins.stream().flatMap(p -> p.getTransients().stream()),
+            Stream.of(TracerContextStorage.CURRENT_SPAN)
+        ).collect(Collectors.toSet());
         UnaryOperator<RestHandler> restWrapper = null;
         for (ActionPlugin plugin : actionPlugins) {
-            UnaryOperator<RestHandler> newRestWrapper = plugin.getRestHandlerWrapper(threadPool.getThreadContext(), headers);
+            UnaryOperator<RestHandler> newRestWrapper = plugin.getRestHandlerWrapper(threadPool.getThreadContext(), headers, transients);
```
reta, line 604: @Deepti24 this change is unclear to me: the span, if propagated, should be set in thread context (transient headers), there should be no custom logic involved (ideally) to carry it forward

Deepti24: Hey @reta , @cwperks has opened a separate PR to fix this issue in thread propagation itself : https://github.com/cwperks/OpenSearch/pull/345/changes
I think that would be smaller fix for this

But to explain the issue, there is a bug in security plugin (due to how restore and stash Context work). 
So I tried to follow the pattern of how headersToCopy are used, tried doing same for transientsToCopy. Then would use that in security plugin as mentioned below:
https://github.com/opensearch-project/security/pull/6000/changes
But yes, ideally we should follow the pattern you mentioned.

cwperks: @reta there's a bug on the restore path described here: https://github.com/opensearch-project/OpenSearch/pull/20822#issuecomment-4032750560

The transient headers are correctly propagating from parent tc -> stashed tc, but not from stashed tc -> restored parent tc.

PR #18195, opened 2025-05-05
Enhance terms lookup query to take a query clause

srikanthpadakanti:
### Description
Enhances the Terms lookup query to accept a query clause instead of only supporting a single docId. This expands functionality by enabling users to define document matches through flexible query logic rather than relying solely on fixed document IDs.

This change includes:

Support for Terms lookup query to use a query clause.
Validation logic to prevent using both id and query simultaneously.
Associated unit test updates to cover query-based lookup behavior. 

### Related Issues
Resolves [#17599](https://github.com/opensearch-project/OpenSearch/issues/17599)

### Check List
- [ X] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #17599, opened 2025-03-16
[Feature Request] Enhancing Terms lookup query to take a query clause rather than docId


navneet1v:
### Is your feature request related to a problem? Please describe

[Terms Lookup query](https://opensearch.org/docs/latest/query-dsl/term/terms/#terms-lookup) is a very powerful query where it allows user to use strings stored in a document in a separate index to be passed as list of terms. But this feature is still limited where in terms lookup query you can just lookup terms from a document and you cannot pass a query which can result in terms.

Example with current capability of Terms lookup query
```

PUT students
{
  "mappings": {
    "properties": {
      "student_id": { "type": "keyword" }
    }
  }
}

PUT students/_doc/1
{
  "name": "Jane Doe",
  "student_id" : "111"
}

PUT students/_doc/2
{
  "name": "Mary Major",
  "student_id" : "222"
}

PUT classes/_doc/101
{
  "name": "CS101",
  "enrolled" : ["111" , "222"]
}

GET students/_search
{
  "query": {
    "terms": {
      "student_id": {
        "index": "classes",
        "id": "101",
        "path": "enrolled"
      }
    }
  }
}
```

### Describe the solution you'd like

The solution I am proposing here is to enhance the terms lookup query to take another query clause to get the terms to be used in final query.

```
curl -XGET "http://localhost:9200/students/_search" -H 'Content-Type: application/json' -d'
{
  "query": {
    "terms": {
      "student_id": {
        "index": "classes",
        "path": "enrolled"
        "query": {
        	// standard OpenSearch query
        }
      }
    }
  }
}
'
```

### Related component

Search:Query Capabilities

### Describe alternatives you've considered

NA

### Additional context

NA

srikanthpadakanti:
@sandeshkr419 If no one is working on this, please assign this to me. I am more than happy to help.

srikanthpadakanti:
Can someone please confirm - if my findings are valid? 

[terms-query-subquery-lookup-limitation_Version3.md](https://github.com/user-attachments/files/21169022/terms-query-subquery-lookup-limitation_Version3.md)

navneet1v:
@srikanthpadakanti I read through the document which you have added. I have some questions:

> - If you use a subquery in `terms_lookup` (such as with the `query:` parameter), the terms can only be fetched when the query rewrite is executed on a shard context.

[...]

PR #20823, opened 2026-03-10
Fix terms lookup subquery to use cluster max_clause_count setting

### LINKED ISSUE #20820: [BUG] TermsQueryBuilder uses wrong setting key for max_clause_count
### Describe the bug

In `TermsQueryBuilder.java` (line 611), the terms lookup fetch-size logic reads the setting key `"indices.query.max_clause_count"` instead of the correct `"indices.query.bool.max_clause_count"`.

```java
// Current (incorrect) — TermsQueryBuilder.java:611
int maxClauseCount = idxSettings.getAsInt("indices.query.max_clause_count", 1024);
```

The correct setting key, as defined in `SearchService.INDICES_MAX_CLAUSE_COUNT_SETTING`, is:

```java
// SearchService.java:398
"indices.query.bool.max_clause_count"
```

Because `"indices.query.max_clause_count"` is never a registered setting, `getAsInt` always falls back to the hardcoded default of `1024`, making any user/operator configuration of `indices.query.bool.max_clause_count` silently ignored when computing the terms lookup fetch size.

### To reproduce

1. Set the cluster setting `indices.query.bool.max_clause_count` to a value other than 1024 (e.g., 4096).
2. Run a `terms` query with a `terms_lookup` that would benefit from a larger fetch size.
3. Observe that the fetch size is still capped at 1024 regardless of the configured value.

### Expected behavior

The terms lookup fetch-size calculation should read `"indices.query.bool.max_clause_count"` (or ideally use the typed `SearchService.INDICES_MAX_CLAUSE_COUNT_SETTING` constant) so that the operator-configured value is respected.

### Suggested fix

```java
// Option A: Use the typed Setting (preferred)
int maxClauseCount = SearchService.INDICES_MAX_CLAUSE_COUNT_SETTING.get(idxSettings);

// Option B: Use the correct string key
int maxClauseCount = idxSettings.getAsInt("indices.query.bool.max_clause_count", 1024);
```

The test in `TermQueryWithDocIdAndQueryTests.java:380` also uses the wrong key and should be updated to match.

### OpenSearch version

main (HEAD)

### Operating system

All

### Additional context

The `.bool.` segment was likely dropped during development. The setting was made dynamically updateable in [#13568](https://github.com/opensearch-project/OpenSearch/pull/13568) (2.16.0), but this code path was not updated to use the correct key.

bcsimoes:
### Description

`TermsQueryBuilder.fetch()` reads `indices.query.max_clause_count` (without `.bool.`) from index settings — a non-existent, unregistered setting path — with a hardcoded fallback of 1024. The actual cluster setting `indices.query.bool.max_clause_count` is never consulted, making the fetch limit unconfigurable.

**Fix:** Replace the raw `idxSettings.getAsInt(...)` call with `IndexSearcher.getMaxClauseCount()`, which already reflects the cluster setting via `SearchService.INDICES_MAX_CLAUSE_COUNT_SETTING`.

**Test:** Added `testTermsLookupFetchLimitUsesClusterMaxClauseCount` validating that the fetch limit mechanism reads from `IndexSearcher.getMaxClauseCount()` (backed by the registered cluster setting) rather than a hardcoded value.

Introduced in PR #18195.

### Related Issues

Resolves #20820
Relates to #17599 (partially — the feature works but the limit is misconfigured)

### Check List

### DIFF
--- a/server/src/test/java/org/opensearch/index/query/TermQueryWithDocIdAndQueryTests.java
+++ b/server/src/test/java/org/opensearch/index/query/TermQueryWithDocIdAndQueryTests.java
@@ -8,6 +8,7 @@
 
 package org.opensearch.index.query;
 
+import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.TotalHits;
 import org.opensearch.action.get.GetRequest;
@@ -372,12 +373,10 @@ public class TermQueryWithDocIdAndQueryTests extends OpenSearchTestCase {
         when(mockClient.admin()).thenReturn(mockAdminClient);
         when(mockAdminClient.indices()).thenReturn(mockIndicesAdminClient);
 
-        // Settings that produce a small fetch size
-        int maxTermsCount = 5, maxResultWindow = 5, maxClauseCount = 5;
+        // Set maxTermsCount and maxResultWindow high so maxClauseCount (3) is the limiting factor
         Settings idxSettings = Settings.builder()
-            .put("index.max_terms_count", maxTermsCount)
-            .put("index.max_result_window", maxResultWindow)
-            .put("indices.query.max_clause_count", maxClauseCount)
+            .put(IndexSettings.MAX_TERMS_COUNT_SETTING.getKey(), 100)
+            .put(IndexSettings.MAX_RESULT_WINDOW_SETTING.getKey(), 100)
             .build();
 
         // Simulate settings response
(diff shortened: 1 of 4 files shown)


server/src/test/java/org/opensearch/index/query/TermQueryWithDocIdAndQueryTests.java
```diff
@@ -372,13 +373,8 @@ public void testTermsLookupSubqueryTotalHitsExceedsFetchSize() throws Exception
...
 
-        // Settings that produce a small fetch size
-        int maxTermsCount = 5, maxResultWindow = 5, maxClauseCount = 5;
-        Settings idxSettings = Settings.builder()
-            .put("index.max_terms_count", maxTermsCount)
-            .put("index.max_result_window", maxResultWindow)
-            .put("indices.query.max_clause_count", maxClauseCount)
-            .build();
+        // Set maxTermsCount and maxResultWindow high so maxClauseCount (3) is the limiting factor
+        Settings idxSettings = Settings.builder().put("index.max_terms_count", 100).put("index.max_result_window", 100).build();
```
gaobinlong: Could we use the setting key like `IndexSettings.MAX_TERMS_COUNT_SETTING.getKey()` instead?

bcsimoes: hey @gaobinlong  gaobinlong thank you for taking time to review this pr, 100% will do in a sec

PR #20825, opened 2026-03-10
Bump `com.netflix.nebula:gradle-info-plugin` to 16.2.1

andrross:
### Check List


andrross:
Just FYI, I added the skip-diff-analyzer label because the jump in major versions seems fine to me. We should keep up with the latest, and it just so happens this plugin has released 4 major versions in the last few years. I encountered no compatibility issues with the latest version.

sandeshkr419:
~@andrross - I'm wondering how are you able to get the gradle check running after adding `skip-diff-analyzer` label.~

~I couldn't get it running on https://github.com/opensearch-project/OpenSearch/pull/20801~

Issue #20555, opened 2026-02-05
[Feature Request] WLM group-level search settings


dzane17:
### Is your feature request related to a problem? Please describe

Workload management in OpenSearch lets administrators group and manage search traffic using workload groups and rules, so that resource policies can be applied consistently to a class of requests. Today, most search behavior is governed by cluster-level defaults, which means tenants generally operate under the same baseline conditions. Callers can override some behavior via request parameters, but that approach is hard to govern at scale and can allow requests to exceed their intended limits if not tightly controlled.

As OpenSearch deployments move toward multitenancy, the need to customize search behavior per tenant increases. Requiring every client to pass the right headers/parameters on every query is operationally difficult and error-prone. Enabling per–workload group overrides aligns with existing WLM patterns, where requests are associated with a workload group and policies are applied automatically and consistently.

### Describe the solution you'd like

Allow WLM groups to optionally define `search_settings` that are applied to all search requests assigned to the group. For example, a WLM group object could look like:
```
{
  "name": "analytics",
  "resiliency_mode": "enforced",
  "resource_limits": {
    "cpu": 0.1,
    "memory": 0.1
  },
  "search_settings": {        // New field in WLM group //
    "timeout": "500ms",
    "cancel_after_time_interval": "5s",
    "max_concurrent_shard_requests": 3
  }
}
```
Possbile settings to onboard:
1. `cancel_after_time_interval`
Ensures that long-running searches are automatically canceled after a fixed interval, preventing runaway queries from consuming cluster resources indefinitely and protecting other tenants from noisy neighbors.

2. `timeout`
Enforces a hard upper bound on how long a search is allowed to execute, helping keep latency predictable for a workload group and avoiding situations where slow or stalled queries tie up search threads.

3. `max_concurrent_shard_requests`
Limits how many shard-level requests a single search can execute in parallel, reducing fan-out pressure on the cluster and preventing high-cardinality queries from overwhelming CPU and thread pools.

4. `batched_reduce_size`
Controls how many shard results are reduced at a time during the reduce phase, helping to manage memory usage for large fan-out searches and reducing peak heap pressure in multi-tenant environments.

5. `phase_took`
Enables or disables detailed per-phase timing information in responses, allowing operators to balance observability needs against response size and overhead for specific workload groups.

6. `max_buckets`
Caps the number of aggregation buckets that a search can produce, protecting the cluster from excessive memory consumption caused by unbounded or poorly designed aggregations.

### Related component

[...]

PR #20830, opened 2026-03-10
Add additional WLM search settings

dzane17:
### Description
Adds support for additional workload group search settings. Follow up to #20536 

1. `cancel_after_time_interval`
Ensures that long-running searches are automatically canceled after a fixed interval, preventing runaway queries from consuming cluster resources indefinitely and protecting other tenants from noisy neighbors.

2. `max_concurrent_shard_requests`
Limits how many shard-level requests a single search can execute in parallel, reducing fan-out pressure on the cluster and preventing high-cardinality queries from overwhelming CPU and thread pools.

3. `batched_reduce_size`
Controls how many shard results are reduced at a time during the reduce phase, helping to manage memory usage for large fan-out searches and reducing peak heap pressure in multi-tenant environments.

Currently WLM settings act strictly as defaults — if the user has explicitly set a value on the request, it is always preserved. This behavior is subject to change in subsequent PRs (pending discussion).

### Related Issues
Part of #20555 

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/wlm/WorkloadGroupSearchSettings.java
+++ b/server/src/main/java/org/opensearch/wlm/WorkloadGroupSearchSettings.java
@@ -8,10 +8,11 @@
 
 package org.opensearch.wlm;
 
+import org.opensearch.common.settings.Setting;
 import org.opensearch.common.unit.TimeValue;
 
 import java.util.Map;
-import java.util.function.Function;
+import java.util.function.Consumer;
 
 /**
  * Registry of valid workload group search settings with their validators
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/opensearch/wlm/WorkloadGroupSearchSettings.java
```diff
@@ -114,4 +120,38 @@ private static String validateTimeValue(String value) {
             return e.getMessage();
         }
     }
+
+    /**
+     * Validates a positive integer string.
+     * @param value the string to validate
+     * @return null if valid, error message if invalid
+     */
+    private static String validatePositiveInt(String value) {
```
cwperks: Let's not reinvent the wheel, I believe we have intSettings where you can specify min value as 1.

server/src/main/java/org/opensearch/wlm/WorkloadGroupSearchSettings.java
```diff
@@ -114,4 +120,38 @@ private static String validateTimeValue(String value) {
...
+            return "must be a valid integer";
+        }
+    }
+
+    /**
+     * Validates batched reduce size (must be >= 2).
+     * @param value the string to validate
+     * @return null if valid, error message if invalid
+     */
+    private static String validateBatchedReduceSize(String value) {
```
cwperks: Same here, let's re-use the existing mechanism from Setting.java

server/src/main/java/org/opensearch/wlm/WorkloadGroupSearchSettings.java
dzane17: Good point. `intSetting` is not ideal since I don't need to create a Setting object. I was able to reuse `Settings.parseInt()`, `Settings.parseTimeValue()` methods though.

jainankitk (commented):
Thanks @dzane17 for raising this PR. At a high level, the management of these settings per workload management group seems like bit of a pain to me. Is it possible to update individual setting value for a workload management group instead of updating the workload management group itself? Also, is it better to call it just `settings` and have search prefix with the setting name, so that we can have `indexing` also part of same json object instead of separate one?

dzane17:
@jainankitk Right now `search_settings` are configured the same way as `resource_limits` in a WLM group. They are unique to each WLM group and can be updated dynamically. The only difference is that search settings are optional fields. Are you recommending the same search settings be shared across multiple WLM groups?

```
{
  "_id" : "GxwBfp3_SSyEJ-MpfwFZWw",
  "name" : "analytics",
  "resiliency_mode" : "enforced",
  "resource_limits" : {
    "cpu" : 0.2,
    "memory" : 0.3
  },
  "search_settings" : {
    "batched_reduce_size" : "400",
    "cancel_after_time_interval" : "10s",
    "max_concurrent_shard_requests" : "10",
    "timeout" : "200ms"
  },
  "updated_at" : 1774646244691
}
```

I think calling it `settings` would introduce ambiguity between general WLM group config vs actual search settings that are enforced upon queries in the group. WLM does not track indexing requests, only searches. So there is no need to add a separate `index_settings` field. Any settings that impact search requests can anyway be clumped in to `search_settings`.

dzane17:
Closing in favor of https://github.com/opensearch-project/OpenSearch/pull/21523

PR #20831, opened 2026-03-10
Make BulkItemRequest immutable

msfroh:
### Description
I was talking with @itschrispeck about some JIT optimization issues in BulkItemRequest's serialization. While looking at the code, the `volatile` keyword on the `primaryResponse` field made me cringe. Why is a `BulkItemRequest` mutable at all?

It turns out that we modify the existing `BulkItemRequest` instances on the primary shard. These modified requests (as part of a `BulkShardRequest`) are sent to the replicas. That is, the `PrimaryExecutionContext` returns the modified-in-place `BulkShardRequest` that it was created with, which is sent to the replicas.

This change makes `BulkItemRequest` immutable. The `PrimaryExecutionContext` collects all of the primary responses, then produces a **new** `BulkShardRequest` that combines the original `BulkItemRequest`s with the responses, which gets forwarded to replicas.

### Related Issues
N/A

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/action/bulk/BulkItemRequest.java
+++ b/server/src/main/java/org/opensearch/action/bulk/BulkItemRequest.java
@@ -36,15 +36,12 @@ import org.apache.lucene.util.Accountable;
 import org.apache.lucene.util.RamUsageEstimator;
 import org.opensearch.action.DocWriteRequest;
 import org.opensearch.common.Nullable;
-import org.opensearch.core.common.Strings;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 import org.opensearch.core.common.io.stream.Writeable;
 import org.opensearch.core.index.shard.ShardId;
-import org.opensearch.core.xcontent.MediaTypeRegistry;
 
 import java.io.IOException;
-import java.util.Objects;
 
 /**
  * Transport request for a Single bulk item
@@ -55,9 +52,9 @@ public class BulkItemRequest implements Writeable, Accountable {
 
     private static final long SHALLOW_SIZE = RamUsageEstimator.shallowSizeOfInstance(BulkItemRequest.class);
 
-    private int id;
-    private DocWriteRequest<?> request;
-    private volatile BulkItemResponse primaryResponse;
+    private final int id;
+    private final DocWriteRequest<?> request;
+    private final BulkItemResponse primaryResponse;
 
     /**
      * @param shardId the shard id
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/opensearch/action/bulk/BulkItemRequest.java
andrross: I feel like a broken record (no pun intended) because I suggest this all the time, but can you make this a `record`?

msfroh: Good idea! Will do!

msfroh:
The comments from the PR bot (https://github.com/opensearch-project/OpenSearch/pull/20831#issuecomment-4035056322) are not bad.

* Null dereference: I believe we're guaranteed that every BulkItemRequest will have a non-`null` response after processing on the primary. Still, maybe it's worth adding a defensive null-check just in case?
* Abort handling: I found that the abort logic was dead code. I searched for references and it was only being called from tests. So, I cleaned it up.
* The missing properties in the cloning logic is tricky -- as the bot called out, if we ever add more properties to `BulkShardRequest`, we could miss copying them here. Maybe it would be safer if we move the logic into `BulkShardRequest` itself and add a dedicated unit test. That might be less of a gotcha.

server/src/main/java/org/opensearch/action/bulk/BulkPrimaryExecutionContext.java
```diff
@@ -347,24 +337,16 @@ public void markAsCompleted(BulkItemResponse translatedResponse) {
         if (translatedResponse.isFailed() == false && requestToExecute != null && requestToExecute != getCurrent()) {
             request.items()[currentIndex] = new BulkItemRequest(request.items()[currentIndex].id(), requestToExecute);
         }
-        getCurrentItem().setPrimaryResponse(translatedResponse);
+        primaryResponses[currentIndex] = translatedResponse;
         currentItemState = ItemProcessingState.COMPLETED;
+        assertInvariants(ItemProcessingState.COMPLETED);
```
andrross, line 342: My AI agent is telling me 2 code paths reach here without setting requestToExecute, which would violate the assertion:

- markOperationAsNoOp — transitions INITIAL → EXECUTED without setting requestToExecute
- failOnMappingUpdate — clears requestToExecute in markAsRequiringMappingUpdate, then transitions to EXECUTED and calls markAsCompleted

Will this cause a regression?

server/src/main/java/org/opensearch/action/bulk/BulkItemRequest.java
andrross: Outdated comment? Probably don't need the variable anymore given the field is final/immutable.

server/src/main/java/org/opensearch/action/bulk/BulkPrimaryExecutionContext.java
msfroh: That might be what's causing the tests to fail since I made this change. Let me look into it...

andrross (approved):

PR #20832, opened 2026-03-10
Fix JVM deoptimization loop due to lambda misprediction

itschrispeck:
### Description
This change resolves a JVM deoptimization loop by refactoring how primaryResponse is written to the stream. 

To do this, we explicitly check if `primaryResponse == null`, avoiding the check within the lambda. This prevents the C2 compiler from generating an `invokedynamic` instruction, which is prone to permanent deoptimization when branch traffic flip-flops. The explicit if/else allows the JVM to compile both paths, relying on CPU branch prediction instead of a JVM trap.

It also fixes a potential race condition where the volatile `primaryResponse` is read once to check, and read again for use, though it may be modified in between. 

This pulls the `writeOptionalWriteable` logic outside, but this follows the convention of other places in the repo.

### Background
We recently found elevated CPU despite no change in load. 
image

Profiling revealed it was due to a deopt/uncommon trap hit at `BulkItemRequest#writeThin`
image

This is an example deopt event from the profile:
```
jdk.Deoptimization {
  startTime = 19:11:35.451 (2026-03-10)
  compileId = 26645
  compiler = "c2"
  method = org.opensearch.action.bulk.BulkItemRequest.writeThin(StreamOutput)
  lineNumber = 142
  bci = 21
  instruction = "ifnonnull"
  reason = "unstable_if"
  action = "none"
  eventThread = "opensearch[esdock-opensearch-metric-mvp-production-cluster-phx-ginuz-sukam][write][T#15]" (javaThreadId = 267)
  stackTrace = [
    org.opensearch.action.bulk.BulkItemRequest.writeThin(StreamOutput) line: 142
    org.opensearch.action.bulk.BulkShardRequest.lambda$writeTo$3(StreamOutput, BulkItemRequest) line: 102
    org.opensearch.core.common.io.stream.StreamOutput.writeArray(Writeable$Writer, Object[]) line: 939
    org.opensearch.action.bulk.BulkShardRequest.writeTo(StreamOutput) line: 99
    org.opensearch.action.support.replication.TransportReplicationAction$ConcreteShardRequest.writeTo(StreamOutput) line: 1589
  ]
}
```

`action = "none"` implies that we've hit the recompilation cutoff for the method. Since lambdas use `invokedynamic`, my understanding is that it's possible to flip-flop between predictions, eventually hitting the cutoff. In this case, the cutoff was hit while the wrong branch was compiled, and each invocation hits the trap. 

### Related Issues
n/a

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/action/bulk/BulkItemRequest.java
+++ b/server/src/main/java/org/opensearch/action/bulk/BulkItemRequest.java
@@ -140,15 +140,8 @@ public class BulkItemRequest implements Writeable, Accountable {
         out.writeVInt(id);
         DocWriteRequest.writeDocumentRequestThin(out, request);
 
-        BulkItemResponse primaryResponse = this.primaryResponse;
-
-        // Explicitly check if primaryResponse is null to avoid branching/mispredictions in a lambda
-        if (primaryResponse != null) {
-            out.writeBoolean(true);
-            primaryResponse.writeThin(out);
-        } else {
-            out.writeBoolean(false);
-        }
+        BulkItemResponse primaryResponse = this.primaryResponse; // Read volatile once
+        out.writeOptionalWriteable((o, resp) -> resp.writeThin(o), primaryResponse);
     }
 
     @Override


andrross (approved):

server/src/main/java/org/opensearch/action/bulk/BulkItemRequest.java
```diff
@@ -139,7 +139,16 @@ public void writeTo(StreamOutput out) throws IOException {
     public void writeThin(StreamOutput out) throws IOException {
         out.writeVInt(id);
         DocWriteRequest.writeDocumentRequestThin(out, request);
-        out.writeOptionalWriteable(primaryResponse == null ? null : primaryResponse::writeThin);
+
+        BulkItemResponse primaryResponse = this.primaryResponse;
+
+        // Explicitly check if primaryResponse is null to avoid branching/mispredictions in a lambda
```
andrross: @itschrispeck @varunbharadwaj This comment will help, but it seems possible a well-meaning future developer could "simplify" this code by using `writeOptionalWriteable()` again. Could we refactor `StreamOutput::writeOptionalWriteable` to avoid the lamba instead of changing the code here?

itschrispeck: ah, `StreamOutput` already has a more fitting method signature, updated to use it. 

The regression point is something we are concerned about too, there's a number of places in the code where the if/else pattern is already used. It might be good to flag the original syntax as a forbidden pattern, can think about this more and create a follow up

PR #20833, opened 2026-03-10
Upgrade to Gradle 9.4

andrross:
### Check List

### DIFF
--- a/plugins/repository-hdfs/build.gradle
+++ b/plugins/repository-hdfs/build.gradle
@@ -152,7 +152,7 @@ for (String fixtureName : ['hdfsFixture', 'haHdfsFixture', 'secureHdfsFixture', 
     final List miniHDFSArgs = []
 
     // See please https://issues.apache.org/jira/browse/HADOOP-19486
-    miniHDFSArgs.add("-javaagent:${-> configurations.agent.incoming.artifactView({}).files.getAt(0)}")
+    miniHDFSArgs.add("-javaagent:${-> configurations.agent.incoming.artifactView({}).files.singleFile}")
 
     // If it's a secure fixture, then depend on Kerberos Fixture and principals + add the krb5conf to the JVM options
     if (fixtureName.equals('secureHdfsFixture') || fixtureName.equals('secureHaHdfsFixture')) {
--- a/libs/agent-sm/agent/build.gradle
+++ b/libs/agent-sm/agent/build.gradle
@@ -68,7 +68,7 @@ task prepareAgent(type: Copy) {
 }
 
 artifacts {
-  add("agentJar", project.jar.archiveFile.get())
+  agentJar(jar)
   agentDist(file("$buildDir/distributions")) {
     builtBy prepareAgent
   }
--- a/test/framework/build.gradle
+++ b/test/framework/build.gradle
@@ -117,7 +117,7 @@ test {
   systemProperty 'tests.gradle_wire_compat_versions', BuildParams.bwcVersions.wireCompatible.join(',')
   systemProperty 'tests.gradle_unreleased_versions', BuildParams.bwcVersions.unreleased.join(',')
 
-  jvmArgs += ["-javaagent:${-> configurations.agent.incoming.artifactView({}).files.getAt(0)}"]
+  jvmArgs += ["-javaagent:${-> configurations.agent.incoming.artifactView({}).files.singleFile}"]
 }
 
 tasks.register("integTest", Test) {


distribution/build.gradle
```diff
@@ -44,6 +44,9 @@ import java.nio.file.Path
 plugins {
   id 'base'
 }
+
+evaluationDependsOn(':libs:agent-sm:agent')
```
andrross: The need for this is coming from the reference to `project(':libs:agent-sm:agent').prepareAgent` below.  It's been fragile in the past, and something with 9.4 makes this break. I have run into this problem before (see #19711), but I could never get that PR to reliably pass. The issue is that as the gradle code is written this project depends on the evaluation of the agent project. The line here adds that dependency explicitly. However, `evaluationDependsOn` is not a recommended practice because it reduces the build's ability to do configuration in parallel. There are better ways to solve this and I may tilt at this windmill later, but for now this does fix the problem.

andrross:
@reta I ran into failures with the previous version that used `evaluationDependsOn`. I went ahead and let kiro refactor how the agent artifacts were being consumed by the other projects. Here's the description:

> Also replace the eager cross-project task reference to
:libs:agent-sm:agent with Gradle-idiomatic patterns:
>
> - Add a consumable agentDist configuration in the agent
>  project that publishes the prepareAgent output directory
 > as an artifact with a Category attribute.
> - Add a matching resolvable agent configuration in the
>   distribution subprojects to consume it via normal
>   dependency resolution.
> - Replace direct task references (project.prepareAgent,
>   project.jar) with lazy alternatives: tasks.named() for
>   task providers, lazy GStrings for deferred path
>   resolution, and closure-based dependsOn.
> 
> This removes the need for an evaluationDependsOn call which forced
> the agent project to be configured before the distribution
> project, violating Gradle best practices around project
> isolation and configuration-time coupling.

We'll see if this works...

reta:
> @reta I ran into failures with the previous version that used `evaluationDependsOn`. I went ahead and let kiro refactor how the agent artifacts were being consumed by the other projects. Here's the description:

Thanks @andrross , I will check it a bit later today, sorry about that

buildSrc/build.gradle
```diff
@@ -291,6 +291,12 @@ if (project != rootProject) {
     suppressPomMetadataWarningsFor("testFixturesApiElements")
     suppressPomMetadataWarningsFor("testFixturesRuntimeElements")
   }
+
+  // Gradle 9.4 changed the default to enforce stricter validation. Disabling
```
andrross, line 295: See https://docs.gradle.org/9.4.0/release-notes.html#stricter-validation-for-published-plugins

reta:
> @reta I ran into failures with the previous version that used `evaluationDependsOn`. I went ahead and let kiro refactor how the agent artifacts were being consumed by the other projects. Here's the description:

Just looked at it, configuration would be an idiomatic approach indeed, thanks @andrross

andrross:
@reta Can you take a look again when you have a few minutes. Not urgent. I ran into a few other failures and had to make some additional tweaks.

plugins/repository-hdfs/build.gradle
```diff
@@ -128,9 +128,11 @@ project(':test:fixtures:krb5kdc-fixture').tasks.preProcessFixture {
 
 
 // Create HDFS File System Testing Fixtures for HA/Secure combinations
+def agentJarProvider = project(':libs:agent-sm:agent').tasks.named('jar').map { it.archiveFile.get() }
 for (String fixtureName : ['hdfsFixture', 'haHdfsFixture', 'secureHdfsFixture', 'secureHaHdfsFixture']) {
   def tsk = tasks.register(fixtureName, org.opensearch.gradle.test.AntFixture) {
-    dependsOn configurations.hdfsFixture, project(':test:fixtures:krb5kdc-fixture').tasks.postProcessFixture, project(':libs:agent-sm:agent').prepareAgent
+    dependsOn configurations.hdfsFixture, project(':test:fixtures:krb5kdc-fixture').tasks.postProcessFixture
+    dependsOn { project(':libs:agent-sm:agent').tasks.named('prepareAgent') }
```
reta: I am wondering if we better switch to configuration based approach here as well, @andrross do you mind if I push small change to your pull request? Thank you.

andrross: Sure go ahead!

reta: Done, wondering what do you think about it? Thanks!

andrross: Looks good! I made a couple small tweaks here: https://github.com/opensearch-project/OpenSearch/pull/20833/changes/f2b556b902d83586aaee60852bffa54f875b841f

PR #20792, opened 2026-03-06
Feat/hunspell ref path

shayush622:
- Add INDEX_REF_PATH_SETTING for package-based hunspell dictionaries
- Add RestHunspellCacheInvalidateAction for cache invalidation endpoint
- Update HunspellService with cache management methods
- Add ref_path validation in MetadataCreateIndexService

### Description
This PR adds support for loading Hunspell dictionaries from package-based directories using a new ref_path parameter, enabling multi-tenant dictionary isolation and hot-reload capabilities.

### Related Issues
Resolves #[20712]
Link to https://github.com/opensearch-project/OpenSearch/issues/20712#event-23132088690.

### Check List
- [x] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #20712, opened 2026-02-23
[RFC] Add ref_path support for Hunspell token filter to enable multi-package same-locale dictionaries; Enable Hot-reload for Hunspell


shayush622:
### The Problem

Currently, the Hunspell token filter only supports the `locale` parameter which loads dictionaries exclusively from `config/hunspell/{locale}/`. HunspellService caches dictionaries in `ConcurrentHashMap`. Unlike synonyms (which reload from disk on every analyzer refresh), Hunspell always returns cached dictionaries. This creates the following issues:

1. __Single Dictionary Per Locale:__ Only one dictionary can exist for each locale name (e.g., `en_US`). If two different applications or tenants need different dictionaries for the same locale/language, this is not possible.

2. __Caching Conflict Risk:__ The current caching mechanism uses only the locale name as the cache key, preventing true isolation of dictionaries with the same locale name.

3. **No Hot-Reload Support**: Once a dictionary is loaded into the cache (`ConcurrentHashMap`), it persists until node restart. Updating dictionary files on disk does not reflect in the running cluster without a full restart. This prevents dynamic dictionary updates for package re-associations or dictionary version changes.

__Example of the Limitation:__

```javascript
# Current state - only one en_US dictionary possible
config/hunspell/en_US/
├── en_US.dic
└── en_US.aff

# Desired state - multiple en_US dictionaries for different packages
config/packages/
├── pkg-123/hunspell/en_US/    
├── pkg-456/hunspell/en_US/        
└── pkg-789/hunspell/en_US/      
```

### Describe the solution you'd like

__Add a new `ref_path` parameter to the Hunspell token filter__ that allows specifying a package-based path to load dictionaries from, with full cache key isolation.

#### New Parameter: `ref_path + locale (separate parameters)`

- ref_path: Package ID only (e.g., "pkg-1234")
- locale: Dictionary locale (e.g., "en_US") - REQUIRED when using ref_path

- __Format:__ `{package-id}`
- __Resolves to:__ `config/packages/{ref_path}/hunspell/{locale}/`
- __Cache Key:__ `ref_path` + ":" + locale ("{packageId}:{locale}" )

#### Usage Example

```json
PUT /index_123
{
  "settings": {
    "analysis": {
      "analyzer": {
        "abc_analyzer": {
          "type": "custom",
          "tokenizer": "standard",
          "filter": ["lowercase", "abc_hunspell"]
        }
      },
      "filter": {
        "abc_hunspell": {
[...]

PR #20840, opened 2026-03-11
feat(hunspell): Add ref_path support for package-based dictionary loa…

shayush622:
### Description

This PR adds support for loading Hunspell dictionaries from package-based directories using a new `ref_path` parameter in the hunspell token filter. This enables multi-tenant dictionary isolation where each package can have its own hunspell dictionaries independent of the traditional `config/hunspell/` location.

**Key changes:**
- **`ref_path` parameter**: New optional parameter in the hunspell token filter that specifies a package ID. When used with `locale`, dictionaries are loaded from `config/packages/{ref_path}/hunspell/{locale}/`
- **Cache key strategy**: Package-based dictionaries use `{packageId}:{locale}` cache keys to avoid collisions with traditional locale-only keys
- **Cache management**: Added `invalidateDictionary()`, `invalidateDictionariesByPackage()`, and `invalidateAllDictionaries()` methods to HunspellService for programmatic cache control
- **Security validation**: Path traversal prevention, separator injection protection, and null byte checks in `validatePackageIdentifier()`
- **Hot-reload support**: `updateable: true` flag enables `_reload_search_analyzers` API support for package-based dictionaries
- **Backward compatible**: Traditional locale-only loading (`config/hunspell/{locale}/`) continues to work unchanged

**Usage example:**
```json
{
  "type": "hunspell",
  "ref_path": "pkg-1234",
  "locale": "en_US"
}
```
This is __Part 1 of 2__ — a follow-up PR will add REST API endpoints for cache info and invalidation. Split from #20792.

### Related Issues

Resolves #[20712](https://github.com/opensearch-project/OpenSearch/issues/20712)

### Related Issues

### Check List


server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -73,6 +125,16 @@ public TokenStream create(TokenStream tokenStream) {
         return new HunspellStemFilter(tokenStream, dictionary, dedup, longestOnly);
     }
 
+    /**
+     * Returns the analysis mode for this filter.
+     * When {@code updateable: true} is set, returns {@code SEARCH_TIME} which enables hot-reload
+     * via the _reload_search_analyzers API.
+     */
+    @Override
+    public AnalysisMode getAnalysisMode() {
```
cwperks: Not too familiar with the token filter factory, but this change is required to make this hot reloadable?

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -81,4 +143,64 @@ public boolean longestOnly() {
...
 
+    /**
+     * Validates that a package identifier or locale is safe and doesn't contain
+     * path traversal sequences, separators, or other dangerous characters.
+     *
+     * @param value The value to validate (package ID or locale)
+     * @param paramName The parameter name for error messages
+     * @throws IllegalArgumentException if validation fails
+     */
+    private static void validatePackageIdentifier(String value, String paramName) {
```
cwperks: Is it possible to simplify this by using a regex with an allowlist of chars like alphanumeric, hyphen and underscore?

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -73,6 +125,16 @@ public TokenStream create(TokenStream tokenStream) {
         return new HunspellStemFilter(tokenStream, dictionary, dedup, longestOnly);
     }
 
+    /**
+     * Returns the analysis mode for this filter.
+     * When {@code updateable: true} is set, returns {@code SEARCH_TIME} which enables hot-reload
+     * via the _reload_search_analyzers API.
```
cwperks: What API is this?

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -146,6 +155,187 @@ public Dictionary getDictionary(String locale) {
...
+        Path packagesBaseDir = env.configDir().resolve("packages");
+
+        // Resolve package directory: config/packages/{packageId}/
+        Path packageDir = packagesBaseDir.resolve(packageId);
+
+        // Security check: ensure path stays under config/packages/ (prevent path traversal attacks)
+        // Both paths must be converted to absolute and normalized before comparison
+        Path packagesBaseDirAbsolute = packagesBaseDir.toAbsolutePath().normalize();
+        Path packageDirAbsolute = packageDir.toAbsolutePath().normalize();
+        if (!packageDirAbsolute.startsWith(packagesBaseDirAbsolute)) {
```
cwperks: Can this ever be false given how `packageDir` is created from `packagesBaseDir.resolve(packageId);`? I suppose its defensive, but the validation from above prevents path traversal.

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -146,6 +155,187 @@ public Dictionary getDictionary(String locale) {
...
+            throw new IllegalArgumentException(
+                String.format(Locale.ROOT, "Package path must be under config/packages directory. Package: [%s]", packageId)
+            );
+        }
+
+        // Additional check: ensure the resolved package directory is exactly one level under packages/
+        // This prevents packageId=".." or "foo/../bar" from escaping
+        if (!packageDirAbsolute.getParent().equals(packagesBaseDirAbsolute)) {
```
cwperks: same here. Can this ever happen?

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -146,6 +155,187 @@ public Dictionary getDictionary(String locale) {
...
+                )
+            );
+        }
+
+        // Load settings from the locale directory if present
+        Settings dictSettings = loadDictionarySettings(dicDir, Settings.EMPTY);
+        boolean ignoreCase = dictSettings.getAsBoolean("ignore_case", defaultIgnoreCase);
+
+        // Find affix and dictionary files
+        Path[] affixFiles = FileSystemUtils.files(dicDir, "*.aff");
```
cwperks: Do we already have logic somewhere in the repo that reads these files?

server/src/test/resources/indices/analyze/conf_dir/packages/test-pkg/hunspell/en_US/en_US.dic
cwperks, line 1: Should we add a smaller file in for this if its for testing?

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
cwperks: I see there is `loadDictionary` already in this class. Can we extend that to support `ref_path` so that the logic can be re-used?

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
cwperks: Is it the ISM API? https://github.com/opensearch-project/index-management/blob/b9f9f58b90df8732f6c77b8c1a2ce91ea50bffad/src/main/kotlin/org/opensearch/indexmanagement/refreshanalyzer/RestRefreshSearchAnalyzerAction.kt#L57

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: Yes, it would be needed to make the dictionaries(hunspell) hot-reloadable. This would be used by a flag inside index-settings: `updatable` defaulting to false which means Analysis.all (INDEX + SEARCH time). if someone explicitly mentions this as true it would mean Analysis.SEARCH i.e. making it a reloadable analyzer

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: yes that is correct: refresh_search_analyzers api; will correct its name in the remarks.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: Yes, that is a good suggestion, thanks. Will implement this.

pgtgrly:
The ref_path support and package-based dictionary loading looks good. However, the cache invalidation methods (invalidateDictionary, invalidateDictionariesByPackage, invalidateAllDictionaries) in HunspellService suggest we're still using node-level caching. Hunspell dictionaries should use index-level caching to align with how all other analyzers work. This simplifies the invalidation story _refresh_search_analyzers handles it naturally without needing explicit cache management APIs.

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
shayush622: yes just added for defensive check.

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
shayush622: Defense-in-depth.

server/src/test/resources/indices/analyze/conf_dir/packages/test-pkg/hunspell/en_US/en_US.dic
shayush622: Ack, will shorten it.  Although, this was made too large seeing the size of the following file: There are a few more comments to be addressed:
 @/OpenSearch/server/src/test/resources/indices/analyze/conf_dir/hunspell/en_US/en_US.dic

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -81,4 +144,45 @@ public boolean longestOnly() {
         return longestOnly;
     }
 
+    /**
+     * Allowlist pattern for safe package identifiers and locales.
+     * Permits alphanumeric characters, hyphens, underscores, and dots (but not leading/trailing dots).
+     * Examples: "pkg-1234", "en_US", "my-package-v2", "en_US_custom"
+     */
+    private static final Pattern SAFE_IDENTIFIER_PATTERN = Pattern.compile("^[a-zA-Z0-9][a-zA-Z0-9._-]*[a-zA-Z0-9]$|^[a-zA-Z0-9]$");
```
cwperks: Can we forbid . entirely to simplify?

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -255,4 +418,83 @@ private static Settings loadDictionarySettings(Path dir, Settings defaults) thro
...
+    }
+
+    /**
+     * Force reloads a dictionary from disk by invalidating cache then loading fresh.
+     *
+     * @param packageId The package ID (e.g., "pkg-1234")
+     * @param locale The locale (e.g., "en_US")
+     * @return The newly loaded Dictionary
+     */
+    public Dictionary reloadDictionaryFromPackage(String packageId, String locale) {
```
cwperks: Where is this called?

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -146,6 +155,146 @@ public Dictionary getDictionary(String locale) {
...
+    /**
+     * Loads a hunspell dictionary from a package directory.
+     * Expects hunspell files at: config/packages/{packageId}/hunspell/{locale}/
+     *
+     * @param packageId The package identifier
+     * @param locale The locale (e.g., "en_US")
+     * @return The loaded Dictionary
+     * @throws Exception if loading fails
+     */
+    private Dictionary loadDictionaryFromPackage(String packageId, String locale) throws Exception {
```
cwperks, line 198: Can we re-use the existing `loadDictionary` function and add a separate arg for `packageDir`? For non-custom dictionaries we can pass `hunspellDir` and otherwise pass in the custom `packageDir`.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -39,29 +39,82 @@
...
-            throw new IllegalArgumentException(String.format(Locale.ROOT, "Unknown hunspell dictionary for locale [%s]", locale));
+        if (refPath != null) {
+            // Package-based loading: ref_path (package ID) + locale (required)
+            if (locale == null) {
+                throw new IllegalArgumentException("When using ref_path, the 'locale' parameter is required for hunspell token filter");
+            }
+
+            // Validate ref_path and locale are safe package/locale identifiers
+            validatePackageIdentifier(refPath, "ref_path");
+            validatePackageIdentifier(locale, "locale");
```
cwperks, line 93: Should we validate refPath and local similarly? Is locale more freeform or is there an enumerated list of values it can be?

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -39,29 +39,82 @@
...
+            // Traditional locale-based loading (backward compatible)
+            // Loads from config/hunspell/{locale}/
+            // Validate locale to prevent path traversal and cache key ambiguity
+            validatePackageIdentifier(locale, "locale");
+            dictionary = hunspellService.getDictionary(locale);
+            if (dictionary == null) {
+                throw new IllegalArgumentException(String.format(Locale.ROOT, "Unknown hunspell dictionary for locale [%s]", locale));
+            }
+        } else {
+            throw new IllegalArgumentException("missing [locale | language | lang] configuration for hunspell token filter");
```
cwperks: Are all of these supported or only `locale`? Let's have the error messages be actionable.

shayush622:
Scoping this PR to focus on ref_path support and package-based dictionary loading only (Phase 1).

**What this PR covers:**
- `ref_path` parameter for loading dictionaries from `config/packages/{packageId}/hunspell/{locale}/`
- Node-level cache with `{packageId}:{locale}` cache key strategy for multi-tenant isolation
- Input validation via regex allowlist
- Shared `loadDictionary` with baseDir parameter for code reuse
- Comprehensive unit tests

**What's deferred to a follow-up PR:**
- Hot-reload support via `_reload_search_analyzers` integration — per discussion with @cwperks, the preferred approach is to reuse the existing `_reload_search_analyzers` API (rather than introducing a new endpoint) where each invocation invalidates the node-level cache and re-reads dictionary data from disk. This is based on the assumption that reload is an infrequent operation. Further design discussion in RFC: #20712

Removing `updateable`, `reloadDictionaryFromPackage`, `invalidateDictionary`, and cache diagnostic methods from this PR to keep it focused on ref_path support.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: Yes sure

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
shayush622: Yes sure, that would actually simplify things, thanks.

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
shayush622: This is a dead code, will remove this, thanks for pointing out.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: Yes, both `ref_path` and `locale` are validated identically using the same `validatePackageIdentifier()` method with a regex allowlist (`^[a-zA-Z0-9][a-zA-Z0-9_-]*[a-zA-Z0-9]$`). This allows alphanumeric characters, hyphens, underscores, and dots (not leading/trailing).

Locale is freeform — there's no enumerated list. Hunspell locales are user-defined directory names (e.g., `en_US`, `de_DE`, `en_US_custom`, `medical_v2`). The existing behavior before this PR also treats locale as freeform — it just resolves `config/hunspell/{locale}/` without validating against any list.

The validation here is purely for security (prevent path traversal, cache key injection) rather than semantic correctness. If a user specifies a non-existent locale like `foo_BAR`, the loading will fail with "Could not find hunspell dictionary [foo_BAR]" which is the existing behavior.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: Done.

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -187,44 +327,55 @@ private void scanAndLoadDictionaries() throws IOException {
      * @return The loaded Hunspell dictionary
      * @throws Exception when loading fails (due to IO errors or malformed dictionary files)
      */
-    private Dictionary loadDictionary(String locale, Settings nodeSettings, Environment env) throws Exception {
+    /**
```
cwperks, line 321: This comment looks to be duplicated up above. Can we remove the comment block above this?

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -187,44 +327,55 @@ private void scanAndLoadDictionaries() throws IOException {
...
+     * package-based loading (baseDir=packageHunspellDir).
+     *
+     * @param locale       The locale of the hunspell dictionary to be loaded
+     * @param nodeSettings The node level settings (pass Settings.EMPTY for package-based loading)
+     * @param env          The node environment
+     * @param baseDir      The base directory containing locale subdirectories with .aff/.dic files
+     * @return The loaded Hunspell dictionary
+     * @throws Exception when loading fails
+     */
+    private Dictionary loadDictionary(String locale, Settings nodeSettings, Environment env, Path baseDir) throws Exception {
```
cwperks, line 334: Nice to see re-use with the baseDir passed in as a param.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -54,14 +78,32 @@ public class HunspellTokenFilterFactory extends AbstractTokenFilterFactory {
...
+            if (locale == null) {
+                throw new IllegalArgumentException("When using ref_path, the 'locale' parameter is required for hunspell token filter");
+            }
+
+            // Validate ref_path and locale are safe package/locale identifiers
+            validatePackageIdentifier(refPath, "ref_path");
+            validatePackageIdentifier(locale, "locale");
+
+            // Load from package directory: config/analyzers/{ref_path}/hunspell/{locale}/
+            dictionary = hunspellService.getDictionaryFromPackage(refPath, locale);
```
cwperks, line 96: Inside `getDictionaryFromPackage`, I can see it does a call to `dictionaries.computeIfAbsent`. Is there any path where the dictionary is reloaded from the fs on `/_refresh_search_analyzers` call?

shayush622: No, in this PR there is no path where the dictionary is reloaded from disk on `_reload_search_analyzers`. 

Currently, `computeIfAbsent` means if the cache key already exists, it returns the cached Dictionary object without hitting disk. Since HunspellService is a node-level singleton, once a dictionary is cached, it stays cached for the lifetime of the node.

When `_reload_search_analyzers` is called, `MapperService.reloadSearchAnalyzers()` calls `registry.buildTokenFilterFactories()` which recreates this `HunspellTokenFilterFactory`. The constructor calls `getDictionaryFromPackage()` → `computeIfAbsent()` → cache HIT → returns the same Dictionary object. So the dictionary is NOT reloaded from disk.

This is intentional for Phase 1 — as discussed that hot-reload integration will be addressed in a follow-up PR.

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -112,16 +113,18 @@ public class HunspellService {
...
     public HunspellService(final Settings settings, final Environment env, final Map<String, Dictionary> knownDictionaries)
         throws IOException {
         this.knownDictionaries = Collections.unmodifiableMap(knownDictionaries);
+        this.env = env;
         this.hunspellDir = resolveHunspellDirectory(env);
         this.defaultIgnoreCase = HUNSPELL_IGNORE_CASE.get(settings);
         this.loadingFunction = (locale) -> {
             try {
-                return loadDictionary(locale, settings, env);
+                return loadDictionary(locale, settings, env, hunspellDir);
```
cwperks, line 127: Won't this loadingFunction fail for custom dictionaries?

cwperks: nvm, the loading func is inline in getDictionaryForPackage

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
cwperks: So there's 2 paths I can see:

1. `getDictionary` - built-in dictionaries, there is a cache lookup here `dictionaries.computeIfAbsent` to avoid re-reading from the gs
2. `getDictionaryFromPackage` - custom dictionaries. In current iteration of the PR it also does a lookup and doesn't re-read

What would be the concern about always re-reading in the case of 2? As far as I understand, `getDictionaryFromPackage` is only called on 1) index creation, 2) index open and 3) call to `_refresh_search_analyzers`. 

I think the alternative would be figuring out a more intelligent way of re-reading (i.e. files changed on disk after map entry entered) or make the dictionaries map into an actual cache where the entries have a ttl.

cwperks: Then on expiry of the entry the loading fn is called.

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
shayush622: Yes sure, removed!

Issue #12336, opened 2024-02-15
[BUG] org.opensearch.upgrades.IndexingIT.testIndexingWithSegRep flaky


peternied:
### Describe the bug

org.opensearch.upgrades.IndexingIT.testIndexingWithSegRep is flaky, [jenkins logs](https://build.ci.opensearch.org/job/gradle-check/33789/testReport/org.opensearch.upgrades/IndexingIT/testIndexingWithSegRep/).

Assertion is from this codeblock:
https://github.com/opensearch-project/OpenSearch/blob/f73c82aa466cf3354b5d2e3e3f46f0613cacbb0f/qa/rolling-upgrade/src/test/java/org/opensearch/upgrades/IndexingIT.java#L119-L121

```
Suppressed: java.lang.AssertionError: expected:<0> but was:<1>
  at org.junit.Assert.fail(Assert.java:89)
  at org.junit.Assert.failNotEquals(Assert.java:835)
  at org.junit.Assert.assertEquals(Assert.java:647)
  at org.junit.Assert.assertEquals(Assert.java:633)
  at org.opensearch.upgrades.IndexingIT.lambda$waitForSearchableDocs$0(IndexingIT.java:121)
  at org.opensearch.test.OpenSearchTestCase.assertBusy(OpenSearchTestCase.java:1106)
  ... 40 more
```

### Related component

Indexing

### To Reproduce

```
REPRODUCE WITH: ./gradlew ':qa:rolling-upgrade:v2.13.0#upgradedClusterTest' --tests "org.opensearch.upgrades.IndexingIT.testIndexingWithSegRep" -Dtests.seed=326EEB751FF1EA22 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=ru-RU -Dtests.timezone=Africa/Timbuktu -Druntime.java=21
```

### Expected behavior

Test pass consistently

### Additional Details

- Might be related to https://github.com/opensearch-project/OpenSearch/issues/7679

peternied:
@andrross @dreamer-89 @sarthakaggarwal97 @mch2 Adding you folks in since you seem to have context on the previous fix - this issue might be from a different source since the callstack don't seem to align with #7679

peternied:
[...]

Issue #11933, opened 2024-01-18
[BUG] org.opensearch.upgrades.RefreshVersionInClusterStateIT.testRefresh is flaky


dblock:
### Describe the bug

```
org.opensearch.client.ResponseException: method [GET], host [http://[::1]:42577], URI [/_cat/nodes?h=id,version], status line [HTTP/1.1 400 Bad Request]
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Values less than -1 bytes are not supported: -2b"}],"type":"illegal_argument_exception","reason":"Values less than -1 bytes are not supported: -2b"},"status":400}
```

```
org.opensearch.client.ResponseException: method [GET], host [http://[::1]:42577], URI [/_cat/nodes?h=id,version], status line [HTTP/1.1 400 Bad Request]
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Values less than -1 bytes are not supported: -2b"}],"type":"illegal_argument_exception","reason":"Values less than -1 bytes are not supported: -2b"},"status":400}
	at __randomizedtesting.SeedInfo.seed([36B948FA59B4B2A6:CC7A3B5B7F5DC201]:0)
	at app//org.opensearch.client.RestClient.convertResponse(RestClient.java:385)
	at app//org.opensearch.client.RestClient.performRequest(RestClient.java:355)
	at app//org.opensearch.client.RestClient.performRequest(RestClient.java:330)
	at app//org.opensearch.upgrades.RefreshVersionInClusterStateIT.testRefresh(RefreshVersionInClusterStateIT.java:29)
	at java.base@21.0.1/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base@21.0.1/java.lang.reflect.Method.invoke(Method.java:580)
	at app//com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1750)
	at app//com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:938)
	at app//com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:974)
	at app//com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:988)
	at app//com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
[...]

Issue #5065, opened 2022-11-03
[BUG] Cluster crushes after upgrade of one node (rolling upgrade) 2.4.0->3.0.0


martin-gaievski:
**Describe the bug**
Cluster return 
> `cluster` failed to wait for cluster health yellow after 40 SECONDS
    IO error while waiting cluster
    503 Service Unavailable

after one out of three nodes got upgraded from 2.4.0 to 3.0.0. It's a test cluster, we use core build-tools to construct it and [upgrade the node](https://github.com/opensearch-project/k-NN/blob/main/qa/rolling-upgrade/build.gradle#L52).

**To Reproduce**
We're seeing this as part of our test BWC workflow in CI [example](https://github.com/opensearch-project/k-NN/actions/runs/3388362234/jobs/5630254342#step:5:1228). It's also reproducible in local dev environment.

Steps to reproduce the behavior locally:
1. Clone k-NN main branch from https://github.com/opensearch-project/k-NN
2. Execute command below from dev guide - https://github.com/opensearch-project/k-NN/blob/main/DEVELOPER_GUIDE.md#backwards-compatibility-testing:
```
./gradlew :qa:bwcTestSuite -Dbwc.version=2.4.0
```
or narrowed version for rolling upgrade
```
./gradlew :qa:rolling-upgrade:testRollingUpgrade -Dbwc.version=2.4.0
```
I've constructed minimal setup that run just one test and test doesn't call any of k-NN specific API. I've captured log files from each cluster node individually, please check attached zip. In that test we started 3 nodes cluster, then node with index 0 got upgraded to 3.0.0 while nodes 1 and 2 were still running 2.4.0. 

[knn-minimal-test.zip](https://github.com/opensearch-project/OpenSearch/files/9933254/knn-minimal-test.zip)

From my understanding what essentially happens is nodes on different version do not discover each other.

I do not see any errors on old 2.4.0 nodes, but on upgraded 3.0.0 node there is this error after node attempted to start:
```
[2022-11-03T21:12:23,365][WARN ][r.suppressed             ] [knnBwcCluster-rolling-0] path: /_cluster/health, params: {wait_for_status=yellow, wait_for_nodes=>=3}
org.opensearch.discovery.ClusterManagerNotDiscoveredException: null
[...]

PR #20844, opened 2026-03-11
Add node-level JVM and CPU runtime metrics

sakrah:
### Description

**Motivation**

OpenSearch's telemetry framework (`MetricsRegistry`) currently exposes no JVM runtime metrics. Operators monitoring production clusters rely on external mechanisms -- the `_nodes/stats` REST API (polling-based, adds HTTP overhead), standalone JMX exporters (separate process, separate config), or the OTel Java agent (bytecode instrumentation, heavyweight dependency). None of these integrate with the telemetry backend that OpenSearch is already configured to use.

This means if a cluster has the OTel plugin (or any future telemetry plugin) enabled, JVM metrics like heap pressure, GC pause time, and thread exhaustion still require a separate collection pipeline. There's no reason for that gap -- the data is already available via `JvmService.stats()`, which caches `JvmStats` snapshots with a 1-second TTL.

**What this PR does**

Adds `NodeRuntimeMetrics`, a `Closeable` component that registers ~30 pull-based gauges through `MetricsRegistry`, covering JVM memory, GC, buffer pools, threads, class loading, and CPU. Since it uses the framework-agnostic `MetricsRegistry` API, these metrics are automatically available through whatever telemetry backend the cluster is configured with -- no additional infrastructure required.

Metric names follow [OpenTelemetry JVM semantic conventions](https://opentelemetry.io/docs/specs/semconv/runtime/jvm-metrics/) (`jvm.memory.used`, `jvm.gc.duration`, `jvm.thread.count`, etc.) so they align with the broader ecosystem and are immediately recognizable to operators already using OTel-based tooling.

**Metrics registered**

| Category | Metrics | Tags |
|----------|---------|------|
| Memory | `jvm.memory.used`, `jvm.memory.committed`, `jvm.memory.limit` | `type` (heap/non_heap), `pool` (per memory pool) |
| GC | `jvm.gc.duration`, `jvm.gc.count` | `gc` (per collector) |
| Buffer pools | `jvm.buffer.memory.used`, `jvm.buffer.memory.limit`, `jvm.buffer.count` | `pool` (direct/mapped) |
| Threads | `jvm.thread.count` (total + per-state breakdown) | `state` (runnable, waiting, etc.) |
| Classes | `jvm.class.count`, `jvm.class.loaded` | — |
| CPU | `jvm.cpu.time`, `jvm.cpu.recent_utilization`, `process.uptime` | — |

**Design**

- All gauge suppliers read through `JvmService.stats()`, which caches `JvmStats` with a 1-second TTL -- a single collection sweep reuses one snapshot across all gauges with no redundant MXBean calls.
- Memory pools, GC collectors, and buffer pools are discovered dynamically from the initial `JvmStats` snapshot and tagged by name, so gauges work across G1, Parallel, CMS, ZGC, and other collectors without configuration.
- Per-state thread counts use a separate synchronized cache (1s TTL) to avoid redundant `getThreadInfo()` calls across the 6 per-state gauges.
- `NodeRuntimeMetrics` implements `Closeable` with an idempotent `AtomicBoolean` guard and is added to `Node`'s `resourcesToClose`.
- `ThreadMXBean` is injectable via a package-private constructor for unit testing.

### Related Issues

_N/A (new functionality)_

### Check List
- [x] New functionality has been documented (Javadoc)
- [ ] API changes companion pull request - N/A (no public API change)

### Testing

Unit tests in `NodeRuntimeMetricsTests` verify:
- All expected gauges are registered with correct names, units, and tags
- Gauge values reflect `JvmStats` snapshot data
- Per-state thread counts are correctly computed from `ThreadInfo[]`
- `close()` releases all gauge handles and is idempotent


server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -1035,6 +1038,14 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
             );
             final MonitorService monitorService = new MonitorService(settings, threadPool, fsServiceProvider);
 
+            final NodeRuntimeMetrics nodeRuntimeMetrics = new NodeRuntimeMetrics(
+                metricsRegistry,
+                monitorService.jvmService(),
+                ProcessProbe.getInstance(),
+                OsProbe.getInstance()
+            );
+            resourcesToClose.add(nodeRuntimeMetrics);
```
msfroh, line 1047: I'm not going to object to it in this PR, just because we *also* have `ClusterManagerMetrics` created in the `Node` constructor.

In the long run, though, I think we should move the creation of metrics to somewhere else. I'm not sure where that somewhere else is yet.

server/src/main/java/org/opensearch/monitor/NodeRuntimeMetrics.java
```diff
@@ -0,0 +1,440 @@
...
+            gaugeHandles.add(
+                registry.createGauge(JVM_MEMORY_LIMIT, "JVM memory pool max", UNIT_BYTES, () -> poolBytes(poolName, false), poolTags)
+            );
+        }
+    }
+
+    private double poolBytes(String poolName, boolean used) {
+        for (JvmStats.MemoryPool pool : jvmService.stats().getMem()) {
+            if (pool.getName().equals(poolName)) {
+                long bytes = (used ? pool.getUsed() : pool.getMax()).getBytes();
```
msfroh: We're keeping track of `used` and `max` here, but we're not using the values from `last_gc_stats`. IMO, those are among the most useful metrics. 

Heap usage alone is not that interesting, since it tends to trend up to around 70%, when garbage collections start to kick in. Then the heap will stick around 70%, just because the garbage collector is kind of lazy. By comparison, the `last_gc_stats` will give you a better sense of how much heap your node is *really* using. If it's trending up, it suggests a possible memory leak. If it's low and stays low, it suggests that your heap could probably be smaller.

msfroh: Also, a boolean parameter on a method that basically says "give me value A or value B, depending on the value of this parameter" tends to be a code smell. (See e.g. https://martinfowler.com/bliki/FlagArgument.html.)

I would make the helper method like `MemoryPool getPoolByName(String name`. Then the per-pool gauges can be added like:

```
gaugeHandles.add(
    registry.createGauge(
        JVM_MEMORY_USED, 
        "JVM memory pool used", 
        UNIT_BYTES, 
        () -> getPoolByName(poolName).getUsed(), 
        poolTags
    )
);
```

server/src/main/java/org/opensearch/node/Node.java
sakrah: Agreed. Will take this as part of a larger discussion

server/src/main/java/org/opensearch/monitor/NodeRuntimeMetrics.java
sakrah: Agreed. Fixed

sakrah:
## CI Failure — Unrelated Flaky Test

The Gradle check failed on `:qa:rolling-upgrade:v2.19.5-remote#twoThirdsUpgradedTest`:

```
process was found dead while waiting for cluster health yellow
```

Root cause is an `AssertionError` in `IndexShard.updateShardState` during a rolling upgrade from 2.19.5 to 3.6.0:

```
fatal error in thread [opensearch[v2.19.5-remote-0][clusterApplierService#updateTask][T#1]], exiting
java.lang.AssertionError: a started primary with non-pending operation term must be in primary mode
  [closed_index_replica_allocation][0]
  at org.opensearch.index.shard.IndexShard.updateShardState(IndexShard.java:929)
```

Related flaky test issues in the rolling-upgrade suite:
- https://github.com/opensearch-project/OpenSearch/issues/5065
- https://github.com/opensearch-project/OpenSearch/issues/11933
- https://github.com/opensearch-project/OpenSearch/issues/12336

Retrying CI with an empty commit.

PR #20852, opened 2026-03-12
Add Liquid Cache runtime initialization and query execution metrics logging

cocosz:
### Description
This change integrates Liquid Cache as the underlying runtime environment for DataFusion query execution and introduces logging to verify cache usage and effectiveness.

The Liquid Cache runtime is initialized once per process and reused across queries. Query execution paths now emit cache usage and execution metrics to help validate that the cache is actively being utilized.

### Motivation

Liquid Cache provides byte-level caching for Parquet file reads, which can significantly reduce repeated I/O for frequently accessed data.

However, verifying whether the cache is actually being used during query execution is difficult without explicit instrumentation. This PR introduces runtime-level logging that surfaces cache usage and execution metrics to aid debugging and validation.

### Architecture Diagram

                                 +----------------------+
                                 |   Java / JNI Caller  |
                                 +----------+-----------+
                                            |
                                            v
                     +-----------------------------------------------+
                     |          query_executor.rs                    |
                     |-----------------------------------------------|
                     | execute_query_with_cross_rt_stream()          |
                     | execute_fetch_phase()                         |
                     +-------------------+---------------------------+
                                         |
                                         v
                     +-----------------------------------------------+
                     |         LiquidOnlyRuntime::init()             |
                     |-----------------------------------------------|
                     | OnceLock singleton                            |
                     | - initializes Liquid Cache once               |
                     | - stores RuntimeEnv                           |
                     | - stores LiquidCacheRef / keepalive handle    |
                     +-------------------+---------------------------+
                                         |
                                         v
                     +-----------------------------------------------+
                     |        Liquid Cache RuntimeEnv                |
                     |-----------------------------------------------|
                     | Byte-level Parquet cache                      |
                     | Cache dir: /var/lib/opensearch/liquid_cache   |
                     +-------------------+---------------------------+
                                         |
                                         v
                     +-----------------------------------------------+
                     |      DataFusion Session / Physical Plan       |
                     |-----------------------------------------------|
                     | SessionStateBuilder                           |
                     | ListingTable / ListingOptions                 |
                     | Projection / Optimizers                       |
                     | execute_stream()                              |
                     +-------------------+---------------------------+
                                         |
                                         v
                     +-----------------------------------------------+
                     |            Parquet / Object Store             |
                     |-----------------------------------------------|
                     | Local FS / storage layer                      |
                     | Reads go through Liquid Cache runtime         |
                     +-----------------------------------------------+

### DIFF
--- a/plugins/engine-datafusion/jni/src/liquid_cache.rs
+++ b/plugins/engine-datafusion/jni/src/liquid_cache.rs
@@ -17,7 +17,7 @@ use datafusion::{
 use liquid_cache_datafusion_local::LiquidCacheLocalBuilder;
 use liquid_cache_datafusion_local::storage::cache::LiquidCache;
 use liquid_cache_datafusion_local::storage::cache_policies::LiquidPolicy;
-use vectorized_exec_spi::log_info;
+use vectorized_exec_spi::{log_info, log_error};
 
 pub const LIQUID_CACHE_DIR: &str = "/tmp/opensearch/liquid_cache";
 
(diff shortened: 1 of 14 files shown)


plugins/engine-datafusion/jni/src/liquid_cache.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    /// Log cache stats if Liquid Cache has been initialized.
+    pub fn log_stats_if_initialized() {
+        if let Some(Ok(runtime)) = LIQUID_ONLY.get() {
+            runtime.log_stats();
+        }
+    }
+
+    /// Reset the cache, clearing all in-memory entries and disk files.
+    pub fn reset_cache(&self) {
```
chaitanya588, line 139: Since the only calling function is within the same module, you can remove "pub" annotation here. This will give us more safer way of accessing the runtime. Similar point applies to log_stats as well.
Only value i am seeing here in retaining public is, here we are clearing up the directory and recreating the directories. For some reason, if we are restarting a crashed program then we may need to reset without creating a run time. Otherwise, i don't see any usage.

plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/core/DataFusionRuntimeEnv.java
```diff
@@ -48,15 +48,39 @@ public final class DataFusionRuntimeEnv implements AutoCloseable {
         Setting.Property.NodeScope
     );
 
+    /**
+     * Controls whether Liquid Cache is enabled for byte-level Parquet caching.
+     */
+    public static final Setting<Boolean> DATAFUSION_LIQUID_CACHE_ENABLED = Setting.boolSetting(
+        "datafusion.liquid_cache.enabled",
+        false,
```
chaitanya588, line 56: Is it intentional to keep this setting as false here? Are we waiting for some kind of perf runs to keep this as true by default?

PR #20854, opened 2026-03-12
Ensure that transient ThreadContext headers with propagators survive  restore

Deepti24:
### Description
This is to fix issue mentioned in this comment: https://github.com/opensearch-project/OpenSearch/pull/20822#issuecomment-4033270180
**Issue details**
* When request reaches this class : Netty4HttpRequestHeaderVerifier, we take a snapshot of context using newStoredContext(). We store it as attribute with key CONTEXT_TO_RESTORE
* Now some spans are added to current thread. For instance: AbstractHttpServerTransport => dispatchRequest
* Now request lands on SecurityRestFilter. Here current thread's context has span info
* SecurityRestFilter would now call restore context stored earlier with key CONTEXT_TO_RESTORE
* Since span got added after we took the snapshot, it would not contain the span id.
* To fix this we should merge the context in restore method.

**Explaining why do we have a SpanReference with null span id as seen in screenshot in above comment (Fix in class ThreadContextBasedTracerContextStorage)**
* Let's say request comes to NativeMessageHandler=> handleRequest
* We start a span with SpanScope spanScope = tracer.withSpanInScope(span). It is auto-closeable
* When the block completes, it auto closes using DefaultSpanScope => close
* This method sets span id within SpanReference as null
* It is never removed from the thread and when threads are re-used, it stays (will attach screen shot on this)

### Related Issues
(https://github.com/opensearch-project/security/issues/5990)
https://github.com/cwperks/OpenSearch/pull/345

### Check List


Deepti24:
image
Screenshot showing that SpanReference stays in the thread with null span
Added a debugger to a thread from the pool that handles requests. It had SpanReference in its thread local

server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java
```diff
@@ -293,10 +293,25 @@ public StoredContext newStoredContext(boolean preserveResponseHeaders, Collectio
         final ThreadContextStruct newContext = threadLocal.get();
 
         return () -> {
+            ThreadContextStruct current = threadLocal.get();
+            ThreadContextStruct enrichedOriginalContext = originalContext;
```
cwperks: nit: I think `enrichedOriginalContext` is a bit misleading of a name for this. maybe `restoredContext`? Also wondering if we should leave a comment about how this block is specifically around propagating transients headers (that have a propagator) from original context -> new context -> back to original restored context.

cwperks:
@Deepti24 the changes lgtm. left one small comment around naming.

cwperks:
please fix the precommit failures

server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java
Deepti24: yes makes sense, I have added the changes. Will take care of build issues as well. Thanks for quick review

server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java
```diff
@@ -293,10 +293,31 @@ public StoredContext newStoredContext(boolean preserveResponseHeaders, Collectio
...
+            // dropped when the security plugin (or any other caller) calls storedContext.restore().
+            // Without this, restore() would blindly overwrite the threadLocal with the original
+            // snapshot, losing any propagated transients that were set after newStoredContext() was called.
+            ThreadContextStruct current = threadLocal.get();
+            ThreadContextStruct restoredContext = originalContext;
+            final Map<String, Object> propagated = propagateTransients(current.transientHeaders, current.isSystemContext);
+            if (!propagated.isEmpty()) {
+                Map<String, Object> merged = new HashMap<>(originalContext.transientHeaders);
+                propagated.forEach(merged::putIfAbsent);
+                restoredContext = new ThreadContextStruct(
```
cwperks: Not sure if the failures in https://build.ci.opensearch.org/job/gradle-check/72451/ are related to this.

Maybe we can do something similar to the line below `threadLocal.set(restoredContext.putResponseHeaders(threadLocal.get().responseHeaders));`?

Is there a setter for transients we can use?

Deepti24: yes test case failure appear to be unrelated
No, there is no setter for this currently. So I have added a method which would add data to transient(by creating a new one)
Please let me know if it looks okay.

Also, I want to backport this to 3.1 and 2.19. I am unable to add relevant labels. Can you please help with the same ?

Deepti24:
@cwperks Based on my understanding, test cases were failing because we use try-with-resource to handle context propagation and restore for threads. Whenever a new thread is given task, it would create a new context and restore it later when it exits. Hence, by the end of the process, we should end up with original context we started with ie. DEFAULT_CONTEXT. In our case, since we were creating new context if there were propagators in restore method, we will never end up with DEFAULT_CONTEXT in the end. Hence, tried to use the pattern used to propagate response headers back from the thread handling network calls ie. thread which wants propagation passes the flag as true
This will also need a fix in security plugin repo: https://github.com/opensearch-project/security/pull/6006

Do let me know if this looks okay

server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java
```diff
@@ -293,10 +305,24 @@ public StoredContext newStoredContext(boolean preserveResponseHeaders, Collectio
...
         return () -> {
+            // Re-apply propagator-declared transients from the current context back into the
+            // snapshot being restored. This ensures that transients written after the snapshot
+            // was taken using newStoredContext (e.g. CURRENT_SPAN set by the tracing infrastructure) are not silently
+            // dropped when the security plugin (or any other caller) calls storedContext.restore().
+            // Without this, restore() would blindly overwrite the threadLocal with the original
+            // snapshot, losing any propagated transients that were set after newStoredContext() was called.
+            ThreadContextStruct current = threadLocal.get();
+            ThreadContextStruct restoredContext = originalContext;
+            if (preserveTransients) {
```
cwperks, line 316: @Deepti24 are you sure this check is needed here? Would we be able to put into the `restoreContext.transientHeaders` map directly or is this map immutable?

cwperks: In general, I think the ThreadContext class could use some cleanup bc it looks like there are competing notions. I wonder if it would make sense in the future to copy a similar `putPersistent` notion to transient headers like `putPersistentTransient` and ensure that its clear that persistent means persisting stashing and restore.

cwperks: @Deepti24 I think we can both 1) make the `SnapshotResiliencyTests` tests and 2) always propagate transients if present.

In those tests were there transients to propagated from new context back to restored context? If there were no transients to propagate then it should still return the same exact ThreadContextStruct as what was stored.

Deepti24: There were transients to propagate back in SnapshotResiliencyTests with key TASK_ID which is causing the context to be copied and failing.
Based on my understanding, transients are immutable as all methods here tend to copy the context whenever we try to add something to it [link to code](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java#L859)
can you please elaborate a bit on 1) make the SnapshotResiliencyTests tests point

cwperks: ah ok, I guess if that's the case then the test assertion is wrong. I think we just need to ensure that the assertion is correct in spirit.

so a human readable assertion would be:

I expect my original context to be restored with the headers prior to storing the context (plus now transients that should propagate).

Right now its simply doing that the same object in memory is restored but we are breaking that now with the transient propagation so I believe the test would need to be changed accordingly.

cwperks: I believe everything but the transients should be the same `Map` in memory so maybe we can check all those but then do a separate check for the transients which we know would propagate. 

Also curious to dig deeper into why `TASK_ID` is set in a stashed context and then restored. Where is that set?

Deepti24: I thought that assertion was part of test cases initially and should be fixed but later on found out it is part of core OS classes not test class [code link of assertion](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/util/concurrent/OpenSearchThreadPoolExecutor.java#L155)
So I was not sure if its a good idea to modify it as its core class. If it is fine to modify it, then that is the cleaner approach
I guess the idea is to ensure we started with a clean context and end up with a clean one so that context is not shared across threads. Probably we can just cleanup the context here rather than assert

Deepti24: > I believe everything but the transients should be the same `Map` in memory so maybe we can check all those but then do a separate check for the transients which we know would propagate.
> 
> Also curious to dig deeper into why `TASK_ID` is set in a stashed context and then restored. Where is that set?

TaskThreadContextStatePropagator class sets the TASK_ID. Quick summary of this:
The TASK_ID is used by TaskResourceTrackingService to track resource usage (CPU, memory) for tasks. When a task spawns work on different threads, the TASK_ID needs to follow that work so resource consumption can be properly attributed back to the originating task.

Looking at flow, looks like RequestHandlerRegistry=> taskExecutionStarted() is called which calls newStoredContext() internally and stores that context. Also adds TASK_ID to current thread. Then in its finally block, it calls restore() where since task is done, it wouldn't want to restore the transient anymore as accounting for task would have been done internally. We starting to restore TASK_ID is incorrect I guess in this context, based on my understanding

cwperks: ok, I think I'm starting to see the intent.

Each thread gets a [ContextPreservingRunnable](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java#L905-L920) that starts with DEFAULT_CONTEXT and after `restore()` should end with default context.

Since we are explicitly not propagating with restore, its not getting back to DEFAULT_CONTEXT. I suppose one solution would be to add our propagation logic in, but skip if originalContext == DEFAULT_CONTEXT.

cwperks: @Deepti24 your solution makes sense. Sorry, had to do a bit of tracing to come to the same conclusion.

server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java
```diff
@@ -252,14 +252,26 @@ public StoredContext newStoredContext(boolean preserveResponseHeaders) {
...
+
     /**
      * Just like {@link #stashContext()} but no default context is set. Instead, the {@code transientHeadersToClear} argument can be used
      * to clear specific transient headers in the new context. All headers (with the possible exception of {@code responseHeaders}) are
      * restored by closing the returned {@link StoredContext}.
      *
      * @param preserveResponseHeaders if set to <code>true</code> the response headers of the restore thread will be preserved.
      */
-    public StoredContext newStoredContext(boolean preserveResponseHeaders, Collection<String> transientHeadersToClear) {
+    public StoredContext newStoredContext(
```
cwperks, line 270: @Deepti24 can we can the ordering so that the 2 bool args are first?

Otherwise lgtm.

server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java
Deepti24: No worries, it was an interesting discussion.

server/src/main/java/org/opensearch/common/util/concurrent/ThreadContext.java
Deepti24: sure will do

Deepti24: done

Deepti24:
@cwperks Will need your review here as well: https://github.com/opensearch-project/security/pull/6006
And also need to back-port this to 3.1 and 2.19.  And if possible to 2.17

cwperks:
> @cwperks Will need your review here as well: [opensearch-project/security#6006](https://github.com/opensearch-project/security/pull/6006) And also need to back-port this to 3.1 and 2.19. And if possible to 2.17

If you're managing a fork that I would suggest to cherry-pick there. We can backport it on the main repos, but there's no plan to release from the 3.1 and 2.17 branches. 2.19 is a release branch so we can backport there.

cwperks:
That's a flaky test. I will re-run the gradle check.

cwperks:
@Deepti24 can you sync this feature branch with main? I believe that will resolve the bwc failure...the 2.19 branch just got bumped to 2.19.6.

Deepti24:
> @Deepti24 can you sync this feature branch with main? I believe that will resolve the bwc failure...the 2.19 branch just got bumped to 2.19.6.

sure, synced it now

cwperks:
@Deepti24 please fix the spotless errors

```

* What went wrong:
      src/test/java/org/opensearch/telemetry/tracing/ThreadContextBasedTracerContextStorageTests.java
          @@ -286,9 +286,6 @@
           ········source·=·new·HashMap<>();
           ········source.put(ThreadContextBasedTracerContextStorage.CURRENT_SPAN,·spanReference);
           ········assertFalse(context.transients(source).isEmpty());
          -········assertEquals(
          -············span,
          -············((SpanReference)·context.transients(source).get(ThreadContextBasedTracerContextStorage.CURRENT_SPAN)).getSpan()
          -········);
          +········assertEquals(span,·((SpanReference)·context.transients(source).get(ThreadContextBasedTracerContextStorage.CURRENT_SPAN)).getSpan());
           ····}
           }
```

Deepti24:
@cwperks I wanted to fix DCO history (I signed using personal id but was supposed to do with corporate id) along with this, is it okay if I run below commands for the same ?

git reset --soft main
git commit -s -m "Fix security plugin issue and add test cases"
git push origin fix-security-plugin-issue --force

Do let me know if it is not recommended

Have fixed checkstyle

cwperks:
@Deepti24 yes you can do that. I also have a mechanism on my side to manually set DCO to pass as long as at least one of the commits is signed. PRs get squashed on merge so a maintainer can assure that the signature is kept in the squashed commit message. I think the requirement is that for all unique contributors on a PR, each contributor needs to sign at least one of their commits.

Deepti24:
> @Deepti24 yes you can do that. I also have a mechanism on my side to manually set DCO to pass as long as at least one of the commits is signed. PRs get squashed on merge so a maintainer can assure that the signature is kept in the squashed commit message. I think the requirement is that for all unique contributors on a PR, each contributor needs to sign at least one of their commits.

@cwperks Sure then please use the DCO corresponding to Signed-off-by: Deepti Chauhan  when you merge

ACCOUNT:
[dchauhan3-blip](https://github.com/opensearch-project/OpenSearch/commits?author=dchauhan3-blip)

Thanks !!

cwperks:
@Deepti24 please rebase to ensure that https://github.com/opensearch-project/OpenSearch/pull/20920 is in your branch. The tests are failing without that fix.

PR #20857, opened 2026-03-13
Disable concurrent search for filter duplicates in significant_text

andrross:
The underlying structured used to find duplicate byte sequences (DuplicateByteSequenceSpotter) is stateful and not thread-safe, so disable concurrent search if this particular feature is being used.

### Related Issues
Part of #14408, specifically `{yaml=search.aggregation/90_sig_text/Dedup noise}`

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/SignificantTextAggregatorFactory.java
+++ b/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/SignificantTextAggregatorFactory.java
@@ -315,6 +315,9 @@ public class SignificantTextAggregatorFactory extends AggregatorFactory {
 
     @Override
     protected boolean supportsConcurrentSegmentSearch() {
-        return true;
+        // The underlying structured used to find duplicate byte sequences (DuplicateByteSequenceSpotter)
+        // is stateful and not thread-safe, so disable concurrent search if this particular feature is
+        // being used.
+        return filterDuplicateText == false;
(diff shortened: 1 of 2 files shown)


andrross:
@msfroh If you don't mind I'd appreciate a look to see if this seems correct

prudhvigodithi:
We can explore supporting `DuplicateByteSequenceSpotter` thread safe as an enhancement to ensure `significant_text` aggs with `filter_duplicate_text` get benefit with concurrent segment search.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/SignificantTextAggregatorFactory.java
```diff
@@ -315,6 +315,9 @@ public void close() {
 
     @Override
     protected boolean supportsConcurrentSegmentSearch() {
-        return true;
+        // The underlying structured used to find duplicate byte sequences (DuplicateByteSequenceSpotter)
```
sohami, line 318: it is unclear why `DuplicateByteSequenceSpotter` needs to be threadsafe for inter-segment concurrent search. I see it being created at per leaf (or segment) and owning ord combination [here](https://github.com/opensearch-project/OpenSearch/pull/20857/changes#diff-96e9cdf20100682e2a2b1aae820dbfc7efaa5b1bb4606f2ebc54617eb2a0acc6R227)

PR #20858, opened 2026-03-13
Reduce ClusterState retention in retry closures

### LINKED ISSUE #15065: [BUG] [Remote Store] [Snapshots] Heavy Heap Usage on Master Node due stuck snapshot deletions for Remote Store clusters
### Describe the bug

If segment uploads on a shard is stuck due to any reason, an ongoing snapshot for the shard remains in the IN_PROGRESS state until we trigger a snapshot delete. Upon delete trigger, the snapshot transitions to the ABORTED state.

On a non Remote Store cluster, each snapshot thread on data node snapshotting that shard, [check](https://github.com/linuxpi/OpenSearch/blob/46722deb86789fb21e7f0f5af71ca4b4b73637fb/server/src/main/java/org/opensearch/repositories/blobstore/BlobStoreRepository.java#L3508) periodically if snapshot is marked aborted. If it identifies the snapshot is marked aborted, it fails the shard snapshot, eventually transitioning the snapshot to a terminal failed state.

On a Remote Store cluster, the code flow is different since its a shallow snapshot. While for remote store shards, the snapshot thread [waits for segment uploads to complete](https://github.com/linuxpi/OpenSearch/blob/46722deb86789fb21e7f0f5af71ca4b4b73637fb/server/src/main/java/org/opensearch/snapshots/SnapshotShardsService.java#L408), if the segment uploads is stuck, this thread remains suspended until segment upload resumes. hence, if the snapshot is aborted during this duration, it wont identify its aborted and wont fail the shard snapshot operations. 

This leads to snapshot being in ABORTED state indefinitely until remote segment uploads resume. Snapshot deletion calls are blocking, and wait indefinitely for deletion to complete. They do that by [queuing up a listener](https://github.com/linuxpi/OpenSearch/blob/f58cf991a0b491d663ffc73a9764b8112951b238/server/src/main/java/org/opensearch/snapshots/SnapshotsService.java#L2666) which gets invoked on deletion completion. this listener holds reference to the cluster state object. subsequent delete calls will queue up their own listeners each holding a reference of the cluster state at that point in time. So multiple cluster state objects end up on the heap which wont be GC'ed until the listeners invoked.

So if snapshot is stuck due to remote segment uploads being stuck, more and more retries to delete the stuck snapshot will get stuck and hold up cluster state reference in heap leading to increasing heap usage on active master.

### Related component

Storage:Snapshots

### To Reproduce

1. Simulate stuck segment uploads for a shard
2. trigger a snapshot, it gets stuck in IN_PROGRESS state
3. trigger a delete for the same snapshot, the delete call will timeout and delete operation will get stuck
4. Keep sending delete snapshot request and see active master jvm increase

### Expected behavior

- Delete snapshot calls can be made async where the call is just a trigger for delete and actual delete can happen in the background
- Evaluate if a check for aborted snapshot is needed in shallow copy snapshot code flow to allow snapshot operations to fail for remote upload stuck shards

### Additional Details

**Thread dump**
```
opensearch[095012d6c1bfe3db7b0f51c4bb31f9d7][snapshot_shards][T#1]" #482 daemon prio=5 os_prio=0 cpu=104.25ms elapsed=2910.84s tid=0x0000ffff584f4ad0 nid=0x198a waiting on condition  [0x0000ffff15a5a000]
   java.lang.Thread.State: WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@17.0.9/Native Method)
        - parking to wait for  <0x0000000610011bf0> (a java.util.concurrent.locks.ReentrantLock$NonfairSync)
        at java.util.concurrent.locks.LockSupport.park(java.base@17.0.9/LockSupport.java:211)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(java.base@17.0.9/AbstractQueuedSynchronizer.java:715)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(java.base@17.0.9/AbstractQueuedSynchronizer.java:938)
        at java.util.concurrent.locks.ReentrantLock$Sync.lock(java.base@17.0.9/ReentrantLock.java:153)
        at java.util.concurrent.locks.ReentrantLock.lock(java.base@17.0.9/ReentrantLock.java:322)
        at org.apache.lucene.search.ReferenceManager.maybeRefreshBlocking(ReferenceManager.java:238)
        at org.opensearch.index.engine.InternalEngine.refresh(InternalEngine.java:1860)
        at org.opensearch.index.engine.InternalEngine.flush(InternalEngine.java:1975)
        at org.opensearch.index.engine.InternalEngine.acquireLastIndexCommit(InternalEngine.java:2199)
        at org.opensearch.index.shard.IndexShard.acquireLastIndexCommit(IndexShard.java:1695)
        at org.opensearch.snapshots.SnapshotShardsService.snapshot(SnapshotShardsService.java:675)
        at org.opensearch.snapshots.SnapshotShardsService$1.doRun(SnapshotShardsService.java:393)
        at org.opensearch.common.util.concurrent.ThreadContext$ContextPreservingAbstractPrioritizedRunnable.doRun(ThreadContext.java:979)
        at org.opensearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:52)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(java.base@17.0.9/ThreadPoolExecutor.java:1136)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(java.base@17.0.9/ThreadPoolExecutor.java:635)
        at java.lang.Thread.run(java.base@17.0.9/Thread.java:840)
```

**Screenshots**
Screenshot 2024-05-22 at 3 29 32 PM

 - OS: Amazon Linux 2
 - Version OS_2.11

**Additional context**
N/A

HarishNarasimhanK:
### 1. Goal

In OpenSearch, snapshot deletion is a cluster manager routed operation. When a delete request is received, the cluster manager creates internal callback objects (listeners) to track the operation and notify the caller once it completes. These listeners inadvertently hold a reference to a large in-memory object called `ClusterState`, which contains the entire cluster's metadata, routing information, and index definitions.

When a snapshot deletion gets stuck or takes a long time to complete, users or automated systems may retry the delete request multiple times. As listeners accumulate from repeated retries, multiple `ClusterState` objects get pinned on the heap, causing the cluster manager node's memory usage to grow until it runs out of memory.

This change fixes the issue by ensuring that the listeners only hold the small pieces of information they actually need (a version number and a node identifier) instead of the entire `ClusterState` object. This allows the large `ClusterState` objects to be garbage collected immediately, preventing the memory buildup.

### 2. Current Workflow

This section traces the lifecycle of a snapshot delete request from the REST API to the point where the listener is stored in `SnapshotsService`.

1. A client sends `DELETE /_snapshot/{repository}/{snapshot}`.

2. The REST layer (`RestDeleteSnapshotAction`) constructs a `DeleteSnapshotRequest` and passes it to `NodeClient`.

3. `NodeClient` dispatches the request to `TransportDeleteSnapshotAction`, which extends `TransportClusterManagerNodeAction`.

4. The base class creates an `AsyncSingleAction` instance to manage the request lifecycle. `AsyncSingleAction` fetches the current `ClusterState` and calls `doStart(clusterState)`.

5. If the local node is the cluster manager, `doStart()` wraps the original listener using `getDelegateForLocalExecute(clusterState)`. This wrapper contains a lambda for retry logic that references the `clusterState` parameter. Due to Java lambda capture semantics, the entire `ClusterState` object is implicitly retained by this lambda for as long as the listener exists.

6. The wrapped listener is passed into `TransportDeleteSnapshotAction.clusterManagerOperation()`, which calls `snapshotsService.deleteSnapshots(request, listener)`. The listener still carries the captured `ClusterState` reference inside its retry lambda.

7. Inside `SnapshotsService`, the deletion is submitted as a cluster state update. Once the cluster state is updated to record the deletion, the listener is stored in the `snapshotDeletionListeners` map (keyed by the deletion UUID) in order to notify the client when the deletion completes.

### 3. Issue with Current Workflow

- The listener stored in `snapshotDeletionListeners` sits in the map until the repository-level deletion reaches a terminal state. If the deletion is stuck (due to slow I/O, stuck segment uploads, large repository cleanup, or any other reason), the listener remains in `snapshotDeletionListeners` indefinitely, and the captured `ClusterState` cannot be garbage collected.

- For each subsequent delete request, `SnapshotsService` adds another listener to `snapshotDeletionListeners` through the same path. As delete requests accumulate, these listeners pile up, each pinning a `ClusterState` object on the heap. The cluster manager node's heap usage grows monotonically with each repeated delete, eventually leading to `OutOfMemoryError`.

### 4. Requirements

- Reduce the size of the data retained by retry closures. Instead of capturing the full `ClusterState` object, closures should only hold the minimal primitives required for retry decisions.

- Preserve existing retry behavior and backward compatibility.

### 5. Approach: Extract Primitives Before Closure Creation

The retry closures in `TransportClusterManagerNodeAction` only need two pieces of information from the `ClusterState` to make retry decisions: the cluster state version (a `long`) and the cluster manager node's ephemeral ID (a `String`). By extracting these values before creating any lambda or anonymous class, the closures capture only these lightweight primitives. The full `ClusterState` object is no longer referenced by any closure and becomes eligible for garbage collection immediately.

#### Sequence Diagram

image

#### Implementation Steps

1. In `getDelegateForLocalExecute` inside `AsyncSingleAction`, the cluster state version and the cluster manager node are extracted before the lambda is created. The lambda now references only these extracted values, so the full `ClusterState` is no longer retained.

2. A new overloaded `retryOnMasterChange` method is added that accepts the version and cluster manager node directly. It extracts the ephemeral ID from the cluster manager node and passes it to the predicate builder.

3. The original `retryOnMasterChange` method that accepts a full `ClusterState` is kept as a convenience bridge. It extracts the version and cluster manager node and delegates to the new overload.

4. The `retry` method signature is updated to accept the version and cluster manager node instead of the full `ClusterState`. It uses the persistent node ID and version to construct the cluster state observer via a new primitives-based constructor.

5. A new overloaded `build` method is added to `ClusterManagerNodeChangePredicate` that accepts the version and ephemeral ID directly. The existing `build` method that accepts a full `ClusterState` is refactored to extract these values and delegate to the new overload.

6. Two new constructors are added to `ClusterStateObserver` that accept the cluster manager node ID and version as primitives, instead of requiring a full `ClusterState` object.

7. The `StoredState` inner class inside `ClusterStateObserver` is refactored to support construction from primitives. The existing constructor that accepts a `ClusterState` now delegates to the new primitives-based constructor.

### 6. Validation

The fix was validated by reproducing the memory retention issue on a local cluster and comparing heap dumps before and after the change.

#### Reproduction Setup

1. Added a `Thread.sleep()` call in `BlobStoreRepository.doDeleteShardSnapshots()` to simulate a long-running deletion that stays stuck.

2. Created 500 indices with heavy mappings (50+ fields each), multiple aliases per index, and a small number of documents per index to inflate the `ClusterState` size.

3. Created one snapshot per index (500 snapshots total) using a filesystem-based snapshot repository.

4. Spammed delete requests for all snapshots repeatedly, so that listeners accumulate in `snapshotDeletionListeners` while the deletion is stuck.

5. Captured heap dumps from the cluster manager node and compared the retained size of listener chains.

#### Results

**Before (without fix)**

image

**After (with fix)**

image

## Related Issues
Resolves #15065 

## Check List


server/src/main/java/org/opensearch/action/support/clustermanager/TransportClusterManagerNodeAction.java
```diff
@@ -266,7 +266,9 @@ protected void doStart(ClusterState clusterState) {
                 } else {
                     if (nodes.getClusterManagerNode() == null) {
                         logger.debug("no known cluster-manager node, scheduling a retry");
-                        retryOnMasterChange(clusterState, null);
+                        // Extract version before creating closure to avoid retaining full ClusterState
+                        final long stateVersion = clusterState.version();
+                        retryOnMasterChange(stateVersion, null, null, null);
```
Adiiigo: You are passing null values, can we check for the NPE that might occur in the calling functions.

server/src/main/java/org/opensearch/action/support/clustermanager/TransportClusterManagerNodeAction.java
```diff
@@ -285,11 +287,23 @@ protected void doStart(ClusterState clusterState) {
             }
         }
 
-        private void retryOnMasterChange(ClusterState state, Throwable failure) {
```
Adiiigo, line 288: Would it make sense to have overloaded function for this as well?

server/src/main/java/org/opensearch/action/support/clustermanager/TransportClusterManagerNodeAction.java
```diff
@@ -298,8 +308,10 @@ private void retry(ClusterState state, final Throwable failure, final Predicate<
                     listener.onFailure(new ClusterManagerNotDiscoveredException(failure));
                     return;
                 }
+                final String persistentNodeId = clusterManagerNode != null ? clusterManagerNode.getId() : null;
```
Adiiigo, line 311: Suggestion: Can we check what will be the repercussions if `persistentNodeId` is passed to `ClusterStateObserver` as null in case `clusterManagerNode` turns out to be null.

server/src/main/java/org/opensearch/cluster/ClusterManagerNodeChangePredicate.java
Adiiigo, line 63: `newClusterManager.getEphemeralId()` might return null and `.equals` would throw NPE. We can check the value first and then perform extraction.

server/src/test/java/org/opensearch/action/support/clustermanager/TransportClusterManagerNodeActionTests.java
```diff
@@ -807,6 +895,87 @@ private ClusterState buildClusterState(ClusterState state, long term, long versi
         return ClusterState.builder(state).version(version).metadata(newMetadata).build();
     }
 
+    /**
+     * Tests that when the clusterStateLatestChecker's GetTermVersion request fails with a
+     * ConnectTransportException, the handleTransportException path correctly triggers a retry
+     * and the request is eventually forwarded to a new master.
+     */
+    public void testClusterStateLatestCheckerHandlesTransportException() throws ExecutionException, InterruptedException {
```
Adiiigo, line 892: You can follow AAA testing writing strategy.

server/src/main/java/org/opensearch/action/support/clustermanager/TransportClusterManagerNodeAction.java
HarishNarasimhanK: > Suggestion: Can we check what will be the repercussions if `persistentNodeId` is passed to `ClusterStateObserver` as null in case `clusterManagerNode` turns out to be null.

When `clusterManagerNode` is null, `persistentNodeId` is passed as null to `StoredState` constructor. This is null safe as the `StoredState` class only uses this field in `isOlderOrDifferentClusterManager()` function, which compares it via `Objects.equals` (which is null-safe).

Also, `persistentNodeId` wouldn't be null if the `clusterManagerNode` is not null, as it is instantiated with the `intern()` function which would itself throw an NPE on a null input, meaning a valid `DiscoveryNode` instance is guaranteed to have a non-null nodeId, so the only way `persistentNodeId` is null is when `clusterManagerNode` itself is null.

server/src/main/java/org/opensearch/cluster/ClusterManagerNodeChangePredicate.java
HarishNarasimhanK: `getEphemeralId()` can never return null because, in the `DiscoveryNode` constructor, the ephemeral ID is stored by calling `ephemeralId.intern()`, which would itself throw a `NullPointerException` if null were passed. Constructors that do not take an explicit ephemeral ID automatically generate one using `UUIDs.randomBase64UUID()`. Therefore, if a `DiscoveryNode` object exists, its ephemeral ID is always a non-null string. Should I still add a null check for the `getEphemeralId()`?

Adiiigo: Got it. Then it wouldn't make sense to keep adding null check everywhere. If we are sure that this will never be null and is handled by the constructor before passing it to the other function, we should be good.

server/src/test/java/org/opensearch/cluster/ClusterManagerNodeChangePredicateTests.java
```diff
@@ -0,0 +1,171 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/*
+ * Licensed to Elasticsearch under one or more contributor
```
andrross: The SPDX header is sufficient assuming this is new code that you created

CHANGELOG.md
```diff
@@ -67,6 +67,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Ensure that transient ThreadContext headers with propagators survive restore ([#169373](https://github.com/opensearch-project/OpenSearch/pull/20854))
 - Handle dependencies between analyzers ([#19248](https://github.com/opensearch-project/OpenSearch/pull/19248))
 - Fix `_field_caps` returning empty results and corrupted field names for `disable_objects: true` mappings ([#20800](https://github.com/opensearch-project/OpenSearch/pull/20800))
+- Reduce ClusterState retention in retry closures of TransportClusterManagerNodeAction ([#20858](https://github.com/opensearch-project/OpenSearch/pull/20858))
```
linuxpi: we dont need change log. change is mostly for user facing changes, this is more like an internal improvement or bug fix

server/src/main/java/org/opensearch/cluster/ClusterStateObserver.java
```diff
@@ -117,6 +117,36 @@ public ClusterStateObserver(
...
+    public ClusterStateObserver(
+        String clusterManagerNodeId,
+        long version,
+        ClusterService clusterService,
+        @Nullable TimeValue timeout,
+        Logger logger,
+        ThreadContext contextHolder
+    ) {
+        this(clusterManagerNodeId, version, clusterService.getClusterApplierService(), timeout, logger, contextHolder);
+    }
```
linuxpi: do we need this ctor? could we have directly passed the `clusterApplierService`.

Issue #13923, opened 2024-06-01
[Design] Splitting Shards In-Place


vikasvb90:
### Is your feature request related to a problem? Please describe

# Introduction

In RFC https://github.com/opensearch-project/OpenSearch/issues/12918, we requested feedback on the proposal of splitting shards in-place without any data loss or without the need to stop write traffic. We mentioned two approaches and expressed bias on the physical shard split approach with supporting data points taken from various benchmarks. In this document we will dive deep in the design of physically splitting shards in-place. 

OpenSearch provides capability to split an index into a new index with more number of shards. This is achieved by first blocking write traffic on the index by putting write block on it and then all shards are recovered locally by executing a split [query](https://github.com/opensearch-project/OpenSearch/blob/2.13/server/src/main/java/org/opensearch/index/shard/ShardSplittingQuery.java#L78) on each shard of the source index. High level steps of splitting of a shard in this process are as follows: 

1. Local snapshot is created on the source shard in which the latest commit is acquired .
2. Segments files from this commit are hardlinked into child shard directories.
3. A new IndexWriter is created and opened in CREATE mode.
4. ShardSplittingQuery is executed which essentially executes a delete by query and deletes documents which do not belong to the respective child shard. This is done by passing the predicate from OperationRouting which provides the shard ID for a given doc ID. If this shard Id doesn’t match with the shard ID of child shard then the document is discarded.
5. IndexWriters are committed and child shards are then bootstrapped by creating new translog UUIDs, updating user data on IndexWriter with max seq no., checkpoint, max auto id timestamp, new history UUID, etc.
6. Engine is opened, translog ops are replayed, shard is set to POST_RECOVERY and shard started action is published to master.
7. Assigned shards node receives the update from master and puts the shard in STARTED state.

### Describe the solution you'd like

# High Level Approach

Since shard split is already implemented, we can reuse shard splitting components like ShardSplittingQuery and hardlinking to build in-place shard split. Major areas which need to be covered to build this are: 

**1. Recovering shards while handling write traffic i.e. online recovery of child shards.
2. Filtering out non-belonging operations on child shards during recovery.
3. Changes in operation routing to route docs to right child shards.
4. Handling shard routing changes of splitting primary on split start, during split and post recovery.**

## Online Recovery of Child Shards

Note: We will leaving out details of some of the actions performed during recovery to avoid overloading this doc with too much information. Goal is to provide a general high level sense of how recovery of in-place shard would look like. 

[...]

PR #20859, opened 2026-03-13
Adding metadata to support shard split and to resolve shard ids based on child shard hash

vikasvb90:
### Description
Introduces metadata infrastructure for online (live) shard splitting in OpenSearch.

**ShardRange** - Represents a hash range owned by a child shard, mapping a shard ID to a contiguous range of hash values. Supports range containment checks.

**SplitShardsMetadata** - Tracks the shard split tree within IndexMetadata. Maintains root-to-child shard mappings, in-progress split state, and active shard IDs. Given a document's routing hash, it resolves the correct child shard by walking the split tree. Ensures shard range consistency.

**primaryTermsMap** in IndexMetadata - Adds a Map<Integer, Long> alongside the existing long[] array to support non-contiguous shard IDs that arise after splits. The map becomes the source of truth for primary term look-ups.

**OperationRouting** - Updated to resolve document routing to the correct child shard by delegating to SplitShardsMetadata.getShardIdOfHash().

### Related Issues

This is based on the in-place shard split [design](https://github.com/opensearch-project/OpenSearch/issues/13923).

### Check List


server/src/main/java/org/opensearch/cluster/metadata/SplitShardsMetadata.java
```diff
@@ -0,0 +1,675 @@
...
+        if (maxShardId != that.maxShardId) return false;
+        if (!inProgressSplitShardIds.equals(that.inProgressSplitShardIds)) return false;
+        if (!Arrays.deepEquals(rootShardsToAllChildren, that.rootShardsToAllChildren)) return false;
+        if (parentToChildShards.size() != that.parentToChildShards.size()) return false;
+        for (Integer key : parentToChildShards.keySet()) {
+            if (!Arrays.deepEquals(parentToChildShards.get(key), that.parentToChildShards.get(key))) {
+                return false;
+            }
+        }
+        return true;
```
andrross, line 551: I think you're missing `activeShardIds`

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
```diff
@@ -1066,6 +1068,7 @@ public Iterator<Setting<?>> settings() {
     private final long aliasesVersion;
 
     private final long[] primaryTerms;
+    private final Map<Integer, Long> primaryTermsMap;
```
andrross: Keeping both structures, and making sure they stay in sync, is a maintenance burden. Is it possible to only have the `primaryTermsMap` and construct the `long[]` variant on-demand when needed for serde with pre 3.6 versions?

server/src/main/java/org/opensearch/cluster/metadata/ShardRange.java
```diff
@@ -0,0 +1,134 @@
...
+
+import java.io.IOException;
+
+/**
+ * Represents the hash range assigned to a shard.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class ShardRange implements Comparable<ShardRange>, ToXContentFragment, Writeable {
```
andrross: How about making this a record given all its fields are final primitives?

```java
public record ShardRange(int shardId, int start, int end) implements Comparable<ShardRange>, ToXContentFragment, Writeable
```

server/src/main/java/org/opensearch/cluster/metadata/ShardRange.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public int getEnd() {
+        return end;
+    }
+
+    public ShardRange copy() {
+        return new ShardRange(shardId, start, end);
+    }
+
+    public boolean contains(long hash) {
```
andrross: Why take a `long` given the fields are of type `int`?

server/src/main/java/org/opensearch/cluster/metadata/ShardRange.java
```diff
@@ -0,0 +1,134 @@
...
+
+    public int getStart() {
+        return start;
+    }
+
+    public int getEnd() {
+        return end;
+    }
+
+    public ShardRange copy() {
```
andrross: This class is immutable. There should be no need to ever make a copy.

server/src/main/java/org/opensearch/cluster/metadata/SplitShardsMetadata.java
vikasvb90: Thanks @andrross  for pointing! Fixing it!

server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
vikasvb90: You are right! In the original code I wrote in my branch (6 months back) that was the plan but somehow primitive map got creeped in along the way. Removing it. Thanks!

server/src/main/java/org/opensearch/cluster/metadata/ShardRange.java
vikasvb90: Yes, makes sense!

vikasvb90:
Thanks @andrross for the review! Addressed your comments. Can you please take a look again?

PR #20871, opened 2026-03-14
Tolerate BWC branch patch version bump

andrross:
When a patch release (e.g., 2.19.5) is published and the release branch is bumped to the next patch (2.19.6), BWC tests on main fail because Version.java still references the old patch version.  This causes all in-flight PRs to fail until Version.java is updated.

This change relaxes the logic so that BWC tests will still pass if the checked out code uses a patch version one greater than expected. This prevents CI failures every time a release branch increments its patch version, but still prevents the main branch from drifting by more than one patch version.

I tested this manually with a local 3.5 branch. Works as expected when I bump 3.5.0 to 3.5.1, but fails with 3.5.2.

### Check List

### DIFF
--- a/buildSrc/src/main/java/org/opensearch/gradle/internal/InternalDistributionBwcSetupPlugin.java
+++ b/buildSrc/src/main/java/org/opensearch/gradle/internal/InternalDistributionBwcSetupPlugin.java
@@ -106,7 +106,7 @@ public class InternalDistributionBwcSetupPlugin implements Plugin<Project> {
                 bwcVersion,
                 distributionProject.name,
                 distributionProject.getProjectPath(),
-                distributionProject.getDistFile(),
+                distributionProject,
                 buildBwcTaskProvider
             );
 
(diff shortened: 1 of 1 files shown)


andrross:
@reta Curious what you think about this. Is there any downside? It would make it much smoother to push changes like #20867 without causing all inflight PRs to break and need a rebase.

reta:
> @reta Curious what you think about this. Is there any downside? It would make it much smoother to push changes like #20867 without causing all inflight PRs to break and need a rebase.

Thanks @andrross , I think this is pretty clever idea, the only issue I see here is that we may let some changes that require version specific logic to slip in, a hypothetical (but real) example:
 - change uses `Version.onOrAfter(2.19.4)`
 - meanwhile, we bump to 2.19.5 (now the pull request would need rebase/merge, bwc fails)
 - with relaxed BWC, the change will go to 2.19.5 under assumption it is still 2.19.4

I certainly see the benefits, the alternative I see (which could be a bit easier / straightforward to implement), is to have some sort of automerge (like dependabot does): on version bump, merge the active pull requests with target branch. 

In any case, certainly +1 to try it out, would love to have an option to cap it by time: relax BWC version fallback to one week, hard fail after (although that would require more work for doubtful benefits).

andrross:
FYI @reta, didn't have much success with this approach. It turns out FullClusterRestart has some test assertions that fail on a patch mismatch: https://build.ci.opensearch.org/job/gradle-check/72679/testReport/

PR #20873, opened 2026-03-15
Add ingest pipeline support for pull-based ingestion

imRishN:
### Description
Adds final_pipeline execution support to the pull-based ingestion path. Documents are transformed by configured ingest pipelines before being written to Lucene.

- Pipeline resolution from index settings with dynamic update
- CompletableFuture sync bridge for async IngestService.executeBulkRequest()
- Guardrails blocking _id and _routing mutations 
- Zero overhead when no pipeline is configured
- Pipeline not invoked for delete operations

### Related Issues
Resolves - 
1. https://github.com/opensearch-project/OpenSearch/issues/20875
2. https://github.com/opensearch-project/OpenSearch/issues/20879
3. https://github.com/opensearch-project/OpenSearch/issues/20880

### Check List


server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,189 @@
...
+     */
+    IngestPipelineExecutor(IngestService ingestService, String index, @Nullable String finalPipeline) {
+        this.ingestService = Objects.requireNonNull(ingestService);
+        this.index = Objects.requireNonNull(index);
+        this.resolvedFinalPipeline = finalPipeline;
+        this.pipelinesResolved = true;
+    }
+
+    /**
+     * Resolves pipeline names from index settings. Called lazily on first document and cached.
```
varunbharadwaj: Are we resolving the pipelines lazily on first document? It seems like we only resolve when the MessageProcessor is initialized? I feel what we do now (resolve on initialization) is better.

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,189 @@
...
+            if (e != null) {
+                future.completeExceptionally(e);
+            } else {
+                future.complete(null);
+            }
+        }, slot -> dropped.set(true), ThreadPool.Names.WRITE);
+
+        // Block until pipeline execution completes (with timeout)
+        try {
+            future.get(PIPELINE_EXECUTION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
```
varunbharadwaj, line 141: I'm wondering if it would be better to add synchronous execution support in IngestService, something like 

```
executePipelineSync(..) {
  CountDownLatch latch = new CountDownLatch(1);
  // execute the pipeline (ex: innerExecute(..))
  latch.await()
}
```
If this is possible, we could possibly execute the pipelines on the same thread avoiding the thread handoff. For async pipelines, it would still continue to wait for the result to be available.

What do you think? Have we already explored this path and run into any other challenges?

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,189 @@
...
+     * Also registers a dynamic settings listener for final_pipeline updates.
+     */
+    void resolvePipelineNames(IndexSettings indexSettings) {
+        if (pipelinesResolved) {
+            return;
+        }
+        updateFinalPipeline(IndexSettings.FINAL_PIPELINE.get(indexSettings.getSettings()));
+
+        // Register dynamic settings listener for final_pipeline updates
+        indexSettings.getScopedSettings().addSettingsUpdateConsumer(IndexSettings.FINAL_PIPELINE, this::updateFinalPipeline);
```
varunbharadwaj: Should we register the listener in the constructor before `resolvePipelineNames` can be called so we don't miss any update?

server/src/main/java/org/opensearch/indices/pollingingest/MessageProcessorRunnable.java
```diff
@@ -116,28 +120,36 @@ public MessageProcessorRunnable(
...
         private final String index;
+        private final IngestPipelineExecutor pipelineExecutor;
 
-        MessageProcessor(IngestionEngine engine) {
-            this(engine, engine.config().getIndexSettings().getIndex().getName());
+        MessageProcessor(IngestionEngine engine, IngestService ingestService) {
+            String indexName = engine.config().getIndexSettings().getIndex().getName();
+            this.engine = engine;
+            this.index = indexName;
+            this.pipelineExecutor = new IngestPipelineExecutor(ingestService, indexName);
```
varunbharadwaj: Is it possible to create a single instance of IngestPipelineExecutor at the IngestionEngineFactory layer and pass it all the way through? Thinking if it can help us avoid duplicate settings update consumer registration across writer threads.

server/src/main/java/org/opensearch/indices/pollingingest/MessageProcessorRunnable.java
```diff
@@ -116,28 +120,36 @@ public MessageProcessorRunnable(
...
+            this.pipelineExecutor.resolvePipelineNames(engine.config().getIndexSettings());
```
varunbharadwaj: Is it possible to call resolvePipelineNames inside the IngestPipelineExecutor constructor instead of exposing it outside?

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * Handles ingest pipeline resolution and execution for pull-based ingestion.
+ *
+ * <p>Resolves configured pipelines from index settings (lazy, cached) and executes them
+ * synchronously by bridging IngestService's async callback API with CompletableFuture.
+ * Only {@code final_pipeline} is supported.
+ */
+public class IngestPipelineExecutor {
```
varunbharadwaj, line 40: We can highlight in the javadocs that ingest pipeline/processors on pull-based ingestion flow does not require INGEST roles and executes the transformations on the current node (request is not forwarded to ingest nodes).

server/src/main/java/org/opensearch/indices/pollingingest/IngestionErrorStrategy.java
```diff
@@ -67,7 +67,8 @@ public static ErrorStrategy parseFromString(String errorStrategy) {
     @PublicApi(since = "3.6.0")
     enum ErrorStage {
         POLLING,
-        PROCESSING
+        PROCESSING,
+        PIPELINE
```
varunbharadwaj: Are we using the new error stage anywhere? PIPELINE will be a part of the PROCESSING stage, maybe we can avoid the new error stage?

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Yeah fixing the comment

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Yes, good point. This also helped to remove some cluttered code.

server/src/main/java/org/opensearch/indices/pollingingest/MessageProcessorRunnable.java
imRishN: Good suggestion. Refactored to create a single IngestPipelineExecutor instance in IngestionEngine and pass it through the chain instead of IngestService. Now this has also eliminated duplicate listener registerations. 

Also, as we discussed is prior PRs, this has in a way cleaned up the wiring and intermediate layers no longer know about IngestService at all. The wiring looks like this now:

```
IngestionEngine (creates single IngestPipelineExecutor) → DefaultStreamPoller (passes executor through) → PartitionedBlockingQueueContainer (passes executor through) → N × MessageProcessorRunnable (all share the same executor)
```

IngestService is now only referenced in IngestionEngine and IngestPipelineExecutor itself. The rest of the pull-based path is decoupled from it.

server/src/main/java/org/opensearch/indices/pollingingest/MessageProcessorRunnable.java
imRishN: Done. Resolution is now fully encapsulated in the constructor

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Yeah missed adding that. Added now

server/src/main/java/org/opensearch/indices/pollingingest/IngestionErrorStrategy.java
imRishN: Added this for fine grained tracking, but removing for now in context of this PR. Will review if this is needed later

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Hmm, executing on same thread would avoid the handoff overhead. But there are a few things I considered - 
1. `IngestService` is a core class used by all push based indexing. Adding a sync execution might modify a stable interface and would need deeper review which could be beyond the scope of this PR. Additionally, `runBulkRequestInBatch()` handles batching, metrics tracking, pipeline chaining, index change detection, and slot management. While we can expose a sync path through all of that but looks non trivial. Wdyt?
2. For async processors, we'd still need a latch/future to block for those cases. The internal `Pipeline.execute()` ->  `IngestDocument.executePipeline()` chain is fundamentally callback-based
3. And there seem to be no practical impact for most processors(low weight simpler ones) as execution time dominates the context switch cost

Even with above nuances, this could be a valid optimization and can be taken up as a follow up when we benchmark our changes. Can create a tracking issue for this. Let me know your thoughts

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,172 @@
...
+ * <p>Unlike push-based indexing, pipeline execution in pull-based ingestion does not require the
+ * node to have the {@code ingest} role. Transformations are executed locally on the node hosting the
+ * shard, and requests are not forwarded to dedicated ingest nodes.
+ */
+public class IngestPipelineExecutor {
+
+    private static final Logger logger = LogManager.getLogger(IngestPipelineExecutor.class);
+
+    // TODO: consider making this configurable via index settings if use cases with slow processors arise
+    static final long PIPELINE_EXECUTION_TIMEOUT_SECONDS = 30;
```
monika48, line 45: +1 to making this configurable in the future.
Also what is this default timeout for pipeline executor in push based ingestion?

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,172 @@
...
+     * Resolves the final pipeline from index settings and registers a dynamic settings listener.
+     *
+     * @param ingestService the ingest service for pipeline execution
+     * @param index the index name
+     * @param indexSettings the index settings to resolve a pipeline from and register listener on
+     */
+    public IngestPipelineExecutor(IngestService ingestService, String index, IndexSettings indexSettings) {
+        this.ingestService = Objects.requireNonNull(ingestService);
+        this.index = Objects.requireNonNull(index);
+        indexSettings.getScopedSettings().addSettingsUpdateConsumer(IndexSettings.FINAL_PIPELINE, this::updateFinalPipeline);
```
monika48, line 65: Can indexSetting be null here?

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Executes final_pipeline on the source map synchronously using CompletableFuture to bridge
+     * IngestService's async callback API.
+     *
+     * @param id document ID
+     * @param sourceMap source map to transform
+     * @return the transformed source map, or null if the document was dropped by the pipeline
+     * @throws Exception if pipeline execution fails
+     */
+    public Map<String, Object> executePipelines(String id, Map<String, Object> sourceMap) throws Exception {
```
monika48, line 103: Can we add metrics for: pipeline success/failure, execution time, document drop, timeout

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,172 @@
...
+            (slot, e) -> future.completeExceptionally(e),
+            (thread, e) -> {
+                if (e != null) {
+                    future.completeExceptionally(e);
+                } else {
+                    future.complete(null);
+                }
+            },
+            slot -> dropped.set(true),
+            ThreadPool.Names.WRITE
```
monika48, line 136: Should we consider using a dedicated thread pool for pipeline execution instead of WRITE pool?
Some benefits of dedicated thread pool:
1. dedicated pool would prevent slow pipelines from blocking indexing operations
pipeline failures won't impact write operations  
2. pipeline processing might need different characteristics (more threads for cpu intensive processors, smaller queue for faster feedback)
3. Independent backpressure control for pipeline processing

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Already tracking this in meta here - https://github.com/opensearch-project/OpenSearch/issues/20885

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: No, the only caller passes engineConfig.getIndexSettings() which is always non-null.

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Push based ingestion has no timeout as IngestService.executeBulkRequest() runs to completion with no time bound. The 30s timeout is specific to our sync bridge and acts as a safety valve against hung pipelines. 
+1 to making it configurable as a follow-up.

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Hmm, the WRITE pool on a pull based only node is effectively idle and push based index()/update()/bulk() operations are rejected by IngestionEngine. So there's no contention in the dedicated pull based node setup.

On mixed nodes (pull based + push based indices on the same node), there could be contention. However, the WRITE pool has a queue depth of 10,000 making starvation unlikely in practice. Additionally, a dedicated thread pool would add configuration complexity. 

The longer term optimization (sync pipeline execution on the processor thread, eliminating the thread pool dependency entirely) would be a better investment since it removes the thread handoff altogether. We can add this as a followup

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
varunbharadwaj: I see, makes sense. I haven't checked deeply but here is what I thought could work last time.

Adding the following in IngestService
```
public void executeBulkRequestSync(
    int numberOfActionRequests,
    Iterable<DocWriteRequest<?>> actionRequests,
    BiConsumer<Integer, Exception> onFailure,
    BiConsumer<Thread, Exception> onCompletion,
    IntConsumer onDropped
) {
    runBulkRequestInBatch(numberOfActionRequests, actionRequests, onFailure, onCompletion, onDropped);
}
```
And calling it as

```
CountDownLatch latch = new CountDownLatch(1);
BiConsumer<Integer, Exception> onFailure = (slot, e) -> {
    ...
};

BiConsumer<Thread, Exception> onCompletion = (thread, e) -> {
    ...
    latch.countDown();
};

ingestService.executeBulkRequestSync(
    numberOfActionRequests,
    actionRequests,
    onFailure,
    onCompletion,
    onDropped
);

latch.await(...)
```

But i'm okay to revisit this topic later on in a subsequent PR if that would be easier/simpler, after we get the end to end functionality working as you suggested 👍

varunbharadwaj: We could leave a TODO to explore this later

imRishN: Sure, Lets defer this to a followup. Added a TODO. The approach you suggested looks right and avoids the threadpool dispatch. Will get back to this

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestPipelineFromKafkaIT.java
```diff
@@ -0,0 +1,967 @@
...
+    public void testDynamicPipelineUpdate() throws Exception {
+        createPipeline("pipeline_v1", "{\"processors\": [{\"set\": {\"field\": \"version\", \"value\": \"v1\"}}]}");
+        createPipeline("pipeline_v2", "{\"processors\": [{\"set\": {\"field\": \"version\", \"value\": \"v2\"}}]}");
+
+        // Produce the first batch and create an index with v1 pipeline
+        produceData("1", "Alice", "25");
+        createIndexWithPipeline("pipeline_v1", 1, 0);
+
+        // Wait for the first document to be indexed with v1
+        waitForState(() -> {
```
varunbharadwaj, line 252: It looks like in most of these validations we wait for N docs to be visible, and validate a field. We could extract this out to a util method? The base class already has a `waitForSearchableDocs` util method which is slightly different, but we could have another one specific for the ingest pipeline validations (maybe with a validation logic supplier/callback if that helps)?

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestPipelineFromKafkaIT.java
```diff
@@ -0,0 +1,967 @@
...
+        });
+    }
+
+    // --- Transformation-specific tests ---
+
+    /**
+     * RENAME: Rename a field from one name to another.
+     * Uses built-in "rename" processor.
+     */
+    public void testTransformRename() throws Exception {
```
varunbharadwaj: Is this test the same as `testPipelineRenamesField`? If yes, we can remove this.

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestPipelineFromKafkaIT.java
```diff
@@ -0,0 +1,967 @@
...
+    }
+
+    /**
+     * STRING_TO_JSON: Parse a JSON string field into a structured object.
+     * Uses built-in "json" processor.
+     */
+    public void testTransformStringToJson() throws Exception {
+        createPipeline("string_to_json_pipeline", "{\"processors\": [{\"json\": {\"field\": \"metadata\"}}]}");
+
+        String payload = "{\"_id\":\"1\",\"_source\":{\"name\":\"alice\",\"metadata\":\"{\\\"key\\\":\\\"value\\\",\\\"count\\\":42}\"},"
```
varunbharadwaj, line 453: Is it straightforward to have a util method to generate these payload strings, maybe using XContentBuilder? We can explore this if you feel it improves readability.

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestPipelineFromKafkaIT.java
```diff
@@ -0,0 +1,967 @@
...
+            Object ts = source.get("timestamp");
+            return ts instanceof Number && ((Number) ts).longValue() == 1739459500000L;
+        });
+    }
+
+    /**
+     * TYPE_CONVERSION: Convert a string field to integer.
+     * Uses built-in "convert" processor with type "integer".
+     */
+    public void testTransformTypeConversion() throws Exception {
```
varunbharadwaj, line 474: I'm wondering if we can reduce the number of tests / simplify if we combine similar tests such as testTransformStringToLong and testTransformTypeConversion into a single test, applying all the processors together after the basic validations for the rest. That can possibly reduce the overall number of tests and simplify a bit to test the core ingest pipeline integration and functionality.

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
varunbharadwaj: We can revisit the behavior of push-based flow to see whether we drop/timeout. Because timeouts result in dropping the message for PBI flow causing potential missed updates, which will require manual intervention as discussed. We can consider if removing the timeout is better - which would be the case if we convert to a synchronous execution model.

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * Creates an IngestPipelineExecutor for the given index.
+     * Resolves the final pipeline from index settings and registers a dynamic settings listener.
+     *
+     * @param ingestService the ingest service for pipeline execution
+     * @param index the index name
+     * @param indexSettings the index settings to resolve a pipeline from and register listener on
+     */
+    public IngestPipelineExecutor(IngestService ingestService, String index, IndexSettings indexSettings) {
```
varunbharadwaj, line 62: Can this index be an alias or has to be a concrete index?

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -0,0 +1,175 @@
...
+            throw new RuntimeException("Ingest pipeline execution failed", e.getCause());
+        }
+
+        if (dropped.get()) {
+            return null;
+        }
+
+        // verify _id and _routing have not been mutated
+        if (Objects.equals(originalId, indexRequest.id()) == false) {
+            throw new IllegalStateException(
```
varunbharadwaj, line 157: In MessageProcessorRunnable.hasExhaustedRetries method, we don't list IllegalStateException. Can we check if this results in wasted retries? If yes, we can add IllegalStateException to the list to avoid retry. If we're adding this, we can also make it a list/set of non-retryable exceptions, as the number is growing.

server/src/main/java/org/opensearch/indices/pollingingest/MessageProcessorRunnable.java
```diff
@@ -169,10 +178,12 @@ protected void process(ShardUpdateMessage shardUpdateMessage, MessageProcessorMe
 
         /**
          * Visible for testing. Get the engine operation from the message.
-         * @param shardUpdateMessage an update message containing payload and pointer for the update
+         *
+         * @param shardUpdateMessage      an update message containing payload and pointer for the update
          * @param messageProcessorMetrics message processor metrics
          * @return the message operation
          */
+        @SuppressWarnings("unchecked")
```
varunbharadwaj: Are we adding @SuppressWarnings for the sourceMap? If so can we just add it for that instead of entire method?

```
@SuppressWarnings("unchecked")
Map<String, Object> sourceMap = (Map<String, Object>) payloadMap.get(SOURCE);
```

server/src/test/java/org/opensearch/indices/pollingingest/MessageProcessorTests.java
```diff
@@ -52,7 +60,12 @@ public void setUp() throws Exception {
 
         documentMapper = mock(DocumentMapper.class);
         when(documentMapperForType.getDocumentMapper()).thenReturn(documentMapper);
-        processor = new MessageProcessorRunnable.MessageProcessor(ingestionEngine, "index");
+        // No pipeline configured — executor with null finalPipeline (pre-resolved, skips lazy resolution)
```
varunbharadwaj: We can update the comment as we don't do lazy resolution anymore

imRishN: Removed

server/src/main/java/org/opensearch/indices/pollingingest/MessageProcessorRunnable.java
imRishN: Done

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Done

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Always concrete index name. getIndexSettings().getIndex().getName() always returns the concrete name

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestPipelineFromKafkaIT.java
imRishN: Yes removing

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestPipelineFromKafkaIT.java
imRishN: Merged few tests

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestPipelineFromKafkaIT.java
imRishN: XContentBuilder would add verbosity without improving clarity IMO. Raw json seems fine with readability and represents exact kafka message. wdyt?

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestPipelineFromKafkaIT.java
imRishN: We can extract a helper that encapsulates our checks. But waitForState already serves the purpose with clean inline logic. Seems nice to have, not a must. Shall we skip this?

PR #20874, opened 2026-03-15
Don't consider retetion leases hold by replicas for remote store back…

gbbafna:
…ed indices

### Related Issues

### Issue details 

1. Primary was on 5d4 . Replica was fd2c
2. Primary relocated to 5d4 to 7ed8f at  17:42:13,292  
3. Replica came up on 5d4  at 17:42:22 . 
4. Retention lease of 5d4 was never removed on the new primary and the seq no didn't increase it ever. 

```
"seq_no":{"max_seq_no":15249281,"local_checkpoint":-1,"global_checkpoint":-1},
"retention_leases":{"primary_term":1,"version":344724,"leases":[{"id":"peer_recovery/nodeA","retaining_seq_no":0,"timestamp":1773411457159,"source":"peer recovery"},{"id":"peer_recovery/nodeB","retaining_seq_no":15250120,"timestamp":1773411457159,"source":"peer recovery"}]}
```

This seq_no=0 retention lease led to shard size growing huge , as deletes never got expunged . 

### Check List

### DIFF
--- a/server/src/internalClusterTest/java/org/opensearch/remotestore/RemoteIndexPrimaryRelocationIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/remotestore/RemoteIndexPrimaryRelocationIT.java
@@ -75,7 +75,7 @@ public class RemoteIndexPrimaryRelocationIT extends IndexPrimaryRelocationIT {
         internalCluster().startDataOnlyNode();
         final Settings settings = Settings.builder()
             .put(indexSettings())
-            .put(IndexService.RETENTION_LEASE_SYNC_INTERVAL_SETTING.getKey(), "100ms")
+            .put(IndexService.RETENTION_LEASE_SYNC_INTERVAL_SETTING.getKey(), "5s")
             .build();
         createIndex("test", settings);
         ensureGreen("test");


server/src/main/java/org/opensearch/index/seqno/ReplicationTracker.java
```diff
@@ -285,10 +285,15 @@ public synchronized Tuple<Boolean, RetentionLeases> getRetentionLeases(final boo
         // the primary calculates the non-expired retention leases and syncs them to replicas
         final long currentTimeMillis = currentTimeMillisSupplier.getAsLong();
         final long retentionLeaseMillis = indexSettings.getRetentionLeaseMillis();
-        final Set<String> leaseIdsForCurrentPeers = routingTable.assignedShards()
-            .stream()
-            .map(ReplicationTracker::getPeerRecoveryRetentionLeaseId)
-            .collect(Collectors.toSet());
+        final Set<String> leaseIdsForCurrentPeers;
+        if (indexSettings.isRemoteStoreEnabled()) {
```
shourya035, line 289: Should we also do this for the remote migration path? Also, `routingTable.primaryShard()` might need a null check since Replication Tracker might get initialized even before primary mode is enabled.

gbbafna: While the null check is a valid defensive programming concern, in practice getRetentionLeases is only called when primaryMode is asserted (line 284), meaning a primary shard should always be assigned. 

>Should we also do this for the remote migration path? 

Great point . In remote migration , the replica wouldn't get assigned on the older primary node - as we would assign seg rep copies only .

server/src/internalClusterTest/java/org/opensearch/remotestore/RemoteIndexPrimaryRelocationIT.java
```diff
@@ -48,4 +69,85 @@ public void testPrimaryRelocationWhileIndexing() throws Exception {
         internalCluster().startClusterManagerOnlyNode();
         super.testPrimaryRelocationWhileIndexing();
     }
+
+    public void testRetentionLeaseAfterPrimaryRelocation() throws Exception {
+        internalCluster().startClusterManagerOnlyNode();
+        internalCluster().startDataOnlyNode();
+        final Settings settings = Settings.builder()
+            .put(indexSettings())
+            .put(IndexService.RETENTION_LEASE_SYNC_INTERVAL_SETTING.getKey(), "100ms")
```
shwetathareja: lease sync interval is set to 100ms , how would it create the same race where node hosting the primary, gets replica now and lease has not expired

gbbafna: With 100ms , the race is not reproduced all the time . Let me increase it to 5 second to repro it all the times. 

Sample failure w/o code changes  - 

```
[2026-03-16T12:44:11,385][INFO ][o.o.i.s.ReplicationTracker] [node_t2] [test][0] no retention leases are expired from current retention leases [RetentionLeases{primaryTerm=1, version=6, leases={peer_recovery/0Ej3oXZIQbqc4Bx6kpXw5g=RetentionLease{id='peer_recovery/0Ej3oXZIQbqc4Bx6kpXw5g', retainingSequenceNumber=1, timestamp=1773645241079, source='peer recovery'}, peer_recovery/BRcgYY3IT2SugOlYBea9BA=RetentionLease{id='peer_recovery/BRcgYY3IT2SugOlYBea9BA', retainingSequenceNumber=1, timestamp=1773645241079, source='peer recovery'}}}]


java.lang.AssertionError: expected exactly one peer recovery retention lease but got [RetentionLease{id='peer_recovery/0Ej3oXZIQbqc4Bx6kpXw5g', retainingSequenceNumber=1, timestamp=1773645241079, source='peer recovery'}, RetentionLease{id='peer_recovery/BRcgYY3IT2SugOlYBea9BA', retainingSequenceNumber=1, timestamp=1773645241079, source='peer recovery'}] expected:<1> but was:<2>

```

PR #20892, opened 2026-03-17
Switch executeQueryPhaseAsync and streamNext from block_on to spawn

aravindsagar:
### Description
Switch `executeQueryPhaseAsync` and `streamNext` from `block_on` to `spawn`.

This makes the JVM thread spawn tokio tasks, and returns immediately after submitting work. This will also help with task cancellation and better collection of query metrics.

Panics are caught using `catch_unwind` and surfaced as errors to the listener.

The changes caused `ThreadLeakError` in org.opensearch.datafusion.DataFusionReaderManagerTests. To prevent this, in `runtime_manager.rs`, split `io_runtime` into a `Handle` (for sharing) and `Mutex<Option<Runtime>>` (for owned shutdown). `shutdown()` now calls `io_runtime.shutdown_timeout(10s)` after `cpu_executor.join_blocking()`, blocking until all datafusion-io-* threads have fully terminated.

### Related Issues

### Check List

### DIFF
--- a/plugins/engine-datafusion/jni/src/lib.rs
+++ b/plugins/engine-datafusion/jni/src/lib.rs
@@ -687,7 +687,7 @@ pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_executeQu
             runtime,
             cpu_executor,
         ),
-        |env, lr, ptr| set_action_listener_ok_global(env, lr, ptr),
+        |env, listener_ref, stream_pointer| set_action_listener_ok_global(env, listener_ref, stream_pointer),
     );
 }
 
@@ -728,7 +728,7 @@ pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_fetchSegm
         "fetchSegmentStats",
         listener_ref,
         async move { util::fetch_segment_statistics(files_meta).await },
-        |env, lr, map| set_action_listener_ok_global_with_map(env, lr, &map),
+        |env, listener_ref, stats_map| set_action_listener_ok_global_with_map(env, listener_ref, &stats_map),
     );
 }
 
@@ -798,7 +798,7 @@ pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_streamNex
                 }
             }
         },
-        |env, lr, ptr| set_action_listener_ok_global(env, lr, ptr),
+        |env, listener_ref, data_pointer| set_action_listener_ok_global(env, listener_ref, data_pointer),
     );
     // Function returns immediately to java - async rust work continues in background
 }


plugins/engine-datafusion/src/test/java/org/opensearch/datafusion/DataFusionReaderManagerTests.java
```diff
@@ -51,7 +53,21 @@
 import static org.opensearch.datafusion.search.cache.CacheSettings.STATISTICS_CACHE_SIZE_LIMIT;
 import static org.opensearch.index.engine.Engine.SearcherScope.INTERNAL;
 
+@ThreadLeakFilters(defaultFilters = true, filters = { DataFusionReaderManagerTests.DataFusionThreadFilter.class })
 public class DataFusionReaderManagerTests extends OpenSearchTestCase {
+
+    /**
+     * Excludes Tokio worker threads from randomizedtesting's thread leak detection.
```
bharath-techie: This is the reason why we switched back to block on. When the tests finish running , the tests should ensure all threads are cleaned up.

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -126,6 +133,72 @@ where
...
+        .call_static_method("java/lang/Thread", "currentThread", "()Ljava/lang/Thread;", &[])?
+        .l()?;
+    let jname = env.new_string(name)?;
+    env.call_method(&thread, "setName", "(Ljava/lang/String;)V", &[(&jname).into()])?;
+    Ok(())
+}
+
+/// Extract a human-readable message from a panic payload.
+fn panic_message(payload: &Box<dyn std::any::Any + Send>) -> String {
+    if let Some(s) = payload.downcast_ref::<&str>() {
```
alchemist51, line 134: Can we use match construct here?

aravindsagar: Not sure if `match` works well here, as each `downcast_ref` call returns a different type

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -128,6 +129,59 @@ where
...
+    listener_ref: GlobalRef,
+    task: Fut,
+    on_ok: FOk,
+)
+where
+    Fut: Future<Output = Result<T, DataFusionError>> + Send + 'static,
+    T: Send + 'static,
+    FOk: FnOnce(&mut JNIEnv, &GlobalRef, T) + Send + 'static,
+{
+    let _ = runtime.spawn(async move {
```
Bukhtawar, line 165: Do we need to maintain this handle in order to abort on task cancellation. I am assuming we will be wrapping this with a task manager that would handle this?

bharath-techie: isn't this ignored rather ?

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -630,22 +686,9 @@ pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_executeQu
...
-            Err(e) => {
-                with_jni_env(|env| {
-                    log_error!("Query execution failed: {}", e);
-                    set_action_listener_error_global(env, &listener_ref, &e);
-                });
-            }
-        }
-    });
+        ),
+        |env, lr, ptr| set_action_listener_ok_global(env, lr, ptr),
```
bharath-techie: This change has made the code less readable.

aravindsagar: I agree. The intention behind the spawn_jni_task abstraction is to make sure that panics are caught and propagated as errors in tasks spawned io_runtime - else the Java side listener may never be called. We could just add catch_unwind everywhere we spawn tasks in io_runtime - but I thought it maybe easy for someone to miss adding this.

Do you have any suggestions for handling this better?

aravindsagar:
Ran some benchmarks with OSB (clickbench workload with `datafusion-ppl` procedure) with different number of search clients against OpenSearch running on a 16-core machine.

Sample OSB invocation:
```
$ docker run --network host \
-v $HOME/benchmarks:/opensearch-benchmark/.benchmark \
-v $HOME/workloads:/opensearch-benchmark/workloads \
opensearchproject/opensearch-benchmark run \
--target-hosts http://localhost:9200 \
--pipeline benchmark-only \
--workload-path=~/workloads/clickbench \
--workload-params="ingest_percentage:5,warmup_iterations:1,test_iterations:5,search_clients:8" \
--test-procedure="datafusion-ppl" \
--exclude-tasks="q20-specific-user,q24-google-urls-sorted,q25-search-phrases-by-time,q26-search-phrases-sorted,q27-search-phrases-multi-sort"
```

Ran the benchmarks against `feature/datafusion` branch, as well as the changes in this PR. For both setups, ran for multiple values of search_clients: 1, 8, 16, 32 and 64. 

Found no significant differences between the 2 approaches (spawn vs block_on). spawn was slightly faster with low values of search_clients, but the difference was practically nil at 64

8 clients: spawn 4% faster
16 clients: spawn 3.5% faster
32 clients: spawn 1.2% faster
64 clients: spawn 0.8% faster

PR #20897, opened 2026-03-17
Resolve actual BWC version from build output

andrross:
When a BWC branch bumps its patch version before Version.java on main is updated, the tests.old_cluster_version system property passed to BWC tests is wrong. This causes test failures in assertions that compare the expected version against values reported by the running cluster or stored in index metadata.

InternalDistributionBwcSetupPlugin already detects when the BWC branch produces a fallback artifact with a different patch version.  Write the actual version to a file so that downstream test tasks can read it at execution time instead of relying on the version inferred from Version.java on main.

### Check List

### DIFF
--- a/buildSrc/src/main/java/org/opensearch/gradle/internal/InternalDistributionBwcSetupPlugin.java
+++ b/buildSrc/src/main/java/org/opensearch/gradle/internal/InternalDistributionBwcSetupPlugin.java
@@ -50,6 +50,7 @@ import org.gradle.language.base.plugins.LifecycleBasePlugin;
 import javax.inject.Inject;
 
 import java.io.File;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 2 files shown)


cwperks (approved):
TY for fixing this issue @andrross! I encountered this multiple times yesterday. It took a second to grasp that fallbackVersion is actually referencing the next unreleased version 2.19.6. For simplicity sake (without actually diving deeper into the code), have we ever considered updating the assertion so that its ok if `tests.old_cluster_version` is of a patch version >= what the main branch is aware of - i.e. permit both 2.19.5 and 2.19.6 at least for a short period before main becomes aware of 2.19.6.

andrross:
> permit both 2.19.5 and 2.19.6 at least for a short period before main becomes aware of 2.19.6.

@cwperks I did consider this, but I was worried about the case where you're running bwc tests against the previous patch version (I'm not sure if we actually do this in practice?) because then the assertions would be of little value if they would accept both versions under test. Also this approach keeps the logic of accepting a patch version mismatch contained within `InternalDistributionBwcSetupPlugin` and other tests don't need to be aware of anything special.

PR #20901, opened 2026-03-17
Add new setting property 'Sensitive' for tiering dynamic settings

cwperks:
### Description

Companion Security PR: https://github.com/opensearch-project/security/pull/6016

This PR adds a new Setting property called `Sensitive` which is intended to help tier settings for authorization for different callers.

Currently, the security repo has dynamic settings through the security index (see [config.yml](https://github.com/opensearch-project/security/blob/main/config/config.yml)) and still have some static cluster settings in `opensearch.yml`. The security index has had some bwc issues (particularly if a new setting is added from one minor version to the next and its persisted in a mixed cluster during rolling upgrade) and some of these settings would be better suited for using the existing settings mechanism.

In order to use the existing settings mechanism, the security plugin needs a mechanism for distinguishing which settings should be toggleable by a securityadmin vs any user with `cluster:admin/settings/put` permission.

I propose adding a new setting property called `Sensitive` so that the security plugin can check a settings update payload to see if any sensitive settings are being updated. If the payload contains a sensitive setting, then only a security admin (`all_access` or `security_manager` roles) should be able to update, otherwise any user with `cluster:admin/settings/put` should be able to update. 

### Related Issues

Related to:

- https://github.com/opensearch-project/OpenSearch/issues/20905
- https://github.com/opensearch-project/security/issues/5219
- https://github.com/opensearch-project/security/issues/6002

### Check List


finnegancarroll:
If this property is reserved for security relevant settings what do you think about forcing sensitive settings to be persistant as well?

This might prevent users from having unexpected security relevant changes due to a cluster restart, blue green deployment, or a process crash on a single node cluster.

cwperks:
> If this property is reserved for security relevant settings what do you think about forcing sensitive settings to be persistant as well?
> 
> This might prevent users from having unexpected security relevant changes due to a cluster restart, blue green deployment, or a process crash on a single node cluster.

That's a good idea. Let me check what it would entail.

cwperks:
@finnegancarroll I was able to block transient in this PR. Please take a look at the latest changes when you have some time.

finnegancarroll (approved):

server/src/main/java/org/opensearch/common/settings/Setting.java
```diff
@@ -177,7 +177,14 @@ public enum Property {
          * Mark this setting as immutable on snapshot restore
          * i.e. the setting will not be allowed to be removed or modified during restore
          */
-        UnmodifiableOnRestore
+        UnmodifiableOnRestore,
+
+        /**
+         * Mark this dynamic setting as sensitive. Sensitive settings require security admin
```
andrross: @cwperks The thing that is a little strange here is that there is no concept of a "security admin" in core OpenSearch. Is this property a no-op if you're running without the security plugin?

An obvious question is whether its possible for the security plugin itself to annotate settings it deems sensitive in order to keep the logic related to this feature together.

cwperks: Its added here so that other plugins can also use this setting property and automatically make the security plugin aware. You're right that with core itself its a noop, but it makes sense in the context of security and I think is the simplest current method for creating 2 tiers of settings for those that can be designated more sensitive then others.

cwperks: There are other things like this in core that are pertinent to security see [RestRequestFilter](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/rest/RestRequestFilter.java)

cwperks: 3 such settings I plan to mark initially are:

- https://github.com/opensearch-project/security/issues/6015
- https://github.com/opensearch-project/security/issues/6002
- `plugins.ml_commons.allow_registering_model_via_url` - https://docs.opensearch.org/latest/ml-commons-plugin/custom-local-models/

cwperks: ^ as mentioned in the linked issue, security actually has a separate setting mechanism with persistence via the security index but there are some problems (namely around bwc) with this.

I would really like to use the setting mechanism from core, but it would be needed to distinguish between "sensitive" settings vs everything else.

andrross: @cwperks Maybe just change the comment to be a little more clear? Something like:

> Marks a setting as sensitive. The Sensitive property has default enforcement but enables plugins to implement different policies for these settings. In practice the security plugin will require higher privileges for modifying sensitive settings. This property can only be applied to settings that also have `{@link Property#Dynamic}`.

cwperks: Updated the comment in the javadoc

PR #20902, opened 2026-03-17
Fix flaky profiler assertions in terms aggregation YAML tests

andrross:
The YAML REST tests for "string profiler via global ordinals" and "numeric profiler" assert exact values for debug fields in the aggregation profile output. These assertions are non-deterministic when concurrent search is enabled.

With concurrent search, a single segment's doc ID space is partitioned across multiple slices, each with its own aggregator
instance and independent profiling counters. ConcurrentAggregationProfiler.reduceProfileResultsTree() does not merge the debug maps across slices — it simply assigns the last slice's debug info, discarding all others. Which slice is "last" depends on iteration order, making the reported values arbitrary.

For the string profiler test, this causes segments_with_single_valued_ords and segments_with_multi_valued_ords to reflect only one slice's view. For the numeric profiler test, total_buckets reports one slice's bucket ordinal count rather than the true total.

A proper fix would require changing the debug info API so that aggregators can declare a merge strategy (sum, max, last, etc.) for each key. The current API uses an untyped Map<String, Object> with no way to distinguish counters from cardinalities, thresholds, or other numeric values, so there is no safe generic merge.

Relax the assertions to verify the fields exist with plausible values without depending on exact counts.

### Related Issues
Part of #14319

### Check List

### DIFF
--- a/rest-api-spec/src/main/resources/rest-api-spec/test/search.aggregation/20_terms.yml
+++ b/rest-api-spec/src/main/resources/rest-api-spec/test/search.aggregation/20_terms.yml
@@ -1074,8 +1074,8 @@ setup:
   - match: { profile.shards.0.aggregations.0.debug.deferred_aggregators: [ max_number ] }
   - match: { profile.shards.0.aggregations.0.debug.collection_strategy: dense }
   - match: { profile.shards.0.aggregations.0.debug.result_strategy: terms }
-  - gt:    { profile.shards.0.aggregations.0.debug.segments_with_single_valued_ords: 0 }
-  - match: { profile.shards.0.aggregations.0.debug.segments_with_multi_valued_ords: 0 }
+  - gte:   { profile.shards.0.aggregations.0.debug.segments_with_single_valued_ords: 0 }
+  - gte:   { profile.shards.0.aggregations.0.debug.segments_with_multi_valued_ords: 0 }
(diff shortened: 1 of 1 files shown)


andrross:
This is the code that replaces the debug info with the last slice processed in concurrent search:

https://github.com/opensearch-project/OpenSearch/blob/8fd9aebb0ca329c88de1cf74e6a037d5247a3593/server/src/main/java/org/opensearch/search/profile/aggregation/ConcurrentAggregationProfiler.java#L112

Here is an example failure: https://build.ci.opensearch.org/job/gradle-check/74123/testReport/junit/org.opensearch.test.rest/ClientYamlTestSuiteIT/test__p0_search_aggregation_20_terms_numeric_profiler_/

This continues to be a pain point with test flakiness. @mch2 @sohami @jainankitk  Can one of you take a look?

bowenlan-amzn (approved):
It'd make more sense to preserve per-slice debug info instead of trying to merge. Profile API is experimental so we're free to change the shape — debug could become a list under concurrent search. Feels like a good first issue.

jainankitk (approved):
Looks fine for fixing the flaky test. But, I am concerned about the following:

> ConcurrentAggregationProfiler.reduceProfileResultsTree() does not merge the debug maps across slices — it simply assigns the last slice's debug info, discarding all others

Seems like a genuine issue we need to fix in Concurrent Segment Search profiling? Maybe create an issue for the same?

PR #10519, opened 2023-10-10
Replace multipart download with parallel file download

andrross:
There are a few open issues with the multi-stream download approach:
 - Recovery stats are not being reported correctly
 - It is incompatible (short of reopening and re-reading the entire file) with the existing Lucene checksum validation logic
 - There are some issues with integrating it with the pending client side encryption work

Given this, I attempted an experiment where I replaced with multi-stream-within-a-single-file approach with simply parallelizing downloads across files (this is how snapshot restore works). I actually got better results with this approach: recovering a ~52GiB shard took about 4.7 minutes with the multi-stream code versus 3.9 minutes with the parallel file approach (r7g.4xlarge EC2 instance, 500MiB/s EBS volume, S3 as remote repository).

I think this is the right approach as it leverages the more battle-tested code path and addresses the three issues listed above. The multi-stream approach still has promise as it will allow us to download very large files faster (whereas this approach they can be the long poll on the transfer operation). However, given that 5GB segments (made up of multiple files in practice) are the norm, we generally aren't dealing with huge files.

### Related Issues
Resolves #10516

### Check List
- [x] New functionality includes testing.
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff
- [x] Commit changes are listed out in CHANGELOG.md file (See: [Changelog](../blob/main/CONTRIBUTING.md#changelog))
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #20906, opened 2026-03-18
fixed concurrent translog download failover

### LINKED ISSUE #10826: [Remote Translog] Download Translog files concurrently
In failover/primary relocations, we download translog from remote. We need to make it  concurrent (still bounded) on a separate threadpool, may be `translog_transfer` to make that faster . We can use concurrent segment download as a prior work to do the same.

ThyTran1402:
### Description
During failover and primary relocations, OpenSearch downloads all translog generations from remote store before opening the engine. Previously this was done sequentially and one generation at a time in a blocking for-loop inside `RemoteFsTranslog.downloadOnce()`. Since translog generations are independent of each other, this was a pure serial I/O bottleneck that scaled linearly with the number of retained generations.

This change parallelises translog generation downloads using the existing `TRANSLOG_TRANSFER` threadpool (already used for uploads), following the same `CountDownLatch` + worker-queue pattern established for concurrent segment downloads for #10519.

### Related Issues
Resolves  #10826 

### Check List


ThyTran1402:
I think the PR is ready for review.

Issue #20876, opened 2026-03-16
[META] OpenSearch Pluggable Indexing


mgodwan:
### Please describe the end goal of this project

We are planning to introduce components related to enable pluggable indexing https://github.com/opensearch-project/OpenSearch/issues/20644, and multiple composable data-formats in OpenSearch. This issue is to track the work related to the feature.

### Abstractions in Core
- [x] Introduce Lucene agnostic Interfaces (Indexer, CatalogSnapshot, FieldCapability, etc.) https://github.com/opensearch-project/OpenSearch/pull/20675
- [x] Introduce DataFormat Plugin and associated interfaces (DataFormat, IndexingExecutionEngine, etc.) https://github.com/opensearch-project/OpenSearch/pull/20675
- [x] Dataformat Registry https://github.com/opensearch-project/OpenSearch/pull/21039
- [x] Introduce Lucene agnostic Searcher/Reader interfaces 
- [ ] Indexer implementation for data formats along with store management  https://github.com/opensearch-project/OpenSearch/pull/21181/
- [ ] Add Merge Handler and Committer interfaces and orchestration logic https://github.com/opensearch-project/OpenSearch/pull/
- [x] Add Catalog Snapshot and File Manager implementation https://github.com/opensearch-project/OpenSearch/pull/20982
- [x] Add changes to support mappers to work with new interfaces https://github.com/opensearch-project/OpenSearch/pull/21001
- [ ] Add committer implementation github.com/opensearch-project/OpenSearch/pull/21103
- [ ] Add File Deleter to work with committer https://github.com/opensearch-project/OpenSearch/pull/21145
- [ ] Managing Internal/External Refreshes
- [ ] Existing ITs and UTs to wire into new implementation as well.

### Composable Data Format Plugin https://github.com/opensearch-project/OpenSearch/pull/20909
- [x] Multiplexing implementation 
- [x] Add support for Row Id Management for coherence across data formats
- [x] Failure handling across data formats during indexing
- [ ] Refresh orchestration
- [ ] JMH benchmark for merge flow

### Commons for Data formats
- [ ] Writer Pool Management to handle concurrency

### Parquet Data Format Plugin
- [x] Introduce Arrow based implementation to buffer incoming documents https://github.com/opensearch-project/OpenSearch/pull/20930
- [x] arrow-rs based Indexing and native writer management
- [x] JNI Bridge for leveraging arrow-rs
- [x] Field Handlers
[...]

PR #20909, opened 2026-03-18
Add sandbox plugin for composite indexing execution engine

alchemist51:
### Description

This PR introduces the `composite-engine` sandbox plugin that implements the `CompositeIndexingExecutionEngine` — the orchestration layer for multi-format indexing as described in RFC [#20644](https://github.com/opensearch-project/OpenSearch/issues/20644)

The composite engine enables an index to write documents to multiple storage formats (e.g., Lucene + Parquet) simultaneously through a single `IndexingExecutionEngine` interface. Format plugins register via the `ExtensiblePlugin` SPI, and the composite engine delegates writes, refresh, and file management to each per-format engine.

***Note***: we have used ExtensiblePlugin SPI model temporarily. Once we have introduced Dataformat Registry, we should be able to get rid of this model.

### What's included

**New sandbox plugin: `sandbox/plugins/composite-engine`**

- `CompositeEnginePlugin` — `ExtensiblePlugin` entry point that discovers `DataFormatPlugin` implementations at node bootstrap, validates index settings, and creates the composite engine. Registers three index settings:
  - `index.composite.enabled` (default `false`)
  - `index.composite.primary_data_format` (default `"lucene"`)
  - `index.composite.secondary_data_formats` (default `[]`)
- `CompositeIndexingExecutionEngine` — Orchestrates indexing across a primary and zero or more secondary per-format engines. Handles writer creation, refresh (flush all writers → build segments → delegate per-format refresh), file deletion, and document input creation.
- `CompositeDataFormat` — A `DataFormat` wrapper over the constituent formats. Uses `Long.MIN_VALUE` priority so concrete formats take precedence.
- `CompositeDocumentInput` — Broadcasts `addField`, `setRowId`, and other metadata operations to all per-format `DocumentInput` instances. Releases the writer back to the pool on `close()`.
- `CompositeWriter` — Delegates `addDoc`, `flush`, `sync`, and `close` to each per-format writer (primary first, then secondaries). Implements `Lock` for pool checkout semantics.
- `CompositeDataFormatWriterPool` — Thread-safe pool of `CompositeWriter` instances with lock-based checkout/release and a `checkoutAll` for flush.
- `RowIdGenerator` — Generates monotonically increasing row IDs for cross-format document synchronization within a writer's segment scope.

**New sandbox lib: `sandbox/libs/composite-engine-lib`**

- `ConcurrentQueue` — Striped concurrent queue using thread-affinity hashing to reduce contention across concurrent indexing threads.
- `LockableConcurrentQueue` — Extends `ConcurrentQueue` with `tryLock`-based polling so writers can be checked out without blocking.

### How format plugins integrate

Format plugins (e.g., Parquet) extend this plugin by:
1. Declaring `extendedPlugins = ['composite-engine']` in their `build.gradle`
2. Implementing `DataFormatPlugin`
3. The `ExtensiblePlugin` SPI discovers them automatically during node bootstrap

### Related issues

Resolves part of #20876

### Check List
- [x] All classes are annotated with `@ExperimentalApi`
- [x] No BWC tests required (sandbox/experimental)
- [x] Commits are signed off (DCO)

### DIFF
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDocumentInput.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDocumentInput.java
@@ -13,6 +13,7 @@ import org.opensearch.index.engine.dataformat.DataFormat;
 import org.opensearch.index.engine.dataformat.DocumentInput;
 import org.opensearch.index.mapper.MappedFieldType;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -47,7 +48,7 @@ public class CompositeDocumentInput implements DocumentInput<List<? extends Docu
     ) {
         this.primaryFormat = Objects.requireNonNull(primaryFormat, "primaryFormat must not be null");
         this.primaryDocumentInput = Objects.requireNonNull(primaryDocumentInput, "primaryDocumentInput must not be null");
-        this.secondaryDocumentInputs = Map.copyOf(
+        this.secondaryDocumentInputs = Collections.unmodifiableMap(
             Objects.requireNonNull(secondaryDocumentInputs, "secondaryDocumentInputs must not be null")
         );
     }
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
@@ -31,7 +31,7 @@ import org.opensearch.index.shard.ShardPath;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.LinkedHashMap;
+import java.util.IdentityHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -275,7 +275,7 @@ public class CompositeIndexingExecutionEngine implements IndexingExecutionEngine
     @Override
     public CompositeDocumentInput newDocumentInput() {
         DocumentInput<?> primaryInput = primaryEngine.newDocumentInput();
-        Map<DataFormat, DocumentInput<?>> secondaryInputMap = new LinkedHashMap<>();
+        Map<DataFormat, DocumentInput<?>> secondaryInputMap = new IdentityHashMap<>();
         for (IndexingExecutionEngine<?, ?> engine : secondaryEngines) {
             secondaryInputMap.put(engine.getDataFormat(), engine.newDocumentInput());
         }
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
@@ -21,8 +21,8 @@ import org.opensearch.index.engine.dataformat.Writer;
 import org.opensearch.index.engine.exec.WriterFileSet;
 
 import java.io.IOException;
-import java.util.AbstractMap;
-import java.util.LinkedHashMap;
+import java.util.Collections;
+import java.util.IdentityHashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
@@ -44,7 +44,8 @@ class CompositeWriter implements Writer<CompositeDocumentInput>, Lockable {
 
     private static final Logger logger = LogManager.getLogger(CompositeWriter.class);
 
-    private final Map.Entry<DataFormat, Writer<DocumentInput<?>>> primaryWriter;
+    private final DataFormat primaryFormat;
+    private final Writer<DocumentInput<?>> primaryWriter;
     private final Map<DataFormat, Writer<DocumentInput<?>>> secondaryWritersByFormat;
     private final ReentrantLock lock;
     private final long writerGeneration;
@@ -87,16 +88,14 @@ class CompositeWriter implements Writer<CompositeDocumentInput>, Lockable {
         this.writerGeneration = writerGeneration;
 
         IndexingExecutionEngine<?, ?> primaryDelegate = engine.getPrimaryDelegate();
-        this.primaryWriter = new AbstractMap.SimpleImmutableEntry<>(
-            primaryDelegate.getDataFormat(),
-            (Writer<DocumentInput<?>>) primaryDelegate.createWriter(writerGeneration)
-        );
+        this.primaryFormat = primaryDelegate.getDataFormat();
+        this.primaryWriter = (Writer<DocumentInput<?>>) primaryDelegate.createWriter(writerGeneration);
 
-        Map<DataFormat, Writer<DocumentInput<?>>> secondaries = new LinkedHashMap<>();
+        Map<DataFormat, Writer<DocumentInput<?>>> secondaries = new IdentityHashMap<>();
         for (IndexingExecutionEngine<?, ?> delegate : engine.getSecondaryDelegates()) {
             secondaries.put(delegate.getDataFormat(), (Writer<DocumentInput<?>>) delegate.createWriter(writerGeneration));
         }
-        this.secondaryWritersByFormat = Map.copyOf(secondaries);
+        this.secondaryWritersByFormat = Collections.unmodifiableMap(secondaries);
         this.rowIdGenerator = new RowIdGenerator(CompositeWriter.class.getName());
     }
 
@@ -106,11 +105,11 @@ class CompositeWriter implements Writer<CompositeDocumentInput>, Lockable {
             throw new IllegalStateException("Cannot add document to writer in state " + state.get());
         }
         // Write to primary first
-        WriteResult primaryResult = primaryWriter.getValue().addDoc(doc.getPrimaryInput());
+        WriteResult primaryResult = primaryWriter.addDoc(doc.getPrimaryInput());
         switch (primaryResult) {
-            case WriteResult.Success s -> logger.trace("Successfully added document in primary format [{}]", primaryWriter.getKey().name());
+            case WriteResult.Success s -> logger.trace("Successfully added document in primary format [{}]", primaryFormat.name());
             case WriteResult.Failure f -> {
-                logger.debug("Failed to add document in primary format [{}]", primaryWriter.getKey().name());
+                logger.debug("Failed to add document in primary format [{}]", primaryFormat.name());
                 return primaryResult;
             }
         }
@@ -141,8 +140,8 @@ class CompositeWriter implements Writer<CompositeDocumentInput>, Lockable {
     public FileInfos flush() throws IOException {
         FileInfos.Builder builder = FileInfos.builder();
         // Flush primary
-        Optional<WriterFileSet> primaryWfs = primaryWriter.getValue().flush().getWriterFileSet(primaryWriter.getKey());
-        primaryWfs.ifPresent(writerFileSet -> builder.putWriterFileSet(primaryWriter.getKey(), writerFileSet));
+        Optional<WriterFileSet> primaryWfs = primaryWriter.flush().getWriterFileSet(primaryFormat);
+        primaryWfs.ifPresent(writerFileSet -> builder.putWriterFileSet(primaryFormat, writerFileSet));
         // Flush secondaries
         for (Writer<DocumentInput<?>> writer : secondaryWritersByFormat.values()) {
             FileInfos fileInfos = writer.flush();
@@ -156,7 +155,7 @@ class CompositeWriter implements Writer<CompositeDocumentInput>, Lockable {
 
     @Override
     public void sync() throws IOException {
-        primaryWriter.getValue().sync();
+        primaryWriter.sync();
         for (Writer<DocumentInput<?>> writer : secondaryWritersByFormat.values()) {
             writer.sync();
         }
@@ -164,7 +163,7 @@ class CompositeWriter implements Writer<CompositeDocumentInput>, Lockable {
 
     @Override
     public void close() throws IOException {
-        primaryWriter.getValue().close();
+        primaryWriter.close();
         for (Writer<DocumentInput<?>> writer : secondaryWritersByFormat.values()) {
             writer.close();
         }
--- a/libs/concurrent-queue/src/jmh/java/org/opensearch/common/queue/LockablePoolBenchmark.java
+++ b/libs/concurrent-queue/src/jmh/java/org/opensearch/common/queue/LockablePoolBenchmark.java
@@ -1,0 +1,253 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.common.queue;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Group;
+import org.openjdk.jmh.annotations.GroupThreads;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.locks.LockSupport;
+import java.util.concurrent.locks.ReentrantLock;
+
+/**
+ * JMH benchmark for {@link LockablePool} measuring:
+ * <ul>
+ *   <li>Isolated checkout/return throughput at varying thread counts</li>
+ *   <li>Mixed workload: concurrent writers + periodic checkoutAll (refresh)</li>
+ *   <li>Aggressive refresh: frequent checkoutAll to stress Phase 3 bulk cleanup</li>
+ *   <li>Writer latency during refresh contention (sample mode)</li>
+ * </ul>
+ */
+@Fork(2)
+@Warmup(iterations = 2, time = 3)
+@Measurement(iterations = 3, time = 5)
+@State(Scope.Benchmark)
+@SuppressWarnings("unused")
+public class LockablePoolBenchmark {
+
+    @Param({ "4", "8" })
+    int concurrency;
+
+    private LockablePool<PoolEntry> pool;
+
+    @Setup(Level.Iteration)
+    public void setup() {
+        AtomicInteger counter = new AtomicInteger(0);
+        pool = new LockablePool<>(() -> new PoolEntry(counter.getAndIncrement()), LinkedList::new, concurrency);
+        // Pre-warm the pool with more entries to populate multiple stripes
+        for (int i = 0; i < concurrency * 4; i++) {
+            PoolEntry e = pool.getAndLock();
+            pool.releaseAndUnlock(e);
+        }
+    }
+
+    // ── Mixed workload: writers + periodic refresh (1s interval) ──
+
+    @Benchmark
+    @Group("mixed_7w_1r")
+    @GroupThreads(7)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MICROSECONDS)
+    public void writers_7w1r(Blackhole bh) {
+        PoolEntry e = pool.getAndLock();
+        bh.consume(simulateWork(e));
+        pool.releaseAndUnlock(e);
+    }
+
+    @Benchmark
+    @Group("mixed_7w_1r")
+    @GroupThreads(1)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MICROSECONDS)
+    public List<PoolEntry> refresh_7w1r() throws InterruptedException {
+        Thread.sleep(1000);
+        return pool.checkoutAll();
+    }
+
+    @Benchmark
+    @Group("mixed_3w_1r")
+    @GroupThreads(3)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MICROSECONDS)
+    public void writers_3w1r(Blackhole bh) {
+        PoolEntry e = pool.getAndLock();
+        bh.consume(simulateWork(e));
+        pool.releaseAndUnlock(e);
+    }
+
+    @Benchmark
+    @Group("mixed_3w_1r")
+    @GroupThreads(1)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MICROSECONDS)
+    public List<PoolEntry> refresh_3w1r() throws InterruptedException {
+        Thread.sleep(1000);
+        return pool.checkoutAll();
+    }
+
+    // ── Aggressive refresh: 10ms interval to stress checkoutAll Phase 3 ──
+    // This makes checkoutAll fire ~100x/sec instead of 1x/sec, amplifying
+    // the difference between per-item remove (old) and bulk removeIf (new).
+
+    @Benchmark
+    @Group("aggressive_7w_1r")
+    @GroupThreads(7)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MICROSECONDS)
+    public void writers_aggressive_7w1r(Blackhole bh) {
+        PoolEntry e = pool.getAndLock();
+        bh.consume(simulateWork(e));
+        pool.releaseAndUnlock(e);
+    }
+
+    @Benchmark
+    @Group("aggressive_7w_1r")
+    @GroupThreads(1)
+    @BenchmarkMode(Mode.Throughput)
+    @OutputTimeUnit(TimeUnit.MICROSECONDS)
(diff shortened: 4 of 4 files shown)


sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
```diff
@@ -0,0 +1,223 @@
...
+        switch (primaryResult) {
+            case WriteResult.Success s -> logger.trace("Successfully added document in primary format [{}]", primaryWriter.getKey().name());
+            case WriteResult.Failure f -> {
+                logger.debug("Failed to add document in primary format [{}]", primaryWriter.getKey().name());
+                return primaryResult;
+            }
+        }
+
+        // Then write to each secondary
+        List<DocumentInput<?>> secondaryInputs = new ArrayList<>(doc.getSecondaryInputs().values());
```
RS146BIJAY: Just wondering if there is a way to associate which input is related to which writers. I think we are assuming here that inputs and writers will be in the same order.

alchemist51: Got it, that should be easy. We can maintain the map like how we maintain dataformat to writer map. We can then use the Dataformat as the common key to find the association.

Will update the PR.

alchemist51: Addressed it.

libs/concurrent-queue/src/main/java/org/opensearch/common/queue/ConcurrentQueue.java
```diff
@@ -0,0 +1,140 @@
...
+                }
+            } finally {
+                lock.unlock();
+            }
+        }
+        return null;
+    }
+
+    boolean remove(T entry) {
+        for (int i = 0; i < concurrency; ++i) {
```
mgodwan, line 126: Can we avoid going through all the entries here by maintaining an auxiliary data structure?

libs/concurrent-queue/src/main/java/org/opensearch/common/queue/LockableConcurrentQueue.java
```diff
@@ -0,0 +1,79 @@
...
+ * on top of {@link ConcurrentQueue}. Entries must implement {@link Lockable} so that
+ * they can be atomically locked when polled and unlocked when returned.
+ * <p>
+ * This is used by the composite writer pool to ensure that a writer is locked
+ * before it is handed out and unlocked when it is returned.
+ *
+ * @param <T> the type of lockable elements held in this queue
+ * @opensearch.experimental
+ */
+public final class LockableConcurrentQueue<T extends Lockable> {
```
mgodwan, line 26: Lets add a JMH benchmark for this.

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDocumentInput.java
```diff
@@ -0,0 +1,111 @@
...
+        );
+        this.onClose = Objects.requireNonNull(onClose, "onClose must not be null");
+    }
+
+    @Override
+    public void addField(MappedFieldType fieldType, Object value) {
+        primaryDocumentInput.addField(fieldType, value);
+        for (DocumentInput<?> input : secondaryDocumentInputs.values()) {
+            input.addField(fieldType, value);
+        }
```
mgodwan: Can we add exceptions with format information about where the addField failed?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeEnginePlugin.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private static final Logger logger = LogManager.getLogger(CompositeEnginePlugin.class);
+
+    /**
+     * Index setting to enable composite indexing for an index.
+     * When {@code true}, the composite engine orchestrates writes across all registered data formats.
+     * Validates that the primary data format is non-empty when enabled.
+     */
+    public static final Setting<Boolean> COMPOSITE_ENABLED = Setting.boolSetting(
+        "index.composite.enabled",
```
mgodwan: Do we need this setting? Would having a composite.secondary.formats and composite.primary.formats serve the purpose?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeEnginePlugin.java
```diff
@@ -0,0 +1,210 @@
...
+                    }
+                }
+            }
+
+            @Override
+            public Iterator<Setting<?>> settings() {
+                return List.<Setting<?>>of(PRIMARY_DATA_FORMAT).iterator();
+            }
+        },
+        Setting.Property.IndexScope
```
mgodwan: Can we mark this as final as well (along with other settings in the plugin)?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeEnginePlugin.java
```diff
@@ -0,0 +1,210 @@
...
+     * {@link DataFormat#priority()} is retained.
+     */
+    private volatile Map<String, DataFormatPlugin> dataFormatPlugins = Map.of();
+
+    /** Creates a new composite engine plugin. */
+    public CompositeEnginePlugin() {}
+
+    @Override
+    public void loadExtensions(ExtensionLoader loader) {
+        List<DataFormatPlugin> formatPlugins = loader.loadExtensions(DataFormatPlugin.class);
```
mgodwan, line 89: This can be served through the DataFormatRegistry defined in core. Do we need this to be extensible?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeEnginePlugin.java
```diff
@@ -0,0 +1,210 @@
...
+    public void onIndexModule(IndexModule indexModule) {
+        Settings settings = indexModule.getSettings();
+        boolean compositeEnabled = COMPOSITE_ENABLED.get(settings);
+        if (compositeEnabled == false) {
+            return;
+        }
+
+        String primaryFormatName = PRIMARY_DATA_FORMAT.get(settings);
+        List<String> secondaryFormatNames = SECONDARY_DATA_FORMATS.get(settings);
+        CompositeIndexingExecutionEngine.validateFormatsRegistered(dataFormatPlugins, primaryFormatName, secondaryFormatNames);
```
mgodwan: Do we need this validation here? Should this be in the initialization of the engine itself?

sandbox/libs/composite-common/src/main/java/org/opensearch/composite/RowIdGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+ * Generates monotonically increasing row IDs for cross-format document synchronization.
+ * Each writer instance gets its own {@code RowIdGenerator} so that row IDs are unique
+ * within a writer's segment scope.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class RowIdGenerator {
```
mgodwan, line 18: Can we move this to a sandbox lib?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
```diff
@@ -0,0 +1,207 @@
...
+     * to the engine's writer pool.
+     *
+     * @param engine           the composite indexing execution engine
+     * @param writerGeneration the writer generation number
+     */
+    @SuppressWarnings("unchecked")
+    public CompositeWriter(CompositeIndexingExecutionEngine engine, long writerGeneration) {
+        this.lock = new ReentrantLock();
+        this.aborted = false;
+        this.flushPending = false;
```
mgodwan: Can this be SetOnce/AtomicBoolean?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeEnginePlugin.java
alchemist51: Yes, we can remove this once the DataformatRegistry construct is present in core.

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class CompositeIndexingExecutionEngine implements IndexingExecutionEngine<CompositeDataFormat, CompositeDocumentInput> {
+
+    private static final Logger logger = LogManager.getLogger(CompositeIndexingExecutionEngine.class);
+
+    private final IndexingExecutionEngine<?, ?> primaryEngine;
+    private final List<IndexingExecutionEngine<?, ?>> secondaryEngines;
```
Bukhtawar: nit: Can we use a set instead?

alchemist51: Fixed

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
alchemist51: Created an Enum to maintain the state now

libs/concurrent-queue/src/main/java/org/opensearch/common/queue/ConcurrentQueue.java
alchemist51: Have made the change, will benchmark the perf.

alchemist51:
```
Benchmark                                                 (concurrency)  (poolSize)  (useIndex)   Mode  Cnt  Score   Error   Units
LockableConcurrentQueueBenchmark.removeAndReAdd_4threads              4          16        true  thrpt    5  4.158 ± 1.506  ops/us
LockableConcurrentQueueBenchmark.removeAndReAdd_4threads              4          16       false  thrpt    5  5.761 ± 1.236  ops/us
LockableConcurrentQueueBenchmark.removeAndReAdd_4threads              8          16        true  thrpt    5  4.777 ± 1.399  ops/us
LockableConcurrentQueueBenchmark.removeAndReAdd_4threads              8          16       false  thrpt    5  6.672 ± 1.315  ops/us
```

Going ahead and removing the concurrentHashMap from this flow.

libs/concurrent-queue/src/jmh/java/org/opensearch/common/queue/LockableConcurrentQueueBenchmark.java
```diff
@@ -0,0 +1,166 @@
...
+ * </ul>
+ */
+@Fork(3)
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 10, time = 1)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MICROSECONDS)
+@State(Scope.Benchmark)
+@SuppressWarnings("unused")
+public class LockableConcurrentQueueBenchmark {
```
mgodwan, line 41: Move this  benchmark into the lib

sandbox/libs/composite-common/src/main/java/org/opensearch/composite/RowIdGenerator.java
```diff
@@ -0,0 +1,59 @@
...
+    public long nextRowId() {
+        return counter.getAndIncrement();
+    }
+
+    /**
+     * Returns the current row ID value without incrementing.
+     *
+     * @return the current row ID
+     */
+    public long currentRowId() {
```
mgodwan, line 47: Can we just keep it visible for testing?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
```diff
@@ -0,0 +1,242 @@
...
+        for (IndexingExecutionEngine<?, ?> delegate : engine.getSecondaryDelegates()) {
+            secondaries.put(delegate.getDataFormat(), (Writer<DocumentInput<?>>) delegate.createWriter(writerGeneration));
+        }
+        this.secondaryWritersByFormat = Map.copyOf(secondaries);
+        this.rowIdGenerator = new RowIdGenerator(CompositeWriter.class.getName());
+    }
+
+    @Override
+    public WriteResult addDoc(CompositeDocumentInput doc) throws IOException {
+        // Write to primary first
```
mgodwan, line 107: check for active?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    /**
+     * Returns whether this writer has a pending flush.
+     *
+     * @return {@code true} if a flush is pending
+     */
+    public boolean isFlushPending() {
+        return getState() == WriterState.FLUSH_PENDING;
```
mgodwan, line 218: Where will this be used?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * Returns the current state of this writer.
+     *
+     * @return the writer state
+     */
+    public WriterState getState() {
+        return state.get();
```
mgodwan, line 227: This is internal. Do we need to expose this?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
```diff
@@ -0,0 +1,242 @@
...
+
+        IndexingExecutionEngine<?, ?> primaryDelegate = engine.getPrimaryDelegate();
+        this.primaryWriter = new AbstractMap.SimpleImmutableEntry<>(
+            primaryDelegate.getDataFormat(),
+            (Writer<DocumentInput<?>>) primaryDelegate.createWriter(writerGeneration)
+        );
+
+        Map<DataFormat, Writer<DocumentInput<?>>> secondaries = new LinkedHashMap<>();
+        for (IndexingExecutionEngine<?, ?> delegate : engine.getSecondaryDelegates()) {
+            secondaries.put(delegate.getDataFormat(), (Writer<DocumentInput<?>>) delegate.createWriter(writerGeneration));
```
mgodwan, line 96: Lets revisit this later if the lookup adds cycles. We can synchronize the order between writers and document inputs in engine

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
```diff
@@ -0,0 +1,242 @@
...
+ * <p>
+ * Constructed from a {@link CompositeIndexingExecutionEngine}, it iterates the engine's
+ * delegates to create per-format writers. The primary format's writer is always first
+ * in the {@code writers} list. A {@code postWrite} callback releases this writer back
+ * to the pool after each write cycle.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class CompositeWriter implements Writer<CompositeDocumentInput>, Lockable {
```
mgodwan: This can be scoped to the current package only

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeEnginePlugin.java
```diff
@@ -0,0 +1,152 @@
...
+                    "Replacing DataFormatPlugin for format [{}] (priority {} > existing {})",
+                    name,
+                    format.priority(),
+                    existingPriority
+                );
+            }
+            registry.put(name, plugin);
+            logger.info("Registered DataFormatPlugin [{}] for format [{}]", plugin.getClass().getName(), name);
+        }
+        this.dataFormatPlugins = Collections.unmodifiableMap(registry);
```
mgodwan: [nit] you can use Maps#copyOf

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
```diff
@@ -0,0 +1,283 @@
...
+    ) {
+        Objects.requireNonNull(dataFormatPlugins, "dataFormatPlugins must not be null");
+        Objects.requireNonNull(indexSettings, "indexSettings must not be null");
+
+        Settings settings = indexSettings.getSettings();
+
+        String primaryFormatName = CompositeEnginePlugin.PRIMARY_DATA_FORMAT.get(settings);
+        List<String> secondaryFormatNames = CompositeEnginePlugin.SECONDARY_DATA_FORMATS.get(settings);
+
+        validateFormatsRegistered(dataFormatPlugins, primaryFormatName, secondaryFormatNames);
```
mgodwan, line 98: What if formats are null?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    @Override
+    public RefreshResult refresh(RefreshInput refreshInput) throws IOException {
+        List<CompositeWriter> dataFormatWriters = writerPool.checkoutAll();
+
+        // Mark each writer as flush-pending before flushing
+        for (CompositeWriter writer : dataFormatWriters) {
+            writer.setFlushPending();
+        }
```
mgodwan, line 185: This can go in DFAwareEngine layer

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
```diff
@@ -0,0 +1,283 @@
...
+            total += engine.getNativeBytesUsed();
+        }
+        return total;
+    }
+
+    @Override
+    public void deleteFiles(Map<String, Collection<String>> filesToDelete) throws IOException {
+        primaryEngine.deleteFiles(filesToDelete);
+        for (IndexingExecutionEngine<?, ?> engine : secondaryEngines) {
+            engine.deleteFiles(filesToDelete);
```
mgodwan: Can we handle exception denoting where the failure happened (if any)?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
```diff
@@ -0,0 +1,283 @@
...
+        if (newSegmentList.isEmpty()) {
+            logger.debug("No new segments produced from flush");
+            return null;
+        }
+
+        logger.debug("Produced {} new segments from flush", newSegmentList.size());
+        refreshedSegments.addAll(newSegmentList);
+
+        // Delegate refresh to each per-format engine
+        RefreshInput emptyInput = RefreshInput.builder().build();
```
mgodwan, line 226: Create writer file set and send it so that engines like lucene can do addIndexes

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    @Override
+    public CompositeDocumentInput newDocumentInput() {
+        CompositeWriter writer = writerPool.getAndLock();
+        DocumentInput<?> primaryInput = primaryEngine.newDocumentInput();
+        Map<DataFormat, DocumentInput<?>> secondaryInputMap = new LinkedHashMap<>();
+        for (IndexingExecutionEngine<?, ?> engine : secondaryEngines) {
+            secondaryInputMap.put(engine.getDataFormat(), engine.newDocumentInput());
+        }
```
mgodwan, line 281: We don't need to tie it to writer now.

alchemist51:
```
REPRODUCE WITH: ./gradlew ':server:internalClusterTest' --tests 'org.opensearch.recovery.RecoveryWhileUnderLoadIT' -Dtests.method='testRecoverWhileUnderLoadAllocateReplicasRelocatePrimariesTest {p0={"cluster.indices.replication.strategy":"SEGMENT"}}' -Dtests.seed=FA2A6A0B41ACE982 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=en-NZ -Dtests.timezone=Atlantic/Madeira -Druntime.java=25

RecoveryWhileUnderLoadIT > testRecoverWhileUnderLoadAllocateReplicasRelocatePrimariesTest > testRecoverWhileUnderLoadAllocateReplicasRelocatePrimariesTest {p0={"cluster.indices.replication.strategy":"SEGMENT"}} FAILED
    java.lang.RuntimeException: java.util.NoSuchElementException: No value present
        at __randomizedtesting.SeedInfo.seed([FA2A6A0B41ACE982:2303C8CAE5033328]:0)
        at org.opensearch.test.OpenSearchIntegTestCase.waitForReplication(OpenSearchIntegTestCase.java:2590)
        at org.opensearch.recovery.RecoveryWhileUnderLoadIT.assertAfterRefreshAndWaitForReplication(RecoveryWhileUnderLoadIT.java:504)
        at org.opensearch.recovery.RecoveryWhileUnderLoadIT.testRecoverWhileUnderLoadAllocateReplicasRelocatePrimariesTest(RecoveryWhileUnderLoadIT.java:236)

        Caused by:
        java.util.NoSuchElementException: No value present
            at java.****/java.util.Optional.get(Optional.java:143)
            at org.opensearch.test.OpenSearchIntegTestCase.getIndexShard(OpenSearchIntegTestCase.java:2613)
            at org.opensearch.test.OpenSearchIntegTestCase.getIndexShard(OpenSearchIntegTestCase.java:2602)
            at org.opensearch.test.OpenSearchIntegTestCase.lambda$waitForReplication$1(OpenSearchIntegTestCase.java:2564)
            at org.opensearch.test.OpenSearchTestCase.assertBusy(OpenSearchTestCase.java:1181)
            at org.opensearch.test.OpenSearchIntegTestCase.waitForReplication(OpenSearchIntegTestCase.java:2558)
            ... 2 more
```

libs/concurrent-queue/src/jmh/java/org/opensearch/common/queue/LockableConcurrentQueueBenchmark.java
alchemist51: Done

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
alchemist51: added validation in the validateFormatsRegistered function

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
alchemist51: Will move it in the follow up PR to move the writerPool into the DFAwareEngine

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
alchemist51: Will relook the refresh once DFAwareEngine is present

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
alchemist51: Done

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
alchemist51: Agree, earlier I had made it synchronous, but since we will have handful of dataformat, this should not be expensive.. We can revisit it once we do bench-marking again

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
alchemist51: Done

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
alchemist51: Done

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
alchemist51: removed public access.

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
alchemist51: done

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeEnginePlugin.java
alchemist51: done

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
alchemist51: Done

sandbox/libs/composite-common/src/main/java/org/opensearch/composite/RowIdGenerator.java
alchemist51: done

Issue #20910, opened 2026-03-18
[BUG] Remote cluster state compatibility failures


andrross:
### Describe the bug

A BWC test for remote cluster state was added in #20221. This is failing intermittently:

https://build.ci.opensearch.org/job/gradle-check/72744/consoleText
https://build.ci.opensearch.org/job/gradle-check/72748/consoleText

1. Build failure (top-level):
> Task :qa:rolling-upgrade:v2.19.6-remote#twoThirdsUpgradedTest FAILED

Execution failed for task ':qa:rolling-upgrade:v2.19.6-remote#twoThirdsUpgradedTest'.
> process was found dead while waiting for cluster health yellow, cluster{:qa:rolling-upgrade:v2.19.6-remote}

2. IndexMetadata XContent deserialization failure (old node reading index metadata blobs written by upgraded cluster-manager):
```
[2026-03-18T11:33:45,850][ERROR][o.o.g.r.RemoteClusterStateService] [v2.19.6-remote-2] Failed to read cluster state from remote
org.opensearch.gateway.remote.RemoteStateTransferException: Download failed for java_for_range
        at org.opensearch.gateway.remote.RemoteIndexMetadataManager.lambda$getWrappedReadListener$3(RemoteIndexMetadataManager.java:159)
        at org.opensearch.core.action.ActionListener$1.onFailure(ActionListener.java:90)
        at org.opensearch.common.remote.RemoteWriteableEntityBlobStore.lambda$readAsync$0(RemoteWriteableEntityBlobStore.java:87)
        at org.opensearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:955)
        ...
Caused by: java.lang.IllegalStateException: Can't get text on a START_ARRAY at -1:702
        at org.opensearch.common.xcontent.json.JsonXContentParser.text(JsonXContentParser.java:99)
        at org.opensearch.core.xcontent.AbstractXContentParser.map(AbstractXContentParser.java:298)
        at org.opensearch.core.xcontent.AbstractXContentParser.mapStrings(AbstractXContentParser.java:282)
        at org.opensearch.cluster.metadata.IndexMetadata$Builder.fromXContent(IndexMetadata.java:2013)
        at org.opensearch.cluster.metadata.IndexMetadata.fromXContent(IndexMetadata.java:1080)
[...]

PR #20912, opened 2026-03-18
Test remote state BWC in both CM version directions

andrross:
The RemotePublicationClusterStateIT test was not deterministically exercising the case where a new-version cluster-manager writes remote cluster state that old-version nodes must deserialize. This meant backwards-incompatible serialization changes could be merged without being caught, since the test would pass whenever an old-version node happened to win the cluster-manager election.

In the mixed-version (one-third upgraded) phase, the test now explicitly tests both directions:
1. Old CM writes state, new nodes read from remote store
2. New CM writes state, old nodes read from remote store

To force a specific version to be cluster-manager, the test uses the voting config exclusions API to repeatedly exclude the current CM until a node of the desired version wins the election. Exclusions are cleared immediately after each re-election so no node leaves the cluster.

Also changes the remote test cluster's dependency on the non-remote test cluster from dependsOn to mustRunAfter, so the remote tests can be run independently without first running the full non-remote suite.

### Related Issues
Related to #20910

### Check List

### DIFF
--- a/qa/rolling-upgrade/build.gradle
+++ b/qa/rolling-upgrade/build.gradle
@@ -132,7 +132,7 @@ for (Version bwcVersion : BuildParams.bwcVersions.wireCompatible) {
   }
 
   tasks.register("${remoteBaseName}#oldClusterTest", StandaloneRestIntegTestTask) {
-    dependsOn "${baseName}#upgradedClusterTest"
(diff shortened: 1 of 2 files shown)


andrross:
Closing as forward compatibility is explicitly not supported here. See #20948

Issue #20914, opened 2026-03-18
[META] OpenSearch DSL Query Executor Plugin


vinaykpud:
### Please describe the end goal of this project

Part of: #19902: Analytics Engine: Query Execution Framework

This meta issue tracks the incremental development of the `dsl-query-executor` sandbox plugin — a front-end to the analytics engine that intercepts OpenSearch `_search` requests, converts DSL queries into Calcite RelNode logical plans,
and executes them through the analytics engine's query pipeline.

## Architecture

<img width="5099" height="2976" alt="Image" src="https://github.com/user-attachments/assets/42ab11a0-1297-4f5d-ae10-8a6df1afe2ae" />

<br/><br/>

* **`SearchActionFilter`** : An ActionFilter that intercepts all `_search` requests at the transport layer. If the action matches `SearchAction`, it dispatches the `SearchRequest` to `TransportDslExecuteAction` via `DslExecuteAction`. All other actions pass through unchanged to the default OpenSearch path.
* **`TransportDslExecuteAction`** :  Receives `QueryPlanExecutor` and `EngineContext` from the Analytics engine. Resolves the target index, creates a `SearchSourceConverter` with the schema from `EngineContext`, converts the DSL to `QueryPlans`, executes them via `DslQueryPlanExecutor`, and builds the final `SearchResponse`.
* **`SearchSourceConverter`** : Converts a `SearchSourceBuilder` DSL query into Calcite `QueryPlans`. Builds its own Calcite planning infrastructure (HepPlanner, RelOptCluster, CalciteCatalogReader) from the SchemaPlus provided by the analytics engine. Looks up the index as a `RelOptTable` and creates `RelNode` logical plans. 
* **`QueryPlans`** : Immutable container holding one or more QueryPlans produced by conversion. Each QueryPlan pairs a Type (HITS or AGGREGATION) with a Calcite RelNode. A single DSL query may produce multiple plans.
* **`DslQueryPlanExecutor`** : Iterates over `QueryPlans`, delegates each RelNode to the analytics engine's QueryPlanExecutor for execution, and collects the results as `ExecutionResults`.
*  **`ExecutionResult`** : Pairs a `QueryPlan` with the `Iterable<Object[]>` rows returned by the executor. Provides convenience accessors for plan type, field names (derived from the RelNode row type), and the raw rows. 
* **`SearchResponseBuilder`** : Builds a standard OpenSearch `SearchResponse` from a list of ExecutionResults. 

[...]

PR #20916, opened 2026-03-18
Add dsl-query-executor sandbox plugin for dsl query execution via analytics engine

vinaykpud:
Introduces a new sandbox plugin related to this [issue](https://github.com/opensearch-project/OpenSearch/issues/20914) that intercepts `_search` requests and routes them through a execution pipeline via the analytics engine.

### Description
This PR introduces the `dsl-query-executor` sandbox plugin skeleton with end-to-end wiring: action filter → transport action → converter → executor → response builder. 

The converter currently produces a `LogicalTableScan` RelNode only, the executor returns empty rows (analytics engine stub), and the response builder constructs an empty `SearchResponse`. Actual DSL-to-RelNode conversion, execution, and response population will be added incrementally in follow-up PRs.

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/20914

### Check List

### DIFF
--- a/sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/result/SearchResponseBuilderTests.java
+++ b/sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/result/SearchResponseBuilderTests.java
@@ -16,7 +16,7 @@ import java.util.List;
 public class SearchResponseBuilderTests extends OpenSearchTestCase {
 
     public void testBuildReturnsEmptyResponse() {
-        SearchResponse response = SearchResponseBuilder.build(List.of(), 42L);
+        SearchResponse response = SearchResponseBuilder.build(List.of(), 42L * 1_000_000);
 
         assertNotNull(response);
         assertEquals(200, response.status().getStatus());


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
```diff
@@ -0,0 +1,62 @@
...
+    @SuppressWarnings("unchecked")
+    public <Request extends ActionRequest, Response extends ActionResponse> void apply(
+        Task task,
+        String action,
+        Request request,
+        ActionRequestMetadata<Request, Response> actionRequestMetadata,
+        ActionListener<Response> listener,
+        ActionFilterChain<Request, Response> chain
+    ) {
+        if (SearchAction.NAME.equals(action)) {
```
expani, line 60: Will we create another ActionFilter for `_msearch`, `_count` and other APIs like `_search_shards` ( TIL ) https://docs.opensearch.org/latest/api-reference/search-apis/search-shards ? 

At a high level, I can see 2 distinct factors here : 
- APIs that actually execute search ( _search, _count, _msearch ) 
- APIs that just do explain or validation ( _validate/query, _search_shards, _explain ) 

Again just a note to keep in the back of mind and not something we should handle right from the start.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
```diff
@@ -0,0 +1,118 @@
...
+
+            SearchResponse response = SearchResponseBuilder.build(results, tookInMillis);
+            listener.onResponse(response);
+        } catch (Exception e) {
+            logger.error("DSL execution failed", e);
+            listener.onFailure(e);
+        }
+    }
+
+    // TODO: add multi-index support:
```
expani: This raises a broader question of Should Frontend plugins resolve the index patterns or Analytics Core plugin do the same in a consistent manner for all plugins ?

For instance, PPL pushes down the index resolution to OpenSearch Core today. I don't see it calling `IndexNameExpressionResolver.concreteIndexNames` 

Should we follow the same route and push down index pattern to AnalyticsPlugin ? 

The downside is the failure will happen a bit late at AnalyticsPlugin compared to eagerly failing in the DSL Convertor plugin.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/DslExecuteAction.java
```diff
@@ -0,0 +1,28 @@
...
+import org.opensearch.action.search.SearchResponse;
+
+/**
+ * Internal action for executing DSL queries through the analytics engine.
+ * Accepts a {@link org.opensearch.action.search.SearchRequest} and returns a {@link SearchResponse}.
+ */
+public class DslExecuteAction extends ActionType<SearchResponse> {
+
+    /** Action name registered with the transport layer. */
+    public static final String NAME = "cluster:internal/dsl/execute";
```
expani: For customers migrating to these new indices, this would mean a new Security permission needs to be added for all users executing searches ? If so, let's add a TODO for handling this.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
```diff
@@ -0,0 +1,118 @@
...
+    @Override
+    protected void doExecute(Task task, SearchRequest request, ActionListener<SearchResponse> listener) {
+        try {
+            String indexName = resolveToSingleIndex(request);
+            long startTime = System.currentTimeMillis();
+
+            SearchSourceConverter converter = new SearchSourceConverter(engineContext.getSchema());
+            QueryPlans plans = converter.convert(request.source(), indexName);
+            List<ExecutionResult> results = planExecutor.execute(plans);
+            long tookInMillis = System.currentTimeMillis() - startTime;
```
expani: I suppose the intent is to capture the time taken to build the RelNode from DSL and also time spent in building the response. 

We can exclude `planExecutor.execute(plans)` and just add the timings it returns along with the time spent in DSL Converter plugin.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/SearchSourceConverter.java
```diff
@@ -0,0 +1,87 @@
...
+    private final RelOptCluster cluster;
+    private final CalciteCatalogReader catalogReader;
+
+    /**
+     * Initializes planning infrastructure from the given schema.
+     *
+     * @param schema Calcite schema with index tables from the analytics engine
+     */
+    public SearchSourceConverter(SchemaPlus schema) {
+        RelDataTypeFactory typeFactory = new SqlTypeFactoryImpl(RelDataTypeSystem.DEFAULT);
```
expani, line 55: The typeFactory should be created from EngineContext instead of depending on the default type system. 

Since, AnalyticsPlugin doesn't expose let's add a TODO to integrate the same in DSL Convertor plugin when available via EngineContext

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/SearchSourceConverter.java
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * Converts DSL for the given index into query plans.
+     *
+     * @param searchSource the DSL query
+     * @param indexName target index
+     * @return one or more query plans
+     */
+    public QueryPlans convert(SearchSourceBuilder searchSource, String indexName) {
+
+        RelOptTable table = catalogReader.getTable(List.of(indexName));
```
expani, line 77: If we choose to leave index pattern resolution to Analytics plugin, then we might need to create RelOptTable using one of it's static factory methods instead of wrapping using a CatalogReader ( which does the same internally )

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/executor/QueryPlans.java
```diff
@@ -0,0 +1,115 @@
...
+
+        /** Returns what part of the response this plan produces. */
+        @Override
+        public Type type() {
+            return type;
+        }
+
+        /** Returns the Calcite logical plan to execute. */
+        @Override
+        public RelNode relNode() {
```
expani, line 58: One query plan has one Type and produces one RelNode.

We will need multiple RelNodes for some aggregations like nested. But, that would break the contract here. 

So, maybe we keep this a Linked List of RelNode inherently capturing the order in which they are executed. 

If you plan to add this as a followup, please add a TODO

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/SearchResponseBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+        // TODO: populate hits and aggregations from results
+        SearchHits hits = SearchHits.empty(true);
+        SearchResponseSections sections = new SearchResponseSections(hits, null, null, false, null, null, 0);
+        return new SearchResponse(
+            sections,
+            null,
+            0,
+            0,
+            0,
+            tookInMillis,
```
expani: We need to also capture the time took for building the search response. Maybe add a TODO for now

sandbox/plugins/dsl-query-executor/build.gradle
```diff
@@ -0,0 +1,28 @@
...
+  description = 'Sandbox plugin for DSL query execution via Calcite logical plans.'
+  classname = 'org.opensearch.dsl.DslQueryExecutorPlugin'
+  extendedPlugins = ['analytics-engine']
+}
+
+dependencies {
+  compileOnly project(':server')
+  compileOnly project(':sandbox:libs:analytics-framework')
+  compileOnly project(':sandbox:plugins:analytics-engine')
+  compileOnly 'org.apache.calcite.avatica:avatica-core:1.27.0'
```
expani, line 22: I see the analytics-framework has marked it as runtime which forced DSLConvertor use this separately
```
runtimeOnly 'org.apache.calcite.avatica:avatica-core:1.27.0'
```

It's good to start but please add a TODO to consume this from Analytics Framework

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
vinaykpud: The approach is different between the current SQL plugin implementation and the new analytics-based architecture.
In the SQL plugin today, table registration is lazy, ie `OpenSearchIndex` is created with just the raw name string from the query and registers as a Calcite table. Schema resolution also happens lazily: when Calcite needs the table's row type (field names + types), it calls `prepareGetMappings`, and OpenSearch internally resolves wildcards, aliases, and multi-index to concrete indices and returns their mappings.

In the new approach, `OpenSearchSchemaBuilder.buildSchema(clusterState)` builds the schema upfront from ClusterState metadata, registering each concrete index as a table entry in the SchemaPlus. Frontend plugins consume this to construct `RelOptTable` using concrete index names. 
There is a related PoC in the SQL plugin: https://github.com/dai-chen/sql-1/pull/10 - I think it also does not support wildcards/aliases/multi-index
   
So with this approach of building SchemaPlus upfront in Analytics Core, frontend plugins would need to resolve index patterns (wildcards, aliases, multi-index) to concrete index names before table lookup.

Yes, The alternative is pushing index patterns down to the Analytics plugin, similar to how the SQL plugin delegates to OpenSearch today, but as you noted, failures would surface later in the pipeline rather than at the DSL conversion stage.

expani: Then shouldn't we use `OpenSearchSchemaBuilder` or via EngineContext to just return the Table for use in Frontend plugins ? 

This way the validation happens upfront and index pattern resolution also stays at a common place.

vinaykpud: For this plugin Yes, given given concrete index/wildcards/aliases/multi-index if Analytics core can return `RelOptTable` for the resolved indices we can use it. It basically simplifies [this](https://github.com/vinaykpud/OpenSearch/blob/dd475b20b2b1f5107922479bef2741ef297b5389/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/SearchSourceConverter.java#L52)

So decision is basically if the front end plugin needs to recreate the Table? If yes returning schema, else return Table.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/executor/QueryPlans.java
vinaykpud: [QueryPlans.java](https://github.com/vinaykpud/OpenSearch/blob/dd475b20b2b1f5107922479bef2741ef297b5389/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/executor/QueryPlans.java#L61) actually holds One or more `QueryPlan`.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
expani: Looks like [Calcite Schema ( which SchemaPlus wraps )](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/schema/Schema.java#L68C3-L68C35)  has  a way to lookup tables which can also be used.

expani: Let's add TODOs so we don't miss this when changing things, Nothing critical.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/executor/QueryPlans.java
expani: Yes but I was wondering how are we linking the multiple RelNodes generated by one Aggregation ?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
vinaykpud: Okay, I will add a TODO for this.
This requires [DefaultPlanExecutor.java](https://github.com/vinaykpud/OpenSearch/blob/7c055e32e55cba261f3d7953f09de96a43cfe654/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java#L22) return time spent.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/executor/QueryPlans.java
expani: Discussed offline, this requires recursive nesting of child query plans.

sandbox/plugins/dsl-query-executor/src/internalClusterTest/java/org/opensearch/dsl/DslQueryExecutorIT.java
```diff
@@ -0,0 +1,79 @@
...
+ * SearchActionFilter → TransportDslExecuteAction → SearchSourceConverter → DslQueryPlanExecutor → SearchResponse.
+ */
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.SUITE, numDataNodes = 1)
+public class DslQueryExecutorIT extends OpenSearchIntegTestCase {
+
+    private static final String INDEX = "test-index";
+
+    @Override
+    protected Collection<Class<? extends Plugin>> nodePlugins() {
+        return List.of(AnalyticsPlugin.class, DslQueryExecutorPlugin.class);
```
mch2, line 33: i think these kinds of cross plugn tests we should move to a /qa test so that we can test the actual plugin bootstrap with a REST IT. ITs within this pkg validate the generated plan.  Am working on introducing one with a test PPL flow in /sandbox/qa - we can add one for DSL as well.

vinaykpud: Sure, for now i will keep it. Once we merge that /qa test we can remove these.

PR #20920, opened 2026-03-19
Remove relative date in IndicesRequestCacheIT

andrross:
The dates in the test are starting to fall out of the now-10y window. It should be fine to use a fixed date for the 'from' since the 'to' uses now.

### Check List


kkewwei:
LGTM.

After analysis, the root cause is as follows:

- Previously, the time range now-10y ~ now had an INTERSECTS relation with the data set (originally WITHIN). After the rewrite, from-to becomes null, so the second time parsing is skipped and the query should be cacheable = true.

- However, since the relation between now-10y ~ now and the data set is still INTERSECTS, `parseSource` will rewrite the dynamic time now again, which results in cacheable = false.

cwperks:
How many other ticking time bombs do we have in the core or any other repo?


Speaking for security, I know we have some demo certs that expire in the future but for the most part the repo will generate certs on the fly for testing.

PR #20921, opened 2026-03-19
Use caching to avoid excessive creation of NonClosingReaderWrapper instances

### LINKED ISSUE #19175: Avoid creating NonClosingReaderWrapper per search query per phase

As of today, for any search query, we create a new `NonClosingReaderWrapper` object per phase ie this will be created twice per query if a query involves a query and fetch phase.
See [here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/shard/IndexShard.java#L2215C63-L2215C86)

For context, `NonClosingReaderWrapper` was probably created so that any external plugin wrapping the underlying reader doesn't have a way to accidentally/erroneously close it.

We recently discovered an issue all such per request reader wrapper gets registered as "parent" for the underlying segment/leaf reader. This caused an performance issue described [here](https://github.com/apache/lucene/issues/14999). In a nutshell, this would cause >30k parents being registered per segment reader, and would cause high wait times during `SynchronizetSet.add`(inside IndexReader) for an index shard. 

We also saw security plugin's DlsFlsFilterLeafReader also being registered as parent during this issue.

**Lucene did ended up fixing it i.e, by removing the logic to explicitly track parent readers, but I think we should avoid creating such reader wrappers per query.**

Instead of creating `NonClosingReaderWrapper` per query, we should instead evaluate whether we can instead  create a shared object per shard per reader.

In addition, it would be great if we can find a way to let external plugin(like security plugin) also avoid creating such wrappers per query.

### Related component

Search:Performance

_No response_

### Additional context

_No response_

kkewwei:
…stances

### Description
Instead of creating a NonClosingReaderWrapper per query/update, we reuse the same NonClosingReaderWrapper by caching.

We can't reuse it in external plugin(like security plugin), in security plugin, the result may depend on conditions such as the user, so the same `NonClosingReaderWrapper` cannot be reused universally.

image

### Related Issues
Resolves #19175

### Check List


kkewwei:
@sgup432 Please help review in your spare time.

kkewwei:
> ❌ Gradle check result for [5f67c69](https://github.com/opensearch-project/OpenSearch/commit/5f67c69c60702b4ddcaf5bc7d7a8e2dc2417e477): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/72823/)
> 

org.opensearch.indices.IndicesRequestCacheIT.testCanCache  #20920

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -2315,7 +2340,8 @@ public MergedSegmentTransferTracker mergedSegmentTransferTracker() {
      *
      * @opensearch.internal
      */
-    private static final class NonClosingReaderWrapper extends FilterDirectoryReader {
+    @PublicApi(since = "3.6.0")
```
sgup432: Why do we need to make this public?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -2245,15 +2249,36 @@ private Engine.Searcher wrapSearcher(Engine.Searcher searcher) {
...
-        );
+        DirectoryReader directoryReader = engineSearcher.getDirectoryReader();
+        final OpenSearchDirectoryReader openSearchDirectoryReader = OpenSearchDirectoryReader.getOpenSearchDirectoryReader(directoryReader);
         if (openSearchDirectoryReader == null) {
             throw new IllegalStateException("Can't wrap non opensearch directory reader");
         }
-        NonClosingReaderWrapper nonClosingReaderWrapper = new NonClosingReaderWrapper(engineSearcher.getDirectoryReader());
+
+        NonClosingReaderWrapper nonClosingReaderWrapper;
+        if (readerWrapperCache == null) {
```
sgup432: I guess this would never be null?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -321,6 +322,7 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
 
     private final IndexingOperationListener indexingOperationListeners;
     private final Runnable globalCheckpointSyncer;
+    private final ConcurrentHashMap<DirectoryReader, NonClosingReaderWrapper> readerWrapperCache = new ConcurrentHashMap<>();
```
sgup432: Lets clear this map during IndexShard.close() as well? Though it would eventually clear it out when all underlying segments are closed, but still good to have?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
kkewwei: Being a static function, `wrapSearcher` is invoked directly within `InternalEngine`. Since `InternalEngine` does not have access to the variables in the corresponding IndexShard, these variables will be set to null upon entry into this function.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
kkewwei: IndexReaderWrapperTests needs to reference NonClosingReaderWrapper, which requires the class to be exposed. However, I believe we can avoid this issue by using DirectoryReader externally instead.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -2245,15 +2249,36 @@ private Engine.Searcher wrapSearcher(Engine.Searcher searcher) {
...
+        DirectoryReader nonClosingReaderWrapper;
+        if (readerWrapperCache == null) {
+            nonClosingReaderWrapper = new NonClosingReaderWrapper(directoryReader);
+        } else {
+            nonClosingReaderWrapper = readerWrapperCache.computeIfAbsent(directoryReader, key -> {
+                try {
+                    OpenSearchDirectoryReader.addReaderCloseListener(key, cacheKey -> readerWrapperCache.remove(key));
+                    return new NonClosingReaderWrapper(directoryReader);
+                } catch (IOException e) {
+                    readerWrapperCache.remove(key);
```
sgup432: I don't think we should modify map inside `ConcurrentHashMap.computeIfAbsent`.  The javadoc clearly says - "_The mapping function must not modify this map during computation_."

I think it can cause deadlock. As `computeIfAbsent` will hold a write lock to a partition/bin, and then doing `remove` would try taking another write lock on the same partition, causing deadlock.

sgup432: Lets move this logic outside the `computeIfAbsent`

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -321,6 +322,7 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
 
     private final IndexingOperationListener indexingOperationListeners;
     private final Runnable globalCheckpointSyncer;
+    private final ConcurrentHashMap<DirectoryReader, DirectoryReader> readerWrapperCache = new ConcurrentHashMap<>();
```
sgup432: Any reason why are we using `DirectoryReader` as a value and not `NonClosingReaderWrapper`?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -2245,15 +2249,36 @@ private Engine.Searcher wrapSearcher(Engine.Searcher searcher) {
...
         Engine.Searcher engineSearcher,
         CheckedFunction<DirectoryReader, DirectoryReader, IOException> readerWrapper
+    ) throws IOException {
+        return wrapSearcher(engineSearcher, readerWrapper, null);
+    }
+
+    public static Engine.Searcher wrapSearcher(
+        Engine.Searcher engineSearcher,
+        CheckedFunction<DirectoryReader, DirectoryReader, IOException> readerWrapper,
+        ConcurrentHashMap<DirectoryReader, DirectoryReader> readerWrapperCache
```
sgup432: Actually this might not be the right way to do ie passing `concurrentHashMap` here. This is essentially a public API so exposing caching logic outside to clients isn't right? This caching logic should be an internal one.

So to avoid passing it here, we would have to probably move the caching logic inside the private method? Let me know what you think.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
kkewwei: `NonClosingReaderWrapper` is declared as private static class, making it inaccessible from the test class. If we using DirectoryReader as NonClosingReaderWrapper, we will no longer be able to write unit tests.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
kkewwei: It make sense to me.

PR #20923, opened 2026-03-19
Add from and size parameters to terms lookup query for pagination support

### LINKED ISSUE #20865: [Feature Request] Enhancing Terms lookup by query to support pagination and other search features

The applicability of [terms lookup by query](https://docs.opensearch.org/3.3/query-dsl/term/terms/#terms-lookup-by-query) is handicapped by the `max_terms_count` limit of 65,536. And although this is configurable, network and memory overhead can become significant even around 10,000.

**Specific problem:** For use cases involving high-cardinality membership sets (e.g., 100,000+ documents in the lookup index), there is currently no way to paginate through the matched documents before value collection.

Add support for **`from` and `size` parameters** within the terms lookup object to enable pagination of the document matching phase before field value collection.

**Proposed syntax:**

```json
{
  "terms": {
    "ASIN": {
      "index": "membership-index",
      "path": "ASIN",
      "query": { "term": { "userId": "user123" } },
      "from": 0,      // ← Skip this many matched documents
      "size": 10000   // ← Collect values from up to this many documents
    }
  }
}
```

**Nice to have:**

Add support for **`sort` clause and `search_after` parameter** within the terms lookup object to enable scrolling.

### Related component

Search:Query Capabilities

NA

### Additional context

https://github.com/opensearch-project/OpenSearch/issues/17599

ShawnQiang1:
## Description
This PR adds support for `from` and `size` parameters in terms lookup queries, enabling pagination when fetching terms from a lookup index. This is particularly useful when dealing with large lookup datasets where fetching all terms at once is not efficient or exceeds system limits.

100% AI generated

### Related Issues
Resolves #20865

### Changes
- Added `from` and `size` fields to `TermsLookup` class
- Updated `TermsLookup` constructors, getters, setters, and builder methods
- Added XContent parsing support for the new parameters
- Updated serialization/deserialization logic (Version.V_3_0_0+)
- Modified `TermsQueryBuilder.fetch()` to use user-specified `from` and `size` parameters
- Added comprehensive YAML REST test cases

### Usage Example
```json
{
  "terms": {
    "ASIN": {
      "index": "membership-index",
      "path": "ASIN",
      "query": { "term": { "userId": "user123" } },
      "from": 0,
      "size": 10000
    }
  }
}
```

### Testing
- Added 4 new test cases in `171_terms_lookup_query.yml`:
  - `size` parameter limits number of terms fetched
  - `from` parameter skips first N terms
  - Combined `from` and `size` parameters
  - `size=0` returns no terms
- All existing and new tests pass
- Code formatted with `./gradlew spotlessApply`

### Check List
- [x] All tests pass
- [x] Commits are signed per the DCO using --signoff
- [x] CHANGELOG.md updated

### DIFF
--- a/server/src/main/java/org/opensearch/indices/TermsLookup.java
+++ b/server/src/main/java/org/opensearch/indices/TermsLookup.java
@@ -64,12 +64,18 @@ public class TermsLookup implements Writeable, ToXContentFragment {
     private final String path;
     private String routing;
     private QueryBuilder query;
+    private int from = 0;
+    private Integer size;
 
     public TermsLookup(String index, String id, String path) {
         this(index, id, path, null);
     }
 
     public TermsLookup(String index, String id, String path, QueryBuilder query) {
+        this(index, id, path, query, 0, null);
+    }
+
+    public TermsLookup(String index, String id, String path, QueryBuilder query, int from, Integer size) {
         if (index == null) {
             throw new IllegalArgumentException("[" + TermsQueryBuilder.NAME + "] index cannot be null or empty for TermsLookup");
         }
@@ -90,6 +96,8 @@ public class TermsLookup implements Writeable, ToXContentFragment {
         this.id = id;
         this.path = path;
         this.query = query;
+        this.from = from;
+        this.size = size;
     }
 
     public String index() {
@@ -117,6 +125,10 @@ public class TermsLookup implements Writeable, ToXContentFragment {
         if (in.getVersion().onOrAfter(Version.V_3_2_0)) {
             query = in.readOptionalWriteable(inStream -> inStream.readNamedWriteable(QueryBuilder.class));
         }
+        if (in.getVersion().onOrAfter(Version.V_3_5_0)) {
+            from = in.readInt();
+            size = in.readOptionalInt();
+        }
     }
 
     @Override
@@ -133,6 +145,10 @@ public class TermsLookup implements Writeable, ToXContentFragment {
         }
         if (out.getVersion().onOrAfter(Version.V_3_2_0)) {
             out.writeOptionalWriteable(query);
+        }
+        if (out.getVersion().onOrAfter(Version.V_3_5_0)) {
+            out.writeInt(from);
+            out.writeOptionalInt(size);
         }
     }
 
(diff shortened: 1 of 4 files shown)


ShawnQiang1:
@navneet1v @andrross

ShawnQiang1:
@navneet1v  I need some advice. Currently, this paginated subquery doesn't seem to detect when the query has finished. What methods do you think we should use to inform the user that the lookup query has returned 0 records?

navneet1v:
> @navneet1v I need some advice. Currently, this paginated subquery doesn't seem to detect when the query has finished. What methods do you think we should use to inform the user that the lookup query has returned 0 records?


@ShawnQiang1 first of all this its a good idea to add the from and size capability to terms lookup query. I don't think we should provide any support if from and size doesn't give any result. The way I am thinking about this problem is if terms lookup give no results, then query should just proceed thinking the terms provided in query is empty by the user. and based on the other query clause user might get 0 or more results. 

If someone want to get deeper into the query they can use explain api to understand the query. I am hoping this should provide details that terms lookup query resulted in 0 values.

@ShawnQiang1 let me know if I answered your question, or I wandered in a direction far away from your original question.

ShawnQiang1:
@navneet1v I think you're right. Pagination search does have a relatively high performance overhead. When people use it, they're usually based on the number of terms retrieved (e.g., the first 1000 terms in the query), and don't typically stop just because the result is empty. So you're correct; we can leave it as is. 
Also, if you have time, could you please review my code? Thank you very much.

server/src/main/java/org/opensearch/indices/TermsLookup.java
```diff
@@ -117,6 +125,10 @@ public TermsLookup(StreamInput in) throws IOException {
         if (in.getVersion().onOrAfter(Version.V_3_2_0)) {
             query = in.readOptionalWriteable(inStream -> inStream.readNamedWriteable(QueryBuilder.class));
         }
+        if (in.getVersion().onOrAfter(Version.V_3_5_0)) {
```
ShawnQiang1: Actually, I'm not sure whether we need this version  Guardian or not. It seems like it could be removed if our feature doesn't require backport.

PR #20928, opened 2026-03-19
Fix case-insensitive wildcard/regexp queries for non-ASCII characters

### LINKED ISSUE #20843: [BUG] wildcard query with case_insensitive: true fails for some languages (Turkish and Ukrainian for example)
### Describe the bug

Retrieving documents using a `wildcard` query with `case_insensitive: true` fails for some languages (Turkish and Ukrainian for example), while working correctly for others (English and German for example). This failure occurs even when the query value and casing match the stored data exactly.

### Related component

Search

### To Reproduce

1. create the index:
```
curl -X PUT "http://localhost:9200/test_idx" -H 'Content-Type: application/json' -d'
{
  "mappings": {
    "dynamic": false,
    "properties": {
      "name": {
        "type": "wildcard",
        "doc_values": false
      }
    }
  }
}
'
```

2. Insert test documents (English, Turkish, Ukranian, German)
```
# English (Control)
curl -X POST "http://localhost:9200/test_idx/_doc/" -H 'Content-Type: application/json' -d'{"name": "Alice Wonderland"}'

# German 
curl -X POST "http://localhost:9200/test_idx/_doc/" -H 'Content-Type: application/json' -d'{"name": "Heinz Meißner"}'

# Turkish
curl -X POST "http://localhost:9200/test_idx/_doc/" -H 'Content-Type: application/json' -d'{"name": "Gökçe İrmak"}'

# Ukrainian
curl -X POST "http://localhost:9200/test_idx/_doc/" -H 'Content-Type: application/json' -d'{"name": "Олександр Зінченко"}'
```

3. Run the GET queries with wildcard + case_insensitive:
use this query template
```
curl -X GET "http://localhost:9200/test_idx/_search?pretty" -H 'Content-Type: application/json' -d'
{
  "query": {
    "wildcard": {
      "name": {
        "value": "*<NameString>*",
        "case_insensitive": true
      }
    }
  }
}
'
```

TEST RESULTS SUMMARY:

Image

for example the failed Turkish query looks like this:
```
curl -X GET "http://localhost:9200/test_idx/_search?pretty" -H 'Content-Type: application/json' -d'
{
  "query": {
    "wildcard": {
      "name": {
        "value": "*Gökçe İrmak*",
        "case_insensitive": true
      }
    }
  }
}
'
{
  "took" : 44,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 0,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  }
}
```

### Expected behavior

All documents should be returned when the wildcard value matches the stored string. The character set should not impact the retrieval capability of the wildcard type, especially when using exact-case strings.

### Additional Details

**Plugins**

**Screenshots**

 - OS: Amazon Linux (AL2023) / Docker
 - Version: OpenSearch 3.5.0 (latest)

**Additional context**
- removing the case_insensitive (or setting it to `false`) returns as expected also in Turkish and Ukranian:
```
curl -X GET "http://localhost:9200/test_idx/_search?pretty" -H 'Content-Type: application/json' -d'
{
  "query": {
    "wildcard": {
      "name": {
        "value": "*Gökçe İrmak*",
        "case_insensitive": false
      }
    }
  }
}
'
{
  "took" : 24,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 1,
      "relation" : "eq"
    },
    "max_score" : 1.0,
    "hits" : [
      {
        "_index" : "test_idx",
        "_id" : "_I5m3ZwBBv_NcRY78s9B",
        "_score" : 1.0,
        "_source" : {
          "name" : "Gökçe İrmak"
        }
      }
    ]
  }
}
```

harshavamsi:
## Summary
- Fixes #20843: Case-insensitive wildcard and regexp queries silently ignore non-ASCII characters (Turkish, Cyrillic, Greek, etc.)
- Root cause: `AutomatonQueries.toCaseInsensitiveChar()` had an early return for codepoints > 128, skipping case folding for all non-ASCII Unicode
- Replaced ASCII-only guard with proper Unicode case folding via `Character.toLowerCase()`/`Character.toUpperCase()`, correctly handling Turkish İ/ı, Cyrillic, Greek sigma, and other scripts

## Test plan
- [x] Unit tests: `AutomatonQueriesTests` — 6 test methods covering ASCII, Turkish (İ/ı), Cyrillic, Greek, no-case-variant characters, and mixed-script strings
- [x] Integration test: `WildcardQueryBuilderTests.testCaseInsensitiveWithNonAsciiCharacters` — verifies wildcard queries with Turkish and Ukrainian text
- [ ] Manual test: Create index with Turkish text, run `case_insensitive: true` wildcard query matching İ↔i

### DIFF
--- a/server/src/test/java/org/opensearch/common/lucene/search/AutomatonQueriesTests.java
+++ b/server/src/test/java/org/opensearch/common/lucene/search/AutomatonQueriesTests.java
@@ -1,0 +1,119 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.common.lucene.search;
+
+import org.apache.lucene.util.automaton.Automaton;
+import org.apache.lucene.util.automaton.CharacterRunAutomaton;
+import org.opensearch.test.OpenSearchTestCase;
+
+/**
+ * Tests for {@link AutomatonQueries}
+ */
+public class AutomatonQueriesTests extends OpenSearchTestCase {
+
+    /**
+     * Test that toCaseInsensitiveChar works correctly for ASCII characters
+     */
+    public void testToCaseInsensitiveCharAscii() {
+        // Test lowercase ASCII
+        Automaton automaton = AutomatonQueries.toCaseInsensitiveChar('a');
+        CharacterRunAutomaton runAutomaton = new CharacterRunAutomaton(automaton);
+        assertTrue("Should match lowercase 'a'", runAutomaton.run("a"));
+        assertTrue("Should match uppercase 'A'", runAutomaton.run("A"));
+        assertFalse("Should not match 'b'", runAutomaton.run("b"));
+
+        // Test uppercase ASCII
+        automaton = AutomatonQueries.toCaseInsensitiveChar('Z');
+        runAutomaton = new CharacterRunAutomaton(automaton);
+        assertTrue("Should match uppercase 'Z'", runAutomaton.run("Z"));
+        assertTrue("Should match lowercase 'z'", runAutomaton.run("z"));
+        assertFalse("Should not match 'y'", runAutomaton.run("y"));
+    }
+
+    /**
+     * Test that toCaseInsensitiveChar works correctly for Turkish characters.
+     * This is a regression test for bug #20843 where case-insensitive wildcard queries
+     * failed for languages with special case-folding rules.
+     */
+    public void testToCaseInsensitiveCharTurkish() {
+        // Test Turkish dotted capital I (İ, U+0130)
+        // In Turkish: İ (U+0130) lowercases to i (U+0069)
+        // Note: Character.toLowerCase uses locale-independent rules
+        Automaton automaton = AutomatonQueries.toCaseInsensitiveChar('\u0130'); // İ
+        CharacterRunAutomaton runAutomaton = new CharacterRunAutomaton(automaton);
+
+        assertTrue("Should match Turkish capital İ (U+0130)", runAutomaton.run("\u0130"));
+        assertTrue("Should match lowercase i (U+0069)", runAutomaton.run("i"));
+
+        // Test lowercase Turkish dotless i (ı, U+0131)
+        automaton = AutomatonQueries.toCaseInsensitiveChar('\u0131'); // ı
+        runAutomaton = new CharacterRunAutomaton(automaton);
+        assertTrue("Should match Turkish lowercase ı (U+0131)", runAutomaton.run("\u0131"));
+        assertTrue("Should match uppercase I (U+0049)", runAutomaton.run("I"));
+    }
+
+    /**
+     * Test that toCaseInsensitiveChar works correctly for Cyrillic characters (Ukrainian/Russian).
+     * This is part of the fix for bug #20843.
+     */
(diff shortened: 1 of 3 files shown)


msfroh:
UnicodeRootLetterICaseConversions

Not sure if this is why you closed this PR, but the reason why (on the linked issue) I suggested that we need a separate "turkic" mode for case-insensitive matching is because the diagram above represents the Unicode ROOT locale behavior.

The correct behavior for Turkish is vertical arrows pointing back and forth between the two left boxes, and vertical arrows pointing back and forth between the two right boxes. There should be no arrows between the left and right boxes.

server/src/test/java/org/opensearch/common/lucene/search/AutomatonQueriesTests.java
```diff
@@ -0,0 +1,119 @@
...
+     * Test the toCaseInsensitiveString method with mixed scripts
+     */
+    public void testToCaseInsensitiveStringMixed() {
+        // Test a string with mixed ASCII and non-ASCII
+        Automaton automaton = AutomatonQueries.toCaseInsensitiveString("İstanbul");
+        CharacterRunAutomaton runAutomaton = new CharacterRunAutomaton(automaton);
+
+        assertTrue("Should match original 'İstanbul'", runAutomaton.run("İstanbul"));
+        assertTrue("Should match lowercase 'istanbul'", runAutomaton.run("istanbul"));
+        assertTrue("Should match uppercase 'ISTANBUL'", runAutomaton.run("ISTANBUL"));
```
urmichm, line 117: Why do you expect this case to match? 
The original word `İstanbul` is in Turkish, with Turkish letter `İ`; it mustn't match the English translation and English letter `I`.

urmichm:
@msfroh regarding a separate "turkic" mode, are there already language-specific modes?

msfroh:
> @msfroh regarding a separate "turkic" mode, are there already language-specific modes?

@urmichm Not yet. AFAIK, from my discussion about this on Lucene issues, this ambiguity only applies to languages with the dotless i. Checking [Wikipedia](https://en.wikipedia.org/wiki/Dotless_I), it says:
> I, or ı, called dotless i, is a letter used in the [Latin-script alphabets](https://en.wikipedia.org/wiki/Latin-script_alphabet) of [Azerbaijani](https://en.wikipedia.org/wiki/Azerbaijani_alphabet), [Crimean Tatar](https://en.wikipedia.org/wiki/Crimean_Tatar_alphabet), [Gagauz](https://en.wikipedia.org/wiki/Gagauz_alphabet), [Kazakh](https://en.wikipedia.org/wiki/Kazakh_alphabets), [Tatar](https://en.wikipedia.org/wiki/Tatar_alphabet) and [Turkish](https://en.wikipedia.org/wiki/Turkish_alphabet). 

So, I don't think we need language-specific modes across the board, but rather I think we need three "leaf" modes:
1. Case-sensitive: This a byte-for-byte match of the Unicode.
2. Case-insenstive, which needs to be disambiguated:
    1. ROOT: Use the ROOT locale for case conversions (with the case-folding rules from Lucene's `Automata.makeCaseInsensitiveChar()`).
    2. Turkic: Use the alternate dotted vs dotless `i` conversions (and `Automata.makeCaseInsensitiveChar()` for everything else).

PR #20930, opened 2026-03-19
Add parquet-data-format plugin

### LINKED ISSUE #21058: [RFC] Pluggable Parquet Engine for OpenSearch
### Summary

This issue tracks the implementation of a Parquet data format plugin (parquet-data-format) for OpenSearch, built on top of the pluggable component architecture proposed in #20644. The plugin provides a complete IndexingExecutionEngine implementation that writes OpenSearch documents into Apache Parquet files using Apache Arrow as the in-memory columnar representation and a 
native Rust writer for Parquet encoding.

### Background

As outlined in the [Pluggable Component Bundling RFC (#20644)](https://github.com/opensearch-project/OpenSearch/issues/20644), OpenSearch's tight coupling to Apache Lucene limits its ability to support workloads optimized for different storage formats. The RFC proposes a DataFormat and IndexingExecutionEngine abstraction that enables pluggable storage formats to coexist alongside Lucene.

This plugin is the first concrete implementation of that abstraction for a columnar format. It demonstrates that the proposed interfaces are sufficient to build a fully functional, non-Lucene storage engine — validating the extensibility model end-to-end from document ingestion through to durable Parquet file output.

### Architecture Overview

The plugin is structured across three layers:

1. **Java orchestration layer** — Implements the DataFormatPlugin, IndexingExecutionEngine, Writer, and DocumentInput interfaces from the data format framework. Manages Arrow vector batching, schema construction from OpenSearch mappings, and lifecycle coordination.

2. **Arrow buffering layer** — Uses Apache Arrow's VectorSchemaRoot as the in-memory columnar buffer. Documents are written row-at-a-time into Arrow vectors, batched into configurable-size Vector Schema Roots (VSRs), and handed off to the native layer via the Arrow C Data Interface (zero-copy).

3. **Native Rust layer** — A JNI-bridged Rust library (parquet_dataformat_jni) that receives Arrow record batches and writes them to Parquet files using the arrow-rs and parquet-rs crates with LZ4 compression and bloom filters.

#### 1. Engine → Writer → DocumentInput Interaction

This flow shows how the OpenSearch indexing pipeline interacts with the plugin's components. The ParquetIndexingEngine creates per-generation ParquetWriter instances and ParquetDocumentInput
objects. During document parsing, field values are collected into FieldValuePair entries. The completed document is passed to the writer, which delegates to the VSR batching layer.

Image

#### 2. VSR State Machine & Write Flow

The plugin uses a double-buffered VSR pool to overlap document ingestion with native Parquet writes. Each VSR follows a strict ACTIVE → FROZEN → CLOSED lifecycle. When the active VSR reaches
the configured row threshold, it is frozen and submitted to a background thread for native export, while a new active VSR immediately begins accepting writes. The VSR rotation threshold (parquet.max\_rows\_per\_vsr) acts as a soft limit. When the active VSR reaches the threshold but the frozen slot is still occupied by a previous background write, rotation is skipped and the active VSR continues accepting writes beyond the threshold. This avoids blocking the indexing thread on native I/O. The resulting memory growth will be handled by OpenSearch's indexing memory controller, which monitors native memory usage (reported via getNativeBytesUsed()) and triggers a flush when memory pressure exceeds acceptable bounds — ensuring the system self-corrects without stalling document ingestion.

Image

#### 3. Native (Rust) Function Structure & Call Flow

The Java-to-Rust bridge uses the Arrow C Data Interface for zero-copy handoff of columnar data. The Rust side maintains a global writer registry (DashMap) keyed by file path, supporting 
concurrent writers across shards.

Image

Zero copy handoff illustration:

Image

#### 4. Field Mapping & Schema Construction

The plugin maps OpenSearch field types to Apache Arrow vector types through a two-layer design: a pluggable field registry and a schema builder that derives the Arrow schema from the index's
live mappings.

**ArrowFieldRegistry** is a static registry populated at class-load time via ParquetFieldPlugin implementations. Two plugins are registered by default:
- CoreDataFieldPlugin — registers all built-in data types (numeric, text, temporal, boolean, binary)
- MetadataFieldPlugin — registers metadata fields (_id, _routing, _ignored, _size)

Each entry maps an OpenSearch field type name (e.g., "integer", "keyword") to a ParquetField implementation that knows the corresponding Arrow vector type and how to write values into it. 
For example, IntegerParquetField maps to Arrow's IntVector (32-bit signed int) and writes values via IntVector.setSafe(rowIndex, value).

ArrowSchemaBuilder constructs the Arrow Schema from the MapperService at writer creation time. It iterates over all mappers in the document mapping, skips unsupported metadata fields (
_source, _field_names, _index, _nested_path), and looks up each mapper's type name in the ArrowFieldRegistry. For every matched field, it creates an Arrow Field with the mapper's name and 
the corresponding Arrow type (always nullable). A _row_id field (long) is appended to every schema for cross-format document correlation with Lucene.

This design is extensible — new field types can be supported by implementing ParquetFieldPlugin and registering additional ParquetField entries without modifying existing code.

### Components

| Component | Location | Description |
|-----------|----------|-------------|
| `ParquetDataFormatPlugin` | `sandbox/plugins/parquet-data-format` | Plugin entry point. Registers the Parquet `DataFormat`, settings, and thread pool. Creates per-shard `ParquetIndexingEngine` instances. |
| `ParquetIndexingEngine` | `...parquet/engine/` | Per-shard engine implementing `IndexingExecutionEngine`. Manages the shared `ArrowBufferPool`, writer creation, refresh, and file deletion. |
| `ParquetWriter` | `...parquet/writer/` | Per-generation writer implementing `Writer<ParquetDocumentInput>`. Delegates document ingestion to `VSRManager` and produces `FileInfos` on flush. |
| `ParquetDocumentInput` | `...parquet/writer/` | Implements `DocumentInput`. Collects `FieldValuePair` entries during document parsing for transfer to Arrow vectors. |
| `VSRManager` | `...parquet/vsr/` | Orchestrates the Arrow batching → Parquet pipeline. Manages VSR rotation, background native writes, and flush/sync operations. |
| `VSRPool` | `...parquet/vsr/` | Double-buffered pool managing one ACTIVE and one FROZEN `ManagedVSR` slot with row-count-based rotation. |
| `ManagedVSR` | `...parquet/vsr/` | RAII wrapper around Arrow `VectorSchemaRoot` with strict `ACTIVE → FROZEN → CLOSED` lifecycle enforcement and Arrow C Data Interface export. |
| `ArrowFieldRegistry` | `...parquet/fields/` | Registry mapping OpenSearch field type names to `ParquetField` implementations. Populated via `ParquetFieldPlugin` SPI. |
| `ArrowSchemaBuilder` | `...parquet/fields/` | Builds Arrow schemas from `MapperService` field mappings, filtering unsupported metadata fields. |
| `ParquetField` (+ subclasses) | `...parquet/fields/core/` | Abstract base for field-level Arrow vector writes. Concrete implementations for all built-in types: integer, long, float, double, boolean, keyword, text, date, IP, binary, etc. |
| `NativeParquetWriter` | `...parquet/bridge/` | Type-safe Java handle for the Rust writer with lifecycle management (create → write → flush → sync). |
| `RustBridge` | `...parquet/bridge/` | Static JNI bridge to the native Rust library. Package-private writer methods, public utility methods. |
| `ArrowBufferPool` | `...parquet/memory/` | Shared Arrow memory allocator per shard with child allocator isolation per VSR. |
| `native-bridge-spi` | `sandbox/libs/native-bridge-spi` | Shared library providing `PlatformHelper` (native lib loading), `ArrowExport` (C Data Interface RAII container), and logging bridge. |
| `parquet_dataformat_jni` | `...parquet/src/main/rust/` | Rust JNI library. `writer.rs` manages `ArrowWriter` instances via `DashMap`. `jni.rs` provides the JNI entry points. Uses `mimalloc` allocator, LZ4 compression, and bloom filters. |

### Configuration

| Setting | Default | Description |
|---|---|---|
| parquet.max_rows_per_vsr | 50000 | Row threshold triggering VSR rotation and background native write. When rotation is skipped due to an in-flight background write, the active VSR grows beyond this limit until the indexing memory controller triggers a flush. |
| parquet.max_native_allocation | 10% | Maximum native memory allocation per shard in terms of percentage of total native memory |

### Follow-ups

- **Merge support**: Merges is out of scope for this RFC and is tracked for follow-up.

### Related

- #20644
- #18416

Contributors: @rayshrey @raghuvanshraj @darjisagar7 

Refer above

### Related component

Indexing

_No response_

### Additional context

_No response_

rayshrey:
### Description
Introduces parquet-data-format plugin.
For more details see the RFC attached below.

### Related Issues
Resolves #21058

### Check List

### DIFF
--- a/sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/ArrowExport.java
+++ b/sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/ArrowExport.java
@@ -23,27 +23,6 @@ import org.apache.arrow.c.ArrowSchema;
 public record ArrowExport(ArrowArray arrowArray, ArrowSchema arrowSchema) implements AutoCloseable {
 
     /**
-     * Creates a new ArrowExport.
-     *
-     * @param arrowArray  the Arrow array to export
-     * @param arrowSchema the Arrow schema to export
-     */
-    public ArrowExport {
-    }
-
-    /** Returns the arrow array. */
-    @Override
-    public ArrowArray arrowArray() {
-        return arrowArray;
-    }
-
-    /** Returns the arrow schema. */
-    @Override
-    public ArrowSchema arrowSchema() {
-        return arrowSchema;
-    }
-
-    /**
      * Returns the memory address of the Arrow array.
      *
      * @return the native memory address of the array
--- a/sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/ParquetFileMetadata.java
+++ b/sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/ParquetFileMetadata.java
@@ -20,32 +20,4 @@ package org.opensearch.parquet.bridge;
  * @param createdBy creator string from the Parquet file footer metadata
  */
 public record ParquetFileMetadata(int version, long numRows, String createdBy) {
-
-    /**
-     * Creates a new ParquetFileMetadata.
-     *
-     * @param version   the Parquet format version
-     * @param numRows   the total number of rows
-     * @param createdBy the creator string from the file footer
-     */
-    public ParquetFileMetadata {
-    }
-
-    /** Returns the Parquet format version. */
-    @Override
-    public int version() {
-        return version;
-    }
-
-    /** Returns the total number of rows. */
-    @Override
-    public long numRows() {
-        return numRows;
-    }
-
-    /** Returns the creator string from the file footer. */
-    @Override
-    public String createdBy() {
-        return createdBy;
-    }
 }
--- a/sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
+++ b/sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
@@ -116,6 +116,9 @@ public class ParquetIndexingEngine implements IndexingExecutionEngine<ParquetDat
 
     @Override
     public RefreshResult refresh(RefreshInput refreshInput) throws IOException {
+        if (refreshInput == null) {
+            return new RefreshResult(List.of());
+        }
         List<Segment> segments = new ArrayList<>(refreshInput.existingSegments());
         long gen = segments.stream().mapToLong(Segment::generation).max().orElse(-1) + 1;
         for (WriterFileSet wfs : refreshInput.writerFiles()) {
@@ -126,7 +129,7 @@ public class ParquetIndexingEngine implements IndexingExecutionEngine<ParquetDat
 
     @Override
     public long getNextWriterGeneration() {
-        return 0;
+        throw new UnsupportedOperationException("getNextWriterGeneration not supported");
     }
 
     @Override
--- a/gradle/missing-javadoc.gradle
+++ b/gradle/missing-javadoc.gradle
@@ -170,7 +170,9 @@ configure([
   project(":libs:opensearch-common"),
   project(":libs:opensearch-core"),
   project(":libs:opensearch-compress"),
-  project(":server")
+  project(":server"),
+  project(":sandbox:libs:native-bridge-spi"),
+  project(":sandbox:plugins:parquet-data-format")
 ]) {
   project.tasks.withType(MissingJavadocTask) {
     // TODO: bump to variable missing level after increasing javadoc coverage
--- a/sandbox/plugins/parquet-data-format/src/test/java/org/opensearch/parquet/bridge/NativeParquetWriterTests.java
+++ b/sandbox/plugins/parquet-data-format/src/test/java/org/opensearch/parquet/bridge/NativeParquetWriterTests.java
@@ -56,8 +56,6 @@ public class NativeParquetWriterTests extends OpenSearchTestCase {
         super.tearDown();
     }
 
-    // --- Normal lifecycle: create → write → close → flush ---
-
     public void testFullLifecycle() throws Exception {
         String filePath = createTempDir().resolve("full.parquet").toString();
         NativeParquetWriter writer = createWriter(filePath);
@@ -100,8 +98,6 @@ public class NativeParquetWriterTests extends OpenSearchTestCase {
         assertTrue("Parquet file should exist after flush", Files.exists(Path.of(filePath)));
     }
 
-    // --- Close without writes ---
-
     public void testFlushWithoutWrite() throws Exception {
         String filePath = createTempDir().resolve("close-only.parquet").toString();
         NativeParquetWriter writer = createWriter(filePath);
@@ -119,17 +115,15 @@ public class NativeParquetWriterTests extends OpenSearchTestCase {
         assertSame(first, writer.getMetadata());
     }
 
-    // --- Flush auto-closes if not yet closed ---
-
     public void testSyncAutoFlushesIfNotFlushed() throws Exception {
-        String filePath = createTempDir().resolve("auto-close.parquet").toString();
+        String filePath = createTempDir().resolve("auto-flush.parquet").toString();
         NativeParquetWriter writer = createWriter(filePath);
 
         try (ArrowExport export = exportData(new int[] { 1 }, new String[] { "alice" }, new long[] { 10L })) {
             writer.write(export.getArrayAddress(), export.getSchemaAddress());
         }
 
-        // flush without explicit close — should auto-close first
+        // sync without explicit close — should auto-close first
         assertNull(writer.getMetadata());
         writer.sync();
         assertNotNull(writer.getMetadata());
@@ -138,7 +132,7 @@ public class NativeParquetWriterTests extends OpenSearchTestCase {
     }
 
     public void testWriteAfterFlushThrows() throws Exception {
-        String filePath = createTempDir().resolve("write-after-close.parquet").toString();
+        String filePath = createTempDir().resolve("write-after-flush.parquet").toString();
         NativeParquetWriter writer = createWriter(filePath);
         writer.flush();
 
@@ -146,8 +140,6 @@ public class NativeParquetWriterTests extends OpenSearchTestCase {
             expectThrows(IOException.class, () -> writer.write(export.getArrayAddress(), export.getSchemaAddress()));
         }
     }
-
-    // --- Error cases: invalid paths, addresses, schema mismatch ---
 
     public void testCreateWriterWithNonExistentDirectory() {
         expectThrows(IOException.class, () -> {
@@ -179,7 +171,7 @@ public class NativeParquetWriterTests extends OpenSearchTestCase {
             writer.write(export.getArrayAddress(), export.getSchemaAddress());
         }
 
-        // Close fails — Parquet ArrowWriter detects row count mismatch between columns
+        // Flush fails — Parquet ArrowWriter detects row count mismatch between columns
         expectThrows(IOException.class, writer::flush);
     }
 
@@ -194,7 +186,7 @@ public class NativeParquetWriterTests extends OpenSearchTestCase {
     }
 
     public void testSyncCalledTwice() throws Exception {
-        String filePath = createTempDir().resolve("double-flush.parquet").toString();
+        String filePath = createTempDir().resolve("double-sync.parquet").toString();
         NativeParquetWriter writer = createWriter(filePath);
 
         try (ArrowExport export = exportData(new int[] { 1 }, new String[] { "alice" }, new long[] { 10L })) {
--- a/sandbox/plugins/parquet-data-format/src/test/java/org/opensearch/parquet/engine/ParquetIndexingEngineTests.java
+++ b/sandbox/plugins/parquet-data-format/src/test/java/org/opensearch/parquet/engine/ParquetIndexingEngineTests.java
@@ -14,6 +14,7 @@ import org.opensearch.common.settings.Settings;
 import org.opensearch.core.index.shard.ShardId;
 import org.opensearch.index.engine.dataformat.FileInfos;
 import org.opensearch.index.engine.dataformat.RefreshInput;
+import org.opensearch.index.engine.dataformat.RefreshResult;
 import org.opensearch.index.engine.dataformat.Writer;
 import org.opensearch.index.mapper.KeywordFieldMapper;
 import org.opensearch.index.mapper.MappedFieldType;
@@ -120,12 +121,18 @@ public class ParquetIndexingEngineTests extends OpenSearchTestCase {
         assertTrue(engine.refresh(RefreshInput.builder().build()).refreshedSegments().isEmpty());
     }
 
+    public void testRefreshWithNullInput() throws Exception {
+        RefreshResult result = engine.refresh(null);
+        assertNotNull(result);
+        assertTrue(result.refreshedSegments().isEmpty());
+    }
+
     public void testGetMergerReturnsNull() {
         assertNull(engine.getMerger());
     }
 
-    public void testGetNextWriterGenerationReturnsZero() {
-        assertEquals(0, engine.getNextWriterGeneration());
+    public void testGetNextWriterGenerationThrows() {
+        expectThrows(UnsupportedOperationException.class, () -> engine.getNextWriterGeneration());
     }
 
     public void testDeleteFiles() throws Exception {


sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Bridge class that allows Rust code to log messages through Java's logging framework via JNI.
+ *
+ * <p>Rust calls this using the fully qualified class name:
+ * {@code org/opensearch/nativebridge/spi/RustLoggerBridge}
+ */
+public class RustLoggerBridge {
+
+    private static final Logger logger = LogManager.getLogger(RustLoggerBridge.class);
```
alchemist51, line 23: This will always have RustLoggerBridge as the class, is it possible to pass the main class from the caller to understand the origin class of the log?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetDataFormatPlugin.java
```diff
@@ -0,0 +1,48 @@
...
+import org.opensearch.index.shard.ShardPath;
+import org.opensearch.parquet.engine.ParquetDataFormat;
+import org.opensearch.parquet.engine.ParquetIndexingEngine;
+import org.opensearch.parquet.fields.ArrowSchemaBuilder;
+import org.opensearch.plugins.Plugin;
+
+/**
+ * Parquet data format plugin for OpenSearch.
+ */
+public class ParquetDataFormatPlugin extends Plugin implements DataFormatPlugin {
```
alchemist51, line 53: Can you define: extendedPlugins = ['composite-engine'] in your build.gradle to make the composite engine plugin aware of this? until @mgodwan  changes for Dataformat registry merge in?

sandbox/plugins/parquet-data-format/src/main/rust/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+#[cfg(any(test, feature = "test-utils"))]
+pub mod test_utils;
+
+#[cfg(test)]
+mod tests;
+
+pub use native_bridge_spi::{log_info, log_error, log_debug};
+
+lazy_static! {
+    pub static ref WRITER_MANAGER: DashMap<String, Arc<Mutex<ArrowWriter<File>>>> = DashMap::new();
```
alchemist51: Are these specific to each IndexShard or are at node level? Should we scope this per IndexShard/Engine?

sandbox/plugins/parquet-data-format/src/main/rust/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+    pub static ref FILE_MANAGER: DashMap<String, File> = DashMap::new();
+}
+
+pub struct NativeParquetWriter;
```
alchemist51: Do we want to write a trait which each writer have to implement?

sandbox/plugins/parquet-data-format/src/main/rust/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+
+impl NativeParquetWriter {
```
alchemist51: nit: writer logic could be moved to a writer.rs file and lib.rs should just define it.

sandbox/plugins/parquet-data-format/src/main/rust/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+            ],
+        )?;
+        Ok(java_metadata)
+    }
+}
+
+// --- JNI exports ---
+
+#[unsafe(no_mangle)]
+pub extern "system" fn Java_org_opensearch_parquet_bridge_RustBridge_initLogger(
```
alchemist51: Do we want to move jni exports also to a separate file?

sandbox/plugins/parquet-data-format/src/main/rust/src/lib.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+#[unsafe(no_mangle)]
+pub extern "system" fn Java_org_opensearch_parquet_bridge_RustBridge_createWriter(
+    mut env: JNIEnv,
+    _class: JClass,
+    file: JString,
+    schema_address: jlong,
+) -> jint {
+    let filename: String = env.get_string(&file).expect("Couldn't get java string!").into();
```
alchemist51: What happens when a failure happens and expect is called? Does it fail the shard? Or is it just a RunTime exception? How do we distinguish?

sandbox/libs/native-bridge-spi/src/main/rust/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+impl LogLevel {
+    fn as_i32(self) -> i32 {
+        self as i32
+    }
+}
+
+/// Initialize the logger from a JNIEnv reference.
+pub fn init_logger_from_env(env: &JNIEnv) {
```
alchemist51, line 32: Which module will initialise it? How can we use this across components? We should make it idempotent otherwise

sandbox/libs/native-bridge-spi/src/main/rust/src/lib.rs
```diff
@@ -0,0 +1,83 @@
...
+
+#[macro_export]
+macro_rules! log_error {
+    ($($arg:tt)*) => {
+        $crate::__internal_log($crate::LogLevel::Error, &format!($($arg)*))
+    };
+}
+
+#[doc(hidden)]
+pub fn __internal_log(level: LogLevel, message: &str) {
```
alchemist51, line 94: Why this needs to be pub?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,61 @@
...
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Type-safe handle for native Parquet writer with lifecycle management.
+ * Delegates to {@link RustBridge} for actual native calls and handles error checking.
+ */
+public class NativeParquetWriter implements Closeable {
+
+    private final AtomicBoolean writerClosed = new AtomicBoolean(false);
```
alchemist51: Should we use SetOnce instead to make sure error is thrown if we are trying to close a already closed Writer?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,61 @@
...
+    private final String filePath;
+    private ParquetFileMetadata metadata;
```
alchemist51: Should this be final?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,61 @@
...
+        int result = RustBridge.flushToDisk(filePath);
+        if (result != 0) {
+            throw new IOException("Failed to flush Parquet file to disk: " + filePath);
+        }
+    }
+
+    @Override
+    public void close() {
+        if (writerClosed.compareAndSet(false, true)) {
+            metadata = RustBridge.closeWriter(filePath);
```
alchemist51: Why are we tying the closeWriter with metadata? Should we have an separate call for fetching metadata?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/RustBridge.java
```diff
@@ -0,0 +1,68 @@
...
+        String platformDir = PlatformHelper.getPlatformDirectory();
+        String libFileName = PlatformHelper.getPlatformLibraryName(LIB_NAME);
+        String resourcePath = "/native/" + platformDir + "/" + libFileName;
+
+        try (InputStream is = RustBridge.class.getResourceAsStream(resourcePath)) {
+            if (is == null) {
+                System.loadLibrary(LIB_NAME);
+                return;
+            }
+            Path tempDir = Files.createTempDirectory("opensearch-native-");
```
alchemist51: nit: do we need the `-` in the end?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public void deleteFiles(Map<String, Collection<String>> filesToDelete) throws IOException {
+        Collection<String> parquetFiles = filesToDelete.get(dataFormat.name());
+        if (parquetFiles == null) {
+            return;
+        }
+        for (String fileName : parquetFiles) {
+            Path filePath = Path.of(fileName);
+            logger.info("Deleting parquet file: {}", filePath);
```
alchemist51: nit: this could be debug

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public void flush() throws IOException {
+        int result = RustBridge.flushToDisk(filePath);
+        if (result != 0) {
+            throw new IOException("Failed to flush Parquet file to disk: " + filePath);
+        }
+    }
+
+    @Override
+    public void close() {
```
alchemist51: This API could be confusing to the user in terms of when metadata is being created, should we rather have a metadata return from the close() call?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
```diff
@@ -0,0 +1,162 @@
...
+                writer.write(export.getArrayAddress(), export.getSchemaAddress());
+            }
+            vsrPool.completeVSR(frozenVSR);
+            vsrPool.unsetFrozenVSR();
+        }
+        ManagedVSR newVSR = vsrPool.getActiveVSR();
+        if (newVSR == null) {
+            throw new IOException("No active VSR available after rotation");
+        }
+        managedVSR.set(newVSR);
```
Bukhtawar, line 163: Do we need the export to be done synchronously since this might block indexing threads?

sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
rayshrey: By original class, do you mean the Java class that calls the native jni functions or the jni functions itself ?

sandbox/libs/native-bridge-spi/src/main/rust/src/lib.rs
rayshrey: Any module can initialise it at this point.
But only the first one that will call it will succeed. Others will be a no-op as the J`AVA_VM `is a OnceLock.
Calling set on an already initialized OnceLock will return an `Err` but the `.ok` call will silently discard that error making it a no-op.

sandbox/libs/native-bridge-spi/src/main/rust/src/lib.rs
rayshrey: Because the macros `log_info!, log_debug!, log_error!` are exported via `#[macro_export]` and expand in the calling crate (e.g. parquet-data-format). When the macro expands there, it generates
$crate::__internal_log(...) — which is a call to this function itself. For that call to compile from an external crate, the function must be pub.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: No harm in calling close mutiple times as it is idempotent.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: We get the metadata only once the writer is closed.
So no point in making it final as we cannot assign it's value in constructor.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/RustBridge.java
rayshrey: Better to add it since the `opensearch-native-` is a prefix which will be used to create a temp directory and not the full path name.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
rayshrey: Ack, changing it

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
rayshrey: Yes makes sense, adding changes to execute the native write in a different background thread.

sandbox/plugins/parquet-data-format/src/main/rust/src/lib.rs
rayshrey: At this level we are not aware of any Shard/Engine, the Rust layer's job is to mostly create the parquet files at the specified path, write to them and finalize those files. Adding IndexShard/Engine visibility will only complicate the logic without much usefulness.

sandbox/plugins/parquet-data-format/src/main/rust/src/lib.rs
rayshrey: Ack, making those changes.

sandbox/plugins/parquet-data-format/src/main/rust/src/lib.rs
rayshrey: Ack, making the changes.

sandbox/plugins/parquet-data-format/build.gradle
```diff
@@ -0,0 +1,174 @@
...
+  ]
+  for (String path : possiblePaths) {
+    if (new File(path).exists()) {
+      return path
+    }
+  }
+  return 'cargo'
+}
+
+tasks.register('installRust', Exec) {
```
alchemist51, line 89: Do we have such tasks in Java side as well?

rayshrey: I don't think we will have such tasks for Java since gradle itself relies on JDK and cannot function without it

alchemist51: I'm not 100% sure if the build.gradle should do all the installations steps as well. Personally, I feel it's job is to just build.. and the user should be responsible for setting up the dependencies. For example, in rustc also we will have more dependencies as we go(datafusion need protobuf) so will we keep adding all of that in the build.gradle? 

@Bukhtawar  @mgodwan  opinion on this?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
alchemist51: But is that the right behaviour? Should it be called multiple times?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
alchemist51: Can the close return the metadata instead of initialising it at close time and then maintaining it in the writer?

sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
alchemist51: I think we need a way to figure out the file from which the error originate(in rust) or the caller Java class(origin). This will get confusing since multiple flows will be using the same RustLoggerBridge otherwise.

sandbox/libs/native-bridge-spi/src/main/rust/src/lib.rs
alchemist51: we can simply make it no-op in case JAVA_VM is already set? and we can treat the error in setting up as an genuine error? wdyt?

sandbox/libs/native-bridge-spi/src/main/rust/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "native-bridge-spi"
+version = "0.1.0"
+edition = "2021"
+description = "Shared Rust utilities for OpenSearch sandbox native plugins"
+license = "Apache-2.0"
+
+[lib]
+crate-type = ["rlib"]
```
alchemist51, line 9: just for my understanding, what does crate-type do?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/RustBridge.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+public class RustBridge {
+
+    private static final String LIB_NAME = "parquet_dataformat_jni";
+
+    static {
+        loadNativeLibrary();
+    }
+
+    private static void loadNativeLibrary() {
```
alchemist51: Can this function go to PlatformHelper lib?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/fields/ArrowSchemaBuilder.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Creates an Arrow Schema from the MapperService.
+     */
+    public static Schema getSchema(MapperService mapperService) {
+        Objects.requireNonNull(mapperService, "MapperService cannot be null");
+        List<Field> fields = new ArrayList<>();
+        for (Mapper mapper : mapperService.documentMapper().mappers()) {
+            if (isUnsupportedMetadataField(mapper)) {
+                continue;
```
alchemist51, line 49: Should we add a log line here for debugging purpose?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/memory/ArrowBufferPool.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+public class ArrowBufferPool implements Closeable {
+
+    private static final Logger logger = LogManager.getLogger(ArrowBufferPool.class);
+
+    private final RootAllocator rootAllocator;
+    private final long maxChildAllocation;
+
+    public ArrowBufferPool() {
+        long maxAllocationInBytes = 10L * 1024 * 1024 * 1024;
```
alchemist51: nit: could be a static constant, if we want to make it configurable, then let's expose a setting also?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public FieldVector getVector(String fieldName) {
+        if (state != VSRState.ACTIVE) {
+            throw new IllegalStateException("Cannot access vector in VSR state: " + state);
+        }
+        Field field = fields.get(fieldName);
+        return field != null ? vsr.getVector(field) : null;
+    }
+
+    public void moveToFrozen() {
```
alchemist51, line 101: Should this be synchronised? Also can we make the VSRState as an Atomic state?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Exports this VSR to Arrow C Data Interface for native handoff.
+     * Only allowed when VSR is FROZEN.
+     */
+    public ArrowExport exportToArrow() {
+        if (state != VSRState.FROZEN) {
+            throw new IllegalStateException("Cannot export VSR in state: " + state + ". Must be FROZEN.");
+        }
+        ArrowArray arrowArray = ArrowArray.allocateNew(allocator);
```
alchemist51, line 116: Can allocateNew throw error?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public void close() {
+        if (state == VSRState.CLOSED) {
+            return;
+        }
+        if (state == VSRState.ACTIVE) {
+            throw new IllegalStateException("Cannot close VSR " + id + ": must freeze first");
+        }
+        state = VSRState.CLOSED;
+        logger.debug("State transition: FROZEN -> CLOSED for VSR {}", id);
```
alchemist51, line 160: do we want to add a assert above? for FROZEN? just to make sure in case some new state arises, the test will start failing.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
```diff
@@ -0,0 +1,153 @@
...
+    private final int maxRowsPerVSR;
+
+    public VSRPool(String poolId, Schema schema, ArrowBufferPool bufferPool) {
+        this.poolId = poolId;
+        this.schema = schema;
+        this.bufferPool = bufferPool;
+        this.activeVSR = new AtomicReference<>();
+        this.frozenVSR = new AtomicReference<>();
+        this.vsrCounter = new AtomicInteger(0);
+        this.maxRowsPerVSR = 50000;
```
alchemist51: nit: could be a constant

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
```diff
@@ -0,0 +1,153 @@
...
+        ManagedVSR current = activeVSR.get();
+        if (current == null || !shouldRotateVSR(current)) {
+            return false;
+        }
+        if (frozenVSR.get() != null) {
+            throw new IOException("Cannot rotate VSR: frozen slot is occupied");
+        }
+        synchronized (this) {
+            current = activeVSR.get();
+            if (current == null || !shouldRotateVSR(current)) {
```
alchemist51: we have a check for this above as well? Can state change in between also?

sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/ArrowExport.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * <p>Wraps an {@link ArrowArray} and {@link ArrowSchema} allocated via the Arrow C Data Interface,
+ * providing memory address accessors for JNI handoff and automatic resource cleanup on close.
+ * Used by {@link org.opensearch.parquet.vsr.ManagedVSR} to export data and schema to the native
+ * Parquet writer without copying.
+ *
+ * <p>Implements {@link AutoCloseable} to ensure native Arrow memory is released and freed
+ * even if an exception occurs during the write path.
+ */
+public record ArrowExport(ArrowArray arrowArray, ArrowSchema arrowSchema) implements AutoCloseable {
```
mgodwan, line 23: is this a general representation to export arrow? Can this be moved to a lib?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    /**
+     * Writes an Arrow batch to the Parquet file.
+     *
+     * @param arrayAddress  the native memory address of the Arrow array
+     * @param schemaAddress the native memory address of the Arrow schema
+     * @throws IOException if the write fails or the writer is closed
+     */
+    public void write(long arrayAddress, long schemaAddress) throws IOException {
```
mgodwan, line 55: Should this be a synchronized method as we expect only 1 thread to call this?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/ParquetFileMetadata.java
```diff
@@ -0,0 +1,56 @@
...
+        return numRows;
+    }
+
+    /** Returns the creator string from the file footer. */
+    @Override
+    public String createdBy() {
+        return createdBy;
+    }
+
+    @Override
```
mgodwan: Why do we need to write these methods? Isn't record providing these?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/RustBridge.java
mgodwan: +1, this can be a common lib

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetDataFormat.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public long priority() {
+        return 0;
+    }
+
+    @Override
+    public Set<FieldTypeCapabilities> supportedFields() {
+        return Set.of();
```
mgodwan: Why is this empty?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/FieldValuePair.java
```diff
@@ -0,0 +1,60 @@
...
+    private final MappedFieldType fieldType;
+    private final Object value;
+
+    /**
+     * Creates a new FieldValuePair.
+     *
+     * @param fieldType the mapped field type
+     * @param value the parsed field value
+     */
+    public FieldValuePair(MappedFieldType fieldType, Object value) {
```
mgodwan, line 35: how will arrays be handled?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/memory/ArrowBufferPool.java
```diff
@@ -0,0 +1,76 @@
...
+ * Arrow memory allocator pool with configurable limits derived from node settings.
+ *
+ * <p>Wraps an Apache Arrow {@link RootAllocator} whose maximum allocation is computed as a
+ * percentage of available non-heap system memory (total physical memory minus JVM max heap),
+ * controlled by the {@code parquet.max_native_allocation} setting (default {@code "10%"}).
+ *
+ * <p>Child allocators are created per {@link org.opensearch.parquet.vsr.ManagedVSR} instance,
+ * each limited to 1/10th of the root allocation, providing memory isolation between batches.
+ */
+public class ArrowBufferPool implements Closeable {
```
alchemist51, line 33: Do we need to extend a Cleaner like how we are doing in search:
```
 /**
 * Base class for type-safe native pointer wrappers.
 * Provides automatic resource management and prevents use-after-close errors.
 * Subclasses must implement {@link #doClose()} to release native resources.
 * Cleaner is used to ensure resources are cleaned up even if the object is not explicitly closed.
 */
```

Code: https://github.com/bharath-techie/OpenSearch/blob/581dacad0f7baf008850cbb4459e95c0270ae152/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/jni/NativeHandle.java

sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
rayshrey: For Rust side, we can explicitly add the origin in the log message itself, but from the Java side all calls are from the RustBridge itself (since that is the class which contains all the native methods). We can iterate over the stack trace to get the origin of who called the RustBridge methods, but iterating over the stacktrace would over-complicating things and will most likely an overhead as well in terms of performance as well. For logging we want it to be as minimal as possible.

For now, adding explicit messages on Rust side to clearly identify the path should be sufficient in my opinion. We can take up more changes on the logger side incrementally - there are some more issues as well such as lazy loading. Will take-up all logger related changes as follow-ups.

sandbox/libs/native-bridge-spi/src/main/rust/Cargo.toml
rayshrey: Setting crate-type = ["rlib"] means this crate produces a .rlib file that can only be consumed by other Rust crates as a dependency. It's not directly loadable by C/Java/JNI.

"rlib" is applicable for native-bridge-spi since it's a shared utility crate consumed by the parquet plugin's Rust code — not loaded directly via JNI. 

The parquet plugin itself uses "cdylib" to produce the .so/.dylib that Java loads.

For reference, the common crate types:
- rlib — Rust-to-Rust dependency (what you have)
- cdylib — C-compatible shared library (.so/.dylib) for JNI/FFI
- staticlib — C-compatible static library (.a)
- lib — lets the compiler choose (defaults to rlib)

sandbox/libs/native-bridge-spi/src/main/rust/src/lib.rs
rayshrey: Have made some changes to return early in case JAVA_VM is already set.
We are not explicitly setting up anything here, things such as attaching the current jvm thread and finding relevant logger classes etc are done in the log method itself. In this we just set up the SetOnce variable JAVA_VM - the only error we can encounter there is it being already set and nothing more.

sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/ArrowExport.java
rayshrey: To create an ArrowWriter we need a ArrowSchema and to write data we need ArrowArray and ArrowSchema both. These are specific to the ingestion part and might not be much useful for other flows. Hence kept them here only.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: All methods of the Native Parquet Writer are supposed to be synchronized.
Documented this class as not being thread safe.
Should we strictly enforce making every method synchronized (might have a very small and unnecessary overhead though) ?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: The getFileMetdata is particularly useful in tests for assertions and that is the main reason I decided on keeping it this way.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: I do not see any issue in calling close multiple times.
Most of the classes in OpenSearch just are a NoOp for close operations (such as TranslogWriter etc)

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/ParquetFileMetadata.java
rayshrey: javadocs check requires javadocs for accessors of records as well, hence had to explicitly write them

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/RustBridge.java
rayshrey: Sure, makes sense. Moved it to PlatformHelper

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetDataFormat.java
rayshrey: Added.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/fields/ArrowSchemaBuilder.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/memory/ArrowBufferPool.java
rayshrey: In the class referenced above (NativeHandle) it is an abstract class and it's lifecycle might be very different from ours. In our case, ArrowBufferPool is a concrete class managed and maintained by the ParquetIndexingEngine - created and closed by the engine itself. Cleaner usually makes sense for objects whose lifecycle is not properly controlled, which is not the case here.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/FieldValuePair.java
rayshrey: To handle arrays, we will need the value to be a List<Object> instead of Object.
While this may be doable, handling arrays would require making every Field in the Schema explicitly to be a List<> and writing to these Lists field comes with a slight performance overhead.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
rayshrey: According to the function signature, they do not throw any IOException or any other exception.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
rayshrey: Not supposed to be called concurrently. 
But for better guardrails made the state volatile.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
rayshrey: Done

sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
mgodwan: Lets add a todo to solve this aftet the scope of this PR

sandbox/plugins/parquet-data-format/benchmarks/src/main/java/org/opensearch/parquet/benchmark/ListVsScalarBenchmark.java
```diff
@@ -0,0 +1,182 @@
...
+                        case 0:
+                            ((IntVector) lv.getDataVector()).setSafe(childIdx, row + f);
+                            break;
+                        case 1:
+                            ((BigIntVector) lv.getDataVector()).setSafe(childIdx, (long) row * 100 + f);
+                            break;
+                        default:
+                            byte[] val = ("value_" + row).getBytes(StandardCharsets.UTF_8);
+                            ((VarCharVector) lv.getDataVector()).setSafe(childIdx, val);
+                            break;
```
mgodwan: [minor] this can be moved to a common method

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
mgodwan: Can we make it setonce?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @param filePath      the path to the Parquet file to write
+     * @param schemaAddress the native memory address of the Arrow schema
+     * @throws IOException if the native writer creation fails
+     */
+    public NativeParquetWriter(String filePath, long schemaAddress) throws IOException {
+        this.filePath = filePath;
+        int result = RustBridge.createWriter(filePath, schemaAddress);
+        if (result != 0) {
+            throw new IOException("Failed to create native Parquet writer for: " + filePath);
```
mgodwan: We should throw an exception from rust directly here

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,97 @@
...
+        int result = RustBridge.write(filePath, arrayAddress, schemaAddress);
+        if (result != 0) {
+            throw new IOException("Failed to write data to Parquet file: " + filePath);
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
+        if (writerClosed.compareAndSet(false, true)) {
+            metadata = RustBridge.closeWriter(filePath);
```
mgodwan: Can we rename finalizeWriter?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    /**
+     * Syncs the Parquet file to disk. Must be called after {@link #close()}.
+     * If close has not been called yet, it will be called first.
+     */
+    public void flush() throws IOException {
```
mgodwan: Can we rename this to sync?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -0,0 +1,97 @@
...
+            throw new IOException("Cannot write to closed Parquet writer: " + filePath);
+        }
+        int result = RustBridge.write(filePath, arrayAddress, schemaAddress);
+        if (result != 0) {
+            throw new IOException("Failed to write data to Parquet file: " + filePath);
+        }
+    }
+
+    @Override
+    public void close() throws IOException {
```
mgodwan: Can we rename this to flush? close impl can essentially do the same

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetDataFormat.java
```diff
@@ -0,0 +1,50 @@
...
+    public long priority() {
+        return 0;
+    }
+
+    @Override
+    public Set<FieldTypeCapabilities> supportedFields() {
+        return ArrowFieldRegistry.getRegisteredFields()
+            .keySet()
+            .stream()
+            .map(type -> new FieldTypeCapabilities(type, Set.of(FieldTypeCapabilities.Capability.COLUMNAR_STORAGE)))
```
mgodwan: Can we add a todo to override FieldRegistry to return capability for each field?

mgodwan: Also, lets add bloom filter

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetDataFormatPlugin.java
mgodwan: No, DataFormatRegistry should handle this

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/ParquetWriter.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @Override
+    public FileInfos flush() throws IOException {
+        ParquetFileMetadata metadata = vsrManager.flush();
+        if (file == null || metadata == null) {
+            return FileInfos.empty();
+        }
+        Path filePath = Path.of(file);
+        WriterFileSet writerFileSet = WriterFileSet.builder()
+            .directory(filePath.getParent())
```
mgodwan: Shall we return path relative to shard root directory only here?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
```diff
@@ -0,0 +1,146 @@
...
+    @Override
+    public void deleteFiles(Map<String, Collection<String>> filesToDelete) throws IOException {
+        Collection<String> parquetFiles = filesToDelete.get(dataFormat.name());
+        if (parquetFiles == null) {
+            return;
+        }
+        for (String fileName : parquetFiles) {
+            Path filePath = Path.of(fileName);
+            logger.debug("Deleting parquet file: {}", filePath);
+            Files.deleteIfExists(filePath);
```
mgodwan: Can we log if file was not deleted?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public Merger getMerger() {
+        return null;
+    }
+
+    @Override
+    public RefreshResult refresh(RefreshInput refreshInput) throws IOException {
+        return new RefreshResult(Collections.emptyList());
```
mgodwan: We should just return a RefreshResult containing existing files from input, and adding the new file based on refresh input from here

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
```diff
@@ -0,0 +1,170 @@
...
+        ArrowArray arrowArray = ArrowArray.allocateNew(allocator);
+        ArrowSchema arrowSchema = ArrowSchema.allocateNew(allocator);
+        Data.exportVectorSchemaRoot(allocator, vsr, null, arrowArray, arrowSchema);
+        return new ArrowExport(arrowArray, arrowSchema);
+    }
+
+    /**
+     * Exports only the schema to Arrow C Data Interface.
+     */
+    public ArrowExport exportSchema() {
```
mgodwan: This can just return the schema, right?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
```diff
@@ -0,0 +1,170 @@
...
+    @Override
+    public void close() {
+        if (state == VSRState.CLOSED) {
+            return;
+        }
+        if (state == VSRState.ACTIVE) {
+            throw new IllegalStateException("Cannot close VSR " + id + ": must freeze first");
+        }
+        assert state == VSRState.FROZEN : "Expected FROZEN state before closing VSR " + id + " but was " + state;
+        state = VSRState.CLOSED;
```
mgodwan: This should be handled through compareAndSet semantics

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
```diff
@@ -0,0 +1,181 @@
...
+ *
+ * <p>Handles row-count-based rotation: when the active VSR reaches the configured
+ * {@code maxRowsPerVSR} threshold, it is frozen and a new active VSR is created.
+ * The frozen slot must be drained (exported and closed) before the next rotation
+ * can occur; attempting to rotate with an occupied frozen slot throws {@link java.io.IOException}.
+ *
+ * <p>Each new VSR receives its own child allocator from the shared {@link ArrowBufferPool},
+ * ensuring memory isolation between batches.
+ */
+public class VSRPool implements AutoCloseable {
```
mgodwan, line 33: Mark this class as not thread safe

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
```diff
@@ -0,0 +1,181 @@
...
+     * @return true if rotation occurred
+     * @throws IOException if rotation needed but frozen slot is occupied
+     */
+    public boolean maybeRotateActiveVSR() throws IOException {
+        ManagedVSR current = activeVSR.get();
+        if (current == null || !shouldRotateVSR(current)) {
+            return false;
+        }
+        if (frozenVSR.get() != null) {
+            throw new IOException("Cannot rotate VSR: frozen slot is occupied");
```
mgodwan: return false here

mgodwan: We should let IMC handle it

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+        if (current.getRowCount() > 0) {
+            freezeVSR(current);
+        }
+        activeVSR.set(createNewVSR());
```
mgodwan: We should do a compareAndSet here

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * Clears the frozen VSR slot after it has been closed.
+     *
+     * @throws IOException if the frozen slot is empty or the VSR is not closed
+     */
+    public void unsetFrozenVSR() throws IOException {
+        ManagedVSR frozen = frozenVSR.get();
+        if (frozen == null) {
+            throw new IOException("unsetFrozenVSR called when frozen VSR is not set");
```
mgodwan: We can throw an IllegalStateException

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
```diff
@@ -0,0 +1,249 @@
...
+     *
+     * @return metadata about the written Parquet file, or null if no data to flush
+     */
+    public ParquetFileMetadata flush() throws IOException {
+        awaitPendingWrite();
+        ManagedVSR currentVSR = managedVSR.get();
+        if (currentVSR == null || currentVSR.getRowCount() == 0) {
+            logger.debug("No data to flush for {}", fileName);
+            return null;
+        }
```
mgodwan: Can this return null if currentVSR is null, but writes happened on previous VSRs from pool? That shouldn't happen ideally as we should return metadata for what was written till now

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
```diff
@@ -0,0 +1,249 @@
...
+        activeVSR.setRowCount(rowIndex + 1);
+    }
+
+    /**
+     * Handles VSR rotation after document addition if row threshold is reached.
+     * Submits frozen VSR write to the background thread pool or runs it on the calling thread,
+     * depending on the {@code runAsync} setting.
+     */
+    public void maybeRotateActiveVSR() throws IOException {
+        awaitPendingWrite();
```
mgodwan: This would cause writes to wait. We need to avoid this

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
```diff
@@ -0,0 +1,249 @@
...
+    /**
+     * Waits for any in-flight background write to complete.
+     * Propagates exceptions from the background write as IOException.
+     */
+    private void awaitPendingWrite() throws IOException {
+        if (pendingWrite == null) {
+            return;
+        }
+        try {
+            pendingWrite.get();
```
mgodwan: We should have a timeout here, and fail if write timeout. We can keep a timeout of 120s.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetDataFormat.java
rayshrey: Done.

sandbox/libs/native-bridge-spi/src/main/java/org/opensearch/nativebridge/spi/ArrowExport.java
rayshrey: Moved it to native-bridge-spi

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: Change it to be SetOnce

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: Done.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: Done.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
rayshrey: On further thoughts made it Atomic

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
rayshrey: Yes, updated it.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/ManagedVSR.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
rayshrey: Fixed this

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
rayshrey: Added support for timeout.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRPool.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/ParquetWriter.java
rayshrey: Done.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
rayshrey: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
mgodwan:
```
            awaitPendingWrite(ROTATION_TIMEOUT, true);
```

Close seems to be ignoring timeout

PR #20936, opened 2026-03-20
Make warmup settings dynamic for pull-based ingestion

kaustubhbutte17:
## Description
Follow-up to #20526. Changes the two warmup settings from `Final` (immutable after index creation) to `Dynamic` (updatable at runtime via the update settings API).

This allows operators to adjust warmup behavior without recreating the index.

## Settings Changed
| Setting | Before | After |
|---------|--------|-------|
| `index.ingestion_source.warmup.timeout` | Final | **Dynamic** |
| `index.ingestion_source.warmup.lag_threshold` | Final | **Dynamic** |

## Use Cases
- **Disable warmup mid-progress** - emergency escape hatch if warmup is taking too long
- **Adjust lag threshold** - tune without recreating the index
- **Change timeout** - increase/decrease without recreating the index

## Dynamic Update Behavior
| Scenario | Behavior |
|----------|----------|
| Disable warmup while in progress (timeout -> -1) | Warmup completes immediately, shard starts serving |
| Change threshold while warmup in progress | New threshold takes effect on next check cycle |
| Change timeout while warmup in progress | New timeout takes effect on next check cycle |
| Enable warmup after shard is already STARTED | Does NOT re-trigger warmup (would be disruptive) |

## Settings Propagation Path
```
Update Settings API -> IngestionEngine.registerDynamicIndexSettingsHandlers() -> DefaultStreamPoller.updateWarmupConfig()
```

## Testing
- `testUpdateWarmupConfigDisableWhileInProgress` - disabling warmup via config update
- `testUpdateWarmupConfigDisableWithRunningPoller` - disabling mid-warmup with running poller, verifies updateWarmupStatus disabled branch
- `testUpdateWarmupConfigThresholdAndTimeoutWhileInProgress` - threshold and timeout update during warmup
- `testUpdateWarmupConfigDoesNotReEnableAfterCompletion` - no re-trigger after shard is serving
- `testWarmupSkippedWhenPollerStartsInPausedState` - paused poller skips warmup
- `testDynamicWarmupSettingsUpdate` (IT) - end-to-end dynamic settings update via API

### DIFF
--- a/server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
+++ b/server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
@@ -1092,6 +1092,7 @@ public class DefaultStreamPollerTests extends OpenSearchTestCase {
 
         // Wait for the poll loop to call updateWarmupStatus() which counts down the latch
         assertTrue("Warmup should complete after being dynamically disabled", warmupPoller.awaitWarmupComplete(30000));
+        assertTrue(warmupPoller.isWarmupComplete());
 
         warmupPoller.close();
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add support for enabling pluggable data formats, starting with phase-1 of decoupling shard from engine, and introducing basic abstractions ([#20675](https://github.com/opensearch-project/OpenSearch/pull/20675))
 
 - Add warmup phase to wait for lag to catch up in pull-based ingestion before serving ([#20526](https://github.com/opensearch-project/OpenSearch/pull/20526))
-- Make warmup settings dynamic for pull-based ingestion ([#20936](https://github.com/opensearch-project/OpenSearch/pull/20936))
 ### Changed
 - Make telemetry `Tags` immutable ([#20788](https://github.com/opensearch-project/OpenSearch/pull/20788))
 - Move Randomness from server to libs/common ([#20570](https://github.com/opensearch-project/OpenSearch/pull/20570))


server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -430,6 +430,31 @@ public boolean awaitWarmupComplete(long timeoutMs) throws InterruptedException {
...
+     * Updates the warmup configuration dynamically.
+     * Called when index settings are changed at runtime.
+     */
+    @Override
+    public void updateWarmupConfig(IngestionSource.WarmupConfig newConfig) {
+        IngestionSource.WarmupConfig oldConfig = this.warmupConfig;
+        this.warmupConfig = newConfig;
+
+        // If warmup was enabled and is now disabled, mark as complete
+        if (oldConfig.isEnabled() && !newConfig.isEnabled() && !warmupComplete) {
```
varunbharadwaj: Would it be possible to just set the warmupConfig to new config here? If warmup is in-progress and we disable warmup, in the next polling loop when `warmupComplete=false`, we can update the logic to still call updateWarmupStatus which would just mark `warmupComplete=true`.  That way the warmupLatch will count down in a single path avoiding confusion. Let me know if this can have any other problems or if you think otherwise.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -3338,6 +3339,19 @@ public void onSettingsChanged() {
                 indexSettings.getSoftDeleteRetentionOperations()
             );
         }
+
+        // Update warmup config if this is an ingestion engine
+        Indexer indexer = getIndexerOrNull();
```
varunbharadwaj: Do we need all these changes? I think we just need to do the following.

1. IndexMetadata - need to change final to dynamic, as you have already done so.
2. DefaultStreamPoller - expose a updateWarmupConfig
3. In IngestionEngine, we already have a `registerDynamicIndexSettingsHandlers`. Here we can add something like 

```
engineConfig.getIndexSettings()
        .getScopedSettings()
        .addSettingsUpdateConsumer(
            IndexMetadata.INGESTION_SOURCE_WARMUP_TIMEOUT_SETTING,
            IndexMetadata.INGESTION_SOURCE_WARMUP_LAG_THRESHOLD_SETTING,
            this::updateWarmupConfig
        );
```

I think we can avoid changes in IndexShard and other layers with this. Let me know if we run into any issues with this, requiring those changes.

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
```diff
@@ -983,4 +983,182 @@ public void testWarmupAwaitReturnsImmediatelyWhenAlreadyComplete() throws Interr
...
+        );
+
+        // Warmup should not be complete yet
+        assertFalse(warmupPoller.isWarmupComplete());
+
+        // Update timeout to a longer value while warmup is in progress
+        IngestionSource.WarmupConfig updatedConfig = new IngestionSource.WarmupConfig(TimeValue.timeValueMinutes(10), 100L);
+        warmupPoller.updateWarmupConfig(updatedConfig);
+
+        // Warmup should still not be complete (we just changed timeout)
```
varunbharadwaj: If we do not start the poller, i think the timeout doesn't matter and warmUp will never be complete? Seems like `testUpdateWarmupConfigThresholdWhileInProgress` also tests the same but for threshold. We could combine these two tests by updating both timeout and threshold in a single test.

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
```diff
@@ -1376,11 +1376,11 @@ public void testWarmupPhase() throws Exception {
         });
 
         // Step 5: Validate all 10 documents are searchable after warmup
-        // Refresh to make all docs visible, then assert directly (not waitForSearchableDocs)
-        // since all 10 docs should already be indexed during the warmup phase
-        client(nodeA).admin().indices().prepareRefresh(indexName).get();
-        long docCount = client(nodeA).prepareSearch(indexName).setSize(0).get().getHits().getTotalHits().value();
-        assertEquals("All 10 documents should be searchable immediately after warmup completes", 10L, docCount);
+        assertBusy(() -> {
```
varunbharadwaj: Curious why we switch to using assertBusy here? Do we observe any failures/flaky behavior without this? We should not be needing assertBusy here.

varunbharadwaj: Discussed offline - to avoid flakiness, we will change the logic to validate polled message count instead and then wait for 10 docs to be visible.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
kaustubhbutte17: Reverted IndexSettings and IndexShard changes. Now using registerDynamicIndexSettingsHandlers in IngestionEngine with a two-setting consumer, thanks for the suggestion.

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
kaustubhbutte17: done

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
```diff
@@ -1375,18 +1375,61 @@ public void testWarmupPhase() throws Exception {
...
-        client(nodeA).admin().indices().prepareRefresh(indexName).get();
-        long docCount = client(nodeA).prepareSearch(indexName).setSize(0).get().getHits().getTotalHits().value();
-        assertEquals("All 10 documents should be searchable immediately after warmup completes", 10L, docCount);
-
-        // Step 6: Verify stats
+        // Step 5: Verify poller metric confirms all 10 messages were processed during warmup
         PollingIngestStats stats = client(nodeA).admin().indices().prepareStats(indexName).get().getIndex(indexName).getShards()[0]
             .getPollingIngestStats();
         assertNotNull(stats);
         assertEquals(10L, stats.getMessageProcessorStats().totalProcessedCount());
```
varunbharadwaj: We can remove totalProcessedCount validation, as it'll have the same flaky behavior

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
```diff
@@ -983,4 +983,189 @@ public void testWarmupAwaitReturnsImmediatelyWhenAlreadyComplete() throws Interr
...
+            1000,
+            1000,
+            10000,
+            indexSettings,
+            new DefaultIngestionMessageMapper(),
+            enabledConfig
+        );
+
+        // Warmup should not be complete yet
+        assertFalse(warmupPoller.isWarmupComplete());
```
varunbharadwaj, line 1048: Do we have a race condition here? Is it guaranteed warmup will not be complete?

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
```diff
@@ -983,4 +983,189 @@ public void testWarmupAwaitReturnsImmediatelyWhenAlreadyComplete() throws Interr
...
+
+        // Dynamically disable warmup (timeout=-1)
+        IngestionSource.WarmupConfig disabledConfig = new IngestionSource.WarmupConfig(TimeValue.timeValueMillis(-1), 100L);
+        warmupPoller.updateWarmupConfig(disabledConfig);
+
+        // isWarmupComplete() checks warmupConfig.isEnabled(), so disabling makes it return true immediately
+        assertTrue(warmupPoller.isWarmupComplete());
```
varunbharadwaj, line 1055: Does this need any changes to wait for the warmup status to update in next polling loop?

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
```diff
@@ -983,4 +983,189 @@ public void testWarmupAwaitReturnsImmediatelyWhenAlreadyComplete() throws Interr
...
+        warmupPoller.start();
+        assertFalse(warmupPoller.isWarmupComplete());
+
+        // Dynamically disable warmup - updateWarmupStatus will handle completion on next poll loop
+        warmupPoller.updateWarmupConfig(new IngestionSource.WarmupConfig(TimeValue.timeValueMillis(-1), 100L));
+
+        // isWarmupComplete() returns true immediately via config check, but we need to verify
+        // the poll loop also calls updateWarmupStatus() which counts down the latch.
+        // Wait briefly for the poll loop to iterate and hit the disabled branch.
+        Thread.sleep(2000);
```
varunbharadwaj: Can we replace the explicit sleep by updating to warmupPoller.awaitWarmupComplete(30s)

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
```diff
@@ -983,4 +983,189 @@ public void testWarmupAwaitReturnsImmediatelyWhenAlreadyComplete() throws Interr
...
+            1000,
+            1000,
+            10000,
+            indexSettings,
+            new DefaultIngestionMessageMapper(),
+            initialConfig
+        );
+
+        // Warmup should not be complete yet
+        assertFalse(warmupPoller.isWarmupComplete());
```
varunbharadwaj, line 1122: Same question here, if we have a race condition.

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
kaustubhbutte17: Updated

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
kaustubhbutte17: The poller is not started (start() is not called). The constructor sets warmupComplete = false because warmup is enabled. Without starting the poller, no poll loop runs, so warmupComplete stays false. hence assertFalse(isWarmupComplete()) works

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
kaustubhbutte17: No because isWarmupComplete() returns warmupComplete || !warmupConfig.isEnabled(). After updateWarmupConfig(disabledConfig), !warmupConfig.isEnabled() is true immediately. So there is no need to wait for the poll loop.

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
kaustubhbutte17: Due to same reason as above (poller is not started) there wont be any race condition

server/src/test/java/org/opensearch/indices/pollingingest/DefaultStreamPollerTests.java
```diff
@@ -983,4 +983,184 @@ public void testWarmupAwaitReturnsImmediatelyWhenAlreadyComplete() throws Interr
...
+
+        // Start the poller
+        warmupPoller.start();
+        assertFalse(warmupPoller.isWarmupComplete());
+
+        // Dynamically disable warmup - updateWarmupStatus will handle completion on next poll loop
+        warmupPoller.updateWarmupConfig(new IngestionSource.WarmupConfig(TimeValue.timeValueMillis(-1), 100L));
+
+        // Wait for the poll loop to call updateWarmupStatus() which counts down the latch
+        assertTrue("Warmup should complete after being dynamically disabled", warmupPoller.awaitWarmupComplete(30000));
```
varunbharadwaj, line 1094: if it timeout and warmup is not complete, do we want to check - asserttrue(warmupPoller.isWarmupComplete());

kaustubhbutte17: sure, added

PR #20952, opened 2026-03-21
CCR: Remove non-primary origin assertions from operation planners

mishail:
Disclaimer: portions of this text were reformatted and edited with the assistance of generative AI tools for clarity and readability. The analysis and technical content originate from the author.

### Description
OpenSearch commit 59be6aee8a07cc858757f49ec7cc30e0586651c9 / https://github.com/opensearch-project/OpenSearch/pull/20585 refactored `InternalEngine` to extract its two large inner static classes into separate top-level files:

- `InternalEngine.IndexingStrategy` → top-level `IndexingStrategy extends OperationStrategy`
- `InternalEngine.DeletionStrategy` → top-level `DeletionStrategy extends OperationStrategy`

It also introduced `IndexingStrategyPlanner` and `DeletionStrategyPlanner` — classes that encapsulate the planning logic previously inlined in `InternalEngine`.

The refactor broke CCR's `ReplicationEngine` in two ways. See https://github.com/opensearch-project/cross-cluster-replication/issues/1646

1. Compilation failure

`ReplicationEngine.kt` references `IndexingStrategy` and `DeletionStrategy`. These were previously inner static classes of `InternalEngine`; they are now top-level classes in `org.opensearch.index.engine` and require explicit imports. Two lines change on the CCR side.

2. Runtime assertion failure
This affects a lot of tests in CCR at the moment. 
```
java.lang.AssertionError: planing as primary but got PRIMARY
    at org.opensearch.index.engine.OperationStrategyPlanner.assertNonPrimaryOrigin(OperationStrategyPlanner.java:41)
    at org.opensearch.index.engine.IndexingStrategyPlanner.planOperationAsNonPrimary(IndexingStrategyPlanner.java:162)
    at org.opensearch.index.engine.InternalEngine.planIndexingAsNonPrimary(InternalEngine.java:1030)
    at org.opensearch.replication.ReplicationEngine.indexingStrategyForOperation(ReplicationEngine.kt:34)
    at org.opensearch.index.engine.InternalEngine.index(InternalEngine.java:929)
    ...
    at org.opensearch.replication.action.replay.TransportReplayChangesAction.performOnPrimary(TransportReplayChangesAction.kt:126)
```

And that happens because of introduction of `OperationStrategyPlanner`s and more specifically because those planners now have their own `assertNonPrimaryOrigin`, which CCR's `ReplicationEngine` has no way to override.

The old way:
- `ReplicationEngine.indexingStrategyForOperation` always calls `planIndexingAsNonPrimary`
- `ReplicationEngine.assertNonPrimaryOrigin` overrides `InternalEngine.assertNonPrimaryOrigin` to return `true`, bypassing the origin check inside `InternalEngine.planIndexingAsNonPrimary` and `planDeletionAsNonPrimary`

After the refactor, `InternalEngine.planIndexingAsNonPrimary` and `planDeletionAsNonPrimary` delegate to private `IndexingStrategyPlanner` / `DeletionStrategyPlanner` instances. Both planners call `assertNonPrimaryOrigin(operation)` internally, which comes from `OperationStrategyPlanner`, which is entirely separate from `InternalEngine.assertNonPrimaryOrigin`. So `ReplicationEngine.assertNonPrimaryOrigin` is never invoked, and the assertion fires.

## Proposed solution:

Remove the `assert assertNonPrimaryOrigin(...)` calls from `IndexingStrategyPlanner.planOperationAsNonPrimary` (line 162) and `DeletionStrategyPlanner.planOperationAsNonPrimary` (line 117).

**Rationale:** These asserts call `assertNonPrimaryOrigin` through the `OperationStrategyPlanner` interface's default method, a path that bypasses `InternalEngine.assertNonPrimaryOrigin` and any overrides defined in `InternalEngine` subclasses (such as `ReplicationEngine`). This breaks `InternalEngine`'s established extensibility contract: subclasses that legitimately override `assertNonPrimaryOrigin` can no longer influence this check. Removing the asserts from the planners restores that contract. The check that matters — the one that IS overridable — remains in `InternalEngine.planIndexingAsNonPrimary` and `planDeletionAsNonPrimary`.

**Pros:**
- Minimal change: 2 assert lines removed in OpenSearch, 2 imports added in CCR
- Restores the pre-existing extensibility contract without adding new API surface
- Rollback is trivial

**Cons:**
- Removes a precondition guard from the planner classes themselves; planners no longer self-enforce that they are called in the correct context
- Does not provide a clean extension point for future subclasses that may need similar customization of planner behavior

### Related Issues
Resolves https://github.com/opensearch-project/cross-cluster-replication/issues/1646

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,16 +21,20 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - WLM group custom search settings - groundwork and timeout ([#20536](https://github.com/opensearch-project/OpenSearch/issues/20536))
 - Expose JVM runtime metrics via telemetry framework ([#20844](https://github.com/opensearch-project/OpenSearch/pull/20844))
 - Add intra segment support for single-value metric aggregations ([#20503](https://github.com/opensearch-project/OpenSearch/pull/20503))
+- Add new setting property 'Sensitive' for tiering dynamic settings ([#20901](https://github.com/opensearch-project/OpenSearch/pull/20901))
 - Add ref_path support for package-based hunspell dictionary loading ([#20840](https://github.com/opensearch-project/OpenSearch/pull/20840))
 - Add support for enabling pluggable data formats, starting with phase-1 of decoupling shard from engine, and introducing basic abstractions ([#20675](https://github.com/opensearch-project/OpenSearch/pull/20675))
 
 - Add warmup phase to wait for lag to catch up in pull-based ingestion before serving ([#20526](https://github.com/opensearch-project/OpenSearch/pull/20526))
+- Add a new static method to IndicesOptions API to expose `STRICT_EXPAND_OPEN_HIDDEN_FORBID_CLOSED` index option ([#20980](https://github.com/opensearch-project/OpenSearch/pull/20980))
+
 ### Changed
 - Make telemetry `Tags` immutable ([#20788](https://github.com/opensearch-project/OpenSearch/pull/20788))
 - Move Randomness from server to libs/common ([#20570](https://github.com/opensearch-project/OpenSearch/pull/20570))
 - Use env variable (OPENSEARCH_FIPS_MODE) to enable opensearch to run in FIPS enforced mode instead of checking for existence of bcFIPS jars ([#20625](https://github.com/opensearch-project/OpenSearch/pull/20625))
 - Update streaming flag to use search request context ([#20530](https://github.com/opensearch-project/OpenSearch/pull/20530))
 - Move pull-based ingestion classes from experimental to publicAPI ([#20704](https://github.com/opensearch-project/OpenSearch/pull/20704))
+- Lazy init stored field reader in SourceLookup ([#20827](https://github.com/opensearch-project/OpenSearch/pull/20827))
 
 ### Fixed
 - Relax index template pattern overlap check to use minimum-string heuristic, allowing distinguishable multi-wildcard patterns at the same priority ([#20702](https://github.com/opensearch-project/OpenSearch/pull/20702))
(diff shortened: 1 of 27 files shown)


mishail:
Another option, if we do want to keep those assertions in Planners, would be adding `protected` factory methods to `InternalEngine` that create the planner instances, allowing subclasses to substitute custom planner implementations:

```java
// In InternalEngine:
protected IndexingStrategyPlanner createIndexingStrategyPlanner(...) {
    return new IndexingStrategyPlanner(...);
}
protected DeletionStrategyPlanner createDeletionStrategyPlanner(...) {
    return new DeletionStrategyPlanner(...);
}
```

mishail:
cc @shank9918

cwperks:
@mishail Can you please sync with main? There are a few test fixes that this feature branch would need to get for CI checks to pass.

cwperks:
> java.lang.AssertionError: planing as primary but got PRIMARY

@mishail Can we please fix this typo while at it? maybe even change the text entirely. At least for me, its hard to tell what the purpose is. 

After inspecting the call sites, I agree these calls are redundant.

cwperks:
@mishail FYI I updated to include the latest commits on main which includes test fixes. You may need to pull before push. I left one comment about updating an error message. That can either be addressed here or a separate PR, its not a blocker.

cwperks:
@mishail how does this fix the issue in CCR repo exactly? I see that they always call [planIndexingAsNonPrimary](https://github.com/opensearch-project/cross-cluster-replication/blob/9b4cec2b7dc79add8aebeb5dd6d2c3abb2befa8b/src/main/kotlin/org/opensearch/replication/ReplicationEngine.kt#L32) regardless of origin type. Presumably the fix lies in CCR repo?

mishail:
> @mishail how does this fix the issue in CCR repo exactly? I see that they always call [planIndexingAsNonPrimary](https://github.com/opensearch-project/cross-cluster-replication/blob/9b4cec2b7dc79add8aebeb5dd6d2c3abb2befa8b/src/main/kotlin/org/opensearch/replication/ReplicationEngine.kt#L32) regardless of origin type. Presumably the fix lies in CCR repo?

@cwperks The problem is that this piece in their ReplicationEngine doesn't work the way it used to work anymore.

```kotlin
    override fun assertNonPrimaryOrigin(operation: Operation): Boolean {
        return true
    }
```

> After the refactor, InternalEngine.planIndexingAsNonPrimary and planDeletionAsNonPrimary delegate to private IndexingStrategyPlanner / DeletionStrategyPlanner instances. Both planners call assertNonPrimaryOrigin(operation) internally, which comes from OperationStrategyPlanner, which is entirely separate from InternalEngine.assertNonPrimaryOrigin. So ReplicationEngine.assertNonPrimaryOrigin is never invoked, and the assertion fires.

cwperks:
@mishail I'm not deep in this area unfortunately, but can the ReplicationEngine in CCR follow same pattern as InternalEngine?

See comment here: https://github.com/opensearch-project/cross-cluster-replication/pull/1647/changes#r2982353370

```
protected DeletionStrategy deletionStrategyForOperation(final Delete delete) throws IOException {
    if (delete.origin() == Operation.Origin.PRIMARY) {
        return planDeletionAsPrimary(delete);
    } else {
        // non-primary mode (i.e., replica or recovery)
        return planDeletionAsNonPrimary(delete);
    }
}
```

mishail:
@cwperks are we going to use https://github.com/opensearch-project/cross-cluster-replication/pull/1647 to fix the issue?

cwperks:
> @cwperks are we going to use [opensearch-project/cross-cluster-replication#1647](https://github.com/opensearch-project/cross-cluster-replication/pull/1647) to fix the issue?

yes and I think that's the appropriate place to put it because this is logic specific to the CCR repo. My understanding from @mohit10011999 is that in CCR all indexing operations to the follower cluster are to be treated as replicas so the nature of that fix is to "spoof" the index operation to always say its origin replica by nature of it being an indexing op on the following cluster.

cwperks:
Do let me know if you have other thoughts though.

PR #20956, opened 2026-03-22
Fix version for disable_objects bug tests

andrross:
### Check List


andrross:
See the build failure here: https://build.ci.opensearch.org/job/gradle-check/73026/testReport/

@newtonne Please correct me if I'm wrong but I believe this test will only reliably work with 3.6+ versions that have the fix.

newtonne:
@andrross Yeah, that’s right. Thanks for the fix. And thanks for the review on the other PR 👍

PR #20969, opened 2026-03-23
Adding dsl-query-executor plugin to sandbox

vinaykpud:
### Description

### Related Issues

### Check List

### DIFF
--- a/sandbox/libs/analytics-framework/build.gradle
+++ b/sandbox/libs/analytics-framework/build.gradle
@@ -30,19 +30,7 @@ dependencies {
   runtimeOnly 'com.google.guava:failureaccess:1.0.2'
   // SLF4J — Calcite's logging facade
   runtimeOnly "org.slf4j:slf4j-api:${versions.slf4j}"
-
-  // Calcite's BuiltInMethod and DefaultRelMetadataProvider eagerly resolve methods via
-  // reflection during RelOptCluster.create(). These optional calcite-core transitive deps
-  // must be in the same classloader as Calcite (i.e., here, not in child plugins).
-  runtimeOnly "commons-codec:commons-codec:${versions.commonscodec}"
-  runtimeOnly "org.apache.commons:commons-lang3:${versions.commonslang}"
   runtimeOnly "commons-io:commons-io:${versions.commonsio}"
-  runtimeOnly 'org.codehaus.janino:janino:3.1.9'
-  runtimeOnly 'org.codehaus.janino:commons-compiler:3.1.9'
-  runtimeOnly 'org.jooq:joou-java-6:0.9.4'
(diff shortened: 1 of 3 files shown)


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/MetricTranslator.java
```diff
@@ -0,0 +1,50 @@
...
+import org.opensearch.dsl.aggregation.AggregationType;
+import org.opensearch.dsl.converter.ConversionException;
+import org.opensearch.search.aggregations.AggregationBuilder;
+import org.opensearch.search.aggregations.InternalAggregation;
+
+/**
+ * Translates a metric aggregation (AVG, SUM, MIN, MAX, etc.) to a Calcite AggregateCall,
+ * and converts raw result values back to OpenSearch InternalAggregation for response building.
+ */
+public interface MetricTranslator<T extends AggregationBuilder> extends AggregationType<T> {
```
nssuresh2007, line 22: Can we rename this as `SingleValueMetricTranslator` since this returns single aggregate call, whereas for multi-value metric aggregations (like stats) we need multiple aggregate calls to be returned and that would need a different base class?

vinaykpud: For that i thought we can have like this

* `MetricTranslator` as top level interface (we may still need this)
* `SingleMetricTranslator` and all single value metrics will extend this
* `MultiValueMetricTranslator`, all multi value metrics will extend this

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
```diff
@@ -0,0 +1,118 @@
...
+        this.planExecutor = new DslQueryPlanExecutor(executor);
+        this.clusterService = clusterService;
+        this.indexNameExpressionResolver = indexNameExpressionResolver;
+    }
+
+    @Override
+    protected void doExecute(Task task, SearchRequest request, ActionListener<SearchResponse> listener) {
+        try {
+            String indexName = resolveToSingleIndex(request);
+            long startTime = System.currentTimeMillis();
```
Bukhtawar, line 81: Lets use nanoTime as wall clock time for measuring elapsed time

vinaykpud: I will update this in https://github.com/opensearch-project/OpenSearch/pull/20916

PR #20974, opened 2026-03-23
Replace CHANGELOG.md with release-notes/AGENTS.md

andrross:
Use an AI agent steering file to generate release notes from commit messages, PR descriptions, and PR labels.

[Here is an example PR](https://github.com/andrross/OpenSearch/pull/216) using kiro-cli with Claude Opus 4.6

This has quite a lot of overlap with the [existing AI-generated release notes](https://github.com/opensearch-project/opensearch-build/blob/main/src/llms/prompts.py#L34) tooling. There are key differences in the prompt. Namely that for this repo I think it's pretty important to filter out commits that are not important to users, as opposed to the strict one-entry-per-commit policy. There are also a policies like filtering out commit/revert pairs, collapsing multiple version updates to the same dependency, etc. This version also puts explanation about judgement calls into the PR description, giving human reviews some specific things to look at.

The other main difference is that this relies on an agent to do the work of getting commits and scraping the GitHub API, as opposed to scripting it. There are some subtleties with getting the correct commits for a release given that the previous release branch overlaps with main for the duration of the release window. The agent seemed to be pretty good at figuring this out.

I would like to see if we can integrate some of these changes into the automation in the build repo and then maybe not need this. At least in the short term this seems like a pretty good improvement over the CHANGELOG!

### Check List


andrross:
@peterzhuamazon @gaiksaya @rishabh6788 What do you think about integrating some of the things in the prompt here into https://github.com/opensearch-project/opensearch-build/blob/main/src/llms/prompts.py ?

andrross:
Closing in favor of https://github.com/opensearch-project/opensearch-build/pull/6055

PR #20976, opened 2026-03-23
Make stale segments cleanup logic depend on map size as well

### LINKED ISSUE #20960: [BUG] Memory build up due to stale segment cleanup not triggering in Remote Store domains
### Describe the bug

In RemoteSegmentStoreDirectory class we maintain a map of the segments that have been uploaded to remote - [segmentsUploadedToRemoteStore](https://github.com/opensearch-project/OpenSearch/blob/fb5d661e5b47846585974cae492f228652e72420/server/src/main/java/org/opensearch/index/store/RemoteSegmentStoreDirectory.java#L115). Entries are added to this map when the [segments are uploaded](https://github.com/opensearch-project/OpenSearch/blob/fb5d661e5b47846585974cae492f228652e72420/server/src/main/java/org/opensearch/index/store/RemoteSegmentStoreDirectory.java#L726)(which is done after refresh) and entries are removed from this map in the [first refresh post commit](https://github.com/opensearch-project/OpenSearch/blob/fb5d661e5b47846585974cae492f228652e72420/server/src/main/java/org/opensearch/index/shard/RemoteStoreRefreshListener.java#L248).

When continuous refreshes keep happening but flushes do not happen, this map starts growing boundlessly which causes a boundless increase in JVM as well.

The specific condition that we are talking about (refreshes happen continuously but flushes do not happen at all) can happen in cases where updates runs frequently, almost every second (so flushOnIdle() condition isn't triggered) but the number of documents being updated are low. Due to this the merge size logic for flush isn't triggered as well. Also since it is a remote domain the translog size threshold for flush isn't breached as well since every refresh trims the translog as well. But since updates increase deleted docs count leading to frequent merges, leading to new small segments being created frequently.

### Related component

Storage:Remote

### To Reproduce

1. Create a remote store enabled index
2. Ingest some documents to that index
3. Start a script to keep updating the documents (previously ingested in step 2) every second. Ensure that you keep running this script continuously for days
4. Monitor the remote store path configured for number of current files present and observe it increasing steadily
5. Monitor the jvm of the domain as well and observe it increasing (might take a week or so to observe the steady increase)

I was able to run the above steps and could confirm no flushes happening for the said index and number of files in remote store growing to 13k in a period of 24hrs.

### Expected behavior

The `segmentsUploadedToRemoteStore` should be cleaned up independent of the flush logic to avoid the memory increase issue.

### Additional Details

 - Reproduced on Linux EC2 with latest OpenSearch code

**Additional context**
N/A

rayshrey:
### Description
Currently stale segment cleanup logic is triggered only during the first refresh after a flush.
With this PR we are adding an additional setting that the cleanup can also get triggered when the map reaches a specific threshold. This threshold is backed by a setting, which can be set to -1 as well to disable this entire flow and fallback to the previous logic.

### Related Issues
Resolves #20960 

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Delegate getMin/getMax methods for ExitableTerms ([#20775](https://github.com/opensearch-project/OpenSearch/pull/20775))
 - Fix terms lookup subquery fetch limit reading from non-existent index setting instead of cluster `max_clause_count` ([#20823](https://github.com/opensearch-project/OpenSearch/pull/20823))
 - Fix array_index_out_of_bounds_exception with wildcard and aggregations ([#20842](https://github.com/opensearch-project/OpenSearch/pull/20842))
+- Fix stale segment cleanup logic for remote store ([#20976](https://github.com/opensearch-project/OpenSearch/pull/20976))
 - Ensure that transient ThreadContext headers with propagators survive restore ([#169373](https://github.com/opensearch-project/OpenSearch/pull/20854))
 - Handle dependencies between analyzers ([#19248](https://github.com/opensearch-project/OpenSearch/pull/19248))
 - Fix `_field_caps` returning empty results and corrupted field names for `disable_objects: true` mappings ([#20800](https://github.com/opensearch-project/OpenSearch/pull/20800))


server/src/main/java/org/opensearch/indices/RemoteStoreSettings.java
```diff
@@ -174,6 +174,19 @@ public class RemoteStoreSettings {
...
 
+    /**
+     * Controls the threshold for the number of segments uploaded to remote store map.
+     * When the map size exceeds this threshold, stale segment cleanup is triggered even without a flush/commit.
+     * {@code -1} disables threshold-based cleanup.
+     */
+    public static final Setting<Integer> CLUSTER_REMOTE_UPLOADED_SEGMENTS_CLEANUP_THRESHOLD_SETTING = Setting.intSetting(
+        "cluster.remote_store.uploaded_segments_cleanup_threshold",
+        10000,
+        -1,
```
skumawat2025, line 188: We should have a constraint to avoid setting this to very low threshold value. This would trigger cleanup on each refresh and will increase s3 calls.

rayshrey: Have modified the validator to have a min and max value for the setting (apart from the -1 to disable it)

server/src/main/java/org/opensearch/index/shard/RemoteStoreRefreshListener.java
```diff
@@ -244,7 +244,9 @@ private boolean syncSegments() {
                 // if a new segments_N file is present in local that is not uploaded to remote store yet, it
                 // is considered as a first refresh post commit. A cleanup of stale commit files is triggered.
                 // This is done to avoid delete post each refresh.
-                if (isRefreshAfterCommit()) {
+                // Also trigger cleanup if the uploaded segments map exceeds the configured threshold,
+                // to prevent unbounded memory growth when flushes do not happen.
+                if (isRefreshAfterCommit() || uploadedSegmentsMapExceedsThreshold()) {
                     remoteDirectory.deleteStaleSegmentsAsync(indexShard.getRemoteStoreSettings().getMinRemoteSegmentMetadataFiles());
```
Bukhtawar, line 250: Shouldn't this like a background cleanup that cleans up stale segments periodically rather than after a threshold has breached

gbbafna: That might be wasteful when we have large number of idle shards and also when there are frequent writes to the index . 

where do you think background clean up might help ?

PR #20946, opened 2026-03-20
Extract some workflow steps to Github and exclude from Jenkins runner

cwperks:
### Description

This PR is part of a series of 3:

- https://github.com/opensearch-project/OpenSearch/pull/20946
- https://github.com/opensearch-project/opensearch-build/pull/6043
- https://github.com/opensearch-project/opensearch-build-libraries/pull/814

Currently trying to experiment with ways to increase the build's success percentage. See the daily report [here](https://metrics.opensearch.org/_dashboards/app/dashboards#/view/87737cf0-1409-11f1-a611-eb8fce42240d?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-7d,to:now))&_a=(description:'',filters:!(('$state':(store:appState),bool:(minimum_should_match:1,should:!((bool:(must:!((term:(invoke_type.keyword:Timer)),(term:(git_reference.keyword:main))))),(bool:(must:!((term:(invoke_type.keyword:'Post%20Merge%20Action')),(prefix:(pull_request_title.keyword:'push%20trigger%20main'))))))),meta:(alias:!n,disabled:!f,index:'2e729590-eb26-11ee-b008-6583d3dd450c',key:bool,negate:!f,type:custom,value:'%7B%22should%22:%5B%7B%22bool%22:%7B%22must%22:%5B%7B%22term%22:%7B%22invoke_type.keyword%22:%22Timer%22%7D%7D,%7B%22term%22:%7B%22git_reference.keyword%22:%22main%22%7D%7D%5D%7D%7D,%7B%22bool%22:%7B%22must%22:%5B%7B%22term%22:%7B%22invoke_type.keyword%22:%22Post%20Merge%20Action%22%7D%7D,%7B%22prefix%22:%7B%22pull_request_title.keyword%22:%22push%20trigger%20main%22%7D%7D%5D%7D%7D%5D,%22minimum_should_match%22:1%7D'))),fullScreenMode:!f,options:(hidePanelTitles:!f,useMargins:!t),query:(language:kuery,query:''),timeRestore:!t,title:'OpenSearch%20Gradle%20Check%20Success%20Rate',viewMode:view)).

The new workflow is this PR extracts the gradle check for the repository-gcs and repository-azure plugins to Github rather then Jenkins. The idea is that these are 1) run in parallel to speed up the core gradle check on Jenkins and 2) eventually we can get to the point where these only run when the code or dependencies have changed instead of on every pr.

[...]

PR #20977, opened 2026-03-24
Run bwcTestSnapshots as separate GitHub Action

andrross:
Similar to #20946, this is just an experiment to see if the `bwcTestSnapshots` target

### Check List

### DIFF
--- a/.github/workflows/bwc-test-snapshots.yml
+++ b/.github/workflows/bwc-test-snapshots.yml
@@ -1,0 +1,66 @@
+name: Backward Compatibility Tests
+on:
+  push:
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  bwc-test-snapshots:
+    if: github.repository == 'opensearch-project/OpenSearch'
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        task: [ 'bwcTestSnapshots2x', 'bwcTestSnapshots3x' ]
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v6
+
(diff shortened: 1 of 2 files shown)


andrross:
> One high-level comment is which version do the 3x and 2x test against?

@cwperks This is determined by a bunch of complicated logic in [BwcVersions.java](https://github.com/opensearch-project/OpenSearch/blob/c0852f84f0647f78a45e71396d6370a39ba4922f/buildSrc/src/main/java/org/opensearch/gradle/BwcVersions.java#L260-L298). I think the reasoning here is that you just have to keep [Version.java](https://github.com/opensearch-project/OpenSearch/blob/cfebc675d42665f08aa2f987b2ee3eefbbea8870/libs/core/src/main/java/org/opensearch/Version.java#L58) up-to-date and the build will automatically pick the correct versions to test against. I'm honestly not sure if this complexity is worth it, given the alternative is to be a little more declarative about what to test even if you have to update it as things change.

Encoding "2x" and "3x" into the tasks names as I've done here means that there would be changes required at the next major version bump.

cwperks:
> This is determined by a bunch of complicated logic in [BwcVersions.java](https://github.com/opensearch-project/OpenSearch/blob/c0852f84f0647f78a45e71396d6370a39ba4922f/buildSrc/src/main/java/org/opensearch/gradle/BwcVersions.java#L260-L298).

Yikes.

yea, I think there's room for simplification (and readability improvements)

In plain text, I would describe the intent of automated bwc checks as follows:

- **Same major version** 
  - **Previous** - Last released minor (on same major)
  - **Next** - Code from feature branch
- **Previous major version**
  -  **Previous** - Last released from last minor of previous major
  - **Next** - Code from feature branch

I would hope the code could read similarly and maybe make "previous" customizable, heck, even "next" could be customizable to make it easy to say trigger a run from 3.2 to 3.4.

andrross:
Closing in favor of #21026

Issue #19440, opened 2025-09-26
[Discuss] Consider changing how classloading works for extensible and extending plugins


cwperks:
**tldr;**

Plugins should control their deps and not have to wonder about whether plugins they extend provide a dep at runtime

**Longer version:** 

ExtensiblePlugin is an interesting sort of extension point for OpenSearch. It allows us to define additional ways of extending core functionality, but writing another extension point outside of the core. Two such common use-cases are `lang-painless` and the `opensearch-job-scheduler`. With ExtensiblePlugins, plugins define relationships to one another.

For instance, the reporting plugin extends the opensearch-job-scheduler in order to get job scheduling functionality to run reports on a schedule.

There is one not so obvious thing about these plugins relations and that is how their classloaders relate.

When the reporting plugin extends the job scheduler, a side-effect is that job-scheduler's dependencies are available to the reporting plugin at _runtime_. However, the same dependencies are **not** available to reporting at _compile time_.

As a result, you find extending plugins setting `compileOnly` declarations on dependencies that they share with their extended plugin.

This causes some headache to plugin developers when extending other plugins because they are at the whim of the extensible plugin deciding to add dependencies that they also rely upon.

I'm opening this issue to discuss if we can change how classloading works for extensible and extending plugins.

In particular, I'd like to explore removing `extendedLoaders` from here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/PluginsService.java#L789

### To Reproduce

Prevent issues such as https://github.com/opensearch-project/ml-commons/issues/4202

### Expected behavior

No jar hell if extensible and extending plugin have the same `implementation` level dep.

### Additional Details

Extensible plugins have been being used much more than from the earlier days of OpenSearch.

dbwiddis:
> As a result, you find extending plugins setting `compileOnly` declarations on dependencies that they share with their extended plugin.
> 
> This causes some headache to plugin developers when extending other plugins because they are at the whim of the extensible plugin deciding to add dependencies that they also rely upon.

Example: https://github.com/opensearch-project/ml-commons/issues/3860

andrross:
> In particular, I'd like to explore removing extendedLoaders from here: [main/server/src/main/java/org/opensearch/plugins/PluginsService.java#L789](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/plugins/PluginsService.java#L789)

[...]

PR #20978, opened 2026-03-24
Allow plugins to copy libs to plugins/lib/ similar to bin/ and config/

cwperks:
### Description

This change allows plugins to copy libs under the `plugins/lib/` directory of the OpenSearch installation so that extending plugins can use these jars specifically without having access to all runtime dependencies of the extended plugin.

This PR solves an issue experienced in the extending plugin <-> extended plugin relationship where the extended plugin provides a library (an interface) that the extending plugin must implement.

In practice, if we have pluginA extends pluginB then the classloader of pluginA is the classloader of pluginB. What this means is that all of pluginB's deps are available to pluginA at runtime even though it only needs a handful of classes related to the SPI interface that the extending plugin must implement.

This change is influenced by how we handle both `config/` and `bin/` files. When plugins create their assembly (the plugin zip file), if it has a high-level config/ directory or bin/ directory then at installation time we automatically copy these to OPENSEARCH_HOME/config/<plugin_name> and OPENSEARCH_HOME/bin/<plugin_name> respectively. This PR does the same with `shared/` now. So if a plugin assembly has a top level `shared/` directory then we now will copy these to OPENSEARCH_HOME/plugins/lib/<plugin_name>.

When a plugin is uninstalled then we will remove these as well.

With new behavior:

```
> tree plugins
plugins
├── lib
│   └── opensearch-job-scheduler
│       └── opensearch-job-scheduler-spi-3.6.0.0-SNAPSHOT.jar
└── opensearch-job-scheduler
    ├── opensearch-job-scheduler-3.6.0.0-SNAPSHOT.jar
    └── plugin-descriptor.properties
```

Old behavior:

```
➜  opensearch-3.6.0-SNAPSHOT git:(main) ✗ pwd
/Users/cwperx/Projects/opensearch/OpenSearch/build/distribution/local/opensearch-3.6.0-SNAPSHOT
➜  opensearch-3.6.0-SNAPSHOT git:(main) ✗ tree plugins/opensearch-job-scheduler
plugins/opensearch-job-scheduler
├── opensearch-job-scheduler-3.6.0.0-SNAPSHOT.jar
├── opensearch-job-scheduler-spi-3.6.0.0-SNAPSHOT.jar
└── plugin-descriptor.properties
```

### Related Issues

Resolves https://github.com/opensearch-project/OpenSearch/issues/19440

### Check List

### DIFF
--- a/distribution/tools/plugin-cli/src/main/java/org/opensearch/tools/cli/plugin/PluginHelper.java
+++ b/distribution/tools/plugin-cli/src/main/java/org/opensearch/tools/cli/plugin/PluginHelper.java
@@ -31,10 +31,12 @@ public class PluginHelper {
      * @throws IOException if any I/O exception occurs while performing a file operation
      */
     public static Path verifyIfPluginExists(Path pluginPath, String pluginName) throws IOException {
-        List<Path> pluginSubFolders = Files.walk(pluginPath, 1).filter(Files::isDirectory).collect(Collectors.toList());
+        List<Path> pluginSubFolders = Files.walk(pluginPath, 1)
+            .filter(Files::isDirectory)
+            .filter(f -> !f.getFileName().toString().equals("lib"))
+            .collect(Collectors.toList());
         for (Path customPluginFolderPath : pluginSubFolders) {
             if (customPluginFolderPath != pluginPath
-                && "lib".equals(customPluginFolderPath.getFileName().toString()) == false
                 && !((customPluginFolderPath.getFileName().toString()).contains(".installing"))
                 && !((customPluginFolderPath.getFileName().toString()).contains(".removing"))) {
                 final PluginInfo info = PluginInfo.readFromProperties(customPluginFolderPath);


reta:
@cwperks I believe this change will significantly weaken the classpath isolation and open up a whole new class of issues.  Presence of additional, possibly conflicting / incompatible JAR files would easily render the server process nonoperational, this is just my opinion.

cwperks:
> @cwperks I believe this change will significantly weaken the classpath isolation and open up a whole new class of issues. Presence of additional, possibly conflicting / incompatible JAR files would easily render the server process nonoperational, this is just my opinion.

Thank you for chiming in @reta as this is exactly the sort of discussion I wanted to provoke. I would actually argue that this is also status quo today.

i.e. when pluginA extends pluginB then pluginB's dependencies are available to pluginA at runtime.

Let's say both plugins depend on guava, then pluginA needs to declare `compileOnly` on this dep while pluginB uses `implementation` or `api`.

If pluginA compiles w/ version X (let's say 32) and pluginB has a runtime dep on version Y (let's say 33), then pluginA is building with the wrong version and uses a different version of the class at runtime because the class from the parent classloader (pluginB's classloader) wins at runtime. Solution for this of course is a version catalog.

I think it would be better to get these libs that multiple plugins need to reference in a single place without leaking other parent plugin runtime dependencies.

As for sageguards, the jarHell check is still in place and also I believe we can prompt the cluster admin on install to specify a lib will be installed as well. I treat plugin installation as an intentional act by the cluster admin, but hear your concern as well.

reta:
> As for sageguards, the jarHell check is still in place and also I believe we can prompt the cluster admin on install to specify a lib will be installed as well. I treat plugin installation as an intentional act by the cluster admin, but hear your concern as well.

This is yet another side of the problem - the plugins cannot reliably cleanup the `lib` folder, since if  plugin A and plugin B copy the same JAR (let say job-scheduler),  then uninstalling either plugin A or plugin B should clean up the  `lib` folder, breaking things. Certainly not something we should be doing (again, my opinion).

cwperks:
> This is yet another side of the problem - the plugins cannot reliably cleanup the `lib` folder, since if plugin A and plugin B copy the same JAR (let say job-scheduler), then uninstalling either plugin A or plugin B should clean up the `lib` folder, breaking things. Certainly not something we should be doing (again, my opinion).

job-scheduler provides the job-scheduler-spi jar so its incumbent on job-scheduler to add and remove it on install/uninstall respectively.

The plugins that rely on job-scheduler simply need to check for the class' existence at runtime.

2 cases:

1. **pluginA extends pluginB** -> pluginB is required to be installed so pluginA is rest assured that the lib/ is there on the runtime classpath
2. **pluginA optionally extends pluginB** -> pluginB is not required to be installed so pluginA would need to check for class existence at runtime if for some reason then need the class at runtime.

reta:
> The plugins that rely on job-scheduler simply need to check for the class' existence at runtime.

We are considering a general case here (I thought job-scheduler is a good simple example): if several plugins copy same JARs to `lib', uninstalling any of them will break others. Asking plugin developers to do additional runtime checks is a regression in developer experience I believe.

cwperks:
> We are considering a general case here (I thought job-scheduler is a good simple example): if several plugins copy same JARs to `lib', uninstalling any of them will break others. Asking plugin developers to do additional runtime checks is a regression in developer experience I believe.

plugins devs would need to be judicious on what goes into these libs to avoid any conflicts. At least in the SPI case, these really are only meant to be the interfaces (and associated classes) that extending plugins need to extend. 

> The plugins that rely on job-scheduler simply need to check for the class' existence at runtime.

To be clear, this is only the case when optionally extending another plugin. Would not be the case for job-scheduler.

cwperks:
Imagine the case where you have, pluginA extends pluginB, pluginB extends pluginC and pluginC extends pluginD (I believe we have an example of this in neural-search). That means pluginsA's has an ancestry of classloaders through pluginD so all runtime deps of pluginB, pluginC and pluginD are available to pluginA.

I believe these should be better isolated and this change would work towards that where the shared libs go into lib and loaded by the root classloader.

Those libraries are intentionally minimal and only supposed to provide the classes that the extending plugins need to implement.

idk if this is the right approach necessarily (I do like the simplicity of it), but I know there are problems with the current way this is done so I'm trying to simplify it.

reta:
>Those libraries are intentionally minimal and only supposed to provide the classes that the extending plugins need to implement.

I feel strongly that risk of breaking things in unpredictable ways overweights the potential convenience of sharing. Plus, I still don't understand how the uninstall mechanism would work (keeping such libs "just in case" is not an option). But that's again, my personal take.

cwperks:
This PR handles uninstall case. When job scheduler is removed it removes the SPI lib.

Job scheduler can only be removed when all plugins that extend it are also removed.

distribution/tools/plugin-cli/src/main/java/org/opensearch/tools/cli/plugin/RemovePluginCommand.java
```diff
@@ -170,6 +171,18 @@ void execute(Terminal terminal, Environment env, String pluginName, boolean purg
...
+        final Path pluginLibDir = env.libDir().resolve(pluginName);
+        if (Files.exists(pluginLibDir)) {
+            if (!Files.isDirectory(pluginLibDir)) {
+                throw new UserException(ExitCodes.IO_ERROR, "lib dir for " + pluginName + " is not a directory");
+            }
+            try (Stream<Path> paths = Files.list(pluginLibDir)) {
+                pluginPaths.addAll(paths.collect(Collectors.toList()));
+            }
+            pluginPaths.add(pluginLibDir);
+            terminal.println(VERBOSE, "removing [" + pluginLibDir + "]");
```
cwperks: @reta FYI this is where the removal happens.

So with `bin/` and `config/` we already have a convention.

On install any plugin that has top-level `bin/` and `config/` already do get these files automatically copied to OPENSEARCH_HOME/bin/<plugin-name> and OPENSEARCH_HOME/config/<plugin-name> so the removal is also based on this convention.

not necessarily trying to persuade, just want to provide the context that this PR does in fact handle the case for removal. At least providing the context to figure out what a "right" solution could look like.

Also if plugins that extend the job-scheduler are installed (say the reporting plugin), then job-scheduler cannot be removed **until** reporting plugin is removed first.

reta: > @reta FYI this is where the removal happens.

Got it @cwperks , I think I misunderstood **who** copies the plugin in first place, the uninstall is cleared out, thank you. The risks still present since plugins can copy anything there (not only SPI). Another thing to keep in mind (this is coming from production deployment), the `lib` folder could be mounted as readonly file system, this will not work.

cwperks: The other solution I was considering was "micro plugins" where a plugin can create a micro plugin for the spi and extend that (w/o providing an implementation) so that its only the necessary classes in the SPI "micro plugin" folder. I tried that out, but there were complications on the install/uninstall path and this path ultimately was much simpler since it copied existing conventions.

reta: What if we keep OS untouched but introduce `plugins/libs` or `plugins/shared/libs`? In this case the OS stay clean, the plugins folder is expected to change with plugin installation.

cwperks: ^ that's clean. I'll look into what that would take and to update classloading logic accordingly.

cwperks: New format:

```
➜  opensearch-3.6.0-SNAPSHOT git:(micro-plugin) ✗ tree plugins
plugins
├── lib
│   └── opensearch-job-scheduler
│       └── opensearch-job-scheduler-spi-3.6.0.0-SNAPSHOT.jar
└── opensearch-job-scheduler
    ├── opensearch-job-scheduler-3.6.0.0-SNAPSHOT.jar
    └── plugin-descriptor.properties
```

cwperks: FYI the classLoading logic is updated as follows:

- If a plugin starts bundling a `lib/` folder w/ a SPI jar.
- On install the `lib/` from top-level directory of plugin zip is copied to `OPENSEARCH_HOME/plugins/lib/<plugin_name>`
- On node-boostrap we create a lib-only loader if a lib exists with the extended plugin name.
- If pluginA extends pluginB and pluginB provides a lib then parentA's classloader will be the lib-only loader
- It pluginB doesn't provide lib (uses existing pattern where SPI is bundled with other deps), then we fallback to pluginA's classloader is pluginB's classloader so all pluginB's runtime dependencies are visible to pluginA
  - This is to ensure current behavior works as intended and any existing ExtensiblePlugin can opt into this new model

distribution/tools/plugin-cli/src/main/java/org/opensearch/tools/cli/plugin/InstallPluginCommand.java
```diff
@@ -884,6 +891,12 @@ private void installPluginSupportFiles(PluginInfo info, Path tmpRoot, Path destB
             // so any installed config files are left on failure too
             installConfig(info, tmpConfigDir, destConfigDir);
         }
+
+        Path tmpLibDir = tmpRoot.resolve("lib");
+        if (Files.exists(tmpLibDir)) {
+            deleteOnFailure.add(destLibDir);
+            installLib(info, tmpLibDir, destLibDir);
```
reta, line 898: I am trying to figure out which plugin JARs are going to be copied to shared libs? We definitely want only SPIs or/and client ones (may be with dependencies), but not all JARs. Could you please help me understand? (or just have a comment here to explain which libs we copy over). Thank you.

cwperks: Ideally, it would only be SPI jars and on the plugin side there would need to be some refactoring after this PR if they want to take advantage of the changes in this PR. Note: plugins do not need to immediately change as this retains the current classloading as well if a plugin does not have a `lib/` folder in their assembly.

Example extensible plugins:

- Job Scheduler
- transport-grpc
- ml-commons
- notifications
- security (optionally extensible)
- knn
- lang-painless
- workload-management
- alerting

See: https://github.com/search?q=org%3Aopensearch-project+extendedPlugins&type=code&p=2

I always use job-scheduler as an example bc I think its the quintessential model of ExtensiblePlugin.

This is the SPI module: https://github.com/opensearch-project/job-scheduler/tree/main/spi

The important class is [JobSchedulerExtension](https://github.com/opensearch-project/job-scheduler/blob/main/spi/src/main/java/org/opensearch/jobscheduler/spi/JobSchedulerExtension.java)

The job scheduler, and all plugins that extend the job scheduler need to reference the same exact class at runtime (from the same classloader).

Currently, job-scheduler has the SPI as an `implementation` dependency and plugins that extend the job-scheduler have is at a `compileOnly` dependency.

The way they reference the same class is that if a plugin extends the job-scheduler, the job-scheduler's classloader is the parent classloader of the extending plugin. In the delegation model for java, if a class is present on a higher classloader then its the one that's chosen at runtime so you can be assured that both the extending and extended plugin use the same class at runtime.

But this necessarily means that all of job-scheduler's runtime dependencies are chosen first when really its only done for the SPI.

job-scheduler happens to be a "slim" plugin, but its not the case with all ExtensiblePlugins and I think it would be better to have a clean classloading model.

reta: Oh I see, basically plugins have to place these shared `JARs` inside own `lib` folder, may be let us assume it is called `shared` (on the plugin side) to clearly highlight the purpose?

cwperks: Yea that sounds reasonable, but keep the same path for installation. i.e. `OPENSEARCH_HOME/plugins/lib/<plugin_name>`?

A few min and I will push and test the change.

cwperks: Made the suggested change

distribution/tools/plugin-cli/src/main/java/org/opensearch/tools/cli/plugin/PluginHelper.java
```diff
@@ -34,6 +34,7 @@ public static Path verifyIfPluginExists(Path pluginPath, String pluginName) thro
         List<Path> pluginSubFolders = Files.walk(pluginPath, 1).filter(Files::isDirectory).collect(Collectors.toList());
         for (Path customPluginFolderPath : pluginSubFolders) {
             if (customPluginFolderPath != pluginPath
+                && "lib".equals(customPluginFolderPath.getFileName().toString()) == false
```
reta: This exclusion seems to be unnecessary, there should be no lib folder for plugins, right? and everything under shared is moved as well?

reta: Oh, my bad, we iterate all folders, added a comment above

distribution/tools/plugin-cli/src/main/java/org/opensearch/tools/cli/plugin/PluginHelper.java
```diff
@@ -34,6 +34,7 @@ public static Path verifyIfPluginExists(Path pluginPath, String pluginName) thro
         List<Path> pluginSubFolders = Files.walk(pluginPath, 1).filter(Files::isDirectory).collect(Collectors.toList());
```
reta:
```suggestion
        List<Path> pluginSubFolders = Files.walk(pluginPath, 1).filter(Files::isDirectory).filter(f -> !f.getFileName().equals("lib" ).collect(Collectors.toList());
```

cwperks: Done, added this and removed the condition below

reta:
Thanks @cwperks , I think it looks good, we probably need CHANGELOG entry. I would suggest to land it in 3.7 so we would have time to try it with some plugins during the development, sounds reasonable? Thank you.

cwperks:
> Thanks @cwperks , I think it looks good, we probably need CHANGELOG entry. I would suggest to land it in 3.7 so we would have time to try it with some plugins during the development, sounds reasonable? Thank you.

yes, sounds good. I will add a CHANGELOG entry as soon as its cleared on the main branch for 3.7. TY @reta!

cwperks:
Now that the CHANGELOG is not required. This one is ready.

cwperks:
@reta can this one be completed now that CHANGELOG is no longer required?

PR #20979, opened 2026-03-24
Cluster state and recovery constructs for in-place shard split

vikasvb90:
### Description
Add cluster state infrastructure for in-place shard split

This PR adds the cluster state update service and supporting POJO changes needed to trigger an in-place shard split. 

Changes:

Cluster state update service:
- MetadataInPlaceShardSplitService: Submits an acked cluster state update task that updates SplitShardsMetadata on the target index and triggers a reroute. Validates that the index exists, the shard is not already splitting or already split, and virtual shards are not enabled on the index.
- InPlaceShardSplitClusterStateUpdateRequest: Request POJO holding index name, shard ID, split-into count, and cause.
- ClusterManagerTask: Added IN_PLACE_SHARD_SPLIT task key for cluster manager task throttling.

Routing POJO changes to support shard split lifecycle:
- ShardRoutingState: Added SPLITTING state for parent shards undergoing an in-place split.
- RecoverySource: Added InPlaceShardSplitRecoverySource type for child shards recovering from a parent shard on the same node.
- UnassignedInfo: Added CHILD_SHARD_CREATED reason for child shards pending allocation after a split.
- AllocationId: Added splitChildAllocationIds and parentAllocationId fields with version-gated serialization (V_3_6_0), factory methods (newSplit, newTargetSplit, cancelSplit, finishSplit), and updated equals/hashCode/toXContent.
- ShardRouting: Added recoveringChildShards and parentShardId fields, new constructor accepting split fields, and query methods (splitting(), isSplitTarget(), getRecoveringChildShards(), getParentShardId()).

The REST API is not exposed in this PR. The routing allocation logic to actually assign child shards to nodes will follow in a subsequent PR.

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceSplitShardService.java
+++ b/server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceSplitShardService.java
@@ -1,0 +1,159 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.cluster.metadata;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ParameterizedMessage;
+import org.opensearch.Version;
+import org.opensearch.action.admin.indices.split.InPlaceSplitShardClusterStateUpdateRequest;
+import org.opensearch.cluster.AckedClusterStateUpdateTask;
+import org.opensearch.cluster.ClusterState;
+import org.opensearch.cluster.ack.ClusterStateUpdateResponse;
+import org.opensearch.cluster.routing.RoutingTable;
+import org.opensearch.cluster.routing.ShardRouting;
+import org.opensearch.cluster.routing.allocation.AllocationService;
+import org.opensearch.cluster.service.ClusterManagerTask;
+import org.opensearch.cluster.service.ClusterManagerTaskThrottler;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.Priority;
+import org.opensearch.common.annotation.ExperimentalApi;
+import org.opensearch.core.action.ActionListener;
+
+import java.util.function.BiFunction;
+
+/**
+ * Service responsible for applying in-place shard split requests to cluster state.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class MetadataInPlaceSplitShardService {
+    private static final Logger logger = LogManager.getLogger(MetadataInPlaceSplitShardService.class);
+
+    private final ClusterService clusterService;
+    private final AllocationService allocationService;
+    private final ClusterManagerTaskThrottler.ThrottlingKey splitShardTaskKey;
+
+    public MetadataInPlaceSplitShardService(final ClusterService clusterService, final AllocationService allocationService) {
+        this.clusterService = clusterService;
+        this.allocationService = allocationService;
+        this.splitShardTaskKey = clusterService.registerClusterManagerTask(ClusterManagerTask.IN_PLACE_SPLIT_SHARD, true);
+    }
+
+    /**
+     * Submits a cluster state update task to split a shard in-place.
+     */
+    public void split(final InPlaceSplitShardClusterStateUpdateRequest request, final ActionListener<ClusterStateUpdateResponse> listener) {
+        clusterService.submitStateUpdateTask(
+            "in-place-split-shard [" + request.getShardId() + "] of index [" + request.getIndex() + "], cause [" + request.cause() + "]",
+            new AckedClusterStateUpdateTask<>(Priority.URGENT, request, listener) {
+                @Override
+                protected ClusterStateUpdateResponse newResponse(boolean acknowledged) {
+                    return new ClusterStateUpdateResponse(acknowledged);
+                }
+
+                @Override
+                public ClusterManagerTaskThrottler.ThrottlingKey getClusterManagerThrottlingKey() {
+                    return splitShardTaskKey;
+                }
+
+                @Override
+                public ClusterState execute(ClusterState currentState) {
+                    return applySplitShardRequest(currentState, request, allocationService::reroute);
+                }
+
+                @Override
+                public void onFailure(String source, Exception e) {
+                    logger.trace(
+                        () -> new ParameterizedMessage(
+                            "[{}] of index [{}] failed to split online",
+                            request.getShardId(),
+                            request.getIndex()
+                        ),
+                        e
+                    );
+                    super.onFailure(source, e);
+                }
+            }
+        );
+    }
+
+    /**
+     * Applies a shard split request to the given cluster state. Updates the split metadata
+     * in the index metadata and triggers a reroute so that child shards get allocated.
+     */
+    static ClusterState applySplitShardRequest(
+        ClusterState currentState,
+        InPlaceSplitShardClusterStateUpdateRequest request,
+        BiFunction<ClusterState, String, ClusterState> rerouteRoutingTable
+    ) {
+        IndexMetadata curIndexMetadata = currentState.metadata().index(request.getIndex());
+        if (curIndexMetadata == null) {
+            throw new IllegalArgumentException("Index [" + request.getIndex() + "] not found");
+        }
+
+        if (curIndexMetadata.getNumberOfVirtualShards() != -1) {
+            throw new IllegalArgumentException(
+                "In-place shard split is not supported on index [" + request.getIndex() + "] with virtual shards enabled"
+            );
+        }
+
+        if (currentState.nodes().getMinNodeVersion().equals(currentState.nodes().getMaxNodeVersion()) == false
+            || currentState.nodes().getMinNodeVersion().before(Version.V_3_7_0)) {
+            throw new IllegalArgumentException(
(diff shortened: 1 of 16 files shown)


server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceShardSplitService.java
```diff
@@ -0,0 +1,131 @@
...
+                }
+            }
+        );
+    }
+
+    /**
+     * Applies a shard split request to the given cluster state. Updates the split metadata
+     * in the index metadata and triggers a reroute so that child shards get allocated.
+     */
+    static ClusterState applyShardSplitRequest(
```
shwetathareja: we should have separate check to not allow split if cluster is in mixed mode during upgrades to have a guardrail

server/src/main/java/org/opensearch/cluster/routing/AllocationId.java
```diff
@@ -137,29 +175,60 @@ public static AllocationId newTargetRelocation(AllocationId allocationId) {
...
+        for (int c = 0; c < numberOfChildShards; c++) {
+            childIds.add(UUIDs.randomBase64UUID());
+        }
+        return new AllocationId(allocationId.getId(), null, childIds, null);
+    }
+
+    /**
+     * Creates a new allocation id for a child shard that is the result of a split.
+     */
+    public static AllocationId newTargetSplit(AllocationId allocationId, String childAllocId) {
```
shwetathareja, line 220: the above method generates a transient value and when do we call this method? Not clear on the flow here.

server/src/main/java/org/opensearch/cluster/routing/ShardRouting.java
```diff
@@ -268,6 +309,34 @@ public boolean relocating() {
...
+     * Returns <code>true</code> if this shard is a child shard created by an in-place split.
+     */
+    public boolean isSplitTarget() {
+        return parentShardId != null;
+    }
+
+    /**
+     * Returns the recovering child shards of this splitting shard, or null if not splitting.
+     */
+    public ShardRouting[] getRecoveringChildShards() {
```
shwetathareja, line 331: >getRecoveringChildShards() returns the raw ShardRouting[] array directly. Callers can mutate the array contents, breaking the immutability contract of ShardRouting. It should return a defensive copy or an unmodifiable view (e.g., Arrays.copyOf or expose as List<ShardRouting>).

can you look into this

server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceShardSplitService.java
vikasvb90: Do you mean for the version lower than the split supported version? In other version upgrades we can support this. And for a version upgrade from a non-supported split version to a supported split version, it is implicitly handled because the [REST](https://github.com/vikasvb90/OpenSearch/blob/split_final/server/src/main/java/org/opensearch/rest/action/admin/indices/RestInPlaceShardSplitHandler.java) action as well as the admin client (invoked from a plugin) won't be available on lower version cluster manager. 

If you want I can still put an additional check here.

vikasvb90: Actually I am thinking that we should prevent splitting a shard in all cases where it is now supposed to move off the node except shard rebalancing case. 
And if split already started then we should cancel it and let the shard relocate in cases above.
What say?

server/src/main/java/org/opensearch/cluster/routing/AllocationId.java
vikasvb90: Ya there are actually missing pieces to avoid making this PR huge. But your concern is very valid. So linking the source code from my personal repo [here](https://github.com/vikasvb90/OpenSearch/blob/split_final/server/src/main/java/org/opensearch/cluster/routing/AllocationId.java#L177). For this comment, 
1. `newSplit` is called on parent shard to generate N random UUIDs and store them in splitChildAllocationIds on the parent's allocation object. 
2. `newTargetSplit` is called on each child shard. Takes one of the generated UUIDs, and creates a link to its parent by storing `parentAllocationId`. 

This is same as relocation flow where in `newRelocation` source shard generates random UUID for target shard where target shard's id becomes this generated UUID. And this is how both are linked together. 

For split, these generations happen in `createRecoveringChildShards` of `RoutingNodes`

server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceShardSplitService.java
shwetathareja: yes makes sense to me.

vikasvb90: I have added multiple checks now to meet shard split criteria now in this class. Cancellation of split will happen later as part of a new cluster state update so that will come in next PR. I will also need to add cancellation reason in `_cat/recovery` API to inform user why split was cancelled.

server/src/main/java/org/opensearch/cluster/routing/ShardRouting.java
vikasvb90: addressed this. Now returning a clone of this. Not creating a deep copy is intentional because same as relocation flow, `RoutingNodes` depend on original objects. But in this case I don't need to return the original array at least to prevent array mutation.

server/src/main/java/org/opensearch/action/admin/indices/split/InPlaceShardSplitClusterStateUpdateRequest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.opensearch.cluster.ack.ClusterStateUpdateRequest;
+import org.opensearch.common.annotation.ExperimentalApi;
+
+/**
+ * Cluster state update request for in-place shard split.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class InPlaceShardSplitClusterStateUpdateRequest extends ClusterStateUpdateRequest<InPlaceShardSplitClusterStateUpdateRequest> {
```
shwetathareja: should we switch it to SplitShard instead of ShardSplit everywhere to keep verb first aligning with general naming convention

vikasvb90: done

server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceSplitShardService.java
```diff
@@ -0,0 +1,159 @@
...
+                "Cannot split shard ["
+                    + shardId
+                    + "] on index ["
+                    + request.getIndex()
+                    + "] because the primary shard is not started, current state: "
+                    + primaryShard.state()
+            );
+        }
+
+        RoutingTable.Builder routingTableBuilder = RoutingTable.builder(currentState.routingTable());
```
kkewwei, line 145: Can we add a log here?

server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceSplitShardService.java
```diff
@@ -0,0 +1,159 @@
...
+
+        ShardRouting primaryShard = currentState.routingTable()
+            .shardRoutingTable(curIndexMetadata.getIndex().getName(), shardId)
+            .primaryShard();
+        if (primaryShard.relocating()) {
+            throw new IllegalArgumentException(
+                "Cannot split shard [" + shardId + "] on index [" + request.getIndex() + "] because it is currently relocating"
+            );
+        }
+        if (primaryShard.started() == false) {
```
kkewwei, line 134: `primaryShard.started() == false` contains `primaryShard.relocating()`

server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceSplitShardService.java
```diff
@@ -0,0 +1,159 @@
...
+            || currentState.nodes().getMinNodeVersion().before(Version.V_3_7_0)) {
+            throw new IllegalArgumentException(
+                "In-place shard split requires all nodes to be on the same version, at or above " + Version.V_3_7_0
+            );
+        }
+
+        int shardId = request.getShardId();
+        SplitShardsMetadata splitShardsMetadata = curIndexMetadata.getSplitShardsMetadata();
+
+        if (splitShardsMetadata.getInProgressSplitShardIds().contains(shardId)) {
```
kkewwei, line 118: Can we remove `isSplitParent` `getInProgressSplitShardIds` check? I see the same checks exist in `splitShard`.

server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceSplitShardService.java
vikasvb90: No it doesn't.

server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceSplitShardService.java
vikasvb90: Not sure what you mean here. 
1. Both are different checks
2. There are no such checks present in request path to inform user early. All checks are later when cluster state gets created which is too late already to update the user.

One suggestion is that when you are referring to a code pointer, please add the link going forward.

server/src/main/java/org/opensearch/cluster/metadata/MetadataInPlaceSplitShardService.java
vikasvb90: Will add a debug log in the follow up PR.

PR #20984, opened 2026-03-24
Add cluster.blocks.read.auto_release setting to control read block auto-release

### LINKED ISSUE #20539: [BUG] `DiskThresholdMonitor` unconditionally auto-releases user-set `index.blocks.read` setting

AndrewCharlesHay:
### Description

Introduces a new dynamic cluster setting `cluster.blocks.read.auto_release` (default: `true`) that controls whether `DiskThresholdMonitor` automatically releases index read blocks when disk usage returns to a healthy level.

Previously, the monitor unconditionally released read blocks whenever disk usage dropped below the high watermark threshold. This caused an issue for warm/cold node use cases where operators intentionally set `index.blocks.read` — the monitor would silently clear those blocks, causing unexpected write/read behavior.

With this change, operators can disable auto-release:

```
PUT _cluster/settings
{
  "persistent": {
    "cluster.blocks.read.auto_release": false
  }
}
```

When disabled, the monitor will not automatically release read blocks, leaving them in place until manually removed by the operator.

### Related Issues

Resolves #20539

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/cluster/routing/allocation/DiskThresholdMonitor.java
+++ b/server/src/main/java/org/opensearch/cluster/routing/allocation/DiskThresholdMonitor.java
@@ -494,19 +494,24 @@ public class DiskThresholdMonitor {
     }
 
     private void handleReadBlocks(ClusterState state, Set<String> indicesToBlockRead, ActionListener<Void> listener) {
-        final Set<String> indicesToReleaseReadBlock = StreamSupport.stream(
-            Spliterators.spliterator(state.routingTable().indicesRouting().entrySet(), 0),
-            false
-        )
-            .map(Map.Entry::getKey)
-            .filter(index -> indicesToBlockRead.contains(index) == false)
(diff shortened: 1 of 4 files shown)


AndrewCharlesHay:
Good catch from the bot — the original code already had the double-listener pattern (both the release phase and the apply phase passed the outer `listener` directly), but my change made it more visible.

Fixed in the latest commit by restructuring `handleReadBlocks` to define the apply phase as a `Runnable` and chain it after the release completes via `ActionListener.wrap`. The outer `listener` is now only ever invoked once — by the apply phase — regardless of whether auto-release is enabled or not.

AndrewCharlesHay:
Gradle check 73204 shows **"Test Result (no failures)"** — the tests all passed. The GitHub Actions wrapper job reported failure but Jenkins itself had no test failures. Requesting a re-run of the gradle-check.

/rerun-gradle-check

Issue #20899, opened 2026-03-17
[Proposal] Simplify `precommit` workflows


andrross:
### Is your feature request related to a problem? Please describe

We currently run the `precommit` task across the full matrix of OS and JDK versions. These tasks are pretty reliable, but sometimes do fail for reasons unrelated to the code in the PR, e.g.

```
FAILURE: Build failed with an exception.
[Incubating] Problems report is available at: file:///Users/runner/work/OpenSearch/OpenSearch/build/reports/problems/problems-report.html

263 actionable tasks: 262 executed, 1 from cache
* What went wrong:
Execution failed for task ':plugins:analysis-nori:missingJavadoc'.
> Could not resolve all files for configuration ':plugins:analysis-nori:compileClasspath'.
   > Could not download lucene-analysis-nori-10.4.0.jar (org.apache.lucene:lucene-analysis-nori:10.4.0)
      > Could not get resource 'https://repo.maven.apache.org/maven2/org/apache/lucene/lucene-analysis-nori/10.4.0/lucene-analysis-nori-10.4.0.jar'.
         > (bad_record_mac) Tag mismatch
```

However, the tasks in precommit are not platform dependent:

1. forbiddenApis — checks bytecode for use of forbidden JDK APIs (via de.thetaphi:forbiddenapis)
2. jarHell — detects duplicate classes on the classpath
3. forbiddenPatterns — scans source files for forbidden text patterns (e.g. tabs, nocommit markers)
4. licenseHeaders — validates license headers on source files
5. filePermissions — checks file permissions are correct
6. thirdPartyAudit — audits third-party dependencies for class-level violations
7. testingConventions — validates test class naming conventions (must end in "Tests" or "IT")
8. dependencyLicenses — verifies LICENSE/NOTICE/SHA files exist for each dependency (only when includeDependencyLicenses=true)
9. loggerUsage — checks for correct logger usage patterns (skipped for :build-tools)

The platform-dependent task within precommit as far as I can see is `compileJava`. That is already covered by the `assemble` workflow that is run across the full matrix as well.

### Describe the solution you'd like

Stop running the `precommit` workflow across all platforms. Rely on the `assemble` workflow for cross-platform compatibility.

### Related component

Build

andrross:
@cwperks @reta would love your opinion here if I'm missing something

cwperks:
I agree that we don't need to run it on the full matrix. Also curious to know if you see any opportunities to speed up the precommit?

[...]

PR #20992, opened 2026-03-24
Conditionally run assemble GitHub action

andrross:
Only trigger the assemble GitHub action if files that could plausibly impact building the distributions have changed.

Part of #20899

### Check List

### DIFF
--- a/.github/workflows/assemble.yml
+++ b/.github/workflows/assemble.yml
@@ -1,5 +1,14 @@
 name: Gradle Assemble
-on: [pull_request]
+on:
+  pull_request:
+    # Skip this check unless files that could plausibly impact a
+    # distribution have changed.
+    paths:
+      - 'distribution/**'
+      - 'buildSrc/**'
+      - '**/*.gradle'
+      - 'gradle/**'
+      - '.github/workflows/assemble.yml'
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/assemble.yml
```diff
@@ -1,5 +1,15 @@
 name: Gradle Assemble
-on: [pull_request]
+on:
+  pull_request:
+    # Skip this check unless files that could plausibly impact a
+    # distribution have changed.
+    paths:
+      - 'distribution/**'
+      - 'buildSrc/**'
+      - 'build.gradle'
```
cwperks: Does this cover all `build.gradle` files?

andrross: I don't think so. I think `**build.gradle` would though?

cwperks: I think we can keep `build.gradle` in here (to match top-level) and also add `**/build.gradle`

reta: Yeah, I was under impression it covers all build.gradle files, we really need that

andrross: I changed this to use `**/*.gradle` which will match any gradle file anywhere in the repo, including root.

PR #20993, opened 2026-03-24
Move precommit to nightly schedule

andrross:
Stop running 11 variants of precommit on every pull request. This check is already run as a part of the overall 'check' workflow. Running across the matrix of OS and JDK versions is intended to ensure the development remains supported on those platforms. It is rather heavyweight to run on every PR, so moving it to a nightly job seems like a reasonable tradeoff. We'll still detect regressions via the issue that is cut on failure.

Part of https://github.com/opensearch-project/OpenSearch/issues/20899

### Check List


.github/workflows/nightly-precommit.yml
```diff
@@ -1,5 +1,7 @@
 name: Gradle Precommit
-on: [pull_request]
+on:
+  schedule:
+    - cron: '0 6 * * *' # Daily at 6:00 AM UTC
```
reta, line 4: I thought that we would only keep one precommit check? (bundled JDK / Ubuntu) on each pull request? Wouldn't that require a split?

andrross: All the precommit tasks run as a part of `check` so I figured we are covered there. It will even bail out quickly if say spotless fails.

andrross: We used to gate the check workflow on precommit passing. Since that changed it's honestly been duplicate work to run precommit on bundled JDK / Ubuntu.

reta: Oh I see, sure, thanks @andrross

cwperks:
Should we consider having a single precommit (ubuntu, 21) and then passing `-x precommit` in the run to Jenkins?

andrross:
> Should we consider having a single precommit (ubuntu, 21) and then passing `-x precommit` in the run to Jenkins?

@cwperks I don't see a huge benefit. We'd be duplicating the work to compile the code across Jenkins and the GHA. The Jenkins job will fail quickly if there is a precommit error. What do you think?

cwperks:
> @cwperks I don't see a huge benefit. We'd be duplicating the work to compile the code across Jenkins and the GHA. The Jenkins job will fail quickly if there is a precommit error. What do you think?

yea that makes sense.

PR #20994, opened 2026-03-25
Fix SegmentReplicationWithNodeToNodeIndexShardTests

andrross:
Two fixes here:
- Wire in the cluster service to prevent an NPE
- Reduce random document count by 100x

The tests were previously indexing 30000-40000 documents. This would get the test into an fsync loop on the test machine and we were seeing frequent [test timeouts in CI][1].

[1]: https://build.ci.opensearch.org/job/gradle-check/73191/testReport/junit/org.opensearch.index.shard/SegmentReplicationWithNodeToNodeIndexShardTests/testPrimaryPromotionWithConcurrentTranslogRecovery/

### Check List

### DIFF
--- a/server/src/test/java/org/opensearch/index/shard/SegmentReplicationWithNodeToNodeIndexShardTests.java
+++ b/server/src/test/java/org/opensearch/index/shard/SegmentReplicationWithNodeToNodeIndexShardTests.java
@@ -541,7 +541,7 @@ public class SegmentReplicationWithNodeToNodeIndexShardTests extends SegmentRepl
         final RecoverySettings recoverySettings = new RecoverySettings(
             Settings.builder()
                 .put(RecoverySettings.INDICES_TRANSLOG_CONCURRENT_RECOVERY_ENABLE.getKey(), true)
-                .put(RecoverySettings.INDICES_TRANSLOG_CONCURRENT_RECOVERY_BATCH_SIZE.getKey(), 11000)
+                .put(RecoverySettings.INDICES_TRANSLOG_CONCURRENT_RECOVERY_BATCH_SIZE.getKey(), 1000)
                 .build(),
             new ClusterSettings(Settings.EMPTY, ClusterSettings.BUILT_IN_CLUSTER_SETTINGS)
         );
@@ -555,7 +555,7 @@ public class SegmentReplicationWithNodeToNodeIndexShardTests extends SegmentRepl
                 MergedSegmentPublisher.EMPTY
             )
         ) {
-            doPrimaryPromotion(shards, randomInt(10), randomIntBetween(30000, 40000));
+            doPrimaryPromotion(shards, randomInt(10), randomIntBetween(1100, 2000));
         }
     }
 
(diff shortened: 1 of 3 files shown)


server/src/test/java/org/opensearch/index/shard/SegmentReplicationWithNodeToNodeIndexShardTests.java
```diff
@@ -555,7 +555,7 @@ public void testPrimaryPromotionWithConcurrentTranslogRecovery() throws Exceptio
                 MergedSegmentPublisher.EMPTY
             )
         ) {
-            doPrimaryPromotion(shards, randomInt(10), randomIntBetween(30000, 40000));
+            doPrimaryPromotion(shards, randomInt(10), randomIntBetween(300, 400));
```
cwperks: I'm not an expert in this area, but is the relationship to [BATCH_SIZE](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/test/java/org/opensearch/index/shard/SegmentReplicationWithNodeToNodeIndexShardTests.java#L544) important?

andrross: Possibly? I'm not an expert here either, but it can't be set smaller that 10000

https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/indices/recovery/RecoverySettings.java#L247

I do think we should either ship this or disable the test ASAP because best case this test takes several minutes to run which is not ideal for a unit test, and worst case it gets killed by the 20 minute timeout and fails.

@guojialiang92 can you weigh in here?

cwperks: I think we should certainly avoid the magic numbers here.

cwperks: I see we create RecoverSettings above:

```
final RecoverySettings recoverySettings = new RecoverySettings(
    Settings.builder()
        .put(RecoverySettings.INDICES_TRANSLOG_CONCURRENT_RECOVERY_ENABLE.getKey(), true)
        .put(RecoverySettings.INDICES_TRANSLOG_CONCURRENT_RECOVERY_BATCH_SIZE.getKey(), 11000)
        .build(),
    new ClusterSettings(Settings.EMPTY, ClusterSettings.BUILT_IN_CLUSTER_SETTINGS)
);
```

particular for `new ClusterSettings(Settings.EMPTY, ClusterSettings.BUILT_IN_CLUSTER_SETTINGS)`, the second arg is BUILT_IN_CLUSTER_SETTINGS. Can we potentially create a test version of the `INDICES_TRANSLOG_CONCURRENT_RECOVERY_BATCH_SIZE` setting and allow the min to go below 10k?

andrross: Not really. `RecoverySettings.INDICES_TRANSLOG_CONCURRENT_RECOVERY_BATCH_SIZE.getKey()` is just a string. The implementation code ultimately just matches on that string. I think it would take some significant surgery to get around the limit for tests.

guojialiang92: Thanks @andrross, I have reviewed the code.
The minimum value is set to `10000` to avoid unnecessary concurrency.
If it is not easy to bypass in the test, I suggest setting the minimum value to `1000` and using the minimum value in the test. The number of documents should be set to `1100`~`2000`. In the local test, the test running time was reduced by `10` times.

PR #20995, opened 2026-03-25
Add filter and filters bucket aggregation in dsl-query-executor plugin

tanik98:
### Description

### Related Issues

### Check List


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationTreeWalker.java
```diff
@@ -154,4 +285,18 @@ private static String granularityKey(List<GroupingInfo> groupings) {
...
+            .collect(Collectors.joining(","));
+        return base + "__filter__" + filterName;
+    }
+
+    private static String filtersGranularityKey(List<GroupingInfo> groupings, String aggName, String filterKey) {
+        String base = groupings.stream()
+            .flatMap(g -> g.getFieldNames().stream())
+            .collect(Collectors.joining(","));
+        return base + "__filter__" + aggName + "/" + filterKey;
+    }
```
nssuresh2007: Can we think about implementing this in a generic manner so that it can be extensible for any other aggregation as well?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationTreeWalker.java
```diff
@@ -93,32 +119,129 @@ private void handleBucket(
...
+            builder.setAggregationName(aggBuilder.getName());
+
+            if (subAggs != null && subAggs.isEmpty() == false) {
+                walkRecursive(subAggs, accumulatedGroupings, combinedFilter, granKey, granularities, ctx);
+            }
+        }
+
+        // Other bucket: NOT(OR(all filters))
+        if (aggBuilder.otherBucket()) {
+            RexNode otherCondition = translator.buildOtherBucketCondition(filterConditions, ctx);
```
nssuresh2007: I think we are creating AND RexNode within this method, whereas we are invoking the FilterTranslator to create NOT RexNode. Can we have a common logic to create for all kinds of filters?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationTreeWalker.java
```diff
@@ -73,16 +86,29 @@ public List<AggregationMetadata> walk(
...
         Map<String, AggregationMetadataBuilder> granularities,
-        RelDataType rowType
+        ConversionContext ctx
     ) throws ConversionException {
         for (AggregationBuilder aggBuilder : aggs) {
             AggregationType<?> type = registry.get(aggBuilder.getClass());
 
-            if (type instanceof BucketTranslator) {
-                handleBucket((BucketTranslator<AggregationBuilder>) type, aggBuilder, currentGroupings, granularities, rowType);
+            if (type instanceof FiltersBucketTranslator) {
```
nssuresh2007: While I understand filter bucket is unique in such that it adds a filter condition, can we avoid adding specific instanceof check for a particular translator, since it breaks the generic nature?
Instead, can we add a new method `getWalkStrategy()` to the `BucketTranslator` and based on walk strategy (FILTER vs STANDARD) we can execute this branch condition. Thoughts?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationTreeWalker.java
tanik98: Generalize the method for more type of aggrgegations.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationTreeWalker.java
tanik98: Moved filter bucket creation to FiltersBucketTranslator.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationTreeWalker.java
tanik98: Added getWalkStartegy and removed instance of check.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/bucket/FilterBucketTranslator.java
```diff
@@ -0,0 +1,105 @@
...
+    public RexNode combineWithParent(RexNode parentCondition, RexNode filterCondition, ConversionContext ctx) {
+        if (parentCondition == null) {
+            return filterCondition;
+        }
+        return ctx.getRexBuilder().makeCall(SqlStdOperatorTable.AND, parentCondition, filterCondition);
+    }
+
+    @Override
+    public InternalAggregation toBucketAggregation(FilterAggregationBuilder agg, List<BucketEntry> buckets) {
+        throw new UnsupportedOperationException("toBucketAggregation not yet implemented");
```
nssuresh2007: Why are we not implementing this method? Lets add reconstruction logic as well in the same PR

tanik98: Was thinking of adding it when we add response convertor, but let me add the implementation now.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationTreeWalker.java
```diff
@@ -93,32 +123,91 @@ private void handleBucket(
...
 
         // Recurse into sub-aggregations
         Collection<AggregationBuilder> subAggs = translator.getSubAggregations(aggBuilder);
         if (subAggs != null && !subAggs.isEmpty()) {
-            walkRecursive(subAggs, accumulatedGroupings, granularities, rowType);
+            walkRecursive(subAggs, accumulatedGroupings, combinedFilter, filterKey, granularities, ctx);
+        }
+    }
+
+    private void handleFiltersBucket(
```
nssuresh2007: nit: Should we name it `handleMultiFiltersBucket` since this is invoked only in that case? Also, please add javadoc for this method

nssuresh2007 (commented):
LGTM. Just few minor comments

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationTreeWalker.java
tanik98: Updated

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/bucket/BucketTranslator.java
```diff
@@ -23,6 +23,29 @@
...
+    /**
+     * Determines how the {@link org.opensearch.dsl.aggregation.AggregationTreeWalker}
+     * processes this bucket translator during tree traversal.
+     */
+    enum WalkStrategy {
+        /** Standard bucket walk — adds grouping columns (e.g., terms). */
+        STANDARD,
+        /** Single filter walk — produces one plan with a filter condition (e.g., filter). */
+        FILTER,
+        /** Multi-filter walk — produces N plans, one per filter, plus optional other bucket (e.g., filters). */
```
bharath-techie, line 35: Do we know the performance of this ? For N plans, lets have an thread to discuss performance and whether we need it initially or whether we need to add restrictions.

tanik98: This we are yet to finalize, if we need to limit the number of filters based on performance numbers.

PR #20996, opened 2026-03-25
Removing feature flags for tiered storage settings

astute-decipher:
### Description

Removes the feature flag gating for tiered storage (writable warm) index settings, promoting them to always-registered built-in settings.

The following settings are moved from `FEATURE_FLAGGED_INDEX_SETTINGS` to `BUILT_IN_INDEX_SETTINGS` in `IndexScopedSettings`:
- `IndexModule.INDEX_STORE_LOCALITY_SETTING`
- `IndexModule.INDEX_TIERING_STATE`
- `IndexModule.IS_WARM_INDEX_SETTING`

These settings were previously gated behind `WRITABLE_WARM_INDEX_EXPERIMENTAL_FLAG`. With this change, they are always registered so that nodes can parse index metadata containing these settings regardless of whether the feature flag is enabled. This ensures cluster stability during rolling upgrades and mixed-flag configurations.

The `FEATURE_FLAGGED_INDEX_SETTINGS` map is now empty and the `FeatureFlags` import is removed.

### Related Issues

### Check List

### DIFF
--- a/.github/workflows/nightly-precommit.yml
+++ b/.github/workflows/nightly-precommit.yml
@@ -35,7 +35,7 @@ jobs:
 
   open-issue-on-failure:
     if: failure() && github.event_name == 'schedule'
-    needs: precommit
+    needs: nightly-precommit
     runs-on: ubuntu-latest
     permissions:
       issues: write


ankitkala (approved):

server/src/main/java/org/opensearch/common/settings/IndexScopedSettings.java
```diff
@@ -305,6 +304,12 @@ public final class IndexScopedSettings extends AbstractScopedSettings {
                 IndexSettings.INDEX_DERIVED_SOURCE_SETTING,
                 IndexSettings.INDEX_DERIVED_SOURCE_TRANSLOG_ENABLED_SETTING,
 
+                // Writable warm / tiering settings - always registered so nodes can parse
+                // index metadata even when the feature flag is disabled
```
linuxpi, line 308: what would be the behavior now in mixed mode clusters? since these settings would be recognized by only a subset of nodes? Could we add a test for it?

astute-decipher: Should not be an issue. We can consider following scenarios :

1. New CM → Old data node: If a new cluster-manager publishes index metadata with these settings, old nodes that don't recognize them won't hit problems during normal cluster state application - `IndexMetadata.readFrom()` deserializes settings as raw key-value pairs via 
`readSettingsFromStream()` without any validation. The archiveBrokenIndexSettings path only runs during gateway recovery (loading from disk), not when receiving cluster state over the wire. And even in that archival case, the index still loads fine - the settings just get 
prefixed with archived..

2. Old CM → New data node: The old CM wouldn't add these settings to index metadata in the first place (flag off), so no conflict.

3. Practical risk: These settings have safe defaults (false, "full", "hot") and won't appear in index metadata unless someone explicitly configures them — which requires the tiered storage feature to be actively used.

Let me know if there is any other case you are referring to here.

linuxpi (approved):

Issue #12113, opened 2024-01-31
[Feature Request] Field level statistics of lucene index files


rishabhmaurya:
### Is your feature request related to a problem? Please describe

It seems impossible to get statistics such as disk consumption of individual lucene segment files per field. We do have a similar API which can aggregate statistics at shard, index and cluster level using index stats API (https://opensearch.org/docs/latest/api-reference/index-apis/stats/). Like - 
```shell
/<index_name>/_stats/segments?level=shards&include_segment_file_sizes&pretty"
```
```json
  "file_sizes" : {
    "nvm" : {
      "size_in_bytes" : 5486,
      "description" : "Norms"
    },
    "fnm" : {
      "size_in_bytes" : 127478,
      "description" : "Fields"
    },
    "kdd" : {
      "size_in_bytes" : 1119640726,
      "description" : "Others"
    },
    "tmd" : {
      "size_in_bytes" : 59232,
      "description" : "Others"
    },
    "fdm" : {
      "size_in_bytes" : 9175,
      "description" : "Others"
    },
    "kdi" : {
      "size_in_bytes" : 2926068,
      "description" : "Others"
    },
    "dvd" : {
      "size_in_bytes" : 1687766934,
      "description" : "DocValues"
    },
    "kdm" : {
      "size_in_bytes" : 10398,
      "description" : "Others"
    },
    "pos" : {
      "size_in_bytes" : 6051226,
      "description" : "Positions"
    },
    "si" : {
      "size_in_bytes" : 1176,
      "description" : "Segment Info"
    },
    "fdt" : {
      "size_in_bytes" : 9388206796,
      "description" : "Field Data"
    },
    "doc" : {
      "size_in_bytes" : 2964755700,
      "description" : "Frequencies"
    },
    "tim" : {
      "size_in_bytes" : 483838256,
      "description" : "Term Dictionary"
    },
    "dvm" : {
      "size_in_bytes" : 117015,
      "description" : "DocValues"
    },
    "tip" : {
      "size_in_bytes" : 20778232,
      "description" : "Term Index"
    },
    "fdx" : {
      "size_in_bytes" : 519052,
      "description" : "Field Index"
    },
    "nvd" : {
      "size_in_bytes" : 1534,
      "description" : "Norms"
    }
  }
```

### Describe the solution you'd like

Introduce similar API for query param to existing index stats API to provide this information at field level which can be aggregated at shard, index and cluster level per field. 
[...]

PR #21000, opened 2026-03-26
fixed accumaltes size when multi files share the same extension 

ThyTran1402:
### Description
- The root cause that map.put() overwrites same-extension files per segments. And there's no per field size API in Lucene, and files are multi field by design in all major Lucene codecs.
- The fixes:
Change map.put() to map.merge() to correctly accumulate sizes when multiple files share the same extension within a segment.
Adds a regression test in InternalEngineTests that:
disables compound files so segment files remain visible individually,
enables fuzzy-set-for-doc-ID to force multiple Lucene files with the same extension in one segment,
verifies the total reported in file_sizes matches the actual sum of segment file lengths on disk.

### Related Issues
- Related to #12113 

### Check List


msfroh:
@ThyTran1402   I don't think this resolves the linked issue, since (as you called out over there) Lucene creates files per data type per segment, but then throws all fields that have that data type in the same file. IMO, we would need to change the Lucene `Codec` class (actually, mostly the various `*Format` interfaces) to get the size of a field. So, I'm going to modify the description to say "Related to #12113", so that the linked issue doesn't get closed when this gets merged.

Otherwise, the change seems pretty reasonable to me. We should definitely sum across all files with the same extension. Similarly, AFAIK, for vector fields, we end up creating per-field files, which would make the numbers wrong without your fix.

The Gradle check is currently failing on the Spotless check. If you run `./gradlew spotlessApply` that should get fixed. Also, you can remove your change to the changelog file, as we've started autogenerating those based on GitHub PR descriptions, issue descriptions, and commit messages.

Take care of those, and I'm happy to merge this. Thanks! :+1:

msfroh:
@ThyTran1402 Incidentally, if you're interested in trying to tackle the Lucene side of the original issue, it's a little trickier, since it's inherently implementation-specific. 

That said, a good place to start might be (say) `NormsProducer`. You could add a method `getFieldSize(FieldInfo fieldInfo)` with a default implementation that returns `-1` (meaning "I don't know"). Then you could implement it in `Lucene90NormsProducer` and actually calculate the size of a field's norms. For the dense case, I think it's `bytesPerNorm * maxDoc`. For the sparse case, it might be `docsWithFieldLength`, though I'm not completely clear on the norm format implementation.

Then you would need to repeat that for `FieldsProducer`, `DocValuesProducer`, `KnnVectorsReader`, and `PointsReader`. (The other formats, `StoredFieldsFormat`, `TermVectorsFormat`, `FieldInfosFormat`, `SegmentInfoFormat`, `LiveDocsFormat`, and `CompoundFormat`, all store things on a per-segment or per-document basis, not per-field.)

I totally understand if you're not interested in that -- it's quite a lot of work -- but that's a rough outline of what it would take to do it (at least, in my opinion).

ThyTran1402:
> @ThyTran1402 I don't think this resolves the linked issue, since (as you called out over there) Lucene creates files per data type per segment, but then throws all fields that have that data type in the same file. IMO, we would need to change the Lucene `Codec` class (actually, mostly the various `*Format` interfaces) to get the size of a field. So, I'm going to modify the description to say "Related to #12113", so that the linked issue doesn't get closed when this gets merged.
> 
> Otherwise, the change seems pretty reasonable to me. We should definitely sum across all files with the same extension. Similarly, AFAIK, for vector fields, we end up creating per-field files, which would make the numbers wrong without your fix.
> 
> The Gradle check is currently failing on the Spotless check. If you run `./gradlew spotlessApply` that should get fixed. Also, you can remove your change to the changelog file, as we've started autogenerating those based on GitHub PR descriptions, issue descriptions, and commit messages.
> 
> Take care of those, and I'm happy to merge this. Thanks! 👍

Sure! I already cleaned up for Spotless check and remove my change on changelog file. It's great that we started autogenerating changes on changelog files.

ThyTran1402:
> @ThyTran1402 Incidentally, if you're interested in trying to tackle the Lucene side of the original issue, it's a little trickier, since it's inherently implementation-specific.
> 
> That said, a good place to start might be (say) `NormsProducer`. You could add a method `getFieldSize(FieldInfo fieldInfo)` with a default implementation that returns `-1` (meaning "I don't know"). Then you could implement it in `Lucene90NormsProducer` and actually calculate the size of a field's norms. For the dense case, I think it's `bytesPerNorm * maxDoc`. For the sparse case, it might be `docsWithFieldLength`, though I'm not completely clear on the norm format implementation.
> 
> Then you would need to repeat that for `FieldsProducer`, `DocValuesProducer`, `KnnVectorsReader`, and `PointsReader`. (The other formats, `StoredFieldsFormat`, `TermVectorsFormat`, `FieldInfosFormat`, `SegmentInfoFormat`, `LiveDocsFormat`, and `CompoundFormat`, all store things on a per-segment or per-document basis, not per-field.)
> 
> I totally understand if you're not interested in that -- it's quite a lot of work -- but that's a rough outline of what it would take to do it (at least, in my opinion).

Hmmm thank you for your information and letting me know about this. `NormsProducer,` `FieldsProducer,` `DocValuesProducer,` `KnnVectorsReader,` `PointsReader` are Lucene's interfaces, not OpenSearch's. I think adding `getFieldSize(FieldInfo fieldInfo)' to them means I need to contribute to the Apache Lucene project itself in the Apache community. I think that can take quite lot of times. But I can look into that later 😃.

ThyTran1402:
I think I run `./gradlew spotlessApply` and it passed on my local. I'm not sure why it still did not pass CI tests here 😢 @msfroh.

msfroh:
> I think I run `./gradlew spotlessApply` and it passed on my local. I'm not sure why it still did not pass CI tests here 😢 @msfroh.

The latest failure looks unrelated to your change, so I've kicked off a retry. If it fails again, I'll look more closely.

msfroh (approved):

PR #21001, opened 2026-03-26
Updating the FieldMappers to insert call addField of DocumentInput

darjisagar7:
### Description
This PR adds the following functionalities:
1. Introducing a new Feature Flag for the Pluggable dataformat feature. #20644 
2. It adds a new IndexSettings named "index.pluggable.dataformat.enabled", which can be used along with feature flag to enable the pluggable dataformat for an Index.
3. This PR also ensures that the FieldMapper are decoupled from the Lucene's code. Based on the pluggable dataformat architecture the data will be added to DocumentInput which will add the data to respective dataformat. 

### Related Issues

### Check List

### DIFF
--- a/modules/mapper-extras/src/main/java/org/opensearch/index/mapper/RankFeatureFieldMapper.java
+++ b/modules/mapper-extras/src/main/java/org/opensearch/index/mapper/RankFeatureFieldMapper.java
@@ -202,11 +202,6 @@ public class RankFeatureFieldMapper extends ParametrizedFieldMapper {
         context.doc().addWithKey(name(), new FeatureField("_feature", name(), value));
     }
 
-    @Override
-    protected void parseCreateFieldForPluggableFormat(ParseContext context) throws IOException {
-        // TODO check how we can support addWithKey for pluggable dataformat
-    }
-
     private static Float objectToFloat(Object value) {
         if (value instanceof Number number) {
             return number.floatValue();
--- a/server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
+++ b/server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
@@ -349,7 +349,9 @@ public abstract class FieldMapper extends Mapper implements Cloneable {
      * @throws UnsupportedOperationException if the mapper does not support pluggable data formats
      */
     @ExperimentalApi
-    protected abstract void parseCreateFieldForPluggableFormat(ParseContext context) throws IOException;
+    protected void parseCreateFieldForPluggableFormat(ParseContext context) throws IOException {
+        throw new UnsupportedOperationException("Field mapper [" + typeName() + "] does not support pluggable data formats");
+    };
 
     private void extractGroupingCriteriaParams(ParseContext context) throws IOException {
         if (context.docMapper() != null && context.docMapper().mappers() != null) {
--- a/modules/mapper-extras/src/test/java/org/opensearch/index/mapper/RankFeatureFieldMapperTests.java
+++ b/modules/mapper-extras/src/test/java/org/opensearch/index/mapper/RankFeatureFieldMapperTests.java
@@ -168,13 +168,14 @@ public class RankFeatureFieldMapperTests extends MapperTestCase {
     }
 
     @LockFeatureFlag(FeatureFlags.PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG)
-    public void testPluggableDataFormatRankFeatureNoOp() throws IOException {
+    public void testPluggableDataFormatRankFeatureThrows() throws IOException {
         Settings pluggableSettings = Settings.builder().put(getIndexSettings()).put("index.pluggable.dataformat.enabled", true).build();
         DocumentMapper mapper = createDocumentMapper(pluggableSettings, fieldMapping(this::minimalMapping));
         CapturingDocumentInput docInput = new CapturingDocumentInput();
-        mapper.parse(source(b -> b.field("field", 10)), docInput);
-
-        boolean hasField = docInput.getCapturedFields().stream().anyMatch(e -> e.getKey().name().equals("field"));
-        assertFalse("RankFeatureFieldMapper pluggable format is no-op", hasField);
+        MapperParsingException e = expectThrows(
+            MapperParsingException.class,
+            () -> mapper.parse(source(b -> b.field("field", 10)), docInput)
+        );
+        assertThat(e.getCause(), instanceOf(UnsupportedOperationException.class));
     }
 }
--- a/modules/mapper-extras/src/test/java/org/opensearch/index/mapper/ScaledFloatFieldMapperTests.java
+++ b/modules/mapper-extras/src/test/java/org/opensearch/index/mapper/ScaledFloatFieldMapperTests.java
@@ -44,6 +44,7 @@ import org.apache.lucene.index.IndexWriterConfig;
 import org.apache.lucene.index.IndexableField;
 import org.apache.lucene.store.Directory;
 import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.common.CheckedConsumer;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.FeatureFlags;
 import org.opensearch.common.xcontent.XContentFactory;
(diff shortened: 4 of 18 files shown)


server/src/main/java/org/opensearch/index/mapper/Mapper.java
```diff
@@ -303,6 +306,18 @@ protected static boolean hasIndexCreated(Settings settings) {
...
     }
 
+    /**
+     * Checks if the optimised index feature is enabled for the given settings.
+     * Requires both the {@link FeatureFlags#PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG} feature flag
+     *
+     * @param settings the index settings to check
+     * @return {@code true} if the pluggable dataformat feature flag and the optimised index setting are both enabled
+     */
+    public static boolean isOptimisedIndexEnabled(Settings settings) {
```
mgodwan: Can you rename this to isPluggableDataFormatEnabled?

server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
```diff
@@ -217,6 +217,7 @@ public T meta(Map<String, String> meta) {
     protected MultiFields multiFields;
     protected CopyTo copyTo;
     protected DerivedFieldGenerator derivedFieldGenerator;
+    protected int isPluggableDataFormatFeatureEnabled;
```
mgodwan: This should follow an AtomicReference#compareAndSet

mgodwan:
1. For mapper plugins, we need to provide a separate way to allow them to provide how those fields should be written in a data format.
2. Wiring of documentInput into ParseContext would come from the engine, right i.e. prepareIndex?

modules/mapper-extras/src/main/java/org/opensearch/index/mapper/SearchAsYouTypeFieldMapper.java
```diff
@@ -665,6 +675,15 @@ protected void parseCreateField(ParseContext context) throws IOException {
         }
     }
 
+    @Override
+    protected void parseCreateFieldForPluggableFormat(ParseContext context) throws IOException {
+        final String value = context.externalValueSet() ? context.externalValue().toString() : context.parser().textOrNull();
+        if (value == null) {
+            return;
+        }
```
Bukhtawar, line 683: Maybe you could further extract the common logic

```
private String extractValue(ParseContext context) throws IOException {
    String value = context.externalValueSet()
        ? context.externalValue().toString()
        : context.parser().textOrNull();
    return value;
}
```

server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
```diff
@@ -366,11 +387,19 @@ protected final void createFieldNamesField(ParseContext context) {
...
+                    context.doc().add(new Field(FieldNamesFieldMapper.NAME, fieldName, FieldNamesFieldMapper.Defaults.FIELD_TYPE));
+                }
             }
         }
     }
 
+    protected final boolean isPluggableDataFormatFeatureEnabled(ParseContext parseContext) {
+        return isPluggableDataFormatEnabled(parseContext.indexSettings().getSettings());
+    }
+
```
Bukhtawar, line 402: We can cache this instead of reading from index settings everytime?

server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
```diff
@@ -217,6 +219,7 @@ public T meta(Map<String, String> meta) {
     protected MultiFields multiFields;
     protected CopyTo copyTo;
     protected DerivedFieldGenerator derivedFieldGenerator;
+    private final AtomicBoolean isPluggableDataFormatFeatureEnabled = new AtomicBoolean(false);
```
Bukhtawar: This seems unused, prefer using `volatile` once you implement the caching logic

server/src/main/java/org/opensearch/index/mapper/FlatObjectFieldMapper.java
```diff
@@ -593,6 +612,19 @@ private void parseObject(XContentParser parser, ParseContext context) throws IOE
...
+    private void parseObjectForPluggableFormat(XContentParser parser, ParseContext context) throws IOException {
+        assert parser.currentToken() == XContentParser.Token.START_OBJECT;
+        parser.nextToken(); // Skip the outer START_OBJECT. Need to return on END_OBJECT.
+
+        LinkedList<String> path = new LinkedList<>(Collections.singleton(fieldType().name()));
+        HashSet<String> pathParts = new HashSet<>();
+        while (parser.currentToken() != XContentParser.Token.END_OBJECT) {
+            parseToken(parser, context, path, pathParts);
+        }
+
```
Bukhtawar: This logic is again redundant

server/src/main/java/org/opensearch/common/util/FeatureFlags.java
```diff
@@ -74,6 +74,17 @@ public class FeatureFlags {
...
+     * Gates the functionality of pluggable dataformat feature.
+     */
+    public static final String PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG = FEATURE_FLAG_PREFIX + "pluggable.dataformat.enabled";
+
+    public static final Setting<Boolean> PLUGGABLE_DATAFORMAT_EXPERIMENTAL_SETTING = Setting.boolSetting(
+        PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG,
+        false,
+        Property.NodeScope
+    );
+
```
Bukhtawar, line 87: Are we gating behind both feature flag and index settings?

modules/mapper-extras/src/main/java/org/opensearch/index/mapper/RankFeatureFieldMapper.java
```diff
@@ -202,6 +202,11 @@ protected void parseCreateField(ParseContext context) throws IOException {
         context.doc().addWithKey(name(), new FeatureField("_feature", name(), value));
     }
 
+    @Override
+    protected void parseCreateFieldForPluggableFormat(ParseContext context) throws IOException {
+        // TODO check how we can support addWithKey for pluggable dataformat
```
mgodwan: We should throw here if it is not supported?

server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
```diff
@@ -332,6 +337,20 @@ public void parse(ParseContext context) throws IOException {
...
+     * Subclasses that support pluggable data formats should override this method to extract the
+     * parsed value and call {@code context.documentInput().addField(fieldType(), value)}.
+     * The default implementation throws {@link UnsupportedOperationException}.
+     *
+     * @param context the parse context carrying the document input
+     * @throws IOException if an I/O error occurs while parsing
+     * @throws UnsupportedOperationException if the mapper does not support pluggable data formats
+     */
+    @ExperimentalApi
+    protected abstract void parseCreateFieldForPluggableFormat(ParseContext context) throws IOException;
```
mgodwan: We can't have this abstract as this breaks everything that implements FieldMapper outside of the plugin. This should throw unsuspported?

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -860,6 +860,37 @@ public KeywordFieldType fieldType() {
...
+    @Override
+    protected void parseCreateFieldForPluggableFormat(ParseContext context) throws IOException {
+        String value = parseKeywordValue(context);
+        if (value == null) {
+            return;
+        }
+        context.documentInput().addField(fieldType(), value);
+    }
+
+    private String parseKeywordValue(ParseContext context) throws IOException {
```
mgodwan, line 893: nice! how are we testing we are testing all such common functionalities continue to work? I see have a CapturingDocumentInput which is being run against the tests written, but how do we ensure that existing tests can also work there?

server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
darjisagar7: With pluggable dataformat if lucene is selected then every functionality that is working today should work as it is so adding the abstract here ensures that this function is Overriden at all the FieldMapper places.

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
darjisagar7: The current tests are working as it is we have not modified them, so that should give us the confidence that existing flow is working as expected. Or do we need some other mechanism to confirm this?

modules/mapper-extras/src/main/java/org/opensearch/index/mapper/RankFeatureFieldMapper.java
darjisagar7: Agree, I can throw an exception here

server/src/main/java/org/opensearch/index/mapper/FieldMapper.java
mgodwan: This breaks at compile time for any plugin implementing FieldMapper

modules/mapper-extras/src/main/java/org/opensearch/index/mapper/RankFeatureFieldMapper.java
```diff
@@ -204,7 +204,7 @@ protected void parseCreateField(ParseContext context) throws IOException {
 
     @Override
     protected void parseCreateFieldForPluggableFormat(ParseContext context) throws IOException {
-        // TODO check how we can support addWithKey for pluggable dataformat
+        throw new UnsupportedOperationException("Field mapper [" + typeName() + "] does not support pluggable data formats");
```
mgodwan: Do we need to override this separately?

mgodwan: given super is throwing the same

darjisagar7: Removed.

mgodwan (approved):

mgodwan:
Could you please resolve conflicts on this? Rest LGTM

PR #21002, opened 2026-03-26
Fix searchable snapshot directory open pinning problem

### LINKED ISSUE #11676: [BUG] 2.11.1 Search node disk runs full after some time
### Describe the bug

We have dedicated search nodes within the cluster with 100GB of SAN-storage as a cache layer. On normal nodes opensearch would have gone red because the watermarks would have been exceeded. No such notice here, it keeps on filling the disks until it is full after which you get errors like:

image

image

We have not set node.search.cache.size because these nods don't do anything other than searching.

### Related component

Storage:Snapshots

### To Reproduce

Leave the cluster running for a while, do several searches to fill the cache -> error.

### Expected behavior

On a dedicated search node use like 90% of the available storage but nothing more. Prevent stability issues.

### Additional Details

**Plugins**
Default docker 2.11.1 images with S3 plugin installed.

**Screenshots**
N/A

Kubernetes with docker images 2.11.1

**Additional context**

bowenlan-amzn:
### Description

When a searchable snapshot directory is opened, the parent/master `IndexInput` — which only exists as a handle for `clone()`/`slice()` and is never used for searches directly — pin its block in the block cache indefinitely, even before any search has happened.

This PR aims to solve above problem first. 

> IndexInput: Lucene type for reading data/index (input from a file in a Directory). It's not thread safe and keeps internal reading position, so `clone()` is the common pattern to operate on same underlying resources but position independently.

```mermaid
   sequenceDiagram
      participant ROE as ReadOnlyEngine
      participant RSD as RemoteSnapshotDirectory
      participant CFS as _0.cfs Master
      participant Meta as Meta Slices (.fnm .dvm .tmd)
      participant Data as Format Masters (.dvd .doc .tim .tip .nrm .fdt ...)
      participant FC as FileCache

      rect rgb(240, 248, 255)
      Note over ROE,FC: Phase 1: Restore — open()

      Note over ROE,CFS: Lucene90CompoundReader init
      ROE->>RSD: openChecksumInput("_0.cfe")
      RSD->>RSD: RSD.inputsToUnpin.add(cfeMaster)
      Note over RSD,FC: readEntries() → fetchBlock → pinned
      Note over RSD: .cfe closed (try-with-resources) → block already unpinned

      ROE->>RSD: openInput("_0.cfs")
      RSD->>CFS: new OnDemandBlockSnapshotIndexInput
      RSD->>RSD: RSD.inputsToUnpin.add(cfsMaster)
      RSD->>CFS: cfs.inputsToUnpin = new ArrayList (sliceList)
      Note over CFS,FC: checkIndexHeader → fetchBlock(0) → pinned<br/>retrieveChecksum → fetchBlock(last) → pinned

      Note over ROE,FC: SegmentCoreReaders opens format readers via CFS

      ROE->>CFS: openChecksumInput(".fnm") → handle.slice()
      CFS->>Meta: buildSlice(".fnm")
      CFS->>CFS: cfs.inputsToUnpin.add(fnmSlice)
      Note over Meta: FieldInfos read & closed → block already unpinned

      ROE->>CFS: openChecksumInput(".dvm") → handle.slice()
      CFS->>Meta: buildSlice(".dvm")
      CFS->>CFS: cfs.inputsToUnpin.add(dvmSlice)
      Note over Meta: DocValues meta read & closed → block already unpinned

      ROE->>CFS: openInput(".dvd") → handle.slice()
      CFS->>Data: buildSlice(".dvd")
      CFS->>CFS: cfs.inputsToUnpin.add(dvdSlice)
      Note over Data,FC: Stored as DocValuesProducer.data (format master)<br/>reads header → fetchBlock → pinned

      ROE->>CFS: openInput(".doc",".tim",".tip") → handle.slice()
      CFS->>Data: buildSlice() for each
      CFS->>CFS: cfs.inputsToUnpin.add(each)
      Note over Data,FC: Stored as BlockTreeTermsReader.termsIn/.indexIn/.docIn<br/>reads headers → fetchBlock → pinned

      ROE->>CFS: openInput(".nrm",".fdt",".fdx",...) → handle.slice()
      CFS->>Data: buildSlice() for each
      CFS->>CFS: cfs.inputsToUnpin.add(each)
      Note over Data,FC: Stored as NormsProducer.data / StoredFieldsReader.fieldsStream<br/>reads headers → fetchBlock → pinned
      end

      rect rgb(255, 248, 240)
      Note over ROE,FC: Phase 2: unpinBlocksAfterRecovery()
      ROE->>RSD: unpinBlocksAfterRecovery()
      loop for each master in RSD.inputsToUnpin
          Note over RSD: .cfe master: unpinBlock → block==null → no-op
          RSD->>CFS: cfsMaster.unpinBlock()
          loop for each slice in cfs.inputsToUnpin
              CFS->>Meta: unpinBlock → block==null → no-op (already closed)
              CFS->>Data: unpinBlock → blockHolder.close()
              Data->>FC: decRef() → block moves to LRU
          end
          CFS->>CFS: cfs.inputsToUnpin = null (releases slice references)
          CFS->>FC: blockHolder.close() → decRef()
      end
      RSD->>RSD: RSD.inputsToUnpin.clear() (releases master references)
      Note over FC: All blocks now in LRU — warm cache, evictable when space needed
      end

      rect rgb(240, 255, 240)
      Note over ROE,FC: Phase 3: Search
      ROE->>Data: clone() format master (e.g. DocValuesProducer.data)
      Note over Data,FC: Search thread reads clone → fetchBlock on demand → decRef after use
      end

```

---

### Another bigger problem on searchable snapshot search time described below.

**Background**

Searchable snapshots work a bit differently from standard Lucene. The actual data lives in remote storage, so we introduce a virtual `IndexInput` to keep Lucene happy. Under the hood, this virtual `IndexInput` works one block at a time — each block is an 8MB chunk fetched from remote and cached locally.

**The problem with ref-counting**

To track whether a block is actively being read, the current implementation uses reference counting. Lucene typically `clone()`s or `slice()`s an `IndexInput` to read from it, and never explicitly calls `close()`. That's fine for standard Lucene — a clone or slice is just a cheap cursor into a memory-mapped region that can always be paged back in from disk. But we don't have all the data on disk for remote store, so we have to explicitly fetch and open each block. Without `close()` being called, reference counts won't go down, so all they really tell us is that a search started — not that it finished.

The actual cleanup relies on a `Cleaner` hooked into JVM GC, which closes `IndexInput` clones once they're no longer referenced — the same way standard Lucene handles it.

**A simpler option**

Instead of all this, why not just let the block cache decide what to evict? The cache should be smart enough to evict the coldest, oldest, and least-used entries. If an evicted block happens to be used by a search, that search fails — and that could be a useful signal. It either means the cache is too small for the workload, or the cache eviction policy  had bad timing and caught an in-flight search off guard. Either way, the right thing to do is surface this as a clear error message on the failed search.

Lucene MemorySegmentIndexInput has some protection of unmap in the middle of reading
https://github.com/apache/lucene/blob/cba40dd8363148463fbc1359509e8affa18ebac6/lucene/core/src/java/org/apache/lucene/store/MemorySegmentIndexInput.java#L681-L694

Lucene IndexInput expects [this contract](https://github.com/apache/lucene/blob/cba40dd8363148463fbc1359509e8affa18ebac6/lucene/core/src/java/org/apache/lucene/store/IndexInput.java#L34-L38) 
```
If you access the cloned IndexInput after closing the original object, any readXXX methods will throw AlreadyClosedException.
```

---

Note:
- The lucene indexinput held in the block of SnapshotIndexInput is a clone of the master one cached. And refcounting only tracks active clones.
https://github.com/opensearch-project/OpenSearch/blob/527ef863ff2d2f3daf9e9dc2107280ff3c319bec/server/src/main/java/org/opensearch/index/store/remote/utils/TransferManager.java#L97
master one opened here
https://github.com/opensearch-project/OpenSearch/blob/527ef863ff2d2f3daf9e9dc2107280ff3c319bec/server/src/main/java/org/opensearch/index/store/remote/utils/TransferManager.java#L179

- decRef to 0 in cache
https://github.com/opensearch-project/OpenSearch/blob/d151cfa5d0f281163071144e31db2607a29112d5/server/src/main/java/org/opensearch/index/store/remote/utils/cache/LRUCache.java#L262-L268

### Related Issues
Resolves #11676

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
+++ b/server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
@@ -49,6 +49,7 @@ import org.opensearch.common.util.io.IOUtils;
 import org.opensearch.index.seqno.SeqNoStats;
 import org.opensearch.index.seqno.SequenceNumbers;
 import org.opensearch.index.store.Store;
+import org.opensearch.index.store.remote.directory.BlockUnpinningDirectory;
 import org.opensearch.index.store.remote.directory.RemoteSnapshotDirectory;
 import org.opensearch.index.translog.DefaultTranslogDeletionPolicy;
 import org.opensearch.index.translog.NoOpTranslogManager;
@@ -140,9 +141,20 @@ public class ReadOnlyEngine extends Engine {
                     ensureMaxSeqNoEqualsToGlobalCheckpoint(seqNoStats);
                 }
                 this.seqNoStats = seqNoStats;
-                this.indexCommit = Lucene.getIndexCommit(lastCommittedSegmentInfos, directory);
+                // For searchable snapshots, wrap the directory so we can track and
+                // unpin all file cache blocks fetched during DirectoryReader.open().
+                // After the reader is opened (but before any searches), we release
+                // the hard references so blocks become LRU-evictable.
+                BlockUnpinningDirectory unpinningDirectory = null;
+                if (FilterDirectory.unwrap(directory) instanceof RemoteSnapshotDirectory) {
+                    unpinningDirectory = new BlockUnpinningDirectory(directory);
+                }
+                Directory openDirectory = unpinningDirectory != null ? unpinningDirectory : directory;
+                this.indexCommit = Lucene.getIndexCommit(lastCommittedSegmentInfos, openDirectory);
                 reader = wrapReader(open(indexCommit), readerWrapperFunction);
-                unpinBlocksAfterRecovery(directory);
+                if (unpinningDirectory != null) {
+                    unpinningDirectory.unpinAndStopTracking();
+                }
                 readerManager = new OpenSearchReaderManager(reader);
                 assert translogStats != null || obtainLock : "mutiple translogs instances should not be opened at the same time";
                 this.translogStats = translogStats != null ? translogStats : translogStats(config, lastCommittedSegmentInfos);
@@ -221,13 +233,6 @@ public class ReadOnlyEngine extends Engine {
         assert Transports.assertNotTransportThread("opening index commit of a read-only engine");
         DirectoryReader reader = DirectoryReader.open(commit);
         return new SoftDeletesDirectoryReaderWrapper(reader, Lucene.SOFT_DELETES_FIELD);
-    }
-
-    private static void unpinBlocksAfterRecovery(Directory directory) {
-        Directory unwrapped = FilterDirectory.unwrap(directory);
-        if (unwrapped instanceof RemoteSnapshotDirectory) {
-            ((RemoteSnapshotDirectory) unwrapped).unpinBlocksAfterRecovery();
-        }
     }
 
     @Override
--- a/server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
+++ b/server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
@@ -39,6 +39,7 @@ import org.apache.lucene.index.SegmentInfos;
 import org.apache.lucene.index.SoftDeletesDirectoryReaderWrapper;
 import org.apache.lucene.search.ReferenceManager;
 import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FilterDirectory;
 import org.apache.lucene.store.Lock;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.common.concurrent.GatedCloseable;
@@ -48,6 +49,7 @@ import org.opensearch.common.util.io.IOUtils;
 import org.opensearch.index.seqno.SeqNoStats;
 import org.opensearch.index.seqno.SequenceNumbers;
 import org.opensearch.index.store.Store;
+import org.opensearch.index.store.remote.directory.RemoteSnapshotDirectory;
 import org.opensearch.index.translog.DefaultTranslogDeletionPolicy;
 import org.opensearch.index.translog.NoOpTranslogManager;
 import org.opensearch.index.translog.Translog;
@@ -140,6 +142,7 @@ public class ReadOnlyEngine extends Engine {
                 this.seqNoStats = seqNoStats;
                 this.indexCommit = Lucene.getIndexCommit(lastCommittedSegmentInfos, directory);
                 reader = wrapReader(open(indexCommit), readerWrapperFunction);
+                unpinBlocksAfterRecovery(directory);
                 readerManager = new OpenSearchReaderManager(reader);
                 assert translogStats != null || obtainLock : "mutiple translogs instances should not be opened at the same time";
                 this.translogStats = translogStats != null ? translogStats : translogStats(config, lastCommittedSegmentInfos);
@@ -218,6 +221,13 @@ public class ReadOnlyEngine extends Engine {
         assert Transports.assertNotTransportThread("opening index commit of a read-only engine");
         DirectoryReader reader = DirectoryReader.open(commit);
         return new SoftDeletesDirectoryReaderWrapper(reader, Lucene.SOFT_DELETES_FIELD);
+    }
+
+    private static void unpinBlocksAfterRecovery(Directory directory) {
+        Directory unwrapped = FilterDirectory.unwrap(directory);
+        if (unwrapped instanceof RemoteSnapshotDirectory) {
+            ((RemoteSnapshotDirectory) unwrapped).unpinBlocksAfterRecovery();
+        }
     }
 
     @Override
--- a/server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
+++ b/server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
@@ -141,19 +141,15 @@ public class ReadOnlyEngine extends Engine {
                     ensureMaxSeqNoEqualsToGlobalCheckpoint(seqNoStats);
                 }
                 this.seqNoStats = seqNoStats;
-                // For searchable snapshots, wrap the directory so we can track and
-                // unpin all file cache blocks fetched during DirectoryReader.open().
-                // After the reader is opened (but before any searches), we release
-                // the hard references so blocks become LRU-evictable.
-                BlockUnpinningDirectory unpinningDirectory = null;
+                this.indexCommit = Lucene.getIndexCommit(lastCommittedSegmentInfos, directory);
                 if (FilterDirectory.unwrap(directory) instanceof RemoteSnapshotDirectory) {
-                    unpinningDirectory = new BlockUnpinningDirectory(directory);
-                }
-                Directory openDirectory = unpinningDirectory != null ? unpinningDirectory : directory;
-                this.indexCommit = Lucene.getIndexCommit(lastCommittedSegmentInfos, openDirectory);
-                reader = wrapReader(open(indexCommit), readerWrapperFunction);
-                if (unpinningDirectory != null) {
-                    unpinningDirectory.unpinAndStopTracking();
+                    BlockUnpinningDirectory unpinningDir = new BlockUnpinningDirectory(directory);
+                    // Ephemeral commit — unpinningDir only needs to live during open();
+                    // indexCommit is long-lived and used by recovery, metadata reads, etc.
+                    reader = wrapReader(open(Lucene.getIndexCommit(lastCommittedSegmentInfos, unpinningDir)), readerWrapperFunction);
+                    unpinningDir.unpinAndStopTracking();
+                } else {
+                    reader = wrapReader(open(indexCommit), readerWrapperFunction);
                 }
                 readerManager = new OpenSearchReaderManager(reader);
                 assert translogStats != null || obtainLock : "mutiple translogs instances should not be opened at the same time";
--- a/server/src/main/java/org/opensearch/index/store/remote/directory/BlockUnpinningDirectory.java
+++ b/server/src/main/java/org/opensearch/index/store/remote/directory/BlockUnpinningDirectory.java
@@ -1,0 +1,82 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.store.remote.directory;
+
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FilterDirectory;
+import org.apache.lucene.store.IOContext;
+import org.apache.lucene.store.IndexInput;
+import org.opensearch.index.store.remote.file.AbstractBlockIndexInput;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A {@link FilterDirectory} wrapper that tracks {@link AbstractBlockIndexInput}
+ * instances opened during searchable snapshot restore so their file cache blocks
+ * can be unpinned in bulk once the reader is fully opened.
+ * <p>
+ * Lifecycle (used inside {@code ReadOnlyEngine} constructor):
+ * <ol>
+ *   <li>Wrap the directory: {@code new BlockUnpinningDirectory(directory)}</li>
+ *   <li>{@code DirectoryReader.open()} — every {@link #openInput} call tracks
+ *       the returned input; slices and clones are tracked automatically via
+ *       the {@code onClone} callback.</li>
+ *   <li>{@link #unpinAndStopTracking()} — releases all held blocks (LRU-evictable)
+ *       and disables further tracking.</li>
+ * </ol>
+ * <p>
+ * Concurrency: {@link #openInput} may be called from multiple threads during
+ * parallel segment opening, so {@code addToTracked} is synchronized. After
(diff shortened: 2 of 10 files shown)


server/src/main/java/org/opensearch/index/store/remote/file/AbstractBlockIndexInput.java
```diff
@@ -446,11 +446,25 @@ private void demandBlock(int blockId) throws IOException {
...
+     * lifecycle -- subsequent reads will re-fetch the block lazily via
+     * {@link #demandBlock}. After this call, {@link #getFilePointer()}
+     * returns 0.
+     * <p>
+     * Called on parent IndexInputs during {@code clone()} so that blocks
+     * pinned during codec reader init don't stay pinned indefinitely.
+     * Calls {@code blockHolder.close()} directly rather than {@link #close()}
+     * to avoid triggering subclass lifecycle cleanup mid-use.
+     */
+    protected void releaseBlock() {
```
finnegancarroll: >lifecycle -- subsequent reads will re-fetch the block lazily

+1 for this approach.

Do we need to further update the existing code to handle this lazy block "re-fetch" logic? For something like:

```
    @Override
    public int readInt() throws IOException {
        if (blockHolder.block != null && Integer.BYTES <= (blockSize - currentBlockPosition())) {
            return blockHolder.block.readInt();
        } else {
            return super.readInt();
        }
    }
```

We need to attempt to re-fetch the block when `blockHolder.block == null` right?

bowenlan-amzn: It will be the same logic like this indexinput just init and without any block yet.

server/src/main/java/org/opensearch/index/store/remote/file/OnDemandBlockSnapshotIndexInput.java
```diff
@@ -132,6 +157,10 @@ protected OnDemandBlockSnapshotIndexInput buildSlice(String sliceDescription, lo
             directory,
             transferManager
         );
+        if (inputsToUnpin != null) {
+            inputsToUnpin.add(slice);
```
andrross: A consequence here is that the child IndexInputs created in this method will never go out of scope (and therefore be cleaned by the cleaner) until the parent is GC'd because the parent is holding references.

bowenlan-amzn: Yes, this is intentional for compound format.

The lifecycle

1. Recovery openInput("_0.cfs") creates CFS master, sets inputsToUnpin = new ArrayList<>(). 
    - Lucene reads the CFS directory, calling buildSlice() for each sub-file → slices added to inputsToUnpin list. 
    - Codec reader open slices and read blocks → blocks pinned in FileCache because these slices will be held by codec reader 
2. Right after open() returns: unpinBlocksAfterRecovery() is called in ReadOnlyEngine. This calls unpinBlock() on CFS master → iterates slices, unpins each → sets inputsToUnpin = null → releases all references. Then RemoteSnapshotDirectory.inputsToUnpin.clear() releases references to masters too.
3. After recovery: inputsToUnpin is null on the CFS master. Any subsequent buildSlice() calls (from search-time clone()) see inputsToUnpin== null and skip the add — no tracking, no extra references.

server/src/main/java/org/opensearch/index/store/remote/directory/RemoteSnapshotDirectory.java
```diff
@@ -70,7 +77,24 @@ public IndexInput openInput(String name, IOContext context) throws IOException {
         if (fileInfo.name().startsWith(VIRTUAL_FILE_PREFIX)) {
             return new ByteArrayIndexInput(fileInfo.physicalName(), fileInfo.metadata().hash().bytes);
         }
-        return new OnDemandBlockSnapshotIndexInput(fileInfo, localStoreDir, transferManager);
+        OnDemandBlockSnapshotIndexInput master = new OnDemandBlockSnapshotIndexInput(fileInfo, localStoreDir, transferManager);
+        inputsToUnpin.add(master);
```
andrross: Is this a memory leak? After `unpinBlocksAfterRecovery()` is called, this will continue tracking all new IndexInputs but never be cleared.

server/src/main/java/org/opensearch/index/store/remote/directory/BlockUnpinningDirectory.java
```diff
@@ -0,0 +1,82 @@
...
+ * </ol>
+ * <p>
+ * Concurrency: {@link #openInput} may be called from multiple threads during
+ * parallel segment opening, so {@code addToTracked} is synchronized. After
+ * {@link #unpinAndStopTracking()} sets {@code tracked} to {@code null},
+ * {@link #openInput} becomes a plain delegation with only a volatile read overhead.
+ *
+ * @opensearch.internal
+ */
+public final class BlockUnpinningDirectory extends FilterDirectory {
```
andrross, line 43: Wow this code looks really good. Whoever wrote it must be really smart.

server/src/main/java/org/opensearch/index/store/remote/directory/BlockUnpinningDirectory.java
```diff
@@ -0,0 +1,82 @@
...
+            addToTracked(blockInput);
+        }
+        return input;
+    }
+
+    private void track(AbstractBlockIndexInput input) {
+        addToTracked(input);
+    }
+
+    private synchronized void addToTracked(AbstractBlockIndexInput input) {
```
andrross: I _think_ the only concurrency you need to handle is concurrent adds to the list. If DirectoryReader.open() does things in a parallel, there's a join barrier before it returns. The iteration in unpinAndStopTracking() is not thread safe and does not need to be. This could just be a `ConcurrentLinkedQueue` which handles concurrent adds and is cheaper than synchronizing the whole method.

server/src/main/java/org/opensearch/index/store/remote/file/AbstractBlockIndexInput.java
```diff
@@ -136,6 +150,18 @@ public void close() throws IOException {
         currentBlockId = 0;
     }
 
+    /**
+     * Unpin the currently held block
+     */
+    public void unpinBlock() {
```
andrross, line 164: My concern here is that we do in fact lose the file pointer position when unpinning:

```java
@Override
public long getFilePointer() {
    if (blockHolder.block == null) return 0L;
    return currentBlockStart() + currentBlockPosition() - offset;
}
```

after unpin() this will return zero instead of where the cursor was prior to unpinning

andrross: I think this change will allow reads to continue as-is. Looks not too complicated: https://github.com/andrross/OpenSearch/commit/9b5e1731450ce9afebfeeefc7b131528d10172a4

server/src/main/java/org/opensearch/index/store/remote/directory/BlockUnpinningDirectory.java
bowenlan-amzn: Absolutely!

server/src/main/java/org/opensearch/index/store/remote/directory/BlockUnpinningDirectory.java
bowenlan-amzn: Updated

server/src/main/java/org/opensearch/index/store/remote/file/AbstractBlockIndexInput.java
bowenlan-amzn: There's an assumption here

IndexInputs opened, sliced, or cloned via `openInput` during shard initialization serve as masters. Masters are never used for reads directly — they exist solely as source handle for subsequent clones. 

We want to only unpin the masters. It's true that the file position will be lost after unpin. But considering masters will not be used anymore after unpin, this is safe.

---

Preserve file position makes sense if the underlying block can be destroyed during read on the indexinput.

andrross: Yeah, my concern is about the assumption holding true in the future. If it doesn't hold true, then there's a subtle correctness bug. `indexinput.readInt(); indexinput.unpin(); indexinput.readInt()` works with this code, it's just the second readInt() returns the wrong value assuming the caller expected the second int in the stream.

As long as the indexinput saves the position properly, then unpin() becomes totally transparent. Any single-threaded usage of an IndexInput can call unpin() at any point and it will still be correct. We won't have to worry about the assumption being violated because no assumption is being made.

bowenlan-amzn: Now explicitly throw if read happens after unpin.

andrross (approved):

server/src/main/java/org/opensearch/index/store/remote/file/AbstractBlockIndexInput.java
andrross: I still think it's reasonable to make unpin transparent: just save your position and fetch the block again if you need it. However, we don't actually need that functionality now, so this is perfectly safe too.

server/src/main/java/org/opensearch/index/store/remote/file/AbstractBlockIndexInput.java
```diff
@@ -136,8 +157,29 @@ public void close() throws IOException {
...
+        try {
+            blockHolder.close();
+        } catch (IOException e) {
+            logger.info("Exception thrown while unpin block", e);
+        }
+    }
+
+    private void ensureNotUnpinned() {
+        if (unpinned) {
+            throw new AlreadyClosedException("This input has been unpinned and must not be read from: " + this);
```
finnegancarroll, line 176: wdyt about an error message here which indicates the file cache is too small/full? Just thinking the concept of pinning might not be clear to users.

server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
```diff
@@ -139,7 +142,15 @@ public ReadOnlyEngine(
                 }
                 this.seqNoStats = seqNoStats;
                 this.indexCommit = Lucene.getIndexCommit(lastCommittedSegmentInfos, directory);
-                reader = wrapReader(open(indexCommit), readerWrapperFunction);
+                if (FilterDirectory.unwrap(directory) instanceof RemoteSnapshotDirectory) {
+                    BlockUnpinningDirectory unpinningDir = new BlockUnpinningDirectory(directory);
+                    // Ephemeral commit — unpinningDir only needs to live during open();
+                    // indexCommit is long-lived and used by recovery, metadata reads, etc.
+                    reader = wrapReader(open(Lucene.getIndexCommit(lastCommittedSegmentInfos, unpinningDir)), readerWrapperFunction);
+                    unpinningDir.unpinAndStopTracking();
```
finnegancarroll, line 150: Why do we need two separate directory implementations here? 

Can we instead migrate `unpinAndStopTracking` to `RemoteSnapshotDirectory` and just invoke it after we open our reader?

finnegancarroll (dismissed):

server/src/main/java/org/opensearch/index/store/remote/file/AbstractBlockIndexInput.java
bowenlan-amzn: This error is not related to file cache size. It is a defensive guard against unexpected usage outside our current assumptions.
The assumption is: any `IndexInput` unpinned after shard initialization will never be read again.

This is mainly for developer, not normal users, so I think it's fine.

server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
bowenlan-amzn: This is a wrapper/decorator to offload the unpin logic away from `RemoteSnapshotDirectory` so that can stay stateless/clean

https://github.com/opensearch-project/OpenSearch/pull/21002#discussion_r3034158945

server/src/main/java/org/opensearch/index/store/remote/file/AbstractBlockIndexInput.java
andrross: Yeah, this exception should be impossible to hit unless there is a bug

server/src/main/java/org/opensearch/index/engine/ReadOnlyEngine.java
andrross: There are two behavior differences with BlockUnpinningDirectory: it starts tracking immediately upon construction, and then it stops once unpinAndStopTracking() is called. RemoteSnapshotDirectory never needs to do this, so keeping a separate wrapper is a clean way to track this state.

andrross (approved):

sandervandegeijn:
Guys this is fantastic news. I would like to thank you immensely, you have made the security team at Wageningen and some other universities in the Netherlands very happy.

@bowenlan-amzn  I'll ping you on linkedin, I'll try to get a small present to you as appreciation from us at Wageningen.

mvukadinoff:
Thanks for the great work on this. Is there any chance this will be backported to Opensearch 2.XX latest patch version? We're using a software (Graylog) which is not officially supporting version 3.X

Issue #20988, opened 2026-03-24
[Feature Request] Use pre-built SNAPSHOT artifacts for BWC tests instead of building from source


andrross:
### Background

The BWC test infrastructure currently builds unreleased versions from source by checking out the corresponding git branch and running a full `./gradlew assemble` inside the checkout. This is implemented in `InternalDistributionBwcSetupPlugin` and `BwcSetupExtension`, which:

1. Clone/checkout the BWC branch into `build/bwc/checkout-{branch}`
2. Resolve the required JDK version from `.ci/java-versions.properties` in the checkout
3. Run a full Gradle build of the distribution archives inside the checkout
4. Use the resulting artifacts to start the BWC test cluster

The original rationale is stated in the code:

> We want to be able to do BWC tests for unreleased versions without relying on and waiting for snapshots.

### Problem

Building from source is expensive and introduces complexity:

- Requires multiple JDK versions to be available (the BWC branch may need a different JDK than the current branch)
- The `BwcSetupExtension` has to parse `.ci/java-versions.properties` from the checkout and set `JAVA_HOME`/`RUNTIME_JAVA_HOME` accordingly
- Git operations can fail or be slow, especially in CI environments
- The nested Gradle invocation is opaque to build caching and difficult to debug

### Proposal

Replace the build-from-source approach with resolution of pre-built SNAPSHOT artifacts from the Maven snapshot repository for unreleased versions.

We now have reliable, regularly published SNAPSHOT artifacts for all active branches. The tradeoff is that the SNAPSHOT artifact may lag behind the BWC branch tip by a small number of commits. I think this is probably fine because the BWC branches are generally not fast-moving and snapshots are updated quickly. Also the entire plugin ecosystem relies on these snapshots, so this is not a new requirement.

### Related component

Build

andrross:
@reta Would appreciate your opinion here as you know this area as well as anyone.

Also I think some of my recent hackery in this area would not be required if we took this approach:

- #20871
- #20897

PR #21008, opened 2026-03-26
Support downloading pre-built SNAPSHOT artifacts for BWC tests

andrross:
Add a `-Dbwc.buildUnreleasedFromSource=false` flag that downloads pre-built SNAPSHOT artifacts from artifacts.opensearch.org for unreleased versions instead of checking out git branches and compiling from source. The default behavior (building from source) is unchanged.

This is the first step towards #20988. This PR adds the option to run the following:

```
./gradlew :qa:full-cluster-restart:v2.19.6#bwcTest -Dbwc.buildUnreleasedFromSource=false
```

and as the parameter suggests this will not build 2.19.6 from source but instead fetch the SNAPSHOT artifact. Currently this does not work with 3.5.1 due to a FIPS issue with that distribution, but I'm still working on that. I believe the code here is correct.

### Check List

### DIFF
--- a/buildSrc/src/main/java/org/opensearch/gradle/DistributionDownloadPlugin.java
+++ b/buildSrc/src/main/java/org/opensearch/gradle/DistributionDownloadPlugin.java
@@ -32,7 +32,6 @@
 
 package org.opensearch.gradle;
 
-import org.opensearch.gradle.OpenSearchDistribution.Platform;
 import org.opensearch.gradle.OpenSearchDistribution.Type;
 import org.opensearch.gradle.docker.DockerSupportPlugin;
 import org.opensearch.gradle.docker.DockerSupportService;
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/java/org/opensearch/gradle/DistributionDownloadPlugin.java
```diff
@@ -221,40 +220,11 @@ private static void setupDownloadServiceRepo(Project project) {
      * coordinates that resolve to the OpenSearch download service through an ivy repository.
      */
     private String dependencyNotation(OpenSearchDistribution distribution) {
-        Version distroVersion = Version.fromString(distribution.getVersion());
         if (distribution.getType() == Type.INTEG_TEST_ZIP) {
             return "org.opensearch.distribution.integ-test-zip:opensearch:" + distribution.getVersion() + "@zip";
         }
 
-        String extension = distribution.getType().toString();
-        String classifier = distroVersion.onOrAfter("1.0.0") ? ":x64" : ":x86_64";
```
andrross, line 230: Subtle change, but I removed the branching here assuming that we have no need to support anything pre-1.0.0.

PR #21011, opened 2026-03-27
Add optional ingest config readers

### LINKED ISSUE #21010: # [Feature Request] Add readOptionalBooleanProperty to Ingest ConfigurationUtils

The ingest processor configuration utility `ConfigurationUtils` is missing a `readOptionalBooleanProperty()` method. Currently, `readBooleanProperty()` always requires a default value to be provided. For processor authors who want to distinguish between "the user explicitly set this to false" vs "the user didn't provide this at all" (tri-state logic), there's no clean way to do it. Other types like `String`, `List`, and `Map` all have `readOptional*` counterparts, but `Boolean` and `Integer` do not.

Add a `readOptionalBooleanProperty()` method to `ConfigurationUtils` that returns `null` if the property is missing, and throws an exception if the property is present but not a boolean. This would bring Boolean in line with other types that already have `readOptional*` methods.

### Related component

Other

- Continue using `readBooleanProperty()` with a default value, but this does not allow distinguishing between missing and explicitly set values.
- Manually checking for property existence and type in each processor, which is error-prone and inconsistent.

### Additional context

Other types like `String`, `List`, and `Map` already have `readOptional*` methods. Adding this for Boolean (and potentially Integer/Double) would improve API consistency and usability.

Manishnemade12:
Add optional boolean and int readers to `ConfigurationUtils` so processors can distinguish unset values from explicit `false` or numeric values.

Add unit coverage for missing, null, valid, and invalid inputs for both helpers.

Tests were not run in this environment because the Gradle wrapper required a network download.

### Description
This change adds two missing optional ingest configuration readers to `ConfigurationUtils`:

- `readOptionalBooleanProperty(...)`
- `readOptionalIntProperty(...)`

These helpers make it possible for ingest processor authors to distinguish between a property that was not provided at all and a property that was explicitly set by the user. That is especially useful for tri-state behavior such as differentiating "unset" from `false`.

The change also keeps integer parsing behavior consistent by routing both required and optional integer reads through the same internal parsing path.

Unit tests were added to cover:

- successful reads for boolean and int values
- missing properties returning `null`
- explicit `null` values returning `null`
- invalid values throwing the expected `OpenSearchParseException`
- configuration removal behavior after reading

### Related Issues
Resolves #21010

### Check List
- [x] API changes companion pull request created, if applicable. Not applicable for this change.
- [x] Public documentation issue/PR created, if applicable. Not applicable for this change.


Manishnemade12:
@cwperks can you please review this pr

Manishnemade12:
@ankitkala @skumawat2025 can you please review this pr

PR #21024, opened 2026-03-27
Add pipeline execution metrics to PollingIngestStats for pull-based ingestion

### LINKED ISSUE #20885: [Feature Request] PollingIngestStats pipeline metrics in pull based ingestion

Add pipeline execution count, failure count to PollingIngestStats

TBA

### Related component

_No response_

_No response_

### Additional context

_No response_

imRishN:
### Description
Add pipeline execution metrics to PollingIngestStats for pull-based ingestion

### Related Issues
Resolves #20885 

### Check List


server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -525,7 +525,13 @@ protected Map<String, String> commitDataAsMap() {
 
     @Override
     public PollingIngestStats pollingIngestStats() {
-        return streamPoller.getStats();
+        PollingIngestStats pollerStats = streamPoller.getStats();
+        // Enrich with pipeline execution metrics from the shared executor
+        return new PollingIngestStats(
```
varunbharadwaj, line 531: Is it possible to pass the pipelineExecutor metrics in the stream poller, when we first construct PollingIngestStats? We can avoid reconstructing another PollingIngestStats here. Wondering if we should first take a look at the same-thread-execution change, and we can avoid some metrics like timeout.

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -51,6 +53,13 @@ public class IngestPipelineExecutor {
     private final String index;
     private volatile String resolvedFinalPipeline;
 
+    // Pipeline execution metrics
+    private final CounterMetric executionCount = new CounterMetric();
+    private final CounterMetric executionTimeNanos = new CounterMetric();
```
varunbharadwaj: Can we check how latency is tracked in the ingest pipeline (and other) flows? I'm guessing we might use MeanMetric instead of CounterMetric.

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -172,4 +191,27 @@ public Map<String, Object> executePipelines(String id, Map<String, Object> sourc
...
+        );
+    }
+
+    /**
+     * Pipeline execution metrics for pull-based ingestion.
+     * Note: totalExecutionCount includes all pipeline invocations (successful, failed, and dropped).
+     * Success count can be derived as: totalExecutionCount - totalFailedCount - totalDroppedCount.
+     */
+    @PublicApi(since = "3.7.0")
+    public record PipelineMetrics(long totalExecutionCount, long totalExecutionTimeInMillis, long totalFailedCount, long totalDroppedCount,
```
varunbharadwaj: Seems PipelineMetrics and PollingIngestStats.PipelineStats is exactly the same? Is it possible to reuse the PollingIngestStats.PipelineStats here so we can also avoid creating redundant intermediate object?

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
imRishN: Same-thread execution is now merged

imRishN: For passing metrics through the poller: I did this intentionally to keep the poller unaware of pipeline concerns. Passing pipelineExecutor into the poller solely for stats collection would add a dependency the poller doesn't need. Wdyt?

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: Yup, changing

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
imRishN: yes updated

server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -120,11 +130,15 @@ public Map<String, Object> executePipelines(String id, Map<String, Object> sourc
             }
         }, slot -> dropped.set(true));
 
+        executionTime.inc(System.nanoTime() - startTimeNanos);
```
varunbharadwaj: Do we want to clear the MeanMetric anytime or continue adding and leave it to shard movement? Assuming the latency here is very low, maybe ~10ms, it should be fine. But if it increases to say 500ms, then we might end up with overflow at very high throughput.

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
varunbharadwaj: It looks like we already pass pipelineExecutor into the poller. But as this is not in the main ingestion path, it should be okay.

varunbharadwaj (approved):

varunbharadwaj:
```
In testFinalPipelineAddsField, pipeline stats are fetched after waitForState completes, but there is no guarantee the stats have been flushed/updated by the time the assertion runs. The totalExecutionCount >= 2 check may be flaky if stats are updated asynchronously or lazily. Consider fetching stats inside waitForState or adding a retry mechanism.
```
@imRishN It looks good to me, but can we double check this. I think the processed count we currently wait for is updated before the ingest pipeline runs and can result in flaky test, as per the AI comment. Adding a waitForSearchableDoc(..) or equivalent can help us ensure the doc is indexed before we validate.

imRishN:
>@imRishN It looks good to me, but can we double check this. I think the processed count we currently wait for is updated before the ingest pipeline runs and can result in flaky test, as per the AI comment. Adding a waitForSearchableDoc(..) or equivalent can help us ensure the doc is indexed before we validate.

1. Metrics are recorded synchronously on the processor thread before indexing.
2. waitForState confirms documents are indexed, which means the processor thread has already passed the metric recording point.

I think we should be good here

PR #21026, opened 2026-03-28
Run BWC tests as separate GitHub Action

### LINKED ISSUE #20988: [Feature Request] Use pre-built SNAPSHOT artifacts for BWC tests instead of building from source

andrross:
This commit adds a GitHub Action to run BWC tests using pre-built snapshots. This avoids the need for any cross-branch coordination. If the versions in these tests become stale, the workflow will add a comment to the issue indicating the problem and how to fix it.

Also add configuration to treat the -SNAPSHOT distributions as changing. This is not needed for the GitHub Action as there is no cache, but is helpful for local testing.

### Related Issues
Resolves #20988

### Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .claude
 CLAUDE.md
 .cursor*
+.kiro*
 
 # intellij files
 .idea/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Add new setting property 'Sensitive' for tiering dynamic settings ([#20901](https://github.com/opensearch-project/OpenSearch/pull/20901))
 - Add ref_path support for package-based hunspell dictionary loading ([#20840](https://github.com/opensearch-project/OpenSearch/pull/20840))
 - Add support for enabling pluggable data formats, starting with phase-1 of decoupling shard from engine, and introducing basic abstractions ([#20675](https://github.com/opensearch-project/OpenSearch/pull/20675))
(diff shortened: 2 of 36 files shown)


andrross:
Closing this in favor of https://github.com/opensearch-project/OpenSearch/pull/21089

Issue #19597, opened 2025-10-10
[FEATURE] Support Jackson 3.x release line


reta:
### Is your feature request related to a problem? Please describe

Jackson has released major new 3.x version [1], this is major breaking change [2] comparing to 2.x release line, including both package and artifact changes.

### Describe the solution you'd like

Consider switching over to Jackson 3.x

### Related component

Libraries

### Describe alternatives you've considered

Keep using Jackson 2.x

### Additional context

[1] https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0
[2] https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-1

### Plugins

- [x] https://github.com/opensearch-project/security/issues/6075
- [x] https://github.com/opensearch-project/ml-commons/issues/4783
- [ ] https://github.com/opensearch-project/anomaly-detection/issues/1701
- [ ] https://github.com/opensearch-project/flow-framework/issues/1361
- [ ] https://github.com/opensearch-project/neural-search/issues/1833
- [ ] https://github.com/opensearch-project/observability/issues/1986
- [ ] https://github.com/opensearch-project/notifications/issues/1164
- [ ] https://github.com/opensearch-project/reporting/issues/1175
- [ ] https://github.com/opensearch-project/search-relevance/issues/448
- [x] https://github.com/opensearch-project/sql/issues/5342
- [ ] https://github.com/opensearch-project/user-behavior-insights/issues/173

andrross:
> 3.0 is a major version update and thereby not API compatible with 2.x; however, since 2.10 new methods have been added in 2.x line to reduce incompatibilities.

Any idea if we're still using a lot of deprecated APIs? If so, it would probably be good to chip away at that with the 2.x version to make the switch to 3.0 a bit easier.

reta:
> Any idea if we're still using a lot of deprecated APIs? If so, it would probably be good to chip away at that with the 2.x version to make the switch to 3.0 a bit easier.

@andrross we use quite a number of deprecated APIs, it is certainly good idea to clean those one up first, thank you.

PR #21029, opened 2026-03-28
Support Jackson 3.x release line

reta:
### Description
Support Jackson 3.x release line. The change does a few things:
 - the OpenSearch Core switches to Jackson 3.x release line at once (x-content, etc..)
 - the OpenSearch `server` module will continue to bundle Jackson 2.x artifacts (along with Jackson 3.x) to not break existing plugins
 - once all plugins switch to Jackson 3.x, the `server` could drop Jackson 2.x artifacts  bundling

### Related Issues
Closes https://github.com/opensearch-project/OpenSearch/issues/19597

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/storage/action/tiering/CancelTieringAction.java
+++ b/server/src/main/java/org/opensearch/storage/action/tiering/CancelTieringAction.java
@@ -1,0 +1,29 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.storage.action.tiering;
+
+import org.opensearch.action.ActionType;
+import org.opensearch.action.support.clustermanager.AcknowledgedResponse;
+
+/**
+ * Action for cancelling ongoing tiering operations.
+ * This action can be used to cancel both hot-to-warm and warm-to-hot migrations
+ * when shards get stuck in RUNNING_SHARD_RELOCATION state.
+ */
+public class CancelTieringAction extends ActionType<AcknowledgedResponse> {
+
+    /** Singleton instance. */
+    public static final CancelTieringAction INSTANCE = new CancelTieringAction();
+    /** Action name for cancel tiering. */
+    public static final String NAME = "indices:admin/_tier/cancel";
+
+    private CancelTieringAction() {
+        super(NAME, AcknowledgedResponse::new);
+    }
+}
--- a/server/src/main/java/org/opensearch/storage/action/tiering/CancelTieringRequest.java
+++ b/server/src/main/java/org/opensearch/storage/action/tiering/CancelTieringRequest.java
@@ -1,0 +1,70 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.storage.action.tiering;
+
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.action.support.clustermanager.AcknowledgedRequest;
+import org.opensearch.core.common.io.stream.StreamInput;
+
+import java.io.IOException;
+
+/**
+ * Request to cancel an ongoing tiering operation for an index.
+ * This request contains the index name and necessary parameters to safely
+ * cancel migrations that may be stuck in RUNNING_SHARD_RELOCATION state.
+ *
+ * Serialization (writeTo, StreamInput constructor), validation, equals/hashCode, and toString
+ * will be added in the implementation PR.
+ */
+public class CancelTieringRequest extends AcknowledgedRequest<CancelTieringRequest> {
+
+    private String index;
+
+    /** Default constructor. */
+    public CancelTieringRequest() {
+        super();
+    }
+
+    /**
+     * Constructs a request from a stream.
+     * @param in the stream input
+     * @throws IOException if an I/O error occurs
+     */
+    public CancelTieringRequest(StreamInput in) throws IOException {
+        super(in);
+        throw new UnsupportedOperationException("Not yet implemented");
+    }
+
+    /**
+     * Constructs a request for the given index.
+     * @param index the index name
+     */
+    public CancelTieringRequest(final String index) {
+        super();
+        this.index = index;
+    }
+
+    @Override
+    public ActionRequestValidationException validate() {
+        throw new UnsupportedOperationException("Not yet implemented");
+    }
+
+    /** Returns the index name. */
+    public String getIndex() {
+        return index;
+    }
+
+    /**
+     * Sets the index name.
+     * @param index the index name
+     */
+    public void setIndex(String index) {
+        this.index = index;
+    }
+}
--- a/server/src/main/java/org/opensearch/storage/action/tiering/HotToWarmTierAction.java
+++ b/server/src/main/java/org/opensearch/storage/action/tiering/HotToWarmTierAction.java
@@ -1,0 +1,31 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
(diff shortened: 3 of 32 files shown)


libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java
```diff
@@ -229,12 +250,55 @@ public XContentBuilder(XContent xContent, OutputStream bos, Set<String> includes
...
+        this.prettyPrint = prettyPrint;
+    }
+
+    /**
+     * Since 3.x release line, Jackson does not allow on the fly changes to the generator
+     * (like changing pretty printer, etc). To workaround and preserve the APIs, we do defer
+     * the generator instantiation till the first usage.
+     */
+    private XContentGenerator generatorInstance() throws IOException {
+        if (generator == null) {
```
reta, line 290: @andrross @cwperks this is probably the most impactful change in the whole pull request, making `generator` initialization lazy. I initially overthought it and used `volatile` + double check locking but looking closely, I have not found the evidence `XContentBuilder` instances are shared across threads (quite expected), simple `null` check should be sufficient.

libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java
```diff
@@ -245,12 +309,16 @@ public OutputStream getOutputStream() {
     }
 
     public XContentBuilder prettyPrint() {
-        generator.usePrettyPrint();
+        this.prettyPrint = true;
```
andrross, line 316: So calling this method (and some others) will have no effect if generatorInstance() has already been invoked on this instance? Is this behavior change going to be impactful? Should these method throw if the generator instance has already been constructed?

reta: Thank you @andrross 

> So calling this method (and some others) will have no effect if generatorInstance() has already been invoked on this instance? 

Correct, but the usage pattern is (all cases in the codebase): `builder = new XContentBuilder().prettyPrint()` (basically `prettyPrint` is always called right after / close to construction site).

> Should these method throw if the generator instance has already been constructed?

I think we shouldn't, the output will be always correct, the formatting in this case looks like minor inconvenience, not an issue at large.

andrross: > I think we shouldn't, the output will be always correct

I'm still on the fence about this. "correct" is in the eye of the caller :) It's certainly _possible_ to imagine use cases that would break if given non-pretty-printed JSON.

libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java
```diff
@@ -245,12 +309,16 @@ public OutputStream getOutputStream() {
...
         return this;
     }
 
     public boolean isPrettyPrint() {
-        return generator.isPrettyPrint();
+        try {
+            return generatorInstance().isPrettyPrint();
```
andrross: Should this always create the generator? In theory if the generator is null at this point you could return the boolean. In fact, if you prevent mutation of the `prettyPrint` field after generator creation then you could always return the field here as its guaranteed to match the generator state.

libs/x-content/src/main/java/org/opensearch/common/xcontent/XObjectWriteContext.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public PrettyPrinter getPrettyPrinter() {
+        return prettyPrinter;
+    }
+
+    @Override
+    public void writeValue(JsonGenerator g, Object value) {
+        // 31-Dec-2009, tatu: Actually, we could just handle some basic
```
andrross, line 53: Looks like this is a copy from Jackson 2.x. Should this be attributed somehow?

libs/x-content/src/main/java/org/opensearch/common/xcontent/XObjectWriteContext.java
```diff
@@ -0,0 +1,101 @@
...
+import tools.jackson.core.util.DefaultPrettyPrinter;
+
+public class XObjectWriteContext extends ObjectWriteContext.Base {
+    private static final SerializedString LF = new SerializedString("\n");
+    private static final DefaultPrettyPrinter.Indenter INDENTER = new DefaultIndenter("  ", LF.getValue());
+    private final PrettyPrinter prettyPrinter;
+
+    public XObjectWriteContext(boolean prettyPrint) {
+        this.prettyPrinter = prettyPrint ? new DefaultPrettyPrinter().withObjectIndenter(INDENTER).withArrayIndenter(INDENTER) : null;
+    }
```
andrross: Bit of a nitpick, but this class looks stateless in the non-pretty print case. Could you do something like the following to avoid creating an instance in what is presumably the normal hot path?

```java
public static XObjectWriteContext create(boolean prettyPrint) {
    if (prettyPrint) {
        return new XObjectWriteContext...
    }
        return DEFAULT_INSTANCE;
    }
}

File: libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java (hunk 2 above)
Comment (reta) [reply libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java:316]: Valid concerns, throwing exception now, thanks!

File: libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java (hunk 3 above)
Comment (reta) [reply libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java]: Always returning `this.prettyPrint;` since it cannot be changed in the generator anymore, good catch!

File: libs/core/src/main/java/org/opensearch/tools/jackson/core/JacksonExceptionTranslator.java [hunk 6]
```diff
@@ -0,0 +1,53 @@
...
+ */
+public final class JacksonExceptionTranslator {
+    private JacksonExceptionTranslator() {}
+
+    /**
+     * Translates Jackson 3 exception hierarchy to Jackson 2 compatible one
+     * @param ex exception to translate
+     * @return translated exception
+     */
+    public static IOException translateOrRethrowReturning(Exception ex) {
```
cwperks: nit: Should we scope the input here to JacksonException and possible rename to translateToIOException?

reta: Good suggestion, thanks @cwperks !

reta:
@andrross @cwperks comments addressed, thanks a lot for the review folks!

rest-api-spec/src/main/resources/rest-api-spec/test/search/30_limits.yml
```diff
@@ -62,6 +65,27 @@ setup:
...
+---
+"Request with size exceeding max integer value (after 3.7.0)":
+  - skip:
+      version: "- 3.6.99"
+      reason: "the error message changes in 3.7.0"
+      features: node_selector
+  - do:
+      node_selector:
+        version: "3.7.0 - "
+      catch:      /Numeric value \(2147483648\) out of range of `int`/
```
andrross: Are the backticks the only difference here? Can we change the assertion to be a regex or something that handles both?

andrross (commented):
What's the risk of a change in Jackson behavior causing a breaking change in the REST API? I think exception names show up in the API, like `json_e_o_f_exception` becoming `unexpected_end_of_input_exception`. Are there more significant differences?

reta:
> What's the risk of a change in Jackson behavior causing a breaking change in the REST API? I think exception names show up in the API, like `json_e_o_f_exception` becoming `unexpected_end_of_input_exception`. Are there more significant differences?

With the exception of error messages / class abbreviations, nothing API (user) facing I am aware of to be honest, the test cases that exercise the APIs are not being affected.

rest-api-spec/src/main/resources/rest-api-spec/test/search/30_limits.yml
reta: > Are the backticks the only difference here? Can we change the assertion to be a regex or something that handles both?

Was  not sure `catch` supports regex, but it seems it does (regular Java pattern), giving it a shot

reta:
@andrross any concerns from your side? thanks!

andrross (approved):

Issue #21012, opened 2026-03-27
Use Advise by Context to ensure IOContext is honored during file read in MMapDirectory


navneet1v:
## Description
Creating this issue to start a discussion on moving to IOContext based read advise for MMap directory as per changes in Lucene: https://github.com/apache/lucene/pull/15040 (version 10.3.1).

### Background/Trail of Changes
Till version 9.12 (OS version 2.19) when MemorySegmentIndexInput was enabled in MMapDirectory it used to honor the read advise. Ref: https://github.com/apache/lucene/blob/b8bfffa36810feea768860be882b278596d0c974/lucene/core/src/java21/org/apache/lucene/store/MemorySegmentIndexInputProvider.java#L45-L86

Lucene Version 10.3.0 : In this change https://github.com/apache/lucene/pull/15040, read advise was no longer getting honored and was falling back to Random, which was then changed to normal with this PR: https://github.com/apache/lucene/pull/15040 . Everything happened in 10.3.0 version.

To go back to old behavior of **honoring the read advise** we have to set 
```
dir.setReadAdvice(MMapDirectory.ADVISE_BY_CONTEXT)
```
which will ensure that read advise is overridden by IOContext. Ref: https://github.com/apache/lucene/blob/9eaf64a84693b58468b48fb9a2671df95e133303/lucene/core/src/java/org/apache/lucene/store/MMapDirectory.java#L157-L170

### Current State (from 3.3 and above)
In current state, what I can see is read advise present in the IOContext is no longer getting honored and all the files are opened and read using NORMAL read advise. Even when we are trying to change the read advise during merge update to read advise is not happening.

### Problem with Current Situation:
Since madvise is always normal, so even when we are trying to update the IOContext for a file we are seeing no impact. Example for vector search we want to keep `.vec` file containing vectors as random access, since reads are random. `.fdt` file has random access too.
But in all these cases we are still doing read advise as normal. 

Hence starting this thread to see what maintainers of OpenSearch think about moving to ADVISE_BY_CONTEXT behavior.

Tagging @msfroh , @andrross , @mch2 , @gbbafna , @shwetathareja

gbbafna:
Thank @navneet1v . I am aligned with your proposal .

navneet1v:
Thanks @gbbafna I can start putting a PR. If folks have other ideas/suggestions please put your thoughts.

navneet1v:
Raised a PR : https://github.com/opensearch-project/OpenSearch/pull/21031

PR #21031, opened 2026-03-28
Updated MMapDirectory to use ReadAdviseByContext rather than default readadvise of Lucene

navneet1v:
### Description
Updated MMapDirectory to use ReadAdviseByContext rather than default readadvise of Lucene. 

More details can be found here: https://github.com/opensearch-project/OpenSearch/issues/21012

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/21012

### Check List
- [ ] ~API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.~
- [ ] ~Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.~

### DIFF
--- a/server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
+++ b/server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
@@ -31,14 +31,18 @@
 
 package org.opensearch.index.store;
 
+import org.apache.lucene.store.DataAccessHint;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.IOContext;
 import org.apache.lucene.store.MMapDirectory;
+import org.apache.lucene.store.MergeInfo;
 import org.apache.lucene.store.NIOFSDirectory;
 import org.apache.lucene.store.NoLockFactory;
+import org.apache.lucene.store.ReadAdvice;
 import org.apache.lucene.util.Constants;
 import org.opensearch.Version;
 import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.common.SuppressForbidden;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.core.index.Index;
 import org.opensearch.core.index.shard.ShardId;
@@ -49,11 +53,14 @@ import org.opensearch.test.IndexSettingsModule;
 import org.opensearch.test.OpenSearchTestCase;
 
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
+import java.util.function.BiFunction;
 import java.util.function.BiPredicate;
 
 import static org.opensearch.test.store.MockFSDirectoryFactory.FILE_SYSTEM_BASED_STORE_TYPES;
@@ -173,6 +180,7 @@ public class FsDirectoryFactoryTests extends OpenSearchTestCase {
             switch (type) {
                 case HYBRIDFS:
                     assertTrue(FsDirectoryFactory.isHybridFs(directory));
+                    mMapDirectoryHasReadAdviceByContext(((FsDirectoryFactory.HybridDirectory) directory).getDelegate());
                     break;
                 // simplefs was removed in Lucene 9; support for enum is maintained for bwc
                 case SIMPLEFS:
@@ -181,6 +189,7 @@ public class FsDirectoryFactoryTests extends OpenSearchTestCase {
                     break;
                 case MMAPFS:
                     assertTrue(type + " " + directory.toString(), directory instanceof MMapDirectory);
+                    mMapDirectoryHasReadAdviceByContext((MMapDirectory) directory);
                     break;
                 case FS:
                     if (Constants.JRE_IS_64BIT) {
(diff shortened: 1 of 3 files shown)


navneet1v:
Tagging @andrross , @msfroh , @gbbafna , @mch2 to review the PR.

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
```diff
@@ -194,4 +204,52 @@ private void doTestStoreDirectory(Path tempDir, String typeSettingValue, IndexMo
...
+            );
+
+            assertEquals(
+                "Advise By context is not be set",
+                MMapDirectory.ADVISE_BY_CONTEXT.apply("test.vec", IOContext.DEFAULT.withHints(DataAccessHint.RANDOM)),
+                readAdvice.apply("test.vec", IOContext.DEFAULT.withHints(DataAccessHint.RANDOM))
+            );
+
+        } catch (Exception e) {
+            fail("Failed to verify read advice is set to ADVISE_BY_CONTEXT: " + e.getMessage());
```
andrross: Just let anything here throw and fail the test. Add `throws Exception` if necessary. This catch block will hide the stack trace.

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
```diff
@@ -194,4 +204,52 @@ private void doTestStoreDirectory(Path tempDir, String typeSettingValue, IndexMo
...
+            @SuppressWarnings("unchecked")
+            BiFunction<String, IOContext, Optional<ReadAdvice>> readAdvice = (BiFunction<
+                String,
+                IOContext,
+                Optional<ReadAdvice>>) getReadAdviceField(mapDirectory);
+
+            // Verify the function behaves identically to ADVISE_BY_CONTEXT
+            // ADVISE_BY_CONTEXT returns the ReadAdvice from the IOContext
+            assertEquals(
+                "Advise By context is not be set",
```
andrross: Typo here and every other assertion in this method

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
```diff
@@ -194,4 +204,52 @@ private void doTestStoreDirectory(Path tempDir, String typeSettingValue, IndexMo
...
+
+        } catch (Exception e) {
+            fail("Failed to verify read advice is set to ADVISE_BY_CONTEXT: " + e.getMessage());
+        }
+
+    }
+
+    @SuppressForbidden(reason = "need reflection to access private readAdvice field for testing")
+    private Object getReadAdviceField(MMapDirectory mMapDirectory) throws Exception {
+        Field readAdviceField = MMapDirectory.class.getDeclaredField("readAdvice");
```
andrross, line 249: This is fragile. Can you wrap an MMapDirectory and intercept `setReadAdvice()` to get this information?

navneet1v: make sense. Let me add a directory for tests.

navneet1v:
> This looks reasonable, but are we concerned at all about potential regressions with using `MADV_RANDOM` for certain files if MGLRU is enabled? That was the whole reason Lucene reverted its change to the default behavior.

By not having this change, we are seeing the regression during search. One thing I can think of here is, if MGLRU is enabled on the Kernel we fallback to NORMAL readadvise, otherwise we can use ADVISE by CONTEXT.

@andrross WDYT?

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
navneet1v: @andrross what typo you are considering here? the idea here is if advise by context is not set we want show the failure that advise by context is not set and fail the test

andrross:
> One thing I can think of here is, if MGLRU is enabled on the Kernel we fallback to NORMAL readadvise, otherwise we can use ADVISE by CONTEXT.

@navneet1v I honestly don't know. Do we have benchmarks that can tell us the optimal configuration with and without MGLRU?

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
navneet1v: @andrross I don't think there is a way to go around this only for tests. We would either need a change in Lucene to expose the readAdvise or we need create another directory in OpenSearch say OpenSearchMMapDirectory which can expose the readAdvise. I am bit reluctant to have a `OpenSearchMMapDirectory` as of now. What other possible approaches you can suggest here?

navneet1v:
> > One thing I can think of here is, if MGLRU is enabled on the Kernel we fallback to NORMAL readadvise, otherwise we can use ADVISE by CONTEXT.
> 
> @navneet1v I honestly don't know. Do we have benchmarks that can tell us the optimal configuration with and without MGLRU?



> > One thing I can think of here is, if MGLRU is enabled on the Kernel we fallback to NORMAL readadvise, otherwise we can use ADVISE by CONTEXT.
> 
> @navneet1v I honestly don't know. Do we have benchmarks that can tell us the optimal configuration with and without MGLRU?

@andrross in 2.19 version of OpenSearch we are using Advise by Context. This change is just making those things exactly same. Plus the regression which was observed in Elastic was during merge and during merge we are already making the read advise sequential

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
andrross: Ah, yeah, there may not be a better way to do this.

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
andrross: So should it say "Advise by context is not set"?

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
```diff
@@ -194,4 +204,52 @@ private void doTestStoreDirectory(Path tempDir, String typeSettingValue, IndexMo
...
+            assertEquals(
+                "Advise By context is not be set",
+                MMapDirectory.ADVISE_BY_CONTEXT.apply("test.tim", IOContext.READONCE),
+                readAdvice.apply("test.tim", IOContext.READONCE)
+            );
+            MergeInfo mergeInfo = new MergeInfo(100, 100L, false, 1);
+            assertEquals(
+                "Advise By context is not be set",
+                MMapDirectory.ADVISE_BY_CONTEXT.apply("test.vec", IOContext.merge(mergeInfo).withHints(FileDataHint.KNN_VECTORS)),
+                readAdvice.apply("test.vec", IOContext.merge(mergeInfo).withHints())
```
andrross: This looks like a bug. Should `FileDataHint.KNN_VECTORS` be set on both IOContexts here?

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
navneet1v: I created a PR in lucene: https://github.com/apache/lucene/pull/15892 to expose this parameter. Lets see what happens there. But as of now there is no way and we had to go via Reflection for tests.

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
navneet1v: actually it doesn't matter. Since readadvise by context doesn't look at the `FileDataHint`, it exists early as merge was added. I will remove extra thing

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
navneet1v: Right. when you said typo I wasn't sure what does it mean. I have fixed the message now.

server/src/test/java/org/opensearch/index/store/FsDirectoryFactoryTests.java
navneet1v: done

navneet1v:
@andrross can we trigger a benchmark run on this PR. In the meanwhile I doing more deep-dive on the issue provided by Lucene to see where and how ES faced issues. Because based on my deep-dive we change the IOContext during merge from RANDOM to SEQUENTIAL so we should face this issue in vectors.

Ref: https://github.com/apache/lucene/blob/726731f602a1b37430cd26995540c8a5d3735708/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java#L196-L201

navneet1v:
@andrross here is what I am able to trace out.

The issue that happened in Elastic Search was with Lucene version 9.12, with Kernel enabled MGLRU. Elastic was always using Random Read advise for their vector files. Ref: https://github.com/elastic/elasticsearch/issues/124499


But with OpenSearch we should not face this issue because:
1. When Opensearch is doing merges in lucene 10.x for vectors we switch the read advise to [sequential](https://github.com/apache/lucene/blob/branch_10_4/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java#L203-L207), same capability was not present in [Lucene 9.12](https://github.com/apache/lucene/blob/branch_9_12/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsReader.java). 

I can try reproducing the same setup with OpenSearch 2.19(using Lucene 9.12) and OpenSearch 3.x where we override the read advice to showcase that we will not see regression when MGLRU is enabled, is the above pointers not enough. Please let me know your thoughts.

navneet1v:
> Looks good.
> 
> Curious where's the search regression coming from? Is it somewhere you define IOContext with DataAccessHint.RANDOM, but without this change it got ignored?

Yes, so vectors are stored in `.vec` file which gets opened with IOContext as Random. But without this change that IOContext was getting ignored, which was causing higher page faults, during search.

PR #21021, opened 2026-03-27
[draft] Using ordinals in multi terms instead of byte values for bucket keys

sandeshkr419:
### Description
Testing out if using ordinals instead of byte values for multi-bucket keys help in speeding up multi-terms aggregations.

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21033, opened 2026-03-29
[perf] Using ordinals in multi term aggs instead of byte values for bucket keys

### LINKED ISSUE #21051: Optimizing Multi-terms over keyword fields using global ordinals

The `MultiTermsAggregator` currently serializes composite bucket keys as variable-length `BytesRef` arrays via `BytesStreamOutput`, storing them in a `BytesRefHash` through `BytesKeyedBucketOrds`. Each unique composite key incurs a full byte-array copy in the hash, plus per-document serialization/deserialization overhead. For high-cardinality multi-terms aggregations, this creates significant JVM memory pressure.

This optimization replaces the serialized `BytesRef` composite keys with packed ordinal tuples when all fields in the aggregation support global ordinals (`ValuesSource.Bytes.WithOrdinals`). Instead of storing `"electronics|laptop|premium"` as ~30+ bytes, we store three `long` ordinals packed into a single `long` (for 2 fields) or a compact `BytesRef` of fixed-width ordinals (for N fields). This eliminates the variable-length byte serialization entirely and reduces per-bucket memory from O(key_length) to O(N * 8) bytes where N is the number of fields.

For mixed-type configurations (keyword + numeric fields), the optimization falls back to the existing `BytesRef`-based approach, ensuring full backward compatibility.

The idea is to replace the serialized `BytesRef` composite keys with packed ordinal tuples when all fields in the aggregation support global ordinals (`ValuesSource.Bytes.WithOrdinals`). Instead of storing `"electronics|laptop|premium"` as ~30+ bytes, we store three `long` ordinals packed into a compact `BytesRef` of fixed-width ordinals (for N fields). This eliminates the variable-length byte serialization entirely and reduces per-bucket memory from O(key_length) to O(N * 8) bytes where N is the number of fields.

POC: https://github.com/opensearch-project/OpenSearch/pull/21032
This was the first iteration of the planned improvement which resulted in a massive gain.

| Metric | Operation | Baseline | Contender | Diff | Unit |
| :--- | :--- | :--- | :--- | :--- | :--- |
| Min Throughput | multi_terms-keyword | 1.30403 | 2.00055 | 0.69653 | ops/s |
| Mean Throughput | multi_terms-keyword | 1.32316 | 2.00091 | 0.67775 | ops/s |
| Median Throughput | multi_terms-keyword | 1.32546 | 2.00083 | 0.67536 | ops/s |
| Max Throughput | multi_terms-keyword | 1.33388 | 2.00161 | 0.66773 | ops/s |
| 50th percentile latency | multi_terms-keyword | 25248.1 | 171.678 | -25076.4 | ms |
| 90th percentile latency | multi_terms-keyword | 34857.1 | 176.62 | -34680.5 | ms |
| 99th percentile latency | multi_terms-keyword | 37025.8 | 212.795 | -36813 | ms |
| 100th percentile latency | multi_terms-keyword | 37146.3 | 232.514 | -36913.7 | ms |
| 50th percentile service time | multi_terms-keyword | 734.633 | 170.571 | -564.062 | ms |
| 90th percentile service time | multi_terms-keyword | 752.307 | 175.58 | -576.727 | ms |
| 99th percentile service time | multi_terms-keyword | 829.96 | 210.92 | -619.039 | ms |
| 100th percentile service time | multi_terms-keyword | 848.685 | 231.12 | -617.564 | ms |
| error rate | multi_terms-keyword | 0 | 0 | 0 | % |

### Related component

Search:Aggregations

**_Optimization 1:_ 2 Keyword packed into a single long**

Packing 2 keyword ordinals into a single long as 126 bits could work perfectly fine for upto a few million cardinality for 2 keywords. This avoid converting to byteref completely as the single long can function as the composite key required to maintain the buckets.

POC: https://github.com/opensearch-project/OpenSearch/pull/21021

**_Optimization 2:_ N Keyword packed into a single long, or 2 longs if required**

Packing multiple buckets into a single long, and extend into a second long if needed. This is inspired from LongLongHash approach but designed for multiple keywords in a multi-aggregation request.

POC: https://github.com/opensearch-project/OpenSearch/pull/21033

----

| Metric | Baseline | Generic N-long optimization | 2 keyword - single long | multiple fields, 1 long (2 if it overflows 1 long) | Unit |
| :--- | :--- | :--- | :--- | :--- | :--- |
| Min Throughput | 1.30403 | 2.00055 | 2.00212 | 2.00086 | ops/s |
| Mean Throughput | 1.32316 | 2.00091 | 2.00346 | 2.00143 | ops/s |
| Median Throughput | 1.32546 | 2.00083 | 2.00313 | 2.00132 | ops/s |
| Max Throughput | 1.33388 | 2.00161 | 2.00615 | 2.00258 | ops/s |
| 50th percentile latency | 25248.1 | 171.678 | 122.332 | 127.833 | ms |
| 90th percentile latency | 34857.1 | 176.62 | 127.911 | 132.191 | ms |
| 99th percentile latency | 37025.8 | 212.795 | 186.781 | 200.869 | ms |
| 100th percentile latency | 37146.3 | 232.514 | 186.826 | 209.26 | ms |
| 50th percentile service time | 734.633 | 170.571 | 121.029 | 126.69 | ms |
| 90th percentile service time | 752.307 | 175.58 | 126.345 | 131.003 | ms |
| 99th percentile service time | 829.96 | 210.92 | 185.912 | 199.778 | ms |
| 100th percentile service time | 848.685 | 231.12 | 186.051 | 208.258 | ms |
| error rate | 0 | 0 | 0 | 0 | % |

### Additional context

While the numbers from 3 approaches is fairly similar as evident from multiple runs you can find on the PRs itself from benchmarks, I prefer the last approach as it can handle cases for very high cardinality easy in 1 (at max 2) longs.

Basically 1M cardinality field will require 20 bits to pack [2^20 ~ 1M], a single long (126 bits) packing will suffice in most observed cases.

While a single long approach for specifically 2 keywords is great and should ideally cover majority cases, the 1/2 long approach for N keywrods offer a more optimized approach for more than 2 fields as well keeping the memory imprint in a predictable manner.

sandeshkr419:
Signed-off-by: Sandesh Kumar <sandeshkr419@gmail.com>

## Description

  `multi_terms` aggregations over keyword fields today serialize composite bucket keys as concatenated strings into a `BytesRefHash`. This causes per-document allocation, GC pressure, and latency in the seconds range — and outright OOM failures for high-cardinality field combinations.

  This PR replaces that with **packed global ordinals**. Lucene already maintains a compact integer (ordinal) per unique term per field. We bit-pack the ordinals for all fields into one or two `long` values and store them in a primitive hash, eliminating byte-array serialization on the hot collection path entirely.

  **Strategy** (`MultiTermsAggregationFactory.selectOrdinalStrategy`):
  - Requires all fields to be `ValuesSource.Bytes.WithOrdinals` (keyword, IP, boolean)
  - Disabled when a star-tree index is active on the query
  - `totalBits ≤ 63` → single `long` via `LongKeyedBucketOrds`
  - `63 < totalBits ≤ 126`, top-level agg only → two `long`s via `LongLongHash`
  - Otherwise → original `BytesRef` key path (unchanged)

  **New classes:**
  - `MultiTermsBucketOrds` — interface for ordinal-keyed bucket storage
  - `PackedOrdinalBucketOrds` — bit-packs ordinals into 1 or 2 longs

  **Aggregator changes** (`MultiTermsAggregator`):
  - Collection reads ordinals from `SortedSetDocValues`, expands multi-valued
    fields into ordinal tuples via cartesian product (scratch buffers, no per-doc
    allocation)
  - Result phase resolves ordinals back to term values via `lookupOrd`
  - `min_doc_count=0` supported via a dedicated zero-doc ordinal fill path
  - Existing bytes-key path unchanged; both paths coexist

  ## Performance

  ClickBench (99.9M docs, 1 shard, r8g.2xlarge):

  | Query | Fields | Baseline | Optimized | Improvement |
  |-------|--------|----------|-----------|-------------|
  | 2-field | `BrowserCountry × BrowserLanguage` | 6,229 ms | 1,413 ms | −77% |
  | 2-field | `BrowserCountry × UTMSource` | 6,104 ms | 1,569 ms | −74% |
  | 3-field | `HitColor × PageCharset × UTMSource` | 5,196 ms | 2,216 ms | −57% |
  | 4-field | `HitColor × BrCountry × BrLanguage × UTMSource` | 5,868 ms | 2,622 ms | −55% |
  | 4-field high-cardinality | `URL × Referer × SearchPhrase × OriginalURL` | OOM | ~240 s | OOM → functional |

  Official OpenSearch Benchmark (Big5) - check comments for multiple runs:

  | Metric | Baseline | Optimized | Improvement |
  |--------|----------|-----------|-------------|
  | p50 latency | 25,248 ms | 122 ms | −99.5% |
  | p50 service time | 735 ms | 121 ms | −83.5% |

  No regression on other agg shapes.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/21051, improves upon naive implementation: https://github.com/opensearch-project/OpenSearch/pull/21021

### Check List


sandeshkr419:

sandeshkr419:

sandeshkr419:

sandeshkr419:

sandeshkr419:

sandeshkr419:

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregationFactory.java
```diff
@@ -142,11 +143,13 @@ protected Aggregator createInternal(
         }
         // TODO: Optimize passing too many value source config derived objects to aggregator
         bucketCountThresholds.ensureValidity();
+        List<ValuesSource> rawValuesSources = configs.stream().map(config -> config.v1().getValuesSource()).toList();
+        MultiTermsBucketOrds ordinalBucketOrds = selectOrdinalStrategy(rawValuesSources, searchContext, cardinality);
```
asimmahmood1: `selectOrdinalStrategy` only receives the raw `ValuesSource` list, so it cannot see each field's `IncludeExclude` configuration. When all fields support ordinals, `getOrdinalLeafCollector` bypasses the legacy `InternalValuesSource` collectors where these filters are applied and collects every ordinal instead. As a result, an all-keyword `multi_terms` request with `include`/`exclude` can return filtered-out tuples. The simplest safe fix is to fall back when any `config.v2()` is non-null (or explicitly apply equivalent ordinal filters), with an all-keyword regression test.

server/src/test/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsOrdinalOptimizationTests.java
```diff
@@ -0,0 +1,734 @@
...
+            iw.addDocument(doc);
+            iw.close();
+
+            try (DirectoryReader unwrapped = DirectoryReader.open(directory); IndexReader reader = wrapDirectoryReader(unwrapped)) {
+                IndexSearcher searcher = newIndexSearcher(reader);
+                MultiTermsAggregationBuilder builder = new MultiTermsAggregationBuilder("test_agg").terms(
+                    asList(kwConfig(KW_FIELD_1), kwConfig(KW_FIELD_2))
+                );
+                InternalMultiTerms result = searchAndReduce(searcher, new MatchAllDocsQuery(), builder, allFieldTypes());
+                // Verify the ordinal path produced correct results (1 bucket with count 1)
```
asimmahmood1, line 448: These strategy-selection tests only assert final bucket contents, so they would still pass if `selectOrdinalStrategy` always returned `null`. The new `getOrdinalBucketOrds()` accessor is not used. Please add a factory/aggregator-level assertion for the selected implementation and fallback. This should also cover `MANY`: tuples fitting in one long still use `LongKeyedBucketOrds`; only the two-long path falls back, so `testNestedMultiTermsFallsBackToByteKeyPath` currently has an incorrect premise.

server/src/test/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsOrdinalOptimizationTests.java
```diff
@@ -0,0 +1,734 @@
...
+    }
+
+    /**
+     * Test with min_doc_count=0 — exercises collectZeroDocOrdinals and the zero-doc fill path
+     * for the ordinal-based collection path.
+     */
+    public void testMinDocCountZeroOrdinalPath() throws IOException {
+        try (Directory directory = newDirectory()) {
+            RandomIndexWriter iw = new RandomIndexWriter(random(), directory);
+            // Only add docs for some combinations; absent combos should appear with count=0
```
asimmahmood1: This test never creates a zero-count bucket. No document contains `y`, so `y` has no global ordinal, and `MatchAllDocsQuery` gives every indexed combination a positive count. The later `>= 0` assertion therefore does not prove that `collectZeroDocOrdinals` added anything. Please index a combination such as `alpha|y`, exclude that document with the query, and explicitly assert that the resulting bucket exists with count `0`.

asimmahmood1 (commented):
I found one correctness issue and two test-coverage gaps in the packed-ordinal path. The focused tests and formatting checks pass, but these cases should be addressed before merge.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregationFactory.java
```diff
@@ -158,14 +161,49 @@ protected Aggregator createInternal(
...
+        for (ValuesSource vs : rawValuesSources) {
+            if ((vs instanceof ValuesSource.Bytes.WithOrdinals) == false) {
+                return null;
+            }
+        }
+        int numFields = rawValuesSources.size();
+        long[] maxOrds = new long[numFields];
+        for (int i = 0; i < numFields; i++) {
+            maxOrds[i] = ((ValuesSource.Bytes.WithOrdinals) rawValuesSources.get(i)).globalMaxOrd(searchContext.searcher());
+        }
```
jainankitk: Do we need to check the existence of global ordinals for each field?

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/PackedOrdinalBucketOrds.java
```diff
@@ -0,0 +1,244 @@
...
+
+    @Override
+    public long add(long owningBucketOrd, long[] ordinals) {
+        assert ordinals.length == numFields;
+        if (singleLongDelegate != null) {
+            return singleLongDelegate.add(owningBucketOrd, packSingleLong(ordinals));
+        } else {
+            assert owningBucketOrd == 0 : "two-long path only supports single owning bucket";
+            return twoLongDelegate.add(packLong1(ordinals), packLong2(ordinals));
+        }
```
jainankitk, line 197: Why does the "two-long path only supports single owning bucket" ?

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/PackedOrdinalBucketOrds.java
```diff
@@ -0,0 +1,244 @@
...
+                int bitsInLong1 = splitOffset - bitOffsets[i];
+                packed |= (ordinals[i] >>> bitsInLong1);
+            }
+            // else: field is entirely in long1
+        }
+        return packed;
+    }
+
+    long[] unpackTwoLongs(long long1, long long2) {
+        long[] ordinals = new long[numFields];
```
jainankitk, line 171: I am wondering if `ordinals` should be passed as buffer by the caller to ensure reuse across documents?

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregator.java
```diff
@@ -140,45 +154,33 @@ public InternalAggregation[] buildAggregations(long[] owningBucketOrds) throws I
...
         long[] otherDocCounts = new long[owningBucketOrds.length];
+
+        SortedSetDocValues[] globalOrdsForLookup = null;
+        if (ordinalBucketOrds != null) {
+            globalOrdsForLookup = new SortedSetDocValues[ordinalSources.size()];
+            List<LeafReaderContext> leaves = context.searcher().getTopReaderContext().leaves();
+            if (leaves.isEmpty() == false) {
+                LeafReaderContext leafCtx = leaves.get(0);
+                for (int i = 0; i < ordinalSources.size(); i++) {
+                    globalOrdsForLookup[i] = ordinalSources.get(i).globalOrdinalsValues(leafCtx);
```
jainankitk, line 165: Could the `globalOrdinalsValues` be uninitialized?

jainankitk (commented):
Thanks @sandeshkr419 for making this change. Performance optimization looks really promising. Minor comments to understand the change better

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/PackedOrdinalBucketOrds.java
sandeshkr419: Unpack runs once per result bucket, not per doc. 
The per-doc hot path already allocates nothing, so not expecting measurable gain.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregator.java
sandeshkr419: No: it's only set when leaves exist, and buildOrdinalBuckets early-returns on the null guard otherwise.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregationFactory.java
sandeshkr419: Fixed.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/PackedOrdinalBucketOrds.java
sandeshkr419: The single-long path uses LongKeyedBucketOrds, which lets you file each key under a parent bucket, so nesting works. 

The two-long path uses LongLongHash (a plain hash of the (long1, long2) pair with no "which parent" slot) so everything would collapse into one shared table. That's only safe when there's exactly one parent: a top-level multi_terms. That's the cardinality == ONE case; nested ones that need two longs fall back to the byte-key path. This optimization with so much high cardinality on nested aggs - seems safer to skip and fallback to legacy flow. :)

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregationFactory.java
sandeshkr419: Bytes.WithOrdinals fields always have a global-ordinal table (similar with keyword terms usage). 

Also, our chat about `eager_global_ordinals` - it only affects when that ordinals is built (refresh vs lazy first-query), not whether it exists - so we don't need special handling here.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregationFactory.java
```diff
@@ -158,14 +161,51 @@ protected Aggregator createInternal(
...
+            if (config.v2() != null) {
+                return null;
+            }
+            ValuesSource vs = config.v1().getValuesSource();
+            if ((vs instanceof ValuesSource.Bytes.WithOrdinals) == false) {
+                return null;
+            }
+            maxOrds[i] = ((ValuesSource.Bytes.WithOrdinals) vs).globalMaxOrd(searchContext.searcher());
+        }
+        return PackedOrdinalBucketOrds.create(searchContext.bigArrays(), cardinality, maxOrds);
```
jainankitk, line 206: `selectOrdinalStrategy()` allocates `PackedOrdinalBucketOrds` (BigArrays-backed) before `new MultiTermsAggregator(...)`. If the constructor throws (e.g. circuit breaker trips allocating `docCounts` in the super constructor), nothing calls `close()` on the already-allocated ords — leaking the memory and its circuit-breaker reservation. The byte path avoids this by building `bucketOrds` inside the constructor after `super()`.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregationFactory.java
jainankitk: On a many-segment shard where nothing else has forced global ordinals yet, `selectOrdinalStrategy` calling `globalMaxOrd` at aggregator setup becomes the thing that eagerly builds the `OrdinalMap` — and it does so even in cases where the code ultimately falls back to the byte path (which never needs it). Have we measured for scenarios where the global ordinals needs to be built synchronously during the `globalMaxOrd`?

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregationFactory.java
sandeshkr419: Thanks for catching this. Fixed in the latest commit.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/MultiTermsAggregationFactory.java
sandeshkr419: Yes — measured it with a cold-vs-warm delta on a dev box (99.9M docs, 37 segments, URL ~18M × Referer ~19.7M ords): the synchronous build for the worst-case pair (URL×Referer, ~18–20M ords over 37 segments) is **~17s** cold vs **~5ms** warm, i.e. the same OrdinalMap a keyword terms agg builds, cached per-shard after the first query.

It's only "wasted" in the fallback shape (paid then unused), but it's a one-time cached build amortized across later queries and removable via eager_global_ordinals, so it's worth paying — **it also unlocks the high-cardinality queries that OOM on baseline today.**

Global ords have to get built, whether its indexing refresh or at the cold search.

asimmahmood1 (approved):

jainankitk (approved):

PR #21039, opened 2026-03-30
Add DataFormatRegistry

mgodwan:
### Description

### Related Issues

### Check List


server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatRegistry.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * Returns all registered data formats that support a specific capability for a field type.
+     *
+     * @param fieldType the field type name
+     * @param capability the capability to check
+     * @return list of data formats supporting the capability for the field type
+     */
+    public List<DataFormat> supportsCapability(String fieldType, FieldTypeCapabilities.Capability capability) {
```
bharath-techie, line 127: Will fieldType eventually be an enum ?

mgodwan: Fields can come from plugins, so enum may not work

mgodwan: MappedFieldType is a representation but the capability check would be before creation of this data type. @darjisagar7 is looking into that flow, so once we close there, we can make this a more specific data type.

PR #21046, opened 2026-03-30
Fix flaky testOpenIndexOverLimit in ClusterShardLimitIT

gagandhakrey:
Use assertAcked() for the close index operation instead of manually checking isAcknowledged(), which could return false when nodes fail to acknowledge within the default timeout on slow environments (e.g., Linux/Java 25 CI runners).

Also scope the pre-close health check to the specific index, and fix a copy-paste bug in the final assertion that was checking for 'snapshot-index' (from testRestoreSnapshotOverLimit) instead of verifying that 'test-index-1' remains in the CLOSE state.

### Description

### Related Issues

### Check List


gagandhakrey:
Fix flaky testOpenIndexOverLimit in ClusterShardLimitIT

The testOpenIndexOverLimit test was intermittently failing in CI on Linux/Java 25 environments due to three issues. First, the close index operation used a manual isAcknowledged() check, which could return false when cluster nodes failed to acknowledge within the default timeout on slow or loaded CI runners — replaced with assertAcked(), which is the established pattern across the test suite and handles timeouts more robustly. Second, the pre-close health check was waiting on the entire cluster rather than scoping it to test-index-1, which is now corrected. Third, the final assertion was a copy-paste bug from testRestoreSnapshotOverLimit — it was checking that "snapshot-index" doesn't exist (a vacuously true assertion since that index is never created in this test), and has been replaced with a meaningful check that test-index-1 remains in CLOSE state after the failed open attempt.

PR #21047, opened 2026-03-30
Add query filter, project, aggregation, sort converters

vinaykpud:
### Description
Adds DSL-to-Calcite converters for the `dsl-query-executor` sandbox plugin. Converts OpenSearch SearchSourceBuilder DSL into Calcite RelNode logical plans for execution by the analytics engine.

 Converters added:
  - FilterConverter — converts query DSL (term query) into LogicalFilter
  - ProjectConverter — converts _source field selection (includes, excludes, wildcards) into LogicalProject
  - SortConverter — converts sort, from, size into LogicalSort with collation and pagination
  - AggregateConverter — converts aggregations into LogicalAggregate (GROUP BY + metrics)
  - PostAggregateConverter — applies post-aggregation sorting from bucket orders (_key, _count, metric-based)
  - CollationResolver — resolves BucketOrder to Calcite RelFieldCollation against the post-agg schema

  Supporting infrastructure:
  - SearchSourceConverter — orchestrator that wires converters into query plans: 
         - Scan → Filter → Project → Sort (HITS) 
         - Scan → Filter → Aggregate → PostAggregate (AGGREGATION)
  - ConversionContext — immutable shared state for converters
  - AggregationTreeWalker — walks nested aggregation DSL tree, groups metrics by granularity level, collects bucket orders
  - AggregationMetadata / AggregationMetadataBuilder — pre-computed metadata per granularity with GROUP BY, aggregate calls, and bucket orders

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/20914

### Check List


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/AbstractMetricTranslator.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public String getAggregateFieldName(T agg) {
+        return agg.getName();
+    }
+
+    // TODO: implement response conversion per metric type (InternalAvg, InternalSum, etc.)
+    @Override
+    public InternalAggregation toInternalAggregation(String name, Object value) {
+        throw new UnsupportedOperationException("toInternalAggregation not yet implemented for " + getClass().getSimpleName());
```
nssuresh2007, line 73: Do we want to add these as separate PR? Any reason to not add them in the same PR?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationType.java
```diff
@@ -0,0 +1,22 @@
...
+package org.opensearch.dsl.aggregation;
+
+import org.opensearch.search.aggregations.AggregationBuilder;
+
+/**
+ * Base type interface for aggregation translators.
+ * Provides type identification for the {@link AggregationRegistry}.
+ * Bucket and metric subtypes define their own contracts.
+ */
+public interface AggregationType<T extends AggregationBuilder> {
```
expani: nit: Should be called AggregationTranslator

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/AggregateConverter.java
```diff
@@ -0,0 +1,34 @@
...
+    public AggregateConverter() {}
+
+    /**
+     * Builds a LogicalAggregate from the given metadata.
+     *
+     * @param input the input plan (scan + filter)
+     * @param metadata pre-computed aggregation metadata for one granularity
+     * @return the LogicalAggregate node
+     */
+    public RelNode convert(RelNode input, AggregationMetadata metadata) {
```
expani, line 31: Looks like a util method only used in UTs. If so, constructor should be private.
Can we move it out of source if this will be restricted to UTs going forward ?

sandbox/plugins/parquet-data-format/src/main/rust/Cargo.lock
```diff
@@ -0,0 +1,1759 @@
+# This file is automatically @generated by Cargo.
```
mch2: unintentional?

sandbox/plugins/dsl-query-executor/src/internalClusterTest/java/org/opensearch/dsl/DslAggregationIT.java
```diff
@@ -0,0 +1,124 @@
...
+            search(
+                new SearchSourceBuilder().size(0)
+                    .aggregation(new TermsAggregationBuilder("by_brand").field("brand").order(BucketOrder.count(true)))
+            )
+        );
+    }
+
+    public void testTermsBucketOrderByMetric() {
+        createTestIndex();
+        assertOk(
```
mch2, line 113: these tests are incomplete with the only assertion being assertOk - are we planning to validate the generated relnode here or in another test suite?  Do these need to be ITs or can we do a more lightweight component level UT?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/AbstractMetricTranslator.java
mch2: +1 - can we introduce things as they are wired?

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/query/TermQueryTranslatorTests.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public void testConvertsTermQueryToEquals() throws ConversionException {
+        RexNode result = translator.convert(QueryBuilders.termQuery("name", "laptop"), ctx);
+
+        assertTrue(result instanceof RexCall);
+        RexCall call = (RexCall) result;
+        assertEquals(SqlKind.EQUALS, call.getKind());
+        assertEquals(2, call.getOperands().size());
+        assertTrue(call.getOperands().get(0) instanceof RexInputRef);
+        // assertTrue(call.getOperands().get(1) instanceof RexLiteral);
```
mch2: why commented?

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/query/QueryRegistryTests.java
```diff
@@ -0,0 +1,52 @@
...
+
+        assertNotNull(result);
+        assertFalse(result instanceof UnresolvedQueryCall);
+    }
+
+    public void testResolvesMatchAllQuery() throws ConversionException {
+        RexNode result = registry.convert(QueryBuilders.matchAllQuery(), ctx);
+
+        assertNotNull(result);
+        assertFalse(result instanceof UnresolvedQueryCall);
```
mch2: can we add a better assert here on the returned RexNode?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/executor/QueryPlans.java
```diff
@@ -106,11 +106,8 @@ public Builder add(QueryPlan plan) {
             return this;
         }
 
-        /** Builds the plans. At least one must have been added. */
+        /** Builds the plans. May be empty (e.g. size=0, no aggs — only metadata response). */
```
mch2: not following this - what is a metadata response in this case?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/MetricTranslator.java
```diff
@@ -0,0 +1,50 @@
...
+    String getAggregateFieldName(T agg);
+
+    /**
+     * Converts a raw result value from execution into an OpenSearch InternalAggregation.
+     *
+     * @param name the aggregation name
+     * @param value the raw value (may be null)
+     * @return the corresponding InternalAggregation
+     */
+    InternalAggregation toInternalAggregation(String name, Object value);
```
expani, line 52: Let's natively use Arrow Types in the interface or a very thin Wrapper over it instead of `Object value` 

Analytics Core will also treat Arrow as a First Class citizen.

expani: Another thing to note would be making this a stream/iterator of `<String,Object>` so that response conversion doesn't have to face any virtual call overhead can be done in bulk.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/bucket/BucketTranslator.java
```diff
@@ -0,0 +1,59 @@
...
+    BucketOrder getBucketOrder(T agg);
+
+    /**
+     * Converts grouped bucket entries into an OpenSearch InternalAggregation for response building.
+     *
+     * @param agg the original aggregation builder
+     * @param buckets the bucket entries with keys, doc counts, and sub-aggs
+     * @return the InternalAggregation
+     */
+    InternalAggregation toBucketAggregation(T agg, List<BucketEntry> buckets);
```
expani: This can be an iterator instead of a List. Will help avoid loading all buckets in memory in case there is some post processing to be applied after conversion to InternalAggregation.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/FieldGrouping.java
```diff
@@ -0,0 +1,52 @@
...
+        this.fieldNames = List.copyOf(fieldNames);
+    }
+
+    @Override
+    public List<String> getFieldNames() {
+        return fieldNames;
+    }
+
+    @Override
+    public List<Integer> resolveIndices(RelDataType inputRowType) throws ConversionException {
```
expani: This looks like an utility method instead of something that will be implemented by different implementation of GroupingInfo, so shouldn't at the interface level.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationMetadata.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.calcite.util.ImmutableBitSet;
+import org.opensearch.search.aggregations.BucketOrder;
+
+import java.util.List;
+
+/**
+ * Pre-computed metadata for one aggregation granularity level.
+ * Contains everything needed to build a single {@code LogicalAggregate}.
+ *
+ * <p>A multi-level aggregation tree (e.g., terms → terms → avg) produces
```
expani, line 21: How will we handle `bucket_selector` that can lead to inner queries ? 
https://docs.opensearch.org/latest/aggregations/pipeline/bucket-selector

Something also has to capture the recursive/DAG like nature of `AggregationMetadata` so that the DSL query can be correctly referenced in terms of RelNode

expani (commented):
Thanks @vinaykpud for the PR. 

Dropping some comments on the high level abstractions. Will go through this further.

sandbox/plugins/dsl-query-executor/src/internalClusterTest/java/org/opensearch/dsl/DslAggregationIT.java
vinaykpud: Currently its assertOk because we don't have response construction in this plugin.

In the unit tests at each converter level we are validating the Logical operator built. We can have lightweight exhaustive UT for [SearchSourceConverter.java‎](https://github.com/opensearch-project/OpenSearch/pull/21047/changes#diff-4a2a3857832a41d8f6d626d14f5f2446f450319f834ff47b39664a275fe1c6af) which basically constructs the entire RelNode.

So at IT level we don't need to validate the RelNode, here we can focus on the actual response.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/AbstractMetricTranslator.java
vinaykpud: > Do we want to add these as separate PR? Any reason to not add them in the same PR?
Yes, this needs to be part of PR which implements response construction

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/executor/QueryPlans.java
vinaykpud: It doesn't makes sense. AI assumed it, I missed.

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/query/QueryRegistryTests.java
vinaykpud: Sure, i will add

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/query/TermQueryTranslatorTests.java
vinaykpud: I think i was testing something and forgot to remove, let me update

sandbox/plugins/parquet-data-format/src/main/rust/Cargo.lock
vinaykpud: Yes, will remove

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/AggregateConverter.java
vinaykpud: No, this is the method used by the `SearchSourceConverter` to convert the Aggregation part.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/FieldGrouping.java
vinaykpud: yeah, makes sense. I will move it to `AggregationMetadataBuilder`

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationMetadata.java
vinaykpud: Thanks for calling out this. We cant handle `bucket_selector` in this.  We need to be able to capture that information here.

`AggregationMetadata` should mirror the original DSL nesting. 

The current `AggregationMetadata` handles  metrics + bucket GROUP BY + sorting. For pipeline aggregations it require the tree relationship between granularity levels so we need to make this nested. 
Can we take up this enhancement in separate PR, I will add the TODO for this.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/bucket/BucketTranslator.java
vinaykpud: Sure

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/MetricTranslator.java
vinaykpud: Sure its a good optimization, I have updated the TODO for this will change

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationMetadata.java
expani: Sure Let's handle this in a follow up iteration

mch2 (approved):

msfroh:
It looks like one of the added tests directly contradicts https://github.com/opensearch-project/OpenSearch/pull/21203, which was merged around the same time.

See https://github.com/opensearch-project/OpenSearch/actions/runs/24427826934/job/71365665229?pr=20635

PR #21048, opened 2026-03-30
[Bugfix] Remove X-Request-Id format restrictions and make size configurable

### LINKED ISSUE #20688: [Feature Request]  Make X-Request-Id validation configurable

Hi OpenSearch Team,

I’m looking at PR #19798 where `X-Request-Id` is currently validated as **hex-only** and **exactly 32 chars**.

* Many gateways / load balancers emit request IDs that are not hex-only.
* Operators often want to pass through an existing correlation ID end-to-end without rewriting.

Before I start implementing anything, I’d like to ask if you would be OK with me opening a PR that removes the **hex-only** restriction and makes **only the length configurable**, while keeping the current behavior as the default.

* Introduce a setting like `http.request_id.max_length` (default: **32**).
* Validation would enforce:

  * header is present / non-empty (as today)
  * length is `<= max_length` (default keeps 32)
* The **hex character check would be removed**, so operators can propagate common request-id formats (UUID/ULID/base64url-like tokens, vendor-specific IDs) unchanged.

### Related component

Search

_No response_

### Additional context

_No response_

finnegancarroll:
### Description
Introduces dynamic `http.request_id.max_length` setting for configuring maximum length for X-Request-Id headers. Removes the alpha-numeric validation of X-Request-Id.

### Related Issues
Resolves #20688

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/http/HttpTransportSettings.java
+++ b/server/src/main/java/org/opensearch/http/HttpTransportSettings.java
@@ -259,6 +259,15 @@ public final class HttpTransportSettings {
         Setting.Property.NodeScope
     );
 
+    public static final Setting<Integer> SETTING_HTTP_REQUEST_ID_MAX_LENGTH = intSetting(
+        "http.request_id.max_length",
+        128,
+        16,
+        1024,
+        Setting.Property.Dynamic,
+        Setting.Property.NodeScope
+    );
+
     // Enable HTTP/3 protocol if supported by the operating system and architecture
     // The HTTP/3 transport is still experimental and should be used with caution.
     public static final Setting<Boolean> SETTING_HTTP_HTTP3_ENABLED = Setting.boolSetting(
--- a/server/src/test/java/org/opensearch/common/util/RequestUtilsTests.java
+++ b/server/src/test/java/org/opensearch/common/util/RequestUtilsTests.java
@@ -8,73 +8,84 @@
 
 package org.opensearch.common.util;
 
+import org.opensearch.common.settings.Settings;
 import org.opensearch.core.common.Strings;
+import org.opensearch.http.HttpTransportSettings;
 import org.opensearch.test.OpenSearchTestCase;
 
 public class RequestUtilsTests extends OpenSearchTestCase {
+
+    private static final int DEFAULT_MAX_LENGTH = HttpTransportSettings.SETTING_HTTP_REQUEST_ID_MAX_LENGTH.getDefault(Settings.EMPTY);
 
     public void testGenerateID() {
         assertTrue(Strings.hasText(RequestUtils.generateID()));
     }
 
     public void testValidateRequestIdValid() {
-        RequestUtils.validateRequestId("a1b2c3d4e5f67890abcdef1234567890");
-        RequestUtils.validateRequestId("ABCDEF1234567890abcdef1234567890");
-        RequestUtils.validateRequestId("00000000000000000000000000000000");
-        RequestUtils.validateRequestId("ffffffffffffffffffffffffffffffff");
-        RequestUtils.validateRequestId("FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF");
+        RequestUtils.validateRequestId("a1b2c3d4e5f67890abcdef1234567890", DEFAULT_MAX_LENGTH);
+        RequestUtils.validateRequestId("ABCDEF1234567890abcdef1234567890", DEFAULT_MAX_LENGTH);
+        RequestUtils.validateRequestId("00000000000000000000000000000000", DEFAULT_MAX_LENGTH);
+        RequestUtils.validateRequestId("ffffffffffffffffffffffffffffffff", DEFAULT_MAX_LENGTH);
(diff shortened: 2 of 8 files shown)


finnegancarroll:
Code analyzer gives the following concern.
```
allows arbitrary characters (including newlines, control chars, or special sequences) in the request ID, which could enable log injection 
```

Gave this a quick test and it seems netty rejects these headers as invalid so we do not need to handle them ourselves.
```
curl -s --http1.1 -H $'X-Request-Id: foo\rbar' \
  -X GET "$BASE/test-index/_search" \
  -d '{"query": {"match_all": {}}}' | jq .
```
```
{
  "error": {
    "root_cause": [
      {
        "reason": "Validation failed for header 'X-Request-Id'"
      }
    ],
    "reason": "Validation failed for header 'X-Request-Id'",
    "caused_by": {
      "reason": "a header value contains prohibited character 0xd at index 3."
    }
  },
  "status": 400
}
```

finnegancarroll:
@sgup432 can you take a look when you have a chance?

server/src/main/java/org/opensearch/http/HttpTransportSettings.java
```diff
@@ -259,6 +259,14 @@ public final class HttpTransportSettings {
         Setting.Property.NodeScope
     );
 
+    public static final Setting<Integer> SETTING_HTTP_REQUEST_ID_MAX_LENGTH = intSetting(
+        "http.request_id.max_length",
+        128,
+        1,
```
sgup432: Do you think we should increase this min value to something like 16 atleast? As we shouldn't allow someone passing a header id with size as 1.

server/src/test/java/org/opensearch/common/util/RequestUtilsTests.java
```diff
@@ -13,68 +13,94 @@
 
 public class RequestUtilsTests extends OpenSearchTestCase {
```
sgup432, line 16: Can we also add some sanity IT tests(apart from this unit tests) around updating the requestId max length dynamically? And test both positive and negative flow. One basic should suffice.

server/src/test/java/org/opensearch/common/util/RequestUtilsTests.java
```diff
@@ -13,68 +13,94 @@
...
 
+    private static final int DEFAULT_MAX_LENGTH = 128;
```
sgup432: I would prefer to fetch this value from the settings itself instead of hardcoding.

server/src/main/java/org/opensearch/http/HttpTransportSettings.java
sgup432: In addition, we should also add a max value here? To avoid issues where a user can pass a very big header erroneously or intentionally?

finnegancarroll: Added a `1024` max to guard against excessively large ids. wdyt about keeping the min at 1? I think enforcing some other min value implies a range for valid id length instead of strict maximum and I can't think of any harm here.

server/src/test/java/org/opensearch/common/util/RequestUtilsTests.java
finnegancarroll: Since this change makes `validateRequestId` trivial I ended up migrating the remaining UTs into integ tests as well.

server/src/main/java/org/opensearch/http/HttpTransportSettings.java
sgup432: Yeah having some min value does help here. I am not sure what is the right value to start with, maybe 16?

server/src/test/java/org/opensearch/common/util/RequestUtilsTests.java
```diff
@@ -1,80 +0,0 @@
...
- * this file be licensed under the Apache-2.0 license or a
- * compatible open source license.
- */
-
-package org.opensearch.common.util;
-
-import org.opensearch.core.common.Strings;
-import org.opensearch.test.OpenSearchTestCase;
-
-public class RequestUtilsTests extends OpenSearchTestCase {
```
sgup432, line 14: Oh I didn't mean to remove this unit test. We still need this one.

I just asked to write a simple dynamic setting IT test(in an appropriate IT class) to test for the flow: 
1. Trigger search request with a request_id length like 64 which is <=max 
2. Update the setting to set the max value to 32 dynamically. 
3. Trigger the search request again, this time it should fail with invalid request_id.

finnegancarroll: Got it. Added back this UT suite and added IT with the above flow.

server/src/main/java/org/opensearch/http/HttpTransportSettings.java
finnegancarroll: Moved this setting to 16 minimum.

bowenlan-amzn (approved):
Looks good! 
Regarding potential character injection concerns — I guess an attacker would need existing access and permissions to the cluster to exploit this, at which point they have far more impactful ways to cause damage.

jainankitk (approved):
Looks good to me!

sgup432 (approved):

andrross (approved):

PR #21056, opened 2026-03-31
Fix flaky testOpenIndexOverLimit in  ClusterShardLimitIT

gagandhakrey:
Use assertAcked() for the close index operation instead of manually checking isAcknowledged(), which could return false when nodes fail to acknowledge within the default timeout on slow environments (e.g., Linux/Java 25 CI runners).

Also scope the pre-close health check to the specific index, and fix a copy-paste bug in the final assertion that was checking for 'snapshot-index' (from testRestoreSnapshotOverLimit) instead of verifying that 'test-index-1' remains in the CLOSE state.

### Description

### Related Issues

### Check List


server/src/internalClusterTest/java/org/opensearch/cluster/shards/ClusterShardLimitIT.java
```diff
@@ -681,7 +684,7 @@ public void testOpenIndexOverLimit() {
             verifyException(dataNodes, counts, e);
         }
         ClusterState clusterState = client.admin().cluster().prepareState().get().getState();
-        assertFalse(clusterState.getMetadata().hasIndex("snapshot-index"));
+        assertEquals(IndexMetadata.State.CLOSE, clusterState.getMetadata().index("test-index-1").getState());
```
cwperks, line 687: Good catch.

PR #21060, opened 2026-03-31
Extended UT Framework for Functional Correctness of query conversion

nssuresh2007:
Changes to test the end to end DSL query conversion without OpenSearch cluster. Currently adds support for match_all and terms with agg for the initial commit.

### Description
This feature introduces a golden-file-based unit test framework for the OpenSearch DSL query executor plugin (`sandbox/plugins/dsl-query-executor/`). The framework validates the correctness of both the forward conversion path (DSL → RelNode) and the reverse conversion path (RelNode execution results → SearchResponse DSL output) by comparing actual outputs against pre-approved golden files. Each golden file encodes a complete test case: the input DSL, the expected RelNode logical plan, simulated execution rows, and the expected DSL output.

### Requirements

- As a developer, I want each golden file to contain all inputs and expected outputs for a single test case, so that test cases are self-contained and easy to review in code reviews.
- As a developer, I want the test framework to automatically discover and parse golden files, so that adding a new test case requires only adding a new golden file.
- As a developer, I want to validate that a given DSL input produces the expected Calcite logical plan, so that regressions in the conversion pipeline are caught automatically.
- As a developer, I want to validate that simulated execution results are correctly converted back into a SearchResponse, so that regressions in the response building logic are caught automatically.
- As a developer, I want to verify that the forward and reverse paths are consistent within a golden file, so that the entire DSL → RelNode → execution → DSL output pipeline is validated end-to-end.
- As a developer, I want golden file test cases covering the hits pipeline scenarios, so that query translation correctness is validated for common query types.
- As a developer, I want golden file test cases covering the aggregation pipeline scenarios, so that aggregation translation correctness is validated for supported metric and bucket types.
- As a developer, I want a way to regenerate golden files when intentional changes are made to the conversion logic, so that maintaining golden files does not become a burden.
- As a developer, I want the golden file tests to run as fast unit tests without requiring a running OpenSearch cluster, so that they can be part of the standard build cycle.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
+++ b/sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
@@ -1,0 +1,323 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.dsl.golden;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import org.apache.calcite.rel.RelNode;
+import org.opensearch.common.xcontent.json.JsonXContent;
+import org.opensearch.core.xcontent.MediaTypeRegistry;
+import org.opensearch.core.common.Strings;
+import org.opensearch.core.xcontent.DeprecationHandler;
+import org.opensearch.core.xcontent.NamedXContentRegistry;
+import org.opensearch.core.xcontent.XContentParser;
+import org.opensearch.dsl.converter.SearchSourceConverter;
+import org.opensearch.dsl.executor.QueryPlans;
+import org.opensearch.dsl.result.ExecutionResult;
+import org.opensearch.dsl.result.SearchResponseBuilder;
+import org.opensearch.search.SearchModule;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.test.OpenSearchTestCase;
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.net.URL;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Golden file tests for the DSL query executor plugin.
+ *
+ * <p>Each test method loads its own specific golden file by name, keeping
+ * dependencies explicit. Tests validate the forward path (DSL → RelNode),
+ * reverse path (ExecutionResult → SearchResponse), and field name consistency.
+ */
+public class DslGoldenFileTests extends OpenSearchTestCase {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final NamedXContentRegistry X_CONTENT_REGISTRY = new NamedXContentRegistry(
+        new SearchModule(org.opensearch.common.settings.Settings.EMPTY, Collections.emptyList()).getNamedXContents()
+    );
+
+    // ---- Forward Path Helper ----
+
+    /**
+     * Loads a golden file, parses the inputDsl into a SearchSourceBuilder,
+     * invokes SearchSourceConverter.convert(), serializes the RelNode via
+     * explain(), and compares against expectedRelNodePlan. In update mode,
+     * overwrites the golden file instead of asserting.
+     */
+    private void runForwardPathTest(String goldenFileName) throws Exception {
+        GoldenTestCase tc = GoldenFileLoader.load(goldenFileName);
+        CalciteTestInfra.InfraResult infra = CalciteTestInfra.buildFromMapping(tc.getIndexName(), tc.getIndexMapping());
+
+        SearchSourceBuilder searchSource = parseSearchSource(tc.getInputDsl());
+        SearchSourceConverter converter = new SearchSourceConverter(infra.schema());
+        QueryPlans plans = converter.convert(searchSource, tc.getIndexName());
+
+        QueryPlans.Type expectedType = QueryPlans.Type.valueOf(tc.getPlanType());
+        List<QueryPlans.QueryPlan> matchingPlans = plans.get(expectedType);
+        assertFalse("No " + expectedType + " plan produced for " + goldenFileName, matchingPlans.isEmpty());
+
+        RelNode relNode = matchingPlans.get(0).relNode();
+        String actualPlan = relNode.explain().trim();
+
+        // Validate schema (row type) field names match executionFieldNames
+        List<String> relNodeFields = relNode.getRowType().getFieldNames();
+        assertEquals(
+            "Forward path schema mismatch for " + goldenFileName
+                + "\nRelNode row type fields: " + relNodeFields
+                + "\nGolden file executionFieldNames: " + tc.getExecutionFieldNames(),
+            tc.getExecutionFieldNames(),
+            relNodeFields
+        );
+
+        if (GoldenFileUpdater.isUpdateMode()) {
+            Path filePath = goldenFilePath(goldenFileName);
+            GoldenFileUpdater.updatePlan(filePath, actualPlan);
+        } else {
+            assertEquals(
+                "Forward path mismatch for " + goldenFileName
+                    + "\nExpected:\n" + tc.getExpectedRelNodePlan()
+                    + "\nActual:\n" + actualPlan,
+                tc.getExpectedRelNodePlan().trim(),
+                actualPlan
+            );
+        }
+    }
+
+    // ---- Reverse Path Helper ----
+
+    /**
+     * Loads a golden file, constructs an ExecutionResult from executionRows
+     * and field names, invokes SearchResponseBuilder.build(), serializes the
+     * SearchResponse to JSON, and compares against expectedOutputDsl ignoring
+     * non-deterministic fields (took, _shards). In update mode, overwrites
+     * the golden file instead of asserting.
+     */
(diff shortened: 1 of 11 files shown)


vinaykpud:
Thanks @nssuresh2007 

This PR adds a "golden file" testing layer that reads JSON files and validates:
- Forward path: SearchSourceBuilder → SearchSourceConverter.convert() → RelNode.explain() → assert matches expected plan string
- Reverse path: mock rows → SearchResponseBuilder.build() → assert JSON matches expected output

I think it's worth considering how this fits with what we already have:

1. Unit tests per converter/translator : these test individual modules for each DSL construct and RelNode construct.
2. internalClusterTest integration tests (DslQueryIT, DslAggregationIT, DslSortIT, DslProjectIT) : these test real execution against a real cluster.

The golden file tests chain SearchSourceConverter + SearchResponseBuilder but still use mocked infrastructure (no cluster), so there's a fair bit of overlap with the existing unit tests. One concern is that maintenance burden could grow over time : for example, if a RelNode plan shape changes due to a refactor (even though semantics are correct), all golden files would need updating.

A couple of alternatives worth considering:

We could extend the existing internalClusterTest suite for each new query shape. These already test real execution against a single-node cluster:

```java
// In DslQueryIT.java
public void testRangeQuery() {
    SearchResponse resp = executeQuery(indexName,
        "{\"query\":{\"range\":{\"price\":{\"gte\":100}}}}");
    assertEquals(RestStatus.OK, resp.status());
    // validate actual hits from real execution
}
```

If we want JSON-driven test cases for convenience, we could build that on top of integration test infrastructure. Two options that come to mind:


1. Java REST QA test : similar to the PPL query tests in https://github.com/opensearch-project/OpenSearch/tree/feature/datafusion/sandbox/qa/analytics-engine-rest. Tests queries against a
real cluster for a standard set of query shapes.
2. YAML REST QA test : since the DSL query executor intercepts standard _search via SearchActionFilter, we're essentially testing the search API with different query bodies. YAML REST tests could be a natural fit here with no custom framework needed (same pattern as https://github.com/opensearch-project/OpenSearch/tree/main/rest-api-spec).

What do you think? Happy to discuss which approach makes the most sense here.

nssuresh2007:
@vinaykpud thanks for the feedback.

This extended unit tests are needed for functional correctness of DSL plugin as a stand alone testing framework without OpenSearch cluster dependency. This requirement is not satisfied by the existing tests:
1. Unit tests - Very class level focused. Does not help in functional correctness verification testing to validate if the entire query conversion flow works or not.
2. Integration tests - Runs tests on actual cluster. While Integration Tests help to cover these test scenarios, IT can still fail if the underlying analytics plugin or backend plugin has issues since it focuses on end to end testing. This cannot help in standalone testing of DSL plugin to certify its functional correctness.

Hence this becomes necessary to have this extended UT to validate if the logic that we are currently adding is correct. Otherwise, we would be able to catch it very late in the development lifecycle until all downstream plugins are functional and testable. Also note that this is not a replacement for UT or IT, but it is complementary to the above.

Please let me know if that helps to address your query.

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
```diff
@@ -0,0 +1,233 @@
...
+
+        // Convert golden file rows to Object[] iterable
+        List<Object[]> rows = new ArrayList<>();
+        for (List<Object> row : tc.getExecutionRows()) {
+            rows.add(row.toArray());
+        }
+        ExecutionResult result = new ExecutionResult(plan, rows);
+
+        // Build SearchResponse
+        var response = SearchResponseBuilder.build(List.of(result), 0L);
```
bharath-techie: Forward path we'll use SearchSourceBuilder to convert to Calcite and handoff - but for reverse path, the plan is to get results as SearchResponseBuilder / SearchResponse to maintain result compatibility ? 

For SQL / PPL , we might directly get back the rows, hence the question.

nssuresh2007: > but for reverse path, the plan is to get results as SearchResponseBuilder / SearchResponse to maintain result compatibility ?

Yes, that is the plan

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
```diff
@@ -0,0 +1,233 @@
...
+        assertFalse("No " + expectedType + " plan produced for " + goldenFileName, matchingPlans.isEmpty());
+
+        RelNode relNode = matchingPlans.get(0).relNode();
+        String actualPlan = relNode.explain().trim();
+
+        if (GoldenFileUpdater.isUpdateMode()) {
+            Path filePath = goldenFilePath(goldenFileName);
+            GoldenFileUpdater.update(filePath, actualPlan, tc.getExpectedOutputDsl());
+        } else {
+            assertEquals(
```
tanik98: In RedNode plan, should we validate schema as well at each level?

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/GoldenFileLoader.java
```diff
@@ -0,0 +1,96 @@
...
+    private static void validate(GoldenTestCase testCase, Path filePath) {
+        requireNonNull(testCase.getTestName(), "testName", filePath);
+        requireNonNull(testCase.getIndexName(), "indexName", filePath);
+        requireNonNull(testCase.getIndexMapping(), "indexMapping", filePath);
+        requireNonNull(testCase.getInputDsl(), "inputDsl", filePath);
+        requireNonNull(testCase.getExpectedRelNodePlan(), "expectedRelNodePlan", filePath);
+        requireNonNull(testCase.getExecutionFieldNames(), "executionFieldNames", filePath);
+        requireNonNull(testCase.getExecutionRows(), "executionRows", filePath);
+        requireNonNull(testCase.getExpectedOutputDsl(), "expectedOutputDsl", filePath);
+    }
```
tanik98: Instead of this, can we use @NonNull in GoldenTestCase class itself?

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
```diff
@@ -0,0 +1,233 @@
...
+        if (GoldenFileUpdater.isUpdateMode()) {
+            Path filePath = goldenFilePath(goldenFileName);
+            GoldenFileUpdater.update(filePath, tc.getExpectedRelNodePlan(), actualOutput);
+        } else {
+            // Remove non-deterministic fields before comparison
+            Map<String, Object> expectedOutput = tc.getExpectedOutputDsl();
+            stripNonDeterministicFields(actualOutput);
+            stripNonDeterministicFields(expectedOutput);
+
+            assertEquals(
```
tanik98: Bucket order also might not be deterministic for cases where OpenSearch use topK approach at a shard level. e.g. terms bucket aggregation with size parameter.

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
```diff
@@ -0,0 +1,233 @@
...
+            throw new IllegalStateException("Golden file not found on classpath: golden/" + goldenFileName);
+        }
+        return Path.of(resource.toURI());
+    }
+
+    /**
+     * Removes non-deterministic fields (took, _shards) from a serialized
+     * SearchResponse map so that comparisons are stable.
+     */
+    private void stripNonDeterministicFields(Map<String, Object> responseMap) {
```
tanik98: Score also we will have to strip as it wo'nt be supported?

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/GoldenFileLoader.java
nssuresh2007: Currently not able to find any library which uses NonNull annotation. Also currently other places in OpenSearch also use `Objects.RequireNonNull` only

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
nssuresh2007: Added check for field names as well to validate the schema: https://github.com/opensearch-project/OpenSearch/pull/21210/changes#diff-4ba3b789a12988b89c5caf2e9a4f2d50adea38008059768576b4303205ef820fR79

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
nssuresh2007: Added check for score as well: https://github.com/opensearch-project/OpenSearch/pull/21210/changes#diff-4ba3b789a12988b89c5caf2e9a4f2d50adea38008059768576b4303205ef820fR225

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/golden/DslGoldenFileTests.java
nssuresh2007: Added sorting before comparing the buckets so that they would be deterministic: https://github.com/opensearch-project/OpenSearch/pull/21210/changes#diff-4ba3b789a12988b89c5caf2e9a4f2d50adea38008059768576b4303205ef820fR275

nssuresh2007:
Apologies, since the forked repo branch got deleted by mistake raised new PR after re-creating it and addressed the comments: https://github.com/opensearch-project/OpenSearch/pull/21210

PR #21065, opened 2026-04-01
Add stats, extended_stats, and value_count support to DSL Calcite plugin

varun-st:
## Summary
Add support for stats, extended_stats, and value_count metric aggregations to the DSL Calcite plugin.

### New Metric Aggregations
• **StatsMetricTranslator**: Implements stats aggregation returning 5 metrics (count, min, max, sum, avg)
• **ExtendedStatsMetricTranslator**: Implements extended_stats aggregation returning 7 metrics (count, min, max, sum, avg, variance, std_deviation)
• **ValueCountMetricTranslator**: Implements value_count aggregation returning document count

#### Stats Aggregation
• Generates 5 AggregateCall objects for COUNT, MIN, MAX, SUM, AVG
• Maps to Calcite's standard SQL aggregate functions
• Returns InternalStats with proper empty aggregation handling

#### Extended Stats Aggregation
• Generates 7 AggregateCall objects including VAR_POP and STDDEV_POP
• Calculates sum_of_squares from variance using formula: sum_of_squares = variance * count + avg² * count
• Uses default sigma value (2.0) for std_deviation_bounds
• Returns InternalExtendedStats with proper empty aggregation handling

#### Value Count Aggregation
• Maps to Calcite's COUNT function
• Returns InternalValueCount with count value

### Key Features
• **Composite Metric Pattern**: Stats and extended_stats use CompositeMetricTranslator interface for multi-value metrics
• **Type Handling**: 
  • COUNT returns BIGINT NOT NULL
  • AVG, VAR_POP, STDDEV_POP return DOUBLE
  • MIN, MAX, SUM preserve input field type
• **Empty Aggregation Support**: Proper handling when no data matches (count=0, min=+∞, max=-∞)
• **Null Handling**: All aggregations ignore null values (matching OpenSearch behavior)

### Testing
• Comprehensive unit tests for all three translators
• Tests cover: valid values, null handling, empty results, type conversions, edge cases
• Integration tests added to DslLogicalPlanIntegrationIT

### Files Added
• StatsMetricTranslator.java
• ExtendedStatsMetricTranslator.java
• ValueCountMetricTranslator.java
• StatsMetricTranslatorTest.java (180 lines, 11 tests)
• ExtendedStatsMetricTranslatorTest.java (197 lines, 11 tests)
• ValueCountMetricTranslatorTest.java (124 lines, 8 tests)

### Files Modified
• AggregationTreeWalker.java: Added handling for composite metrics
• AggregationResponseBuilder.java: Added buildCompositeMetric() method
• CompositeMetricTranslator.java: New interface for multi-value metrics
• AggregationRegistry.java: Registered new aggregation types
• DslLogicalPlanIntegrationIT.java: Added integration tests

## Testing
All unit tests pass. Integration tests verify end-to-end functionality with OpenSearch.

### DIFF
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationMetadataBuilder.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationMetadataBuilder.java
@@ -12,6 +12,7 @@ import org.apache.calcite.rel.RelCollations;
 import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.util.ImmutableBitSet;
@@ -94,7 +95,10 @@ public class AggregationMetadataBuilder {
         boolean noGroupBy = groupings.isEmpty();
         List<AggregateCall> allCalls = new ArrayList<>();
         for (AggregateCall call : aggregateCalls) {
-            if (noGroupBy) {
+            // COUNT is always non-nullable (returns 0 for empty sets), while other aggregations
+            // like AVG, MIN, MAX, SUM return null for empty sets when there's no GROUP BY
+            boolean isCount = call.getAggregation().getKind() == SqlKind.COUNT;
+            if (noGroupBy && !isCount) {
                 RelDataType nullableType = typeFactory.createTypeWithNullability(call.getType(), true);
                 allCalls.add(AggregateCall.create(
                     call.getAggregation(), call.isDistinct(), call.isApproximate(),
@@ -108,6 +112,10 @@ public class AggregationMetadataBuilder {
         List<String> allFieldNames = new ArrayList<>(aggregateFieldNames);
 
         if (implicitCountRequested) {
+            RelDataType bigIntNotNull = typeFactory.createTypeWithNullability(
+                typeFactory.createSqlType(SqlTypeName.BIGINT),
+                false
+            );
             allCalls.add(AggregateCall.create(
                 SqlStdOperatorTable.COUNT,
                 false,
@@ -116,7 +124,7 @@ public class AggregationMetadataBuilder {
                 List.of(),
                 -1,
                 RelCollations.EMPTY,
-                typeFactory.createSqlType(SqlTypeName.BIGINT),
+                bigIntNotNull,
                 IMPLICIT_COUNT_NAME
             ));
             allFieldNames.add(IMPLICIT_COUNT_NAME);
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/AvgMetricTranslator.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/AvgMetricTranslator.java
@@ -11,6 +11,8 @@ package org.opensearch.dsl.aggregation.metric;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.opensearch.search.aggregations.metrics.AvgAggregationBuilder;
+import org.opensearch.search.aggregations.InternalAggregation;
+import java.util.Map;
 
 /** Translates AVG metric aggregation to Calcite. */
 public class AvgMetricTranslator extends AbstractMetricTranslator<AvgAggregationBuilder> {
@@ -31,5 +33,11 @@ public class AvgMetricTranslator extends AbstractMetricTranslator<AvgAggregation
     @Override
     protected String getFieldName(AvgAggregationBuilder agg) {
         return agg.field();
+    }
+
+    @Override
+    public InternalAggregation toInternalAggregation(String name, Map<String, Object> values) {
+        // TODO: Implement conversion from DataFusion result to InternalAvg
+        throw new UnsupportedOperationException("toInternalAggregation not yet implemented for AvgMetricTranslator");
     }
 }
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/ExtendedStatsMetricTranslator.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/ExtendedStatsMetricTranslator.java
@@ -1,0 +1,118 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.dsl.aggregation.metric;
+
+import org.apache.calcite.rel.RelCollations;
+import org.apache.calcite.rel.core.AggregateCall;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.sql.SqlAggFunction;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.opensearch.dsl.converter.ConversionException;
+import org.opensearch.search.DocValueFormat;
+import org.opensearch.search.aggregations.InternalAggregation;
+import org.opensearch.search.aggregations.metrics.ExtendedStatsAggregationBuilder;
+import org.opensearch.search.aggregations.metrics.InternalExtendedStats;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Translator for extended_stats aggregation.
+ * Extended stats produces: count, min, max, sum, sum_of_squares, variance, std_deviation, std_deviation_bounds.
+ * Note: avg, stddev are not computed in Calcite - InternalExtendedStats calculates them dynamically.
+ * We compute variance to derive sum_of_squares (Calcite lacks SUM(x²) function).
+ */
+public class ExtendedStatsMetricTranslator implements MetricTranslator<ExtendedStatsAggregationBuilder> {
+
+    private static final int METRIC_COUNT = 5;
+    private static final double DEFAULT_SIGMA = 2.0;
+    private static final String COUNT_SUFFIX = "_count";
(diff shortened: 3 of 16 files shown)


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/AvgMetricTranslator.java
```diff
@@ -32,4 +34,9 @@ protected SqlAggFunction getAggFunction() {
     protected String getFieldName(AvgAggregationBuilder agg) {
         return agg.field();
     }
+
+    @Override
+    public InternalAggregation toInternalAggregation(String name, Map<String, Object> values) {
+        throw new UnsupportedOperationException("toInternalAggregation not yet implemented for AvgMetricTranslator");
```
nssuresh2007, line 41: Is this intentional? Any reason why we are not implementing this?

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/aggregation/metric/ExtendedStatsMetricTranslatorTest.java
```diff
@@ -0,0 +1,182 @@
+/*
```
varun-st, line 1: The directory structure of these test files need to be updated. I'll update it in a subsequent revision.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/ExtendedStatsMetricTranslator.java
```diff
@@ -0,0 +1,118 @@
...
+        RelDataTypeField field = rowType.getField(fieldName, true, false);
+        if (field == null) {
+            throw new ConversionException("Field '" + fieldName + "' not found");
+        }
+        int fieldIndex = field.getIndex();
+        String baseName = agg.getName();
+
+        List<AggregateCall> calls = new ArrayList<>(METRIC_COUNT);
+
+        calls.add(createAggregateCall(SqlStdOperatorTable.COUNT, Collections.emptyList(),
```
nssuresh2007: Should it be count(fieldIndex)? How do we differentiate between count(*) vs count(field)?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationMetadataBuilder.java
```diff
@@ -94,7 +95,8 @@ public AggregationMetadata build(RelDataType inputRowType, RelDataTypeFactory ty
         boolean noGroupBy = groupings.isEmpty();
         List<AggregateCall> allCalls = new ArrayList<>();
         for (AggregateCall call : aggregateCalls) {
-            if (noGroupBy) {
+            boolean isCount = call.getAggregation().getKind() == SqlKind.COUNT;
+            if (noGroupBy && !isCount) {
```
nssuresh2007, line 101: nit: Please add a comment on why count is excluded (since count is always non-nullable) which will help in better understanding of the code

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/AvgMetricTranslator.java
varun-st: I have added the implementation only for Aggregations that were introduced part of this change. I'll add a TODO item to add this as a followup.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationMetadataBuilder.java
varun-st: I have a comment in the new revision.

nssuresh2007 (approved):

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/StatsMetricTranslator.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.Map;
+
+/**
+ * Translator for stats aggregation.
+ * Stats produces multiple metrics: count, min, max, sum.
+ * Note: avg is not computed in Calcite - InternalStats calculates it as sum/count.
+ */
+public class StatsMetricTranslator implements MetricTranslator<StatsAggregationBuilder> {
+
+    private static final int METRIC_COUNT = 4;
```
vinaykpud, line 35: we should support all 5 aggregation: https://docs.opensearch.org/latest/aggregations/metric/stats/

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/StatsMetricTranslator.java
```diff
@@ -0,0 +1,108 @@
...
+
+        calls.add(createAggregateCall(SqlStdOperatorTable.COUNT, Collections.singletonList(fieldIndex),
+            field.getType(), baseName + COUNT_SUFFIX));
+        calls.add(createAggregateCall(SqlStdOperatorTable.MIN, Collections.singletonList(fieldIndex),
+            field.getType(), baseName + MIN_SUFFIX));
+        calls.add(createAggregateCall(SqlStdOperatorTable.MAX, Collections.singletonList(fieldIndex),
+            field.getType(), baseName + MAX_SUFFIX));
+        calls.add(createAggregateCall(SqlStdOperatorTable.SUM, Collections.singletonList(fieldIndex),
+            field.getType(), baseName + SUM_SUFFIX));
+
```
vinaykpud, line 67: include avg also

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/StatsMetricTranslator.java
```diff
@@ -0,0 +1,108 @@
...
+        return calls;
+    }
+
+    private AggregateCall createAggregateCall(SqlAggFunction function, List<Integer> argList,
+            RelDataType type, String name) {
+        return AggregateCall.create(
```
vinaykpud, line 73: we should have a common place to convert to `AggregateCall`, `AbstractMetricTranslator` also need this

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/ExtendedStatsMetricTranslator.java
```diff
@@ -0,0 +1,118 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Translator for extended_stats aggregation.
+ * Extended stats produces: count, min, max, sum, sum_of_squares, variance, std_deviation, std_deviation_bounds.
+ * Note: avg, stddev are not computed in Calcite - InternalExtendedStats calculates them dynamically.
```
vinaykpud, line 31: We cant support partial. Either we can support or not support.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/ExtendedStatsMetricTranslator.java
```diff
@@ -0,0 +1,118 @@
...
+import org.opensearch.search.aggregations.metrics.InternalExtendedStats;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Translator for extended_stats aggregation.
+ * Extended stats produces: count, min, max, sum, sum_of_squares, variance, std_deviation, std_deviation_bounds.
```
vinaykpud, line 30: Full list here : https://docs.opensearch.org/latest/aggregations/metric/extended-stats/
This just covers partial, please validate

vinaykpud:
Please edit the description, looks like AI generated but very verbose, just mention what changes we are introducing as part of this.

vinaykpud:
Please raise PR again the main

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/ExtendedStatsMetricTranslator.java
varun-st: All the stats should be covered. We use the underlying engine to compute only the stats that need to be computed directly from the raw data. The stats that can be derived for the other stats are already handled in https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/metrics/InternalExtendedStats.java.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/ExtendedStatsMetricTranslator.java
varun-st: We are supporting all the stats. Please refer to the previous comment for details.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/StatsMetricTranslator.java
varun-st: All the stats should be covered. We use the underlying engine to compute only the stats that need to be computed directly from the raw data. The stats that can be derived for the other stats are already handled in https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/metrics/InternalStats.java.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/StatsMetricTranslator.java
varun-st: Average is also included. Please refer to the response in the above comment.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/metric/StatsMetricTranslator.java
varun-st: Except for 2-3 parameters, almost everything else is dynamic in nature and need to be passed as a parameter. I am not sure how much of a value it would be extract this into a util method. Please let me know your thoughts.

Issue #20882, opened 2026-03-16
[Feature Request] Execute ingest pipelines synchronously on processor thread for pull-based ingestion 


imRishN:
### Is your feature request related to a problem? Please describe
PR #20873 introduced ingest pipeline support for pull-based ingestion using a `CompletableFuture` sync bridge over `IngestService.executeBulkRequest()`, which dispatches to the WRITE thread pool. This introduced unnecessary thread handoff, a 30-second timeout that risks data loss on slow pipelines, and potential WRITE pool contention on mixed nodes. 

### Describe the solution you'd like

Execute ingest pipelines synchronously on the processor thread for pull-based ingestion, eliminating the thread pool dispatch and CompletableFuture bridge.

PR #21068, opened 2026-04-01
Execute ingest pipelines synchronously on processor thread for pull-based ingestion

imRishN:
### Description
Add `executeBulkRequestSync()` to `IngestService` that calls `runBulkRequestInBatch()` directly on the calling thread, bypassing thread pool dispatch. Replace the `CompletableFuture` sync bridge in `IngestPipelineExecutor` with direct sync call. This eliminates thread handoff to WRITE pool and removes 30-second timeout.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/20882

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Handle dependencies between analyzers ([#19248](https://github.com/opensearch-project/OpenSearch/pull/19248))
 - Restore default `shard_path_type` to FIXED for snapshot repositories ([#20643](https://github.com/opensearch-project/OpenSearch/issues/20643))
 - Fix `_field_caps` returning empty results and corrupted field names for `disable_objects: true` mappings ([#20800](https://github.com/opensearch-project/OpenSearch/pull/20800))
+- Fix race condition in PeerFinder where concurrent connection attempts could fail ([#21055](https://github.com/opensearch-project/OpenSearch/pull/21055))
+
 
 ### Dependencies
 - Bump shadow-gradle-plugin from 8.3.9 to 9.3.1 ([#20569](https://github.com/opensearch-project/OpenSearch/pull/20569))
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/opensearch/indices/pollingingest/IngestPipelineExecutor.java
```diff
@@ -118,34 +108,18 @@ public Map<String, Object> executePipelines(String id, Map<String, Object> sourc
...
-            throw new RuntimeException("Ingest pipeline execution timed out after [" + PIPELINE_EXECUTION_TIMEOUT_SECONDS + "] seconds", e);
-        } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();
-            throw new RuntimeException("Ingest pipeline execution was interrupted", e);
-        } catch (ExecutionException e) {
-            throw new RuntimeException("Ingest pipeline execution failed", e.getCause());
+        // Execute pipeline synchronously on the calling thread — no thread pool dispatch
+        ingestService.executeBulkRequestSync(1, Collections.singletonList(indexRequest), (slot, e) -> failureRef.set(e), (thread, e) -> {
+            if (e != null) {
+                failureRef.compareAndSet(null, e);
```
varunbharadwaj: Is it possible to have non-null exception here? If e is non null, should we wrap it with the new exception so we don't lose information?

imRishN: Done

server/src/main/java/org/opensearch/ingest/IngestService.java
```diff
@@ -831,6 +831,20 @@ protected void doRun() {
         });
     }
 
+    /**
+     * Same as {@link #executeBulkRequest} but executes synchronously on the calling thread
+     * without dispatching to a thread pool.
+     */
```
msfroh, line 839: Can you add a comment here to make it clear that it's unsafe to invoke this method from a `TransportAction`, since it may be running on a request thread?

imRishN: Sure, added

PR #21079, opened 2026-04-02
Add merge support for Parquet data format plugin

Shailesh-Kumar-Singh:
## Description
  
  Adds streaming k-way merge for the Parquet data format plugin, along with writer sort-on-finalize support. Produces a single globally-sorted Parquet file from N input segment files.
  
  ### Writer Sort-on-Finalize
  
  When a segment file is finalized, accumulated unsorted data must be sorted before the file is ready for queries:
  
  - **Small files (≤ 32 MB):** Sorted in-memory using Arrow's built-in lexicographic sort
  - **Large files (> 32 MB):** External merge sort — each batch is sorted individually and written as a temporary chunk file, then k-way merged into the final sorted output
  
  This bounds memory to a single batch regardless of total file size.
  
  ### Index Sort Configuration
  
  The merge respects OpenSearch's `index.sort.*` settings:
  
  - **Sort columns:** Multi-column sort via `index.sort.field` — supports `long`, `date`, `integer`, `double`, `float`, `keyword`, and their variants. Integer/float types are widened to 64-bit for uniform comparison.
  - **Reverse sort:** Per-column ascending/descending direction via `index.sort.order`.
  - **Null handling:** Explicit null-first / null-last via `index.sort.missing` (OpenSearch defaults to nulls-last regardless of sort direction). Uses explicit null representations rather than sentinel values to avoid collision with real data.
  
  When no sort columns are configured, files are concatenated sequentially (unsorted merge).
  
  ### K-Way Merge — Three-Tier Cascade
  
  A min-heap tracks the current sort value of each input file. The algorithm pops the smallest entry and decides how much data to emit using three tiers, choosing the coarsest granularity possible:
  
  1. **Tier 1 — Drain:** When only one file remains, stream all its batches directly to output. Zero comparisons.
  2. **Tier 2 — Whole-Batch Emit:** If the last row of the current batch ≤ the heap top (next file's smallest value), emit the entire batch as a zero-copy Arrow slice. Repeat for consecutive batches. This is the dominant tier when sort ranges
  don't overlap — one comparison per 100K-row batch instead of 100K individual comparisons.
  3. **Tier 3 — Binary Search:** When a batch partially overlaps with the heap top, binary search finds the exact boundary row (O(log₂ B) ≈ 17 extractions for 100K rows). Emit the prefix, advance the cursor, re-insert into the heap.
  
  ### Pipelined Write Path
  
  Three concurrent stages so merge computation, compression, and disk IO overlap:
  
  1. **Merge loop** (calling thread) — k-way merge produces Arrow record batch slices, pads them to the union schema, and buffers them in memory. When ~1M rows accumulate, triggers a flush to Stage 2.
  2. **Column encoding** (Rayon work-stealing pool, 4 threads) — on flush, all buffered batches are concatenated into a single large batch, a sequential `___row_id` column is appended, and the batch is decomposed into individual leaf columns.
  Each column is then compressed and encoded in parallel across 4 threads via work-stealing — columns that finish quickly free their threads to pick up remaining work, naturally load-balancing across columns of varying complexity.
  3. **Disk write** (Tokio async runtime) — encoded row groups are sent over a bounded channel (capacity 2) to a background task, which dispatches each write to a blocking thread through a rate-limited wrapper (default 20 MB/s) to avoid
  starving concurrent indexing. The background task does not wait for the current write to complete before accepting the next row group — so disk write for row group N happens concurrently with encoding of row group N+1.
  
  The bounded channel between Stage 2 and Stage 3 provides natural backpressure — at most 2 encoded row groups can be queued. If the merge loop produces faster than disk can absorb, the channel fills and the merge loop blocks, preventing
  unbounded memory growth. Peak memory ~300-350 MB regardless of input size.
  
  ### Schema Evolution & Row ID
  
  Input files may have different column sets due to mapping changes. Union schema is computed once; batches from files missing a column get null-padded. `___row_id` is stripped from inputs and rewritten with globally sequential values `[0,
  total_rows)` for cross-format joins.


sandbox/plugins/parquet-data-format/src/main/rust/src/native_settings.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub reverse_sorts: Vec<bool>,
+}
+
+impl NativeSettings {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn get_compression_type(&self) -> &str {
+        self.compression_type.as_deref().unwrap_or("ZSTD")
```
chaitanya588: In the NativeSettings.java, we are already keeping the default values from the settings. Then do we need to keep default values here as well?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetSettings.java
```diff
@@ -9,40 +9,81 @@
...
+    public static final Setting<ByteSizeValue> DICT_SIZE_BYTES = Setting.byteSizeSetting(
+        "parquet.dict_size_bytes",
+        new ByteSizeValue(2, ByteSizeUnit.MB),
+        Setting.Property.IndexScope
+    );
+
+    /** Compression codec for Parquet files, e.g. ZSTD, SNAPPY, LZ4_RAW (default ZSTD). */
+    public static final Setting<String> COMPRESSION_TYPE = Setting.simpleString(
+        "parquet.compression_type",
+        "ZSTD",
```
chaitanya588: we need to change the compression type to "LZ4_RAW" here as well.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package org.opensearch.parquet.bridge;
+
+/**
+ * Immutable settings passed to the native Rust writer via JNI.
+ * The Rust side reads values through the getter methods.
+ * All fields are nullable; the native side falls back to defaults when null.
+ */
+public class NativeSettings {
```
chaitanya588, line 16: Wire up the bloom filter settings here.

sandbox/plugins/parquet-data-format/src/main/rust/src/parquet_merge_stream.rs
```diff
@@ -0,0 +1,1439 @@
...
+/// reverse flags.
+#[unsafe(no_mangle)]
+pub extern "system" fn Java_org_opensearch_parquet_bridge_RustBridge_mergeParquetFilesInRust(
+    mut env: JNIEnv,
+    _class: JClass,
+    input_files: JObject,
+    output_file: JString,
+    index_name: JString,
+) -> jint {
+    let input_files_vec = match convert_java_list_to_vec(&mut env, input_files) {
```
chaitanya588: This method is returning int where as in java layer we are seeing void. Can we make the signature in sync for avoiding any unwanted behaviour in future?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
```diff
@@ -90,8 +95,27 @@ public ParquetIndexingEngine(
...
+            .compressionLevel(indexSettings.getValue(ParquetSettings.COMPRESSION_LEVEL))
+            .pageSizeBytes(indexSettings.getValue(ParquetSettings.PAGE_SIZE_BYTES).getBytes())
+            .pageRowLimit(indexSettings.getValue(ParquetSettings.PAGE_ROW_LIMIT))
+            .dictSizeBytes(indexSettings.getValue(ParquetSettings.DICT_SIZE_BYTES).getBytes())
+            .rowGroupSizeBytes(indexSettings.getValue(ParquetSettings.ROW_GROUP_SIZE_BYTES).getBytes())
+            .build();
+        try {
+            RustBridge.onSettingsUpdate(config);
+        } catch (Exception e) {
+            logger.error("Failed to push Parquet settings to Rust store", e);
```
chaitanya588: Is there any reason to consume this exception silently?

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -93,48 +154,198 @@ impl NativeParquetWriter {
...
+            temp_filename, output_filename, sort_columns, reverse_sorts, nulls_first
+        );
+
+        if sort_columns.is_empty() {
+            log_info!("No sort columns specified, renaming temp file to final");
+            std::fs::rename(temp_filename, output_filename)?;
+            return Ok(());
+        }
+
+        Self::sort_and_write(temp_filename, output_filename, sort_columns, reverse_sorts, nulls_first)
```
chaitanya588: Is this going to work with large files as well? In the previous commit, i am seeing this same implementation for smaller sized files. But that is missing here.

sandbox/plugins/parquet-data-format/src/main/rust/src/parquet_merge_stream.rs
```diff
@@ -0,0 +1,1439 @@
...
+            }
+            arrow::datatypes::TimeUnit::Microsecond => {
+                col.as_primitive::<DurationMicrosecondType>().value(row)
+            }
+            arrow::datatypes::TimeUnit::Nanosecond => {
+                col.as_primitive::<DurationNanosecondType>().value(row)
+            }
+        },
+        other => {
+            return Err(MergeError::Logic(format!(
```
chaitanya588: In OpenSearch today we have index sort column for string type as well. Then this will fail right?

sandbox/plugins/parquet-data-format/src/main/rust/src/parquet_merge_stream.rs
```diff
@@ -0,0 +1,1439 @@
...
+fn get_sort_values(
+    batch: &RecordBatch,
+    row: usize,
+    col_indices: &[usize],
+    dtypes: &[ArrowDataType],
+    nulls_first: &[bool],
+) -> MergeResult<Vec<i64>> {
+    let mut values = Vec::with_capacity(col_indices.len());
+    for (i, (col_idx, dtype)) in col_indices.iter().zip(dtypes.iter()).enumerate() {
+        let nf = nulls_first.get(i).copied().unwrap_or(false);
```
chaitanya588: As of now in OpenSearch, if setting is not specified then default value is depending upon the sort order. Where as here we are always taking the default nulls_first value as a fixed one (false in this case) without taking sort order into account. This will give different default behavior b/w opensearch and parquet. To make it consistent we should follow the same logic as that of opensearch.

Shailesh-Kumar-Singh: No. 
In Opensearch also documents missing the sort field are always sorted last by default, regardless of sort order.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
Shailesh-Kumar-Singh: The `merge_sorted_chunks` for large files was also loading everything into memory (concatenating all chunks then sorting), so the `large file` path had no actual memory advantage over the `small file` path. It was doing extra work (chunk + sort + merge) for the same memory footprint.

sandbox/plugins/parquet-data-format/src/main/rust/src/parquet_merge_stream.rs
chaitanya588: Assume that we have a field "test_int_field" as a indexsort field, then we will be ending up with this chain. Here in setMissingValue we are taking order also into account.
 
```
"test_int_field" (string in index settings)
  → NumberFieldMapper.NumberFieldType (via mapping lookup)
    → SortedNumericIndexFieldData (via fielddataBuilder)
      → IndexNumericFieldData.sortField()
        → IntValuesComparatorSource (XFieldComparatorSource subclass)
          → missingObject(null, reverse) — the XOR logic
            → SortedNumericSortField.setMissingValue(Integer.MIN_VALUE or Integer.MAX_VALUE)
```

Shailesh-Kumar-Singh: > Here in setMissingValue we are taking order also into account.

We do not take order into account that is why we are doing XOR
ascending: false ^ false = false → MAX_VALUE : **nulls sort last**
descending: false ^ true = true → MIN_VALUE : **nulls sort last**

In both cases, nulls come last irrespective of sort order.

```
source.sortMissingLast(missingValue) ^ reverse ? SortedSetSortField.STRING_LAST : SortedSetSortField.STRING_FIRST
```

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
Shailesh-Kumar-Singh: We will have to instead call our streaming parquet merge flow by passing all the small files, that will merge and return the sorted large file we want.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -37,12 +38,23 @@ public class NativeParquetWriter {
...
      * @throws IOException if the native writer creation fails
      */
-    public NativeParquetWriter(String filePath, long schemaAddress) throws IOException {
+    public NativeParquetWriter(
+        String filePath,
+        String indexName,
+        long schemaAddress,
+        List<String> sortColumns,
+        List<Boolean> reverseSorts,
+        List<Boolean> nullsFirst
```
rayshrey: Can we create a separate class out of these values. 
Something which takes in indexSettings and returns these values.
Would be much more cleaner to have validation logics for these fields in that class and would be more maintainable in the sense that id we want to add new configs we only need to make changes in that class instead of changing multiple classes. Would make code simpler in VSRManager as well (getting these configs from indexSettings).

gradle.properties
```diff
@@ -32,3 +32,4 @@ systemProp.jdk.tls.client.protocols=TLSv1.2,TLSv1.3
 
 # jvm args for faster test execution by default
 systemProp.tests.jvm.argline=-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m
+systemProp.sandbox.enabled=true
```
rayshrey: I hope this is unintentional, let's not push this :)

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
```diff
@@ -0,0 +1,80 @@
...
+        public Builder indexName(String v) { this.indexName = v; return this; }
+        public Builder compressionType(String v) { this.compressionType = v; return this; }
+        public Builder compressionLevel(Integer v) { this.compressionLevel = v; return this; }
+        public Builder pageSizeBytes(Long v) { this.pageSizeBytes = v; return this; }
+        public Builder pageRowLimit(Integer v) { this.pageRowLimit = v; return this; }
+        public Builder dictSizeBytes(Long v) { this.dictSizeBytes = v; return this; }
+        public Builder rowGroupSizeBytes(Long v) { this.rowGroupSizeBytes = v; return this; }
+        public Builder bloomFilterEnabled(Boolean v) { this.bloomFilterEnabled = v; return this; }
+        public Builder bloomFilterFpp(Double v) { this.bloomFilterFpp = v; return this; }
+        public Builder bloomFilterNdv(Long v) { this.bloomFilterNdv = v; return this; }
```
rayshrey: Do we need any validation logic here?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetSettings.java
```diff
@@ -9,40 +9,81 @@
...
     public static final String DEFAULT_MAX_NATIVE_ALLOCATION = "10%";
-    /** Default maximum number of rows per VectorSchemaRoot before rotation. */
     public static final int DEFAULT_MAX_ROWS_PER_VSR = 50000;
 
-    /** Maximum native memory allocation for Arrow buffers, as a percentage of non-heap memory. */
+    /** Group setting prefix for all Parquet settings. */
+    public static final Setting<Settings> PARQUET_SETTINGS = Setting.groupSetting(
+        "parquet.",
+        Setting.Property.IndexScope
+    );
```
rayshrey: How does this work exactly ? Without this will other settings work or not ?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
Shailesh-Kumar-Singh: sure, let me see how we can do this.

gradle.properties
Shailesh-Kumar-Singh: yes its just for testing, will be removed in next commit.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -134,30 +167,267 @@ impl NativeParquetWriter {
...
+        );
+
+        if sort_columns.is_empty() {
+            log_info!("No sort columns specified, renaming temp file to final");
+            std::fs::rename(temp_filename, output_filename)?;
+            return Ok(());
+        }
+
+        let file_size = std::fs::metadata(temp_filename)?.len();
+        const MAX_MEMORY_SIZE: u64 = 32 * 1024 * 1024;
```
rayshrey: How did we come up with this number ? Also is file size the correct metric here, since the size will be the encoded and compressed size, which might amplify to a higher number of rows when doing `sort_small_file`. Can we take the decision based on number of rows or something similar.

Also can we consider making this value (along with other values such as batch size etc) configurable via node settings or something like that so that benchmarking becomes easier (Both jmh benchmarks and full end-to-end benchmarks as well)

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -134,30 +167,267 @@ impl NativeParquetWriter {
...
-                    Ok(Some(FinalizeResult { metadata: parquet_metadata, crc32 }))
+                    let writer = mutex.into_inner().unwrap();
+                    match writer.close() {
+                        Ok(_) => {
+                            log_info!("Successfully closed temp writer for: {}", temp_filename);
+                            // _file is dropped here, closing the file handle
+
+                            Self::sort_and_rewrite_parquet(&temp_filename, &filename, &index_name, &sort_columns, &reverse_sorts, &nulls_first)?;
+
+                            let _ = std::fs::remove_file(&temp_filename);
```
rayshrey: What happens if deletion of temp file fails ?

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -134,30 +167,267 @@ impl NativeParquetWriter {
...
-                    let parquet_metadata = writer.finish()?;
-                    let file_metadata = parquet_metadata.file_metadata();
-                    log_debug!("Successfully finalized writer for file: {}, num_rows={}", filename, file_metadata.num_rows());
-                    let crc32 = writer.inner().checksum();
-                    log_debug!("CRC32 for file {}: {:#010x}", filename, crc32);
-                    Ok(Some(FinalizeResult { metadata: parquet_metadata, crc32 }))
+                    let writer = mutex.into_inner().unwrap();
+                    match writer.close() {
+                        Ok(_) => {
+                            log_info!("Successfully closed temp writer for: {}", temp_filename);
```
rayshrey: nit - log_debug

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -134,30 +167,267 @@ impl NativeParquetWriter {
...
+
+    fn sort_and_rewrite_parquet(
+        temp_filename: &str,
+        output_filename: &str,
+        index_name: &str,
+        sort_columns: &[String],
+        reverse_sorts: &[bool],
+        nulls_first: &[bool],
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        log_info!(
```
rayshrey: Let's avoid info logs for every file level operation and use debug logs instead.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -134,30 +167,267 @@ impl NativeParquetWriter {
...
+        log_info!("Using in-memory sort for small file: {}", temp_filename);
+
+        let file = File::open(temp_filename)?;
+        let builder = ParquetRecordBatchReaderBuilder::try_new(file)?;
+        let arrow_reader = builder.with_batch_size(2048).build()?;
+
+        let mut batches = Vec::new();
+        for batch_result in arrow_reader {
+            batches.push(batch_result?);
+        }
```
rayshrey: Just trying to understand, we are ultimately loading the whole file into memory, right ?
Then why are we loading it in batches of smaller sizes ?

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -134,30 +167,267 @@ impl NativeParquetWriter {
...
+        if chunk_paths.is_empty() {
+            log_info!("No data to sort in file: {}", temp_filename);
+            std::fs::rename(temp_filename, output_filename)?;
+            return Ok(());
+        }
+
+        log_info!("Created {} sorted chunks, merging via streaming k-way merge", batch_count);
+
+        // Use the streaming merge to produce the final sorted file
+        crate::merge::merge_sorted(
```
rayshrey: nit: we can simply import(use) in the beginning and just write `merge_sorted` instead of `crate::merge::merge_sorted` here.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/io_task.rs
```diff
@@ -0,0 +1,183 @@
...
+// =============================================================================
+// Process-wide shared Tokio runtime for async IO
+// =============================================================================
+
+static IO_RUNTIME: OnceLock<Runtime> = OnceLock::new();
+
+fn get_io_runtime() -> &'static Runtime {
+    IO_RUNTIME.get_or_init(|| {
+        tokio::runtime::Builder::new_multi_thread()
+            .worker_threads(4)
```
rayshrey: nit: Move this 4 to a constant as well

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/context.rs
```diff
@@ -0,0 +1,221 @@
...
+                    let col_writer = writer_iter.next().ok_or_else(|| {
+                        MergeError::Logic("Fewer column writers than leaf columns".into())
+                    })?;
+                    leaves_and_writers.push((leaf, col_writer));
+                }
+            }
+        }
+
+        let chunk_results: Vec<
+            Result<parquet::arrow::arrow_writer::ArrowColumnChunk, parquet::errors::ParquetError>,
```
rayshrey, line 173: Are we blocking here till we get the encoded columns ?
Can we do this in async as well ?

Shailesh-Kumar-Singh:
addressing few AI review comments which can be ignored:

>The constructor extracts sortColumns, reverseSorts, and nullsFirst independently from index settings. If these lists have different sizes (e.g., sort fields are set but sort orders are not), the native bridge may receive mismatched arrays, potentially causing undefined behavior or panics in Rust. There is no validation that all three lists have the same length.

Already validated by OpenSearch core in `IndexSortConfig` constructor, it throws `IllegalArgumentException` on size mismatch at index creation time. These settings are `Final`, so they can't change after.


>After the binary search, cursor.row_idx is set to run_end and then cursor.advance()
increments it by 1, so the next row processed is run_end + 1. However, the slice
already includes run_end (length is run_end - run_start + 1), so run_end is
consumed. This is correct only if advance() moves past run_end. But if run_end == 
run_start (single row), run_len is 1 and the slice is correct. The issue is that
cursor.row_idx should be set to run_end before calling advance(), which increments
to run_end + 1 — this is consistent. However, verify that when lo == hi - 1
terminates the loop, the boundary is not off-by-one: rows [run_start..=lo] are <= 
heap_top, and row hi is > heap_top. The slice should be run_end - run_start + 1 rows
only if row at lo is confirmed <= heap_top. Consider adding an explicit check for
the lo value after the binary search to confirm it is still <= heap_top before
including it in the slice.

The existing logic is correct. The invariant guarantees `rows[lo] <= heap_top < rows[hi]` at every iteration, `lo` only advances on `<=` comparisons, so it cannot overshoot. Adding a post-loop verification check is dead code.

>The new indexName field added to MergeInput has no null check or validation in the MergeInput compact constructor or the Builder.build() method. A null indexName passed to the Rust bridge could cause a NullPointerException or undefined native behavior.

indexName is sourced from `indexSettings.getIndex().getName()` which is guaranteed non-null by OpenSearch core

>The NativeSettings is built with always-present default values for compressionType,                                                                                                                                                               
  compressionLevel, etc., but the NativeSettings Javadoc states "All fields are                                                                                                                                                                     
  nullable; the native side falls back to defaults when null." Passing non-null                                                                                                                                                                     
  defaults here overrides any Rust-side defaults unconditionally. If a user has not                                                                                                                                                                 
  explicitly configured a setting, the Java default (e.g., "LZ4_RAW", 2) will always                                                                                                                                                                
  be pushed, preventing the Rust layer from ever applying its own defaults. Consider                                                                                                                                                                
  only setting fields that are explicitly configured by the user.      
  
Java is the single source of truth for settings. `Setting.get()` always returns a resolved value (user-configured or default). Rust-side defaults are a defensive fallback, not a competing source of truth

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
Shailesh-Kumar-Singh: No. These values come from OpenSearch Setting definitions in ParquetSettings.java, which already validate ranges and types at index creation time (e.g., compressionLevel is bounded 1-9 via Setting.intSetting).

sandbox/plugins/parquet-data-format/src/main/rust/src/parquet_merge_stream.rs
Shailesh-Kumar-Singh: Yes this was a todo, support for String, Float32/64, Int 8 etc. was pending, have pushed now.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/context.rs
Shailesh-Kumar-Singh: >Are we blocking here till we get the encoded columns ?

Yes `get_merge_pool().install(|| ... .collect())` blocks the calling thread until all columns finish encoding.
>Can we do this in async as well ?

How? 
We need all encoded column chunks before we can send a complete row group to the I/O task.
OR 
If you meant making the encoding itself async, encoding is pure CPU-bound work, Rayon's      
  `par_iter` is already parallelizing the encoding across columns. Async would help when threads are sitting idle waiting on I/O.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
Shailesh-Kumar-Singh: added an error log, I do not want to propagate it as a hard error since the final sorted file was already written successfully, the temp file is just cleanup.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
Shailesh-Kumar-Singh: made this value configurable through node level settings.

sandbox/plugins/parquet-data-format/src/main/rust/src/native_settings.rs
Shailesh-Kumar-Singh: Java defaults are the source of truth for production. Rust defaults are a safety net for tests and edge cases where SETTINGS_STORE is empty (e.g., unit tests that call Rust directly without Java pushing settings first).
Making fields non-optional would break NativeSettings::default() which is used as a fallback in writer.rs and context.rs.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
Shailesh-Kumar-Singh: yes, we can load the full RecordBatch containing all rows as well, done.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/StreamingParquetMergeStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Implements merging of Parquet files.
+ */
+public class StreamingParquetMergeStrategy implements ParquetMergeStrategy {
```
mgodwan: I don't see any streaming specific logic in this java class. We should rename accordingly.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/StreamingParquetMergeStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private static final Logger logger = LogManager.getLogger(StreamingParquetMergeStrategy.class);
+
+    private final DataFormat dataFormat;
+    private final String indexName;
+    private final Path shardDataPath;
+
+    public StreamingParquetMergeStrategy(DataFormat dataFormat, String indexName, Path shardDataPath) {
```
mgodwan: I am assuming we would replace the shardDataPath with an object store in further PRs?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/NativeParquetMergeStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+                .forEach(file -> filePaths.add(shardDataPath.resolve(writerFileSet.directory()).resolve(file)))
+        );
+
+        String outputDirectory = shardDataPath.resolve(files.getFirst().directory()).toString();
+        String mergedFilePath = getMergedFilePath(writerGeneration, outputDirectory);
+        String mergedFileName = getMergedFileName(writerGeneration);
+
+        try {
+            // Merge files in Rust
+            RustBridge.mergeParquetFilesInRust(filePaths, mergedFilePath, indexName);
```
mgodwan, line 66: Should this return the final path for the created file?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
```diff
@@ -123,7 +130,7 @@ public void addDocument(ParquetDocumentInput doc) throws IOException {
             parquetField.createField(fieldType, activeVSR, pair.getValue());
         }
         int rowIndex = activeVSR.getRowCount();
-        BigIntVector rowIdVector = (BigIntVector) activeVSR.getVector("_row_id");
+        BigIntVector rowIdVector = (BigIntVector) activeVSR.getVector("___row_id");
```
mgodwan: Can we use the constant for ROW_ID here?https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/dataformat/DocumentInput.java#L25

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
```diff
@@ -0,0 +1,176 @@
...
+    private final String indexName;
+    private final String compressionType;
+    private final Integer compressionLevel;
+    private final Long pageSizeBytes;
+    private final Integer pageRowLimit;
+    private final Long dictSizeBytes;
+    private final Long rowGroupSizeBytes;
+    private final Boolean bloomFilterEnabled;
+    private final Double bloomFilterFpp;
+    private final Long bloomFilterNdv;
```
mgodwan, line 26: Do we need boxed types here?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
```diff
@@ -0,0 +1,176 @@
...
+ */
+
+package org.opensearch.parquet.bridge;
+
+/**
+ * Immutable settings passed to the native Rust writer via JNI.
+ * The Rust side reads values through the getter methods.
+ * All fields are nullable; the native side falls back to defaults when null.
+ */
+public class NativeSettings {
```
mgodwan, line 16: This can be a record?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private final String indexName;
+    private final String compressionType;
```
mgodwan, line 19: How are the values validated here?

gradle.properties
mgodwan: Please remove this

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/NativeParquetMergeStrategy.java
Shailesh-Kumar-Singh: The caller constructs and passes the merged file output path, Rust writes to that exact path. 
Hence, no need to return the path back.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/StreamingParquetMergeStrategy.java
Shailesh-Kumar-Singh: Yes.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/StreamingParquetMergeStrategy.java
Shailesh-Kumar-Singh: fine, renaming to `NativeParquetMergeStrategy`

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
Shailesh-Kumar-Singh: Given that this has 12 nullable fields, The builder pattern gives named setters which is more readable and less error prone because we have many same-typed params here.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
Shailesh-Kumar-Singh: yes, as these fields are all optional, the native side checks for null and falls back to its own defaults when a value isn't provided.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
Shailesh-Kumar-Singh: in ParquetSettings validation happens, NativeSettings is just a transport object that carries already-validated values to the FFI boundary.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -134,30 +153,276 @@ impl NativeParquetWriter {
...
+    ) -> Result<(), Box<dyn std::error::Error>> {
+        log_debug!("Using streaming merge sort for large file: {}", temp_filename);
+
+        let file = File::open(temp_filename)?;
+        let builder = ParquetRecordBatchReaderBuilder::try_new(file)?;
+        let arrow_reader = builder.with_batch_size(batch_size).build()?;
+
+        let mut chunk_paths: Vec<String> = Vec::new();
+        let mut batch_count = 0;
+        let temp_dir = std::env::temp_dir();
```
mgodwan: We should not use temp dir as this can be a separate file system itself. We should write to the same parquet directory with a prefix like temp_.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/unsorted.rs
```diff
@@ -0,0 +1,95 @@
...
+use super::schema::{projection_indices_excluding_row_id, ColumnMapping};
+
+/// Unsorted merge: reads each input file sequentially, pads to union schema,
+/// rewrites `___row_id` with globally sequential values. No sorting performed.
+pub fn merge_unsorted(
+    input_files: &[String],
+    output_path: &str,
+    index_name: &str,
+) -> MergeResult<()> {
+    log_info!(
```
mgodwan: Make this a trace log

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/unsorted.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    for path in input_files {
+        let file = File::open(path)?;
+        let builder = ParquetRecordBatchReaderBuilder::try_new(file)?;
+        let schema = builder.schema().clone();
+        let parquet_descr = builder.parquet_schema().clone();
+
+        let projection_indices = projection_indices_excluding_row_id(&schema);
+        let projection = parquet::arrow::ProjectionMask::roots(&parquet_descr, projection_indices);
+        let reader = builder.with_batch_size(BATCH_SIZE).with_projection(projection).build()?;
```
mgodwan: Can BATCH_SIZE be configurable?

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/io_task.rs
```diff
@@ -0,0 +1,186 @@
...
+pub const OUTPUT_FLUSH_ROWS: usize = 1_000_000;
+
+/// Disk write rate limit in MB/s.
+pub const RATE_LIMIT_MB_PER_SEC: f64 = 20.0;
+
+/// Number of threads in the shared Rayon pool for parallel column encoding.
+const RAYON_NUM_THREADS: usize = 4;
+
+/// Number of Tokio worker threads for async IO.
+const TOKIO_WORKER_THREADS: usize = 4;
```
mgodwan: These should be configurable. Can we start with having a factor of number of cores for the thread?

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/io_task.rs
```diff
@@ -0,0 +1,186 @@
...
+}
+
+/// Spawns the background IO task on the shared Tokio runtime.
+///
+/// The IO task owns the `SerializedFileWriter` and receives encoded row groups
+/// over a bounded channel. Each disk write is dispatched to `spawn_blocking`
+/// but is **not** awaited immediately — this allows the merge loop to prepare
+/// the next row group while the current one is still being flushed to disk.
+pub fn spawn_io_task(
+    writer: SerializedFileWriter<RateLimitedWriter<File>>,
```
mgodwan: We can leverage existing crc writer to avoid extra reads later

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/heap.rs
```diff
@@ -0,0 +1,189 @@
...
+            arrow::datatypes::TimeUnit::Nanosecond => col.as_primitive::<DurationNanosecondType>().value(row),
+        }),
+
+        // Float types → SortKey::Float
+        ArrowDataType::Float64 => SortKey::Float(col.as_primitive::<Float64Type>().value(row)),
+        ArrowDataType::Float32 => SortKey::Float(col.as_primitive::<Float32Type>().value(row) as f64),
+
+        // String types → SortKey::Bytes
+        ArrowDataType::Utf8 => SortKey::Bytes(col.as_string::<i32>().value(row).as_bytes().to_vec()),
+        ArrowDataType::LargeUtf8 => SortKey::Bytes(col.as_string::<i64>().value(row).as_bytes().to_vec()),
```
mgodwan, line 163: We should randomize generation of all data types and use them for sort tests

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/context.rs
```diff
@@ -0,0 +1,221 @@
...
+            let m = concat_batches(&self.data_schema, self.output_chunks.as_slice())?;
+            self.output_chunks.clear();
+            m
+        };
+        let n = merged.num_rows();
+
+        let with_id = append_row_id(&merged, self.next_row_id, &self.output_schema)?;
+        drop(merged);
+
+        let col_writers = self
```
mgodwan, line 158: We should close them in case of any errors?

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/sorted.rs
```diff
@@ -0,0 +1,216 @@
...
+    let pool = get_merge_pool();
+    let direction_label = if reverse_sorts.iter().all(|&r| !r) {
+        "ascending"
+    } else if reverse_sorts.iter().all(|&r| r) {
+        "descending"
+    } else {
+        "mixed"
+    };
+
+    log_info!(
```
mgodwan: We can make most of the info logs as debug/trace

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/heap.rs
```diff
@@ -0,0 +1,189 @@
...
+    }
+    Ordering::Equal
+}
+
+// =============================================================================
+// HeapItem for k-way merge
+// =============================================================================
+
+#[derive(Debug)]
+pub struct HeapItem {
```
mgodwan, line 92: Given this file contains heapitem, and sort utils, can we name the file as heap_utils.rs?

mgodwan:
Thanks for the changes @Shailesh-Kumar-Singh 
Like the overall documentation, and clean separation of concerns.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
mgodwan: How are the settings validated then during the create index?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
mgodwan: You can still use builder for ctor. Getters can be avoided

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/context.rs
Shailesh-Kumar-Singh: these are in-memory encoders with no OS resources, so drop would have been fine. But still added explicit cleanup anyways via pair_leaves_with_writers which returns unconsumed writers on error so the caller can close them before propagating.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/heap.rs
Shailesh-Kumar-Singh: the file has HeapItem and sort comparison functions that are coupled to the heap logic, not utility methods, so heap.rs should be fine.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/io_task.rs
Shailesh-Kumar-Singh: sure we can start with ( num of cores / 8 ) for Rayon & Tokio each

sandbox/plugins/parquet-data-format/src/main/rust/src/ffm.rs
```diff
@@ -326,3 +332,84 @@ pub unsafe extern "C" fn parquet_merge_files(
...
+// Parquet reader (for test verification)
+// ---------------------------------------------------------------------------
+
+/// Reads a parquet file and returns its contents as a JSON string.
+/// Each row is a JSON object. The result is a JSON array of objects.
+/// The JSON bytes are written into `out_buf`, actual length into `out_len`.
+/// Returns 0 on success.
+#[ffm_safe]
+#[no_mangle]
+pub unsafe extern "C" fn parquet_read_as_json(
```
mgodwan, line 346: This can just return as arrow batches. Why do we need json here?

Shailesh-Kumar-Singh: as this is just used for tests (sort order verification on small files) JSON avoids the FFI memory lifecycle management that Arrow batches would need

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
Shailesh-Kumar-Singh: that would be validated in ParquetSettings right, like ParquetSettings.java declares its constraints, e.g. Setting.intSetting("index.parquet.merge_batch_size", 100_000, 1, ...) enforces min=1 etc

sandbox/plugins/parquet-data-format/src/main/rust/src/ffm.rs
mgodwan: What is the additional overhead?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeSettings.java
mgodwan: i am concerned about compression type, and alike settings. We may add some setting values which will not be supported until the engine creation happens.

mgodwan: I will create a TODO for the same.

mgodwan (approved):
Capturing some additional tasks to be covered here:

* Tests with randomization for sorting across different data types.
* https://github.com/opensearch-project/OpenSearch/pull/21079#discussion_r3169946311
* https://github.com/opensearch-project/OpenSearch/pull/21079#discussion_r3161181799
* https://github.com/opensearch-project/OpenSearch/pull/21079/changes#r3169979457
* Add back streaming crc writer tests
* tokio runtime is still doing both io/cpu work. we need to look into decoupling of the same.

PR #21084, opened 2026-04-02
Adding per query memory tracking in Datafusion plugin

himshikhagupta:
### Description
1. Add per query memory pool which wraps the global memory pool, and implements DaatFusion's `MemoryPool`
2. This will be used for tracking the memory usage for each query, as reported by Datafusion
3. To handle cases where DF may be reporting higher memory due to double counting, additional check to verify mimalloc rss stats if there is room to grow.

### Check List


plugins/engine-datafusion/jni/src/query_metrics.rs
```diff
@@ -0,0 +1,523 @@
...
+                // memory actually reserved by mimalloc (not process-wide RSS).
+                let (current_rss, _) = get_mimalloc_rss();
+                if current_rss.saturating_add(additional) <= pool_limit() {
+                    log_info!(
+                        "QueryMemoryPool: inner pool denied {}B, but mimalloc has headroom \
+                         (rss={}MB, limit={}MB) — allowing",
+                        additional,
+                        current_rss / (1024 * 1024),
+                        pool_limit() / (1024 * 1024),
+                    );
```
Bukhtawar: Looks like the comment suggests `current_commit` to be more representative :) ?

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -914,9 +933,13 @@ pub extern "system" fn Java_org_opensearch_datafusion_jni_NativeBridge_executeFe
             exclude_fields,
             runtime,
             cpu_executor,
+            query_pool.map(|p| p as std::sync::Arc<dyn datafusion::execution::memory_pool::MemoryPool>),
         ).await {
             Ok(stream_ptr) => stream_ptr,
             Err(e) => {
+                // Snapshot stats and mark completed — Java's error handler
+                // will call drain_completed_query to retrieve native memory stats.
+                query_metrics::stop_query_tracking(context_id);
```
Bukhtawar: Can this be handled automatically as a part of QueryMemoryPool::drop as the query ends

himshikhagupta: QueryMemoryPool may not always be called depending on if DF needs to block/unblock memory. Explicit start and stop for query tracking seemed like a better approach rather than depending on MemoryPool.

plugins/engine-datafusion/jni/src/query_metrics.rs
himshikhagupta: Looks like a miss on the comment. I was using current_commit initially but changed to current_rss. This was because commit can be higher even if memory is not being used. rss will give exactly what is being used so switched to this. Wdyt?

plugins/engine-datafusion/jni/src/query_metrics.rs
```diff
@@ -0,0 +1,523 @@
...
+    });
+    QUERY_TRACKERS.insert(context_id, tracker);
+    Some(query_pool)
+}
+
+/// Mark a query as completed and log its final metrics.
+/// The tracker stays in the registry so Java can retrieve it via JNI.
+/// This mirrors Java's TaskResourceTrackingService pattern where stats are
+/// snapshotted onto the Task object before removal, and listeners consume
+/// the enriched object.
```
himshikhagupta: Right now query is marked as completed and left in the map. Will integrate with stats to emit aggregated metrics and add logic to cleanup completed queries in the map in the next change.

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -297,6 +242,42 @@ fn log_task_metrics(operation: &str, metrics: &tokio_metrics::TaskMetrics) {
...
 
+/// Log overall plugin memory usage from mimalloc's heap stats.
+/// This captures ALL Rust-side allocations (not just DataFusion's cooperative pool).
+#[allow(dead_code)]
+fn log_mimalloc_stats() {
+    let mut elapsed_ms: usize = 0;
+    let mut user_ms: usize = 0;
+    let mut system_ms: usize = 0;
+    let mut current_rss: usize = 0;
+    let mut peak_rss: usize = 0;
```
alchemist51: this is process rss right?

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -297,6 +242,42 @@ fn log_task_metrics(operation: &str, metrics: &tokio_metrics::TaskMetrics) {
...
+    let mut current_commit: usize = 0;
+    let mut peak_commit: usize = 0;
```
alchemist51: What does commit means? How are we consuming this?

plugins/engine-datafusion/jni/src/lib.rs
```diff
@@ -297,6 +242,42 @@ fn log_task_metrics(operation: &str, metrics: &tokio_metrics::TaskMetrics) {
...
+    let mut page_faults: usize = 0;
+
+    unsafe {
+        libmimalloc_sys::mi_process_info(
+            &mut elapsed_ms,
+            &mut user_ms,
+            &mut system_ms,
```
alchemist51: Do we need elapsed, user and system ms when we are not logging it?

plugins/engine-datafusion/jni/src/lib.rs
alchemist51:
```
        let inner = SessionContext::new_with_state(state);
        register_selector_aggregates(&inner);
        register_iox_scalar_functions(&inner);
        if let Some(default_catalog) = self.default_catalog {
            inner.register_catalog(DEFAULT_CATALOG, default_catalog);
        }

        IOxSessionContext::new(inner, self.exec, recorder, memory_monitor)
```

Can we not create a custom context which will handle all the stop, start and drop internally like influx: https://github.com/influxdata/influxdb3_core/blob/main/iox_query/src/exec/context.rs#L310?

plugins/engine-datafusion/jni/src/query_metrics.rs
```diff
@@ -0,0 +1,523 @@
...
+use vectorized_exec_spi::log_info;
+
+use datafusion::execution::memory_pool::{MemoryConsumer, MemoryPool, MemoryReservation};
+use datafusion::common::DataFusionError;
+
+// ---------------------------------------------------------------------------
+// Global pool limit — set once at createGlobalRuntime, read by QueryMemoryPool
+// ---------------------------------------------------------------------------
+
+static POOL_LIMIT: OnceLock<usize> = OnceLock::new();
```
alchemist51: Can this be present in the Runtime?

plugins/engine-datafusion/jni/src/query_metrics.rs
```diff
@@ -0,0 +1,523 @@
...
+        match self.inner.try_grow(reservation, additional) {
+            Ok(()) => {
+                self.track_grow(additional);
+                Ok(())
+            }
+            Err(e) => {
+                // Inner pool said no — check mimalloc's committed memory against
+                // the configured pool limit as a fallback. current_commit reflects
+                // memory actually reserved by mimalloc (not process-wide RSS).
+                let (current_rss, _) = get_mimalloc_rss();
```
alchemist51: if we calculate here, won't this be on hot path? Do we want a thread which will keep on updating and we fetching it here?

plugins/engine-datafusion/jni/src/query_metrics.rs
```diff
@@ -0,0 +1,523 @@
...
+                if current_rss.saturating_add(additional) <= pool_limit() {
+                    log_info!(
+                        "QueryMemoryPool: inner pool denied {}B, but mimalloc has headroom \
+                         (rss={}MB, limit={}MB) — allowing",
+                        additional,
+                        current_rss / (1024 * 1024),
+                        pool_limit() / (1024 * 1024),
+                    );
+                    self.inner.grow(reservation, additional);
```
alchemist51: How the memory will look in the inner memory pool? For example:

inner and outer were reporting 3GB(4GB limit). We got 1GB try grow. We realised that current usage is actually 2GB only.. How the inner value will now be updated? This one will make it 4GB and the inner will throw an error?

plugins/engine-datafusion/jni/src/lib.rs
Bukhtawar: Basically retaining it as more of a defensive check but also stopping the tracking earlier for the happy flows

alchemist51: @himshikhagupta  @Bukhtawar  @bharath-techie wdyt?

plugins/engine-datafusion/jni/src/query_metrics.rs
```diff
@@ -0,0 +1,523 @@
...
+pub struct QueryMemoryPool {
+    /// The shared global memory pool that enforces the overall limit.
+    inner: Arc<dyn MemoryPool>,
+    /// Current bytes reserved by this query.
+    current_bytes: AtomicUsize,
+    /// Peak bytes reserved by this query.
+    peak_bytes: AtomicUsize,
+}
+
+impl QueryMemoryPool {
```
Bukhtawar: Can we rename to call this file `query_memory_pool_tracker`?

plugins/engine-datafusion/jni/src/query_metrics.rs
```diff
@@ -0,0 +1,523 @@
...
+            &mut current_rss,
+            &mut peak_rss,
+            &mut current_commit,
+            &mut peak_commit,
+            &mut page_faults,
+        );
+    }
+
+    (current_rss, peak_rss)
+}
```
Bukhtawar: Can this be cached, as we are looking this up for all subsequent queries post the first try_grow failure?

plugins/engine-datafusion/jni/src/lib.rs
himshikhagupta: With the existing code, it would be rss for the allocator initialized in datafusion. With the common allocator now this will change.

plugins/engine-datafusion/jni/src/lib.rs
himshikhagupta: These are standard variables. Need to initialize all since that is the only way the process_info API works.

plugins/engine-datafusion/jni/src/query_metrics.rs
himshikhagupta: We are only doing this if DF MemoryPool rejects the call, in that case we want to get current rss usage. Again, with the common allocator this logic will change.

alchemist51: I think with the common allocator, this will involve not just reading mimalloc but also checking other components memory and maybe doing a delta? essentially this method call will bulk up more.

plugins/engine-datafusion/jni/src/query_metrics.rs
alchemist51: I think we have a similar thread for this ongoing here: https://github.com/opensearch-project/OpenSearch/pull/21084#discussion_r3049494759

plugins/engine-datafusion/jni/src/query_metrics.rs
himshikhagupta: done

himshikhagupta:
Addressed comments:
1. Added a new QueryTrackerContext which handles query tracking behind the scenes.
2. Removed the mimalloc RSS fallback from QueryMemoryPool since that is going to change with the new common allocator design. Will address this change in a separate PR.

@Bukhtawar @alchemist51

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -161,6 +161,10 @@ public Iterable<Object[]> execute(RelNode logicalFragment, Object context) {
             EngineBridge<byte[], ? extends EngineResultStream, RelNode> bridge =
                 (EngineBridge<byte[], ? extends EngineResultStream, RelNode>) plugin.bridge(engine, snapshot.getRef());
 
+            if (context instanceof Long) {
```
alchemist51, line 164: Why this change is needed?

alchemist51:
LGTM otherwise!

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
himshikhagupta: Need to pass context from all layers for supporting the JNI methods

himshikhagupta:
mainline PR corresponding to this https://github.com/opensearch-project/OpenSearch/pull/21211

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
Bukhtawar: I think we need to handle this better. Why do we think the `Object` context would just be a context id?

himshikhagupta:
Closing this in favor of mainline PR https://github.com/opensearch-project/OpenSearch/pull/21211

PR #21086, opened 2026-04-02
Refactor and simplify the `_source` parsing in FetchSourceContext

### LINKED ISSUE #20612: [Feature Request] [BUG] `_source` include / exclude validation

urmichm:
### Description
Refactoring and simplification of `FetchSourceContext.java` to better align with current Java best practices.
The first part of resolving the #20612 issue. 

To keep this PR focused and easier to understand, the remaining work will be addressed in follow-up PRs.

### Related Issues
Partially Resolves #20612

### Check List


server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
```diff
@@ -133,101 +133,110 @@ public static FetchSourceContext parseFromRestRequest(RestRequest request) {
...
+                String[] includes = new String[] { parser.text() };
+                return new FetchSourceContext(true, includes, null);
+            }
+            case XContentParser.Token.START_ARRAY -> {
+                ArrayList<String> list = new ArrayList<>();
+                while ((token = parser.nextToken()) != XContentParser.Token.END_ARRAY) {
+                    list.add(parser.text());
+                }
+                String[] includes = list.toArray(new String[0]);
+                return new FetchSourceContext(true, includes, null);
```
urmichm, line 157: The `parseSourceArray` method is not used here to keep the original implementation of unchecked parsing. This PR focuses solely on refactoring.

urmichm:
> The error message in the default case of fromXContent now includes VALUE_STRING and START_ARRAY as valid tokens, which is an improvement. However, the old error message only listed VALUE_BOOLEAN and START_OBJECT. This is a behavioral change in error messages that could affect users or tests relying on specific error message content.

The error message has been reverted to its original form

urmichm:
> The fromXContent method uses a switch expression with arrow cases but the method signature requires a return value. The comment // MUST never reach here after the switch block suggests the compiler requires a statement after the switch, but since all cases either return or throw, this may cause a compilation error depending on Java version. The switch does not have an explicit return after it, which could be a compile-time issue if the compiler cannot prove exhaustiveness.

I have removed the confusing comment

> In parseSourceObject, the switch statement uses arrow-case syntax (Java 14+ pattern), but the START_ARRAY and VALUE_STRING cases do not have explicit break or return statements after their blocks. With arrow-case syntax this is fine, but it's worth verifying there is no unintended fall-through behavior, especially since the default case throws an exception that should only be reached for truly unknown tokens.

The project is compiled with the appropriate Java version, and this should not trigger any issues.

cc: @bowenlan-amzn

urmichm:
> Logic change in parseFromRestRequest: `fetchSource == null ? true : fetchSource` was replaced with `fetchSource == null || fetchSource`. These are logically equivalent, but the change deserves scrutiny....
https://github.com/opensearch-project/OpenSearch/pull/21086#issuecomment-4180281277

Reverted the change to keep PR simple

bowenlan-amzn:
@urmichm the codecov is not that good, but definitely not because your refactor.
Will let you decide whether some more coverage is good with this PR or easier to do it in your next PR. 
I'm good to merge as it is now. cc @sandeshkr419

urmichm:
Hi @bowenlan-amzn 
I have unit tests in the next PR,  let's keep this one as refactoring only. 
The next one will contain all the necessary tests

PR #21091, opened 2026-04-03
Add histogram and date_histogram aggregation support

varun-st:
## Description

Adds support for histogram and date_histogram bucket aggregations to the DSL query executor plugin.

### Core Implementation
• **Expression Grouping Framework**: Introduces ExpressionGrouping interface for aggregations that compute bucket values using formulas (histogram, date_histogram)
• **Projected Column Names**: Generates unique column names for computed expressions to avoid conflicts with source fields
• **Bucket Translators**: Implements HistogramBucketTranslator and DateHistogramBucketTranslator with gap filling support for min_doc_count=0
• **Response Builder**: New AggregationResponseBuilder handles nested aggregation response construction with proper granularity key matching

### Key Features
• ✅ Histogram aggregation with configurable interval and offset
• ✅ Date histogram with calendar and fixed intervals
• ✅ Gap filling for missing buckets
• ✅ Nested aggregations (histogram within terms, etc.)
• ✅ Sibling aggregations
• ✅ Bucket sorting with order parameter
• ✅ Consistent granularity key generation between planning and execution phases

### Bug Fixes
• Fixed granularity key generation to use projected column names consistently across planning and result phases
• Fixed AggregationMetadata to store correct schema column names for expression groupings
• Fixed child filters in nested aggregations to use actual column names from SQL results

### Testing
• Unit tests for HistogramGrouping and DateHistogramGrouping
• Unit tests for bucket translators with gap filling scenarios
• Integration tests covering nested and sibling histogram/date_histogram aggregations
• Unit tests for ComparisonUtils, CollationResolver, and AggregationResponseBuilder

## Example Usage

json
{
  "aggs": {
    "price_histogram": {
      "histogram": {
        "field": "price",
        "interval": 50,
        "min_doc_count": 0
      },
      "aggs": {
        "avg_quantity": {
          "avg": { "field": "quantity" }
        }
      }
    }
  }
}

### DIFF
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
@@ -13,6 +13,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.action.search.SearchResponse;
+import org.opensearch.action.search.SearchResponseSections;
+import org.opensearch.action.search.ShardSearchFailure;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.analytics.EngineContext;
@@ -25,12 +27,15 @@ import org.opensearch.core.index.Index;
 import org.opensearch.dsl.converter.SearchSourceConverter;
 import org.opensearch.dsl.executor.DslQueryPlanExecutor;
 import org.opensearch.dsl.executor.QueryPlans;
+import org.opensearch.dsl.result.AggregationResponseBuilder;
 import org.opensearch.dsl.result.ExecutionResult;
-import org.opensearch.dsl.result.SearchResponseBuilder;
+import org.opensearch.search.SearchHits;
+import org.opensearch.search.aggregations.InternalAggregations;
 import org.opensearch.tasks.Task;
 import org.opensearch.transport.TransportService;
 
 import java.util.List;
+import java.util.stream.Collectors;
 
 /**
  * Coordinates DSL query execution: converts SearchSourceBuilder to Calcite RelNode plans,
@@ -85,12 +90,57 @@ public class TransportDslExecuteAction extends HandledTransportAction<SearchRequ
             List<ExecutionResult> results = planExecutor.execute(plans);
             long tookInMillis = System.currentTimeMillis() - startTime;
 
-            SearchResponse response = SearchResponseBuilder.build(results, tookInMillis);
+            InternalAggregations aggregations = buildAggregations(results, request, converter);
+
+            // TODO: Build hits from HITS results
+            SearchHits hits = SearchHits.empty(true);
+
+            SearchResponseSections sections = new SearchResponseSections(
+                hits,
+                aggregations,
+                null,
+                false,
+                null,
+                null,
+                0
+            );
+
+            SearchResponse response = new SearchResponse(
+                sections,
+                null,
+                aggregations != null ? 1 : 0,
+                aggregations != null ? 1 : 0,
+                0,
+                tookInMillis,
+                ShardSearchFailure.EMPTY_ARRAY,
+                SearchResponse.Clusters.EMPTY
+            );
+
             listener.onResponse(response);
         } catch (Exception e) {
             logger.error("DSL execution failed", e);
             listener.onFailure(e);
         }
+    }
+
+    private InternalAggregations buildAggregations(
+            List<ExecutionResult> results,
+            SearchRequest request,
+            SearchSourceConverter converter) throws Exception {
+
+        List<ExecutionResult> aggResults = results.stream()
+            .filter(r -> r.getType() == QueryPlans.Type.AGGREGATION)
+            .collect(Collectors.toList());
+
+        if (aggResults.isEmpty() || request.source().aggregations() == null) {
+            return null;
+        }
+
+        AggregationResponseBuilder builder = new AggregationResponseBuilder(
+            converter.getAggregationRegistry(),
+            aggResults
+        );
+        return builder.build(request.source().aggregations().getAggregatorFactories());
     }
 
     // TODO: add multi-index support:
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/DateHistogramGrouping.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/DateHistogramGrouping.java
@@ -1,0 +1,110 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.dsl.aggregation;
+
+import org.apache.calcite.avatica.util.TimeUnit;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.opensearch.dsl.converter.ConversionException;
+import org.opensearch.search.aggregations.bucket.histogram.DateHistogramInterval;
+
+import java.util.List;
+
+public class DateHistogramGrouping implements ExpressionGrouping {
+
+    private static final String BUCKET_SUFFIX = "date_histogram_bucket";
+
+    private final String aggregationName;
+    private final String fieldName;
+    private final DateHistogramInterval calendarInterval;
+    private final DateHistogramInterval fixedInterval;
+    private final String uniqueId;
+
+    public DateHistogramGrouping(String aggregationName, String fieldName, DateHistogramInterval calendarInterval,
+                                  DateHistogramInterval fixedInterval) {
+        this.aggregationName = aggregationName;
+        this.fieldName = fieldName;
+        this.calendarInterval = calendarInterval;
+        this.fixedInterval = fixedInterval;
+
+        String intervalStr = calendarInterval != null ? calendarInterval.toString() : fixedInterval.toString();
+        String key = aggregationName + ":" + fieldName + ":" + intervalStr;
+        this.uniqueId = Integer.toHexString(key.hashCode());
+    }
+
+    @Override
+    public List<String> getFieldNames() {
+        return List.of(fieldName);
+    }
+
+    @Override
+    public String getProjectedColumnName() {
+        return aggregationName + "$$" + fieldName + "$$" + uniqueId + "$$" + BUCKET_SUFFIX;
+    }
+
+    @Override
+    public RexNode buildExpression(RelDataType inputRowType, RexBuilder builder) throws ConversionException {
+        RelDataTypeField field = inputRowType.getField(fieldName, true, false);
+        if (field == null) {
+            throw new ConversionException("Field not found: " + fieldName);
+        }
+
+        RexNode fieldRef = builder.makeInputRef(field.getType(), field.getIndex());
+
+        if (calendarInterval != null) {
+            return buildCalendarExpression(fieldRef, builder);
+        } else {
+            return buildFixedExpression(fieldRef, builder);
+        }
+    }
+
+    private RexNode buildCalendarExpression(RexNode fieldRef, RexBuilder builder) {
+        TimeUnit unit = mapToTimeUnit(calendarInterval.toString());
+        return builder.makeCall(SqlStdOperatorTable.FLOOR, fieldRef, builder.makeFlag(unit));
+    }
+
+    private RexNode buildFixedExpression(RexNode fieldRef, RexBuilder builder) {
+        long intervalMs = parseFixedInterval(fixedInterval.toString());
+        RexNode intervalLiteral = builder.makeBigintLiteral(java.math.BigDecimal.valueOf(intervalMs));
+
+        RexNode divided = builder.makeCall(SqlStdOperatorTable.DIVIDE, fieldRef, intervalLiteral);
+        RexNode floored = builder.makeCall(SqlStdOperatorTable.FLOOR, divided);
+        return builder.makeCall(SqlStdOperatorTable.MULTIPLY, floored, intervalLiteral);
+    }
+
+    private TimeUnit mapToTimeUnit(String interval) {
+        return switch (interval) {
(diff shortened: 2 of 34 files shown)


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
```diff
@@ -85,14 +90,59 @@ protected void doExecute(Task task, SearchRequest request, ActionListener<Search
...
+            SearchResponse response = new SearchResponse(
+                sections,
+                null,
+                aggregations != null ? 1 : 0,
+                aggregations != null ? 1 : 0,
+                0,
+                tookInMillis,
+                ShardSearchFailure.EMPTY_ARRAY,
+                SearchResponse.Clusters.EMPTY
+            );
```
nssuresh2007, line 117: Can't we update SearchResponseBuilder to handle this?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/CollationResolver.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.Map;
+
+/**
+ * Resolves {@link BucketOrder} objects to Calcite {@link RelFieldCollation} using
+ * the actual post-aggregation schema from the {@code LogicalAggregate} node.
+ *
+ * All field lookups are name-based against the real {@link RelDataType} —
+ * zero positional assumptions about the output schema layout.
+ */
+public final class CollationResolver {
```
nssuresh2007, line 32: Where is this class used? Could not see any usages

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/DateHistogramGrouping.java
```diff
@@ -0,0 +1,110 @@
...
+            default -> throw new IllegalArgumentException("Unsupported calendar interval: " + interval);
+        };
+    }
+
+    private long parseFixedInterval(String interval) {
+        String unit = interval.substring(interval.length() - 1);
+        long value = Long.parseLong(interval.substring(0, interval.length() - 1));
+
+        return switch (unit) {
+            case "ms" -> value;
```
nssuresh2007, line 102: As pointed out in [Automated PR review](https://github.com/opensearch-project/OpenSearch/pull/21091#issuecomment-4181101399), we are extracting single character and this case may not reach.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    private static Map<List<Object>, List<Object[]>> groupByKeys(
+            List<Object[]> rows, int keyCount, Map<String, Integer> colIndex) {
+        Map<List<Object>, List<Object[]>> grouped = new HashMap<>();
+
+        for (Object[] row : rows) {
+            List<Object> key = new ArrayList<>(keyCount);
+            for (int i = 0; i < keyCount; i++) {
+                key.add(row[i]);
```
nssuresh2007, line 224: How do we know using first element in each group by would be sufficient for uniqueness?
For eg. if `GROUP BY status_code, hostname` is the query, then using status_code as the key would not be unique and would overwrite each other right?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+        List<BucketEntry> buckets = new ArrayList<>();
+        Collection<AggregationBuilder> subAggs = translator.getSubAggregations(agg);
+
+        for (Map.Entry<List<Object>, List<Object[]>> entry : grouped.entrySet()) {
+            Map<String, Object> childFilter = new HashMap<>(parentKeyFilter);
+            for (int i = 0; i < currentGroupColumns.size(); i++) {
+                childFilter.put(currentGroupColumns.get(i), entry.getKey().get(i));
+            }
+
+            long docCount = ((Number) entry.getValue().get(0)[colIndex.get("_count")]).longValue();
```
nssuresh2007, line 166: Can we always assume that _count will be present?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+    public AggregationResponseBuilder(AggregationRegistry registry, List<ExecutionResult> aggResults) {
+        this.registry = registry;
+        this.granularityMap = new HashMap<>();
+        for (ExecutionResult result : aggResults) {
+            String key = granularityKey(result);
+            granularityMap.put(key, result);
+        }
+    }
+
+    public InternalAggregations build(Collection<AggregationBuilder> originalAggs) throws ConversionException {
```
nssuresh2007, line 51: Is there some scenarios where ordering might matter here?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+        return grouped;
+    }
+
+    private static String granularityKey(ExecutionResult result) {
+        RelNode relNode = result.getPlan().relNode();
+        if (relNode instanceof LogicalAggregate agg) {
+            int groupCount = agg.getGroupCount();
+            return result.getFieldNames().stream()
+                .limit(groupCount)
+                .collect(Collectors.joining(","));
```
nssuresh2007, line 238: If the field names are in this order `["country", "status", "_count", "avg_size"]` where group by is based on country and status, we expect this ordering since we want first two columns to be country and status as per my understanding. Is the ordering of these column names guaranteed?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+/**
+ * Converts execution results into OpenSearch InternalAggregations format.
+ * Simplified version for the new architecture using Iterable of Object arrays.
+ */
+public final class AggregationResponseBuilder {
```
nssuresh2007, line 37: nit: Please add java doc and comments explaining what each part of this code does for easier understanding

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/bucket/HistogramBucketTranslator.java
```diff
@@ -0,0 +1,108 @@
...
+public class HistogramBucketTranslator implements BucketTranslator<HistogramAggregationBuilder> {
+
+    @Override
+    public Class<HistogramAggregationBuilder> getAggregationType() {
+        return HistogramAggregationBuilder.class;
+    }
+
+    @Override
+    public GroupingInfo getGrouping(HistogramAggregationBuilder agg) {
+        return new HistogramGrouping(agg.getName(), agg.field(), agg.interval(), 0.0);
```
nssuresh2007, line 40: Why do we hardcode the offset here to 0?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/bucket/DateHistogramBucketTranslator.java
```diff
@@ -0,0 +1,117 @@
...
+    public InternalAggregation toBucketAggregation(DateHistogramAggregationBuilder agg, List<BucketEntry> buckets) {
+        List<InternalDateHistogram.Bucket> dateHistogramBuckets;
+
+        if (agg.minDocCount() == 0 && buckets.size() > 1) {
+            dateHistogramBuckets = fillEmptyBuckets(buckets, agg);
+        } else {
+            dateHistogramBuckets = convertBuckets(buckets, agg);
+        }
+
+        long minDocCount = agg.minDocCount() == 0 ? 1 : agg.minDocCount();
```
nssuresh2007, line 63: Why do we set minDocCount to 1 here?

nssuresh2007 (commented):
1. Can you also please check if we need validations to ensure that we throw error if invalid data types are passed?
2. Can you please paste the complete input and output you had tested in the Testing Details section above?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+            if (idx == null || !ComparisonUtils.valuesEqual(row[idx], entry.getValue())) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private static Map<List<Object>, List<Object[]>> groupByKeys(
+            List<Object[]> rows, int keyCount, Map<String, Integer> colIndex) {
+        Map<List<Object>, List<Object[]>> grouped = new HashMap<>();
```
vinaykpud, line 219: HashMap has no iteration order guarantee so buckets will come out in arbitrary order. We need to preserve order.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    public InternalAggregations build(Collection<AggregationBuilder> originalAggs) throws ConversionException {
+        List<InternalAggregation> aggs = buildLevel(originalAggs, new ArrayList<>(), Map.of());
+        return InternalAggregations.from(aggs);
+    }
+
+    @SuppressWarnings("unchecked")
+    private List<InternalAggregation> buildLevel(
```
vinaykpud, line 57: Can we have this Response Builder for both Metric and Aggregation (whatever we already have now) as separate PR? Because just for existing metric and term aggregation there will be some edge cases so lets not mix with new aggregation/query. Lets have a first working response builder in a separate PR.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/bucket/HistogramBucketTranslator.java
varun-st: I remember offset being categorised as P1 initially and hence, skipped the implementation for it. I have added support in the newer revision.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/DateHistogramGrouping.java
varun-st: Yes, it is a miss. I have updated it in the newer revision.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/CollationResolver.java
varun-st: This class is required for one of the follow up PRs. I'll remove it here and add it back when it is actually used. Thanks.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
varun-st: I have added this in the next revision.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
varun-st: The order does not matter for functional correctness. But, keeping it ordered might help with keep the results predictable. So, I'll update it in the next revision.

Ref - https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/AggregatorFactories.java#L385-L386.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
varun-st: I'll separate it out when I raise this PR against `main` branch.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
varun-st: We add `_count` explicitly in all aggregation calls - https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/AggregationMetadataBuilder.java#L137-L152.

I'll add a defensive check regardless.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
varun-st: Ack. I have updated this in the new revision.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
varun-st: We are using all the Group By fields here as the key (`keyCount`). So, I guess it should guarantee it. Please let me know if I am missing something.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/result/AggregationResponseBuilder.java
varun-st: Yes, it is guaranteed that the group by field names would first.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/TransportDslExecuteAction.java
varun-st: I have moved these responsibilities to SearchResponseBuilder in the newer revision.

PR #21097, opened 2026-04-03
Memoize source serialization in AbstractSearchAsyncAction

### LINKED ISSUE #19951: [BUG] High latency when fanning out large requests to many shards.
### Describe the bug

We're issuing a search request that is a ~2MB JSON (lots of filters) against a 1000-shard, 25-node OpenSearch index.  The index has 2TB at the time of issue but we're indexing and will eventually have 30TB, so reducing shard count isn't an option.  Single-shard execution is relatively quick (~500ms) but we're observing 6s latency when issuing the query against all shards.  We've set the pre_filter_shard_size and max_concurrent_shard_requests to 999999 so the fanout should be as large as possible and all shards execute in parallel. We think we've narrowed down the source of the high E2E latency to request serialization.

IIUC, request serialization occurs synchronously for every shard the request is being fanned out to.  Given a lot of the work is in serializing the search source and the source shouldn't vary between shards, does it make sense to have a caching layer to avoid duplicating this work?  If so, we'd be happy to contribute.

### Related component

Search:Performance

### To Reproduce

Create a 1000-shard cluster, issue a query ANDing a bunch of large TermInSet queries (number of characters-wise, so large terms) against it.

### Expected behavior

We expect all shard requests to be emitted at roughly the same time.

### Additional Details

**Plugins**
gcs-repository
a custom plugin with custom query types, rescorers, highlighters, and admission controller action filters.

**Screenshots**

Plot of shard parallelism based on debug logging we added for start and end times:
Image

justinborromeo-glean:
### Description

AbstractSearchAsyncAction is the coordinator-side action responsible for fanning out ShardSearchRequests to each shard. The status quo is that ShardSearchRequest objects are created for each shard then serialized and sent. We've found that on our OpenSearch 2.19.1 clusters with many shards (e.g. 1000), this synchronous source serialization for large (O(MB)) requests adds up and blows up latency. We observed latencies of 10-15s for this step.

This change adds memoization to this step. On its first serialization, the serialized source is stored. This stored source is used for all subsequent shard search requests.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/19951

### Check List


justinborromeo-glean:
Not sure if I'm missing something obvious but the Gradle check shows no test failures so not sure why it shows up as failed.

andrross:
@justinborromeo-glean The console output of the build has the failure details:

```

* What went wrong:
```

server/src/main/java/org/opensearch/action/search/SearchRequest.java
```diff
@@ -533,6 +537,8 @@ public SearchRequest searchType(String searchType) {
      */
     public SearchRequest source(SearchSourceBuilder sourceBuilder) {
         this.source = Objects.requireNonNull(sourceBuilder, "source must not be null");
```
andrross, line 539: `SearchSourceBuilder` itself is mutable. How do we guarantee there are no changes to it after memoization?

andrross: I would be all for making SearchSourceBuilder immutable by the way, but I suspect that's an enormous task...

justinborromeo-glean: Agree SSB immutability would be a big change.

I don't think there's an easy way to guarantee there are no changes to SSB after mutation.  We could introduce something like a generation that's incremented every time SSB is changed but there are a lot of places SSB is mutated and it's not feasible to instrument all the sub-builders (e.g. QueryBuilder).  Hash comparison on this path is too expensive.

Do you have any other ideas?  If that's blocking, then I don't think this change is viable.

justinborromeo-glean:
Going to close this since there isn't a path to getting this landed

PR #21106, opened 2026-04-03
Split gradle check into parallel check and bwc jobs

andrross:
Run `check` and `bwcTestSnapshots` as separate matrix entries in the gradle-check GitHub Actions workflow. Remove bwcTestSnapshots as a dependency of the check task.

### Check List

### DIFF
--- a/.github/workflows/gradle-check.yml
+++ b/.github/workflows/gradle-check.yml
@@ -65,6 +65,14 @@ jobs:
       issues: write # To create an issue if check fails on push.
     runs-on: ubuntu-latest
     timeout-minutes: 130
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - label: check
+            gradle_check_command: 'check -Dtests.coverage=true'
+          - label: bwc
+            gradle_check_command: 'bwcTestSnapshots -Dbwc.buildUnreleasedFromSource=false'
     steps:
       - name: Checkout OpenSearch repo
         uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


andrross:
@cwperks Is it possible to test this? Due to `pull_request_target` it seems to always use the version of the workflow checked into main instead of what is in the PR.

cwperks:
> @cwperks Is it possible to test this? Due to `pull_request_target` it seems to always use the version of the workflow checked into main instead of what is in the PR.

Unfortunately I don't think we have a way to do this unless we can authorize someone's fork to start the gradle check.

cwperks:
@andrross ML repo has similar issues. Maybe @ylwu-amzn or @dhrubo-os would know?

dhrubo-os:
> @andrross ML repo has similar issues. Maybe @ylwu-amzn or @dhrubo-os would know?

Yeah, we've hit the same issue in ml-commons. Unfortunately there's no clean way to test pull_request_target workflow changes from a fork PR — GitHub always uses the workflow from the base branch. In ml-commons we'
ve typically had to merge and validate on the next PR. 

@peterzhuamazon how do you test when you raise PRs in ML-Commons in CI workflow?

andrross:
> merge and validate on the next PR

@dhrubo-os YOLO!

@cwperks I suspect that's what we'll need to do here. What do you think? All the checks will (eventually) pass on this PR, it's just that it's running the old version of the workflow.

cwperks:
Yea it's ok with me. IMO it's exactly what we have version control for. 

The only other thing I would check is if we could somehow use act (https://github.com/nektos/act) and somehow get the authority to trigger Jenkins from local.

PR #21111, opened 2026-04-03
Fix ShardRoutingTests for SPLITTING state

andrross:
The testEqualsIgnoringVersion test could fail when a randomly generated shard was in SPLITTING state. Case 4 attempted to set a recovery source on the shard, but the ShardRouting constructor only allows recovery sources on UNASSIGNED or INITIALIZING shards. Add SPLITTING to the guard condition so the case is skipped.

Fixes test failures like https://build.ci.opensearch.org/job/gradle-check/73794/testReport/junit/org.opensearch.cluster.routing/ShardRoutingTests/testEqualsIgnoringVersion/

### Check List

### DIFF
--- a/server/src/test/java/org/opensearch/cluster/routing/ShardRoutingTests.java
+++ b/server/src/test/java/org/opensearch/cluster/routing/ShardRoutingTests.java
@@ -215,8 +215,8 @@ public class ShardRoutingTests extends OpenSearchTestCase {
                     }
                     break;
                 case 4:
-                    // change recovery source (only works for inactive primaries)
-                    if (otherRouting.active() || otherRouting.primary() == false) {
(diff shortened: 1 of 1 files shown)


andrross:
@vikasvb90 FYI

vikasvb90:
Thanks @andrross for fixing the test! PR suggestion by bot is actually correct. A splitting shard is supposed to be active as well. While breaking up code across multiple PRs, SPLITTING state got pushed without change in method `active`. This was there in my main code though [CP](https://github.com/vikasvb90/OpenSearch/blob/split_final/server/src/main/java/org/opensearch/cluster/routing/ShardRouting.java#L351)

vikasvb90:
@andrross  Raised this [PR](https://github.com/opensearch-project/OpenSearch/pull/21120) to fix this.

PR #21122, opened 2026-04-04
Add fips.mode configuration to control truststore check in FIPS mode for GCS repository

### LINKED ISSUE #20405: [BUG] GCS snapshot repository plugin doesn't work after FIPS mode is activated
### Describe the bug

After we upgraded to v3.4.0 we faced that snapshots are broken with GCS repository plugin:
```
response: '{"error":{"root_cause":[{"type":"repository_exception","reason":"[name-opensearch-dev] could not read repository data from index blob"}],"type":"repository_exception","reason":"[name-opensearch-dev] could not read repository data from index blob","caused_by":{"type":"i_o_exception","reason":"java.lang.IllegalStateException: FIPS mode is active but no custom truststore is configured. Please configure gcs.client.<client-name>.truststore.path and gcs.client.<client-name>.truststore.secure_password settings.","caused_by":{"type":"illegal_state_exception","reason":"FIPS mode is active but no custom truststore is configured. Please configure gcs.client.<client-name>.truststore.path and gcs.client.<client-name>.truststore.secure_password settings."}}},"status":500}',
```

### Related component

Storage:Snapshots

### To Reproduce

1. Have GCS repository configured in opensearch

### Expected behavior

We expect snapshots working

### Additional Details

**Plugins**
repository-gcs

**Additional context**

Richard-coco:
## Description
[GCS] Fix FIPS mode truststore handling without opt-out

This commit removes the previously proposed FIPS bypass switch and implements a compliant solution for default Google certificates in FIPS mode.

When running in FIPS mode (BCFIPS provider active) and no custom truststore is configured:
- The code now dynamically creates a FIPS-compliant BCFKS KeyStore
- Copies all trusted Google root certificates from the built-in Google trust store into the BCFKS container
- Maintains full FIPS compliance without any opt-out or security relaxation

The existing FIPS conversion scripts provided by OpenSearch are intended for user-managed certificates and JVM trust stores, not for the bundled Google certificates inside the GCS client library. Runtime conversion to BCFKS is the cleanest, most user-friendly approach to maintain out-of-the-box functionality.

This change ensures the GCS plugin works correctly in FIPS mode while adhering to all security requirements.

## Related Issues
Fixes #20405


Richard-coco:
This change adds a fips.mode setting (default: true) to control strict FIPS truststore validation behavior.
The default value preserves existing security behavior.
The setting is only for operational flexibility in regulated environments.
This is a safe, backward-compatible change and requires a security review approval.

reta:
> This change adds a fips.mode setting (default: true) to control strict FIPS truststore validation behavior. The default value preserves existing security behavior. The setting is only for operational flexibility in regulated environments. This is a safe, backward-compatible change and requires a security review approval.

Thanks @Richard-coco , I am not entirely onboard with this change. The whole idea behind enabling FIPS is to make sure we could harden the distribution and all its components. Adding such ad-hoc opt-outs calls for issues, if the distribution or some of its components could not be run in the FIPS mode, well, the distribution has to turn it off. For operational flexibility, migrating to FIPS compatible trust store would be preferred (I believe we do provide some scripts already).

Richard-coco:
@reta Thanks for your guidance and feedback.

I've removed the FIPS bypass switch and updated the code to use a proper FIPS-compliant BCFKS trust store for Google's default certificates.

The scripts you mentioned work great for user-managed certificates, but they don't apply to the built-in Google certificates that come bundled with the GCS library.
To keep things simple for users, this change creates the BCFKS trust store dynamically at runtime, which meets FIPS requirements while maintaining the out-of-the-box experience.

reta:
Thanks @Richard-coco , would you be up adding test case to `GoogleCloudStorageServiceFipsTests` for this particular configuration? (no trust store, FIPS mode) Thank you.

Richard-coco:
@reta Sure, the corresponding test case for no trust store in FIPS mode has been added and pushed.
Thank you for the review!

Richard-coco:
@reta The gradle-check failed on the unrelated arrow-flight-rpc plugin test, which is a flaky test unrelated to my GCS plugin changes. Could you please re-run the CI for me? Thanks a lot!

PR #21123, opened 2026-04-04
Fix #20612: validate ambiguous empty _source.include with exclude

### LINKED ISSUE #20612: [Feature Request] [BUG] `_source` include / exclude validation

Richard-coco:
### Description
This PR adds validation to reject the ambiguous combination of empty _source.include and non-empty _source.exclude configuration.
When users submit a request with "include": [] and "exclude": ["xxx"], it creates semantic ambiguity, so this change throws an IllegalArgumentException to clarify usage.

### Related Issues
Resolves #20612

### Check List

### DIFF
--- a/server/src/test/java/org/opensearch/search/fetch/subphase/FetchSourcePhaseTests.java
+++ b/server/src/test/java/org/opensearch/search/fetch/subphase/FetchSourcePhaseTests.java
@@ -61,7 +61,11 @@ public class FetchSourcePhaseTests extends OpenSearchTestCase {
     }
 
     public void testBasicFiltering() throws IOException {
-        XContentBuilder source = XContentFactory.jsonBuilder().startObject().field("field1", "value").field("field2", "value2").endObject();
+        XContentBuilder source = XContentFactory.jsonBuilder()
+            .startObject()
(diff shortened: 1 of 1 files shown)


Richard-coco:
Closing this PR as @urmichm has already started working on the core validation logic for this issue (#20612). I appreciate the discussion and will focus on other unaddressed issues in the project. Thanks!

Issue #21117, opened 2026-04-04
[BUG] Maven Snapshot Publication Failing


cwperks:
### Describe the bug

Maven snapshot publication has been failing since https://github.com/opensearch-project/OpenSearch/pull/21057 was merged into core. 

See full failure here: https://github.com/opensearch-project/OpenSearch/actions/runs/23969727021/job/69916495842

```
Caused by:
  process didn't exit successfully: `/home/runner/work/OpenSearch/OpenSearch/sandbox/plugins/analytics-backend-datafusion/target/release/build/substrait-859e6d0cbc9375a9/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=FORCE_REBUILD
  cargo:rerun-if-changed=substrait
  cargo:rerun-if-changed=substrait/text/dialect_schema.yaml
  cargo:rerun-if-changed=substrait/text/simple_extensions_schema.yaml
  cargo:rerun-if-changed=substrait/proto/substrait/parameterized_types.proto
  cargo:rerun-if-changed=substrait/proto/substrait/extensions/extensions.proto
  cargo:rerun-if-changed=substrait/proto/substrait/type.proto
  cargo:rerun-if-changed=substrait/proto/substrait/capabilities.proto
  cargo:rerun-if-changed=substrait/proto/substrait/algebra.proto
  cargo:rerun-if-changed=substrait/proto/substrait/extended_expression.proto
  cargo:rerun-if-changed=substrait/proto/substrait/function.proto
  cargo:rerun-if-changed=substrait/proto/substrait/type_expressions.proto
  cargo:rerun-if-changed=substrait/proto/substrait/plan.proto

  --- stderr
  Error: Custom { kind: NotFound, error: "Could not find `protoc`. If `protoc` is installed, try setting the `PROTOC` environment variable to the path of the `protoc` binary. To install it on Debian, run `apt-get install protobuf-compiler`. It is also available at https://github.com/protocolbuffers/protobuf/releases  For more information: https://docs.rs/prost-build/#sourcing-protoc" }
warning: build failed, waiting for other jobs to finish...

> Task :sandbox:plugins:analytics-backend-datafusion:buildRustLibrary FAILED


[Incubating] Problems report is available at: file:///home/runner/work/OpenSearch/OpenSearch/build/reports/problems/problems-report.html
FAILURE: Build failed with an exception.
1187 actionable tasks: 1186 executed, 1 from cache

* What went wrong:
[...]

PR #21125, opened 2026-04-04
Disable sandbox by default for publication, require passing -Dsandbox.enabled=true

cwperks:
### Description

This PR solves 2 issues:

- Changes the default value of `sandbox.enabled` to false to ensure its intentional for anyone to build or run with sandbox.
- Ensures that publication tasks for sandbox plugins only run when `sandbox.enabled` is set to true

This is related to the introduction of the sandbox datafusion plugin which requires rust + cargo to assemble and run. Without this change, all devs will now need to ensure they have the new pre-requisites on their machines to build and run the core repo even if not using the sandbox plugin.

### Related Issues

Related to conversation on https://github.com/opensearch-project/OpenSearch/issues/21117

### Check List


bharath-techie:
Hi @cwperks ,
One challenge I see here is that this parameter should be set to true for all the gradle checks and CI. Some of these are in `.github` workflows , not sure if we will need more changes outside this for CI etc.

Since we'll move these plugins eventually out of sandbox, and the only thing we are saving is `protoc, rust` prerequisite installation for the time being, wondering if we should do this change or not.

Tagging @Bukhtawar @peterzhuamazon as well.

Bukhtawar:
> Hi @cwperks , One challenge I see here is that this parameter should be set to true for all the gradle checks and CI. Some of these are in `.github` workflows , not sure if we will need more changes outside this for CI etc.
> 
> Since we'll move these plugins eventually out of sandbox, and the only thing we are saving is `protoc, rust` prerequisite installation for the time being, wondering if we should do this change or not.
> 
> Tagging @Bukhtawar @peterzhuamazon as well.

I believe this is a one-time installation for builds so I am hoping it doesn't hurt much as long as we can have the build scripts install it while adding a clear pre-req in the development guide?

cwperks:
Ack I will close this. btw, we are working on splitting the core gradle check to reduce the time to completion for all tests (see https://github.com/opensearch-project/OpenSearch/pull/21106). I think it would be a logical split to run the sandbox tests as their own CI check. Have you experienced any issues while developing on the datafusion feature branch wrt getting feedback from the gradle check?

Bukhtawar:
+1 to have independency CI checks for sandbox

PR #21133, opened 2026-04-06
Add `upgrade` param to force merge API

### LINKED ISSUE #21132: [Feature Request] Use forceMerge API to upgrade older Lucene segments to the latest

When upgrading across multiple major versions of OpenSearch (1.x), we found an index having shards with multiple segment files from prior Lucene versions. 
While an `_upgrade` API does exist, it is not documented (like in Elasticsearch), and only a passing mentioned of it is made in the project's Rust client  (Ref: https://docs.rs/opensearch/latest/opensearch/indices/struct.IndicesUpgrade.html )
```
The _upgrade API is no longer useful and will be removed. Instead, see _reindex API.
```
But to have users perform a reindex, which is very inconvenient... especially if the index is not managed behind an alias. 

A workaround is to force-merge old segments with the new Lucene engine. This works great if you are well short of your max segment size (typically 5GB).

In this particular case, we already had older segments at 5GB and the only solution was to raise the `index.merge.policy.max_merged_segment` to 10GB. This increased downtime and storage demands and doesn't fix the problem in the longterm where this would then likely be doubled to 20GB.

Since the `_upgrade` API uses the same components as the `_forcemerge` API, we should consider using the `_forcemerge` API and provide an additional request parameter that can serve this purpose. 
It could be on the lines of discarding segments with only deleted documents (i.e.  `only_expunge_deletes`) and be very specific -- `only_upgrade_lucene`

### Related component

Indexing

_No response_

### Additional context

_No response_

kkhatua:
### Description
Exposes the existing Lucene segment upgrade capability through the _forcemerge REST endpoint via a new only_upgrade_lucene boolean query parameter. When set to true, old Lucene segments are rewritten to the latest version without consolidating segment count.

This connects the same upgrade path that the legacy _upgrade API uses, by calling engine.forceMerge with upgrade=true and maxNumSegments=MAX_VALUE. Since the _upgrade API seems to be deprecated (no documentation), using the forcemerge API seems to be the right approach as it does similar steps with `only_expunge_deletes` parameter. Also, introducing as a query parameter makes it easier to backport if desired.

In addition, the API will reject any conflicting combinations with `max_num_segments` and `only_expunge_deletes` parameters. Added tests to validate this restriction.

### Related Issues
Resolves #21132 

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/action/admin/indices/forcemerge/ForceMergeRequest.java
+++ b/server/src/main/java/org/opensearch/action/admin/indices/forcemerge/ForceMergeRequest.java
@@ -72,12 +72,14 @@ public class ForceMergeRequest extends BroadcastRequest<ForceMergeRequest> {
         public static final boolean ONLY_EXPUNGE_DELETES = false;
         public static final boolean FLUSH = true;
         public static final boolean PRIMARY_ONLY = false;
+        public static final boolean ONLY_UPGRADE_LUCENE = false;
     }
 
     private int maxNumSegments = Defaults.MAX_NUM_SEGMENTS;
     private boolean onlyExpungeDeletes = Defaults.ONLY_EXPUNGE_DELETES;
     private boolean flush = Defaults.FLUSH;
     private boolean primaryOnly = Defaults.PRIMARY_ONLY;
+    private boolean onlyUpgradeLucene = Defaults.ONLY_UPGRADE_LUCENE;
 
     private static final Version FORCE_MERGE_UUID_VERSION = Version.V_3_0_0;
 
@@ -114,6 +116,7 @@ public class ForceMergeRequest extends BroadcastRequest<ForceMergeRequest> {
                 "As of legacy version 7.7 [" + Engine.FORCE_MERGE_UUID_KEY + "] is no longer optional in force merge requests."
             );
         }
+        onlyUpgradeLucene = in.getVersion().onOrAfter(Version.V_3_7_0) ? in.readBoolean() : Defaults.ONLY_UPGRADE_LUCENE;
     }
 
     /**
@@ -189,6 +192,23 @@ public class ForceMergeRequest extends BroadcastRequest<ForceMergeRequest> {
     }
 
     /**
+     * Should the force merge only upgrade Lucene segments written by an older version to the latest Lucene version.
+     * When set to {@code true}, no actual segment merging is performed — only segments with an older Lucene version
+     * are rewritten. Defaults to {@code false}.
+     */
+    public boolean onlyUpgradeLucene() {
+        return onlyUpgradeLucene;
+    }
+
+    /**
+     * See {@link #onlyUpgradeLucene()}.
+     */
+    public ForceMergeRequest onlyUpgradeLucene(boolean onlyUpgradeLucene) {
+        this.onlyUpgradeLucene = onlyUpgradeLucene;
+        return this;
+    }
+
+    /**
      * Should this task store its result after it has finished?
      */
     public void setShouldStoreResult(boolean shouldStoreResult) {
@@ -212,6 +232,8 @@ public class ForceMergeRequest extends BroadcastRequest<ForceMergeRequest> {
             + flush
             + "], primaryOnly["
             + primaryOnly
+            + "], onlyUpgradeLucene["
+            + onlyUpgradeLucene
             + "]";
     }
 
@@ -229,6 +251,9 @@ public class ForceMergeRequest extends BroadcastRequest<ForceMergeRequest> {
         } else {
             out.writeOptionalString(forceMergeUUID);
         }
+        if (out.getVersion().onOrAfter(Version.V_3_7_0)) {
+            out.writeBoolean(onlyUpgradeLucene);
+        }
     }
 
     @Override
@@ -242,6 +267,8 @@ public class ForceMergeRequest extends BroadcastRequest<ForceMergeRequest> {
             + flush
             + ", primaryOnly="
             + primaryOnly
+            + ", onlyUpgradeLucene="
+            + onlyUpgradeLucene
             + '}';
     }
 }
--- a/server/src/main/java/org/opensearch/rest/action/admin/indices/RestForceMergeAction.java
+++ b/server/src/main/java/org/opensearch/rest/action/admin/indices/RestForceMergeAction.java
@@ -77,11 +77,20 @@ public class RestForceMergeAction extends BaseRestHandler {
         mergeRequest.onlyExpungeDeletes(request.paramAsBoolean("only_expunge_deletes", mergeRequest.onlyExpungeDeletes()));
         mergeRequest.flush(request.paramAsBoolean("flush", mergeRequest.flush()));
         mergeRequest.primaryOnly(request.paramAsBoolean("primary_only", mergeRequest.primaryOnly()));
+        mergeRequest.onlyUpgradeLucene(request.paramAsBoolean("only_upgrade_lucene", mergeRequest.onlyUpgradeLucene()));
         if (mergeRequest.onlyExpungeDeletes() && mergeRequest.maxNumSegments() != ForceMergeRequest.Defaults.MAX_NUM_SEGMENTS) {
             deprecationLogger.deprecate(
                 "force_merge_expunge_deletes_and_max_num_segments_deprecation",
                 "setting only_expunge_deletes and max_num_segments at the same time is deprecated and will be rejected in a future version"
             );
+        }
+        if (mergeRequest.onlyUpgradeLucene()) {
+            if (mergeRequest.maxNumSegments() != ForceMergeRequest.Defaults.MAX_NUM_SEGMENTS) {
+                throw new IllegalArgumentException("cannot set only_upgrade_lucene and max_num_segments at the same time");
+            }
+            if (mergeRequest.onlyExpungeDeletes()) {
+                throw new IllegalArgumentException("cannot set only_upgrade_lucene and only_expunge_deletes at the same time");
+            }
         }
         if (request.paramAsBoolean("wait_for_completion", true)) {
             return channel -> client.admin().indices().forceMerge(mergeRequest, new RestToXContentListener<>(channel));
--- a/server/src/main/java/org/opensearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/opensearch/index/shard/IndexShard.java
@@ -1726,14 +1726,18 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
             logger.trace("force merge with {}", forceMerge);
         }
         Indexer engine = getIndexer();
-        engine.forceMerge(
-            forceMerge.flush(),
-            forceMerge.maxNumSegments(),
-            forceMerge.onlyExpungeDeletes(),
-            false,
-            false,
-            forceMerge.forceMergeUUID()
-        );
+        if (forceMerge.onlyUpgradeLucene()) {
+            engine.forceMerge(forceMerge.flush(), Integer.MAX_VALUE, false, true, false, forceMerge.forceMergeUUID());
+        } else {
+            engine.forceMerge(
+                forceMerge.flush(),
+                forceMerge.maxNumSegments(),
+                forceMerge.onlyExpungeDeletes(),
+                false,
+                false,
+                forceMerge.forceMergeUUID()
+            );
+        }
     }
 
     /**
--- a/server/src/main/java/org/opensearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/opensearch/index/shard/IndexShard.java
@@ -1729,14 +1729,18 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
             logger.trace("force merge with {}", forceMerge);
         }
         Indexer engine = getIndexer();
-        engine.forceMerge(
-            forceMerge.flush(),
-            forceMerge.maxNumSegments(),
-            forceMerge.onlyExpungeDeletes(),
-            false,
-            false,
-            forceMerge.forceMergeUUID()
-        );
+        if (forceMerge.onlyUpgradeLucene()) {
+            engine.forceMerge(forceMerge.flush(), Integer.MAX_VALUE, false, true, false, forceMerge.forceMergeUUID());
+        } else {
+            engine.forceMerge(
+                forceMerge.flush(),
+                forceMerge.maxNumSegments(),
+                forceMerge.onlyExpungeDeletes(),
+                false,
+                false,
+                forceMerge.forceMergeUUID()
+            );
+        }
     }
 
     /**
--- a/server/src/test/java/org/opensearch/action/admin/indices/forcemerge/ForceMergeRequestTests.java
+++ b/server/src/test/java/org/opensearch/action/admin/indices/forcemerge/ForceMergeRequestTests.java
@@ -44,13 +44,13 @@ public class ForceMergeRequestTests extends OpenSearchTestCase {
     public void testDescription() {
         ForceMergeRequest request = new ForceMergeRequest();
         assertEquals(
-            "Force-merge indices [], maxSegments[-1], onlyExpungeDeletes[false], flush[true], primaryOnly[false]",
+            "Force-merge indices [], maxSegments[-1], onlyExpungeDeletes[false], flush[true], primaryOnly[false], onlyUpgradeLucene[false]",
             request.getDescription()
         );
 
         request = new ForceMergeRequest("shop", "blog");
         assertEquals(
-            "Force-merge indices [shop, blog], maxSegments[-1], onlyExpungeDeletes[false], flush[true], primaryOnly[false]",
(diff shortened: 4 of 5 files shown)


kkhatua:
Testing performed with upgrade of segments from OpenSearch 2.19.5
```
# Startup 2.19.5 Cluster
export OPENSEARCH_JAVA_HOME=$(/usr/libexec/java_home -v 21) && ./bin/opensearch -d -p pid

# Created 3 indices with 10 documents and delete none, 1 and 10 documents from them respectively
 ./ingest_docs.sh dummy-.... 

# Ingesting 7 docs into `dummy-10docs-10deletes` as it has 0 docs 
 ./ingest_docs.sh dummy-10docs-10deletes 7

# Verifying indices & segments
[kkhatua@ github/OpenSearch] curl -X GET "localhost:9200/_cat/indices/dummy*?v"
health status index                        uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   dummy-10docs-10deletes       cuB5LOFqQZ-U9QDjFxG-WA   1   0          7           20     45.4kb         45.4kb
green  open   dummy-10docs-0deletes        n4TbC6MJQ5CEn0SXAplF6Q   1   0         10            0       20kb           20kb
green  open   dummy-10docs-1deletes        h9ZM8z_PRBC8JsZZ6p9Kew   1   0          9            2     27.2kb         27.2kb
[kkhatua@ github/OpenSearch] curl 'http://localhost:9200/_cat/segments/dummy*?v'
index                  shard prirep ip        segment generation docs.count docs.deleted   size size.memory committed searchable version compound
dummy-10docs-10deletes 0     p      127.0.0.1 _0               0          0           10 23.8kb           0 false     true       9.12.3  true
dummy-10docs-10deletes 0     p      127.0.0.1 _1               1          0           10  3.5kb           0 false     true       9.12.3  true
dummy-10docs-10deletes 0     p      127.0.0.1 _2               2          7            0 17.9kb           0 false     true       9.12.3  true
dummy-10docs-0deletes  0     p      127.0.0.1 _0               0         10            0 19.8kb           0 false     true       9.12.3  true
dummy-10docs-1deletes  0     p      127.0.0.1 _0               0          9            1 23.8kb           0 false     true       9.12.3  true
dummy-10docs-1deletes  0     p      127.0.0.1 _1               1          0            1  3.1kb           0 false     true       9.12.3  true


# Terminated 2.19.5 
[kkhatua@ testbed/opensearch-2.19.5] kill $(cat pid)

# Spinning up the custom 3.7.0-SNAPSHOT that uses Lucene 10.4.0 engine
JAVA_HOME=/Users/kkhatua/Library/Java/JavaVirtualMachines/corretto-24.0.2/Contents/Home ./gradlew run --data-dir=/Users/kkhatua/labratory/testbed/opensearch-2.19.5/data

# Verifying
[kkhatua@ github/OpenSearch] curl -X GET "localhost:9200"
{
  "name" : "runTask-0",
  "cluster_name" : "runTask",
  "cluster_uuid" : "YU3KNRfuSUu9K3w6r9szEQ",
  "version" : {
    "distribution" : "opensearch",
    "number" : "3.7.0-SNAPSHOT",
    "build_type" : "tar",
    "build_hash" : "70013743dda121b6efe3f8285890719694b870b4",
    "build_date" : "2026-04-05T06:19:45.059138Z",
    "build_snapshot" : true,
    "lucene_version" : "10.4.0",
    "minimum_wire_compatibility_version" : "2.19.0",
    "minimum_index_compatibility_version" : "2.0.0"
  },
  "tagline" : "The OpenSearch Project: https://opensearch.org/"
}
[kkhatua@ github/OpenSearch] curl -X GET "localhost:9200/_cat/segments/dummy*?v"
index                  shard prirep ip        segment generation docs.count docs.deleted   size size.memory committed searchable version compound
dummy-10docs-10deletes 0     p      127.0.0.1 _2               2          7            0 17.9kb           0 true      true       9.12.3  true
dummy-10docs-0deletes  0     p      127.0.0.1 _0               0         10            0 19.8kb           0 true      true       9.12.3  true
dummy-10docs-1deletes  0     p      127.0.0.1 _0               0          9            1 23.8kb           0 true      true       9.12.3  true

# Testing forbidden combination of simultaneous expunge and upgrade 
[kkhatua@ github/OpenSearch] curl -X POST "localhost:9200/dummy*/_forcemerge?only_expunge_deletes&only_upgrade_lucene"
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"cannot set only_upgrade_lucene and only_expunge_deletes at the same time"}],"type":"illegal_argument_exception","reason":"cannot set only_upgrade_lucene and only_expunge_deletes at the same time"},"status":400}%

# Testing expunging deletes
[kkhatua@ github/OpenSearch] curl -X POST "localhost:9200/dummy*/_forcemerge?only_expunge_deletes"
{"_shards":{"total":3,"successful":3,"failed":0}}%

# Verifying
[kkhatua@ github/OpenSearch] curl -X GET "localhost:9200/_cat/segments/dummy*?v"
index                  shard prirep ip        segment generation docs.count docs.deleted   size size.memory committed searchable version compound
dummy-10docs-10deletes 0     p      127.0.0.1 _2               2          7            0 17.9kb           0 true      true       9.12.3  true
dummy-10docs-0deletes  0     p      127.0.0.1 _0               0         10            0 19.8kb           0 true      true       9.12.3  true
dummy-10docs-1deletes  0     p      127.0.0.1 _0               0          9            1 23.8kb           0 true      true       9.12.3  true

# Testing only_upgrade_lucene
[kkhatua@ github/OpenSearch] curl -X POST "localhost:9200/dummy*/_forcemerge?only_upgrade_lucene="
{"_shards":{"total":3,"successful":3,"failed":0}}%

# Verifying indices and segments
[kkhatua@ github/OpenSearch] curl -X GET "localhost:9200/_cat/indices/dummy*?v"
health status index                  uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   dummy-10docs-10deletes 5ROAHkbiQt-OQVCSO08Jzg   1   0         14            0     39.8kb         39.8kb
green  open   dummy-10docs-1deletes  AZi4VzpxS2SiLE4oRxWFyw   1   0         18            0     48.2kb         48.2kb
green  open   dummy-10docs-0deletes  zqX1_DcJREawijEojr775Q   1   0         20            0     44.4kb         44.4kb

[kkhatua@ github/OpenSearch] curl -X GET "localhost:9200/_cat/segments/dummy*?v"
index                  shard prirep ip        segment generation docs.count docs.deleted   size size.memory committed searchable version compound
dummy-10docs-10deletes 0     p      127.0.0.1 _2               2          7            0 17.9kb           0 false     true       9.12.3  true
dummy-10docs-10deletes 0     p      127.0.0.1 _3               3          7            0 17.3kb           0 true      false      10.4.0  false
dummy-10docs-0deletes  0     p      127.0.0.1 _0               0         10            0 19.8kb           0 false     true       9.12.3  true
dummy-10docs-0deletes  0     p      127.0.0.1 _1               1         10            0 19.1kb           0 true      false      10.4.0  false
dummy-10docs-1deletes  0     p      127.0.0.1 _0               0          9            1 23.8kb           0 false     true       9.12.3  true
dummy-10docs-1deletes  0     p      127.0.0.1 _2               2          9            0 18.7kb           0 true      false      10.4.0  false

# After running searches against ALL the dummy* indices to force closing references on any active `IndexReader`s 
[kkhatua@ github/OpenSearch] curl -X GET "localhost:9200/_cat/segments/dummy*?v"
index                  shard prirep ip        segment generation docs.count docs.deleted   size size.memory committed searchable version compound
dummy-10docs-10deletes 0     p      127.0.0.1 _3               3          7            0 17.3kb           0 true      true       10.4.0  false
dummy-10docs-0deletes  0     p      127.0.0.1 _1               1         10            0 19.1kb           0 true      true       10.4.0  false
dummy-10docs-1deletes  0     p      127.0.0.1 _2               2          9            0 18.7kb           0 true      true       10.4.0  false
```

andrross:
@kkhatua Why does it have to be "only_upgrade_lucene"? Can it just be an optional "upgrade_lucene" boolean? If you wanted the "only" behavior you can specific a huge number of segments but that wouldn't be a requirement.

It may need to be mutually exclusive with "only_expunge_deletes", but that option already has weird semantics.

kkhatua:
> @kkhatua Why does it have to be "only_upgrade_lucene"? Can it just be an optional "upgrade_lucene" boolean? If you wanted the "only" behavior you can specific a huge number of segments but that wouldn't be a requirement.
> 
> It may need to be mutually exclusive with "only_expunge_deletes", but that option already has weird semantics.

You hit the nail on the head ! 😄 

The `only_` prefix is precisely for indicating the mutually exclusive behavior. We don't always pick the most conservative parameter names (e.g `cancel_after_time_interval`) , but in this case I used the `only_expunge_deletes` as a precedence.

andrross:
> > @kkhatua Why does it have to be "only_upgrade_lucene"? Can it just be an optional "upgrade_lucene" boolean? If you wanted the "only" behavior you can specific a huge number of segments but that wouldn't be a requirement.
> 
> > 
> 
> > It may need to be mutually exclusive with "only_expunge_deletes", but that option already has weird semantics.
> 
> 
> 
> You hit the nail on the head ! 😄 
> 
> 
> 
> The `only_` prefix is precisely for indicating the mutually exclusive behavior. We don't always pick the most conservative parameter names (e.g `cancel_after_time_interval`) , but in this case I used the `only_expunge_deletes` as a precedence. 

@kkhatua But I don't think this needs mutually exclusive behavior, does it? `only_expunge_deletes` will need to continue to be it's weird mutually-exclusive self, but `upgrade_lucene` can just be another option in force merge right?

kkhatua:
True... I'm enforcing exclusivity with `max_num_segments` and `only_expunge_deletes` because both of these will itself rewrite new segments  only under favorable conditions of target segments not exceeding maxSize when merging, and segments with deleted docs. It can appear confusing if segments are not "upgraded" to the new version.

The key problem we are trying to solve is for maxSized segments being written into the new formats which otherwise would have been done untouched by the periodic background merges. I want to avoid the problem of having the force segments to be artificually merged into larger and larger segment sizes just for sake of version currency.

sgup432:
@kkhatua 
> Since the _upgrade API seems to be deprecated (no documentation)

I don't think this is right? The `_upgrade` API is still available to use and not deprecated. I checked `RestUpgradeAction` and don't see any `@deprecation` tags there.

So do we still need this change? IMO we don't need this. We can simply use `POST /{index}/_upgrade?only_ancient_segments=false`. Can you verify this against the same test scenarios you covered earlier?

Plus the `_upgrade` API also handles the flow(see `updateSettings()` in `TransportUpgradeAction`) where we update the cluster state to record the new version info. And I don't see this PR handles that. There might be other stuff which we might be missing.

kkhatua:
@sgup432 
The upgrade API doesn't have any documentation and Elastic itself has moved to what they call an Upgrade Assistant with Elasticsearch 8.x [Ref](https://www.elastic.co/docs/deploy-manage/upgrade/prepare-to-upgrade/upgrade-assistant) .  Given that all the primitives needed to explicitly manage segments exists in the forceMerge API, this should be incorporated into that rather than maintaining a second endpoint that is redundant. Index Managment plugin can be used by users to manage version currency.

Also, we don't need to update the cluster state to record the new version info. Keep in mind that the functionality and behavior is identical to other operations like `only_expunge_deletes` and _**none**_ of these require to update the cluster state. So, I don't think we're missing anything here. The upgrade steps I tested above show the versions correctly tracked at the segment level. Also, from an overhead perspective,  the update settings will trigger a cluster state update task, which isn't needed IMO.

andrross:
> Given that all the primitives needed to explicitly manage segments exists in the forceMerge API, this should be incorporated into that rather than maintaining a second endpoint that is redundant.

@kkhatua I'm not sure a second endpoint is redundant. You're shimming in a semantically separate operation into force merge (the same thing that `only_expunge_deletes` did). This will result in three separate APIs that are the same in name only.

andrross:
> The key problem we are trying to solve is for maxSized segments being written into the new formats which otherwise would have been done untouched by the periodic background merges. I want to avoid the problem of having the force segments to be artificually merged into larger and larger segment sizes just for sake of version currency.

@kkhatua The behavior I would expect is that if you provide the `upgrade=true` parameter but the segments would otherwise not be modified by the merge policy, then the existing segments would be re-written as-is just with the upgrade process done to them. In other words, regardless of whether you called `_forcemerge` and `_forcemerge?upgrade=true`, you would be guaranteed to get the same resulting segment topology, it's just with `upgrade=true` you'd be guaranteed that every segment has been upgraded. Is that not possible?

andrross:
To be clear on my thinking here, I think either of the following two options would be okay:

- Add `upgrade=true` parameter to _forcemerge. If specified it still operates according to the merge policy and the provided max number of segments, but also guarantees that all segments are upgraded, even if some of them are not actually merged into larger segments (those segments would just be re-written one-for-one).
- Make the `_upgrade` API work as expected and ensure it is properly documented/tested/supported.

kkhatua:
@andrross  
Looked at the code more closely and I'm in agreement with your point. The primary motivation was that segments hitting the max segment size for merges would stay in the old format as the engine upgraded, unless you forced a resize to a large segment. This can hurt performance, e.g. with concurrent segment search. But it appears I was wrong to assume that there is absolutely no way of forcing a segment to upgrade and outgrow the size simultaneously when `maxNumSegments` can be tailored to just match the requisite number (though it isn't guaranteed). 
Hence, the feature still has value and of the 2 options, the `upgrade=true` option is preferable IMO.

Summarzing the changes:
1. Renamed the parameter `only_upgrade_lucene` to **`upgrade`** and made it an orthogonal modifier rather than an exclusive mode that it previously was (i.e. we can now use it in combination with other options).
2. It now **composes with `max_num_segments`** instead of rejecting it:
     - `upgrade=true` alone rewrites each old-version segment one-for-one and otherwise preserves the segment topology — same result as a plain `_forcemerge`, just with every segment guaranteed on the current Lucene version (this is the behavior you described).
3. `upgrade=true&max_num_segments=N` upgrades first, then consolidates to `≤ N`.
4. The only remaining restriction is `upgrade` + `only_expunge_deletes`, which is rejected because expunge-deletes takes the `forceMergeDeletes` engine path that doesn't upgrade.

server/src/main/java/org/opensearch/action/admin/indices/forcemerge/ForceMergeRequest.java
```diff
@@ -188,6 +191,25 @@ public ForceMergeRequest primaryOnly(boolean primaryOnly) {
...
     }
 
+    /**
+     * Should the force merge upgrade Lucene segments written by an older Lucene version to the current
+     * Lucene version. This composes with {@link #maxNumSegments(int)}: when {@code max_num_segments} is
+     * not set, every old-version segment is rewritten one-for-one to the current version and the segment
+     * topology is otherwise preserved (no consolidation). When {@code max_num_segments} is set, segments
+     * are upgraded and then merged down to at most that many segments. Defaults to {@code false}.
+     */
+    public boolean upgrade() {
```
andrross, line 204: There's a builder class (`ForceMergeRequestBuilder`) that should be updated too for completeness.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestForceMergeAction.java
```diff
@@ -77,12 +77,19 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
                 "force_merge_expunge_deletes_and_max_num_segments_deprecation",
                 "setting only_expunge_deletes and max_num_segments at the same time is deprecated and will be rejected in a future version"
             );
         }
+        // "upgrade" composes with "max_num_segments" (upgrade segments, then optionally consolidate),
+        // but it is incompatible with "only_expunge_deletes": the engine only upgrades segments on the
+        // forceMerge path, whereas only_expunge_deletes takes the forceMergeDeletes path.
+        if (mergeRequest.upgrade() && mergeRequest.onlyExpungeDeletes()) {
+            throw new IllegalArgumentException("cannot set upgrade and only_expunge_deletes at the same time");
+        }
```
andrross: Should this validation be in `ForceMergeRequest.validate()` (inherited from the parent ActionRequest class)? As this code is written, a plugin that invoked a force merge via the transport request would never hit this validation and would be allowed to specify both upgrade and only expunge deletes.

server/src/main/java/org/opensearch/action/admin/indices/forcemerge/ForceMergeRequest.java
kkhatua: Agree... added `ForceMergeRequestBuilder#setUpgrade(boolean)`  and tests around it.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestForceMergeAction.java
kkhatua: Moved the `upgrade` and  `only_expunge_deletes` check into `ForceMergeRequest#validate()` and removed the `IllegalArgumentException` i was throwing for the combination, so transport/plugin callers are covered too.

PR #21142, opened 2026-04-06
Added support for search_after pagination using LogicalFilter

abhishek00159:
### Summary

This PR implements search_after pagination support in the DSL query executor plugin by converting pagination parameters into Calcite LogicalFilter predicates. This enables efficient
cursor-based pagination for large result sets without the performance overhead of deep pagination using from/size.
* In Lucene-based systems (like OpenSearch), search_after is applied during result collection, not after a full sort. Internally, it uses a Top-K selection algorithm backed by a priority queue (heap), which avoids sorting the entire dataset. This leads to a time complexity of approximately: O(N · log K), where N = total number of matching documents, K = page size (number of results requested).

* In contrast, a Apache Calcite-based implementation models search_after as a filter condition followed by sorting. The execution flow becomes: Filter → Sort → Limit. This reduces the dataset first (from N → M), but then performs a full sort on the filtered results. The resulting time complexity is approximately: O(N + M · log M) ≈ O(M log M), where: M = number of rows after filtering. Further optimizations may apply if indexes are available, but in general, this approach still requires sorting the reduced dataset.

Overall, we can expect some performance difference with the Calcite implementation, and there isn’t a direct alternative approach to implement it. However, the expected performance impact is likely to be minimal.

### Problem Statement

The DSL query executor previously lacked support for search_after pagination, limiting its ability to efficiently paginate through large datasets. The from/size approach becomes 
increasingly expensive for deep pagination as it requires skipping all previous results.

### Solution

Enhanced the FilterConverter to translate search_after values into equivalent filter conditions based on the sort order. The converter now:

1. Validates pagination requirements: Ensures search_after is used with sort fields and that value counts match
2. Generates filter predicates: Converts pagination values into comparison operators (>, <) based on sort direction
3. Handles multi-field sorting: Creates OR conditions with proper precedence for complex sort orders
4. Combines with existing queries: Merges search_after filters with existing query conditions using AND logic

### Technical Details

#### FilterConverter Logic

For a single sort field with ASC order:
search_after: [100]
sort: price ASC                                                                                                                                                                      
→ Filter: price > 100                                                                                                                                                                
                                                                                                                                                                                     

For multiple sort fields:
search_after: ["brandX", 1200]
sort: brand ASC, price DESC                                                                                                                                                          
→ Filter: (brand > "brandX") OR (brand = "brandX" AND price < 1200)                                                                                                                  
                                                                                                                                                                                     

This approach ensures correct pagination semantics where results continue from the last seen document according to the sort order.

#### Key Implementation Points

• **Operator selection**: ASC sorts use >, DESC sorts use <
• **Multi-field precedence**: Earlier sort fields take precedence with equality checks cascading to later fields
• **Type safety**: Uses Calcite's RexBuilder to ensure type-correct literal creation
• **Error handling**: Validates field existence, sort presence, and value count matching

### Changes

#### Core Logic (FilterConverter.java)

• Extended isApplicable() to check for both query and search_after presence
• Added buildSearchAfterCondition() to construct filter predicates from pagination values
• Implemented multi-field sort handling with proper OR/AND combinations
• Added validation for sort requirements and value count matching
• Combined query and search_after conditions using AND operator

#### Integration Tests (DslSearchAfterIT.java) - Not Working as of Now

• testSearchAfterWithSingleSort(): Validates basic single-field pagination
• testSearchAfterWithMultipleSorts(): Tests multi-field sort precedence
• testSearchAfterWithQuery(): Ensures search_after works with query filters

#### Unit Tests (FilterConverterTests.java)

• Tests for ASC/DESC sort direction handling
• Multi-field sort combination logic
• Query + search_after condition merging
• Error cases: missing sort, value count mismatch, unknown fields
• Sort order vs schema order validation

### Testing

Unit Tests: 7 new test cases covering:
• Single and multi-field sorting
• ASC/DESC direction handling
• Query combination logic
• Error conditions and validation


### Manual Testing:
Index creation:
```
% curl -k -X PUT "localhost:9200/test-index" \            
-H "Content-Type: application/json" -d'
{                        
  "settings": {               
    "number_of_shards": 1,
    "number_of_replicas": 0   
  },                     
  "mappings": {               
    "properties": {      
      "name": { "type": "keyword" },
      "age":  { "type": "integer" }
    }                         
  }
}'
{"acknowledged":true,"shards_acknowledged":true,"index":"test-index"}%                                                                                                               
```
Test data for index:
```
% curl -k -X POST "localhost:9200/test-index/_bulk" \
-H "Content-Type: application/json" -d'
{ "index": { "_id": 1 } }
{ "name": "alice", "age": 25 }
{ "index": { "_id": 2 } }
{ "name": "bob",   "age": 30 }
{ "index": { "_id": 3 } }
{ "name": "john",  "age": 30 }
{ "index": { "_id": 4 } }
{ "name": "david", "age": 35 }
{ "index": { "_id": 5 } }
{ "name": "zara",  "age": 40 }
'
{"took":787,"errors":false,"items":[{"index":{"_index":"test-index","_id":"1","_version":1,"result":"created","_shards":{"total":1,"successful":1,"failed":0},"_seq_no":0,"_primary_term":1,"status":201}},{"index":{"_index":"test-index","_id":"2","_version":1,"result":"created","_shards":{"total":1,"successful":1,"failed":0},"_seq_no":1,"_primary_term":1,"status":201}},{"index":{"_index":"test-index","_id":"3","_version":1,"result":"created","_shards":{"total":1,"successful":1,"failed":0},"_seq_no":2,"_primary_term":1,"status":201}},{"index":{"_index":"test-index","_id":"4","_version":1,"result":"created","_shards":{"total":1,"successful":1,"failed":0},"_seq_no":3,"_primary_term":1,"status":201}},{"index":{"_index":"test-index","_id":"5","_version":1,"result":"created","_shards":{"total":1,"successful":1,"failed":0},"_seq_no":4,"_primary_term":1,"status":201}}]}%                                
% curl -k -X POST "localhost:9200/test-index/_refresh"
{"_shards":{"total":1,"successful":1,"failed":0}}%     
```
Search Query with search_after pagination:
```
% curl -k -X GET "localhost:9200/test-index/_search" \
-H "Content-Type: application/json" -d'
{
  "size": 2,
  "sort": [
    { "age": "asc" },
    { "name": "asc" }
  ],
  "search_after": [30, "bob"],
  "_source": ["name", "age"]
}' 
```

Verification of Calcite Rel node creation:
* Sort order is broken due to known issue in Sort, will be raising fix in another PR for that.
```
LogicalSort(sort0=[$1], sort1=[$0], dir0=[ASC], dir1=[ASC], fetch=[2])
  LogicalProject(name=[$1], age=[$0])
    LogicalFilter(condition=[OR(>($0, CAST(30):INTEGER), AND(=($0, CAST(30):INTEGER), >($1, CAST('bob':VARCHAR):VARCHAR)))])
      LogicalTableScan(table=[[test-index]])
```

### Check List
- [Yes] Functionality includes testing.
- [Yes ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [Yes ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

### DIFF
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/FilterConverter.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/FilterConverter.java
@@ -10,15 +10,21 @@ package org.opensearch.dsl.converter;
 
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.logical.LogicalFilter;
+import org.apache.calcite.rel.type.RelDataTypeField;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 import org.opensearch.dsl.query.QueryRegistry;
 import org.opensearch.index.query.MatchAllQueryBuilder;
+import org.opensearch.search.builder.SearchSourceBuilder;
+import org.opensearch.search.sort.FieldSortBuilder;
+import org.opensearch.search.sort.SortOrder;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 
 /**
- * Converts the DSL {@code query} clause to a {@link LogicalFilter}.
- * Skips match_all since a table scan already returns all rows.
+ * Converts the DSL {@code query} clause and {@code search_after} to a {@link LogicalFilter}.
  */
 public class FilterConverter extends AbstractDslConverter {
 
@@ -35,13 +41,83 @@ public class FilterConverter extends AbstractDslConverter {
 
     @Override
     protected boolean isApplicable(ConversionContext ctx) {
-        return ctx.getSearchSource().query() != null
-            && !(ctx.getSearchSource().query() instanceof MatchAllQueryBuilder);
+        return hasQuery(ctx) || hasSearchAfter(ctx);
     }
 
     @Override
     protected RelNode doConvert(RelNode input, ConversionContext ctx) throws ConversionException {
-        RexNode condition = queryRegistry.convert(ctx.getSearchSource().query(), ctx);
+        List<RexNode> conditions = new ArrayList<>();
+
+        if (hasQuery(ctx)) {
+            conditions.add(queryRegistry.convert(ctx.getSearchSource().query(), ctx));
+        }
+
+        if (hasSearchAfter(ctx)) {
+            conditions.add(buildSearchAfterCondition(input, ctx));
+        }
+
+        RexNode condition = conditions.size() == 1 ? conditions.get(0)
+            : ctx.getRexBuilder().makeCall(SqlStdOperatorTable.AND, conditions);
         return LogicalFilter.create(input, condition);
     }
+
+    private RexNode buildSearchAfterCondition(RelNode input, ConversionContext ctx) throws ConversionException {
+        SearchSourceBuilder ss = ctx.getSearchSource();
+        Object[] values = ss.searchAfter();
+
+        if (ss.sorts() == null || ss.sorts().isEmpty()) {
+            throw new ConversionException("search_after requires sort");
+        }
+        if (values.length != ss.sorts().size()) {
+            throw new ConversionException("search_after values must match sort fields");
+        }
+
+        List<RexNode> orConditions = new ArrayList<>();
+        for (int sortIdx = 0; sortIdx < ss.sorts().size(); sortIdx++) {
+            if (!(ss.sorts().get(sortIdx) instanceof FieldSortBuilder fieldSort)) {
+                throw new ConversionException("search_after only supports field sorts");
+            }
+
+            RelDataTypeField field = input.getRowType().getField(fieldSort.getFieldName(), false, false);
+            if (field == null) {
+                throw new ConversionException("Field not found: " + fieldSort.getFieldName());
+            }
+
+            RexNode fieldRef = ctx.getRexBuilder().makeInputRef(field.getType(), field.getIndex());
+            RexNode literal = ctx.getRexBuilder().makeLiteral(values[sortIdx], field.getType(), true);
+            RexNode comparison = fieldSort.order() == SortOrder.ASC
+                ? ctx.getRexBuilder().makeCall(SqlStdOperatorTable.GREATER_THAN, fieldRef, literal)
+                : ctx.getRexBuilder().makeCall(SqlStdOperatorTable.LESS_THAN, fieldRef, literal);
+
+            List<RexNode> equals = new ArrayList<>();
+            for (int j = 0; j < sortIdx; j++) {
+                FieldSortBuilder prev = (FieldSortBuilder) ss.sorts().get(j);
+                RelDataTypeField pf = input.getRowType().getField(prev.getFieldName(), false, false);
+                equals.add(ctx.getRexBuilder().makeCall(SqlStdOperatorTable.EQUALS,
+                    ctx.getRexBuilder().makeInputRef(pf.getType(), pf.getIndex()),
+                    ctx.getRexBuilder().makeLiteral(values[j], pf.getType(), true)));
+            }
+
+            if (equals.isEmpty()) {
+                orConditions.add(comparison);
+            } else {
+                RexNode allEquals = equals.size() == 1 ? equals.get(0)
+                    : ctx.getRexBuilder().makeCall(SqlStdOperatorTable.AND, equals);
+                orConditions.add(ctx.getRexBuilder().makeCall(SqlStdOperatorTable.AND, allEquals, comparison));
+            }
+        }
+
+        return orConditions.size() == 1 ? orConditions.get(0)
+            : ctx.getRexBuilder().makeCall(SqlStdOperatorTable.OR, orConditions);
+    }
+
+    private static boolean hasQuery(ConversionContext ctx) {
+        return ctx.getSearchSource().query() != null
+            && !(ctx.getSearchSource().query() instanceof MatchAllQueryBuilder);
+    }
+
+    private static boolean hasSearchAfter(ConversionContext ctx) {
+        return ctx.getSearchSource().searchAfter() != null && ctx.getSearchSource().searchAfter().length > 0;
+    }
 }
+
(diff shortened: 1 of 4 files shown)


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/FilterConverter.java
```diff
@@ -35,13 +41,83 @@ public FilterConverter(QueryRegistry queryRegistry) {
...
+        if (hasSearchAfter(ctx)) {
+            conditions.add(buildSearchAfterCondition(input, ctx));
+        }
+
+        RexNode condition = conditions.size() == 1 ? conditions.get(0)
+            : ctx.getRexBuilder().makeCall(SqlStdOperatorTable.AND, conditions);
         return LogicalFilter.create(input, condition);
     }
+
+    private RexNode buildSearchAfterCondition(RelNode input, ConversionContext ctx) throws ConversionException {
```
vinaykpud, line 64: Lot of steps done as part of this function, some one line comments will help

abhishek00159: Ack, will update. This is verifying that sort fields matches with search_after fields. And then create condition for filter according to nested structure.

bharath-techie:
> In contrast, a Apache Calcite-based implementation models search_after as a filter condition followed by sorting. The execution flow becomes: Filter → Sort → Limit. This reduces the dataset first (from N → M), but then performs a full sort on the filtered results. The resulting time complexity is approximately: O(N + M · log M) ≈ O(M log M), where: M = number of rows after filtering. Further optimizations may apply if indexes are available, but in general, this approach still requires sorting the reduced dataset.

Datafusion does support pushing down sort, filter , limits and projections.  So we should try to pushdown in all these cases.
We should not be limited by calcite if pushing down any of these becomes a concern because of it. Will the plan get optimized in DF layer ?

bharath-techie:
Also there is another case with re-index where we were thinking to use PIT with search after instead of scroll. 

Is PIT with search after coming in successive PR ?

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/converter/FilterConverterTests.java
```diff
@@ -81,4 +82,109 @@ public void testUnsupportedQueryProducesFilterWithUnresolvedCondition() throws C
...
+
+        assertTrue(result instanceof LogicalFilter);
+        RexCall condition = (RexCall) ((LogicalFilter) result).getCondition();
+        assertEquals(SqlKind.GREATER_THAN, condition.getKind());
+    }
+
+    public void testSearchAfterWithSingleSortDesc() throws ConversionException {
+        SearchSourceBuilder source = new SearchSourceBuilder()
+            .sort("price", SortOrder.DESC)
+            .searchAfter(new Object[]{100});
```
bharath-techie, line 101: In datafusion we can represent it as limit with offsets , `SELECT price FROM tbl WHERE price = '100' LIMIT 100,10`

Can we try maybe to atleast get a e2e picture - on how DF physical plan will look like for some of these calcite plans. ?

abhishek00159: Yes, I am working on getting DF physical plan. But I am able to get queryDAG:
For query:
```
curl -X GET "localhost:9200/test_index/_search" -H 'Content-Type: application/json' -d '
{
  "sort": [
    { "score": "asc" }  ],
  "search_after": [85]
}'
```
```
QueryDAG(queryId=6c35384a-e154-49cb-9fc5-debd225d574b)
Stage 0 [root]
  OpenSearchSort(sort0=[$4], dir0=[ASC], fetch=[2], viableBackends=[[datafusion]])
    OpenSearchFilter(condition=[ANNOTATED_PREDICATE(id=0, backends=[datafusion], >($4, CAST(85:BIGINT):BIGINT))], viableBackends=[[datafusion]])
      OpenSearchTableScan(table=[[parquet_simple]], viableBackends=[[datafusion]])
  Alternatives (1):
  [0] OpenSearchSort(sort0=[$4], dir0=[ASC], fetch=[2], viableBackends=[[datafusion]])
        OpenSearchFilter(condition=[ANNOTATED_PREDICATE(id=0, backends=[datafusion], >($4, CAST(85:BIGINT):BIGINT))], viableBackends=[[datafusion]])
          OpenSearchTableScan(table=[[parquet_simple]], viableBackends=[[datafusion]])
```
Also why offset can't be used for search_after as search_after only works with sort feature and search_after can be field with Integer, string, date, etc format, So we need filter on this field to get the subset of data and also search_after is property of sort so it can only be applied on query with sort. So eventually we need to use filter and sort everytime search_after is used. Now if we use limit which is currently in process of implementation we can further get the size of the page which can be applied after sort.
In case of above query `SELECT price FROM tbl WHERE price = '100' LIMIT 100,10` it will only give me any 10 record where price is 100 and even if we use sort in this offset will be of no use as search_after work only with size(Limit).
I will check once have physical plan if any pushdown is happening over limit which is optimizing it.

abhishek00159:
> Also there is another case with re-index where we were thinking to use PIT with search after instead of scroll.
> 
> Is PIT with search after coming in successive PR ?

@bharath-techie 
Point in Time (PIT) [DF does not natively support maintaining a frozen state needed for scroll] so we will not support it.

abhishek00159:
> > In contrast, a Apache Calcite-based implementation models search_after as a filter condition followed by sorting. The execution flow becomes: Filter → Sort → Limit. This reduces the dataset first (from N → M), but then performs a full sort on the filtered results. The resulting time complexity is approximately: O(N + M · log M) ≈ O(M log M), where: M = number of rows after filtering. Further optimizations may apply if indexes are available, but in general, this approach still requires sorting the reduced dataset.
> 
> Datafusion does support pushing down sort, filter , limits and projections. So we should try to pushdown in all these cases. We should not be limited by calcite if pushing down any of these becomes a concern because of it. Will the plan get optimized in DF layer ?

@bharath-techie as we are using filter + sort + limits + projections so eventually DF will still do pushing down. I can share what optimization in DF layer are happening post pushing down.

bharath-techie:
> Point in Time (PIT) [DF does not natively support maintaining a frozen state needed for scroll] so we will not support it.

Why not. ? We maintain the context at OpenSearch right, we have datafusion reader too frozen at point in time, we just need to ensure we are able to scroll through the entire data 0 to N with limit , sort and offset.

sandbox/plugins/dsl-query-executor/src/test/java/org/opensearch/dsl/converter/FilterConverterTests.java
bharath-techie: So how are we planning to use search_after with datafusion, am still kind of confused. 

Since we are in `feature/datafusion` branch, we should be able to create substrait for the same DAG right ?

PR #21145, opened 2026-04-06
File Deletions Support through IndexFileDeleter in CatalogSnapshotManager

rayshrey:
### Description

### Related Issues

### Check List


server/src/main/java/org/opensearch/index/engine/exec/commit/SafeBootstrapCommitter.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    protected SafeBootstrapCommitter(CommitterConfig config) throws IOException {
+        EngineConfig engineConfig = config.engineConfig();
+        if (engineConfig == null || engineConfig.getStore() == null) {
+            throw new IllegalArgumentException("SafeBootstrapCommitter requires a non-null EngineConfig with a valid Store");
+        }
+        Store store = engineConfig.getStore();
+        if (engineConfig.getTranslogConfig() != null) {
+            Path translogPath = engineConfig.getTranslogConfig().getTranslogPath();
+            if (translogPath != null) {
```
mgodwan: Should we assert the case when translogPath is null?

server/src/main/java/org/opensearch/index/engine/exec/commit/SafeBootstrapCommitter.java
```diff
@@ -0,0 +1,73 @@
...
+public abstract class SafeBootstrapCommitter implements Committer {
+
+    /**
+     * Constructs the committer, validating config and trimming unsafe commits.
+     * If translog config is absent, safe bootstrap trimming is skipped.
+     *
+     * @param config the committer configuration
+     * @throws IOException if commit discovery or rewriting fails
+     */
+    protected SafeBootstrapCommitter(CommitterConfig config) throws IOException {
```
mgodwan, line 44: Make a note here that current store bootstrap is continued to be used, and have a todo to decouple the store impl as well and rely on Committer provided static methods

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneCommitter.java
```diff
@@ -139,9 +169,99 @@ IndexWriter getIndexWriter() {
...
+     * Deserializes CatalogSnapshots from Lucene IndexCommits in the given store.
+     * Static-friendly — does not depend on instance fields.
+     */
+    static List<CatalogSnapshot> deserializeCatalogSnapshots(Store store) throws IOException {
+        Function<String, String> resolver = fn -> store.shardPath().getDataPath().resolve(fn).toString();
+        List<IndexCommit> commits;
+        try {
+            commits = DirectoryReader.listCommits(store.directory());
+        } catch (IndexNotFoundException e) {
+            return List.of();
```
mgodwan: We expect that there should at least be one commit always, right? due to store bootstrap? Should we fail in this case?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneCommitter.java
```diff
@@ -139,9 +169,99 @@ IndexWriter getIndexWriter() {
...
+        }
+        List<CatalogSnapshot> result = new ArrayList<>();
+        for (IndexCommit ic : commits) {
+            String serialized = ic.getUserData().get(CatalogSnapshot.CATALOG_SNAPSHOT_KEY);
+            if (serialized != null && serialized.isEmpty() == false) {
```
mgodwan, line 239: Can you add a note when it can be null (i.e. empty store recovery)

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneCommitter.java
```diff
@@ -139,9 +169,99 @@ IndexWriter getIndexWriter() {
...
+        }
+        // Read persisted global checkpoint from translog checkpoint file
+        String translogUUID = commits.getLast().getUserData().get(Translog.TRANSLOG_UUID_KEY);
+        long globalCheckpoint = Translog.readGlobalCheckpoint(translogPath, translogUUID);
+        CatalogSnapshot safeCommit = CombinedCatalogSnapshotDeletionPolicy.findSafeCommitPoint(commits, globalCheckpoint);
+        // Rewrite at the safe commit point
+        Function<String, String> resolver = fn -> store.shardPath().getDataPath().resolve(fn).toString();
+        List<IndexCommit> indexCommits = DirectoryReader.listCommits(store.directory());
+        IndexCommit safeIndexCommit = null;
+        for (IndexCommit ic : indexCommits) {
```
mgodwan: Do we need to recompute this? This already happened in deserializeCatalogSnapshots

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneCommitter.java
```diff
@@ -139,9 +169,99 @@ IndexWriter getIndexWriter() {
...
+            if (serialized != null && serialized.isEmpty() == false) {
+                CatalogSnapshot cs = DataformatAwareCatalogSnapshot.deserializeFromString(serialized, resolver);
+                if (cs.getGeneration() == safeCommit.getGeneration()) {
+                    safeIndexCommit = ic;
+                    break;
+                }
+            }
+        }
+        if (safeIndexCommit == null) {
+            return;
```
mgodwan: Shouldn't we fail this if safe commit is null?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneIndexingExecutionEngine.java
```diff
@@ -163,8 +164,28 @@ public DataFormat getDataFormat() {
...
+    public Map<String, Collection<String>> deleteFiles(Map<String, Collection<String>> filesToDelete) throws IOException {
+        Collection<String> luceneFiles = filesToDelete.get(LUCENE_FORMAT_NAME);
+        if (luceneFiles == null || luceneFiles.isEmpty()) {
+            return Map.of();
+        }
+        Directory directory = store.directory();
+        Collection<String> failed = new ArrayList<>();
+        for (String fileName : luceneFiles) {
+            try {
+                directory.deleteFile(fileName);
```
mgodwan: Do we need to actually call deleteFile on directory here? Given, IndexWriter already has wiring into the FileDeleter lucene internally maintains, should we not just have a deletion policy which handles that directly?

server/src/main/java/org/opensearch/index/engine/exec/coord/IndexFileDeleter.java
```diff
@@ -0,0 +1,458 @@
...
+    private void executeDeletesWithRetry(Map<String, Collection<String>> filesByFormat) throws IOException {
+        // Filter out files that were re-referenced since decRefFiles ran
+        Map<String, Collection<String>> safeToDelete = new HashMap<>();
+        synchronized (this) {
+            for (Map.Entry<String, Collection<String>> entry : filesByFormat.entrySet()) {
+                String formatName = entry.getKey();
+                Map<String, AtomicInteger> formatRefs = fileRefCounts.get(formatName);
+                Collection<String> stillDead = new HashSet<>();
+                for (String file : entry.getValue()) {
+                    if (formatRefs == null || formatRefs.containsKey(file) == false) {
```
mgodwan, line 318: In what cases can this be revived?

mgodwan: If no case is there, we should assert that formatRefs does not contain key, and fail assertion otherwise

server/src/main/java/org/opensearch/index/engine/exec/coord/IndexFileDeleter.java
```diff
@@ -0,0 +1,458 @@
...
+    @Override
+    public void close() throws IOException {
+        List<CatalogSnapshot> snapshotsToRelease;
+        synchronized (this) {
+            snapshotsToRelease = new ArrayList<>(committedSnapshots);
+            committedSnapshots.clear();
+            pendingDeletes.clear();
+        }
+        for (CatalogSnapshot cs : snapshotsToRelease) {
+            cs.decRef();
```
mgodwan: Won't this delete all the files?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneCommitter.java
```diff
@@ -128,6 +129,35 @@ public SafeCommitInfo getSafeCommitInfo() {
...
+        for (String fileName : store.directory().listAll()) {
+            if (fileName.startsWith(IndexFileNames.SEGMENTS) == false) {
+                continue;
+            }
+            try {
+                SegmentInfos sis = SegmentInfos.readCommit(store.directory(), fileName);
+                Map<String, String> commitUserData = sis.getUserData();
+                boolean match = snapshotUserData.entrySet().stream().allMatch(e -> e.getValue().equals(commitUserData.get(e.getKey())));
+                if (match) {
+                    store.directory().deleteFile(fileName);
```
mgodwan: This will mess with the way IndexWriter internally maintains all of this.

We can do something like

```
class MyPolicy implements IndexDeletionPolicy {

        private Map<String, IndexCommit> trackedCommits = new HashMap<>();

        @Override
        public void onInit(List<? extends IndexCommit> commits) throws IOException {
            onCommit(commits);
        }

        @Override
        public void onCommit(List<? extends IndexCommit> commits) throws IOException {
            for (IndexCommit commit : commits) {
                trackedCommits.putIfAbsent(commit.getUserData().get("snapshot_generation"), commit);
            }
        }

        public void purgeCommit(String generation) throws IOException {
            IndexCommit commit = trackedCommits.get(generation);
            if (commit != null) {
                commit.delete();
                trackedCommits.remove(generation);
            }
        }
    }
```

Committer's deleteCommit may just call purgeCommit from this policy and lucene can internally track this

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneIndexingExecutionEngine.java
mgodwan: Lucene would actually throw a NoSuchFileException here. closing reader on lucene would take care of this delete as well.

mgodwan (approved):

PR #21147, opened 2026-04-06
Added support for DSL Exists query to convert in RelNode format

abhishek00159:
### Description
Implements the `QueryTranslator` interface to convert OpenSearch `ExistsQueryBuilder` to Calcite `IS NOT NULL` RexNode.

### 1. Created ExistsQueryTranslator
**Key Logic:**
- Extracts field name from `ExistsQueryBuilder`
- Looks up field in schema to get field index and throw exception if boost is passed with not default value or field is not passed.
- Creates `IS NOT NULL($fieldIndex)` RexNode using Calcite's `SqlStdOperatorTable.IS_NOT_NULL`

**Calcite Output:**
```
LogicalFilter(condition=[IS NOT NULL($N)])
  LogicalTableScan(table=[[index-name]])
```
### 2. Registered ExistsQueryTranslator
**File:** `src/main/java/org/opensearch/dsl/query/QueryRegistryFactory.java`

Added registration in the `create()` method:
```java
registry.register(new ExistsQueryTranslator());
```

This makes the translator available for all exists queries.

### 3. Added Integration Tests
**File:** `src/internalClusterTest/java/org/opensearch/dsl/DslLogicalPlanIntegrationIT.java`

Added three comprehensive tests.
1. Exists query test
2. Exists query with bool.
3. Exist query not supported for boost.

### Related Issues
NA

### Check List
- [Done ] Functionality includes testing.
- [NA ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ NA] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

### DIFF
--- a/sandbox/plugins/dsl-query-executor/src/internalClusterTest/java/org/opensearch/dsl/DslQueryIT.java
+++ b/sandbox/plugins/dsl-query-executor/src/internalClusterTest/java/org/opensearch/dsl/DslQueryIT.java
@@ -54,4 +54,16 @@ public class DslQueryIT extends DslIntegTestBase {
             client().search(new SearchRequest(INDEX, "test-index-2").source(new SearchSourceBuilder())).actionGet()
         );
     }
+
+    public void testExistsQuery() {
+        createTestIndex();
+        assertOk(search(new SearchSourceBuilder().query(QueryBuilders.existsQuery("name"))));
+    }
+
+    public void testExistsQueryWithBoostFails() {
+        createTestIndex();
+        expectThrows(Exception.class, () ->
+            search(new SearchSourceBuilder().query(QueryBuilders.existsQuery("name").boost(2.0f)))
+        );
+    }
 }
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/query/ExistsQueryTranslator.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/query/ExistsQueryTranslator.java
@@ -1,0 +1,47 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.dsl.query;
+
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexNode;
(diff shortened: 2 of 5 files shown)


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/query/ExistsQueryTranslator.java
```diff
@@ -0,0 +1,47 @@
...
+    public RexNode convert(QueryBuilder query, ConversionContext ctx) throws ConversionException {
+        ExistsQueryBuilder existsQuery = (ExistsQueryBuilder) query;
+        String fieldName = existsQuery.fieldName();
+        float boost = existsQuery.boost();
+
+        RelDataTypeField field = ctx.getRowType().getField(fieldName, false, false);
+        if (field == null) {
+            throw new ConversionException("Field '" + fieldName + "' not found in schema");
+        }
+        if (boost != AbstractQueryBuilder.DEFAULT_BOOST) {
```
vinaykpud, line 40: boost is checked after field lookup. If someone passes both an invalid field and a non-default boost, they'll get "field not found" instead of "boost unsupported." Consider validating boost first since it's cheaper (no schema lookup) and gives a more actionable error message. This is minor but easy to fix.

sandbox/plugins/dsl-query-executor/src/internalClusterTest/java/org/opensearch/dsl/DslQueryIT.java
```diff
@@ -54,4 +54,16 @@ public void testFailsForMultipleIndices() {
             client().search(new SearchRequest(INDEX, "test-index-2").source(new SearchSourceBuilder())).actionGet()
         );
     }
+
```
vinaykpud, line 57: The PR description mentions "Exists query with bool" as one of three integration tests, but the actually two integration tests. The bool combination test is missing. Do we need that?

abhishek00159: Sure, will update it.

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/query/ExistsQueryTranslator.java
abhishek00159: Sure, updating in next revision.

PR #21160, opened 2026-04-08
Analytics Plugin Planner Marking Phase with Calcite RBO and CBO integration

expani:
### Description

Adds the coordinator-side query planner for the Analytics Plugin — the marking phase that converts logical Calcite RelNode trees into annotated OpenSearchXxx nodes with per-operator, per-predicate, per-aggregate-call, and per-expression viable backend lists.

analytics-framework (SPI):

- New capability model: BackendCapabilityProvider, FilterCapability (Standard/FullText/Expression), AggregateCapability (with AggregateDecomposition for future partial state customization), ProjectCapability (Scalar/Opaque), WindowCapability,
FilterOperator, AggregateFunction, FieldType, OperatorCapability, DelegationType, ShuffleCapability, FragmentConvertor.

These must natura

- AnalyticsSearchBackendPlugin refactored: name() abstract, three provider getters (getCapabilityProvider(), getSearchExecEngineProvider(), getFragmentConvertor()) — each is empty by default so backends

analytics-engine (planner):
- CapabilityRegistry — singleton built at plugin startup, pre-indexed lookups for filter/aggregate/scalar/opaque/window backends by (operator, fieldType, format)
- FieldStorageResolver — resolves per-field storage from IndexMetadata: primary_data_format for doc values, lucene for explicitly indexed fields
- Five marking rules using CapabilityRegistry — no backend iteration in rules
- OpenSearchAggregateSplitRule (Volcano CBO) — splits SINGLE → PARTIAL+FINAL on distribution trait mismatch; decomposition deferred to plan forking (see TODO in rule javadoc)
- Custom RelNode types carrying viableBackends per operator and per-expression annotations
- PlannerImpl.markAndOptimize() — HepPlanner (RBO) + VolcanoPlanner (CBO)

analytics-backend-datafusion: Updated to implement getSearchExecEngineProvider() per new SPI.

Tests (70 passing): 34 new planner tests across AggregateRuleTests, FilterRuleTests, ProjectRuleTests. MockBackend base class eliminates BackendCapabilityProvider boilerplate in test subclasses.

### Check List


bowenlan-amzn:
### Code review

Found 1 issue:

1. `PlannerImpl.java` is missing the mandatory Apache-2.0 SPDX license header. Every other new Java file in this PR (61 files) includes the standard header block. This file starts directly with the `package` declaration.

https://github.com/opensearch-project/OpenSearch/blob/7e66435540a62b32b629ddf6d993735faea11f4f/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java#L1-L3

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/OperatorCapability.java
```diff
@@ -0,0 +1,28 @@
...
+    SCAN,
+    FILTER,
+    /** Can evaluate filter predicates on derived/expression columns (e.g. HAVING on aggregate output). */
+    FILTER_ON_EXPRESSIONS,
+    AGGREGATE,
+    SORT,
+    JOIN,
+    WINDOW,
+    PROJECT,
+    COORDINATOR_REDUCE
```
mch2: This isn't really a relational operator capability. the reduce would occur on a plan fragment containing multiple operators, more of an engine capability?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateDecomposition.java
```diff
@@ -0,0 +1,60 @@
...
+ * <ul>
+ *   <li>COVAR_POP(x, y): partial emits SUM(x*y), SUM(x), SUM(y), COUNT;
+ *       final expression: (SUM(x*y) - SUM(x)*SUM(y)/COUNT) / COUNT</li>
+ *   <li>HLL distinct count: partial emits a single HLL sketch accumulator;
+ *       final expression: HLL_MERGE(sketches) → cardinality estimate</li>
+ * </ul>
+ *
+ * @opensearch.internal
+ */
+public interface AggregateDecomposition {
```
mch2, line 39: Can we get by without this? The same engine will produce the partial and compute the final - the schema can be determined from the vsr?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateFunction.java
```diff
@@ -0,0 +1,87 @@
...
+    PERCENTILE_CONT(Type.STATE_EXPANDING, SqlKind.PERCENTILE_CONT),
+    PERCENTILE_DISC(Type.STATE_EXPANDING, SqlKind.PERCENTILE_DISC),
+    COLLECT(Type.STATE_EXPANDING, SqlKind.COLLECT),
+    LISTAGG(Type.STATE_EXPANDING, SqlKind.LISTAGG),
+
+    // Approximate — probabilistic, fixed-size state
+    APPROX_COUNT_DISTINCT(Type.APPROXIMATE, SqlKind.OTHER);
+
+    /** Category of aggregate function. Affects execution strategy (shuffle vs map-reduce). */
+    public enum Type {
```
mch2, line 46: `Affects execution strategy` - these are hints for the planner yes?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateCapability.java
```diff
@@ -0,0 +1,78 @@
...
+        this(function, fieldType, formats, null);
+    }
+
+    public static AggregateCapability simple(AggregateFunction function, FieldType fieldType,
+                                              Set<String> formats) {
+        assert function.getType() == AggregateFunction.Type.SIMPLE;
+        return new AggregateCapability(function, fieldType, formats);
+    }
+
+    public static AggregateCapability statistical(AggregateFunction function, FieldType fieldType,
```
mch2: is this going to expand? as is not sure we we need these assertions / separate methods.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FilterOperator.java
```diff
@@ -0,0 +1,114 @@
...
+
+    public SqlKind getSqlKind() {
+        return sqlKind;
+    }
+
+    /**
+     * Returns a Calcite {@link SqlFunction} for this full-text operator.
+     * Only valid for operators of type {@link Type#FULL_TEXT}.
+     */
+    public SqlFunction toSqlFunction() {
```
mch2, line 82: its a bit hard to follow where if anywhere these functions are used - can we introduce when needed? same goes for the rest of the pr.

mch2 (commented):
thanks @expani, i'm still going through this - publishing comments thus far...

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/OperatorCapability.java
expani: Makes sense this can be separated. Probably this and arrow compatibility can become engine capability

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateDecomposition.java
expani: Maybe but the RelNode's rowtype won't match with the data generated by VSR so kept it for consistency. 
Let me add a TODO to revisit once we implement approximate functions via DataFusion

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateFunction.java
expani: Yes this will be useful in determining whether to do a partial agg before or after shuffle. If the aggregate functions are STATE_EXPANDING, it will increase the size of partial state and thus shuffled data. Such a decision can be taken by the planner

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateCapability.java
expani: Likely not much as most of aggregate functions we implement should fall in one of these 4 categories.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FilterOperator.java
expani: For now, mostly in UTs within FilterRuleTests. 
I can clean this up further.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateCapability.java
expani: Will remove the asserts.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DelegationType.java
```diff
@@ -0,0 +1,33 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.analytics.spi;
+
+/**
+ * Types of work that can be delegated between backends.
+ *
+ * <p>A backend declares {@code canDelegate(FILTER)} if it has a custom
```
bowenlan-amzn, line 14: It seems large amount of complexity of this PR comes from delegation logic, which makes me believe this delegation mechanism **should not** be merged in with this PR which is to establish the general foundation planning logic. Basically enough planning logic to make the datafusion parquet side work end to end.

Could we wait till both backends functioning and then do this delegation logic end to end in one PR? I guess the shape of this delegation logic originates from some prototype, and I doubt we really need to implement like this in the end.

expani: There are several pieces at work here : 
- Feasibility check on the backends on Filter Predicates, Aggregate Functions and generic capabilities to Scan, Sort and perform Coordinator Reduce. These are not linked to delegation. 
- Given that marking phase will do these checks, I have included the Planner to also check whether one backend can delegate to other ***only for Filters.*** So, delegation is only restricted to the class `OpenSearchFilterRule` in this PR which is around 50-60 lines of code at best. 
- I have not included the fragment conversion for delegation in this PR which would add a lot further. I also left out checking the feasibility to delegate anything other than filter ( Project, Aggregate; etc ) 


Some things we would still need in the planner irrespective of how Delegation gets implemented : 
- Feasibility check of whether one backend can delegate and another can accept those delegation. 
- If a delegation is performed, which Delegation API makes sense to be used. For instance, Filter Delegation b/w DF and Lucene have 2 sub-cases, Pure AND filter delegation as well as Kleen's 3 way merge for complex boolean expressions ( OR, NOT; etc ) This will also be determined by Planner going forward but hasn't been included in this PR


Let me know your thoughts if these are enough to start with.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -78,7 +78,7 @@ public Iterable<Object[]> execute(RelNode logicalFragment, Object context) {
         List<Object[]> rows = new ArrayList<>();
         try (var dataFormatAwareReader = dataFormatAwareEngine.acquireReader()) {
             ExecutionContext ctx = new ExecutionContext(tableName, task, dataFormatAwareReader.get());
-            try (SearchExecEngine<ExecutionContext, EngineResultStream> engine = provider.createSearchExecEngine(ctx)) {
+            try (SearchExecEngine<ExecutionContext, EngineResultStream> engine = provider.getSearchExecEngineProvider().createSearchExecEngine(ctx)) {
```
vinaykpud: nit: Although its not directly related to this PR
can we rename `provider` → `backendPlugin`, the current name reads awkwardly with the SPI (provider.getSearchExecEngineProvider()), and the variable holds an AnalyticsSearchBackendPlugin, not a provider.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -0,0 +1,111 @@
+package org.opensearch.analytics.planner;
```
vinaykpud, line 9: Missing license header

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -0,0 +1,312 @@
...
+                // Derived column — only child viable backends + their delegation targets
+                List<String> anyFormat = registry.filterBackendsAnyFormat(operator, fieldType);
+                List<String> delegationAcceptors = registry.delegationAcceptors(DelegationType.FILTER);
+                for (String name : childViableBackends) {
+                    if (anyFormat.contains(name)) {
+                        fieldViable.add(name);
+                    }
+                }
+                // Delegation targets reachable from child viable backends
+                List<String> delegationSupporters = registry.delegationSupporters(DelegationType.FILTER);
```
vinaykpud: when would a backend need to delegate a filter on a derived column? The backend already has the value in its pipeline. Is the delegation check here needed?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -0,0 +1,312 @@
...
+                if (childViableBackends.stream().anyMatch(delegationSupporters::contains)) {
+                    for (String name : anyFormat) {
+                        if (!fieldViable.contains(name) && delegationAcceptors.contains(name)) {
+                            fieldViable.add(name);
+                        }
+                    }
+                }
+            } else {
```
vinaykpud: for non-derived fields path format-aware lookups for doc values and index formats looks redundant because `filterBackendsAnyFormat` returns a superset. Could this be simplified to just `filterBackendsAnyFormat`, or I am I missing something?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -0,0 +1,312 @@
...
+
+    /**
+     * Determines which backends can evaluate this leaf predicate.
+     * Extracts all field references, looks up their {@link FieldStorageInfo} from the child,
+     * checks backend format support, operator capability, and operator+fieldType support.
+     * Intersects across all referenced fields.
+     */
+    private List<String> resolveViableBackends(
+        RexCall predicate,
+        RelDataType inputRowType,
```
vinaykpud: unused field

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -0,0 +1,312 @@
...
+
+    /**
+     * Recursively walks the condition tree. Boolean connectives (AND, OR, NOT) are
+     * preserved — we recurse into their children. Leaf predicates are wrapped in
+     * {@link AnnotatedPredicate} with viable backends resolved from child's field storage.
+     */
+    private RexNode annotateCondition(
+        RexNode condition,
+        RelDataType inputRowType,
+        List<FieldStorageInfo> fieldStorage,
```
vinaykpud: can we keep consistency of naming? `fieldStorage` -> `fieldStorageInfos` ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -0,0 +1,312 @@
...
+        }
+        if (operator == null) {
+            throw new IllegalStateException("Unrecognized filter operator [" + predicate.getKind() + "]");
+        }
+
+        Set<String> viableSet = new HashSet<>(registry.operatorBackends(OperatorCapability.FILTER));
+
+        for (int fieldIndex : fieldIndices) {
+            if (fieldIndex >= fieldStorage.size()) {
+                continue;
```
vinaykpud: I think this shouldn't happen since Calcite validates field references, but if it does, should we  throw  error here instead of silently ignoring

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTableScanRule.java
```diff
@@ -0,0 +1,113 @@
...
+
+        // TODO: table name can be an index pattern — needs IndexNameExpressionResolver
+        String tableName = scan.getTable().getQualifiedName().getLast();
+
+        IndexMetadata indexMetadata = context.getClusterState().metadata().index(tableName);
+        if (indexMetadata == null) {
+            throw new IllegalArgumentException("Index [" + tableName + "] not found in cluster state");
+        }
+
+        String primaryFormat = indexMetadata.getSettings().get("index.composite.primary_data_format", "lucene");
```
vinaykpud: Unused primaryFormat

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTableScanRule.java
```diff
@@ -0,0 +1,113 @@
...
+
+        // Viable backends: must support SCAN and be able to read ALL requested fields
+        // (natively or via delegation to another backend that can read the field)
+        List<String> scanCapable = registry.operatorBackends(OperatorCapability.SCAN);
+        List<String> delegationSupporters = registry.delegationSupporters(DelegationType.SCAN);
+        List<String> delegationAcceptors = registry.delegationAcceptors(DelegationType.SCAN);
+        List<String> viableBackends = new ArrayList<>(scanCapable);
+
+        for (FieldStorageInfo field : fieldStorage) {
+            if (field.isDerived()) {
```
vinaykpud, line 73: isn't this check is always false for tablescan?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateRule.java
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * <p>The split into PARTIAL + FINAL is NOT done here. It happens via
+ * {@link OpenSearchAggregateSplitRule} which fires when Volcano detects
+ * a distribution trait mismatch (RANDOM input needing SINGLETON output).
+ *
+ * @opensearch.internal
+ */
+public class OpenSearchAggregateRule extends RelOptRule {
+
+    private static final Logger logger = LogManager.getLogger(OpenSearchAggregateRule.class);
```
vinaykpud: FilterRule uses LOGGER. This uses logger. Minor but inconsistent across the PR.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateRule.java
```diff
@@ -0,0 +1,212 @@
...
+                    + funcNames
+                    + " not supported by any viable backend among "
+                    + childViableBackends
+            );
+        }
+
+        logger.debug("Aggregate viable backends: {} (child viable: {})", viableBackends, childViableBackends);
+
+        RelTraitSet aggregateTraits = child.getTraitSet();
+        if (aggregateTraits.size() > 0) {
```
vinaykpud: do we need this check? I think  Calcite always initializes trait sets with at least the convention trait.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateRule.java
```diff
@@ -0,0 +1,212 @@
...
+        CapabilityRegistry registry = context.getCapabilityRegistry();
+
+        if (aggCall.getArgList().isEmpty()) {
+            return new ArrayList<>(registry.operatorBackends(OperatorCapability.AGGREGATE));
+        }
+
+        List<String> callViable = null;
+        for (int fieldIndex : aggCall.getArgList()) {
+            if (fieldIndex >= childFieldStorage.size()) {
+                continue;
```
vinaykpud: I think if this happens a throw would catch the bug early instead of silently ignoring.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchProjectRule.java
```diff
@@ -0,0 +1,224 @@
...
+            boolean canHandleAll = true;
+            for (RexNode expr : annotatedExprs) {
+                if (!(expr instanceof AnnotatedProjectExpression annotation)) {
+                    continue;
+                }
+                if (annotation.getViableBackends().contains(candidateName)) {
+                    continue;
+                }
+                boolean canDelegate = delegationSupporters.contains(candidateName)
+                    && annotation.getViableBackends().stream().anyMatch(delegationAcceptors::contains);
```
vinaykpud, line 207: this delegation check pattern which checks can delegates repeats across the rules can this be a utility method like `canHandleOrDelegate`?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -0,0 +1,312 @@
...
+                List<String> anyFormat = registry.filterBackendsAnyFormat(operator, fieldType);
+                List<String> delegationAcceptors = registry.delegationAcceptors(DelegationType.FILTER);
+                for (String name : childViableBackends) {
+                    if (anyFormat.contains(name)) {
+                        fieldViable.add(name);
+                    }
+                }
+                // Delegation targets reachable from child viable backends
+                List<String> delegationSupporters = registry.delegationSupporters(DelegationType.FILTER);
+                if (childViableBackends.stream().anyMatch(delegationSupporters::contains)) {
```
vinaykpud: Also this logic of  add delegation targets to a viable list also repeated across the rules, can be extracted to utility

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/CapabilityRegistry.java
```diff
@@ -0,0 +1,303 @@
...
+            if (!caps.arrowCompatibleOperators().isEmpty()) {
+                arrowCompatibleIndex.put(name, caps.arrowCompatibleOperators());
+            }
+        }
+    }
+
+    // ---- Eager lookups ----
+
+    public List<String> operatorBackends(OperatorCapability capability) {
+        return operatorIndex.getOrDefault(capability, List.of());
```
vinaykpud, line 142: nit: Can we consider returning unmodifiable lists from the lookup methods?

expani: Should be fine as we are creating a copy even if the caller chooses to modify it.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
expani: Resolved

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchProjectRule.java
expani: Resolved

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateRule.java
expani: yes throwing an exception instead

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateRule.java
expani: Removed it

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateRule.java
expani: Made it consistent

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTableScanRule.java
expani: Yes throwing an exception instead

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTableScanRule.java
expani: removed

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
expani: removed

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
expani: added

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
expani: renamed

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/OperatorCapability.java
expani: Kept only sort and coordinator reduce for now. 
Will remove these further when we implement Delegation for execution

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
expani: Great catch. 
This remained before I had added Delegation as a capability. 
Removed it as it isn't required anymore.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
expani: To check if the filter function is supported by the backend accepting it. This has been refactored as well so now happens centrally in CapabilityRegistry

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
expani: renamed

bowenlan-amzn (commented):
Putting down the story of what we are trying to achieve across backends.

Take this query as the example:

```sql
SELECT status, message
FROM logs
WHERE status = 200 AND MATCH_PHRASE(message, 'timeout error')
```

The marking phase identifies per-expression capabilities:
- status = 200 → [datafusion, lucene]
- MATCH_PHRASE(message, 'timeout error') → [lucene]

When the logical plan is lowered to a physical plan, a plan rewrite rule sees the incompatible backends and splits the filter into separate operators:

```
DataFusionFilter(status = 200)           → [datafusion]
  └─ DataFusionScan                      → [datafusion]
       └─ LucenePreFilter(MATCH_PHRASE)  → [lucene]
```

The local executor walks:

1. LucenePreFilter: runs MATCH_PHRASE against Lucene's inverted index
   Input:  predicate (field=message, query='timeout error')
   Output: Bitmap (docId → match/no-match)

2. DataFusionScan: reads parquet, skips rows not in bitmap
   Input:  parquet files + Bitmap from step 1
   Output: Arrow RecordBatch (only matching rows)

3. DataFusionFilter: evaluates status = 200
   Input:  Arrow RecordBatch
   Output: Arrow RecordBatch (filtered)

The contract between backends is a Bitmap — Lucene produces it, DataFusion consumes it at scan time as a row selection filter.

expani:
@bowenlan-amzn Thanks for the review. 

I am unable to understand if this is to address something in this existing PR or you just wanted to document the offline discussion. 

Also, `       └─ LucenePreFilter(MATCH_PHRASE)  → [lucene]` Seems incorrectly placed inside the scan. Is it representing a filter pushdown ?

mch2 (approved):
thanks @expani , lgtm.

bowenlan-amzn:
> @bowenlan-amzn Thanks for the review.
> 
> I am unable to understand if this is to address something in this existing PR or you just wanted to document the offline discussion.
> 
> Also, ` └─ LucenePreFilter(MATCH_PHRASE) → [lucene]` Seems incorrectly placed inside the scan. Is it representing a filter pushdown ?

Lucene/OpenSearch is not equivalent to DataFusion — they serve fundamentally different purposes. Treating Lucene/OpenSearch as a general SQL engine in the planner is not justified.

The planner marks and rewrites which operations are suitable for Lucene/OpenSearch. The execution engine is responsible for implementing the specific logic to carry them out. 

Delegation is an execution-layer concern, not a planning-layer concern.

This distinction will become clearer when we get to the execution layer — we can refactor at that point.

expani:
>Lucene/OpenSearch is not equivalent to DataFusion — they serve fundamentally different purposes. Treating Lucene/OpenSearch as a general SQL engine in the planner is not justified.

`OpenSearch(Lucene+ Painless + Mustache + Geo + ....) ` is a subset of DataFusion as it can do Scan, Filter, Project and Aggregate.  Yes it's not a general SQL Engine but treating it as a special case is also wrong. 

Thinking in terms of Relational Algebra most of the DSL Queries can be represented as relational operators. Painless being project, FullText/OtherLuceneQuery being filters; Metric/Bucket/Pipeline aggs being Aggregates;etc. 

This PR handles it by having capabilities where `OpenSearch(Lucene) <the backend plugin>` only declares the ability to run Scan, Filter and Project and DataFusion declares the capabilities similar to a general SQL Engine and will only increase as we implement more features ( Joins, SubQueries, Window; etc ) 

>Delegation is an execution-layer concern, not a planning-layer concern.

It is a concern of both layers. Planner determines the *Feasability* and possible alternative plans that can be generated due to Delegation either for better performance or for correctness. 

It can also determine the delegation strategy suitable in each alternative. 

There will be a data node local planner that picks the best alternative based on the Shard Stats AND ONLY THEN does execution layer comes into picture. 

>This distinction will become clearer when we get to the execution layer

I understand that this PR doesn't show the whole picture and it was intentional to keep the content short. Once we have implemented Pure Parquet Search, I think Delegation can be swiftly started upon both at planner and DataNode Execution.

bowenlan-amzn:
> Thinking in terms of Relational Algebra most of the DSL Queries can be represented as relational operators. Painless being project, FullText/OtherLuceneQuery being filters; Metric/Bucket/Pipeline aggs being Aggregates;etc. This PR handles it by having capabilities where OpenSearch(Lucene) <the backend plugin> only declares the ability to run Scan, Filter and Project and DataFusion declares the capabilities similar to a general SQL Engine

Yes, from the real cases, it's pretty clear Lucene/OpenSearch only provides add-on abilities and DataFusion here is the only execution engine of the plan.

expani:
>Yes, from the real cases, it's pretty clear Lucene/OpenSearch only provides add-on abilities and DataFusion here is the only execution engine of the plan.

That's not completely true. Consider the scenario where there are no parquet files and the entire data is present in Lucene indices and doc values. If a user submits a query similar to 

```
select url, timestamp from http_logs where status = 200 limit 100
```

We will not be using DataFusion as there are no Aggregates so the only execution engine is OpenSearch(Lucene) backend

Even if there is a query with aggregate but consider no shards in the Index have deleted docs and URL is a keyword field
```
select count() from http_logs group by url limit 100
```

Such queries can simply be converted to use the TermsIndex postings in Lucene without ever going via datafusion. 

Datafusion is an accelerator for certain query shapes but definitely not the only execution engine.

bowenlan-amzn:
> That's not completely true. Consider the scenario where there are no parquet files and the entire data is present in Lucene indices and doc values.

If all data is in Lucene and the query shape can be fully executed in Lucene/OpenSearch, the right move is to route to OpenSearch's existing execution path — not to model as a SQL engine within the new analytics planner framework. Because that's what already exists today through the SQL plugin. So I don't think your point of Lucene also acting as an engine stands here.

The analytics engine exists for query shapes that can't be served well by the existing path — large-scale aggregations, joins, columnar scans. That's where DataFusion adds value.

expani:
Okay for the sake of this discussion, creating one abbrevation: 

OpenSearch(Lucene) Backend plugin : LBP

>If all data is in Lucene and the query shape can be fully executed in Lucene, the right move is to route to OpenSearch's existing execution path

The examples I shared above can also be present as a sub-query or as part of a Join ( left or right side ). Then there is no routing possible. 

Take the case
```
SELECT a.url, a.timestamp
   FROM http_logs a
   JOIN http_logs b ON a.url = b.url AND a.timestamp = b.timestamp
   WHERE a.status = 200 AND b.status = 200
   LIMIT 100
```

Equivalent Calcite Logical Plan
```
LogicalSort(fetch=[100])
     LogicalProject(url=[$0], timestamp=[$1])
       LogicalJoin(condition=[AND(=($0, $3), =($1, $4))], joinType=[inner])
         LogicalFilter(condition=[=($2, 200)])
           LogicalTableScan(table=[[http_logs]])
         LogicalFilter(condition=[=($2, 200)])
           LogicalTableScan(table=[[http_logs]])
```

Here DataFusion will help only with Joins , Coordinator Reduce and Sorting whereas the Scan and Filter are done via LBP. We cannot route this to Core OpenSearch rather we will reuse the existing Mappers ( In Core OpenSearch ) for generating queries per field type ***via LBP*** 

As we add support for more field types, query types in LBP then it will be similar to Lucene in Core OpenSearch except the aggregation framework. 

Even for the aggregation framework, we will include the optimizations done as part of [tryPreCompute](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/aggregations/AggregatorBase.java#L236)

Now, the key question is how much time will RelNode conversion and planning add to query latency ? If it's a lot then yes this path cannot be used for everything. But, with plan caching ( similar to prepared statements ) and optimizing the conversion I think we can keep it less than 1ms, let's see how it turns out.

This opens up a clear transition of decoupling Lucene from OpenSearch and making it a backend/engine that can be switched with other libraries like Tantivy if required for certain field types. Also, opens up adding other engines like Velox alongside DataFusion. I am not even covering the use-cases for KNN ( possibly Lance format ), [Time series DB](https://github.com/opensearch-project/time-series-db/) and other workloads yet.

bowenlan-amzn:
> The examples I shared above can also be present as a sub-query or as part of a Join ( left or right side ). Then there is no routing possible.

My previous comment was specifically about your earlier examples (simple filters on Lucene-only data), which can and should be routed. Your new JOIN example goes back to the previous discussion and proves why DataFusion is the engine and Lucene is the data access layer here.

I will summarize in one simple sentence: DataFusion backend executes the plan and calls into Lucene/OpenSearch to perform specific operations and get back data. They are not symmetric backends. 

Happy to move forward with this PR and refine as the execution path lands.

---

> Even for the aggregation framework, we will include the optimizations done as part of tryPreCompute
This opens up a clear transition of decoupling Lucene from OpenSearch and making it a backend/engine that can be switched with other libraries like Tantivy if required for certain field types. Also, opens up adding other engines like Velox alongside DataFusion. I am not even covering the use-cases for KNN ( possibly Lance format ), [Time series DB](https://github.com/opensearch-project/time-series-db/?rgh-link-date=2026-04-15T18%3A15%3A06Z) and other workloads yet.

Agreed on the vision — Tantivy could be represented as special operators if it's Lucene-like, Velox another execution engine. The asymmetry holds regardless of which specific backends are plugged in.

expani:
@bowenlan-amzn My 2 cents on the summarization is that Lucene is a subset of DataFusion where it can handle a subset of relational operators ( Scan, Filter, Project ) and in special cases also handle aggregate ( StarTree, precompute optimizations in current aggregation framework ) 

IMO instead of calling them asymmetric, calling them a subset-superset of one another is a better model. 

Words aside, I think delegation fragment conversion and data node execution orchestration should clear out a lot of questions ( Pure parquet with Lucene indices ).
Post this, the implementation for the use case of pure Lucene Index + DocValues with DataFusion used for accelerating aggregations should clear out the complete structure.

PR #21161, opened 2026-04-08
Fix tiering validation regressions

abishekgiri:
## Summary
- complete hot-to-warm requests after partial validation instead of leaving the listener unresolved
- harden warm-tier capacity validation when disk usage data is missing and preserve deterministic rejection order
- add focused coverage for the transport response path and validator edge cases

## Validation
- JAVA_HOME=/Users/abishekkumargiri/Library/Java/JavaVirtualMachines/openjdk-21.0.1/Contents/Home ./gradlew :server:precommit :server:test --tests "org.opensearch.indices.tiering.TieringRequestValidatorTests" --tests "org.opensearch.action.admin.indices.tiering.TransportHotToWarmTieringActionUnitTests"

### DIFF
--- a/server/src/main/java/org/opensearch/action/admin/indices/tiering/TransportHotToWarmTieringAction.java
+++ b/server/src/main/java/org/opensearch/action/admin/indices/tiering/TransportHotToWarmTieringAction.java
@@ -105,11 +105,7 @@ public class TransportHotToWarmTieringAction extends TransportClusterManagerNode
             clusterInfoService.getClusterInfo(),
             diskThresholdSettings
         );
-
-        if (tieringValidationResult.getAcceptedIndices().isEmpty()) {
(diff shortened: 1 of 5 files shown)


abishekgiri:
Adding a bit more reviewer context here:
- fixes the transport path so partially valid hot-to-warm requests always return a response
- makes warm-tier capacity validation resilient when disk usage data is unavailable and keeps rejection ordering deterministic
- adds focused coverage for the transport action and validator edge cases

Validated with:
JAVA_HOME=/Users/abishekkumargiri/Library/Java/JavaVirtualMachines/openjdk-21.0.1/Contents/Home ./gradlew :server:precommit :server:test --tests "org.opensearch.indices.tiering.TieringRequestValidatorTests" --tests "org.opensearch.action.admin.indices.tiering.TransportHotToWarmTieringActionUnitTests"

PR #21165, opened 2026-04-08
Add partition strategy setting and partition assignment for flexible shard-to-partition mapping

### LINKED ISSUE #21259: [Feature Request] Add partition_strategy setting and PartitionAssignment utility

Pull-based ingestion enforces a strict 1:1 mapping between shards and source partitions — shard N consumes partition N. There is no configuration to control this behavior or to allow a shard to consume from multiple partitions. Users who need fewer shards than source partitions must maintain external infrastructure to re-partition.

Introduce a PartitionStrategy enum (fixed/auto) on IngestionSource, a new `index.ingestion_source.partition_strategy` index setting, and a PartitionAssignment utility that computes modulo-based partition-to-shard assignment (partition % numShards == shardId). Include  validation warning for multi-shard auto without key-partitioned source.

### Related component

_No response_

_No response_

### Additional context

_No response_

imRishN:
### Description
Introduces the foundation for flexible shard-to-partition mapping in pull-based ingestion. Currently, each shard is hardcoded to consume from exactly one source partition (shard N = partition N), forcing users to maintain intermediate topics when source partition count differs from shard count.

This adds:
  - SourcePartitionStrategy enum (SIMPLE/MODULO) on IngestionSource
  - index.ingestion_source.source_partition_strategy setting (default: simple)
  - SourcePartitionAssignment utility for computing partition-to-shard assignment

SIMPLE preserves current 1:1 behavior. MODULO assigns partitions using modulo-based distribution (partition P -> shard P % numShards), enabling a single shard to consume from multiple source partitions.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/21259

### Check List


server/src/main/java/org/opensearch/cluster/metadata/IndexMetadata.java
```diff
@@ -960,6 +960,20 @@ public Iterator<Setting<?>> settings() {
         Property.Final
     );
 
+    /**
+     * Defines the strategy for mapping source stream partitions to OpenSearch shards.
+     * "fixed" (default): 1:1 mapping where shard N consumes partition N.
+     * "auto": each shard consumes all partitions where partition % numShards == shardId.
```
imRishN: Note - Even after concurrent poller https://github.com/opensearch-project/OpenSearch/issues/21266, the shard will consume all partitions where partition % numShards == shardId, whether via 1 poller or K pollers is an internal detail

server/src/main/java/org/opensearch/indices/pollingingest/SourcePartitionAssignment.java
```diff
@@ -0,0 +1,83 @@
...
+        if (numSourcePartitions <= 0) {
+            throw new IllegalArgumentException("Number of source partitions must be positive, got: " + numSourcePartitions);
+        }
+        if (shardId < 0 || shardId >= numShards) {
+            throw new IllegalArgumentException("Shard ID [" + shardId + "] must be >= 0 and < numShards [" + numShards + "]");
+        }
+        // TODO - support "RANGE" below when we implement https://github.com/opensearch-project/OpenSearch/issues/21267
+        switch (strategy) {
+            case FIXED:
+                if (shardId >= numSourcePartitions) {
```
monika48, line 52: I know this is the existing behavior, but I've seen people miss this and then spend a while wondering why their data isn't showing up in the index. 
Could we surface a warning at index creation when FIXED is chosen with numSourcePartitions > numShards? Something in validateIngestionSourceSettings that says "you have N partitions but only M shards, partitions M..N-1 won't be consumed

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -1300,6 +1300,32 @@ static void validateIngestionSourceSettings(Settings settings, ClusterState stat
 
         // Settings validation to the mapper
         IngestionMessageMapper.validateSettings(mapperType, mapperSettings);
+
+        // Partition strategy validation
+        if (IndexMetadata.INGESTION_SOURCE_PARTITION_STRATEGY_SETTING.exists(settings)) {
+            IngestionSource.PartitionStrategy partitionStrategy = IndexMetadata.INGESTION_SOURCE_PARTITION_STRATEGY_SETTING.get(settings);
+            if (partitionStrategy == IngestionSource.PartitionStrategy.AUTO) {
```
monika48: The version check only fires for AUTO, but the setting key index.ingestion_source.partition_strategy itself is brand new. In a mixed-version cluster during a rolling upgrade, if this setting (even =fixed) ends up in index metadata that gets replicated to a 3.6 node, won't that node reject it as an unknown setting key?

monika48: I think we need to guard the presence of the setting, not just the AUTO value.
Two options I can think of:
1. Move the minNodeVersion.before(V_3_7_0) check out to the outer if, so any use of this setting is rejected on a mixed cluster.
2. Only persist the setting when it's AUTO (non default), so old nodes never see the key unless the cluster is fully upgraded.

server/src/main/java/org/opensearch/cluster/metadata/IngestionSource.java
```diff
@@ -203,9 +213,44 @@ public String toString() {
...
+            + '\''
             + '}';
     }
 
+    /**
+     * Strategy for mapping source stream partitions to OpenSearch shards.
+     */
+    public enum PartitionStrategy {
+        FIXED("fixed"),
+        AUTO("auto");
```
varunbharadwaj: We were discussing the alternatives like static strategy where user defines the mappings. Wondering if we need a better name for supporting new strategies in the future? `auto` might be confusing if we have new options in future. What do you think? Thinking if the current one-to-one can be called as DEFAULT, and then have MODULO as the other one. We can have a CUSTOM for users specifying the mappings. We can replace these if we are able to come up with better names.

Can we also finalize on the pros and cons of the different strategies we discussed to see if modulo based approach can work for us before adding it here? Mainly on new partition visibility with coordination across shards and capability to re-use existing partitions when custom sharding is used.

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -1300,6 +1300,32 @@ static void validateIngestionSourceSettings(Settings settings, ClusterState stat
...
+                Version minNodeVersion = state.nodes().getMinNodeVersion();
+                if (minNodeVersion.before(Version.V_3_7_0)) {
```
varunbharadwaj: As partition strategy is a static setting that does not change during the life of an index, this validation is helping us in case an user creates a new index mid-rollout?

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -1300,6 +1300,32 @@ static void validateIngestionSourceSettings(Settings settings, ClusterState stat
...
+                            + Version.V_3_7_0
+                            + "] or later, but the minimum node version is ["
+                            + minNodeVersion
+                            + "]"
+                    );
+                }
+                int numShards = IndexMetadata.INDEX_NUMBER_OF_SHARDS_SETTING.get(settings);
+                if (numShards > 1) {
+                    logger.warn(
+                        "partition_strategy [auto] with [{}] shards requires the source topic to be key-partitioned "
```
varunbharadwaj: >partition_strategy [auto] with [{}] shards requires the source topic to be key-partitioned by the document ID field for correctness

If one documentID is written to multiple partitions, it is also a problem in the current 'fixed' mode right? This is more a general pull-based ingestion warning?

server/src/main/java/org/opensearch/indices/pollingingest/PartitionAssignment.java
```diff
@@ -0,0 +1,83 @@
...
+     * @param strategy            the partition assignment strategy
+     * @return unmodifiable list of partition IDs assigned to this shard
+     * @throws IllegalArgumentException if numSourcePartitions is less than numShards for FIXED strategy,
+     *                                  or if no partitions are assigned to the shard
+     */
+    public static List<Integer> assignPartitions(int shardId, int numShards, int numSourcePartitions, PartitionStrategy strategy) {
+        if (numSourcePartitions <= 0) {
+            throw new IllegalArgumentException("Number of source partitions must be positive, got: " + numSourcePartitions);
+        }
+        if (shardId < 0 || shardId >= numShards) {
```
varunbharadwaj: will we ever have shardId < 0 || shardId >= numShards? If this is for just extra safety, we can maybe leave a comment mentioning it.

server/src/main/java/org/opensearch/indices/pollingingest/SourcePartitionAssignment.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param shardId             the OpenSearch shard ID
+     * @param numShards           total number of shards in the index
+     * @param numSourcePartitions total number of partitions in the source stream
+     * @param strategy            the partition assignment strategy
+     * @return unmodifiable list of partition IDs assigned to this shard
+     * @throws IllegalArgumentException if numSourcePartitions is less than numShards for FIXED strategy,
+     *                                  or if no partitions are assigned to the shard
+     */
+    public static List<Integer> assignPartitions(int shardId, int numShards, int numSourcePartitions, PartitionStrategy strategy) {
```
varunbharadwaj: How do we plan to retrieve numSourcePartitions? Do we lookup the streaming source?

server/src/main/java/org/opensearch/indices/pollingingest/PartitionAssignment.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Computes which source stream partitions a given OpenSearch shard should consume,
+ * based on the configured {@link PartitionStrategy}.
+ */
+public class PartitionAssignment {
```
varunbharadwaj: Maybe `PartitionAssignment` naming is fine, but let us discuss and see if this works for both Kafka and Kinesis. We can double check the naming convention used earlier and follow the same. Maybe something like `StreamPartition` or `StreamShard` instead of `Partition` can be used. We can discuss further.

server/src/main/java/org/opensearch/indices/pollingingest/SourcePartitionAssignment.java
imRishN: Good suggestion. The precise check needs `IngestionConsumerFactory.getSourcePartitionCount()`, which is added in the multi-partition consumer factory PR (#21222). Index creation runs on the cluster manager, which doesn't have the consumer factory connection initialize, the partition count is only known once a data node sets up its consumer for shard initialization. Adding a TODO for now at validation site and will have this once we wire the dependency

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
imRishN: Great catch. Moved the version check to guard the presence of setting

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
imRishN: Yup, exactly. Without the check, a new index created on a 3.7 node during rollout would replicate metadata containing the unknown setting to 3.6 nodes, and would fail to open the index with unknown setting error. This guards helps there

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
imRishN: Yup, moved warning a bit up to check always while using pull based

server/src/main/java/org/opensearch/indices/pollingingest/PartitionAssignment.java
imRishN: Yes, updating

server/src/main/java/org/opensearch/indices/pollingingest/SourcePartitionAssignment.java
imRishN: Yes, adding this in next PR - #21222

server/src/main/java/org/opensearch/indices/pollingingest/PartitionAssignment.java
imRishN: Good point, I was thinking same. Its not public api so thought it should be fine as its just implementation details. But to avoid confusion later, I will rename this to StreamPartitionAssignment

server/src/main/java/org/opensearch/cluster/metadata/IngestionSource.java
imRishN: We can rename auto to `modulo`. `Default` for 1:1 mapping can get misleading in future as feature matures because users would usually tend to use default if they are not aware of full details. We can keep fixed for 1:1. Wdyt?

On different strategies like - modulo, range, custom, etc - 

On modulo - works really well and is correct for common use cases, but with increasing shard count - every shard's partition assignment shifts. And there's no clean way to "split" partition data across new shards without re-reading from the source.              

"Range" would solve this naturally - as these are subsets and splitting a range maintains existing assignments for everything outside the split. And adding a partition extends the rightmost range with no other shard's mapping changes.

We can support both and let the user decide what works best for them. Lmk your thoughts here?

server/src/main/java/org/opensearch/cluster/metadata/IngestionSource.java
```diff
@@ -62,7 +64,8 @@ private IngestionSource(
         TimeValue pointerBasedLagUpdateInterval,
         IngestionMessageMapper.MapperType mapperType,
         Map<String, Object> mapperSettings,
-        WarmupConfig warmupConfig
+        WarmupConfig warmupConfig,
+        PartitionStrategy partitionStrategy
```
varunbharadwaj: Should this be SourcePartitionStrategy as per the new naming?

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -1277,6 +1277,44 @@ private static void validateSearchOnlyReplicasSettings(Settings indexSettings) {
...
+                        + "appears in multiple source partitions, it may be indexed into multiple shards.",
+                    numShards
+                );
+            }
+        }
+
+        // Partition strategy validation. The setting key itself was introduced in V_3_7_0; reject any explicit
+        // value (including [simple], the default) on mixed clusters where some nodes don't
+        // recognize the key — otherwise index metadata replicated to older nodes would carry
+        // an unknown setting.
```
varunbharadwaj: >otherwise index metadata replicated to older nodes would carry an unknown setting

I think carrying unknown setting is fine, but the real issue this validation solves is when using different partition strategies, the older version nodes might not be aware of it and fallback to the default 1-to-1 mapping, reading from an incorrect partition, till the node is upgraded. We can maybe update the comment to highlight this.

server/src/main/java/org/opensearch/cluster/metadata/IngestionSource.java
imRishN: yeah, updated

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
imRishN: updated

varunbharadwaj (approved):

PR #19403, opened 2025-09-24
Asynchronous FlightServerChannel to avoid thread contention per request

rishabhmaurya:
### Description
Addresses contention per request when multiple threads are flushing the response to the same `FlightServerChannel`. This can happen in search path when concurrent segment search is enabled where multiple threads will try to flush corresponding batches to the `FlightServerChannel`. This can cause thread to be blocked as it is currently protected under synchronized block.

**Solution**
The better way to handle it is to process the flushing of response, lets call it `BatchTask`, asynchronously. This can be done by having single threaded executor per request, which can keep adding `BatchTask` from multiple threads for the given request to its queue and process them linearly. 

- Added event loop group which maintains set of single threaded executors with configurable count. 
- one event loop from event loop group is assigned to the `FlightServerChannel` in a round robin way when a request is received at the server. This guarantees linear processing.
- One event loop can be associated with multiple channels. This is to avoid too many single threaded executor service to be created.
- All type of events on `FlightServerChannel` i.e. regular batch flush, complete stream and error are handled asynchronously now.
- Since its asynchronous, we need to pass the `FlightTransportChannel` to the `BatchTask` to release it when the stream is either completed or results in an error, which was not the case before and could be released in the finally block.

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [x] Functionality includes testing.
~~- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.~~
~~- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.~~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21167, opened 2026-04-08
Fix FlightOutboundHandler clearing caller's ThreadContext

Hailong-am:
### Description

`FlightOutboundHandler` uses `stashContext()` to capture the thread context before dispatching work to a plain `ExecutorService`. However, `stashContext()` clears the calling thread's context as a side effect. Since the context is only restored on the executor thread, the caller loses its `ThreadContext` (security headers, request metadata, etc.) after calling `sendResponseBatch()`, `completeStream()`, or `sendErrorResponse()`.

This replaces the manual `stashContext()` / `restore()` pattern with `ThreadContext.preserveContext()`, which is the idiomatic OpenSearch mechanism for cross-thread context propagation. It:
- Captures context at wrap time via `newStoredContext(false)` **without** clearing the caller
- At run time, stashes the executor thread's context, restores the captured context, runs the task, then cleans up

### Changes

**`FlightOutboundHandler.java`:**
- Replace `stashContext()` with `preserveContext()` wrapping the executor lambdas in `sendResponseBatch()`, `completeStream()`, and `sendErrorResponse()`
- Remove `StoredContext` field from `BatchTask` record
- Remove manual `storedContext.restore()` calls from `processBatchTask()`, `processCompleteTask()`, `processErrorTask()`

**`FlightOutboundHandlerTests.java` (new):**
- Unit tests verifying caller's `ThreadContext` is preserved after each method call
- Tests context propagation to executor thread via `preserveContext()`
- Tests context preservation across multiple batch sends

**`FlightOutboundHandlerContextPropagationTests.java` (new):**
- Integration tests using real `FlightTransport` and `StreamTransportService`
- End-to-end verification of context propagation through stream response batches
- Tests context preservation through error response paths

**`FlightTransportChannelTests.java`:**
- Updated `BatchTask` constructor call to match new signature (removed `storedContext` parameter)

### Related Issues

Resolves #21166

https://github.com/opensearch-project/OpenSearch/pull/19403

### Check List

For more information on following Developer Certification of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/plugins/arrow-flight-rpc/src/test/java/org/opensearch/arrow/flight/transport/FlightOutboundHandlerTests.java
+++ b/plugins/arrow-flight-rpc/src/test/java/org/opensearch/arrow/flight/transport/FlightOutboundHandlerTests.java
@@ -1,0 +1,213 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.arrow.flight.transport;
+
+import org.opensearch.Version;
+import org.opensearch.common.util.concurrent.ThreadContext;
+import org.opensearch.core.transport.TransportResponse;
+import org.opensearch.test.OpenSearchTestCase;
+import org.opensearch.threadpool.TestThreadPool;
+import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.transport.StatsTracker;
+import org.opensearch.transport.TransportMessageListener;
+import org.junit.After;
+import org.junit.Before;
+
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class FlightOutboundHandlerTests extends OpenSearchTestCase {
+
+    private ThreadPool threadPool;
+    private FlightOutboundHandler handler;
+    private ExecutorService executor;
+    private FlightServerChannel mockFlightChannel;
+    private TransportMessageListener mockListener;
+
+    private static final String HEADER_KEY = "test-header";
+    private static final String HEADER_VALUE = "test-value";
+
+    @Before
+    @Override
+    public void setUp() throws Exception {
+        super.setUp();
+        threadPool = new TestThreadPool(getTestName());
+        executor = Executors.newSingleThreadExecutor();
+        handler = new FlightOutboundHandler("test-node", Version.CURRENT, new String[0], new StatsTracker(), threadPool);
+
+        mockFlightChannel = mock(FlightServerChannel.class);
+        when(mockFlightChannel.getExecutor()).thenReturn(executor);
+
+        mockListener = mock(TransportMessageListener.class);
+        handler.setMessageListener(mockListener);
+    }
+
+    @After
+    @Override
+    public void tearDown() throws Exception {
+        executor.shutdown();
+        assertTrue(executor.awaitTermination(5, TimeUnit.SECONDS));
+        threadPool.shutdown();
+        super.tearDown();
+    }
+
+    public void testSendResponseBatchPreservesCallerThreadContext() throws Exception {
+        ThreadContext threadContext = threadPool.getThreadContext();
+        threadContext.putHeader(HEADER_KEY, HEADER_VALUE);
+
+        CountDownLatch latch = new CountDownLatch(1);
+        doAnswer(invocation -> {
+            latch.countDown();
+            return null;
+        }).when(mockListener).onResponseSent(anyLong(), anyString(), any(TransportResponse.class));
+
+        handler.sendResponseBatch(
+            Version.CURRENT,
+            Collections.emptySet(),
+            mockFlightChannel,
+            mock(FlightTransportChannel.class),
+            1L,
+            "test-action",
+            mock(TransportResponse.class),
+            false,
+            false
+        );
+
+        // Verify the caller's thread context is NOT cleared
+        assertEquals(
+            "Caller's thread context should be preserved after sendResponseBatch",
+            HEADER_VALUE,
+            threadContext.getHeader(HEADER_KEY)
+        );
+    }
+
+    public void testCompleteStreamPreservesCallerThreadContext() throws Exception {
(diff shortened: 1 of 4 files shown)


Hailong-am:
@rishabhmaurya would you mind to help to review this fix?

Hailong-am:
Breaking change brought in by PR https://github.com/opensearch-project/OpenSearch/pull/21039/changes#diff-e13f61d0a17c4541e43c1ade6f9b452696340e76d728970322fa3b85dd9b5aaa, Not related to my change.

rishabhmaurya:
logically it makes sense. I will check the code in details later in the day. 
I'm curious what is your use case?

Hailong-am:
> logically it makes sense. I will check the code in details later in the day. I'm curious what is your use case?

@rishabhmaurya thanks. My case is we are building streaming experience for chat and sending back tool call result to user in the middle of whole chat flow, so the flow is tool A called and send back toolA's result and continue reasoning and call other tools until we got a final answer.  To achieve sending tool result back we leverage FlightOutBoundHandler and we still want to keep the thread context as it contains information which need to tool call.

rishabhmaurya:
@Hailong-am thanks for the explanation, makes a lot of sense. Usually in search case, the thread executes an action and rest of the logic is handled by action listener once response is received. I'm taking a look now

rishabhmaurya:
@Hailong-am are you restoring the preserved context once response is received so that response handler has the context? From the initial looks of it, it doesn't seems like that's the case.

Hailong-am:
> @Hailong-am are you restoring the preserved context once response is received so that response handler has the context? From the initial looks of it, it doesn't seems like that's the case.


 @rishabhmaurya Yes, the context is restored on the executor thread before the response is processed. Here's the flow:                                                                                   
                                                                                                                                                                                                         
   `preserveContext()` wraps the runnable with a `ContextPreservingRunnable` which, at execution time:
   1. Stashes the executor thread's existing context (clean slate)
   2. Restores the caller's captured context onto the executor thread
   3. Runs the task — so `processBatchTask` / `processCompleteTask` / `processErrorTask` all see the correct `ThreadContext`
   4. Restores the executor thread's original context when done

   This matters because `getHeaderBuffer()` reads from `threadPool.getThreadContext()` to serialize headers into the Flight response (line 288). With `preserveContext`, that call sees the caller's
   context, so the response carries the correct headers.

   The reason we need this explicitly is that `FlightServerChannel` uses a plain JDK `ExecutorService`, not OpenSearch's `ThreadPool.executor()` which wraps runnables with context propagation automatically.

Hailong-am:
@rishabhmaurya Yes, the context is restored on the executor thread before the response is processed. Here's the flow:

`preserveContext()` wraps the runnable with a `ContextPreservingRunnable` which, at execution time:
1. Stashes the executor thread's existing context (clean slate)
2. Restores the caller's captured context onto the executor thread
3. Runs the task — so `processBatchTask` / `processCompleteTask` / `processErrorTask` all see the correct `ThreadContext`
4. Restores the executor thread's original context when done

This matters because `getHeaderBuffer()` reads from `threadPool.getThreadContext()` to serialize headers into the Flight response (line 288). With `preserveContext`, that call sees the caller's context, so the response carries the correct headers.

The reason we need this explicitly is that `FlightServerChannel` uses a plain JDK `ExecutorService`, not OpenSearch's `ThreadPool.executor()` which wraps runnables with context propagation automatically.

---

### Old Flow (`stashContext`) — broken

```
Caller Thread                          Executor Thread
─────────────                          ───────────────
ThreadContext: {key: "value"}
        │
        ▼
stashContext()
  ├─ captures snapshot ──────────────▶ StoredContext
  └─ CLEARS caller's context ✗
        │
ThreadContext: {} ← EMPTY!             
        │                              
        ▼                              
submit task to executor ──────────────▶ executor.execute(() -> {
        │                                  │
        ▼                                  ▼
caller continues with                  storedContext.restore()
EMPTY context ✗                        ThreadContext: {key: "value"} ✓
                                           │
                                           ▼
                                       processBatchTask()
                                       getHeaderBuffer() ← sees correct context
                                           │
                                           ▼
                                       task.close()
                                       storedContext.close()
                                       ThreadContext: ??? ← stale, no cleanup
```

**Problems:**
1. Caller loses context immediately after `stashContext()`
2. Executor thread's pre-existing context not cleaned up before `restore()`
3. Executor thread context not restored after task completes

---

### New Flow (`preserveContext`) — fixed

```
Caller Thread                          Executor Thread
─────────────                          ───────────────
ThreadContext: {key: "value"}
        │
        ▼
preserveContext(runnable)
  └─ newStoredContext(false)
     captures snapshot ──────────────▶ ContextPreservingRunnable
     does NOT clear caller ✓               │
        │                                  │
ThreadContext: {key: "value"} ✓            │
        │                                  │
        ▼                                  │
submit wrapped task to executor ──────────▶ executor.execute(wrapped -> {
        │                                  │
        ▼                                  ▼
caller continues with                  1. stashContext()
INTACT context ✓                          └─ saves executor's own context
                                          └─ clears executor thread (clean slate)
                                           │
                                           ▼
                                       2. restore() caller's captured context
                                          ThreadContext: {key: "value"} ✓
                                           │
                                           ▼
                                       3. processBatchTask()
                                          getHeaderBuffer() ← sees correct context
                                           │
                                           ▼
                                       4. try-with-resources closes stashed context
                                          └─ executor thread's original context restored
                                          ThreadContext: {original executor ctx} ✓
```

**Fixed:**
1. Caller's context preserved throughout
2. Executor thread starts clean before restoring caller's context
3. Executor thread's own context restored after task completes

plugins/arrow-flight-rpc/src/test/java/org/opensearch/arrow/flight/transport/FlightOutboundHandlerTests.java
```diff
@@ -0,0 +1,206 @@
...
+
+        assertEquals(
+            "Caller's thread context should be preserved after sendErrorResponse",
+            HEADER_VALUE,
+            threadContext.getHeader(HEADER_KEY)
+        );
+    }
+
+    public void testSendResponseBatchPropagatesContextToExecutorThread() throws Exception {
+        ThreadContext threadContext = threadPool.getThreadContext();
```
rishabhmaurya, line 139: here are we asserting the propagated header?

rishabhmaurya:
@Hailong-am LGTM, thanks for fixing it. One minor comment on the test.

rishabhmaurya (approved):

plugins/arrow-flight-rpc/src/test/java/org/opensearch/arrow/flight/transport/FlightOutboundHandlerTests.java
rishabhmaurya: you can take this up as a follow up

Hailong-am: @rishabhmaurya here is the follow up PR to address the unit test https://github.com/opensearch-project/OpenSearch/pull/21200

Issue #21025, opened 2026-03-27
[Feature Request] Add a size limit on BitsetFilterCache


sgup432:
### Is your feature request related to a problem? Please describe

In one of the issue we encountered, we saw BitsetFilterCache consumed almost ~29gb of heap and caused cluster instability due to node drops. BitsetFilterCache is unbounded by default. It is used primarily for nested document queries. Caches Bitset objects to identify parent vs child documents within lucene segments.

### Describe the solution you'd like

Put a size limit: Some X% of heap. Not beyond 10% IMO

### Related component

Search:Resiliency

### Describe alternatives you've considered

1. Other alternative would be to use tiered cache for this. Have a larger disk cache and a smaller on-heap cache. Challenges involves serializing Query object.

2. Disable the cache? Compute results on the fly. Or store the parent/child information in the disk/lucene itself as a metadata maybe?

sgup432:
It is important to note that this cache is a two level cache with structure `Cache<Segment, Cache<Query, Value>>`, so don't think we can use the internal `Cache` size limit here to impose max size. Options are to either implement a custom cache implementation for this with our own size tracking logic OR flatten the cache to Cache<segment + query, V>(this would also solve the issue mentioned [here](https://github.com/opensearch-project/OpenSearch/issues/20190))

ShawnQiang1:
This cache appears to be index-level, so we also need to consider having a global state management system. Perhaps we could refer to IndicesFieldDataCache for implementation?

sgup432:
@ShawnQiang1 This cache is actually operates on a lucene segment level and not index-level, as the cache key is tied to a segment. So in that way, it is more closer to lucene's query cache implementation than IndicesFieldDataCache. 

I already have a [PR](https://github.com/apache/lucene/pull/15558/) out to flatten(and optimize) the lucene's query cache in lucene. We can probably do similar for this as well which will allow to put a max size limit as well. Or find an easier route with the current structure to cap the max size for now.

andrross:
Thanks @sgup432. An unbounded cache seems bad! What are the next steps here?

sgup432:
@andrross We need to make a code change to put a size limit on this cache as suggested above. I will try finding sometime for this to get a PR out within a week.

PR #21179, opened 2026-04-08
BitsetFilterCache refactor: Change it to a node level cache with a size limit

sgup432:
### Description

Earlier, BitsetFilterCache was created per index and without any size limit for each or all of the caches combined.

- This PR changes this cache to a node level cache shared across the segment, and then puts a size limit. Defaults to 5% of the total heap available on the node, also a static node level setting.

- It is also changing(or flattening) this cache structure from `Cache<K, Cache<K1, V1>>` to `Cache<CK, V>` where CK is a composite key `<Segment, Query>`.  This logic was needed to be able to put a size limit on the cache.
Other way to handle this would have been to implement our own custom cache and manage our own logic to calculate the size and impose a limit, but that's a even bigger change compared to the current approach taken.

We have now `IndicesBitsetFilterCache`, a node level cache. `BitsetFilterCache` is still used at an index level, but it eventually uses `IndicesBitsetFilterCache` to store results, so it is only a thin layer now. Kept it to avoid backward compatibility issues.

### Related Issues
https://github.com/opensearch-project/OpenSearch/issues/21025

### Check List


server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
```diff
@@ -0,0 +1,403 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.ToLongBiFunction;
+
+/**
+ * Node-level cache for {@link BitSet} based filters. Manages a single flat cache shared across
+ * all indices on the node, with a configurable size limit and async stale entry cleanup.
+ *
+ * @opensearch.api
+ */
+@ExperimentalApi
```
jainankitk, line 71: `@ExperimentalApi` on the new class but `@PublicApi` on the setting constant. `INDEX_LOAD_RANDOM_ACCESS_FILTERS_EAGERLY_SETTING` is delegated from `BitsetFilterCache` (which is `@PublicApi`) to `IndicesBitsetFilterCache`
(which is `@ExperimentalApi`). Can this cause confusion for the users?

server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
```diff
@@ -0,0 +1,403 @@
...
+        Set<IndexReader.CacheKey> staleSnapshot = new HashSet<>(staleCacheKeys);
+        staleCacheKeys.removeAll(staleSnapshot);
+        registeredKeys.removeAll(staleSnapshot);
+
+        for (BitsetCacheKey key : cache.keys()) {
+            if (staleSnapshot.contains(key.readerCacheKey)) {
+                cache.invalidate(key);
+            }
+        }
+    }
```
jainankitk, line 199: This method iterates over all cache keys which is O(n) over the entire cache. On a node with many indices and segments, this could be expensive. The old code was O(1) — just `loadedFilters.invalidate(coreCacheReader)`. Should we C]consider maintaining a reverse index from `CacheKey` → `Set<BitsetCacheKey>` for O(1) purge.

server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
```diff
@@ -0,0 +1,403 @@
...
+    @Override
+    public void onClose(IndexReader.CacheKey ownerCoreCacheKey) {
+        staleCacheKeys.add(ownerCoreCacheKey);
+    }
+
+    @Override
+    public void close() {
+        cacheCleaner.close();
+        clear();
+    }
```
jainankitk, line 168: The old code called `loadedFilters.invalidateAll()` on index close. There should be a 
mechanism for index-level cleanup — perhaps `IndicesBitsetFilterCache` should expose a `clearForIndex(Index)` method.

jainankitk (commented):
I noticed that code reviewer added some concerns here - https://github.com/opensearch-project/OpenSearch/pull/21179#issuecomment-4210212716, in addition to my comments. Also, I am wondering if it is worthwhile breaking this down into 2 PRs (3 might be too much)?

1. Sub-PR theme: Introduce IndicesBitsetFilterCache node-level cache with size limit
1. Sub-PR theme: Wire IndicesBitsetFilterCache into IndexService and IndicesService + Sub-PR theme: Update test infrastructure to use IndicesBitsetFilterCache

server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
sgup432: >This method iterates over all cache keys which is O(n) over the entire cache. 

Yeah this is intentional as the cache structure has changed. Also note that this happens on an async thread, and it is similar to how Request cache works.

server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
sgup432: I didn't remove it from BitsetFilterCache as it would be a breaking change.
So perhaps, I can instead mark `BitsetFilterCache.INDEX_LOAD_RANDOM_ACCESS_FILTERS_EAGERLY_SETTING` as deprecated. What do you think?

server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
sgup432: Actually we don't need this i.e. doing explicit index level clean up on `IndicesBitsetFilterCache` now.

Earlier, we were creating one `BitsetFilterCache` per index. So once index was closed, we called `close()` or `invalidateAll()` to clear up the cache as it is not needed.

Now with this change, we don't have a cache per index. Instead we maintain a node-level common cache shared across indices. Once the index is closed, lucene will trigger close() on segment readers, and we will collect those stale readers in a set, and clear those in background eventually during a periodic cleanup.

So in a nutshell, we handle index level cleanup now via background scheduler. This `close()` is basically to clear the whole node level cache which happens during node shutdown.

server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
sgup432: I have marked `BitsetFilterCache.INDEX_LOAD_RANDOM_ACCESS_FILTERS_EAGERLY_SETTING` as deprecated. So we are good on this one.

server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
sgup432: >On a node with many indices and segments, this could be expensive. The old code was O(1) — just loadedFilters.invalidate(coreCacheReader)

Again to clarify, the older code was O(1) because we had one cache per index. So we have a common cache for all the indices. Plus the cache key structure has changed to `Cache<<Segment, Query>, Value>>` to efficiently calculate the size.  So we don't have a way now to do O(1) cleanups. And as mentioned above, we do background cleanups on async thread, so current search threads won't be blocked as such.

>Should we C]consider maintaining a reverse index from CacheKey → Set<BitsetCacheKey> for O(1) purge.

To be honest, it is not worth it. By maintaining this reverse structure would take additional memory. Plus requires extra handling as the reverse index must be kept in sync with the primary cache atomically. And as mentioned above, the background thread is simple, correct, and fast enough for realistic cache sizes(5% of heap as this one).

jainankitk (approved):

PR #21181, opened 2026-04-09
Add Data Format aware engine

mgodwan:
### Description

Introduces `DataFormatAwareEngine` — a Lucene-decoupled indexing engine that enables pluggable data formats in OpenSearch. When the feature flag `opensearch.experimental.feature.pluggable.dataformat.enabled` and index setting `index.pluggable.dataformat.enabled` are both set to `true`, `IndexShard` uses `DataFormatAwareEngine` (via `DataFormatAwareIndexerFactory`) instead of the traditional `EngineBackedIndexer` + `InternalEngine` path.

`DataFormatAwareEngine` implements the `Indexer` interface and orchestrates the same distributed system concerns as `InternalEngine` — sequence number management, translog durability, refresh/flush lifecycle, checkpoint tracking — but delegates all format-specific work through pluggable abstractions: `IndexingExecutionEngine`, `Committer`, `EngineReaderManager`, and `CatalogSnapshotManager`.

The `CompositeIndexingExecutionEngine` (in the composite-engine sandbox plugin) orchestrates across a primary data format (currently Parquet) and optional secondary formats (Lucene in future). Each format plugin provides its own `IndexingExecutionEngine` implementation via the `DataFormatRegistry`.

- `DataFormatAwareEngine` — format-agnostic engine implementing `Indexer`, owns seq-no tracking, translog, writer pool, catalog snapshot lifecycle
- `IndexerFactory` / `DataFormatAwareIndexerFactory` / `EngineBackedIndexerFactory` — factory abstraction for choosing between DFAE and the legacy Lucene path
- `IndexingThrottler` — throttling support for the pluggable engine

`DataFormatAwareEngineTests` covers:
- Seq-no management: primary assignment, replica out-of-order delivery, checkpoint advancement, gap handling with stall/fill
- Translog: ops written, sync persists checkpoint, recovery-origin skips translog write, `newChangesSnapshot` reads ops back, `countNumberOfHistoryOperations`
- Refresh: catalog snapshot creation with exact segment/file/numRows verification, generation advancement, segment accumulation, `lastRefreshedCheckpoint` tracking
- Flush: catalog snapshot commit, `shouldPeriodicallyFlush`, `writeIndexingBuffer` triggers refresh, non-waiting flush concurrency
- Reader acquisition: format-specific readers present after refresh, snapshot consistency, concurrent access, lifecycle
- Concurrency: concurrent indexing (seq-no correctness), concurrent refresh+flush, end-to-end stress
- Lifecycle: close, flushAndClose, failEngine, operations after close/fail throw
- Stats: `getSeqNoStats` after sync, persisted vs processed checkpoint divergence

### Related Issues

### Check List


server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -8,59 +8,934 @@
...
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    private Engine.IndexResult indexIntoEngine(Engine.Index index) throws IOException {
+        Engine.IndexResult indexResult;
+        if (index.origin() == Engine.Operation.Origin.PRIMARY) {
+            // Assign a sequence number on the primary
+            final long seqNo = localCheckpointTracker.generateSeqNo();
+            index = new Engine.Index(
+                index.uid(), index.parsedDoc(), seqNo, index.primaryTerm(),
```
RS146BIJAY: By using Engine.Index inside DataFormatAwareEngine, are we not taking a dependecy on Lucene here? So what I meant is, inside Engine.index constructor, we are passing a Term as the first parameter which is obtained from index.uid(). Term is a Lucene concept: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/Term.java. Should DataAwareEngine should have a direct dependency on a Lucene construct here or can we keep this independent? Asking because if we keep a dependency on Lucene term for id during ingestion, update API for say ParquetWriter or composite writer will accept a Term instance for id like: update(Term uid, CompositeDocumentInput doc), which seems incorrect.

mgodwan: Agreed. 
I didn't introduce another indirection here to avoid additional object creation for existing engine. 

> update API for say ParquetWriter or composite writer will accept a Term instance for id like
We can just extract the byte[] and pass it? Why do we need to have the dependency flow in there?

server/src/main/java/org/opensearch/index/engine/EngineConfig.java
```diff
@@ -117,6 +120,9 @@ public final class EngineConfig {
     private final Supplier<DocumentMapperForType> documentMapperForTypeSupplier;
     private final ClusterApplierService clusterApplierService;
     private final MergedSegmentTransferTracker mergedSegmentTransferTracker;
+    private final DataFormatRegistry dataFormatRegistry;
+    private final MapperService mapperService;
+    private CommitterFactory committerFactory;
```
Bukhtawar: Wouldn't this amount to a breaking change given its's a `PublicApi`

mgodwan: Since the constructor accepts a builder, this should be ok. Didn't see any issues with japicmp target either.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -8,80 +8,954 @@
...
+                return indexResult;
+            }
+        } catch (RuntimeException | IOException e) {
+            maybeFailEngine("index id[" + index.id() + "] origin[" + index.origin() + "]", e);
+            throw e;
+        }
+    }
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    private Engine.IndexResult indexIntoEngine(Engine.Index index) throws IOException {
```
Bukhtawar, line 407: nit: Maybe we could move this and other IndexingExecution methods to another class for better maintainability and testability.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -8,80 +8,954 @@
...
+        VersionType versionType,
+        Engine.Operation.Origin origin,
+        long ifSeqNo,
+        long ifPrimaryTerm
+    ) {
+        throw new UnsupportedOperationException("delete operation not supported.");
+    }
+
+    @Override
+    public void refresh(String source) throws EngineException {
```
Bukhtawar, line 521: Do we need to care about internal/external readers?

server/src/test/java/org/opensearch/index/engine/DataFormatAwareEngineTests.java
```diff
@@ -0,0 +1,1560 @@
...
+
+/**
+ * Tests for {@link DataFormatAwareEngine} covering the orchestration layer:
+ * sequence number management, translog integration, refresh/flush lifecycle,
+ * catalog snapshot management, concurrency, and failure handling.
+ *
+ * <p>Uses mock data format components to isolate the DFAE orchestration logic
+ * from any specific data format implementation.</p>
+ */
+public class DataFormatAwareEngineTests extends OpenSearchTestCase {
```
Bukhtawar, line 87: How can we ensure that existing indexing tests with Lucene as the primary format(only) runs fine without regression in functionality?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
mgodwan: Yes. tracking this in meta here https://github.com/opensearch-project/OpenSearch/issues/20876

Bukhtawar (approved):
Thanks for the changes, looks pretty good.

server/src/test/java/org/opensearch/index/engine/DataFormatAwareEngineTests.java
mgodwan: Once the required features are available (merges, updates, etc), the plan is to enable tests written in IndexShardIT, InternalEngineMergeIT, etc. to run against the composite plugin. I'm taking that as an acceptance criteria for the overall indexing.

server/src/main/java/org/opensearch/index/engine/EngineConfig.java
Bukhtawar: nit: final?

mgodwan:
Flaky test:

 - org.opensearch.indices.IndicesRequestCacheCleanupIT.testCacheClearAPIRemovesStaleKeysWhenStalenessThresholdIsLow

mgodwan:
Gradle check is returning errors:

```
Also:   org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: 87d21a12-9ba5-4bb6-85fb-51045965c112
com.onepassword.jenkins.plugins.exception.OnePasswordException: Error retrieving secret op://opensearch-infra-secrets/dockerhub-production-readonly-credentials/username:
[ERROR] 2026/04/14 17:43:30 could not read secret 'op://opensearch-infra-secrets/dockerhub-production-readonly-credentials/username': error initializing client: Unknown: (504) (Unknown), An unknown error occurred.
```

PR #21183, opened 2026-04-09
[Pull-based Ingestion] Fix document mapper usage in ingestion engine

varunbharadwaj:
### Description
The ingestion engine currently retrieves an instance of document mapper and always uses it until next engine initialization. This is a bug as newly added fields and other mapping updates do not reflect till the shard is reinitialized. This PR fixes this by always retrieving document mapper instance for every message.

### DIFF
--- a/plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
+++ b/plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
@@ -1385,6 +1385,110 @@ public class IngestFromKafkaIT extends KafkaIngestionBaseIT {
         waitForSearchableDocs(10, List.of(nodeA));
     }
 
+    public void testKafkaIngestionWithMappingUpdate() throws Exception {
+        final String nodeA = internalCluster().startNode();
+
+        createIndex(
+            indexName,
+            Settings.builder()
+                .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
+                .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
+                .put("ingestion_source.type", "kafka")
+                .put("ingestion_source.pointer.init.reset", "earliest")
+                .put("ingestion_source.param.topic", topicName)
+                .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
+                .put("index.replication.type", "SEGMENT")
+                .put("ingestion_source.error_strategy", "drop")
+                .build(),
+            "{\"dynamic\":\"strict\",\"properties\":{\"product\":{\"type\": \"text\"},"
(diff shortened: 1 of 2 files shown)


plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
```diff
@@ -1385,6 +1385,110 @@ public void testWarmupPhase() throws Exception {
...
+                .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
+                .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
+                .put("ingestion_source.type", "kafka")
+                .put("ingestion_source.pointer.init.reset", "earliest")
+                .put("ingestion_source.param.topic", topicName)
+                .put("ingestion_source.param.bootstrap_servers", kafka.getBootstrapServers())
+                .put("index.replication.type", "SEGMENT")
+                .put("ingestion_source.error_strategy", "drop")
+                .build(),
+            "{\"dynamic\":\"strict\",\"properties\":{\"product\":{\"type\": \"text\"},"
```
pseudo-nymous, line 1403: From my understanding, in pull-based ingestion, setting `dynamic = true` allows new fields to be indexed based on inferred types, but the mapping itself is not updated (unlike in push-based ingestion). Since Lucene maintains its own field schema, this can lead to inconsistencies if the same field appears with different types later, and such issues would only surface at indexing time.

Given this, should we consider explicitly disabling dynamic = true for pull-based ingestion and informing users during index creation?

Also, how does `dynamic = strict_allow_templates` behave in pull-based ingestion? If it follows the same pattern (i.e., no mapping updates), we should notify users similarly.

pseudo-nymous (approved):

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
varunbharadwaj: Yeah, this is a good point. Currently we just say we don't support dynamic mapping but don't enforce it in indexing path. It would be good to maybe add some validation on PBI index creation to avoid such issues. We can take it as a follow up 👍

kaustubhbutte17 (approved):

msfroh (approved):

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/IngestFromKafkaIT.java
varunbharadwaj: As @msfroh mentioned, maybe it is worth looking into adding dynamic mapping support instead of disabling it entirely. We can think about it more.

PR #21191, opened 2026-04-09
native memory based admission control

pradeep-L:
### Description

This change introduces a Native Memory Based Admission Controller that throttles incoming transport requests (search, indexing, cluster admin) based on actual physical memory utilization on the node. The controller reads native memory usage from OsProbe. which parses MemAvailable from /proc/meminfo on Linux systems. The NativeMemoryBasedAdmissionController evaluates this usage against configurable per-action-type thresholds and rejects the requests with HTTP 429 .  This gives the operators a more accurate signal for memory pressure since it accounts for OS-level page cache and slab memory rather than relying solely on JVM heap usage.

### Related Issues

### Check List

### DIFF
--- a/distribution/packages/src/common/systemd/opensearch.service
+++ b/distribution/packages/src/common/systemd/opensearch.service
@@ -145,6 +145,9 @@ ReadOnlyPaths=-/etc/os-release -/usr/lib/os-release -/etc/system-release
 ## Allow read access to Linux IO stats
 ReadOnlyPaths=/proc/self/mountinfo /proc/diskstats
 
+## Allow read access to Linux memory stats
+ReadOnlyPaths=/proc/meminfo
+
 ## Allow read access to control group stats
 ReadOnlyPaths=/proc/self/cgroup /sys/fs/cgroup/cpu /sys/fs/cgroup/cpu/-
 ReadOnlyPaths=/sys/fs/cgroup/cpuacct /sys/fs/cgroup/cpuacct/- /sys/fs/cgroup/memory /sys/fs/cgroup/memory/-
--- a/distribution/packages/src/common/systemd/opensearch.service
+++ b/distribution/packages/src/common/systemd/opensearch.service
@@ -145,8 +145,8 @@ ReadOnlyPaths=-/etc/os-release -/usr/lib/os-release -/etc/system-release
 ## Allow read access to Linux IO stats
 ReadOnlyPaths=/proc/self/mountinfo /proc/diskstats
 
-## Allow read access to Linux memory stats
-ReadOnlyPaths=/proc/meminfo
+## Allow read access to native memory stats (RssAnon via /proc/self/status)
+ReadOnlyPaths=/proc/self/status
 
 ## Allow read access to control group stats
 ReadOnlyPaths=/proc/self/cgroup /sys/fs/cgroup/cpu /sys/fs/cgroup/cpu/-
--- a/server/src/internalClusterTest/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionForClusterManagerIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionForClusterManagerIT.java
@@ -25,6 +25,7 @@ import org.opensearch.node.ResourceUsageCollectorService;
 import org.opensearch.node.resource.tracker.ResourceTrackerSettings;
 import org.opensearch.plugins.Plugin;
 import org.opensearch.ratelimitting.admissioncontrol.controllers.CpuBasedAdmissionController;
+import org.opensearch.ratelimitting.admissioncontrol.controllers.NativeMemoryBasedAdmissionController;
 import org.opensearch.ratelimitting.admissioncontrol.enums.AdmissionControlActionType;
 import org.opensearch.ratelimitting.admissioncontrol.enums.AdmissionControlMode;
 import org.opensearch.ratelimitting.admissioncontrol.stats.AdmissionControllerStats;
@@ -47,6 +48,8 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import static org.opensearch.ratelimitting.admissioncontrol.AdmissionControlSettings.ADMISSION_CONTROL_TRANSPORT_LAYER_MODE;
 import static org.opensearch.ratelimitting.admissioncontrol.settings.CpuBasedAdmissionControllerSettings.CLUSTER_ADMIN_CPU_USAGE_LIMIT;
+import static org.opensearch.ratelimitting.admissioncontrol.settings.NativeMemoryBasedAdmissionControllerSettings.CLUSTER_ADMIN_NATIVE_MEMORY_USAGE_LIMIT;
+import static org.opensearch.ratelimitting.admissioncontrol.settings.NativeMemoryBasedAdmissionControllerSettings.NATIVE_MEMORY_BASED_ADMISSION_CONTROLLER_TRANSPORT_LAYER_MODE;
 import static org.opensearch.test.hamcrest.OpenSearchAssertions.assertAcked;
 import static org.hamcrest.Matchers.equalTo;
 
@@ -114,7 +117,7 @@ public class AdmissionForClusterManagerIT extends OpenSearchIntegTestCase {
     }
 
     public void testAdmissionControlEnforced() throws Exception {
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 99, new IoUsageStats(98));
+        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 99, new IoUsageStats(98), 90);
 
         // Write API on ClusterManager
         assertAcked(prepareCreate("test").setMapping("field", "type=text").setAliases("{\"alias1\" : {}}"));
@@ -149,7 +152,7 @@ public class AdmissionForClusterManagerIT extends OpenSearchIntegTestCase {
 
     public void testAdmissionControlEnabledOnNoBreach() throws InterruptedException {
         // CPU usage is less than threshold 50%
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 35, new IoUsageStats(98));
+        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 35, new IoUsageStats(98), 90);
 
         // Write API on ClusterManager
         assertAcked(prepareCreate("test").setMapping("field", "type=text").setAliases("{\"alias1\" : {}}").execute().actionGet());
@@ -174,7 +177,7 @@ public class AdmissionForClusterManagerIT extends OpenSearchIntegTestCase {
     public void admissionControlDisabledOnBreach(Settings admission) throws InterruptedException {
         client().admin().cluster().prepareUpdateSettings().setTransientSettings(admission).execute().actionGet();
 
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 97, new IoUsageStats(98));
+        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 97, new IoUsageStats(98), 90);
 
         // Write API on ClusterManager
         assertAcked(prepareCreate("test").setMapping("field", "type=text").setAliases("{\"alias1\" : {}}").execute().actionGet());
@@ -188,7 +191,7 @@ public class AdmissionForClusterManagerIT extends OpenSearchIntegTestCase {
     }
 
     public void testAdmissionControlResponseStatus() throws Exception {
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 99, new IoUsageStats(98));
+        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 99, new IoUsageStats(98), 90);
 
         // Write API on ClusterManager
         assertAcked(prepareCreate("test").setMapping("field", "type=text").setAliases("{\"alias1\" : {}}"));
--- a/server/src/internalClusterTest/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionForClusterManagerIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionForClusterManagerIT.java
@@ -25,7 +25,6 @@ import org.opensearch.node.ResourceUsageCollectorService;
 import org.opensearch.node.resource.tracker.ResourceTrackerSettings;
 import org.opensearch.plugins.Plugin;
 import org.opensearch.ratelimitting.admissioncontrol.controllers.CpuBasedAdmissionController;
-import org.opensearch.ratelimitting.admissioncontrol.controllers.NativeMemoryBasedAdmissionController;
 import org.opensearch.ratelimitting.admissioncontrol.enums.AdmissionControlActionType;
 import org.opensearch.ratelimitting.admissioncontrol.enums.AdmissionControlMode;
 import org.opensearch.ratelimitting.admissioncontrol.stats.AdmissionControllerStats;
@@ -48,8 +47,6 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import static org.opensearch.ratelimitting.admissioncontrol.AdmissionControlSettings.ADMISSION_CONTROL_TRANSPORT_LAYER_MODE;
 import static org.opensearch.ratelimitting.admissioncontrol.settings.CpuBasedAdmissionControllerSettings.CLUSTER_ADMIN_CPU_USAGE_LIMIT;
-import static org.opensearch.ratelimitting.admissioncontrol.settings.NativeMemoryBasedAdmissionControllerSettings.CLUSTER_ADMIN_NATIVE_MEMORY_USAGE_LIMIT;
-import static org.opensearch.ratelimitting.admissioncontrol.settings.NativeMemoryBasedAdmissionControllerSettings.NATIVE_MEMORY_BASED_ADMISSION_CONTROLLER_TRANSPORT_LAYER_MODE;
 import static org.opensearch.test.hamcrest.OpenSearchAssertions.assertAcked;
 import static org.hamcrest.Matchers.equalTo;
 
@@ -117,7 +114,14 @@ public class AdmissionForClusterManagerIT extends OpenSearchIntegTestCase {
     }
 
     public void testAdmissionControlEnforced() throws Exception {
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 99, new IoUsageStats(98), 90);
+        cMResourceCollector.collectNodeResourceUsageStats(
+            clusterManagerNodeId,
+            System.currentTimeMillis(),
+            97,
+            99,
+            new IoUsageStats(98),
+            90
+        );
 
         // Write API on ClusterManager
         assertAcked(prepareCreate("test").setMapping("field", "type=text").setAliases("{\"alias1\" : {}}"));
@@ -152,7 +156,14 @@ public class AdmissionForClusterManagerIT extends OpenSearchIntegTestCase {
 
     public void testAdmissionControlEnabledOnNoBreach() throws InterruptedException {
         // CPU usage is less than threshold 50%
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 35, new IoUsageStats(98), 90);
+        cMResourceCollector.collectNodeResourceUsageStats(
+            clusterManagerNodeId,
+            System.currentTimeMillis(),
+            97,
+            35,
+            new IoUsageStats(98),
+            90
+        );
 
         // Write API on ClusterManager
         assertAcked(prepareCreate("test").setMapping("field", "type=text").setAliases("{\"alias1\" : {}}").execute().actionGet());
@@ -177,7 +188,14 @@ public class AdmissionForClusterManagerIT extends OpenSearchIntegTestCase {
     public void admissionControlDisabledOnBreach(Settings admission) throws InterruptedException {
         client().admin().cluster().prepareUpdateSettings().setTransientSettings(admission).execute().actionGet();
 
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 97, new IoUsageStats(98), 90);
+        cMResourceCollector.collectNodeResourceUsageStats(
+            clusterManagerNodeId,
+            System.currentTimeMillis(),
+            97,
+            97,
+            new IoUsageStats(98),
+            90
+        );
 
         // Write API on ClusterManager
         assertAcked(prepareCreate("test").setMapping("field", "type=text").setAliases("{\"alias1\" : {}}").execute().actionGet());
@@ -191,7 +209,14 @@ public class AdmissionForClusterManagerIT extends OpenSearchIntegTestCase {
     }
 
     public void testAdmissionControlResponseStatus() throws Exception {
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 99, new IoUsageStats(98), 90);
+        cMResourceCollector.collectNodeResourceUsageStats(
+            clusterManagerNodeId,
+            System.currentTimeMillis(),
+            97,
+            99,
+            new IoUsageStats(98),
+            90
+        );
 
         // Write API on ClusterManager
         assertAcked(prepareCreate("test").setMapping("field", "type=text").setAliases("{\"alias1\" : {}}"));
--- a/server/src/internalClusterTest/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionForClusterManagerIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionForClusterManagerIT.java
@@ -10,6 +10,7 @@ package org.opensearch.ratelimitting.admissioncontrol;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.lucene.util.Constants;
 import org.opensearch.action.admin.indices.alias.get.GetAliasesRequest;
 import org.opensearch.action.admin.indices.alias.get.GetAliasesResponse;
 import org.opensearch.action.support.clustermanager.term.GetTermVersionAction;
@@ -25,6 +26,7 @@ import org.opensearch.node.ResourceUsageCollectorService;
 import org.opensearch.node.resource.tracker.ResourceTrackerSettings;
 import org.opensearch.plugins.Plugin;
 import org.opensearch.ratelimitting.admissioncontrol.controllers.CpuBasedAdmissionController;
+import org.opensearch.ratelimitting.admissioncontrol.controllers.NativeMemoryBasedAdmissionController;
 import org.opensearch.ratelimitting.admissioncontrol.enums.AdmissionControlActionType;
 import org.opensearch.ratelimitting.admissioncontrol.enums.AdmissionControlMode;
 import org.opensearch.ratelimitting.admissioncontrol.stats.AdmissionControllerStats;
@@ -47,6 +49,8 @@ import java.util.concurrent.atomic.AtomicReference;
 
 import static org.opensearch.ratelimitting.admissioncontrol.AdmissionControlSettings.ADMISSION_CONTROL_TRANSPORT_LAYER_MODE;
 import static org.opensearch.ratelimitting.admissioncontrol.settings.CpuBasedAdmissionControllerSettings.CLUSTER_ADMIN_CPU_USAGE_LIMIT;
+import static org.opensearch.ratelimitting.admissioncontrol.settings.NativeMemoryBasedAdmissionControllerSettings.CLUSTER_ADMIN_NATIVE_MEMORY_USAGE_LIMIT;
+import static org.opensearch.ratelimitting.admissioncontrol.settings.NativeMemoryBasedAdmissionControllerSettings.NATIVE_MEMORY_BASED_ADMISSION_CONTROLLER_TRANSPORT_LAYER_MODE;
 import static org.opensearch.test.hamcrest.OpenSearchAssertions.assertAcked;
 import static org.hamcrest.Matchers.equalTo;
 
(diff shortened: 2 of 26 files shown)


server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -253,6 +253,44 @@ private String readSingleLine(final Path path) throws IOException {
         return lines.get(0);
     }
 
+    /**
+     * Returns the available memory in bytes on Linux by reading {@code MemAvailable} from {@code /proc/meminfo}.
+     * Available memory is a better estimate than free memory as it accounts for reclaimable page cache and slab memory.
+     * Returns -1 if not on Linux or if the value cannot be read.
+     */
+    public long getAvailableMemorySize() {
+        if (!Constants.LINUX) {
```
himshikhagupta: Why are we limiting to Linux only systems?

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -253,6 +253,44 @@ private String readSingleLine(final Path path) throws IOException {
...
+     *
+     * @return the available memory in bytes, or -1 if not found
+     * @throws IOException if an I/O exception occurs reading {@code /proc/meminfo}
+     */
+    @SuppressForbidden(reason = "access /proc/meminfo")
+    long readAvailableMemoryFromProcMeminfo() throws IOException {
+        final List<String> lines = Files.readAllLines(PathUtils.get("/proc/meminfo"));
+        for (final String line : lines) {
+            if (line.startsWith("MemAvailable:")) {
+                final String[] parts = line.split("\\s+");
```
himshikhagupta, line 283: How is this different from what the existing method `getFreePhysicalMemorySize` is already giving?

server/src/main/java/org/opensearch/node/resource/tracker/NodeResourceUsageTracker.java
```diff
@@ -66,14 +67,25 @@ public IoUsageStats getIoUsageStats() {
...
      * Checks if all of the resource usage trackers are ready
      */
     public boolean isReady() {
         if (Constants.LINUX) {
-            return memoryUsageTracker.isReady() && cpuUsageTracker.isReady() && ioUsageTracker.isReady();
+            return memoryUsageTracker.isReady() && cpuUsageTracker.isReady() && ioUsageTracker.isReady()
+                && nativeMemoryUsageTracker.isReady();
         }
-        return memoryUsageTracker.isReady() && cpuUsageTracker.isReady();
+        return memoryUsageTracker.isReady() && cpuUsageTracker.isReady() && nativeMemoryUsageTracker.isReady();
```
himshikhagupta: Should we initialize the nativememorytracker if the metric itself is not available?

server/src/main/java/org/opensearch/node/NodeResourceUsageStats.java
```diff
@@ -71,6 +76,7 @@ public void writeTo(StreamOutput out) throws IOException {
         if (out.getVersion().onOrAfter(Version.V_2_13_0)) {
             out.writeOptionalWriteable(this.ioUsageStats);
         }
+        out.writeDouble(this.nativeMemoryUtilizationPercent);
```
himshikhagupta: Do we need a version check here?

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+
+    /**
+     * Check if the configured resource usage limits are breached for the action
+     */
+    private boolean isLimitsBreached(String actionName, AdmissionControlActionType admissionControlActionType) {
+        // check if cluster state is ready
+        if (clusterService.state() != null && clusterService.state().nodes() != null) {
+            long maxMemoryLimit = this.getMemoryRejectionThreshold(admissionControlActionType);
```
himshikhagupta, line 91: How are we mapping the native memory usage to action type?

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
pradeep-L: Free Physical memory is RAM that is completely unused,
Available memory is RAM that can be used without causing pressure (free + reclaimable).
we are considering available memory for our use case.

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java
pradeep-L: we can configure different values for indexing , search and cluster admin action types.

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
pradeep-L: For available memory space we have direct signal in linux /proc/meminfo.
But for other operating systems like windows or mac we dont have direct equivalent to fetch the available memory . hence restricting this to only linux systems.

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java
himshikhagupta: configuration makes sense when we know how much is each flow occupying right? Search and Indexing we may be able to get with the plugin memory tracking, but how will we get cluster admin action memory?

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -253,6 +253,44 @@ private String readSingleLine(final Path path) throws IOException {
...
+     * Returns -1 if not on Linux or if the value cannot be read.
+     */
+    public long getAvailableMemorySize() {
+        if (!Constants.LINUX) {
+            return -1;
+        }
+        try {
+            return readAvailableMemoryFromProcMeminfo();
+        } catch (Exception e) {
+            logger.debug("error reading available memory from /proc/meminfo", e);
```
himshikhagupta: nit: this should be logger.warn?

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -253,6 +253,44 @@ private String readSingleLine(final Path path) throws IOException {
...
+
+    /**
+     * Reads {@code MemAvailable} from {@code /proc/meminfo}.
+     *
+     * @return the available memory in bytes, or -1 if not found
+     * @throws IOException if an I/O exception occurs reading {@code /proc/meminfo}
+     */
+    @SuppressForbidden(reason = "access /proc/meminfo")
+    long readAvailableMemoryFromProcMeminfo() throws IOException {
+        final List<String> lines = Files.readAllLines(PathUtils.get("/proc/meminfo"));
```
himshikhagupta: readAllLines looks inefficient since we only want one value. Better to use Files.lines() which will read one line at a time and you can exit when you find the required value.

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -253,6 +253,44 @@ private String readSingleLine(final Path path) throws IOException {
...
+        for (final String line : lines) {
+            if (line.startsWith("MemAvailable:")) {
+                final String[] parts = line.split("\\s+");
+                if (parts.length >= 2) {
```
himshikhagupta, line 284: If this is not true, do we keep looping or error out?

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -0,0 +1,61 @@
...
+     * Get current native memory usage percentage using OS-level physical memory stats.
+     * Prefers available memory (MemAvailable) over free memory (MemFree) as it provides
+     * a more accurate measure of actual memory pressure on the node.
+     */
+    @Override
+    public long getUsage() {
+        OsProbe osProbe = OsProbe.getInstance();
+        long totalMemory = osProbe.getTotalPhysicalMemorySize();
+        if (totalMemory <= 0) {
+            LOGGER.debug("Unable to retrieve total physical memory size");
```
himshikhagupta: Should this be warn/error

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -0,0 +1,61 @@
...
+            return 0;
+        }
+        long availableMemory = osProbe.getAvailableMemorySize();
+        long unusedMemory;
+        if (availableMemory >= 0) {
+            // Use available memory (includes reclaimable cache) for a more accurate picture
+            unusedMemory = availableMemory;
+        } else {
+            LOGGER.debug("for non linux systems we will not measure this native memory");
```
himshikhagupta: Should we do a Linux check first and skip the execution based on that?

server/src/main/java/org/opensearch/node/NodeResourceUsageStats.java
```diff
@@ -60,6 +64,7 @@ public NodeResourceUsageStats(StreamInput in) throws IOException {
         } else {
             this.ioUsageStats = null;
         }
+        this.nativeMemoryUtilizationPercent = in.readDouble();
```
himshikhagupta: For this also version check is needed?

server/src/main/java/org/opensearch/node/NodeResourceUsageStats.java
```diff
@@ -71,6 +76,9 @@ public void writeTo(StreamOutput out) throws IOException {
         if (out.getVersion().onOrAfter(Version.V_2_13_0)) {
             out.writeOptionalWriteable(this.ioUsageStats);
         }
+        if (out.getVersion().onOrAfter(Version.V_3_4_0)) {
```
himshikhagupta: Should this be 3.5?

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
```diff
@@ -0,0 +1,118 @@
...
+    );
+
+    public NativeMemoryBasedAdmissionControllerSettings(ClusterSettings clusterSettings, Settings settings) {
+        this.transportLayerMode = NATIVE_MEMORY_BASED_ADMISSION_CONTROLLER_TRANSPORT_LAYER_MODE.get(settings);
+        clusterSettings.addSettingsUpdateConsumer(NATIVE_MEMORY_BASED_ADMISSION_CONTROLLER_TRANSPORT_LAYER_MODE, this::setTransportLayerMode);
+        this.searchNativeMemoryUsageLimit = SEARCH_NATIVE_MEMORY_USAGE_LIMIT.get(settings);
+        this.indexingNativeMemoryUsageLimit = INDEXING_NATIVE_MEMORY_USAGE_LIMIT.get(settings);
+        this.clusterAdminNativeMemoryUsageLimit = CLUSTER_ADMIN_NATIVE_MEMORY_USAGE_LIMIT.get(settings);
+        clusterSettings.addSettingsUpdateConsumer(SEARCH_NATIVE_MEMORY_USAGE_LIMIT, this::setSearchNativeMemoryUsageLimit);
+        clusterSettings.addSettingsUpdateConsumer(INDEXING_NATIVE_MEMORY_USAGE_LIMIT, this::setIndexingNativeMemoryUsageLimit);
```
himshikhagupta, line 91: Consumer for CLUSTER_ADMIN_NATIVE_MEMORY_USAGE_LIMIT missing

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionControlService.java
```diff
@@ -73,6 +75,7 @@ private void initialize() {
         if (Constants.LINUX) {
             registerAdmissionController(IO_BASED_ADMISSION_CONTROLLER);
         }
+        registerAdmissionController(NATIVE_MEMORY_BASED_ADMISSION_CONTROLLER);
```
himshikhagupta: Should native memory controller be registered only if its linux similar to IO?

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: Have changed the wordings, the native memory tracker will not be initialized if the system is not linux . we have the check here 

```if (Constants.LINUX) {
            registerAdmissionController(IO_BASED_ADMISSION_CONTROLLER);
            registerAdmissionController(NATIVE_MEMORY_BASED_ADMISSION_CONTROLLER);
        }
     ```

File: server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java (hunk 5 above)
Comment (pradeep-L) [reply server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java:91]: For cluster admin , its more like a guardrail if CPU or memory is at 95% admin APIs should be rejected.

File: server/src/main/java/org/opensearch/node/resource/tracker/NodeResourceUsageTracker.java [hunk 15]
```diff
@@ -135,20 +166,29 @@ ResourceTrackerSettings getResourceTrackerSettings() {
     protected void doStart() {
         cpuUsageTracker.doStart();
         memoryUsageTracker.doStart();
-        ioUsageTracker.doStart();
+        if (Constants.LINUX) {
+            ioUsageTracker.doStart();
+            nativeMemoryUsageTracker.doStart();
+        }
```
himshikhagupta, line 187: Lets not change existing setup for ioUsageTracker.

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Class for Native Memory Based Admission Controller in OpenSearch, which aims to provide
+ * admission control based on the actual physical memory usage on the node.
+ * It provides methods to apply admission control if configured limit has been reached.
+ */
+public class NativeMemoryBasedAdmissionController extends AdmissionController {
+    public static final String NATIVE_MEMORY_BASED_ADMISSION_CONTROLLER = "global_native_memory_usage";
+    private static final Logger LOGGER = LogManager.getLogger(NativeMemoryBasedAdmissionController.class);
+    public NativeMemoryBasedAdmissionControllerSettings settings;
```
himshikhagupta: Does this need to be public?

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
```diff
@@ -0,0 +1,126 @@
...
+    public AdmissionControlMode getTransportLayerAdmissionControllerMode() {
+        return transportLayerMode;
+    }
+
+    public Long getSearchNativeMemoryUsageLimit() {
+        return searchNativeMemoryUsageLimit;
+    }
+
+    public void setSearchNativeMemoryUsageLimit(Long searchNativeMemoryUsageLimit) {
+        this.searchNativeMemoryUsageLimit = searchNativeMemoryUsageLimit;
```
himshikhagupta, line 108: Do we need validations on min and max values here? Can someone set it to be higher than 100%? How does system behave then?

server/src/test/java/org/opensearch/monitor/os/OsProbeTests.java
```diff
@@ -310,6 +310,95 @@ List<String> readSysFsCgroupCpuAcctCpuStat(String controlGroup) throws IOExcepti
...
+                for (final String line : lines) {
+                    if (line.startsWith("MemAvailable:")) {
+                        final String[] parts = line.split("\\s+");
+                        if (parts.length >= 2) {
+                            return Long.parseLong(parts[1]) * 1024;
+                        }
+                    }
+                }
+                return -1;
+            }
```
himshikhagupta: Why do we need this impl in test? It should call the real method otherwise we will skip if that method is executing correctly

server/src/main/java/org/opensearch/node/NodeResourceUsageStats.java
```diff
@@ -71,6 +80,9 @@ public void writeTo(StreamOutput out) throws IOException {
         if (out.getVersion().onOrAfter(Version.V_2_13_0)) {
             out.writeOptionalWriteable(this.ioUsageStats);
         }
+        if (out.getVersion().onOrAfter(Version.V_3_7_0)) {
```
himshikhagupta, line 83: can we add a test for BWC serialization

server/src/internalClusterTest/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionForClusterManagerIT.java
```diff
@@ -114,7 +114,14 @@ public void init() {
     }
 
     public void testAdmissionControlEnforced() throws Exception {
-        cMResourceCollector.collectNodeResourceUsageStats(clusterManagerNodeId, System.currentTimeMillis(), 97, 99, new IoUsageStats(98));
+        cMResourceCollector.collectNodeResourceUsageStats(
```
himshikhagupta, line 122: can we add a test for native memory checks, ensuring all settings are working, similar to what we have for other resources here.

server/src/main/java/org/opensearch/node/resource/tracker/NodeResourceUsageTracker.java
pradeep-L: done

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
pradeep-L: Its up to the customer if he sets this value as 100 then the search request will not be rejected and system will oom out. this is the pattern in other controllers as well.

server/src/internalClusterTest/java/org/opensearch/ratelimitting/admissioncontrol/AdmissionForClusterManagerIT.java
pradeep-L: added tests for native memory settings

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
himshikhagupta: I understand other controllers may not be validating, but since we are building a new controller, we can add checks that were missing before. Allowed values should be between 0-100.

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
bharath-techie: Better to add this as a comment as well in the code for reference ?

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
```diff
@@ -0,0 +1,126 @@
...
+ * Settings related to native memory based admission controller.
+ * @opensearch.internal
+ */
+public class NativeMemoryBasedAdmissionControllerSettings {
+
+    /**
+     * Default parameters for the NativeMemoryBasedAdmissionControllerSettings
+     */
+    public static class Defaults {
+        public static final long NATIVE_MEMORY_USAGE_LIMIT = 80;
```
bharath-techie: How will this be configured ? Should we take the account of JVM as well, as in , the native memory usage limit should be based on percentage of `total memory - jvm memory` ?

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
```diff
@@ -0,0 +1,126 @@
...
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.ratelimitting.admissioncontrol.AdmissionControlSettings;
+import org.opensearch.ratelimitting.admissioncontrol.enums.AdmissionControlMode;
+
+/**
+ * Settings related to native memory based admission controller.
+ * @opensearch.internal
+ */
+public class NativeMemoryBasedAdmissionControllerSettings {
```
bharath-techie, line 21: In general today we use circuit breakers for memory based rejections , can you add why we are using transport AC and if we'll extend circuit breakers as well ?

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java
bharath-techie: `sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java `- Just a heads up, I think we already have mapped to `Search` AC type. I think bulk action as well, we can check similarly, if nothing has changed it'll get mapped to `Indexing`.

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
pradeep-L: available memory space is ( total memory - jvm memory - all memory allocated to other process)

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
pradeep-L: yes, we will add native memory based circuit breaker. Circuit Breaker is reactive , cancels requests during execution phase. Admission control is proactive, before the start of the execution.

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
pradeep-L: added

distribution/packages/src/common/systemd/opensearch.service
```diff
@@ -145,6 +145,9 @@ ReadOnlyPaths=-/etc/os-release -/usr/lib/os-release -/etc/system-release
 ## Allow read access to Linux IO stats
 ReadOnlyPaths=/proc/self/mountinfo /proc/diskstats
 
+## Allow read access to Linux memory stats
+ReadOnlyPaths=/proc/meminfo
```
himshikhagupta: Do you need to clean this up if this is not being used anymore?

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -236,6 +237,64 @@ String readProcLoadavg() throws IOException {
...
+     * @return the {@code RssAnon} value in bytes, or {@code -1L} if unavailable
+     */
+    public long getProcessRssAnon() {
+        if (Constants.LINUX == false) {
+            return -1L;
+        }
+        try {
+            return readRssAnonFromProcSelfStatus();
+        } catch (IOException e) {
+            logger.debug("failed to read /proc/self/status", e);
```
himshikhagupta: exceptions should be warn

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -236,6 +237,64 @@ String readProcLoadavg() throws IOException {
...
+                            if (kb < 0L) {
+                                return -1L;
+                            }
+                            return kb * 1024L;
+                        } catch (NumberFormatException nfe) {
+                            logger.debug("malformed RssAnon value in /proc/self/status", nfe);
+                            return -1L;
+                        }
+                    }
+                    logger.debug("RssAnon line has unexpected shape: [{}]", line);
```
himshikhagupta: Can we ensure all errors are marked as either error or warn? 
For debug logs, maybe add `isDebugEnabled` check

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -0,0 +1,183 @@
...
+        long rssAnon = rssAnonSupplier.getAsLong();
+        if (rssAnon < 0L) {
+            LOGGER.debug("Native memory poll skipped: RssAnon unavailable from /proc/self/status");
+            return 0L;
+        }
+
+        long heapCommitted = heapCommittedSupplier.getAsLong();
+        long usage = Math.max(0L, rssAnon - heapCommitted);
+
+        long cap = nativeMemoryCapSupplier.getAsLong();
```
himshikhagupta: Why do we need to calculate cap in every invocation?

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -0,0 +1,183 @@
...
+        return percent;
+    }
+
+
+    /**
+     * Computes {@code Non_Heap_Base = max(0, totalPhysicalMemory - Runtime.maxMemory())}, the
+     * reference value against which {@code parquet.max_native_allocation} percentages resolve.
+     */
+    long computeNonHeapBase() {
+        return Math.max(0L, OsProbe.getInstance().getTotalPhysicalMemorySize() - Runtime.getRuntime().maxMemory());
```
himshikhagupta: Why are we using total physical memory here?

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -0,0 +1,183 @@
...
+
+    /**
+     * Resolves both plugin settings to byte values and returns the 20% headroom-padded sum.
+     * Returns {@code 0L} when neither plugin setting contributes a positive value, so
+     * {@link #getUsage()} can short-circuit without dividing.
+     */
+    long computeNativeMemoryCap(Settings settings) {
+        long df = resolveDataFusionContribution(settings);
+        long base = computeNonHeapBase();
+        long pq = resolveParquetContribution(settings, base);
```
himshikhagupta: Will this still hold once other rust based plugins also start occupying more memory?

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -0,0 +1,183 @@
...
+ * platforms the polling loop is not started and {@link OsProbe#getProcessRssAnon()} returns
+ * {@code -1} without touching the filesystem.
+ */
+public class AverageNativeMemoryUsageTracker extends AbstractAverageUsageTracker {
+
+    private static final Logger LOGGER = LogManager.getLogger(AverageNativeMemoryUsageTracker.class);
+
+    static final String DATAFUSION_MEMORY_POOL_LIMIT_KEY = "datafusion.memory_pool_limit_bytes";
+    static final String PARQUET_MAX_NATIVE_ALLOCATION_KEY = "parquet.max_native_allocation";
+    static final double HEADROOM_FACTOR = 1.2;
```
himshikhagupta: Did we do some benchmarking to come up with? Otherwise this should be configurable.

server/src/main/resources/org/opensearch/bootstrap/security.policy
```diff
@@ -235,6 +235,9 @@ grant {
   permission java.io.FilePermission "/proc/self/mountinfo", "read";
   permission java.io.FilePermission "/proc/diskstats", "read";
 
+  // memory stats on Linux
+  permission java.io.FilePermission "/proc/meminfo", "read";
```
himshikhagupta: check if this is required now?

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -0,0 +1,183 @@
...
+ * <p>Activation is already gated to Linux in {@link NodeResourceUsageTracker}; on non-Linux
+ * platforms the polling loop is not started and {@link OsProbe#getProcessRssAnon()} returns
+ * {@code -1} without touching the filesystem.
+ */
+public class AverageNativeMemoryUsageTracker extends AbstractAverageUsageTracker {
+
+    private static final Logger LOGGER = LogManager.getLogger(AverageNativeMemoryUsageTracker.class);
+
+    static final String DATAFUSION_MEMORY_POOL_LIMIT_KEY = "datafusion.memory_pool_limit_bytes";
+    static final String PARQUET_MAX_NATIVE_ALLOCATION_KEY = "parquet.max_native_allocation";
```
Bukhtawar: Shouldn't this be based on how much native memory this process is allowed to consume

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -0,0 +1,183 @@
...
+     */
+    long computeNonHeapBase() {
+        return Math.max(0L, OsProbe.getInstance().getTotalPhysicalMemorySize() - Runtime.getRuntime().maxMemory());
+    }
+
+    /**
+     * Resolves the DataFusion native-pool contribution. Absent, zero, negative, or unparseable
+     * values all collapse to {@code 0L}.
+     */
+    long resolveDataFusionContribution(Settings settings) {
```
Bukhtawar: I wouldn't base it out per plugin pool, as this should be done by individual plugins

Bukhtawar (commented):
```
// Inputs
  long rssAnon = getProcessRssAnon();
  long heapCommitted = getHeapCommittedBytes();
  long nativeMemoryLimit = getNativeMemoryLimitSetting();
  double bufferPercent = getNativeMemoryBufferSetting();   // default 15%, configurable

  // Effective budget (reserve buffer for growth between polls + JVM non-heap overhead)
  long buffer = (long) (nativeMemoryLimit * bufferPercent / 100.0);
  long effectiveBudget = nativeMemoryLimit - buffer;

  // Calculation
  long nativeUsed = Math.max(0, rssAnon - heapCommitted);
  double utilization = (double) nativeUsed / effectiveBudget * 100.0;

  Settings

  node.native_memory.limit: 15500m
  node.native_memory.buffer_percent: 15    # default 15
```

Would this work better?

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: the parquet memory allocation now is calculated by 10% of total native memory available, So to calculate this i need to get the total physical memeory and total heap committted .

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: As we had discussed earlier, For now we will add the memory allocated to datafusion and parquet plugins. and use this as cap memory for native memory percentage used calculation.

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: this is not benchmarking factor. total native memory = ( max memory allocated to datafusion + max memory allocated to parquet ) * 1.2 ( adding 20% buffer )

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: now we read this cap from node settings

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: now we read this native memory cap from settings

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: have removed this and made native memory cap configurable

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: have made this native memory cap configurable instead of reading from each plugin

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
pradeep-L: have made native memory cap configurable rather than per plugin

server/src/main/java/org/opensearch/node/NodeResourceUsageStats.java
pradeep-L: done

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java
```diff
@@ -0,0 +1,134 @@
...
+                return this.settings.getClusterAdminNativeMemoryUsageLimit();
+            default:
+                throw new IllegalArgumentException(
+                    String.format(
+                        Locale.ROOT,
+                        "Admission control not Supported for AdmissionControlActionType: %s",
+                        admissionControlActionType.getType()
+                    )
+                );
+        }
```
Bukhtawar, line 132: How does individual accounting work? Isn't it node level?

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
```diff
@@ -0,0 +1,132 @@
...
+     * By default it will use the default cluster admin native memory usage limit.
+     */
+    public static final Setting<Long> CLUSTER_ADMIN_NATIVE_MEMORY_USAGE_LIMIT = Setting.longSetting(
+        "admission_control.cluster_admin.native_memory_usage.limit",
+        Defaults.CLUSTER_ADMIN_NATIVE_MEMORY_USAGE_LIMIT,
+        0,
+        100,
+        Setting.Property.Dynamic,
+        Setting.Property.NodeScope
+    );
```
Bukhtawar, line 79: Is this mostly from a configuration/settings update perspective? What actions does it restrict?

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java
pradeep-L: The reason there are separate settings for indexing, search, and cluster admin is because
Different request types have very different resource usage. OpenSearch does not want one overloaded workload to destroy the entire node.
but under layer it used Node level metrics only CPU, JVM heap. The native memory Admisison control follows this design principle.

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/settings/NativeMemoryBasedAdmissionControllerSettings.java
pradeep-L: cluster admin actions include Cluster state management , stats, hot threads , Shard allocation/recovery, Coordination tasks like elections, join.

server/src/main/java/org/opensearch/ratelimitting/admissioncontrol/controllers/NativeMemoryBasedAdmissionController.java
Bukhtawar: But we don't really know how much each of index/search consumes right?

pradeep-L: Yes , For a cluster with indexing heavy can restrict search request at 70 % and allow indexing to go upto 95%. 
To start with i have set all these 3 parameter to 95% default value.
Default values for CPU and IO.
admission_control.search.cpu_usage.limit	95
admission_control.indexing.cpu_usage.limit	95
admission_control.cluster.admin.cpu_usage.limit	95
admission_control.search.io_usage.limit	95
admission_control.indexing.io_usage.limit	95
admission_control.cluster_admin.io_usage.limit	100

Bukhtawar (approved):
Thanks, can you also verify `AdmissionControllerStats` shows rejection per search/indexing/admin

server/src/main/java/org/opensearch/node/NodeResourceUsageStats.java
pradeep-L: done

Issue #21088, opened 2026-04-02
[Feature Request] Support request-level timeout that returns partial results from completed shards


sekubos:
### Is your feature request related to a problem? Please describe

OpenSearch has two timeout mechanisms for search, neither of which fully addresses the need to bound total request latency while preserving partial results:

| Mechanism | Scope | Returns partial results? |
|---|---|---|
| timeout parameter | Per-shard | Yes — but shard-level only. Total request latency includes coordinator overhead, fan-out, merge/ reduce, and is affected by max_concurrent_shard_requests and shard count. Cannot predictably bound wall-clock time. |
| cancel_after_time_interval | Request-level (coordinator) | No — cancels the entire request without returning any results. |

There is no mechanism that combines request-level scope with partial result delivery.

The timeout parameter is difficult to reason about for bounding total request latency because:
- Coordinator overhead (fan-out, merge, reduce) is not included
- The number of shards queried in parallel depends on max_concurrent_shard_requests, node count, and shard allocation
- Resharding changes the relationship between shard timeout and request latency, requiring re-tuning
- Network latency between coordinator and data nodes is variable

### Describe the solution you'd like

Introduce a request-level timeout parameter (e.g., `request_timeout`) that:

1. Starts a timer at the coordinating node when the request is received
2. When the deadline is reached, stops waiting for outstanding shard responses
3. Merges and returns results from shards that have already responded
4. Indicates in the response which shards completed vs. were still pending

Example usage:
```
GET /my-index/_search?request_timeout=800ms&allow_partial_search_results=true
{
  "query": { "match": { "content": "search term" } }
}
```

Example response when the request-level timeout fires:
```
json
{
  "timed_out": true,
  "_shards": {
    "total": 40,
    "successful": 28,
    "skipped": 0,
    "failed": 0,
    "pending": 12
  }
```

### Related component

Search

### Additional context

Use-cases: 

Applications with strict end-to-end latency SLAs (e.g., "search must return within 800ms") need a single, predictable knob to bound total request time while returning the best available results. Today, the workaround is:

1. Set a shard-level timeout conservatively below the latency budget
2. Layer an application-level HTTP timeout on top
3. Tune max_concurrent_shard_requests to make the relationship more predictable
4. Re-tune all of the above every time the index is resharded

This is fragile and operationally expensive. A request-level timeout with partial result delivery would replace this entire pattern with a single parameter.

kkhatua:
@sekubos  one question about this.
[...]

PR #21192, opened 2026-04-09
Add timed_out shard count to _shards response

### LINKED ISSUE #21087: [Feature Request] Add timed_out shard count to shards response section for search timeout observability

When using the `timeout` request parameter with `allow_partial_search_results=true`, the search response only includes a top-level `timed_out: true` boolean. All shards — including those that timed out — are reported as "successful" in the` _shards` section. There is no way to determine how many shards actually timed out (1 shard vs 99% of them). 

This makes it impossible to assess result completeness. A response where 1 out of 40 shards timed out is fundamentally different 
from one where 39 out of 40 timed out, but both return the same `timed_out: true` with "`successful": 40`.

Example current response:
```
json
{
  "timed_out": true,
  "_shards": {
    "total": 40,
    "successful": 40,
    "skipped": 0,
    "failed": 0
  }
}
```

Add a timed_out count to the _shards response object, alongside the existing total, successful, skipped, and failed fields. 
Example:

```
json
"_shards": {
  "total": 40,
  "successful": 40,
  "skipped": 0,
  "failed": 0,
  "timed_out": 12
}
```

Timed-out shards would remain counted as `successful `for backward compatibility. The new field provides explicit visibility into how many shards returned truncated results.

### Related component

Search

1. `profile=true` parameter — Returns per-shard timing breakdowns, allowing you to infer which shards exceeded the timeout by comparing query time against the timeout value. However, profiling adds significant overhead and is not viable for production traffic.

2. `terminate_after` instead of timeout — Uses a document-count limit per shard instead of a time limit. Returns a 
terminated_early: true flag, which provides clearer truncation visibility. However, it bounds work (document count) rather than time, so it doesn't address the same use case. A shard scanning 1000 documents could take 5ms or 500ms depending on query complexity.

### Additional context

Use-cases: 
Applications using shard-level timeouts to bound query latency need to make runtime decisions about result quality — whether to display results with a "partial results" warning, retry the query, or fall back to cached results. Without knowing the ratio of timed-out shards, there is no way to distinguish between nearly-complete and nearly-empty result sets.

This is particularly important for:
- E-commerce search (showing partial product results with confidence indicators)
- Log analytics (knowing whether aggregation results are statistically meaningful)
- Real-time dashboards (deciding whether to display data or show a "data incomplete" state)

dzane17:
### Description

Adds a `timed_out` integer field to the `_shards` section of search responses, reporting how many shards had their query phase time out during Lucene document collection. This addresses the observability gap where `timed_out: true` at the top level gives no indication of whether 1 shard or 99% of shards timed out.

Timed-out shards remain counted as `successful` for backward compatibility (same behaviour as `skipped` shards). The new field is additive and does not change existing response semantics.

### Scope

Only counts shards that timed out during the shard query phase due to the `timeout` request parameter or `search.default_search_timeout` cluster setting. Does not cover `cancel_after_time_interval` (which cancels the entire request without returning a response) or the proposed `request_timeout` in #21088.

### Response format

```json
{
  ...
  "timed_out": true,
  "_shards": {
    "total": 10,
    "successful": 10,
    "skipped": 0,
    "failed": 0,
    "timed_out": 3    // new field
  }
}
```

### Invariants

- `timed_out >= 0`
- `timed_out > 0` iff `timed_out` boolean is `true`

### Testing

No timeout (timed_out count is 0):
```bash
curl -s 'http://localhost:9200/test-index/_search?timeout=10s' \
  -d '{"query": {"match_all": {}}, "size": 1}'
```
```json
{
  "timed_out": false,
  "_shards": {
    "total": 3,
    "successful": 3,
    "skipped": 0,
    "failed": 0,
    "timed_out": 0
  }
}
```

With timeout (all 3 shards timed out):
```bash
curl -s 'http://localhost:9200/test-index/_search?timeout=100ms' \
  -d '{"query": {"match_all": {}}, "size": 1}'
```
```json
{
  "timed_out": true,
  "_shards": {
    "total": 3,
    "successful": 3,
    "skipped": 0,
    "failed": 0,
    "timed_out": 3
  }
}
```

### Related Issues
Resolves #21087

### Check List

### DIFF
--- a/modules/lang-mustache/src/test/java/org/opensearch/script/mustache/MultiSearchTemplateResponseTests.java
+++ b/modules/lang-mustache/src/test/java/org/opensearch/script/mustache/MultiSearchTemplateResponseTests.java
@@ -223,6 +223,7 @@ public class MultiSearchTemplateResponseTests extends AbstractXContentTestCase<M
             .field("successful", 2)
             .field("skipped", 0)
             .field("failed", 0)
+            .field("timed_out", 0)
             .endObject()
             .startObject("hits")
             .startObject("total")
--- a/modules/lang-mustache/src/test/java/org/opensearch/script/mustache/SearchTemplateResponseTests.java
+++ b/modules/lang-mustache/src/test/java/org/opensearch/script/mustache/SearchTemplateResponseTests.java
@@ -220,6 +220,7 @@ public class SearchTemplateResponseTests extends AbstractXContentTestCase<Search
             .field("successful", 0)
             .field("skipped", 0)
             .field("failed", 0)
+            .field("timed_out", 0)
             .endObject()
             .startObject("hits")
             .startObject("total")
--- a/server/src/internalClusterTest/java/org/opensearch/search/SearchTimeoutIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/search/SearchTimeoutIT.java
@@ -96,6 +96,12 @@ public class SearchTimeoutIT extends ParameterizedStaticSettingsOpenSearchIntegT
             .get();
         assertTrue(searchResponse.isTimedOut());
         assertEquals(0, searchResponse.getFailedShards());
+        // Verify timed_out count in _shards is positive and consistent with the boolean
+        assertTrue("timedOutShards should be > 0 when timed_out is true", searchResponse.getTimedOutShards() > 0);
+        assertTrue(
+            "timedOutShards should be <= successful - skipped",
+            searchResponse.getTimedOutShards() <= searchResponse.getSuccessfulShards() - searchResponse.getSkippedShards()
+        );
     }
 
     public void testSimpleDoesNotTimeout() throws Exception {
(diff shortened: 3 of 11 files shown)


kkhatua:
@dzane17 
~~Shouldnt `total` = `successful` + ... + **`timed_out`**~~

Never mind... i didn't realize that you're preserving the old behaviour where `timed_out` shards are also reported as successful. 

Do your tests cover partial `timed_out`s and not just non-zero check? Not sure if the tests are covering partial timeouts because then `success` > `timed_out` > 0

dzane17:
@kkhatua That's correct. Timed-out shards are already counted as successful. Changing that would break clients that check `successful == total` to determine if the search completed without errors. Also you can say the shard did succeed. The results are just incomplete.

`SearchPhaseControllerTests.testTopDocsStatsSomeTimedOut` tests partial timeout.

kkhatua:
Tagging @jainankitk  , @cwperks  for review

PR #21193, opened 2026-04-09
Improve RoutingNode invariant assertion efficiency

andrross:
[BalanceUnbalancedClusterTests take 10 or 11 minutes][1] in CI. The reason is because of O(N^2) assertion behavior. This change moves the invariant assertions out of the individual muting calls up to the RoutingNodes level where it can do the assertion after bulk mutations. This change is safe because assertions are only moved out of package-private methods that are only called by RoutingNodes; the public API of RoutingNode still enforces all the same invariants.

[1]: https://build.ci.opensearch.org/job/gradle-check/74280/testReport/org.opensearch.cluster.routing.allocation/BalanceUnbalancedClusterTests/

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/cluster/routing/RoutingNodes.java
+++ b/server/src/main/java/org/opensearch/cluster/routing/RoutingNodes.java
@@ -174,6 +174,7 @@ public class RoutingNodes implements Iterable<RoutingNode> {
                 }
             }
         }
+        assert nodesToShards.values().stream().allMatch(RoutingNode::invariant);
     }
 
     private void addRecovery(ShardRouting routing) {
@@ -563,6 +564,7 @@ public class RoutingNodes implements Iterable<RoutingNode> {
         assert unassignedShard.unassigned() : "expected an unassigned shard " + unassignedShard;
         ShardRouting initializedShard = unassignedShard.initialize(nodeId, existingAllocationId, expectedSize);
         node(nodeId).add(initializedShard);
+        assert node(nodeId).invariant();
         inactiveShardCount++;
         if (initializedShard.primary()) {
             inactivePrimaryCount++;
@@ -591,6 +593,7 @@ public class RoutingNodes implements Iterable<RoutingNode> {
         ShardRouting target = source.getTargetRelocatingShard();
         updateAssigned(startedShard, source);
         node(target.currentNodeId()).add(target);
+        assert node(target.currentNodeId()).invariant();
         assignedShardsAdd(target);
         addRecovery(target);
         changes.relocationStarted(startedShard, target);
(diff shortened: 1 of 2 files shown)


andrross:
> The invariant() assertion is added after add, remove, and update calls in RoutingNodes, but not after all mutation paths. For example, promoteActiveReplicaShardToPrimary and other private methods that mutate RoutingNode state may not have assertions added. It should be verified that all mutation paths through RoutingNodes that call package-private RoutingNode methods now have corresponding invariant() assertions.

I think this is wrong. `promoteActiveReplicaShardToPrimary` goes through the `updateAssigned` path that has the assertions. I believe all mutation paths have the assertion.

andrross:
This version took 23 seconds instead of 11 minutes: https://build.ci.opensearch.org/job/gradle-check/74295/testReport/org.opensearch.cluster.routing.allocation/BalanceUnbalancedClusterTests/

Still not great, but better!

server/src/main/java/org/opensearch/cluster/routing/RoutingNodes.java
```diff
@@ -563,6 +564,7 @@ public ShardRouting initializeShard(
         assert unassignedShard.unassigned() : "expected an unassigned shard " + unassignedShard;
         ShardRouting initializedShard = unassignedShard.initialize(nodeId, existingAllocationId, expectedSize);
         node(nodeId).add(initializedShard);
+        assert node(nodeId).invariant();
```
cwperks, line 567: There are already assertions within the method, do we really need another assert here? Can this simply be `node(nodeId).invariant();`?

Should we also look to avoid assertions for RuntimeExceptions and throw a different type of exception?

andrross: `assert` is actually useful here. The code invoked in invariant() is quite expensive and we want to avoid executing it in non-test environments. The `assert` keyword means this line is a no-op unless assertions are enabled on the JVM.

server/src/main/java/org/opensearch/cluster/routing/RoutingNodes.java
```diff
@@ -174,6 +174,7 @@ public RoutingNodes(ClusterState clusterState, boolean readOnly) {
                 }
             }
         }
+        assert nodesToShards.values().stream().allMatch(RoutingNode::invariant);
```
cwperks, line 177: ok, took a while to understand the optimization but I see how in the constructor it calls `add` for each shard and every time `add` is called its running [`invariant()`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/routing/RoutingNode.java#L176-L182) twice which iterates through the shards (added up to that point in the constructor) leading to the n^2 runtime.

cwperks: Do you think there's a way we can skip the `invariant` checks for the mutating operations in the constructor specifically and do this check at the end? I'm not terribly deep in these classes, but having the invariant check within the mutating operation makes sense to me but should be skipped for the sake of this constructor until the very end.

andrross: That's possible, but given the coupling between these classes and that the mutating methods are package private, it seems fine to me for RoutingNodes to be responsible for calling the invariant check.

cwperks (approved):

PR #21194, opened 2026-04-09
Fix breaking changes by delegating and deprecating

msfroh:
### Description
We had a couple of attempts to address the breaking changes detector by restoring old methods and throwing a runtime exception. Instead of maintaining backward compatibility with existing APIs, this actually straight-up broke those APIs.

### Related Issues
N/A

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/IndexModule.java
+++ b/server/src/main/java/org/opensearch/index/IndexModule.java
@@ -268,7 +268,7 @@ public final class IndexModule {
     private final AnalysisRegistry analysisRegistry;
     private final EngineFactory engineFactory;
     private final EngineConfigFactory engineConfigFactory;
-    private SetOnce<Function<IndexService, CheckedFunction<DirectoryReader, DirectoryReader, IOException>>> indexReaderWrapper =
+    private final SetOnce<Function<IndexService, CheckedFunction<DirectoryReader, DirectoryReader, IOException>>> indexReaderWrapper =
         new SetOnce<>();
     private final Set<IndexEventListener> indexEventListeners = new HashSet<>();
     private final Map<String, TriFunction<Settings, Version, ScriptService, Similarity>> similarities = new HashMap<>();
@@ -653,6 +653,7 @@ public final class IndexModule {
         WARM_TO_HOT;
     }
 
+    @Deprecated(forRemoval = true)
     public IndexService newIndexService(
         IndexService.IndexCreationContext indexCreationContext,
         NodeEnvironment environment,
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/opensearch/index/IndexModule.java
```diff
@@ -679,11 +680,37 @@ public IndexService newIndexService(
         RemoteStoreSettings remoteStoreSettings,
         Supplier<Integer> clusterDefaultMaxMergeAtOnceSupplier
     ) throws IOException {
-        throw new UnsupportedOperationException(
-            "This API is removed in OpenSearch version 3.4.0. " + "Use the new overloaded newIndexService() method instead."
+        return newIndexService(
```
andrross, line 683: I can only assume, but I suspect the original author expected this method to not be used outside of this repository so it was "safe" to break the behavior. Have you run into cases where this method version was needed? I'd be concerned that the behavior is still broken (and untested) with the default and/or null versions.

msfroh: It's entirely theoretical, and IMO this method should not really be part of our "public" API anyway.

That said, the previous approach violates the spirit of the "detect breaking changes" check by turning the (hypothetical) compile-time breakage into a (hypothetical) runtime breakage. Mostly, I'm concerned about this pattern propagating to an API where it does matter.

I am probably engaging in a slippery-slope fallacy, but I dislike what I see as a bad precedent.

reta: I am 100% with @msfroh here, keeping public (or user exposed) APIs backward compatible is our promise (at least, it was). Any changes on that surface have to take it into account, - yes, it requires more careful approach but it is worth it.

reta (approved):

server/src/main/java/org/opensearch/index/IndexModule.java
```diff
@@ -712,15 +739,41 @@ public IndexService newIndexService(
...
+            translogFactorySupplier,
+            clusterDefaultRefreshIntervalSupplier,
+            fixedRefreshIntervalSchedulingEnabled,
+            shardLevelRefreshEnabled,
+            recoverySettings,
+            remoteStoreSettings,
+            replicator,
+            segmentReplicationStatsProvider,
+            clusterDefaultMaxMergeAtOnceSupplier,
+            null
```
msfroh, line 769: Actually, @andrross is right -- passing `null` here will result in a `NullPointerException` in the `IndexService` constructor.

I'm going to try adding at least *a* test for these various methods to make sure that we can create an `IndexService`. (I can't guarantee that the `IndexService` will behave properly, though.)

andrross (approved):

PR #21201, opened 2026-04-10
[DSL Plugin] Add support for pipeline aggregations

tanik98:
### Description

### Related Issues

### Check List


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/pipeline/BucketsPathResolver.java
```diff
@@ -0,0 +1,161 @@
...
+                }
+            }
+
+            // Search all granularities by metric name
+            String metricName = pathElements.getLast().name;
+            for (AggregationMetadataBuilder candidate : granularities.values()) {
+                if (candidate.containsAggregateField(metricName)) {
+                    return candidate;
+                }
+            }
```
nssuresh2007, line 157: Is this a safety fallback, or do we expect any scenario to land here? Since this may have perf concerns since we would iterate through every granularity here

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/pipeline/sibling/AvgBucketTranslator.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public Type type() {
+        return Type.SIBLING;
+    }
+
+    @Override
+    public RelNode translate(AvgBucketPipelineAggregationBuilder builder, RelNode input,
+                             ConversionContext ctx) throws ConversionException {
+        int colIndex = BucketsPathResolver.resolve(builder.getBucketsPaths()[0], input);
```
nssuresh2007, line 51: nit: Lets add a TODO that when we implement bucket_selector or bucket_script which can support more than one bucketsPaths, we need to relook at this instead of fetching 0th element

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/pipeline/BucketsPathResolver.java
```diff
@@ -0,0 +1,161 @@
...
+        }
+        return granularities.values().iterator().next();
```
nssuresh2007, line 159: I believe we would reach here only for invalid scenarios, since bucket path would be present for pipeline aggregation. Can we throw conversionException if we reach here?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/converter/SearchSourceConverter.java
```diff
@@ -110,15 +114,35 @@ public QueryPlans convert(SearchSourceBuilder searchSource, String indexName) th
...
+                builder.add(new QueryPlans.QueryPlan(QueryPlans.Type.AGGREGATION, agg));
+
+                // Sibling pipelines operate on the original base (before parent modifications)
+                for (PipelineAggregationBuilder pipelineBuilder : metadata.getPipelineBuilders()) {
+                    PipelineTranslator<PipelineAggregationBuilder> translator = resolvePipelineTranslator(pipelineBuilder);
+                    if (translator.type() == PipelineTranslator.Type.SIBLING) {
+                        RelNode pipelinePlan = translator.translate(pipelineBuilder, basePlan, ctx);
+                        builder.add(new QueryPlans.QueryPlan(QueryPlans.Type.AGGREGATION, pipelinePlan));
+                    }
+                }
```
nssuresh2007, line 145: Can we add a PipelineAggsConverter class which can do this logic?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/aggregation/pipeline/parent/BucketSortTranslator.java
```diff
@@ -0,0 +1,112 @@
...
+        RexBuilder rexBuilder = ctx.getRexBuilder();
+        List<FieldSortBuilder> sorts = builder.sorts();
+
+        // Apply gap policy to each sort field column.
+        // SKIP filters out rows with null sort values; INSERT_ZEROS replaces nulls with 0.
+        RelNode gapHandled = input;
+        for (FieldSortBuilder sort : sorts) {
+            int colIndex = resolveColumn(sort.getFieldName(), gapHandled);
+            gapHandled = GapPolicyHandler.apply(builder.gapPolicy(), gapHandled, colIndex, ctx);
+        }
```
nssuresh2007, line 61: Wouldn't this create one LogicalProject or LogicalFilter for each sort field? That would result in multiple sort nodes in the RelNode plan. Is that understanding correct?
If yes, although it can be optimized at later steps, do we need to think of creating a single node with all the condition clauses together?

PR #21202, opened 2026-04-10
Sandbox/foyer page cache

vishwasgarg18:
### Description
A disk page cache backed by [Foyer](https://github.com/foyer-rs/foyer) for caching Parquet column chunk reads on local NVMe, to be used by DataFusion's `TieredObjectStore` (as of now). Can be later integrated during write paths etc. as well.

### Related Issues

### Check List
- [✅] Unit Tests - 

```
glvishwa@bcd074a8699e rust % cargo test
   Compiling opensearch-page-cache v0.1.0 (/Volumes/workplace/OpenSearchOpenSourceNischayWs/OpenSearch/sandbox/libs/page-cache/src/main/rust)
    Finished `test` profile [optimized + debuginfo] target(s) in 38.33s
     Running unittests src/lib.rs (/Volumes/workplace/OpenSearchOpenSourceNischayWs/OpenSearch/sandbox/libs/dataformat-native/rust/target/debug/deps/opensearch_page_cache-c4d94011e167d8e8)

running 33 tests
test foyer::tests::evict_prefix_on_nonexistent_prefix_is_noop ... ok
test foyer::tests::cache_is_usable_after_clear ... ok
test foyer::tests::clear_on_empty_cache_is_safe ... ok
test foyer::tests::concurrent_evict_and_put_does_not_panic ... ok
test foyer::tests::after_evict_prefix_new_put_is_retrievable ... ok
test foyer::tests::evict_prefix_does_not_affect_other_files ... ok
test foyer::tests::ffm_create_with_invalid_utf8_returns_error ... ok
test foyer::tests::ffm_create_with_null_ptr_returns_error ... ok
test foyer::tests::ffm_destroy_with_negative_ptr_returns_error ... ok
test foyer::tests::ffm_destroy_with_zero_ptr_returns_error ... ok
test foyer::tests::evict_prefix_removes_all_ranges_for_file ... ok
test foyer::tests::evict_prefix_twice_is_safe ... ok
test foyer::tests::clear_removes_all_entries ... ok
test foyer::tests::ffm_create_destroy_lifecycle_no_leak ... ok
test foyer::tests::ffm_create_returns_positive_pointer ... ok
test foyer::tests::get_returns_exact_bytes_that_were_put ... ok
test foyer::tests::get_returns_none_for_unknown_key ... ok
test foyer::tests::large_value_round_trips_correctly ... ok
test foyer::tests::get_returns_none_for_wrong_range_on_known_path ... ok
test foyer::tests::key_index_is_empty_after_clear ... ok
test foyer::tests::key_index_has_no_entry_for_evicted_file ... ok
test foyer::tests::event_remove_after_evict_prefix_does_not_panic_or_corrupt_key_index ... ok
test range_cache::tests::range_key_contains_separator ... ok
test range_cache::tests::range_keys_for_different_paths_do_not_share_prefix ... ok
test range_cache::tests::range_keys_for_same_path_share_index_prefix ... ok
test foyer::tests::put_and_get_work_after_cache_nears_capacity ... ok
test foyer::tests::multiple_files_are_independent ... ok
test foyer::tests::put_same_key_twice_replaces_value ... ok
test foyer::tests::multiple_ranges_for_same_file_are_independent ... ok
test foyer::tests::concurrent_puts_to_different_files_do_not_corrupt ... ok
test foyer::tests::replace_event_does_not_duplicate_key_in_key_index ... ok
test foyer::tests::concurrent_put_and_get_same_file_does_not_panic ... ok
test foyer::tests::lru_eviction_removes_stale_keys_from_key_index ... ok

test result: ok. 33 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.58s

   Doc-tests opensearch_page_cache

running 1 test
test /Volumes/workplace/OpenSearchOpenSourceNischayWs/OpenSearch/sandbox/libs/page-cache/src/main/rust/src/range_cache.rs - range_cache::range_cache_key (line 74) ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.84s
```

- [ ] Functionality includes testing. -> Pending


sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
+/*
```
nishchay21: sandbox/libs/page-cache/src/main/rust/
├── Cargo.toml
└── src/
    ├── lib.rs              — re-exports, module declarations
    ├── traits.rs           — PageCache trait (get/put/evict/clear)
    ├── foyer_cache.rs      — FoyerCache implements PageCache
    └── ffm.rs              — extern "C" lifecycle functions

can we have something like above. FoyerCache implements PageCache. Tomorrow if you want moka or a custom LRU, just add another impl. CachedObjectStore is generic over P: PageCache.

sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+/// # Safety
+/// `dir_ptr` must point to `dir_len` consecutive valid UTF-8 bytes.
+#[no_mangle]
+pub unsafe extern "C" fn foyer_create_cache(
+    disk_bytes: u64,
+    dir_ptr: *const u8,
+    dir_len: u64,
+) -> i64 {
+    match std::str::from_utf8(std::slice::from_raw_parts(dir_ptr, dir_len as usize)) {
+        Ok(dir) => Box::into_raw(Box::new(FoyerCache::new(disk_bytes as usize, dir))) as i64,
```
nishchay21: This uses Box::into_raw / Box::from_raw. If the cache pointer is ever shared across threads on the Java side (e.g. multiple shards accessing the same cache), Box is unsound — only one owner allowed. Should use Arc for consistency and safety.

sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+use dashmap::DashMap;
+use foyer::{HybridCache, HybridCacheBuilder, DirectFsDeviceOptionsBuilder, LruConfig};
+use serde::{Deserialize, Serialize};
+
+// ── Value wrapper ─────────────────────────────────────────────────────────────
+
+/// Newtype wrapper around a byte slice that satisfies Foyer's `StorageValue`
+/// bound (`serde::Serialize + Deserialize`). Not part of the public API.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+struct CachedBytes(#[serde(with = "serde_bytes")] Vec<u8>);
```
nishchay21: Can we add tests. Need at minimum: put+get round-trip, get miss returns None, evict_file removes all ranges for a path, clear empties everything, concurrent put+get from multiple threads, FFM create+destroy lifecycle (no leak).

sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// After this call the handle is invalid and must not be used again.
+///
+/// # Safety
+/// `ptr` must be a value returned by [`foyer_create_cache`] that has not
+/// already been passed to this function.
+#[no_mangle]
+pub unsafe extern "C" fn foyer_destroy_cache(ptr: i64) {
+    if ptr > 0 {
+        drop(Box::from_raw(ptr as *mut FoyerCache));
```
nishchay21: should use Arc::from_raw instead of Box::from_raw, and add #[ffm_safe].

sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+/// as an opaque token by the caller — do not dereference or free it directly.
+/// Pass it back to [`foyer_destroy_cache`] when the cache is no longer needed.
+///
+/// Returns `-1` if `dir_ptr` is not valid UTF-8; returns a positive pointer
+/// on success.
+///
+/// # Safety
+/// `dir_ptr` must point to `dir_len` consecutive valid UTF-8 bytes.
+#[no_mangle]
+pub unsafe extern "C" fn foyer_create_cache(
```
nishchay21: If FoyerCache::new panics (it uses expect twice), unwinding across the FFI boundary is undefined behavior. Use native_bridge_common::ffm_safe macro for panic catching

sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+    disk_bytes: u64,
+    dir_ptr: *const u8,
+    dir_len: u64,
+) -> i64 {
```
nishchay21: Add Null check on directory pointer

sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    /// Insert a byte range into the cache.
+    ///
+    /// Foyer writes the entry to disk asynchronously in the background.
+    /// The entry is also recorded in the key index to support per-file eviction.
+    pub fn put(&self, path: &str, start: u64, end: u64, data: Bytes) {
+        let k = Self::make_key(path, start, end);
+        self.inner.insert(k.clone(), CachedBytes::from_slice(&data));
+        self.key_index.entry(path.to_string()).or_default().push(k);
```
nishchay21: key_index grows unbounded. When Foyer evicts entries from disk via LRU, the corresponding keys in key_index are never cleaned up can we see if we can clean these entries

sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+        format!("{}:{}-{}", path, start, end)
+    }
+
+    /// Look up a cached byte range by file path and byte offsets.
+    ///
+    /// Returns `Some(Bytes)` on a cache hit and `None` on a miss.
+    /// Blocks the calling thread while Foyer performs the disk lookup.
+    pub fn get(&self, path: &str, start: u64, end: u64) -> Option<Bytes> {
+        let k = Self::make_key(path, start, end);
+        self._runtime.block_on(async {
```
nishchay21: This is an issue if called from async thread can block threads.

sandbox/modules/foyer-page-cache/rust/src/lib.rs
```diff
@@ -0,0 +1,194 @@
...
+    /// `disk_bytes` capacity stored in `disk_dir`. Blocks the calling thread
+    /// until Foyer has completed asynchronous setup.
+    ///
+    /// # Panics
+    /// Panics if the Tokio runtime cannot be created or if Foyer fails to
+    /// build the cache (e.g. insufficient disk space or invalid path).
+    pub fn new(disk_bytes: usize, disk_dir: impl Into<PathBuf>) -> Self {
+        let disk_dir = disk_dir.into();
+        let rt = tokio::runtime::Runtime::new()
+            .expect("[foyer-page-cache] failed to create Tokio runtime");
```
nishchay21: If the tokio runtime or HybridCache build fails, this panics and crashes the JVM. Should return Result and let the FFM layer convert to an error code. See #[ffm_safe] pattern in native-bridge-common for panic catching.

sandbox/modules/foyer-page-cache/rust/src/lib.rs
vishwasgarg18: Agreed, updated.

Theoretically, on java side only upon close (node shutdown), the cache is destroyed. By that point, all shards should have stopped. So in practice, with correct usage, `Box` would work.

i understand that Arc is safer and is future proof for any changes (deallocating the cache heap memory midway between a query).

sandbox/modules/foyer-page-cache/rust/src/lib.rs
vishwasgarg18: `#[ffm_safe]` strictly requires `-> i64` return type. Hence we can't add it to foyer_destroy_cache function. 

Updated the Box to Arc though.

sandbox/modules/foyer-page-cache/rust/src/lib.rs
vishwasgarg18: Makes sense, updated.

sandbox/modules/foyer-page-cache/rust/src/lib.rs
vishwasgarg18: Yes!

sandbox/modules/foyer-page-cache/rust/src/lib.rs
vishwasgarg18: Yes, updated.
There is an event_listener - `on_leave` in `Foyer` which have integrated on our Foyer Cache to remove stale entries in case of any eviction on Foyer.

sandbox/modules/foyer-page-cache/rust/src/lib.rs
vishwasgarg18: Updated.

sandbox/modules/foyer-page-cache/rust/src/lib.rs
vishwasgarg18: Added!

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
```
vishwasgarg18: Pending usage, will do it in another change when integrating foyer cache with Parquet warm indexes (read paths for now).

vishwasgarg18 (commented):
.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,99 @@
...
+                ValueLayout.JAVA_LONG   // dir_len: u64
+            )
+        );
+
+        // void foyer_destroy_cache(i64 ptr)
+        FOYER_DESTROY_CACHE = linker.downcallHandle(
+            lib.find("foyer_destroy_cache").orElseThrow(),
+            FunctionDescriptor.ofVoid(ValueLayout.JAVA_LONG)
+        );
+        logger.info("[FoyerBridge] FFM handles resolved");
```
nishchay21: Nit:  — remove [FoyerBridge] prefix, logger name already includes class.

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * Destroy a cache previously created by {@link #createCache}.
+     *
+     * <p>After this call the handle is invalid and must not be used again.
+     *
+     * @param ptr the handle returned by {@link #createCache}
+     */
+    public static void destroyCache(long ptr) {
```
nishchay21, line 115: foyer_destroy_cache returns void. Inconsistent with codebase convention where all FFM functions return i64 (0=success, <0=error). Consider returning i64 for error reporting.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Create a Foyer cache with the given disk capacity and storage directory.
+     *
+     * @param diskBytes maximum disk space the cache may use, in bytes
+     * @param diskDir   path to the directory where Foyer stores cache data
+     * @return an opaque handle representing the cache instance; always positive on success
+     * @throws RuntimeException if the native call fails or the directory is invalid
+     */
+    public static long createCache(long diskBytes, String diskDir) {
```
nishchay21: no input validation. Add Objects.requireNonNull(diskDir, "diskDir") and if (diskBytes <= 0) throw new IllegalArgumentException("diskBytes must be positive").

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,99 @@
...
+        try (var call = new NativeCall()) {
+            var dir = call.str(diskDir);
+            long ptr = call.invoke(FOYER_CREATE_CACHE, diskBytes, dir.segment(), dir.len());
```
nishchay21: no check if returned ptr <= 0. Javadoc says "always positive on success" but code doesn't enforce. Add: if (ptr <= 0) throw new IllegalStateException("foyer_create_cache failed, ptr=" + ptr).

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Destroy a cache previously created by {@link #createCache}.
+     *
+     * <p>After this call the handle is invalid and must not be used again.
+     *
+     * @param ptr the handle returned by {@link #createCache}
+     */
+    public static void destroyCache(long ptr) {
+        NativeCall.invokeVoid(FOYER_DESTROY_CACHE, ptr);
```
nishchay21: destroyCache uses NativeCall.invokeVoid() but createCache uses try (var call = new NativeCall()).  Inconsistent. Pick one pattern.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
```diff
@@ -0,0 +1,78 @@
...
+ *       lifecycle components — native cache is destroyed last.
+ * </ol>
+ *
+ * @opensearch.experimental
+ */
+public final class PageCacheService implements Closeable {
+
+    private static final Logger logger = LogManager.getLogger(PageCacheService.class);
+
+    private final long cachePtr;
```
nishchay21: private final long cachePtr — should be volatile if accessed from multiple threads. Or document that PageCacheService is thread-safe because cachePtr is set once in constructor and never changes.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Creates a {@code PageCacheService} with the given disk quota and directory.
+     *
+     * @param diskBytes maximum disk space the cache may use, in bytes
+     * @param diskDir   directory in which Foyer stores cache data
+     * @return a new service instance; {@link #close()} must be called at shutdown
+     */
+    public static PageCacheService create(long diskBytes, String diskDir) {
```
nishchay21: PageCacheService.create() — no validation on inputs. diskBytes <= 0 or diskDir == null would propagate to native and crash. Validate here.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
```diff
@@ -0,0 +1,78 @@
...
+        return cachePtr;
+    }
+
+    /**
+     * Destroys the native cache. Called by {@code Node.close()} after all plugins
+     * have been stopped.
+     */
+    @Override
+    public void close() {
+        if (cachePtr != 0) {
```
nishchay21: close() — checks cachePtr != 0 but FoyerBridge.destroyCache also logs. Again duplicate logging. Also, close() can be called multiple times (Closeable contract). After first close, cachePtr is still non-zero — second close would double-destroy. Add AtomicBoolean closed guard, or use NativeHandle pattern.

sandbox/libs/page-cache/src/main/rust/src/foyer/ffm.rs
```diff
@@ -0,0 +1,43 @@
...
+        .map_err(|e| format!("invalid UTF-8 in dir path: {}", e))?;
+    Ok(Arc::into_raw(Arc::new(FoyerCache::new(disk_bytes as usize, dir))) as i64)
+}
+
+/// Destroy a [`FoyerCache`] previously created by [`foyer_create_cache`].
+///
+/// # Safety
+/// `ptr` must be a value returned by [`foyer_create_cache`] not yet destroyed.
+#[no_mangle]
+pub unsafe extern "C" fn foyer_destroy_cache(ptr: i64) {
```
nishchay21: foyer_destroy_cache — missing #[ffm_safe]. If Arc::from_raw panics (e.g. double-free), it's UB. Add #[ffm_safe] for consistency.

sandbox/libs/page-cache/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+impl EventListener for KeyIndexListener {
+    type Key   = String;
+    type Value = Vec<u8>;
+
+    fn on_leave(&self, reason: Event, key: &String, _value: &Vec<u8>) {
+        match reason {
+            Event::Evict | Event::Replace | Event::Remove => {
+                if let Some(colon_pos) = key.rfind(':') {
```
nishchay21: If the path itself contains : (Windows paths like C:\...), this breaks.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
chaitanya588: invokeVoid is a static method which doesn't have arena involvement as well. It is the practice we are following as well for invoking such fire and forgot kind of call patterns. So, this should be ok.

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBridge.java
vishwasgarg18: I had referred to the datafusion code to decide on the return type void.

But I think it makes sense to have i64 return type, incase of any panic in rust side will unwind across the FFM boundary → undefined behavior in C ABI.

Let's also callout in other piece of code as well.




Examples in datafusion.

```
#[ffm_safe]
#[no_mangle]
pub unsafe extern "C" fn df_create_global_runtime(
    memory_pool_limit: i64,
    spill_dir_ptr: *const u8,
    spill_dir_len: i64,
    spill_limit: i64,
) -> i64 {
    let spill_dir = str_from_raw(spill_dir_ptr, spill_dir_len).map_err(|e| format!("df_create_global_runtime: {}", e))?;
    api::create_global_runtime(memory_pool_limit, spill_dir, spill_limit)
        .map_err(|e| e.to_string())
}

#[no_mangle]
pub unsafe extern "C" fn df_close_global_runtime(ptr: i64) {
    api::close_global_runtime(ptr);
}

```


```
#[ffm_safe]
#[no_mangle]
pub unsafe extern "C" fn df_create_reader(
    table_path_ptr: *const u8,
    table_path_len: i64,
    files_ptr: *const *const u8,
    files_len_ptr: *const i64,
    files_count: i64,
) -> i64 {
    let table_path = str_from_raw(table_path_ptr, table_path_len).map_err(|e| format!("df_create_reader: {}", e))?;
    let mut filenames = Vec::with_capacity(files_count as usize);
    for i in 0..files_count as usize {
        let ptr = *files_ptr.add(i);
        let len = *files_len_ptr.add(i);
        filenames.push(str_from_raw(ptr, len).map_err(|e| format!("df_create_reader: {}", e))?.to_string());
    }
    let mgr = get_rt_manager()?;
    api::create_reader(table_path, filenames, &mgr).map_err(|e| e.to_string())
}

#[no_mangle]
pub unsafe extern "C" fn df_close_reader(ptr: i64) {
    api::close_reader(ptr);
}

```

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
vishwasgarg18: We do have checks at the rust layer.

```
foyer_create_cache

if dir_ptr.is_null() {
        return Err("dir_ptr is null".to_string());
    }
```

```
`disk_bytes` is declared as `u64` (an unsigned 64-bit integer). It can never be negative.
If `disk_bytes == 0` is passed, Foyer's `FsDeviceBuilder::with_capacity(0).build()` will fail with a clear error inside the `#[ffm_safe]` boundary.
```

*Having said that, I think it's better to add at the java layer to fail fast, updating.*

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
vishwasgarg18: Added in the pagecacheservice.

sandbox/libs/page-cache/src/main/rust/src/traits.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use bytes::Bytes;
+
+/// A disk page cache that stores arbitrary byte ranges keyed by `(path, start, end)`.
+///
+/// Implementations are expected to be `Send + Sync` so they can be shared
+/// across async tasks and threads.
+pub trait PageCache: Send + Sync {
+    /// Look up a cached byte range. Returns `Some(Bytes)` on hit, `None` on miss.
+    fn get(&self, path: &str, start: u64, end: u64)
```
chaitanya588: As of today for the existing page cache for storing block based index inputs, we have keys like this. Here also, foyer hybrid cache is going to expect a string key like this. Then do we need to pass on start-end byte range for the PageCache? Then in the future for reusing PageCache for block based cache entries, how will you support them?
```
/data/nodes/0/cache/a1b2c3d4-e5f6-7890-abcd-ef1234567890/0/_0.cfs_block_0
/data/nodes/0/cache/a1b2c3d4-e5f6-7890-abcd-ef1234567890/0/_0.cfs_block_1
```

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
vishwasgarg18: Yes, makes sense. Updated.

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub fn new(disk_bytes: usize, disk_dir: impl Into<PathBuf>) -> Self {
+        let disk_dir = disk_dir.into();
+        let key_index: Arc<DashMap<String, Vec<String>>> = Arc::new(DashMap::new());
+        let listener = Arc::new(KeyIndexListener { key_index: Arc::clone(&key_index) });
+
+        let rt = tokio::runtime::Runtime::new()
+            .expect("[page-cache] failed to create Tokio runtime");
+        let dir_clone = disk_dir.clone();
+        let inner = rt.block_on(async move {
+            HybridCacheBuilder::<String, Vec<u8>>::new()
```
chaitanya588, line 193: Few more options are missing here:
1. .with_policy(HybridCachePolicy::WriteOnInsertion)
2. UringIoEngineConfig: Instead of PsyncIoEngineConfig, if operating system supports io uring, then we should go with that configuration.
```
fn build_io_engine_config() -> Box<dyn IoEngineConfig> {
    #[cfg(target_os = "linux")]
    {
        UringIoEngineConfig::new().boxed()
    }
    #[cfg(not(target_os = "linux"))]
    {
        PsyncIoEngineConfig::new().boxed()
    }
}

```
3. We have multiple options to control BlockEngineConfig settings in granular way. Ex: block size should be 64MB, otherwise we will not be able to write the complete block into the disk and it will resultant to cache miss. 


Refer to https://github.com/chaitanya588/parquet-datafusion (parquet-datafusion/src/foyer_caching_store_s3.rs)

sandbox/libs/page-cache/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: In new implementation the separator is `\x1F` (ASCII Unit Separator, decimal 31), which cannot appear in any filesystem path or object-store URL — S3, GCS, and Azure all percent-encode it as `%1F`.  

This separator can also be used any other use case ( example - lucene based block store etc.)

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
vishwasgarg18: No, `volatile` on `cachePtr` is not required.

As per the java doc - 
```
An object is considered to be completely initialized when its constructor finishes. 
A thread that can only see a reference to an object after that object has been 
completely initialized is guaranteed to see the correctly initialized values for 
that object's `final` fields.
```
https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.5.2:~:text=An%20object%20is%20considered%20to%20be%20completely%20initialized%20when%20its%20constructor%20finishes.%20A%20thread%20that%20can%20only%20see%20a%20reference%20to%20an%20object%20after%20that%20object%20has%20been%20completely%20initialized%20is%20guaranteed%20to%20see%20the%20correctly%20initialized%20values%20for%20that%20object%27s%20final%20fields.


Added the java doc, thanks.

sandbox/libs/page-cache/src/main/rust/src/traits.rs
vishwasgarg18: Updated the `PageCache` trait to use an opaque `CacheKey` rather than `(path, start, end)`, so it is not tied to byte-range semantics. For Parquet, `range_cache_key(path, start, end)` builds the key.

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: > .with_policy(HybridCachePolicy::WriteOnInsertion)

The `.memory(1)` keeps a check that foyer is directly written to disk. It's also good to add `HybridCachePolicy::WriteOnInsertion)` to explicitly write it to disk immediately.


> UringIoEngineConfig: Instead of PsyncIoEngineConfig, if operating system supports io uring, then we should go with that configuration.

Got it. Based on the linux version (>=5.1) and OS (linux/mac etc.) updated the Engine Config.


> We have multiple options to control BlockEngineConfig settings in granular way. Ex: block size should be 64MB, otherwise we will not be able to write the complete block into the disk and it will resultant to cache miss.

Thanks for pointing it, updated.

vishwasgarg18 (commented):
.

sandbox/libs/dataformat-native/rust/Cargo.toml
```diff
@@ -40,6 +41,9 @@ log = "0.4"
 # Allocator
 mimalloc = { version = "0.1.48", default-features = false }
 
+bytes = "1.5"
+foyer = "0.22"
```
mgodwan: Why do we need a foyer dependency directly in the native dataformat lib?

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,108 @@
...
+
+        // i64 foyer_destroy_cache(i64 ptr)  — 0=success, <0=error pointer
+        FOYER_DESTROY_CACHE = linker.downcallHandle(
+            lib.find("foyer_destroy_cache").orElseThrow(),
+            FunctionDescriptor.of(
+                ValueLayout.JAVA_LONG,  // return: 0=ok, <0=error
+                ValueLayout.JAVA_LONG   // ptr
+            )
+        );
+        logger.info("FFM handles resolved");
```
mgodwan: [nit] Make this message more explicit as to what handles have been resolved

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,108 @@
...
+     * @throws RuntimeException if the native call fails or the directory is invalid
+     */
+    public static long createCache(long diskBytes, String diskDir) {
+        try (var call = new NativeCall()) {
+            var dir = call.str(diskDir);
+            long ptr = call.invoke(FOYER_CREATE_CACHE, diskBytes, dir.segment(), dir.len());
+            if (ptr <= 0) {
+                throw new IllegalStateException("foyer_create_cache returned invalid pointer: " + ptr);
+            }
+            logger.info("Cache created: ptr={}, diskBytes={}, dir={}", ptr, diskBytes, diskDir);
```
mgodwan: [nit] we should not log pointer address

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+public final class PageCacheService implements Closeable {
+
+    private static final Logger logger = LogManager.getLogger(PageCacheService.class);
+
+    private final long cachePtr;
+    /** Guards against double-close. {@link Closeable} contract requires idempotent close(). */
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+
+    private PageCacheService(long cachePtr) {
```
mgodwan: Can we instead maintain a Cache interface, which internally maintains the pointer. The current implementation ties the page cache service to a native implementation which can be avoided here

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Create a Foyer cache with the given disk capacity and storage directory.
+     *
+     * @param diskBytes maximum disk space the cache may use, in bytes
+     * @param diskDir   path to the directory where Foyer stores cache data
+     * @return an opaque handle representing the cache instance; always positive on success
+     * @throws RuntimeException if the native call fails or the directory is invalid
+     */
+    public static long createCache(long diskBytes, String diskDir) {
```
mgodwan: Does this need to be static? Can we not have a cache interface, and the FoyerCache instance can hold the handle/pointer and other services/layers can rely on the same

sandbox/libs/page-cache/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -0,0 +1,226 @@
...
+///
+/// On Linux with kernel ≥ 5.1, uses `io_uring` for non-blocking, batched disk
+/// I/O — significantly better throughput under concurrent reads than
+/// `pread`/`pwrite`. Falls back to POSIX synchronous I/O when:
+/// - The target platform is not Linux (e.g. macOS developer builds), or
+/// - The running Linux kernel is < 5.1 (io_uring was introduced in 5.1).
+fn build_io_engine_config() -> Box<dyn IoEngineConfig> {
+    #[cfg(target_os = "linux")]
+    {
+        if kernel_version_at_least(5, 1) {
```
mgodwan: log here for debugging

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -0,0 +1,226 @@
...
+                    // (put succeeds but entry is not stored, resulting in a cache miss).
+                    BlockEngineConfig::new(
+                        FsDeviceBuilder::new(dir_clone)
+                            .with_capacity(disk_bytes)
+                            .build()
+                            .expect("[page-cache] FsDevice build failed")
+                    )
+                    .with_block_size(64 * 1024 * 1024) // 64 MB — matches Parquet row-group and Lucene block sizes
+                )
+                .build()
```
mgodwan, line 217: How is the observability for foyer cache tracked? Do we need to enable a metrics registry as well?

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
```diff
@@ -0,0 +1,101 @@
...
+ * <p>Thread-safe. {@code cachePtr} is {@code final} and written exactly once in
+ * the constructor; per JLS §17.5, {@code final} fields have safe-publication
+ * semantics — all threads that obtain a reference to this object after construction
+ * are guaranteed to see the correct value without {@code volatile} or additional
+ * synchronisation. {@link #close()} is guarded by an {@link java.util.concurrent.atomic.AtomicBoolean}
+ * to make it idempotent and safe for concurrent or repeated invocations.
+ *
+ * @opensearch.experimental
+ */
+public final class PageCacheService implements Closeable {
```
Bukhtawar: This looks more like a `NativeHandle`

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
vishwasgarg18: Updated. Resolving.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
vishwasgarg18: Yes, updated in the rest of the code as well. Thanks for pointing.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
vishwasgarg18: Yes.
Added a `PageCache` interface inside the PageCacheservice. Now, PageCacheService no longer holds any cache implementation details. And FoyerPageCache implements PageCache.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/foyer/FoyerBridge.java
vishwasgarg18: Yes, as mentioned above created `PageCache` interface inside the lifecycle manager (`PageCacheService`) and `FoyerPageCache` implementing the `PageCache`.

> Does this need to be static? 

I think FoyerBridge.createCache and FoyerBridge.destroyCache are correctly static. FoyerBridge is pure stateless FFM call wrapper with no instance state.

sandbox/libs/dataformat-native/rust/Cargo.toml
vishwasgarg18: Actually, we don't need it. Put it in page-cache.

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: 1. On-demand snapshots via `_nodes/stats` (in progress):  We are enhancing the `_nodes/stats` API to include Foyer cache statistics, disk usage, hit/miss counts, and capacity by querying the `HybridCache` instance directly at request time. No metrics registry is needed for this.

We will have to build something like DefaultStatsCounter to get metrics like hit/miss etc that is not provided by hybridCache out of the box.

3. Continuous monitoring via `mixtrics`: Foyer provides a pluggable metrics registry abstraction called `mixtrics` (https://github.com/foyer-rs/mixtrics). To enable continuous monitoring, a registry backed by  OpenTelemetry can be injected via `.with_metrics_registry(...)` on the cache builder.

As per my understanding we will have to write the Rust equivalent of `telemetry-otel`  (a java plugin) initializing the Rust OTel SDK with an OTLP exporter pointing to the required beaglestone collector.

**I think we can take continuous monitoring as a followup.**

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCacheService.java
vishwasgarg18: Yes, correct. Renamed it to `PageCacheHandle`. And kept `PageCache` interface inside `PageCacheHandle`

vishwasgarg18 (commented):
.

vishwasgarg18 (commented):
.

sandbox/libs/page-cache/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -0,0 +1,234 @@
...
+                    // DataFusion reads Parquet row groups of up to 64 MB; Lucene blocks are
+                    // also 64 MB. A block_size smaller than the entry causes a silent drop
+                    // (put succeeds but entry is not stored, resulting in a cache miss).
+                    BlockEngineConfig::new(
+                        FsDeviceBuilder::new(dir_clone)
+                            .with_capacity(disk_bytes)
+                            .build()
+                            .expect("[page-cache] FsDevice build failed")
+                    )
+                    .with_block_size(64 * 1024 * 1024) // 64 MB — matches Parquet row-group and Lucene block sizes
```
bharath-techie: Should this be configurable via settings ?

sandbox/libs/page-cache/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -0,0 +1,234 @@
...
+/// - The target platform is not Linux (e.g. macOS developer builds), or
+/// - The running Linux kernel is < 5.1 (io_uring was introduced in 5.1).
+/// - https://kernelnewbies.org/Linux_5.1
+fn build_io_engine_config() -> Box<dyn IoEngineConfig> {
+    #[cfg(target_os = "linux")]
+    {
+        let release = std::fs::read_to_string("/proc/sys/kernel/osrelease")
+            .unwrap_or_else(|_| "unknown".to_string());
+        let release = release.trim();
+        if kernel_version_at_least(5, 1) {
```
bharath-techie: The choice of the engine after 5.1 - does it make sense to always go with io uring ? or should this be configurable

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -0,0 +1,234 @@
...
+        let listener = Arc::new(KeyIndexListener { key_index: Arc::clone(&key_index) });
+
+        let rt = tokio::runtime::Runtime::new()
+            .expect("[page-cache] failed to create Tokio runtime");
+        let dir_clone = disk_dir.clone();
+        let inner = rt.block_on(async move {
+            HybridCacheBuilder::<String, Vec<u8>>::new()
+                .with_name("page-cache")
+                .with_event_listener(listener)
+                .memory(1)
```
bharath-techie, line 196: Add comments maybe for this as well on why we are doing this

sandbox/libs/page-cache/src/main/rust/src/foyer/foyer_cache.rs
nishchay21: I agree this should be configurable

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -0,0 +1,234 @@
...
+    ///
+    /// # Panics
+    /// Panics if the Tokio runtime cannot be created or if Foyer fails to
+    /// build the cache (e.g. insufficient disk space or invalid path).
+    pub fn new(disk_bytes: usize, disk_dir: impl Into<PathBuf>) -> Self {
+        let disk_dir = disk_dir.into();
+        let key_index: Arc<DashMap<String, Vec<String>>> = Arc::new(DashMap::new());
+        let listener = Arc::new(KeyIndexListener { key_index: Arc::clone(&key_index) });
+
+        let rt = tokio::runtime::Runtime::new()
```
bharath-techie, line 187: Just for my understanding, why do we need a tokio rt just for the creation of foyer cache ?

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCache.java
```diff
@@ -0,0 +1,38 @@
...
+ * callers are introduced. Planned additions include:
+ * <ul>
+ *   <li>{@code evict(String prefix)} — invalidate entries when a shard closes</li>
+ *   <li>{@code diskCapacityBytes()} — for budget management and stats</li>
+ *   <li>{@code diskUsageBytes()} — for stats reporting</li>
+ * </ul>
+ *
+ * @opensearch.experimental
+ */
+public interface PageCache extends Closeable {
```
bharath-techie: Why is this called page cache - isn't foyer implementation transparent hybrid cache ?

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: Foyer's `HybridCacheBuilder.build()` is an async function. During initialization we do real work like that requires runtime engine. Example - 
1. Creates the io_uring ring
2. Probes kernel features
3. Spawns background I/O worker tasks


------------------------
For more context 

Node.java (Java thread, startup)

     ->  FoyerBridge.createCache()  
               -> FFM ->  foyer_create_cache() [Rust]
                                         -> // build() is async, need a runtime to drive it
                                                   -> rt = Runtime::new()
                                                   -> rt.block_on(builder.build())
                                                           -> async init: open disk, probe io_uring,
                                          -> drop(rt)        // ephemeral, no longer needed
                                         -> return cache ptr

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCache.java
vishwasgarg18: The intent is build format and implementation-agnostic interface that any cache backing can implement. Foyer today and perhaps custom LRU based cache tomorrow. The name needs to hold at that level of generality.


> isn't foyer implementation transparent hybrid cache ?

You're right that Foyer is technically a hybrid cache (DRAM + disk) but we've explicitly disabled the DRAM tier with `.memory(1)`. So, essentially it's disk-only for our use case. Also we don't want the interface name to reflect that detail at all.


> Confusion with OS Page Cache
`PageCache` does sound confusingly close to the Linux OS page cache, fair point. Essentially we're doing the same thing as the OS page cache but at a different tier. the OS caches file pages in DRAM to avoid disk I/O, we cache file pages on local SSD to avoid object storage fetches. Same concept, different tier. 

**I'll add a Javadoc comment to make this explicit.**


--------


Other names considered
1. RangeCache -> Too specific to Parquet file format. Tomorrow any cache implementation is expected to present here.
2. FormatCache -> Again, doesn't fit for lucene and any generic use case.
3. DiskCache -> Too generic, doesn't talk about what is cached.

sandbox/libs/page-cache/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: Done. I have kept PageCacheSettings in `org.opensearch.index.store.remote.filecache;` since we have placed the page cache in sandbox/lib.

sandbox/libs/page-cache/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: I have updated this as a setting, default values of io engine is decided basis on the linux version, but can be overriden as settings incase even though the linux version >=5.1 but we need to set different io engine.

vishwasgarg18 (commented):
Addressed comments!

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCache.java
bharath-techie: Isn't all implementations block caches , since it caches the blocks of byte ranges. 

> You're right that Foyer is technically a hybrid cache (DRAM + disk) 

Also I was implying on the "transparent" hybrid cache - hybrid cache was not the suggestion here.

Page cache somehow doesn't sound right to me, why not `BlockCache` if we are keeping it generic or something better.

vishwasgarg18 (commented):
Update the page-cache to block cache.

sandbox/libs/page-cache/src/main/java/org/opensearch/pagecache/PageCache.java
vishwasgarg18: Decided to go ahead with Block Cache.

bharath-techie (approved):
LGTM

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBlockCache.java
```diff
@@ -0,0 +1,94 @@
...
+ * Foyer-backed implementation of {@link BlockCache}.
+ *
+ * <p>Holds the native cache handle privately. Callers interact with this class
+ * through the {@link BlockCache} interface. Native-aware callers that need the
+ * underlying handle must cast to {@code FoyerBlockCache} and call
+ * {@link #nativeCachePtr()}.
+ *
+ * @opensearch.experimental
+ */
+public final class FoyerBlockCache implements BlockCache {
```
Bukhtawar, line 29: Foyer is a native store implementation. We shouldn't have this in libs

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBridge.java
```diff
@@ -0,0 +1,126 @@
...
+ * exposed here — they are called directly from the native layer without
+ * crossing the Java boundary.
+ *
+ * <p>{@link #createCache} returns an opaque {@code long} handle that represents
+ * the native cache instance. The handle must be passed to {@link #destroyCache}
+ * exactly once when the cache is no longer needed.
+ *
+ * @opensearch.experimental
+ */
+public final class FoyerBridge {
```
Bukhtawar, line 39: Lets not map it to the implementation yet. Foyer should be exposed only in the native repo

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBlockCache.java
vishwasgarg18: We have moved foyer block cache as a separate plugin now.

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBridge.java
vishwasgarg18: We have moved foyer block cache as a separate plugin now.

PR #20742, opened 2026-02-27
Validation of the `_source` to reject contradicting and ambiguous requests

urmichm:
### Description
**Validation** of the `_source` object added to avoid confusion and reject contradicting requests:
- `"_source": { "includes": "text", "excludes": ["title", "text"] }` The `text` field is defined in both `includes` and `excludes`. Contradiction. 🚫

**Deprecation** logs added for ambiguous requests:
- `"_source": {}`  At lease one of `includes` or `excludes` shall be defined.  
- `"_source": []` Explicitly defined empty array of `includes` is ambiguous.
- `"_source": { "includes": [], "excludes": ["title"] }` or `_source: { "includes": ["title"], "excludes": [] }` Explicitly defined empty array of `excludes` or `includes` is ambiguous.

To include the whole `_source` object or to exclude it completely, **the existing** `boolean` logic is encouraged.
`"_source": true` and `"_source": false`

Unit tests and [yml tests](https://github.com/opensearch-project/OpenSearch/blob/main/rest-api-spec/src/main/resources/rest-api-spec/test/README.md) added to test behaviour of parsing implementation. 
Existing tests have been extended.  

Follow-up to Pull Request https://github.com/opensearch-project/OpenSearch/pull/21086

### Related Issues
Resolves #20612

### Check List
- [x] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21203, opened 2026-04-10
Validation of the `_source` object in request and deprecation of ambiguous usage 

### LINKED ISSUE #20612: [Feature Request] [BUG] `_source` include / exclude validation

urmichm:
### Description
**Validation** of the `_source` object added to avoid confusion and reject contradicting requests:
- `"_source": { "includes": "text", "excludes": ["title", "text"] }` The `text` field is defined in both `includes` and `excludes`. Contradiction. 🚫

**Deprecation** logs added for ambiguous requests:
- `"_source": {}`  At lease one of `includes` or `excludes` shall be defined.  
- `"_source": []` Explicitly defined empty array of `includes` is ambiguous.
- `"_source": { "includes": [], "excludes": ["title"] }` or `_source: { "includes": ["title"], "excludes": [] }` Explicitly defined empty array of `excludes` or `includes` is ambiguous.

To include the whole `_source` object or to exclude it completely, **the existing** `boolean` logic is encouraged.
`"_source": true` and `"_source": false`

Unit tests and [yml tests](https://github.com/opensearch-project/OpenSearch/blob/main/rest-api-spec/src/main/resources/rest-api-spec/test/README.md) added to test behaviour of parsing implementation. 
Existing tests have been extended.  

Follow-up to Pull Request https://github.com/opensearch-project/OpenSearch/pull/21086

### Related Issues
Resolves #20612

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
+++ b/server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
@@ -50,7 +50,6 @@ import org.opensearch.rest.RestRequest;
 
 import java.io.IOException;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.Map;
@@ -283,7 +282,6 @@ public class FetchSourceContext implements Writeable, ToXContentObject {
             }
         }
         if (includes.length == 0 && excludes.length == 0) {
-            // no valid field names -> empty or unrecognized fields; deprecated
             deprecationLogger.deprecate(
                 "empty_source_object",
                 "An empty object was provided as [_source]. Provide at least one of ["
@@ -296,8 +294,7 @@ public class FetchSourceContext implements Writeable, ToXContentObject {
         return new FetchSourceContext(true, includes, excludes);
     }
 
-    private static String[] parseSourceFieldArray(XContentParser parser, ParseField parseField)
-        throws IOException {
+    private static String[] parseSourceFieldArray(XContentParser parser, ParseField parseField) throws IOException {
         Set<String> sourceArr = new LinkedHashSet<>(); // preserves the order, removes the duplicates
         while (parser.nextToken() != XContentParser.Token.END_ARRAY) {
             if (parser.currentToken() == XContentParser.Token.VALUE_STRING) {
--- a/server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
+++ b/server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
@@ -47,12 +47,13 @@ import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.rest.RestRequest;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.List;
+import java.util.HashSet;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Function;
 
+// ISSUE-20612 Source Validation
 /**
  * Context used to fetch the {@code _source}.
  *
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
urmichm, line 340: I would like to update this method, so it does not generate deprecated objects. However, I am not sure if it would result in _breaking change_. Could you share your opinion?

cc: @bowenlan-amzn @andrross

urmichm:
> The validateAmbiguousFields() method is called in the constructor FetchSourceContext(boolean, String[], String[]), but the constructor FetchSourceContext(StreamInput in) does NOT call validateAmbiguousFields(). This means deserialized instances from the wire could bypass validation. Consider whether validation should also be applied during deserialization, or if it's safe to assume wire data is always valid.

Initially, there was validation on `StreamInput` constructor, but based on previous comments (Closed PR https://github.com/opensearch-project/OpenSearch/pull/20742), the `StreamInput` constructor used mostly for backward compatibility and previously serialized objects. That's why the validation shall not be performed there.

urmichm:
> The AMBIGUOUS_FIELD_MESSAGE constant uses {} as a placeholder (SLF4J/log4j style), but OpenSearchException and ParsingException constructors may not support this format string substitution. Verify that the exception message is correctly formatted with the actual field name substituted, rather than displaying the literal {} in error responses.

ParsingException extends OpenSearchException, which uses the `{}` format

server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
```diff
@@ -206,9 +266,17 @@ private static FetchSourceContext parseSourceObject(XContentParser parser) throw
                 }
                 case XContentParser.Token.VALUE_STRING -> {
                     if (INCLUDES_FIELD.match(currentFieldName, parser.getDeprecationHandler())) {
-                        includes = new String[] { parser.text() };
+                        String includeEntry = parser.text();
+                        if (excludes.contains(includeEntry)) {
+                            throw new ParsingException(parser.getTokenLocation(), AMBIGUOUS_FIELD_MESSAGE, includeEntry);
```
andrross: Bit of a nitpick here, but I think disallowing contradicting includes/excludes is semantic validation, rather than a parsing validation. You've added a validation check to the constructor. Do we need this validation here, as opposed to letting it fall to the check in the constructor?

server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
```diff
@@ -87,6 +96,19 @@ public FetchSourceContext(StreamInput in) throws IOException {
...
 
+    /**
+     * The same entry cannot be both included and excluded in _source.
+     * Since the constructors are public, this validation is required to be called in the constructor.
+     * */
+    private void validateAmbiguousFields() {
+        Set<String> includeSet = new HashSet<>(Arrays.asList(this.includes));
+        for (String exclude : this.excludes) {
+            if (includeSet.contains(exclude)) {
+                throw new OpenSearchException(AMBIGUOUS_FIELD_MESSAGE, exclude);
```
andrross: Should this be an IllegalArgumentException?

server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
andrross: Would you update this to do `"source": true` if includes and excludes are empty?

In general it seems reasonable to change this to not generate deprecated formats.

Side note, in the code today if I create an object from the JSON `"source":true` and then invoke this toXContent method, will it generate `"source": {"includes":[], "excludes":[]}` ?

urmichm: Yes, if the arrays `includes` and `excludes` are empty, we must have `"source": true`, but we have `"source": {"includes":[], "excludes":[]}`. Updating the method is a good approach, in my opinion. 

By updating the method, we do not need to update the unrelated unit tests of `RequestConvertersTests` :) 

The update is here: https://github.com/opensearch-project/OpenSearch/pull/21203/changes/e47acbb4111b971839547ba8d4b87f84f55e9bf4

server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
urmichm: The validation in the constructor is required because the constructor is `public` and used in many places, which sometimes duplicates the parsing.

I think we can gradually refactor the code to use the existing `fromXContent` method and maybe the new methods `parseSourceObject` and `parseSourceArray`. With this idea in mind, the constructors could be deprecated (in the future) in favour of those methods, which is why I kept validation in the methods.

server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
urmichm: I can change this. I was not sure about `IllegalArgumentException`...  `OpenSearchException` looked more appropriate :)

server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
andrross: You will still need the constructor, even if they eventually are private. I'm thinking it would be better to keep semantic validation that throws an IllegalArgumentException in the constructor as opposed to throwing a parsing exception here. It would consolidate the semantic validation into a single place as well.

server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
urmichm: Updated in https://github.com/opensearch-project/OpenSearch/pull/21203/changes/5e984334d3f94dbc7b3aa894d7977a51bdbb5571

server/src/main/java/org/opensearch/search/fetch/subphase/FetchSourceContext.java
urmichm: Okay, I agree!

urmichm:
@andrross The pipeline is green :)
Could you have a look once more?

andrross (approved):

PR #21204, opened 2026-04-10
Add tiered object storage crate for warm node routing

nishchay21:
## Summary

Introduces two Rust crates under `sandbox/` that provide the foundational
file registry and object store routing for warm nodes to read parquet files
from remote stores (S3, GCS, Azure, FS).

## Components

**opensearch-tiered-storage** (`sandbox/modules/tiered-storage/src/main/rust/`)
- `FileRegistry` trait — 8-method interface for file tracking with RAII ref counting
- `TieredStorageRegistry` — DashMap-backed implementation with atomic ref counts,
  metrics counters, and `ReadGuard` that auto-acquires on creation / auto-releases on drop
- `TieredObjectStore` — `ObjectStore` impl that routes reads between local and remote
  stores based on registry metadata. Owns `remote_stores` and resolves cached store
  references at registration time (single DashMap lookup on warm read path)
- FFM bridge — `extern "C"` functions with `#[ffm_safe]` panic catching for Java interop

**opensearch-remote-store** (`sandbox/libs/remote-store/src/main/rust/`)
- Factory pattern for creating S3/GCS/Azure/FS backends from JSON config
- `RemoteObjectStore` logging wrapper

### Check List
- [x] Commits are signed off


Bukhtawar (commented):
Can we have tiered-storage as a part of the data-format plugin itself? Also we might have to re-home various cloud vendor based implementations.

nishchay21:
> Can we have tiered-storage as a part of the data-format plugin itself? Also we might have to re-home various cloud vendor based implementations.

@Bukhtawar placed remote-store in libs/ so it's shareable — open to moving to a different common directory if there's a preferred convention.

On the tiered-storage crate kept as a separate module so any data-format plugin — parquet, CSV, or future formats — can share the same TieredObjectStore without duplicating routing logic. The parquet plugin will create and return the TieredObjectStore implementation — the code just lives in a common place so other plugins can reuse it too. Would love to hear your thoughts on this.

Bukhtawar:
Anything common across native formats place it in `libs/tiered-store` anything specific to data format move to respective data format plugins. The module isn't really an OpenSearch module, so lets avoid the pattern. Anything related to cloud vendor, we might have to place them in the respective repository plugins(S3/GCP/Azure). I realise there might be challenges since we typically are using a single `object_store` crate. The object_store crate with features = ["aws", "gcp", "azure"] pulls in all three cloud SDKs. On a deployment that only uses S3, you're compiling and shipping GCS and Azure code. 

I was thinking 

```
sandbox/libs/tiered-store/src/main/rust/     ← shared crate (no cloud SDKs)
├── Cargo.toml
├── src/
│   ├── lib.rs
│   ├── types.rs                    ← from tiered-storage (FileLocation, TieredFileEntry, ReadGuard, errors)
│   ├── registry/
│   │   ├── mod.rs                  ← from tiered-storage
│   │   ├── traits.rs               ← FileRegistry trait
│   │   └── tiered_registry.rs      ← TieredStorageRegistry (DashMap impl)
│   ├── tiered_object_store.rs      ← from tiered-storage (ObjectStore routing)
│   ├── remote_object_store.rs      ← from remote-store (logging wrapper)
│   ├── factory.rs                  ← StoreFactory with trait-based backend registry (no hardcoded match)
│   ├── backends/
│   │   ├── mod.rs
│   │   └── fs.rs                   ← from remote-store (always available, useful for testing)
│   └── ffm.rs                      ← from tiered-storage (FFM bridge)

plugins/repository-s3/src/main/rust/                  ← S3-specific crate
├── Cargo.toml                      ← depends on object_store with "aws" feature + tiered-native-format
├── src/
│   ├── lib.rs
│   └── s3.rs                       ← from remote-store/backends/s3.rs + extern "C" entry point

plugins/repository-gcs/src/main/rust/                 ← GCS-specific crate
├── Cargo.toml                      ← depends on object_store with "gcp" feature + tiered-native-format
├── src/
│   ├── lib.rs
│   └── gcs.rs                      ← from remote-store/backends/gcs.rs + extern "C" entry point

plugins/repository-azure/src/main/rust/               ← Azure-specific crate
├── Cargo.toml                      ← depends on object_store with "azure" feature + tiered-native-format
├── src/
│   ├── lib.rs
│   └── azure.rs                    ← from remote-store/backends/azure.rs + extern "C" entry point

```

nishchay21:
@Bukhtawar here is what I was thinking, How it wires together at startup and remote can live independent of tiered as well. We can start for having this in warm but later service can be exposed for other usecases. Making sure current code base is untouched 
- Node.java → pluginsService.filterPlugins(NativeRemoteObjectStoreProvider.class) → discovers S3/GCS/Azure/FS plugins
- Node.java → new NativeRemoteObjectStoreService(providers) → binds to Guice
- Tiered storage module injects NativeRemoteObjectStoreService
- On warm shard open → service.ensureStore(repoName, "s3", configJson) → S3 plugin creates Rust ObjectStore → pointer cached
- ts_add_remote_store_ptr(tieredStorePtr, repoName, storePtr) → registers in Rust TieredObjectStore
- DataFusion reads parquet → TieredObjectStore routes to remote S3 ObjectStore 
```
=== RUST SIDE ===

sandbox/libs/tiered-storage/src/main/rust/         ← shared crate (NO cloud SDKs)
├── Cargo.toml                                      ← object_store (base only)
├── src/
│   ├── lib.rs
│   ├── types.rs                                    ← FileLocation, TieredFileEntry, ReadGuard
│   ├── registry/
│   │   ├── traits.rs                               ← FileRegistry trait
│   │   └── tiered_registry.rs                      ← TieredStorageRegistry (DashMap)
│   ├── tiered_object_store.rs                      ← ObjectStore routing (local vs remote)
│   └── ffm.rs                                      ← FFM bridge (ts_* functions)

sandbox/plugins/repository-s3/src/main/rust/        ← S3 crate (aws feature only)
sandbox/plugins/repository-gcs/src/main/rust/       ← GCS crate (gcp feature only)
sandbox/plugins/repository-azure/src/main/rust/     ← Azure crate (azure feature only)
sandbox/plugins/repository-fs/src/main/rust/        ← FS crate (testing, no cloud SDK)

=== JAVA SIDE ===

server/src/main/java/org/opensearch/
├── plugins/
│   └── NativeRemoteObjectStoreProvider.java        ← SPI: stateless factory per backend
│                                                      repositoryType(), createNativeStore(), destroyNativeStore()
├── repositories/
│   └── NativeRemoteObjectStoreService.java         ← Node-level service: aggregates providers,
│                                                      caches pointers, ensureStore(), getStorePtr()

sandbox/plugins/repository-s3/src/main/java/
└── .../s3/native_store/
    └── S3NativeObjectStorePlugin.java              ← implements NativeRemoteObjectStoreProvider
                                                       FFM → s3_create_store / s3_destroy_store

sandbox/plugins/repository-gcs/src/main/java/
└── .../gcs/native_store/
    └── GcsNativeObjectStorePlugin.java             ← implements NativeRemoteObjectStoreProvider
                                                       FFM → gcs_create_store / gcs_destroy_store

sandbox/plugins/repository-azure/src/main/java/
└── .../azure/native_store/
    └── AzureNativeObjectStorePlugin.java           ← implements NativeRemoteObjectStoreProvider
                                                       FFM → azure_create_store / azure_destroy_store

sandbox/plugins/repository-fs/src/main/java/
└── .../fs/native_store/
    └── FsNativeObjectStorePlugin.java              ← implements NativeRemoteObjectStoreProvider
                                                       FFM → fs_create_store / fs_destroy_store

```

sandbox/plugins/native-repository-azure/src/main/java/org/opensearch/repositories/azure/native_store/AzureNativeObjectStorePlugin.java
```diff
@@ -0,0 +1,81 @@
...
+import java.lang.foreign.SymbolLookup;
+import java.lang.foreign.ValueLayout;
+import java.lang.invoke.MethodHandle;
+
+/**
+ * Stateless factory that creates native (Rust) Azure ObjectStore instances via FFM.
+ *
+ * @opensearch.experimental
+ */
+public class AzureNativeObjectStorePlugin extends Plugin implements NativeRemoteObjectStoreProvider {
```
mgodwan, line 44: Won't this add dependency on FFI/M (Java25) beyond sandbox. Are we ok with it?

sandbox/plugins/native-repository-fs/src/main/rust/src/fs.rs
```diff
@@ -0,0 +1,53 @@
...
+    let config: FsConfig = serde_json::from_str(config_json)?;
+    let store = LocalFileSystem::new_with_prefix(&config.base_path)?;
+    Ok(Arc::new(store))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
```
mgodwan, line 40: Can we test read/write operations as well for the store implementation?

server/src/main/java/org/opensearch/repositories/NativeRemoteObjectStoreService.java
```diff
@@ -0,0 +1,182 @@
...
+            }
+        });
+        stores.clear();
+    }
+
+    /**
+     * Cached native store pointer paired with the provider that created it,
+     * so {@link #close()} can call the correct {@code destroyNativeStore}.
+     */
+    private static final class StoreEntry {
```
mgodwan: This should be a NativeHandle? https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/jni/NativeHandle.java

sandbox/plugins/native-repository-azure/src/main/java/org/opensearch/repositories/azure/native_store/AzureNativeObjectStorePlugin.java
nishchay21: The plugins are in sandbox/plugins/, so FFM stays in sandbox. The SPI (NativeRemoteObjectStoreProvider) in server/ has no FFM dependency — it's just a Java interface. The NativeRemoteObjectStoreService in server/ also has no FFM — it just holds long pointers. FFM is only in the sandbox plugin implementations.

Am I missing something ?

mgodwan: I missed the sandbox prefix. I thought you had modified the existing repo plugins

server/src/main/java/org/opensearch/repositories/NativeRemoteObjectStoreService.java
nishchay21: Since NativeHandle is in sandbox/libs/analytics-framework/ and the service is in server/, we can't directly depend on it. Happy to add a TODO for migrating to NativeHandle when the service moves to sandbox, or create a lightweight equivalent in server/ if preferred.

sandbox/plugins/native-repository-s3/src/internalClusterTest/java/org/opensearch/repositories/s3/native_store/S3NativeStoreIT.java
```diff
@@ -0,0 +1,96 @@
...
+ * when the native-repository-s3 sandbox plugin is loaded via ExtensiblePlugin.
+ *
+ * <p>Run with:
+ * <pre>
+ * ./gradlew :sandbox:plugins:native-repository-s3:internalClusterTest \
+ *   --tests "*.S3NativeStoreIT" -Dsandbox.enabled=true
+ * </pre>
+ */
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.TEST, numDataNodes = 0)
+public class S3NativeStoreIT extends OpenSearchIntegTestCase {
```
Bukhtawar: numDataNodes=0?

nishchay21: The test calls internalCluster().startNode() explicitly to control when the node starts for now

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
```diff
@@ -0,0 +1,1467 @@
...
+                {
+                    let rp = Path::from(remote_path);
+                    let store: Arc<dyn ObjectStore> = Arc::clone(store);
+                    drop(guard);
+                    return store.get_ranges(&rp, ranges).await;
+                }
+            }
+        }
+
+        self.local.get_ranges(location, ranges).await
```
Bukhtawar, line 261: Avoid redundant logic. Extract as below
```
fn resolve_remote(&self, path: &str) -> Option<(Path, Arc<dyn ObjectStore>)> {
    let guard = self.registry.get(path)?;
    if guard.location() != FileLocation::Remote {
        return None;
    }
    let binding = guard.remote_binding()?;
    let rp = Path::from(binding.remote_path.as_ref());
    let store = Arc::clone(&binding.store);
    drop(guard); // release before I/O
    Some((rp, store))
}
```

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -1371,14 +1372,26 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
             );
 
             final SegmentReplicationStatsTracker segmentReplicationStatsTracker = new SegmentReplicationStatsTracker(indicesService);
+
+            // Only pass the FS native provider to RepositoriesModule — S3/GCS/Azure
+            // wire their own native providers via ExtensiblePlugin, not through here.
+            NativeRemoteObjectStoreProvider fsNativeProvider = null;
+            for (NativeRemoteObjectStoreProvider p : pluginsService.filterPlugins(NativeRemoteObjectStoreProvider.class)) {
+                if ("fs".equals(p.repositoryType())) {
```
mgodwan: [nit] We can use the constant here

plugins/repository-azure/src/main/java/org/opensearch/repositories/azure/AzureRepository.java
```diff
@@ -109,12 +111,26 @@ public static final class Repository {
     private final AzureStorageService storageService;
     private final boolean readonly;
 
+    /** Native (Rust) object store — created during construction if native provider is available. */
+    private volatile NativeStoreRepository nativeStore = NativeStoreRepository.EMPTY;
```
mgodwan, line 115: Do we need a map here? Confirming this currently one native impl gets tied to a specific repo.
Do we need to handle a mode for handling multiple backends based on the index? @Bukhtawar ?

nishchay21: If we have multiple natives [different Engines] than having map is better for this

Bukhtawar: Do you mean native repo per data format? Wondering if the same native repo can handle multiple data formats(with native dependencies)?

server/src/main/java/org/opensearch/repositories/fs/FsRepository.java
```diff
@@ -197,4 +232,15 @@ public List<Setting<?>> getRestrictedSystemRepositorySettings() {
...
+    @Override
+    protected void doClose() {
+        nativeStore.close();
+        super.doClose();
+    }
+
+    @Override
+    public long getNativeStorePtr() {
+        return nativeStore.getPointer();
+    }
```
Bukhtawar, line 247: I would avoid core to have to deal with native pointer

server/src/main/java/org/opensearch/repositories/Repository.java
```diff
@@ -672,4 +672,25 @@ default void reload(RepositoryMetadata repositoryMetadata) {}
...
+     * <p>Only sandbox tiered storage code should call this method. The default
+     * implementation returns {@code -1} (no native store).
+     *
+     * @return native store pointer ({@code > 0}), or {@code -1} if not available
+     * @throws IllegalStateException if the native store has been closed
+     * @opensearch.experimental
+     */
+    default long getNativeStorePtr() {
+        return -1;
+    }
```
Bukhtawar: This should be encapsulated in a `NativeRemoteStoreRepository` and not be exposed outside

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -1371,14 +1373,26 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
...
+            // Only pass the FS native provider to RepositoriesModule — S3/GCS/Azure
+            // wire their own native providers via ExtensiblePlugin, not through here.
+            NativeRemoteObjectStoreProvider fsNativeProvider = null;
+            for (NativeRemoteObjectStoreProvider p : pluginsService.filterPlugins(NativeRemoteObjectStoreProvider.class)) {
+                if (FsRepository.TYPE.equals(p.repositoryType())) {
+                    fsNativeProvider = p;
+                    break;
+                }
+            }
+
```
Bukhtawar: Sorry why can't `FsRepository` implement `ExtensiblePlugin` and `loadExtensions` is it because `FsRepository` is not a plugin? Either case move the logic to RepositoryModule

nishchay21: FsRepository is not a plugin that is why. Previously had this logic on RepositoryModule but moved to node.java so that only FS is passed forward and others are not so that RepositoryModule should not be aware about S3/GCS/Azure as providers. Also plugin service is not available in repositoriesModule. Let me know your thoughts

server/src/main/java/org/opensearch/repositories/fs/FsRepository.java
nishchay21: This is the FSRepository giving the native pointer back like the other plugins . Will return nativeStore instead

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
```diff
@@ -0,0 +1,357 @@
...
+    ) {
+        let remote_arc: Option<Arc<str>> = remote_path.map(Arc::from);
+        let repo_arc: Option<Arc<str>> = repo_key.map(Arc::from);
+
+        self.registry.update(path, move |e| {
+            e.location = location;
+            e.remote_path = remote_arc;
+            e.repo_key = repo_arc;
+            e.remote_store = store;
+        });
```
Bukhtawar, line 141: Extract validation logic from register and ensure we validate remote before updation

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
```diff
@@ -0,0 +1,357 @@
...
+        let path_str = location.as_ref();
+        let entry = TieredFileEntry::new(FileLocation::Local, None, None, None, Some(size));
+        self.registry.register(path_str, entry);
+
+        native_bridge_common::log_debug!(
+            "TieredObjectStore: put_opts registered LOCAL path='{}', size={}",
+            path_str,
+            size
+        );
+        Ok(result)
```
Bukhtawar, line 210: Should we register before writing to local

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    /// Delete: remove from registry only, NO local delete.
+    /// Local file deletion is handled by the Java layer (CompositeDirectory).
+    // TODO: Consider deferred removal (schedule + sweep) instead of force-remove
+    // when eviction lifecycle is added.
+    async fn delete(&self, location: &Path) -> OsResult<()> {
+        let path_str = location.as_ref();
+        self.registry.remove(path_str, true);
+        Ok(())
+    }
```
Bukhtawar, line 292: Should we delete in remote by first checking if its remote only. For correctness do we need to ensure local is deleted?

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
nishchay21: If you register BEFORE writing:
- Write fails (disk full, I/O error) → registry says file is LOCAL → but file doesn't exist
- Concurrent reader sees LOCAL entry → tries to read from local → NotFound Inconsistent state

Register after guarantees:
- If write fails → no registry entry → file doesn't exist (consistent)
- If write succeeds → registry entry created → readers can find it

Let me know your thoughts ?

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
nishchay21: We dont take care of delete from here:
- Remote deletion is handled by Java's RemoteSegmentStoreDirectory garbage collection
- Local deletion is handled by Java's DataFormatAwareTieredDirectory.deleteFile() which calls localDirectory.deleteFile()
- Rust TieredObjectStore is a read-path component — it doesn't own the files

Bukhtawar (approved):
Thanks for the changes. LGTM

plugins/repository-azure/src/main/java/org/opensearch/repositories/azure/AzureRepository.java
Bukhtawar: Will follow up separately on this

PR #21217, opened 2026-04-13
Skip zone awareness when auto-expand replicas set to all

### LINKED ISSUE #14603: [Feature Request] Skip zone/host awareness when auto-expand set to "all"

Today, if index is set to index.auto_expand_replicas with upper bound to "all", the cluster will automatically expand to same # of replicas as # of data nodes. This is will easily conflicts with shard allocation awareness if there's an uneven distribution of nodes across zones or hosts, which cause some unassigned shards(cluster yellow state).  

It will be good to ignore the same hots allocation decider and allocation awareness decider if using auto expand to all feature. As both actually aim to achieve resilience goal of avoiding uneven or concentration of shards on single failure host or zone, with all copies distributed to every nodes, it's already achieve the goal and zone/rack awareness cannot improve the resilience but only create conflict.  

### Related component

ShardManagement:Resiliency

_No response_

### Additional context

_No response_

ZiwenWan:
### Description
When `index.auto_expand_replicas` is set to `0-all`, the `AwarenessAllocationDecider` was blocking shard placement on clusters where all nodes reside in the same zone. This caused shards to remain unassigned (under-replication) even though every available node should be a valid target.

This change makes the decider return `YES` immediately when the index is configured to auto-expand to all nodes, bypassing zone awareness constraints in that specific case.

**Changes:**
- `AutoExpandReplicas`: added `autoExpandToAll()` helper method
- `AwarenessAllocationDecider`: added early-return in `underCapacity()` when auto-expand-to-all is detected
- `AwarenessAllocationTests`: added unit test `testIgnoredByAutoExpandReplicasToAll`
- `AwarenessAllocationIT`: added integration test `testAwarenessZonesWithAutoExpand`

> Implementation ported from #16031, original work by @amberzsy. Thank you!

### Related Issues
Resolves #14603

### Checklist
- [ ] API changes companion PR created (N/A — no REST API changes)
- [ ] Public documentation issue/PR created (N/A)

### DIFF
--- a/server/src/internalClusterTest/java/org/opensearch/cluster/allocation/AwarenessAllocationIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/cluster/allocation/AwarenessAllocationIT.java
@@ -533,4 +533,68 @@ public class AwarenessAllocationIT extends OpenSearchIntegTestCase {
             assertAcked(client().admin().indices().prepareUpdateSettings("test-idx").setSettings(newsettings));
         });
     }
+
+    public void testAwarenessZonesWithAutoExpand() {
+        Settings commonSettings = Settings.builder()
+            .put(AwarenessReplicaBalance.CLUSTER_ROUTING_ALLOCATION_AWARENESS_BALANCE_SETTING.getKey(), true)
+            .put(AwarenessAllocationDecider.CLUSTER_ROUTING_ALLOCATION_AWARENESS_FORCE_GROUP_SETTING.getKey() + "zone.values", "a")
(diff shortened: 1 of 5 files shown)


ZiwenWan:
@Bukhtawar @andrross would you mind taking another look since you stamped the original PR? Thanks!!

msfroh (approved):
Thanks, @ZiwenWan . The logic makes sense.

We'll need to check the integration tests to understand if the failures are related to this change. The latest test run failures *seem* unrelated, but I may be mistaken.

karenyrx (approved):

PR #21228, opened 2026-04-14
Add version 3.6.1 on main

andrross:
### Check List


cwperks:
In what instances do we also update [bwcVersions](https://github.com/opensearch-project/OpenSearch/blob/main/.ci/bwcVersions)?

andrross:
> In what instances do we also update [bwcVersions](https://github.com/opensearch-project/OpenSearch/blob/main/.ci/bwcVersions)?

@cwperks  As far as I can tell that file is unused. We didn't update it for 3.5.1 so I did not update it here. It is used by the `verifyVersions` gradle task but nothing in our CI system invokes that target.

sandeshkr419 (approved):

PR #21230, opened 2026-04-15
Use binary serde for resource usage headers

### LINKED ISSUE #17407: [Feature Request] Optimizing Resource Usage Header Performance for Search Requests

Query-level resource usage tracking was introduced as part of [#12399](https://github.com/opensearch-project/OpenSearch/issues/12399) and implemented in [PR #13172](https://github.com/opensearch-project/OpenSearch/pull/13172). Following discussions with the community, we opted to piggyback shard-level resource usage data in the response header to be aggregated on the coordinator node. We did some benchmarks last year and didn't see big resource usage impact for normal search workloads. However, recently, we have started to see some challenges:
- As mentioned in https://github.com/opensearch-project/OpenSearch/issues/16635, although the cpu usage overhead to do serialization/deserialization wouldn't be noticeable in most normal workloads, profiling with minimal queries has shown an approximate 7% increase in CPU usage due to serialization and deserialization overhead.
- When reindexing huge amounts of shards, there will be a huge number of frequent short-lived scroll requests being sent. In this case, the resource tracking updates accumulate quickly and can consume a non-trivial amount of resources.

We need to think about how to enhance the performance of the resource usage header injection.

I propose we do the following items to improve the performance of resource usage headers:
1. Replace JSON-based headers with a lightweight, delimited string format like: <action>,<taskid>,<parentid>,<nodeid>,<cpu>,<memory usage>. We should also refactor the transport protocol to allow writing binary header values to further speed up.
2. We can also introduce a configurable sampling mechanism to limit resource tracking to a certain percentage of shards if the number of shards exceeds a certain threshold, this will reduce overhead while maintaining representative insights.
3. Only include resource usage in the response header when usage exceeds a configurable threshold.
4. Move the logic to deserialize/parse the resource usage data into an async flow in Query Insights, instead of processing it within the search execution path, to reduce query latency.

### Related component

Search:Query Insights

N/A

### Additional context

https://github.com/opensearch-project/OpenSearch/issues/11522
https://github.com/opensearch-project/OpenSearch/issues/12399

dzane17:
### Description

Replaces JSON serialization/deserialization of `TaskResourceInfo` in the `TASK_RESOURCE_USAGE` response header with Base64-encoded binary using the existing `Writeable` interface. This reduces CPU overhead on the search path where the resource usage header is parsed for every shard.

### Problem

The `TASK_RESOURCE_USAGE` header carries per-shard resource usage data (CPU, memory) from data nodes to the coordinator. The current implementation serializes via `toXContent` (JSON) and deserializes via `XContentParser` (Jackson) on every shard response. Since a single search request can hit thousands of shards, this serde cost is multiplied accordingly. Profiling has shown ~7% CPU overhead from this serde.

### Solution

Serialize `TaskResourceInfo` using `writeTo`/`readFromStream` (binary), Base64-encode the result into a string. Base64 encoding is necessary because `ThreadContext` response headers only support string values. Even if `ThreadContext` supported raw `byte[]` headers, benchmarks show the additional speedup is marginal (see data below), making the Base64 approach a good tradeoff that avoids modifying the public `ThreadContext` class.

### Backwards compatibility

This change is fully BWC across a rolling upgrade.
- Data nodes serialize with binary only when the receiving coordinator is on `V_3_7_0` or later (resolved from the parent task ID via cluster state), otherwise serialize using the original JSON format.
- On the deserialization path, coordinators try binary first, then fall back to JSON so they can read headers from new and old data nodes. This try/catch fallback is acceptable because the majority of the time the domain will not be mid-upgrade.

**Rolling upgrade verification:** Verified end-to-end with a 3-node cluster running released `3.6.1`, upgrading one node at a time to this PR (`3.7.0`). Search workload driven against all three nodes at every phase: 852 searches total, 0 failures, no hangs, no stuck tasks, no parse errors on any node.

### Benchmark Results (JMH, per 1000 shards)

The end-to-end round-trip (serialize on data node + deserialize on coordinator) is **5.9x faster with Binary+Base64 compared to JSON**. Raw binary column is included to show the theoretical outcome if `ThreadContext` supported `byte[]` headers natively.

| Operation | JSON | Binary+Base64 | Raw Binary | JSON→Base64 | JSON→Raw |
|---|---|---|---|---|---|
| Serialize | 227 μs | 95 μs | 81 μs | 2.4x | 2.8x |
| Deserialize | 785 μs | 220 μs | 139 μs | 3.6x | 5.6x |
| Round-trip | 1,647 μs | 279 μs | 252 μs | 5.9x | 6.5x |

Each test result ran:
- 3 warmup iterations
- 3 measurement iterations

One iteration means JMH ran the target operation continuously for 10 seconds, then divided the total time by the number of completed operations to get a time per operation. Each operation performs 1000 TaskResourceInfo serde cycles — one per shard — to simulate a large search request hitting 1000 shards. The final values in the table are the average ns/op across all 3 measurement iterations.

### Header Size Reduction

Based on a typical `TaskResourceInfo`:

```json
{
  "action": "indices:data/read/search[phase/query]",
  "taskId": 1234567,
  "parentTaskId": 1234566,
  "nodeId": "U0rMsZg9RGOxdnVfMtMNVA",
  "taskResourceUsage": {
    "cpuTimeInNanos": 15000000,
    "memoryInBytes": 2048000
  }
}
```

Format | Size
-- | --
JSON | ~196 bytes
Binary+Base64 | ~136 bytes

**Binary+Base64 reduces bytes sent over the wire by ~31% compared to JSON.**

### Related Issues
Resolves #17407

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/tasks/TaskResourceTrackingService.java
+++ b/server/src/main/java/org/opensearch/tasks/TaskResourceTrackingService.java
@@ -17,14 +17,15 @@ import org.opensearch.ExceptionsHelper;
 import org.opensearch.action.search.SearchShardTask;
 import org.opensearch.common.SuppressForbidden;
 import org.opensearch.common.inject.Inject;
+import org.opensearch.common.io.stream.BytesStreamOutput;
 import org.opensearch.common.settings.ClusterSettings;
 import org.opensearch.common.settings.Setting;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.concurrent.ConcurrentCollections;
 import org.opensearch.common.util.concurrent.ConcurrentMapLong;
 import org.opensearch.common.util.concurrent.ThreadContext;
-import org.opensearch.common.xcontent.XContentHelper;
-import org.opensearch.core.common.bytes.BytesArray;
+import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.tasks.resourcetracker.ResourceStats;
 import org.opensearch.core.tasks.resourcetracker.ResourceStatsType;
 import org.opensearch.core.tasks.resourcetracker.ResourceUsageInfo;
@@ -32,16 +33,13 @@ import org.opensearch.core.tasks.resourcetracker.ResourceUsageMetric;
 import org.opensearch.core.tasks.resourcetracker.TaskResourceInfo;
 import org.opensearch.core.tasks.resourcetracker.TaskResourceUsage;
 import org.opensearch.core.tasks.resourcetracker.ThreadResourceInfo;
-import org.opensearch.core.xcontent.DeprecationHandler;
-import org.opensearch.core.xcontent.MediaTypeRegistry;
-import org.opensearch.core.xcontent.NamedXContentRegistry;
-import org.opensearch.core.xcontent.XContentParser;
 import org.opensearch.threadpool.RunnableTaskExecutionListener;
 import org.opensearch.threadpool.ThreadPool;
 
 import java.io.IOException;
 import java.lang.management.ManagementFactory;
 import java.util.ArrayList;
+import java.util.Base64;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -322,7 +320,7 @@ public class TaskResourceTrackingService implements RunnableTaskExecutionListene
                 )
                 .build();
             // Remove the existing TASK_RESOURCE_USAGE header since it would have come from an earlier phase in the same request.
-            threadPool.getThreadContext().updateResponseHeader(TASK_RESOURCE_USAGE, taskResourceInfo.toString());
+            threadPool.getThreadContext().updateResponseHeader(TASK_RESOURCE_USAGE, serializeToBase64(taskResourceInfo));
         } catch (Exception e) {
             logger.debug("Error during writing task resource usage: ", e);
         }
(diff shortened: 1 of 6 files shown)


ansjcy (approved):
LGTM, the numbers are pretty impressive. 2 caveats:

- we need to accept the resource usgae data loss during rolling upgrade
- Please add change log entry to reflect this.

jainankitk (commented):
@kaushalmahi12 - What do you think about this change? Mostly looks good to me.

kkhatua:
> [ansjcy](https://github.com/ansjcy) left a comment
> LGTM, the numbers are pretty impressive. 2 caveats:
> we need to accept the resource usgae data loss during rolling upgrade
> Please add change log entry to reflect this.

@dzane17 
Given this is a breaking change and you have UTs to cover the data loss during a rolling upgrade, were you able to verify this with an actual upgrade ?

I'm of the opinion that we should have a flag just incase something breaks and allow a rollback to the old format of the usage headers.

dzane17:
@kkhatua @ansjcy I decided to add bwc functionality so we will no longer will lose resource tracking info during upgrade and tested a local rolling upgrade. Details are in the description.

Also added a dynamic cluster setting to revert to JSON serialization in an emergency.

kkhatua:
Thanks, @dzane17 

@kaushalmahi12  could you take a look at the PR?

server/src/main/java/org/opensearch/tasks/TaskResourceTrackingService.java
```diff
@@ -321,13 +364,36 @@ public void writeTaskResourceUsage(SearchShardTask task, String nodeId) {
                     )
                 )
                 .build();
+            String headerValue = canCoordinatorReadBinaryHeader(task) ? serializeToBase64(taskResourceInfo) : taskResourceInfo.toString();
```
cwperks: We should probably come up with a generic way of doing this across the board. The Security plugin actually has logic with this optimization as well for internodal transport. See https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java

dzane17: Good point, I extracted the serde methods to `WriteableBase64.java` so other consumers can use it.

server/src/main/java/org/opensearch/tasks/TaskResourceTrackingService.java
```diff
@@ -87,6 +119,14 @@ public boolean isTaskResourceTrackingEnabled() {
         return taskResourceTrackingEnabled;
     }
 
+    public void setBinaryResourceUsageHeaderEnabled(boolean binaryResourceUsageHeaderEnabled) {
+        this.binaryResourceUsageHeaderEnabled = binaryResourceUsageHeaderEnabled;
+    }
+
+    public boolean isBinaryResourceUsageHeaderEnabled() {
```
cwperks, line 126: Where is the usage of this?

cwperks: btw the pattern in the security plugin is to put base64 encoding to a different TC header altogether for node-to-node transport and the convention is to use the TC header by appending `_header`

for example when user is serialized (user is transient) the transient header is ` _opendistro_security_user` and the base64 serialized header is `_opendistro_security_user_header`: https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/transport/SecurityInterceptor.java#L355-L358

I'd love to work towards a convention longer term in the project, but can be done separately from this PR.

cwperks (approved):

server/src/main/java/org/opensearch/tasks/TaskResourceTrackingService.java
dzane17: `isBinaryResourceUsageHeaderEnabled` is used now for consistency.

That is a good point. Ideally we should have a standard across core and plugins. This PR is a start for binary serialization. Header management is still a WIP.

cwperks (approved):

PR #21232, opened 2026-04-15
Share single FormatChecksumStrategy instance per shard between engine and store

ask-kamal-nayan:
### Description                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                       
ParquetDataFormatPlugin.getFormatDescriptors() creates a new PrecomputedChecksumStrategy() on every call. Since the directory and engine call it independently, they end up with different instances. Also DataFormatPlugin.lastDescriptors where getting updated as new Shards are created as these are node level singleton class. Checksums registered by the engine during writes are invisible to the directory during uploads, forcing an O(n) full-file CRC32 fallback instead of O(1) lookup.                                                                                                                             
                                                                                                                                                                                                                                                     
  #### Fix: 
Create strategies once per shard in IndexService.createShard() via DataFormatRegistry.createChecksumStrategies() and pass the same map to both DataFormatAwareStoreDirectory and IndexingEngineConfig.                                        
                                                                                                                                                                                                                                                     
  ### Changes                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                     
  - Add DataFormatRegistry.createChecksumStrategies() — single creation point                                                                                                                                                                        
  - Add checksumStrategies to IndexingEngineConfig record                                                                                                                                                                                            
  - Remove FormatChecksumStrategy param from DataFormatPlugin.indexingEngine()                                                                                                                                                                    
  - DataFormatAwareStoreDirectory accepts pre-built strategies instead of DataFormatRegistry                                                                                                                                                                                                    
                                                                                                                                                                                                                                                     
  ### Testing                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                     
  - All existing unit tests updated and passing                                                                                                                                                                                                      
  - Server, sandbox, and test framework compile clean


server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatRegistry.java
```diff
@@ -148,8 +143,8 @@ public Set<DataFormat> getRegisteredFormats() {
      * @return unmodifiable map of format name to descriptor, or empty map if no pluggable data format is configured
      */
     public Map<String, DataFormatDescriptor> getFormatDescriptors(IndexSettings indexSettings) {
```
mgodwan: I don't see this public method now getting used. Can we remove this?

mgodwan (changes requested):

mgodwan (approved):

PR #21240, opened 2026-04-15
feat: Native Arrow transport path for VectorSchemaRoot in Flight plugin

vamsimanohar:
## Context

While building a query engine POC using Apache DataFusion, I found that query results are already in native Arrow format (`VectorSchemaRoot`) on both the sending and receiving sides. However, the current Flight transport path forces serialization for native arrow format.

```
Worker DataFusion → VectorSchemaRoot → writeTo() → bytes → VarBinary vector → Flight → VarBinary vector → read(StreamInput) → bytes → Coordinator DataFusion
```

The data starts as Arrow, gets serialized to bytes, wrapped in a VarBinary vector, sent over Flight, unwrapped, deserialized back to bytes, and then reconstructed — all redundant when both ends already speak Arrow natively.

This PR adds a zero-serialization path so Arrow data can flow directly:

```
Worker DataFusion → VectorSchemaRoot → Flight → VectorSchemaRoot → Coordinator DataFusion
```

### Server side (sending)
- **`ArrowBatchResponse`** — marker interface for `TransportResponse` instances carrying native Arrow data. When `FlightOutboundHandler` detects this, it sends the `VectorSchemaRoot` directly via `putNext()`, bypassing `VectorStreamOutput` serialization.
- **`FlightServerChannel.sendArrowBatch()`** — sends the caller's `VectorSchemaRoot` directly. Tracks ownership via `externalRoot` flag so `close()` doesn't free externally-owned roots.

### Client side (receiving)
- **`ArrowStreamHandler`** — interface for response handlers that can consume native `VectorSchemaRoot` data via `readArrow()`. Handlers implementing this receive typed Arrow data instead of byte streams.
- **`FlightTransportResponse.resolveArrowStreamHandler()`** — walks the decorator chain (`MetricsTrackingResponseHandler → ContextRestoreResponseHandler → TraceableTransportResponseHandler → original handler`) to find the `ArrowStreamHandler`. Result is cached after first resolution.

### Handler chain walking
- Added `TransportResponseHandler.getDelegate()` default method to enable generic decorator chain introspection. Implemented in `ContextRestoreResponseHandler`, `TraceableTransportResponseHandler`, and `MetricsTrackingResponseHandler`.

### Key design decisions
- **Caller retains ownership** of `VectorSchemaRoot` on the send side — Flight reads from it but does not close it.
- **`FlightStream` reuses its root** on the receive side — `readArrow()` consumers must deep-copy if they need to hold data across `next()` calls, or process inline for zero-copy.
- **Fallback preserved** — non-`ArrowBatchResponse` responses continue to use the existing byte serialization path. Non-`ArrowStreamHandler` handlers continue to use `VectorStreamInput` deserialization.

## Test plan

- [x] `NativeArrowTransportIT.testSingleBatchNativeArrow` — 1 batch, 3 rows, verifies typed columns (VarChar, Int) and actual data values end-to-end
- [x] `NativeArrowTransportIT.testMultipleBatchesNativeArrow` — 3 batches, 2 rows each, verifies multi-batch streaming with data verification
- [x] `ArrowBatchResponseTests` — unit tests for ownership, serialization bypass, and error handling
- [x] All existing flight integration tests pass (`FlightTransportIT`, `ClientSideChaosIT`, `SubAggregationIT`, etc.)
- [x] All existing flight unit tests pass

### DIFF
--- a/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/ArrowStreamHandler.java
+++ b/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/ArrowStreamHandler.java
@@ -1,0 +1,36 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.arrow.flight.transport;
+
+import org.apache.arrow.vector.VectorSchemaRoot;
+import org.opensearch.core.transport.TransportResponse;
+
+/**
+ * Marker interface for {@link org.opensearch.transport.TransportResponseHandler} instances
+ * that can consume native Arrow columnar data. When the Arrow Flight transport detects a handler
+ * implementing this interface, it passes the received {@link VectorSchemaRoot} directly —
+ * bypassing the byte-oriented {@link VectorStreamInput} deserialization path.
+ *
+ * <p>Implementations must also implement {@code read(StreamInput)} as a fallback
+ * for non-Flight transports (e.g., Netty4).
+ *
+ * @opensearch.experimental
+ */
+public interface ArrowStreamHandler<T extends TransportResponse> {
+
+    /**
+     * Reads a response directly from a native Arrow {@link VectorSchemaRoot}.
+     * Called instead of {@code read(StreamInput)} when the Flight transport
+     * receives native Arrow data (i.e., the server sent an {@link ArrowBatchResponse}).
+     *
+     * @param root the Arrow columnar data for this batch
+     * @return the deserialized response
+     */
+    T readArrow(VectorSchemaRoot root);
+}
--- a/plugins/arrow-flight-rpc/docs/native-arrow-transport-design.md
+++ b/plugins/arrow-flight-rpc/docs/native-arrow-transport-design.md
@@ -1,0 +1,42 @@
+# Native Arrow Transport Path — Design
+
+## Problem
+
+The current Arrow Flight transport serializes all responses through `VectorStreamOutput`,
+which stuffs bytes into a `VarBinaryVector`. This is wasteful when the response already
+contains a `VectorSchemaRoot` (e.g., query results from DataFusion). The data gets
+serialized to bytes, packed into a binary vector, sent over Flight (which does its own
+Arrow IPC), then unpacked and deserialized on the other side.
+
+## Solution: Approach A — Dual-Path in FlightOutboundHandler
+
+Add an `instanceof` check in the outbound handler. If the response implements
+`ArrowBatchResponse`, extract the `VectorSchemaRoot` and call `putNext()` directly.
+Otherwise, use the existing byte path unchanged.
+
+### New Interfaces
+
+1. **`ArrowBatchResponse`** — marker interface for responses carrying native Arrow data
+   - `VectorSchemaRoot getArrowRoot()` — returns the native Arrow data
+   - `Schema getArrowSchema()` — returns the Arrow schema
+
+2. **`ArrowStreamHandler<T>`** — marker interface for handlers that can receive native Arrow
+   - `T readArrow(VectorSchemaRoot root)` — reads a batch from native Arrow data
+
+### Server-Side Changes
+
+- `FlightOutboundHandler.processBatchTask()` — instanceof check before `writeTo()`
+- `FlightServerChannel.sendArrowBatch()` — new method that sends VectorSchemaRoot directly
+
+### Client-Side Changes
+
+- `FlightTransportResponse.nextResponse()` — instanceof check on handler before VectorStreamInput
+
+### Fallback
+
+`ArrowBatchResponse` still implements `writeTo(StreamOutput)` for Netty4 transport.
+The fallback path uses `ArrowStreamWriter` to serialize to IPC bytes.
+
+### Scope
+
+Changes are entirely within the `arrow-flight-rpc` plugin. No core OpenSearch changes.
--- a/plugins/arrow-flight-rpc/src/internalClusterTest/java/org/opensearch/arrow/flight/NativeArrowTransportIT.java
+++ b/plugins/arrow-flight-rpc/src/internalClusterTest/java/org/opensearch/arrow/flight/NativeArrowTransportIT.java
@@ -1,0 +1,427 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.arrow.flight;
+
+import org.apache.arrow.memory.BufferAllocator;
+import org.apache.arrow.memory.RootAllocator;
+import org.apache.arrow.vector.FieldVector;
+import org.apache.arrow.vector.IntVector;
+import org.apache.arrow.vector.VarCharVector;
+import org.apache.arrow.vector.VectorSchemaRoot;
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.FieldType;
+import org.apache.arrow.vector.types.pojo.Schema;
+import org.opensearch.action.ActionRequest;
+import org.opensearch.action.ActionRequestValidationException;
+import org.opensearch.action.ActionType;
+import org.opensearch.action.support.ActionFilters;
+import org.opensearch.action.support.TransportAction;
+import org.opensearch.arrow.flight.transport.ArrowBatchResponse;
+import org.opensearch.arrow.flight.transport.ArrowStreamHandler;
+import org.opensearch.arrow.flight.transport.FlightStreamPlugin;
+import org.opensearch.cluster.node.DiscoveryNode;
+import org.opensearch.common.inject.Inject;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.action.ActionResponse;
+import org.opensearch.core.common.io.stream.StreamInput;
+import org.opensearch.core.common.io.stream.StreamOutput;
+import org.opensearch.plugins.ActionPlugin;
+import org.opensearch.plugins.Plugin;
+import org.opensearch.tasks.Task;
+import org.opensearch.test.OpenSearchIntegTestCase;
+import org.opensearch.threadpool.ThreadPool;
+import org.opensearch.transport.StreamTransportResponseHandler;
+import org.opensearch.transport.StreamTransportService;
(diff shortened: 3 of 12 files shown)


vamsimanohar:
### Responses to review suggestions

**1. Allocator leak in test (`handleStreamRequest`)**

This is test-only code. The core issue is that `sendResponseBatch()` is async — it dispatches to an executor that calls `putNext()` later. The server has no callback to know when `putNext()` completes, so it can't safely close the allocator/root immediately after sending. Closing prematurely causes the executor to read freed/zeroed memory (confirmed by test failure when we tried `response.close()` after send).

In a real production consumer (e.g., the lakehouse distributed query engine), DataFusion owns the `VectorSchemaRoot` and manages its own memory lifecycle. The ownership model for production use will be addressed when we integrate the first real consumer. Deferring this for now.

**2. Cache `resolveArrowStreamHandler()` (per-batch decorator chain walk)**

Fixed in `ae33c2b`. The resolved handler is now cached after the first call to `nextResponse()`, avoiding repeated decorator chain traversal on every batch.

**3. Thread safety of `externalRoot`**

The class Javadoc states: "This implementation is not thread safe; consumer must ensure to invoke sendBatch serially and call completeStream() at the end." `sendArrowBatch()` and `close()` are not called concurrently — `close()` runs after `completeStream()` or on cancellation, both of which happen after all batches are sent.

**4. Mixed byte/arrow path (`externalRoot` never reset)**

A single stream is either all native Arrow or all byte path. Mixing `sendArrowBatch()` and `sendBatch()` on the same channel is not a supported use case and would not make semantic sense. The `externalRoot` flag correctly reflects the ownership for the stream's lifetime.

Additional detail on why we're not fixing this:

- **Can't happen in practice** — a `FlightServerChannel` is created per stream, and a stream is either all native Arrow batches or all byte batches. The server handler picks one path based on the response type and uses it consistently. There's no scenario where `sendArrowBatch()` and `sendBatch()` would be mixed on the same channel.
- **Same pattern as existing code** — `sendBatch()` doesn't validate this either. Neither path checks what the previous batch type was. The channel is a fire-once-per-stream construct, not a general-purpose reusable channel.
- **Class Javadoc already establishes the contract** — `FlightServerChannel` Javadoc states: "This implementation is not thread safe; consumer must ensure to invoke sendBatch serially and call completeStream() at the end." The caller controls what goes through the channel and is responsible for using it correctly.
- Adding a guard for mixed paths would add complexity for a scenario that can't arise and would be inconsistent with how the byte path already works.

**5. Schema mismatch in subsequent batches**

This is the caller's responsibility, same as the existing byte path (`sendBatch`) which also doesn't validate schema consistency between batches. Adding validation only to the native Arrow path would be inconsistent.

**6. Document root reuse contract in `ArrowStreamHandler.readArrow()` Javadoc**

The `ArrowStreamHandler` interface Javadoc already documents that the root is borrowed from the Flight stream. Will add an explicit note about reuse in a follow-up if needed.

**7. Gradle check failure**

The failure is `FlightOutboundHandlerTests.testSendResponseBatchPropagatesContextToExecutorThread` — a pre-existing flaky test. Confirmed by running it on the base commit without any of our changes — same failure. Unrelated to this PR.

vamsimanohar:
@rishabhmaurya @reta @finnegancarroll — requesting review.

rishabhmaurya:
thanks @vamsimanohar for working on it. 

> However, the current Flight transport path forces unnecessary serialization:

I won't call it unnecessary as the very first use case we onboarded was streaming aggregations, where we have to do ser/de to/from StreamOutput/Input. Maybe oneday we will start making use of Arrow format within opensearch aggregation directly and call it unnecessary :)

The main question here is - who is doing buffer management? Java or DF? 
I'm of the opinion that Java should be doing all buffer management and pass the allocators to DF where it needs. 

I think the changes can be simplified a lot as it was always designed to support this primary use case. I will add details in my next comment in a while (running the approach with kiro to implement and check).

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/ArrowStreamHandler.java
```diff
@@ -0,0 +1,36 @@
...
+ * that can consume native Arrow columnar data. When the Arrow Flight transport detects a handler
+ * implementing this interface, it passes the received {@link VectorSchemaRoot} directly —
+ * bypassing the byte-oriented {@link VectorStreamInput} deserialization path.
+ *
+ * <p>Implementations must also implement {@code read(StreamInput)} as a fallback
+ * for non-Flight transports (e.g., Netty4).
+ *
+ * @opensearch.experimental
+ */
+public interface ArrowStreamHandler<T extends TransportResponse> {
```
mch2, line 25: nitpick - this is an alternate to Writeable's handler.read(StreamInput) - lets call it something like `ArrowStreamReader` to not confuse with transport response handlers.

mch2:
> The main question here is - who is doing buffer management? Java or DF?

Ideally, the buffer management should be java side with all allocation a child of the same root. Right now [DataFusionService](https://github.com/opensearch-project/OpenSearch/blob/66bc444ef3956331bddc5a34421eb9e063bf508d/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionService.java#L44) is creating its own root allocator java side.

Not an expert here, but I'm assuming your concern @rishabhmaurya is the `externalRoot = true` piece where FlightServerChannel skips closing the root and doesn't claim ownership?  Regardless of whether Java or a backend engine allocated the VSR, should flight close it or make it the caller's responsibility?

vamsimanohar:
> thanks @vamsimanohar for working on it.
> 
> > However, the current Flight transport path forces unnecessary serialization:
> 
> I won't call it unnecessary as the very first use case we onboarded was streaming aggregations, where we have to do ser/de to/from StreamOutput/Input. Maybe oneday we will start making use of Arrow format within opensearch aggregation directly and call it unnecessary :)
> 
My bad. It slipped into description from the context of my POC and the discussion with agent. Anyways edited it. 👍 I know we can't avoid it for existing usecases.

vamsimanohar:
> The main question here is - who is doing buffer management? Java or DF? I'm of the opinion that Java should be doing all buffer management and pass the allocators to DF where it needs.
> 
> I think the changes can be simplified a lot as it was always designed to support this primary use case. I will add details in my next comment in a while (running the approach with kiro to implement and check).

Thanks for the feedback @rishabhmaurya 

I could see few problems with the current implementation particularly on sending side.

**Server-side(Sending Arrow Batch):** 
sendResponseBatch() is async, so the caller can't immediately close the root after calling sendResponseBatch and there is no callback as well. externalRoot=true was added so the channel doesn't close caller-owned roots, but then nobody closes them.

Either we provide a completion callback so the caller can close it after consumption, or FlightServerChannel takes ownership and closes it. I think FlightServerChannel closing it is fine, unless the caller is using the batch for further computation. Depends on the use case. I think once we send a batch from one node to another there won't be further computation. I should maybe revert the externalRoot change. Let me know if your idea is to introduce new interfaces altogether.

 ==============================================================================
  
**Client-side:** 
FlightStream reuses one VectorSchemaRoot across next() calls, so consumers must either process each batch inline before calling next(), or copy it to hold multiple batches. The consumer can use any BufferAllocator for the copy, so memory tracks under whichever pool they choose. The question is whether  to copy or read inline, which depends on the use case. [Not a big concern]

==============================================================================

**Multiple allocator pools:** 
There are three independent memory pools today — Rust GreedyMemoryPool (DataFusion execution), DataFusionService.RootAllocator (Java-side Arrow buffers), and Flight's RootAllocator (transport buffers). Today, DataFusion creates RecordBatches in Rust and transfers them to a child allocator of DataFusionService.RootAllocator via the C Data Interface import(zero copy). But during execution, these pools are blind to each other. There's no unified budget. Do we need coordinated limits...what do you think?
  
Will give it a thought on these problems.

rishabhmaurya:
@vamsimanohar Yes - I like the way you're going about it, however to avoid back n forth, I created some rough changes on how I think this should addressed, please take a look as a reference while you work on it - https://github.com/opensearch-project/OpenSearch/pull/21253 ; it have few tests and passes the checks but there might be issues. 


**Concerns with the current approach (which you're mostly aware of, still listing them here)**

**1. Transport-level branching**

`FlightOutboundHandler.processBatchTask()` and `FlightTransportResponse.nextResponse()` both have `instanceof` checks to branch between byte and native Arrow paths. This couples the transport layer to specific response types. The transport should be agnostic — the response type should control its own serialization behavior.

**2. Core server changes for handler introspection**

The PR adds `getDelegate()` to `TransportResponseHandler` (core interface), `TransportService.ContextRestoreResponseHandler`, and `TraceableTransportResponseHandler`. These changes exist solely to walk the decorator chain and discover if the underlying handler supports native Arrow. This is invasive — a plugin-level feature should not require changes to core transport interfaces.

**3. Buffer ownership confusion**

`FlightServerChannel` introduces an `externalRoot` boolean to track whether the current root is owned by the caller or the channel. This flag is channel-wide but ownership varies per-batch. If a channel ever mixes byte and native Arrow batches (the API allows it), or if `close()` is called at the wrong time, roots can leak or be double-freed.

**4. No pipelining support**

The PR passes the `VectorSchemaRoot` directly from the producer to the transport. Since Flight binds to one root via `start()`, and `putNext()` reads from that root, the producer cannot queue batches ahead. If the producer modifies the root for the next batch before the executor calls `putNext()`, the data is corrupted. This is a race condition for any non-trivial streaming use case.

#### Proposed alternative

**Key tenet**

- **Response controls ser/de**: The response type decides how to serialize/deserialize — not the transport.
- **Java owns buffer management**: The channel's allocator is the parent. Producers create roots from it. The framework manages the shared root bound to Flight.
- **Zero-copy via transfer**: `TransferPair.transfer()` moves buffer pointers from producer vectors to the shared root — no `memcpy`.
- **Pipelining safe**: Each batch has independent buffers. The executor transfers and sends them serially.
- **No core changes**: Everything within the `arrow-flight-rpc` plugin.

**How it works**

**Send side**: API developer extends `ArrowBatchResponse`. Its `writeTo()` is a final no-op. `FlightOutboundHandler` detects this type, creates a shared root on the first batch, and calls `transferTo()` to zero-copy the producer's buffers into the shared root before `putNext()`. All transfer happens on the executor — safe for pipelined production.

**Receive side**: `ArrowBatchResponse(StreamInput in)` calls `((VectorStreamInput) in).getRoot()`. The handler's `read(StreamInput)` constructs the response via this path. `FlightTransportResponse` always creates the same `VectorStreamInput` — no branching, no factory selection. The handler decides whether to call `getRoot()` (native) or byte methods (existing path).

**Allocator access**: `ArrowFlightChannel` interface with `getAllocator()`. `ArrowFlightChannel.from(channel)` unwraps the `TaskTransportChannel`/`BaseTcpTransportChannel` wrapper chain to find the underlying `FlightServerChannel`.

vamsimanohar:
Closing this in favour of https://github.com/opensearch-project/OpenSearch/pull/21253

PR #21242, opened 2026-04-16
Add an event-driven scheduler and stage execution for analytics engine

mch2:
### Description
 Introduces a coordinator-side execution layer that walks a QueryDAG,
  constructs StageExecution instances, and drives them through state
  transitions via listeners.

  Core components:
  - AbstractStageExecution: thread-safe state machine with metrics
  - StageExecutionBuilder: classifies stages and dispatches to schedulers
  - PlanWalker: DAG walker with per-parent listener wiring
  - ShardScanStageExecution: fan-out dispatch to data nodes
  - LocalStageExecution: coordinator-local backend-provided compute
  - QueryScheduler: pool manager + cancellation wiring

  Backend SPI additions (analytics-framework):
  - ExchangeSink, LocalStageContext, LocalStageRequest (for datafusion/native engine reduce at coordinator)

  Stage classification (derived on Stage from the handles it holds):
  - TargetResolver → SHARD_FRAGMENT
  - ExchangeSinkProvider → COORDINATOR_REDUCE
  - otherwise → LOCAL_PASSTHROUGH

  Flow: DefaultPlanExecutor → QueryScheduler → PlanWalker. The walker builds a StageExecution per stage, wires a listener on each child so the parent starts when all children succeed, and starts leaves first. Child
   failures cascade upward via failFromChild and terminate the query.

Note: This doesn't yet contain e2e tests that is dependent on substrait generation, and https://github.com/opensearch-project/OpenSearch/pull/21306/changes

### Related Issues
related - https://github.com/opensearch-project/OpenSearch/issues/19902

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -9,130 +9,163 @@
...
+        EngineContext engineContext,
+        NodeClient client,
+        Scheduler scheduler
+    ) {
+        super(AnalyticsQueryAction.NAME, transportService, actionFilters, in -> {
+            throw new UnsupportedOperationException("Transport path not implemented yet");
+        });
+        this.capabilityRegistry = capabilityRegistry;
         this.clusterService = clusterService;
+        this.searchExecutor = threadPool.executor(ThreadPool.Names.SEARCH);
```
mch2, line 90: changing this to pass threadPool to context so we can decide on pool lower down.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -0,0 +1,155 @@
...
+
+    /**
+     * Collects all batches from the result stream into a single {@link FragmentExecutionResponse}.
+     * Field names are captured from the first batch. Polls the shard task for cancellation
+     * at each batch boundary.
+     *
+     * @param stream the result stream to drain
+     * @param task   the shard task to poll for cancellation (nullable)
+     */
+    FragmentExecutionResponse collectResponse(EngineResultStream stream, @Nullable AnalyticsShardTask task) {
```
Bukhtawar, line 155: I am assuming we will start streaming back batches to the coordinator as they get finalised on the data node once we have streaming transport?

mch2: yup exactly - https://github.com/opensearch-project/OpenSearch/pull/21240 should introduce this capability so we can stream arrow all the way until the final result is converted for front-ends

Bukhtawar (approved):
Overall LGTM , Thanks Marc

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -0,0 +1,178 @@
...
+        long startNanos = System.nanoTime();
+        try (GatedCloseable<Reader> gatedReader = compositeEngine.acquireReader()) {
+            SearchShardTask searchShardTask = null; // TODO: real task for cancellation
+            ExecutionContext ctx = new ExecutionContext(request.getShardId().getIndexName(), searchShardTask, gatedReader.get());
+            ctx.setFragmentBytes(selectedPlan.getFragmentBytes());
+
+            AnalyticsSearchBackendPlugin backend = backends.get(selectedPlan.getBackendId());
+
+            // createSearchExecEngine calls prepare() internally — do NOT call prepare() again
+            try (SearchExecEngine<ExecutionContext, EngineResultStream> engine = backend.getSearchExecEngineProvider().createSearchExecEngine(ctx)) {
```
bharath-techie:
```
   // fork the execution in the search thread pool
                runAsync(getExecutor(executorName, shard), () -> executeDfsPhase(request, task, keepStatesInContext), listener);
```

Do we need an equivalent of this in the data nodes ^ where are we forking the execution to a different threadpool ?

mch2: yes this should be search ThreadPool, will be covered in a separate PR where we are building out the data node search service equivalent.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/StageExecution.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @opensearch.internal
+ */
+public interface StageExecution {
+
+    int getStageId();
+
+    State getState();
+
+    StageMetrics getMetrics();
```
expani, line 34: This is also storing the stageId and it's contained in StageExecution which also stores it above. 

Maybe there is a chance to avoid duplication ?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/ExchangeSource.java
```diff
@@ -0,0 +1,30 @@
...
+package org.opensearch.analytics.backend;
+
+/**
+ * Read-only interface for consuming accumulated results from a stage exchange.
+ * Consumers (parent stages, the walker's completion listener) read from this;
+ * they never write to it.
+ *
+ * @see ExchangeSink for the write-side counterpart
+ */
+public interface ExchangeSource {
```
expani, line 25: I think the sink should have these as there won't be a case where we feed it but don't extract data out of it ?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/ExchangeSource.java
```diff
@@ -0,0 +1,30 @@
...
+
+    /**
+     * Return all accumulated rows in insertion order.
+     * Converts columnar Arrow batches to row-oriented {@code Object[]} arrays.
+     */
+    Iterable<Object[]> readResult();
```
expani: Can be ArrowVSR

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryAction.java
```diff
@@ -0,0 +1,38 @@
...
+ * {@link QueryPlanExecutor#execute(Object, Object)}, not through transport.
+ *
+ * <p>Future: the transport path ({@code doExecute}) will accept query strings
+ * for remote invocation.
+ *
+ * @opensearch.internal
+ */
+public class AnalyticsQueryAction extends ActionType<ActionResponse> {
+
+    public static final String NAME = "indices:data/read/analytics/query";
```
expani, line 32: Can we add a TODO for Security Migration here as this would change the AuthZ model ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/LocalStageScheduler.java
```diff
@@ -0,0 +1,142 @@
...
+                    + "(stageId="
+                    + stage.getStageId()
+                    + ")"
+            );
+        }
+
+        // Select the first plan alternative whose backendId matches a registered backend.
+        StagePlan chosenPlan = null;
+        AnalyticsSearchBackendPlugin backend = null;
+        for (StagePlan plan : stage.getPlanAlternatives()) {
```
expani: Can we add a TODO to change DAG to not produce the case of having alternatives for stages that are not a leaf ? This way selecting an alternative won't be required for Coordinator Reduce. 

We can tackle it later

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/PlanWalker.java
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * Phase 1: Build the execution graph without starting any stages.
+     * All stages are in {@link StageExecution.State#CREATED} state.
+     * Listeners are wired. The graph is inspectable for EXPLAIN.
+     *
+     * @return the fully-wired execution graph
+     */
+    public ExecutionGraph build() {
+        Map<Integer, StageExecution> executions = new ConcurrentHashMap<>();
```
expani: Won't be multi threaded during building so it can use a HashMap ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryAction.java
mch2: I would prefer we add a task on a meta issue instead - TODOs like this will get lost, will make/update one.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/ExchangeSource.java
mch2: This was introduced because callers of each API don't need or care about the other. I could be convinced to merge these, but there's another case to keep them separate that I think makes more sense once the coordinator-reduce API lands.

 A coordinator-reduce stage receives input from N child stages and aggregates them into one output. The interface I'm leaning toward is something like this that would replace `ExchangeSinkProvider` from back-ends.
```java
  interface BackendFragmentExecutor {
      ExchangeSink input(int childStageId);   // per-child write handle
      ExchangeSource output();                // 1 aggregate read handle
      void close();
  }
```

 In that model each per-child sink is purely write-only. The children feed batches into the sink for their stageId, and the backend routes/aggregates internally. There's no meaningful readResult on any individual
child sink.  The read side is the aggregated output of the whole reduce.
Keeping them split lets the same interface cleanly describe both the 1-sink-per-stage case (RowProducingSink implements both) and the N-sinks-per-stage case (per-child sinks are pure write).

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
mch2: Will do this in PR 2.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/PlanWalker.java
mch2: you're right it won't be right now, though we could potentially update stage order in future from multiple threads.  Will change for now thats a much harder feature.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/ExchangeSource.java
mch2: updated this.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/LocalStageScheduler.java
mch2: ack, added in an assert as well - same with the other TODO will spin up a meta issue for these.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryAction.java
expani: Agreed this makes sense to have a dedicated Meta on it's own that discusses different changes required from security perspective.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/ExchangeSource.java
expani: I think you are trying to highlight the case of Join since there are multiple child stages for the same parent. The only case I can think where we would need to pull data at Coordinator for Joins is if the data size is considerably small and a shuffle becomes costlier. 

In such cases, we would need to feed 2 StreamingTables in DataFusion, the Join algo chooses to buffer or incrementally reduce and then we read the final result. So, in my head the CoordinatorReduceFragment ( Join part ) manages the 2 StreamingTables and is a Sink that is fed the input and the readResult is the output post Join.

expani: I think we can start with keeping this separate and naturally see if it makes sense while implementing Joins.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryAction.java
mch2: https://github.com/opensearch-project/OpenSearch/issues/21325

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
```diff
@@ -0,0 +1,98 @@
...
+ * the {@link ExchangeSink} view to child stages and the walker reads
+ * results via the {@link ExchangeSource} view.
+ */
+public class RowProducingSink implements ExchangeSink, ExchangeSource {
+
+    private final List<VectorSchemaRoot> batches = new ArrayList<>();
+    private final List<String> fieldNames = new ArrayList<>();
+
+    @Override
+    public void feed(VectorSchemaRoot batch) {
```
vinaykpud: RowProducingSink.feed() can be called concurrently from multiple SEARCH pool threads when multiple shards respond at the same time right? will this be threadsafe as ArrayList is not thread-safe

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/backend/ExchangeSource.java
mch2: yup exactly for join type cases.  The top level can be modeled as a single sink and internally manage multiple StreamingTables (one per child schema).  But we'd still need to expose input surfaces for each child stream.

I think the shape of this is basically what i have above as `BackendFragmentExecutor`.  Or rename it to something sink related and update the api, collapsing ExchangeSink into a merged interface.

```
  interface ExchangeSink {
      ExchangeSink childInput(int childStageId);  // per-child feed handle
      Iterable<VSR> readResult();                 // combined output
      void close();
  }
```

Working on getting this going in the next day or so and we can look at it.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/PlanWalker.java
```diff
@@ -0,0 +1,250 @@
...
+     */
+    public void cancelAll(String reason) {
+        ExecutionGraph g = this.graph;
+        if (g == null) return;
+        for (StageExecution exec : g.allExecutions()) {
+            StageExecution.State state = exec.getState();
+            if (state == StageExecution.State.RUNNING || state == StageExecution.State.CREATED) {
+                try {
+                    exec.cancel(reason);
+                } catch (Exception ignore) {}
```
vinaykpud: On Exception can we log a warning so that we know the reason.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
```diff
@@ -0,0 +1,98 @@
...
+            int batchRows = batch.getRowCount();
+            if (rowIndex < offset + batchRows) {
+                return toJavaValue(batch.getVector(colIdx), rowIndex - offset);
+            }
+            offset += batchRows;
+        }
+        return null;
+    }
+
+    private static Object toJavaValue(FieldVector vector, int index) {
```
vinaykpud: Can this be a utility? seems like https://github.com/opensearch-project/OpenSearch/pull/21242/changes#diff-9da75654d22fef13732745f1c33e8814b73e9cb7770900f925f4cecc742fac77R206 is identical

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
mch2: you're right, but this is going to be replaced with backend/local sink implementations that better manage concurrency and actually perform a reduce.  Will update this to synchronize for now as it would still be used for pure fan-in cases without reduce.

mch2:
```
WORKFLOW_URL
time passed: 3540
```

gc having issues...

PR #21246, opened 2026-04-16
Make sandbox-check always succeed

andrross:
The sandbox check does not appear to be stable yet. I don't think we want to block PRs on this check failure, but I'd like to avoid normalizing the pattern of merging PRs with failing checks. This change makes sure the sandbox check will always show as green but emit a message if it fails.

### Check List


cwperks:
I believe the change in https://github.com/cwperks/OpenSearch/pull/351 will fix it. I will raise a PR and ask for a review from an expert in the sandbox plugins in development.

andrross:
@mch2 has promised me that sandbox tests are fixed, so closing this PR

PR #21247, opened 2026-04-16
Use pluggable DirectoryFactory in shard store validation (draft)

krishna-ggk:
### Description

`Store.readMetadataSnapshot()` and `Store.tryOpenIndex()` hardcode `new NIOFSDirectory(path)` when reading Lucene segment metadata during shard allocation. This prevents `IndexStorePlugin.DirectoryFactory` implementations (e.g., encryption plugins - [pr-169](https://github.com/opensearch-project/opensearch-storage-encryption/pull/169)) from reading their custom-formatted segment files during the gateway allocation path after a node restart.

**Impact:** After a node restart, the shard allocator calls `TransportNodesListShardStoreMetadataHelper` and `TransportNodesGatewayStartedShardHelper` to validate shard data on disk. These helpers call `Store.readMetadataSnapshot()` / `Store.tryOpenIndex()`, which create a plain `NIOFSDirectory` — bypassing any plugin-provided directory. If the plugin writes `segments_*` and `.si` files in a custom format (e.g., encrypted), the plain directory cannot read them, and the shard is reported as having `no_valid_shard_copy`, preventing recovery.

### Changes

- **`Store.java`**: Added overloaded `readMetadataSnapshot()` and `tryOpenIndex()` that accept `DirectoryFactory` + `IndexSettings`. Added private `openDirectory()` helper that delegates to the factory when provided, falling back to `NIOFSDirectory`.
- **`IndicesService.java`**: Exposed `directoryFactories` via `getDirectoryFactories()` getter.
- **`TransportNodesListShardStoreMetadataHelper.java`**: Resolves `DirectoryFactory` from index settings and passes it to `Store.readMetadataSnapshot()`.
- **`TransportNodesGatewayStartedShardHelper.java`**: Resolves `DirectoryFactory` from index settings and passes it to `Store.tryOpenIndex()`.
- **`StoreTests.java`**: Added 5 unit tests covering custom factory invocation, null-factory fallback, and wrapping factory behavior.

### Testing

- All new unit tests pass
- Verified with [opensearch-storage-encryption](https://github.com/opensearch-project/opensearch-storage-encryption) plugin: `testNodeRestartWithEncryptedIndices` now passes with encrypted `.si`/`segments_*` files (previously failed with `no_valid_shard_copy`)
- Existing `StoreTests` (e.g., `testCanOpenIndex`, `testMetadataSnapshotStreaming`) continue to pass

### Check List


krishna-ggk:
Seeking early feedback if there are any major concerns with the approach - @shwetathareja  @itiyamas @udabhas @RajatGupta02

server/src/main/java/org/opensearch/gateway/TransportNodesGatewayStartedShardHelper.java
```diff
@@ -114,6 +124,47 @@ public static GatewayStartedShard getShardInfoOnLocalNode(
...
+        IndicesService indicesService,
+        Settings settings,
+        ClusterService clusterService
+    ) {
+        IndexService indexService = indicesService.indexService(shardId.getIndex());
+        if (indexService != null) {
+            return indexService.getIndexSettings();
+        }
+        IndexMetadata metadata = clusterService.state().metadata().index(shardId.getIndex());
+        if (metadata != null) {
```
krishna-ggk, line 162: Question: Will this be expensive to be invoked in recover path?

PR #21249, opened 2026-04-16
Fix for RemoteIndexPrimaryRelocationIT Flaky test

divyaruhil:
### Description
The test was failing with IllegalArgumentException: [move_allocation] can't move 0, failed to find it on node due to a race condition. The parent class IndexPrimaryRelocationIT was:

1. Fetching cluster state once at the beginning
2. Tracking the shard location manually with `relocationSource = relocationTarget` after each iteration
3. This caused stale state when the actual shard location differed from the tracked location

**Fix Applied:**

1. Fetch fresh cluster state before each relocation: Added `ClusterState currentState = client().admin().cluster().prepareState().get().getState(); `inside the loop
2. Get current shard location from fresh state : Query the current routing table to find where the primary shard actually is
3. Removed stale state tracking: Deleted `relocationSource = relocationTarget;` since we now fetch the actual location each time

### Related Issues
Resolves #[20872](https://github.com/opensearch-project/OpenSearch/issues/20872)

### Check List


divyaruhil:
Hi @andrross, just a gentle nudge—please review the PR whenever it’s convenient for you. Thankyou !

server/src/internalClusterTest/java/org/opensearch/indices/recovery/IndexPrimaryRelocationIT.java
```diff
@@ -78,12 +79,25 @@ public void testPrimaryRelocationWhileIndexing() throws Exception {
...
+            // Fetch fresh cluster state to get current shard location and available nodes
+            ClusterState currentState = client().admin().cluster().prepareState().get().getState();
+            DiscoveryNode[] dataNodes = currentState.getNodes().getDataNodes().values().toArray(new DiscoveryNode[0]);
+
+            ShardRouting primaryShard = currentState.getRoutingTable().shardRoutingTable("test", 0).primaryShard();
+            if (primaryShard == null || primaryShard.currentNodeId() == null) {
+                logger.warn("--> [iteration {}] primary shard not found or not assigned, retrying", i);
+                i--; // retry this iteration
+                continue;
+            }
```
andrross: Under what scenario can these values be null?

The issue here is that this is effectively an unbounded retry that can go until the test timeout limit is reached. It is also busy-waiting because there is no backoff. But unless there is a valid case for these values to be null, it would be better to just remove the check and let an NPE fail the test if a bug caused these to be null. Same applies for the `relocationSource == null` check below.

divyaruhil: @andrross You're absolutely right. These null checks are unnecessary , that create an unbounded retry loop.

The actual fix for the flakiness was fetching fresh cluster state at the start of each iteration, which ensures we always have the current shard location. After ensureGreen() completes, the primary shard is guaranteed to be assigned, so these values should never be null.

Actually i added that as part of PR Code Suggestions. 

I've removed both null check blocks .

divyaruhil:
Hi @andrross, please let me know if there are any more review comments. Would appreciate it if you could review when you have time.

andrross (approved):

PR #21253, opened 2026-04-17
Native Arrow path support in stream transport with zero-copy transfer

rishabhmaurya:
## Context

Adds a native Arrow transport path to the Flight transport plugin, enabling zero-copy transfer of Arrow data without byte serialization. This is an alternative to #21240 that keeps all changes within the `arrow-flight-rpc` plugin.

## Problem

The existing byte-serialized path (`VectorStreamOutput.ByteSerialized`) serializes Arrow data into a `VarBinaryVector`, which is then sent via Flight `putNext()`. The byte-serialized path is essential for streaming aggregation where results are produced as OpenSearch objects and need serialization into Arrow format. However, for use cases like DataFusion integration where data already originates as native Arrow vectors (potentially from C data import), a direct transfer path avoids the serialization round-trip.

## Solution

Introduce `ArrowBatchResponse` — an abstract base class that API developers extend. When the framework detects this response type, it performs a zero-copy `transferTo()` of the producer's vectors into the channel's shared root, bypassing serialization entirely.

### Key design decisions

**Producer's allocator for the shared root (same-allocator transfer).** The shared root is created from the first batch's producer allocator. This ensures same-allocator transfer, which avoids an Arrow Java bug where `BufferLedger.transferOwnership()` of foreign-backed buffers (from C data import via `wrapForeignAllocation`) doesn't properly free the `ArrowArray` C struct (128 bytes per batch). Same-allocator transfer sidesteps this entirely.

**Long-lived allocator requirement.** The allocator used for producer roots must outlive the gRPC stream. gRPC's zero-copy write path (`ArrowBufRetainingCompositeByteBuf`) retains ArrowBuf references beyond `putNext()` and even beyond `completed()` — they are released asynchronously by gRPC's Netty event loop. Closing the allocator while gRPC still holds these retained references causes memory accounting errors. 

**Producer root closed after transfer.** Each batch's producer root is closed by the framework after `transferTo()` moves its buffers into the shared root. The producer's buffers are empty after transfer, so close is safe and immediate.

### Challenges investigated

**gRPC zero-copy buffer lifecycle.** `putNext()` with `setUseZeroCopy(true)` creates `ArrowBufRetainingCompositeByteBuf` which retains ArrowBufs independently of the shared root. When `transferTo()` replaces the shared root's buffers on the next batch, the old ArrowBufs are released from the root's side but kept alive by gRPC's retain (refcount > 0). The byte-serialized path avoids this because it reuses the same ArrowBuf across batches (overwriting contents via `setSafe()`), so gRPC's retained ByteBufs always point to valid, live memory. For the native arrow path, the allocator must be long-lived so gRPC can release the retained ArrowBufs back to it at any time.

## Upstream issues discovered

During this work we identified two issues in Arrow Java that are worth reporting upstream:

1. **Arrow Java: `BufferLedger.transferOwnership()` leaks `ArrowArray` C struct for foreign-backed buffers.** When `ArrayImporter.importArray()` imports data via the C Data Interface, it allocates a 128-byte `ArrowArray` C struct buffer from the data allocator, wrapped in `ReferenceCountedArrowArray`. During cross-allocator `transferOwnership`, the accounting for the data buffers moves correctly, but the `ReferenceCountedArrowArray` refcount never reaches 0 because the foreign allocation cleanup path isn't triggered. This leaks 128 bytes per imported array per cross-allocator transfer. Same-allocator transfer avoids this because `transferOwnership` returns the same buffer (no new allocation). Present in Arrow Java 18.1.0 and latest main.

2. **Arrow Flight Java: `ServerStreamListener.completed()` is fire-and-forget with no flush guarantee.** `completed()` enqueues HTTP/2 trailers to gRPC's `WriteQueue` but returns immediately without waiting for pending data frames to be flushed. Meanwhile, `ArrowBufRetainingCompositeByteBuf` holds retained references to ArrowBufs that are only released when gRPC's Netty event loop processes the write. There is no callback mechanism (`ServerStreamListener` doesn't expose `setOnCloseHandler`, which exists on `ServerCallStreamObserver` but is not accessible through the Flight API) to know when gRPC has fully released all buffer references. This means allocators backing the stream's data cannot be safely closed immediately after `completed()`. A `setOnCloseHandler` or similar API on `ServerStreamListener` would allow producers to defer cleanup until gRPC is truly done with the buffers.

## Design doc

See `plugins/arrow-flight-rpc/docs/native-arrow-transport-design.md`

### DIFF
--- a/plugins/arrow-flight-rpc/docs/native-arrow-transport-design.md
+++ b/plugins/arrow-flight-rpc/docs/native-arrow-transport-design.md
@@ -1,0 +1,117 @@
+# Native Arrow Transport Path
+
+## Overview
+
+The Arrow Flight transport supports a native Arrow path where typed `VectorSchemaRoot` data
+flows directly over Flight without byte serialization. This is for APIs that produce
+Arrow-columnar data natively (e.g., query engines like DataFusion).
+
+The existing byte-serialized path (`writeTo`/`read` via `StreamOutput`/`StreamInput`) is unchanged.
+
+## Quick Start
+
+### 1. Define your response
+
+Extend `ArrowBatchResponse`. No `writeTo`/`read` override needed — the framework handles it.
+
+```java
+public class MyQueryResponse extends ArrowBatchResponse {
+    public MyQueryResponse(VectorSchemaRoot root) { super(root); }
+    public MyQueryResponse(StreamInput in) throws IOException { super(in); }
+}
+```
+
+### 2. Server-side handler — produce Arrow data
+
+```java
+void handleRequest(MyRequest request, TransportChannel channel, Task task) throws IOException {
+    // Get the channel's allocator. Use this directly for producer roots
+    // (not a child allocator) to avoid Arrow's cross-allocator transfer bug
+    // with foreign-backed buffers from C data import.
+    BufferAllocator allocator = ArrowFlightChannel.from(channel).getAllocator();
+
+    Schema schema = new Schema(List.of(
+        new Field("name", FieldType.nullable(new ArrowType.Utf8()), null),
+        new Field("score", FieldType.nullable(new ArrowType.FloatingPoint(FloatingPointPrecision.DOUBLE)), null)
+    ));
+
+    try {
+        for (int i = 0; i < batchCount; i++) {
+            VectorSchemaRoot root = VectorSchemaRoot.create(schema, allocator);
+            // populate vectors...
+            channel.sendResponseBatch(new MyQueryResponse(root));
+            // root is now owned by the framework — don't reuse or close it
+        }
+        // Cleanup callback runs on executor after all batches are flushed.
+        channel.completeStream();
+    } catch (Exception e) {
+        channel.sendResponse(e);
+    }
+}
+```
+
+### 3. Client-side handler — consume Arrow data
+
+```java
+class MyQueryHandler implements StreamTransportResponseHandler<MyQueryResponse> {
+
+    public MyQueryResponse read(StreamInput in) throws IOException {
+        return new MyQueryResponse(in);
+    }
+
+    public void handleStreamResponse(StreamTransportResponse<MyQueryResponse> stream) {
+        MyQueryResponse response;
+        while ((response = stream.nextResponse()) != null) {
+            VectorSchemaRoot root = response.getRoot();
+            VarCharVector names = (VarCharVector) root.getVector("name");
+            Float8Vector scores = (Float8Vector) root.getVector("score");
+            // process typed vectors...
+        }
+        stream.close();
+    }
+
+    public void handleException(TransportException exp) { /* handle error */ }
+    public String executor() { return ThreadPool.Names.GENERIC; }
+}
+```
+
+## Allocator Management
+
+The allocator used for producer roots must be **long-lived** — it must outlive the gRPC
+stream. This is because gRPC's zero-copy write path (`ArrowBufRetainingCompositeByteBuf`)
+retains ArrowBuf references beyond `putNext()` and `completed()`, releasing them
+asynchronously on the Netty event loop. Closing the allocator while gRPC still holds
+these references causes memory accounting errors.
+
+Use `ArrowFlightChannel.from(channel).getAllocator()` to get the channel's allocator,
+or use your own long-lived application allocator. Do not create and close a child
+allocator per request.
+
+### Important: C Data Import and allocator choice
+
+If your producer imports data via Arrow's C Data Interface (`Data.importIntoVector`,
+`Data.importIntoVectorSchemaRoot`), the imported buffers are foreign-backed. Arrow Java
+has a bug where cross-allocator `transferOwnership` of foreign-backed buffers doesn't
+properly release the internal `ArrowArray` C struct buffer (128 bytes per import call),
+causing a memory leak in the source allocator.
+
+The framework creates the shared Flight root from the **producer's allocator** (the
+allocator of the first batch's vectors), ensuring same-allocator transfer which avoids
+this bug. All subsequent batches should use the same allocator.
+
+## Ownership Contract
+
+| Resource | Created by | Closed by |
+|----------|-----------|-----------|
+| Channel allocator | Framework | Framework (on channel close) |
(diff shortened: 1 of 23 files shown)


vamsimanohar (commented):
Hey @rishabhmaurya , I think this PR is almost there. There is no pointing copying the same to my PR.
should I continue? 
Feel free to close my PR unless you want me to look into any things that you feel are missing and I could work on.

rishabhmaurya:
Sounds good @vamsimanohar. Thanks for sharing your ideas on other PR on which we are able to build up.

plugins/examples/stream-transport-example/src/internalClusterTest/java/org/opensearch/example/stream/NativeArrowStreamTransportExampleIT.java
```diff
@@ -0,0 +1,180 @@
...
+        super.setUp();
+        internalCluster().ensureAtLeastNumDataNodes(2);
+    }
+
+    @Override
+    protected Collection<Class<? extends Plugin>> nodePlugins() {
+        return List.of(StreamTransportExamplePlugin.class, FlightStreamPlugin.class);
+    }
+
+    @AwaitsFix(bugUrl = "")
```
bowenlan-amzn, line 51: Seems a miss to remove, same for the other test

plugins/arrow-flight-rpc/docs/native-arrow-transport-design.md
```diff
@@ -0,0 +1,133 @@
...
+causing a memory leak in the source allocator.
+
+**To avoid this**, create producer roots from the **channel allocator directly** (not a
+child allocator). The framework creates the shared Flight root from the same allocator
+as the producer, ensuring same-allocator transfer which avoids the bug.
+
+If you need per-request memory isolation, create a child allocator for tracking but
+use the channel allocator for the actual Arrow data buffers.
+
+### Using `completeStream(Closeable)` for cleanup
```
bowenlan-amzn: This part seems not in the source code?

bowenlan-amzn (commented):
Overall looks good!

plugins/examples/stream-transport-example/src/internalClusterTest/java/org/opensearch/example/stream/NativeArrowStreamTransportExampleIT.java
rishabhmaurya: this is actually intentional. These test don't add much value as these are just examples but do help while development.

plugins/arrow-flight-rpc/docs/native-arrow-transport-design.md
rishabhmaurya: hmm! will remove it

plugins/examples/stream-transport-example/src/internalClusterTest/java/org/opensearch/example/stream/NativeArrowStreamTransportExampleIT.java
bowenlan-amzn: Make sense. 
Seems we have quite a few examples plugins :) All the tests within that should be disabled automatically.

bowenlan-amzn (approved):

plugins/examples/stream-transport-example/src/internalClusterTest/java/org/opensearch/example/stream/NativeArrowStreamTransportExampleIT.java
```diff
@@ -0,0 +1,180 @@
...
+    @LockFeatureFlag(STREAM_TRANSPORT)
```
mch2, line 52: do we still need the fflag for stream transport?

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
```diff
@@ -151,8 +152,28 @@ private void processBatchTask(BatchTask task) {
...
+                    // This avoids an Arrow bug where cross-allocator transferOwnership of foreign-backed
+                    // buffers (from C data import) doesn't properly free the ArrowArray C struct.
+                    // The producer's allocator must be long-lived (not closed per-request).
+                    sharedRoot = VectorSchemaRoot.create(
+                        arrowResponse.getRoot().getSchema(),
+                        arrowResponse.getRoot().getFieldVectors().get(0).getAllocator()
+                    );
+                }
+                arrowResponse.transferTo(sharedRoot);
+                arrowResponse.getRoot().close();  // release producer's buffers — safe, they've been moved
```
mch2, line 170: Thinking maybe we should document that the producer:
1. must create a new vsr per batch, given this transfer is processed in an async executor
2. doesn't need to close its root with each batch / post iteration

mch2 (approved):
thanks @rishabhmaurya for introducing this capability.  We'll be integrating it into analytics-engine's scheduler within next few days to test it out :)

plugins/examples/stream-transport-example/src/internalClusterTest/java/org/opensearch/example/stream/NativeArrowStreamTransportExampleIT.java
rishabhmaurya: yes - its an optional plugin and behind this feature flag.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
rishabhmaurya: that's a good point. I think I asked agent to document this behaviour as part of API itself, but let me double check.

rishabhmaurya: so in native-arrow request handling logic, `ArrowBatchResponse` will be the response type and its documentation covers it. I hope it addresses it.

PR #21256, opened 2026-04-17
Reject DSL search on indices with pluggable dataformat enabled

KhishorekumarBS:
### Description
Modifies `SearchActionFilter` to check `index.pluggable.dataformat.enabled` on target indices. Indices not using pluggable dataformat should go through the normal Lucene search path while DSL search on indices using pluggable dataformat should be rejected.

Notes on `pluggable.dataformat.enabled`:
- `pluggable.dataformat.enabled` is immutable and cannot be changed once set
- If index level `pluggable.dataformat.enabled` is not set, cluster level pluggable.dataformat.enabled will be added on the index by default

Reference PR - https://github.com/opensearch-project/OpenSearch/pull/21435

**Behavior:**                                                                                                  
  - No target indices have the setting `pluggable.dataformat.enabled = true` →  Use normal search path
  - If any index have `pluggable.dataformat.enabled = true` → Reject request

### Changes
  - `SearchActionFilter`: Added `ClusterService` and `IndexNameExpressionResolver` to resolve concrete indices   
  and check their settings before deciding the execution path. 
  - `DslQueryExecutorPlugin`: Pass new dependencies to `SearchActionFilter`. 
  - `SearchActionFilterTests`: Added tests for all-pluggable bypass, mixed-index rejection, and refactored       
  existing tests to use shared mock helpers.  

### Check List
- [Done] Functionality includes testing.
- [NA] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [NA] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

### DIFF
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
@@ -8,36 +8,41 @@
 
 package org.opensearch.dsl.action;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.opensearch.action.ActionRequest;
 import org.opensearch.action.search.SearchAction;
 import org.opensearch.action.search.SearchRequest;
-import org.opensearch.action.search.SearchResponse;
 import org.opensearch.action.support.ActionFilter;
 import org.opensearch.action.support.ActionFilterChain;
 import org.opensearch.action.support.ActionRequestMetadata;
+import org.opensearch.cluster.ClusterState;
+import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
+import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.action.ActionResponse;
+import org.opensearch.core.index.Index;
+import org.opensearch.index.IndexSettings;
 import org.opensearch.tasks.Task;
-import org.opensearch.transport.client.node.NodeClient;
 
 /**
- * Intercepts all {@code _search} requests and dispatches them to {@link DslExecuteAction}
- * for execution through the Calcite pipeline. Non-search actions pass through unchanged.
+ * Intercepts {@code _search} requests targeting indices with {@code index.pluggable.dataformat.enabled}
+ * and rejects them with an error directing users to PPL instead.
  */
 public class SearchActionFilter implements ActionFilter {
+
+    private static final Logger logger = LogManager.getLogger(SearchActionFilter.class);
 
     /** Runs after the Security plugin's authorization filter (order 0). */
     static final int FILTER_ORDER = 1;
 
-    private final NodeClient client;
+    private final ClusterService clusterService;
+    private final IndexNameExpressionResolver indexNameExpressionResolver;
 
-    /**
-     * Creates a filter that dispatches intercepted searches via the given client.
-     *
-     * @param client node client for dispatching to {@link DslExecuteAction}
-     */
-    public SearchActionFilter(NodeClient client) {
-        this.client = client;
+    public SearchActionFilter(ClusterService clusterService, IndexNameExpressionResolver indexNameExpressionResolver) {
+        this.clusterService = clusterService;
+        this.indexNameExpressionResolver = indexNameExpressionResolver;
     }
 
     @Override
@@ -46,7 +51,6 @@ public class SearchActionFilter implements ActionFilter {
     }
 
     @Override
-    @SuppressWarnings("unchecked")
     public <Request extends ActionRequest, Response extends ActionResponse> void apply(
         Task task,
         String action,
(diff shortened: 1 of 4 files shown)


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
```diff
@@ -59,9 +71,44 @@ public <Request extends ActionRequest, Response extends ActionResponse> void app
...
+            // Non-search actions pass through unchanged.
             chain.proceed(task, action, request, listener);
         }
     }
+
+    private int countPluggableDataFormatIndices(Index[] concreteIndices) {
+        int count = 0;
+        for (Index index : concreteIndices) {
+            IndexMetadata indexMetadata = clusterService.state().metadata().index(index);
+            if (indexMetadata != null && Mapper.isPluggableDataFormatEnabled(indexMetadata.getSettings())) {
```
mgodwan: Can we use the `isPluggableDataFormatEnabled` in IndexSettings instead?

mgodwan: or move to some other util rather than relying Mapper

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
```diff
@@ -59,9 +71,44 @@ public <Request extends ActionRequest, Response extends ActionResponse> void app
...
+                return;
+            }
+
+            if (pluggableCount == concreteIndices.length) {
+                // All indices use pluggable dataformat — bypass DSL plugin, use normal Lucene search path.
+                chain.proceed(task, action, request, listener);
+            } else {
+                // No pluggable dataformat indices — route through DSL/Calcite analytics engine.
+                client.execute(DslExecuteAction.INSTANCE, searchRequest, (ActionListener<SearchResponse>) listener);
+            }
```
nssuresh2007: Shouldn't the if-else branching be in the opposite way? i.e. if `pluggableCount == concreteIndices.length` we need to route it to DSL/Calcite layer?

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
```diff
@@ -59,9 +71,44 @@ public <Request extends ActionRequest, Response extends ActionResponse> void app
...
             SearchRequest searchRequest = (SearchRequest) request;
-            client.execute(DslExecuteAction.INSTANCE, searchRequest, (ActionListener<SearchResponse>) listener);
+
+            // Resolve index expressions (wildcards, aliases) to concrete indices and check
+            // how many have pluggable dataformat enabled via index.pluggable.dataformat.enabled.
+            Index[] concreteIndices = indexNameExpressionResolver.concreteIndices(clusterService.state(), searchRequest);
+            int pluggableCount = countPluggableDataFormatIndices(concreteIndices);
+
+            // Mixed indices (some pluggable, some not) are unsupported — reject the request.
+            if (pluggableCount > 0 && pluggableCount < concreteIndices.length) {
```
nssuresh2007: nit: Would it be better to have a log line here for debugging?

KhishorekumarBS (commented):
Addressed review comments

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
KhishorekumarBS: Addressed

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
KhishorekumarBS: Right. Flipped it

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
KhishorekumarBS: Added

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
```diff
@@ -55,13 +59,33 @@ public <Request extends ActionRequest, Response extends ActionResponse> void app
...
+            if (hasPluggableDataFormatIndex(state, concreteIndices)) {
+                logger.debug("Rejecting DSL search: at least one target index has pluggable dataformat enabled");
+                // TODO: Route to DSL/Calcite engine once supported:
+                // client.execute(DslExecuteAction.INSTANCE, searchRequest, (ActionListener<SearchResponse>) listener);
+                listener.onFailure(
+                    new UnsupportedOperationException("DSL is not supported currently. Please use PPL for search.")
+                );
+                return;
+            }
+        }
```
nssuresh2007, line 78: 1. Can't we comment this entire block and just directly go to chain.proceed()? Since we do not unnecessarily need to evaluate the index settings.
2. We can comment/remove all the other changes and just keep it minimal

KhishorekumarBS: If we let search requests on parquet indices to go through chain.proceed() , it'll use the lucene path and might encounter some errors or inconsistent behavior

nssuresh2007 (approved):
LGTM

sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java
```diff
@@ -55,13 +59,33 @@ public <Request extends ActionRequest, Response extends ActionResponse> void app
...
+        chain.proceed(task, action, request, listener);
+    }
+
+    private boolean hasPluggableDataFormatIndex(ClusterState state, Index[] concreteIndices) {
```
nssuresh2007, line 82: nit: We can move this to a Util class?

Bukhtawar (commented):
I feel we should reject early based on `cluster.restrict.pluggable.dataformat` Once we support mixed mode we might change the planner logic to reject it.

PR #21257, opened 2026-04-17
Added support for regex, where it will be converted to custom UDF and…

abhishek00159:
### Add regex query support via UnresolvedQueryCall

This PR adds support for RegexpQueryBuilder in the DSL query executor by wrapping it in an UnresolvedQueryCall that preserves Lucene regex semantics.

Changes:
• **New translator**: RegexpQueryTranslator converts RegexpQueryBuilder to UnresolvedQueryCall instead of standard SQL SIMILAR TO
• **Registry update**: Registered the new translator in QueryRegistryFactory
• **Tests**: Added unit tests covering basic regex patterns and flags
• **Integration test**: Added testRegexpQueryWithUnresolvedNode to verify end-to-end functionality

Rationale:
Direct translation to SQL regex is lossy because Lucene regex operates on terms in the inverted index while SQL regex operates on full field values. Using UnresolvedQueryCall 
preserves the correct Lucene semantics and routes execution to Lucene.

### Check List
- [ YES ] Functionality includes testing.
- [ YES ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ NA ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

### DIFF
--- a/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/query/RegexpQueryTranslator.java
+++ b/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/query/RegexpQueryTranslator.java
@@ -1,0 +1,34 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.dsl.query;
+
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.opensearch.dsl.converter.ConversionContext;
+import org.opensearch.dsl.converter.ConversionException;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.RegexpQueryBuilder;
+
+/**
+ * Converts {@link RegexpQueryBuilder} to {@link UnresolvedQueryCall}.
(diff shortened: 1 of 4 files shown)


sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/query/RegexpQueryTranslator.java
```diff
@@ -0,0 +1,34 @@
...
+import org.opensearch.dsl.converter.ConversionException;
+import org.opensearch.index.query.QueryBuilder;
+import org.opensearch.index.query.RegexpQueryBuilder;
+
+/**
+ * Converts {@link RegexpQueryBuilder} to {@link UnresolvedQueryCall}.
+ * Translation to SQL SIMILAR TO is lossy — Lucene regex operates on terms in inverted index,
+ * SQL regex operates on full field values. UnresolvedQueryCall preserves Lucene semantics.
+ */
+public class RegexpQueryTranslator implements QueryTranslator {
```
tanik98, line 23: Why do we need this translator, if we are converting it to UnresolvedQueryCall? As anyway its getting converted to UnresolvedQueryCall by defualt, if it doesn't find registered query shape.
https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/query/QueryRegistry.java#L56

PR #21270, opened 2026-04-17
Fix worker gone test failure in sandbox

cwperks:
### Description

Reliably reproduce the test failure with `./gradlew :sandbox:plugins:analytics-backend-datafusion:test -Dtests.seed=218506E88D711883 -Dtests.jvms=1 -i`

  Fixes a flaky test failure in `DatafusionSearchExecEngineTests` where `testEngineExecuteAggregation` and `testEngineExecuteSelectAll` fail with `RuntimeException: Execution error: Worker gone`.

  **Root cause:** `DataFusionJniBridgeTests` called `NativeBridge.shutdownTokioRuntimeManager()` at the end of both `testRuntimeLifecycle()` and `testReaderLifecycle()`. This permanently shuts down the `DedicatedExecutor`'s CPU thread pool inside the Rust `RuntimeManager`. Because the native side stores the runtime manager in a `OnceLock`, subsequent calls to `initTokioRuntimeManager()` from other test classes are no-ops — the shut-down manager stays in place. When `DatafusionSearchExecEngineTests` or `DataFusionQueryExecutionTests` run later in the same JVM (Gradle does not set `forkEvery`, so all test classes share one JVM), `CrossRtStream` tries to spawn work on the dead executor and gets `JobError::WorkerGone`.

  The failure is order-dependent: it only occurs when `DataFusionJniBridgeTests` runs before the other test classes. Since OpenSearch's randomized test runner controls class ordering via seed, the failure is flaky — it depends on the seed drawn for that CI run.

  **Fix:** Remove the `shutdownTokioRuntimeManager()` calls from `DataFusionJniBridgeTests`. Per-test resources (`DataFusionRuntime` via `createGlobalRuntime`, readers) are still properly cleaned up. Only the process-level `RuntimeManager` is left alive for the duration of the JVM, which is the correct behavior for a shared singleton.

### Check List


andrross (approved):

andrross:
@mch2 Can you look at this?

mch2 (approved):
Thanks @cwperks for the fix!

PR #18921, opened 2025-08-05
Add build-tooling to run in the FIPS environment

beanuwave:
### Description
Provides additional build tooling to support builds in FIPS env, including a CLI trust-store installer to override `$JAVA_HOME/lib/security/cacerts`

- add demo/test CLI configurator with the ability to:
-- migrate JVM's default SSL trust store to a BCFKS-formatted one
-- use an existing PKCS#11 trust store
-- display installed 'KeyStore' providers
-- show help
-- execute above commands interactively or in script mode
- add BC libs to standalone REST tests.
- print out 'java.security.properties' for reproducibility information

### Related Issues
Resolves [RFC](https://github.com/opensearch-project/security/issues/4254)

### Check List
- [x] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #18491, opened 2025-06-11
remaining changes from 14912

beanuwave:
### Description
remaining changes from original [PR#14912](https://github.com/opensearch-project/OpenSearch/pull/14912)

### Related Issues
Resolves #[4254]

### Check List
- [x] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #18376, opened 2025-05-27
Fips smoke test

beanuwave:
### Description
Create project to run smoke tests to verify server can handle some basic functionality under FIPS JVM.

### Related Issues
Resolves [RFC](https://github.com/opensearch-project/security/issues/4254)

### Check List
- [x] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #17907, opened 2025-04-11
Add FIPS build tooling

beanuwave:
### Description
Provides the additional build tooling to test & build OS in FIPS JVM with `-Pcrypto.standard=FIPS-140-3`or the (deprecated) `-Ptests.fips.enabled=true `parameter. Code changes required to run successfully in FIPS mode are not included in this commit, they will be addressed in separate upcoming PRs.

Note:
The codecov report may appear inaccurate. Tests for the new `Randomness` logic exist but are only executed when the build is run with the new FIPS parameters enabled.

### Related Issues
[RFC](https://github.com/opensearch-project/security/issues/4254)

### Check List
- [ ] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21278, opened 2026-04-18
Add nightly workflow check to run tests and integration test with FIPS enforcement (-Pcrypto.standard=FIPS-140-3)

cwperks:
### Description

This PR adds a new workflow to run nightly to run unit tests and integration tests with FIPS enforcement.

The core repo has many tests that are skipped with the absence of `-Pcrypto.standard=FIPS-140-3` which is not provided when we run the core gradle check on every PR. Since FIPS is opt-in for cluster operators that require running clusters in FIPS compliant stacks (hardware, OS and applications all with FIPS compliance) this should not be the default for the core gradle check with every PR. Regardless, we still want to ensure these tests are run on a cadence to ensure coverage so this PR proposes running them nightly with a Github workflow on a github runner.

This will ensure that tests matching `**/*FipsTests.java` and `**/*FipsIT.java` have coverage.

**Note**: This only runs FIPS-specific tests atm and not all integ tests. Another change would be required to verify if Github runner would be capable of running all integ tests.

### Related Issues

Related to series of FIPS PRs in core:

- https://github.com/opensearch-project/OpenSearch/pull/17393
- https://github.com/opensearch-project/OpenSearch/pull/17507
- https://github.com/opensearch-project/OpenSearch/pull/17907
- https://github.com/opensearch-project/OpenSearch/pull/18376
- https://github.com/opensearch-project/OpenSearch/pull/18491
- https://github.com/opensearch-project/OpenSearch/pull/18921

### Check List


cwperks:
@beanuwave can you let me know what you think of this when you have a moment?

gradle/fips.gradle
beanuwave, line 161: By running the new task locally, I've discovered 3 issues:
The flag was being read via `System.getProperty()`, which reads JVM client properties that aren't forwarded to the Gradle daemon. Solution would be to use a gradle flag here: `-Ptests.fips.only=true`. 
Also all the tests still ran because `include()` is additive - `OpenSearchTestBasePlugin` adds `**/*Tests.class` to every test task, and calling include `**/*FipsTests.class` on top simply adds to that set rather than replacing it. Finally, with the narrowed include set, disabled`testingConventions` ensures that gradle modules won't fail because of empty test suites.

```suggestion
    if (BuildParams.isInFipsJvm() && project.findProperty('tests.fips.only') == 'true') {
      project.afterEvaluate {
        project.tasks.withType(Test).configureEach { testTask ->
          testTask.setIncludes(['**/*FipsTests.class', '**/*FipsIT.class'])
          testTask.filter.failOnNoMatchingTests = false
        }
        project.tasks.matching { it.name == 'testingConventions' }.configureEach { it.enabled = false }
      }
    }
```

beanuwave (commented):
@cwperks 
Thanks, this looks like a solid foundation! I’d like to highlight a couple of pitfalls that could also be addressed here:

- `SecureReactorNetty4HttpServerTransport*` tests may trigger `QuicClosedChannelException` due to randomized HTTP protocol selection. This can be fixed by overriding `createBuilderWithPort()` function.
- `RestClientBuilderIntegTests` currently uses the `RestClientFipsAwareTestCase` interface to switch between the FIPS context on/off and will not be included in the new nightly workflow.

Both issues can be addressed by following the existing *FipsTests.java & *FipsIT.java patterns.

PR #21279, opened 2026-04-19
Add PIT + search_after as alternative to scroll for local reindex

ankitkala:
## Description

Adds support for using Point in Time (PIT) + `search_after` instead of the traditional scroll API for `_reindex` (local only).

### Motivation

The scroll API holds search context on data nodes, which blocks segment merges and doesn't support resumability. PIT provides a lighter-weight alternative:
- Doesn't block segment merges
- Supports resumability (save last sort values + PIT ID)
- Uses stateless pagination via `search_after`
- Takes a consistent snapshot at PIT creation time

### Usage

```
POST /_reindex?use_pit=true
```

### Changes

- `ClientPitHitSource` — new hit source that creates a PIT, paginates with `search_after` + `_shard_doc` sort, and cleans up the PIT on completion
- `AbstractBulkByScrollRequest` — added `usePit` flag (transient, not serialized to avoid wire format changes)
- `AbstractAsyncBulkByScrollAction` — wires PIT source when `usePit=true`
- `RestReindexAction` — parses `use_pit` query parameter
- `ReindexRequestBuilder` — builder support for `usePit`

### Tests

- `ReindexWithPitIT` — 10 integration test scenarios covering basic reindex, query filters, pagination, snapshot consistency, maxDocs, multi-index, version conflicts, and PIT vs scroll equivalence
- `ClientPitHitSourceTests` — unit tests for the PIT hit source

### Limitations

- PIT + slicing (`slices > 1`) is not supported yet
- Remote reindex stays scroll-based
- `usePit` flag is transient (not serialized) to avoid wire format changes

Signed-off-by: ankikala <ankikala@amazon.com>

### DIFF
--- a/server/src/main/java/org/opensearch/index/reindex/ClientPitHitSource.java
+++ b/server/src/main/java/org/opensearch/index/reindex/ClientPitHitSource.java
@@ -1,0 +1,298 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.reindex;
+
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ParameterizedMessage;
+import org.opensearch.ExceptionsHelper;
+import org.opensearch.action.bulk.BackoffPolicy;
+import org.opensearch.action.search.CreatePitAction;
+import org.opensearch.action.search.CreatePitRequest;
+import org.opensearch.action.search.CreatePitResponse;
+import org.opensearch.action.search.DeletePitAction;
+import org.opensearch.action.search.DeletePitRequest;
+import org.opensearch.action.search.DeletePitResponse;
+import org.opensearch.action.search.SearchRequest;
+import org.opensearch.action.search.SearchResponse;
+import org.opensearch.action.search.ShardSearchFailure;
+import org.opensearch.common.document.DocumentField;
+import org.opensearch.common.unit.TimeValue;
+import org.opensearch.core.action.ActionListener;
+import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.core.concurrency.OpenSearchRejectedExecutionException;
+import org.opensearch.core.xcontent.MediaType;
+import org.opensearch.core.xcontent.MediaTypeRegistry;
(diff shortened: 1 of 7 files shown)


ankitkala:
also needs performance benchmarks

server/src/main/java/org/opensearch/index/reindex/ClientPitHitSource.java
```diff
@@ -0,0 +1,298 @@
...
+import java.util.function.Consumer;
+
+import static java.util.Collections.emptyList;
+import static java.util.Collections.unmodifiableList;
+import static org.opensearch.core.common.util.CollectionUtils.isEmpty;
+
+/**
+ * A scrollable source of hits using Point in Time (PIT) + search_after instead of scroll API.
+ * <p>
+ * PIT provides a lightweight, consistent view of the data at a point in time without holding
```
alchemist51, line 50: PIT also has a context right? Also the context which is create for a PIT is stored inmemory so I don't think it works once a node goes down(where the shard for the PIT is held)

server/src/main/java/org/opensearch/index/reindex/ClientPitHitSource.java
```diff
@@ -0,0 +1,298 @@
...
+        this.firstSearchRequest = firstSearchRequest;
+        this.keepAlive = firstSearchRequest.scroll() != null
+            ? firstSearchRequest.scroll().keepAlive()
+            : TimeValue.timeValueMinutes(5);
+        firstSearchRequest.allowPartialSearchResults(false);
+    }
+
+    @Override
+    public void doStart(RejectAwareActionListener<Response> searchListener) {
+        if (logger.isDebugEnabled()) {
```
alchemist51, line 84: nit: why logger.isDebugEnabled is needed?

server/src/main/java/org/opensearch/index/reindex/ClientPitHitSource.java
```diff
@@ -0,0 +1,298 @@
...
+        executeSearch(searchListener);
+    }
+
+    private void executeSearch(RejectAwareActionListener<Response> searchListener) {
+        // Build a search request with PIT + search_after
+        SearchRequest searchRequest = buildPitSearchRequest();
+        client.search(searchRequest, new ActionListener<SearchResponse>() {
+            @Override
+            public void onResponse(SearchResponse searchResponse) {
+                // Update PIT ID (may change between requests)
```
alchemist51, line 124: Didn't get this part. Can you please explain for my understanding.

PR #21283, opened 2026-04-20
Analytics Plugin Planner — DAG Construction, Plan Forking, and Fragment Conversion

expani:
Builds on the marking phase from #21160 to complete the coordinator-side planning pipeline for the pure Parquet/DataFusion query path. Where PR1 produced an annotated logical plan with per-operator viable backend lists, this PR takes that output through to serialized backend-specific plan bytes ready for the Scheduler.

### What this adds

Query DAG construction — the marked+optimized plan is cut at exchange boundaries into a tree of stages. Each stage carries a target resolver (how to dispatch to shards), an exchange sink provider (how the coordinator reduces results), and other metdata. For this PR only the SINGLETON gather path is implemented; HASH/RANGE shuffle is deferred.

Plan forking — for each stage, one alternative plan is generated per viable backend. Annotations are narrowed to the chosen backend throughout the entire operator tree, producing self-consistent alternatives where every node references the same backend. Delegation-aware forking (where operator and annotation backends differ) is deferred.

Fragment conversion — annotations are stripped from each alternative and the backend's conversion API is called via a composable sequence. The key design decision:
backends never traverse the plan — analytics-engine orchestrates conversion by calling small composable methods in order:
- convertShardScanFragment — data node inner fragment (Filter, Scan, etc.)
- attachPartialAggOnTop — wraps with partial aggregate state
- convertFinalAggFragment — reduce stage reading Arrow batches, with a StageInputScan leaf for schema inference (replaced at execution time with a streaming Arrow batch
source)
- attachFragmentOnTop — any operators above the final agg (Sort, Project)

This keeps backend implementations simple and makes the conversion pipeline explicit and independently testable.

Capability model improvements 

FilterCapability, ProjectCapability, and AggregateCapability now declare supported field types as a Set<FieldType> instead of a single type, reducing declaration verbosity from O(N_ops × N_types) to O(N_ops) per backend. 

ProjectCapability.Scalar gains a supportsLiteralEvaluation flag for routing pure
literal expressions. A pre-marking ReduceExpressionsRule pass eliminates constant predicates before marking rules fire, running in the same HepPlanner instance as the
marking rules.

Bug fixes
getOutputFieldStorage() in filter, sort, and exchange reducer nodes was returning empty inside HEP execution because getInput() returns a HepRelVertex wrapper. Fixed to unwrap before checking. This was exposed by composed pipeline tests (Aggregate(Filter(Scan))).

Test improvements — pipeline shape assertion helpers, composed query shape tests across all operator combinations, dedicated scan and sort rule test files, and per- operator annotation assertion helpers extracted to common methods. 89 tests passing.

### Check List

References: #21160


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ExchangeSink.java
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Coordinator-side sink that accepts Arrow Record Batches streamed from data nodes
+ * and runs the root stage computation (final aggregate, sort, etc.) over them.
+ *
+ * <p>Implementations are backend-specific and created via {@link ExchangeSinkProvider}.
+ * Methods will be added as part of the Scheduler and streaming transport implementation.
+ *
+ * @opensearch.internal
+ */
+public interface ExchangeSink {}
```
expani, line 20: Intentionally empty. Will be added in Scheduler PR

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchStageInputScan.java
```diff
@@ -0,0 +1,89 @@
...
+ * Replaces the child subtree after DAG construction cuts at exchange boundaries.
+ *
+ * <p>The Scheduler feeds Arrow Record Batches from the child stage into the coordinator
+ * sink — this node signals where that input arrives in the root fragment.
+ *
+ * @opensearch.internal
+ */
+public class OpenSearchStageInputScan extends AbstractRelNode implements OpenSearchRelNode {
+
+    private final int childStageId;
```
mch2, line 35: A parent with multiple children is represented as multiple StageInputScans correct? ie for joins we'd have more than one leaf scan node.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchRelNode.java
```diff
@@ -8,14 +8,24 @@
...
 
 /**
  * Marker interface for all OpenSearch custom RelNodes that carry backend assignment
  * and per-column storage metadata.
  *
+ * <p>TODO: consider making this an abstract class storing {@code viableBackends} centrally,
+ * with a default {@link #copyResolved} that returns {@code this} when already narrowed to
+ * a single backend — avoids unnecessary copies during plan forking.
+ *
+ * <p>TODO: when JMH benchmarks show RelNode copy/strip as a hotspot, consider
```
mch2, line 25: this makes sense to do - is the JMH showing it as a hotspot?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FragmentConvertor.java
```diff
@@ -10,75 +10,84 @@
...
  *
- * <p>Backend never traverses the full plan — analytics-engine dictates exactly
- * what to convert and in what context (scan source, shuffle source, in-memory
- * source, shuffle sink).
+ * <p>Composable pipeline for multi-shard aggregate with sort at coordinator:
+ * <ol>
+ *   <li>{@code convertShardScanFragment(tableName, Filter(Scan))} → data node inner bytes</li>
+ *   <li>{@code attachPartialAggOnTop(PartialAgg, innerBytes)} → data node bytes</li>
+ *   <li>{@code convertFinalAggFragment(FinalAgg(StageInputScan))} → reduce stage inner bytes</li>
+ *   <li>{@code attachFragmentOnTop(Sort, innerBytes)} → reduce stage bytes</li>
```
mch2, line 27: is this going to work with substrait to compose byte[] across calls?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchStageInputScan.java
expani: Yes a Join will have multiple StageInputScans. 
This is essentially conveying the schema of the severed child stage to Calcite.
Will add TODO if missed for Joins

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchRelNode.java
expani: Didn't check with JMH yet. 
This refactor required was a lot so deferred with a TODO until this becomes a problem. 
Giving these are cheap constructor calls, hoping it won't add anything significant.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchStageInputScan.java
expani: Added TODO so that we don't miss handling fragment conversion

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FragmentConvertor.java
expani: Substrait has an AggregationPhase that represents Partial/Final and other Agg modes. DataFusion just ignores them when converting into it's Logical plan. 

Which is why we had to write the `partial_agg_optimizer` in the initial POC. 

The long term goal is to use a custom Substrait Consumer in DataFusion.

But in the interest of time we will ignore the attaching part in DataFusion and only use those as indicators to trigger the `partial_agg_optimizer`

So, when `attachPartialAggOnTop(PartialAgg, innerBytes)` is called, DF backend will ignore innerBytes and just convert PartialAgg into Substrait which is the whole RelNode.

expani: During execution at the Shard, `partial_agg_optimizer` will handle stripping the Final Mode

mch2 (approved):

mch2:
Thanks @expani  - I pushed a few license fixes to get the build to green.  Also our sandbox check has been failing recently due to a mimalloc issue rust side causing crashes.  Have commented that out for now with a TODO so we aren't blocking changes.

PR #21286, opened 2026-04-20
Add DynamicLimitPool for runtime memory pool limit changes

gaurav-amz:
Replace GreedyMemoryPool with DynamicLimitPool in the DataFusion runtime. The pool stores its limit in an AtomicUsize shared with a DynamicLimitHandle, allowing the limit to be changed at runtime without restarting the process.

Rust changes:
- New memory.rs: DynamicLimitPool + DynamicLimitHandle with AtomicUsize limit
- api.rs: Use DynamicLimitPool, store handle in DataFusionRuntime, add get_memory_pool_usage/limit and set_memory_pool_limit functions
- ffm.rs: Export df_get_memory_pool_usage, df_get_memory_pool_limit, df_set_memory_pool_limit as extern C functions

Java changes:
- NativeBridge.java: Add FFM method handles and static methods for the 3 new memory pool functions
- DataFusionService.java: Expose getMemoryPoolUsage(), getMemoryPoolLimit(), setMemoryPoolLimit() methods

Runtime behavior:
- Increasing limit takes effect immediately for new allocations
- Decreasing limit only affects new allocations; existing reservations drain naturally as queries complete
- Without set_limit calls, behavior is identical to GreedyMemoryPool

### Description

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -214,8 +217,9 @@ pub fn create_global_runtime(
         .with_max_temp_directory_size(spill_limit as u64)
         .with_mode(DiskManagerMode::Directories(vec![PathBuf::from(spill_dir)]));
 
+    let (dynamic_pool, dynamic_limit_handle) = DynamicLimitPool::new(memory_pool_limit as usize);
```
himshikhagupta, line 152: Should this wrap the GreedyMemoryPool instead of replacing?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionService.java
```diff
@@ -106,6 +106,29 @@ public NativeRuntimeHandle getNativeRuntime() {
         return handle;
     }
 
+    /**
+     * Returns the current memory pool usage in bytes.
+     */
+    public long getMemoryPoolUsage() {
+        return NativeBridge.getMemoryPoolUsage(getNativeRuntime().get());
+    }
```
himshikhagupta, line 132: Who will be the consumer of these methods?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionService.java
```diff
@@ -106,6 +106,29 @@ public NativeRuntimeHandle getNativeRuntime() {
...
+     */
+    public long getMemoryPoolLimit() {
+        return NativeBridge.getMemoryPoolLimit(getNativeRuntime().get());
+    }
+
+    /**
+     * Sets the memory pool limit at runtime. Takes effect for new allocations only.
+     * Existing reservations that exceed the new limit are NOT reclaimed.
+     */
+    public void setMemoryPoolLimit(long newLimitBytes) {
```
himshikhagupta, line 149: Are we planning to add cluster setting for this?

sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
```diff
@@ -0,0 +1,194 @@
...
+        &self,
+        reservation: &MemoryReservation,
+        additional: usize,
+    ) -> Result<(), DataFusionError> {
+        let limit = self.dynamic_limit.load(Ordering::SeqCst);
+        self.used
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |used| {
+                let new_used = used + additional;
+                (new_used <= limit).then_some(new_used)
+            })
```
himshikhagupta: The limit is loaded before fetch_update, but another thread could call set_limit between the load and the CAS. Would this create a race condition where the new limit never gets applied?

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -190,9 +191,11 @@ pub async fn create_object_metas(
 }
 
 /// Opaque runtime handle returned to the caller.
-/// Contains the DataFusion RuntimeEnv (memory pool, disk spill, cache).
+/// Contains the DataFusion RuntimeEnv (memory pool, disk spill, cache)
+/// and a handle to change the memory pool limit at runtime.
 pub struct DataFusionRuntime {
     pub runtime_env: datafusion::execution::runtime_env::RuntimeEnv,
+    pub dynamic_limit_handle: DynamicLimitHandle,
```
himshikhagupta: Does this need to be pub?

sandbox/plugins/analytics-backend-datafusion/rust/src/ffm.rs
```diff
@@ -70,6 +70,32 @@ pub unsafe extern "C" fn df_close_global_runtime(ptr: i64) {
...
+pub unsafe extern "C" fn df_get_memory_pool_usage(runtime_ptr: i64) -> i64 {
+    if runtime_ptr == 0 { return 0; }
+    api::get_memory_pool_usage(runtime_ptr)
+}
+
+/// Returns current memory pool limit in bytes.
+/// Java: MethodHandle(JAVA_LONG → JAVA_LONG)
+#[no_mangle]
+pub unsafe extern "C" fn df_get_memory_pool_limit(runtime_ptr: i64) -> i64 {
+    if runtime_ptr == 0 { return 0; }
```
himshikhagupta: What does it mean if its 0? Do we need exception handling here?
Consider using the #[ffm_safe] pattern with error return

sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
```diff
@@ -0,0 +1,194 @@
...
+    fn shrink(&self, _reservation: &MemoryReservation, shrink: usize) {
+        self.used.fetch_sub(shrink, Ordering::Relaxed);
+    }
+
+    fn try_grow(
+        &self,
+        reservation: &MemoryReservation,
+        additional: usize,
+    ) -> Result<(), DataFusionError> {
+        let limit = self.dynamic_limit.load(Ordering::SeqCst);
```
himshikhagupta: Is using SeqCst overkill here since this method will be called for every memory allocation? Do we know the overhead/pitfalls of using SeqCst vs  Release/Acquire?

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
gaurav-amz: No, GreedyMemoryPool stores its limit as a plain usize with no setter method. There's no way to change it after construction, so wrapping it wouldn't give us runtime mutability. DynamicLimitPool reimplements the same try_grow/grow/shrink logic with an AtomicUsize for the limit. Without set_limit calls, behavior is identical to GreedyMemoryPool.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionService.java
gaurav-amz: This is exposed to stats API or admission control to get the memory usage if needed.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionService.java
gaurav-amz: Yes. Follow up PR will have this change to wire a clusterSettings.addSettingsUpdateConsumer() that calls setMemoryPoolLimit().

sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
gaurav-amz: Not a real race condition. If set_limit fires between the load and the CAS, at most one allocation uses the stale limit. CAS itself is atomic on used, so the pool state stays consistent. The very next try_grow call reads the new limit. This is the same pattern GreedyMemoryPool uses

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
gaurav-amz: Good catch. Will change to pub(crate), it's will only be accessed within the crate (by ffm.rs via api.rs).

sandbox/plugins/analytics-backend-datafusion/rust/src/ffm.rs
gaurav-amz: Yes. Will change all three FFM functions to use #[ffm_safe]. They will return a proper error string for null pointers instead of silently returning 0. This will be consistent with the other FFM functions in ffm.rs

sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
```diff
@@ -0,0 +1,194 @@
...
+        assert!(reservation.try_grow(2048).is_ok());
+
+        // Decrease limit below current usage
+        handle.set_limit(1024);
+
+        // Existing reservation is NOT reclaimed
+        assert_eq!(pool.reserved(), 2048);
+
+        // But new allocations fail
+        let consumer2 = MemoryConsumer::new("test2");
```
bharath-techie, line 215: So the existing queries which needs new allocations will still fail right ?

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
himshikhagupta: I am worried about maintainability here. What if GreedyMemoryPool changes something, how do we plan to keep this identical in the long term?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionService.java
himshikhagupta: Can we include the cluster setting in this PR only to ensure this flow is logically complete.

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
gaurav-amz: Test suite covers the MemoryPool behaviour contract directly — overflow in try_grow, saturation in grow, dynamic increase, dynamic decrease with existing reservations, and concurrent set_limit races. Any upstream change that altered observable MemoryPool semantics would trip one of these.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionService.java
gaurav-amz: Done

sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
gaurav-amz: Yes, This matches GreedyMemoryPool's semantics — the only difference is that the limit can change at runtime. An operator doing a live decrease should expect some in-flight queries to fail where the query won't spill to disk. There is no safe way to handle this.

sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
gaurav-amz: Yes, addressed in a later revision. The hot used counter now uses Relaxed (same as GreedyMemoryPool upstream) and dynamic_limit uses Release/Acquire, which is the minimum needed to publish a set_limit store to subsequent try_grow.

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -231,12 +234,26 @@ pub fn create_global_runtime(
...
+            "memory_pool_limit must be non-negative, got {}",
+            memory_pool_limit
+        )));
+    }
+    if spill_limit < 0 {
+        return Err(DataFusionError::Configuration(format!(
+            "spill_limit must be non-negative, got {}",
+            spill_limit
+        )));
+    }
```
himshikhagupta, line 146: Should these checks be done when cluster setting is updated?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -173,6 +183,35 @@ public List<Setting<?>> getSettings() {
...
+            return;
+        }
+        try {
+            service.setMemoryPoolLimit(newLimitBytes);
+            logger.info("Updated DataFusion memory pool limit to {}B", newLimitBytes);
+        } catch (IllegalStateException e) {
+            // Service has been stopped/closed (e.g., during node shutdown). The listener is
+            // still registered on ClusterSettings because there is no removeSettingsUpdateConsumer
+            // API; swallow the race so cluster-state application does not log a spurious failure.
+            logger.debug("Ignoring memory pool limit update to {}B; service is not running", newLimitBytes);
```
himshikhagupta: nit: should be warn so we atleast know some update was missed

himshikhagupta (commented):
Few comments, LGTM otherwise.

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
gaurav-amz: Setting layer already validates (minValue=0L via Setting.longSetting) and the Rust checks are defense for non cluster setting callers.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
gaurav-amz: Ack.

himshikhagupta:
LGTM!

himshikhagupta (approved):

bharath-techie (approved):
Lgtm

PR #21294, opened 2026-04-20
Fix YAML parser corrupting string values that resemble booleans

andrross:
Jackson 2.x (via SnakeYAML 1.x) recognized 18 YAML 1.1 boolean
forms as VALUE_BOOLEAN tokens: yes/Yes/YES, no/No/NO, true/True/
TRUE, false/False/FALSE, on/On/ON, off/Off/OFF. It only did this
for unquoted (plain) scalars — quoted strings like "no" stayed as
VALUE_STRING.

Jackson 3.x only recognizes lowercase true/false as VALUE_BOOLEAN.
The other 16 forms are emitted as VALUE_STRING regardless of
quoting. The migration (https://github.com/opensearch-project/OpenSearch/pull/21029) added a text() override in
YamlXContentParser using equalsIgnoreCase to normalize these, but
this matched all 48 case permutations of true/false rather than the
specific 16 forms, and could not distinguish quoted from unquoted
scalars, corrupting strings like document IDs.

Fix by subclassing Jackson's YAMLParser (OpenSearchYamlParser) to
override _decodeScalar, which has access to the ScalarEvent and its
isPlain() method. For unquoted scalars matching the 16 SnakeYAML
boolean forms, we return VALUE_TRUE/VALUE_FALSE — exactly restoring
Jackson 2.x behavior. Quoted scalars are unaffected. This requires
no overrides in YamlXContentParser and no changes to any caller.

### Check List

### DIFF
--- a/libs/x-content/src/main/java/org/opensearch/common/xcontent/yaml/YamlXContentParser.java
+++ b/libs/x-content/src/main/java/org/opensearch/common/xcontent/yaml/YamlXContentParser.java
@@ -37,11 +37,7 @@ import org.opensearch.common.xcontent.json.JsonXContentParser;
 import org.opensearch.core.xcontent.DeprecationHandler;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 
-import java.io.IOException;
-
-import org.snakeyaml.engine.v2.resolver.ScalarResolver;
 import tools.jackson.core.JsonParser;
-import tools.jackson.dataformat.yaml.YAMLParser;
 
 public class YamlXContentParser extends JsonXContentParser {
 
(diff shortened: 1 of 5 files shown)


andrross:
FYI @reta 

Here is an example of the failures I've seen: https://build.ci.opensearch.org/job/gradle-check/74684/testReport/junit/org.opensearch.action.get/MultiGetResponseTests/testFromXContent/

reta:
> FYI @reta
> 
> Here is an example of the failures I've seen: https://build.ci.opensearch.org/job/gradle-check/74684/testReport/junit/org.opensearch.action.get/MultiGetResponseTests/testFromXContent/

Thanks @andrross , we could also be very explicit with checking "True", "true", "False", "false", but up to you, thanks for fixing it! Very sad I didn't catch it before

reta (approved):

libs/x-content/src/main/java/org/opensearch/common/xcontent/yaml/YamlXContentParser.java
```diff
@@ -55,29 +66,36 @@ public XContentType contentType() {
...
      *
-     * The better way to deal with that would be be to provide own dedicated {@link ScalarResolver}
+     * The better way to deal with that would be to provide a dedicated {@link ScalarResolver}
      * implementation but unfortunately {@link YAMLParser} does not provide a way to set or override it
      * currently.
      */
     @Override
-    public String text() throws IOException {
-        String text = super.text();
+    public boolean isBooleanValue() throws IOException {
```
reta: @andrross from the test failures, I suspect parser never gets to this point, we may fallback to text() again (I will also take a look shortly).

andrross: Ok, I think I fixed it. Added back `text()` override, but only normalize the text if it is one of the valid boolean forms, otherwise it falls through and returns the original text.

reta: We don't need `booleanValue` and `isBooleanValue` anymore, right?

andrross: So I think the approach of normalizing text() just doesn't work. The latest test failure (https://build.ci.opensearch.org/job/gradle-check/74826/) we hit a version of the Norway problem...there is a string "On" that gets turned into "true". I'm going to push another commit that takes a different approach...

reta (commented):
Thanks @andrross , I wish there was an easier way to configure own scalar parsers ...

reta:
@andrross just FYI, the issue has been addressed here [1], I will create a followup issue to reflect that once new version is released.

[1] https://github.com/FasterXML/jackson-dataformats-text/pull/627

PR #21305, opened 2026-04-21
[BUG] Bulk request got stuck in primary phase when index being deleted

### LINKED ISSUE #21304: [BUG] Bulk request got stuck in primary phase when index being deleted
### Describe the bug

We have encountered a scenario where deleting an index during an ongoing bulk request occasionally causes the request to hang：
```
indices:data/write/bulk                              w8BJJfxzSNOFNqdPkRdY-Q:544205   -                               transport 1776686293827 11:58:13 19.4h      192.168.92.74 opensearch-node-3 requests[500], indices[log1]
indices:data/write/bulk[s]                           w8BJJfxzSNOFNqdPkRdY-Q:544208   w8BJJfxzSNOFNqdPkRdY-Q:544205   transport 1776686293827 11:58:13 19.4h      192.168.92.74 opensearch-node-3 requests[170], index[log1][2]
indices:data/write/bulk[s]                           GYc3Nq7nRjm8kqrGLn2STw:2205626  w8BJJfxzSNOFNqdPkRdY-Q:544208   transport 1776686293827 11:58:13 19.4h      192.168.88.71 opensearch-node-2 requests[170], index[log1][2]
indices:data/write/bulk[s][p]                        GYc3Nq7nRjm8kqrGLn2STw:2205627  GYc3Nq7nRjm8kqrGLn2STw:2205626  direct    1776686293827 11:58:13 19.4h      192.168.88.71 opensearch-node-2 requests[170], index[log1][2]
```
Task details:
```
{
  "completed": false,
  "task": {
    "node": "GYc3Nq7nRjm8kqrGLn2STw",
    "id": 2205627,
    "type": "direct",
    "action": "indices:data/write/bulk[s][p]",
    "status": {
      "phase": "primary"
    },
    "description": "requests[170], index[log1][2]",
    "start_time": "2026-04-20T11:58:13.827Z",
    "start_time_in_millis": 1776686293827,
    "running_time": "19.5h",
    "running_time_in_nanos": 70245793552353,
    "cancellable": false,
    "cancelled": false,
    "parent_task_id": "GYc3Nq7nRjm8kqrGLn2STw:2205626",
    "headers": {},
    "resource_stats": {
		// ...
    }
  }
```

Here is the related trace log:
```
[2026-04-21T07:30:22,485][TRACE][o.o.a.b.TransportShardBulkAction] [opensearch-node-2] [[log1][2]] op [indices:data/write/bulk[s]] completed on primary for request [BulkShardRequest [[log1][2]] containing [175] requests]
[2026-04-21T07:30:22,485][TRACE][o.o.a.b.TransportShardBulkAction] [opensearch-node-2] [[log1][2]] sending op [indices:data/write/bulk[s]] to replica [log1][2], node[Ly-RRJ46TdCQ6Y014G_7Xw], [R], s[STARTED], a[id=uXDJ6e9bTZ21fDTZjNq1sA] for request [BulkShardRequest [[log1][2]] containing [175] requests]
[2026-04-21T07:30:22,512][TRACE][o.o.a.b.TransportShardBulkAction] [opensearch-node-2] [[log1][2]] failure while performing [indices:data/write/bulk[s]] on replica [log1][2], node[Ly-RRJ46TdCQ6Y014G_7Xw], [R], s[STARTED], a[id=uXDJ6e9bTZ21fDTZjNq1sA], request [BulkShardRequest [[log1][2]] containing [175] requests]
org.opensearch.index.shard.PrimaryShardClosedException: CurrentState[CLOSED] Primary closed
        at org.opensearch.action.support.replication.PendingReplicationActions.lambda$close$1(PendingReplicationActions.java:136) ~[opensearch-3.6.0]
        at org.opensearch.action.support.replication.PendingReplicationActions.lambda$cancelActions$2(PendingReplicationActions.java:141) ~[opensearch-3.6.0]
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184) ~[?:?]
        at java.base/java.util.concurrent.ConcurrentHashMap$KeySpliterator.forEachRemaining(ConcurrentHashMap.java:3573) ~[?:?]
        at java.base/java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:762) ~[?:?]
        at java.base/java.util.stream.ReferencePipeline$7$1.accept(ReferencePipeline.java:276) ~[?:?]
        at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708) ~[?:?]
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509) ~[?:?]
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499) ~[?:?]
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:151) ~[?:?]
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:174) ~[?:?]
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:?]
        at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:596) ~[?:?]
        at org.opensearch.action.support.replication.PendingReplicationActions.lambda$cancelActions$3(PendingReplicationActions.java:141) ~[opensearch-3.6.0]
        at org.opensearch.common.util.concurrent.ThreadContext$ContextPreservingRunnable.run(ThreadContext.java:916) ~[opensearch-3.6.0]
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[?:?]
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[?:?]
        at java.base/java.lang.Thread.run(Thread.java:1583) [?:?]
```

### Related component

Indexing

### To Reproduce

Start two threads: one executes the bulk request, and the other deletes the index after a random sleep. The issue can be reproduced after waiting for a short period.

### Expected behavior

Bulk task never stuck

### Additional Details

**Plugins**
All default plugins

**Screenshots**
See above logs

 - OS: Linux
 - Version OpenSearch 3.6.0

xiaoyuan0821:
### Description
Bulk request got stuck in primary phase when index being deleted

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/21304

### Check List


server/src/main/java/org/opensearch/action/support/replication/TransportWriteAction.java
```diff
@@ -586,6 +586,9 @@ public void failShardIfNeeded(
                     exception,
                     listener
                 );
+            } else {
+                // always call listener
```
kkewwei, line 590: It makes sense to me.  Moving the commit above here would be more appropriate.

>  If a write action fails due to the closure of the primary shard
then the replicas should not be marked as failed since they are
still up-to-date with the (now closed) primary shard

kkewwei (approved):
LGTM

server/src/main/java/org/opensearch/action/support/replication/TransportWriteAction.java
xiaoyuan0821: If exception is not instance of `PrimaryShardClosedException`, listener will be called in `shardStateAction.remoteShardFailed`, we only need to add additional call in the else branch

andrross (approved):

AndrewCharlesHay:
Correctness fix looks right — the bug was that `failShardIfNeeded` dropped the listener silently when the primary was already closed, leaving the bulk request hanging on primary phase until it timed out. Always invoking the listener with `onResponse(null)` in the else branch is the standard async-safety pattern.

Security/reliability notes:

1. **Hung requests are a DoS primitive.** Before this fix, a user issuing a bulk request targeting a deleted/deleting index would have their request sit in the primary phase until some upstream timeout fired. Under concurrent write load against an index being deleted, that's a clean way to exhaust the primary-shard action queue and affect other tenants on the same node. Good candidate for a backport to maintenance branches — the fix is one line and the worst-case impact is cluster-wide write starvation.

2. **Test asserts `callbackCount == 1` but not the value passed.** `listener.onResponse(null)` vs `listener.onFailure(ex)` both increment the counter. Consider tightening the test to verify it's specifically an `onResponse(null)` — otherwise a future refactor that converts this to an `onFailure` with a bogus exception would pass the test but change caller-visible behavior. `ActionListener.wrap(callbackCount::incrementAndGet)` doesn't let you distinguish success vs failure paths; switch to separate runnables or capture the callback value.

3. **Silent `null` response**: bulk callers interpret `null` as "no failure" — downstream logic needs to be OK treating this as a no-op. Worth a skim of the primary-closed callers to confirm none of them unconditionally dereference the `TransportReplicationAction.ReplicaResult` they get back. If any do, you'd be swapping "hangs forever" for "NPE at a different layer".

4. **Nit**: the comment `// always call listener` is thin. Expanding to `// When the primary is already closed there's no shard to fail, but the caller still needs the listener invoked or the outer request hangs on primary phase. See #<original-issue>.` makes the invariant obvious to future readers.

Worth backporting — this looks like a latent hang bug in every version since `failShardIfNeeded` was introduced.

PR #21306, opened 2026-04-21
Integrating parquet-data-format to search path

vinaykpud:
### Description
  - Integrate parquet data format into the search path by wiring DataFusionPlugin as a SearchBackEndPlugin that declares "parquet" as its supported format. The DataFormatRegistry resolves format names to DataFormat objects and creates reader managers for each backend.
  - Fix `DataFormatAwareEngine.refresh()` to only call `commitNewSnapshot `when data actually changes. Previously it committed on every refresh cycle (even idle ones), which incremented the catalog generation, deleted the old snapshot, and removed the reader causing search to fail with "No DataFusion reader available".
  - Change `SearchBackEndPlugin.getSupportedFormats()` to return `List<String>` (format names) instead of `List<DataFormat>`, simplifying backend declarations and eliminating the need for mock DataFormat objects.

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -66,13 +63,26 @@ public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<Data
...
     private volatile DataFusionService dataFusionService;
+    private volatile DataFormatRegistry dataFormatRegistry;
 
     /**
      * Creates the DataFusion plugin.
      */
     public DataFusionPlugin() {}
 
+    @Override
+    public Collection<Object> createComponents(DataFormatRegistry dataFormatRegistry) {
```
bharath-techie: Lets combine both the `createComponents` similar to this - https://github.com/opensearch-project/OpenSearch/blob/feature/datafusion/plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DataFusionPlugin.java#L118

bharath-techie: We are creating reader managers as a checked function as part of `DataFormatRegistry` ,  which will be called way later during shard initialization, so there is no cyclic dependency if you combine create components. [ Even today plugin's create components get called after init of df registry ]

server/src/main/java/org/opensearch/plugins/SearchBackEndPlugin.java
```diff
@@ -45,4 +47,15 @@ public interface SearchBackEndPlugin<R> {
...
     EngineReaderManager<?> createReaderManager(ReaderManagerConfig settings) throws IOException;
+
+    /**
+     * Called after the {@link DataFormatRegistry} is available. Plugins can use this
+     * to receive the registry and resolve data format names to {@link org.opensearch.index.engine.dataformat.DataFormat} objects.
+     *
+     * @param dataFormatRegistry the data format registry
+     * @return additional components to bind in Guice, or empty
+     */
+    default Collection<Object> createComponents(DataFormatRegistry dataFormatRegistry) {
```
bharath-techie: Lets remove this as well.

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshot.java
```diff
@@ -240,4 +240,15 @@ public CatalogSnapshot cloneNoAcquire() {
      */
     public abstract Collection<String> getFiles(boolean includeSegmentsFile) throws IOException;
 
+    @Override
+    public boolean equals(Object o) {
```
mgodwan: Why is this needed? Shouldn't `==` check suffice for same process comparison?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionReaderManager.java
```diff
@@ -86,10 +86,15 @@ public void beforeRefresh() throws IOException {}
...
-        if (didRefresh == false) return;
         if (readers.containsKey(catalogSnapshot)) return;
-        DatafusionReader reader = new DatafusionReader(directoryPath, catalogSnapshot.getSearchableFiles(dataFormat.name()));
-        readers.put(catalogSnapshot, reader);
+        if (didRefresh) {
+            var files = catalogSnapshot.getSearchableFiles(dataFormat.name());
+            DatafusionReader reader = new DatafusionReader(directoryPath, files);
+            readers.put(catalogSnapshot, reader);
+        } else if (!readers.isEmpty()) {
+            DatafusionReader latest = readers.values().iterator().next();
```
bharath-techie: Why is this change required ? 

Also `readers.values().iterator().next();` might be wrong right, since order is not maintained in hashmap.

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshot.java
```diff
@@ -240,4 +240,15 @@ public CatalogSnapshot cloneNoAcquire() {
...
+        if (this == o) return true;
+        if (o instanceof CatalogSnapshot == false) return false;
+        return generation == ((CatalogSnapshot) o).generation;
+    }
+
+    @Override
+    public int hashCode() {
+        return Long.hashCode(generation);
```
bharath-techie: Can we follow the same convention as other opensearch objects for these ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
vinaykpud: How can we combine both? changing the existing [createComponents](https://github.com/opensearch-project/OpenSearch/blob/feature/datafusion/plugins/engine-datafusion/src/main/java/org/opensearch/datafusion/DataFusionPlugin.java#L118) signature is a core API change affects all plugins.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionReaderManager.java
vinaykpud: I haven't verified this change, generated by AI, but the issue I was facing intially was like bellow
1. when refresh is called, a snapshot instance is saved
2. but during the search time it looks for the latest snapshot instance basically different instance which is not saved

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
vinaykpud: are you suggesting to to have `createComponents` method in the SearchBackEndPlugin and combine both?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsExtension.java
```diff
@@ -0,0 +1,150 @@
...
+ * SPI extension discovered by analytics-engine via {@code META-INF/services}.
+ * <p>
+ * Receives the fully-initialized {@link DataFusionPlugin} instance via its single-arg
+ * constructor (supported by {@code PluginsService.createExtension()}), so it has access
+ * to the {@link DataFusionService} created during plugin lifecycle.
+ * <p>
+ * Declares all analytics query capabilities (operators, filters, aggregates) and
+ * creates per-shard execution engines.
+ */
+public class DataFusionAnalyticsExtension implements AnalyticsSearchBackendPlugin {
```
expani: nit: DataFusionAnalyticsBackendPlugin

server/src/main/java/org/opensearch/plugins/SearchBackEndPlugin.java
```diff
@@ -45,4 +59,41 @@ public interface SearchBackEndPlugin<R> {
...
+     * @param xContentRegistry the registry for extensible xContent parsing
+     * @param environment the environment for path and setting configurations
+     * @param nodeEnvironment the node environment used coordinate access to the data paths
+     * @param namedWriteableRegistry the registry for {@link org.opensearch.core.common.io.stream.NamedWriteable} object parsing
+     * @param indexNameExpressionResolver A service that resolves expression to index and alias names
+     * @param repositoriesServiceSupplier A supplier for the service that manages snapshot repositories
+     * @param dataFormatRegistry the data format registry for resolving format names to DataFormat objects
+     * @return components to bind in Guice, or empty
+     */
+    default Collection<Object> createComponents(
```
expani, line 88: I am assuming this is to pass `DataFormatRegistry` and other custom stuff related to AnalyticsPlugin ? 

If so, can we keep the method restricted to only those classes ? 

And let plugins get Client, ClusterService and other common parameters via the existing way that all other OpenSearch plugins do.

expani: We can add a comment if this was done for some specific reason and not split specifically for dataFormatRegistry

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -36,17 +33,16 @@
...
  * <p>
- * Initializes the {@link DataFusionService} at node startup and creates
- * per-shard {@link DatafusionSearchExecEngine} instances via the
- * {@link AnalyticsSearchBackendPlugin} SPI.
+ * Owns the {@link DataFusionService} lifecycle (memory pool, native runtime).
+ * Analytics query capabilities are declared in {@link DataFusionAnalyticsExtension},
+ * which is SPI-discovered and receives this plugin instance via its constructor.
  */
-public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<DatafusionReader>, AnalyticsSearchBackendPlugin {
+public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<DatafusionReader> {
```
expani, line 45: nit : Unrelated to this PR, can we add a TODO to rename `SearchBackEndPlugin` as the name is confusing with AnalyticsSearchBackendPlugin

expani (commented):
Thanks @vinaykpud for integrating this

server/src/main/java/org/opensearch/plugins/SearchBackEndPlugin.java
expani: This looked like the right way of going about it. 
Not sure why this was changed again

server/src/main/java/org/opensearch/plugins/SearchBackEndPlugin.java
vinaykpud: We can do either way. @bharath-techie suggested to combine them like this to one `createComponents` specific to the to follow similar to the [TelemetryAwarePlugin](https://github.com/opensearch-project/OpenSearch/blob/bbdf73ac00f6b97d167f581655478fb2c0e0adbf/server/src/main/java/org/opensearch/plugins/TelemetryAwarePlugin.java#L63)

Another way is to use [`Plugin.createComponents`](https://github.com/opensearch-project/OpenSearch/blob/bbdf73ac00f6b97d167f581655478fb2c0e0adbf/server/src/main/java/org/opensearch/plugins/Plugin.java#L142) for all the common parameters  

and have `SearchBackEndPlugin.createComponents` just for `DataFormatRegistry`.

Initially i had like this and then combined like now.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
vinaykpud: We can do as part of this PR

Since it's mostly related to the underlying query engine and data format reader management, we can rename `SearchBackEndPlugin` to `EngineReaderPlugin`. DataFusionPlugin and LucenePlugin will implement this.

Other alternatives
1. BackEndReaderPlugin

expani: Not sure on the name, maybe add a TODO for now we can deal with it later :)

server/src/main/java/org/opensearch/plugins/SearchBackEndPlugin.java
expani: Okay taking the telemetry example if we need a MetricsRegistry later then we will add it here as another parameter ? 

> and have SearchBackEndPlugin.createComponents just for DataFormatRegistry.

Yeah this makes sense as DataFusion is telling that I need DataFormatRegistry and TelemetryAwarePlugin can have a different SPI like `SearchBackEndPlugin` maybe `MetricPublishingPlugin` for injecting the MetricsRegistry. 

In the interest of time, we can add a TODO to revisit later.

vinaykpud: >  Okay taking the telemetry example if we need a MetricsRegistry later then we will add it here as another parameter ?

Yes

expani (approved):

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -586,7 +586,13 @@ public void refresh(String source) throws EngineException {
 
                         RefreshInput refreshInput = new RefreshInput(existingSegments, newSegments);
                         RefreshResult result = indexingExecutionEngine.refresh(refreshInput);
-                        catalogSnapshotManager.commitNewSnapshot(result.refreshedSegments());
+                        // Only commit a new snapshot when data actually changed. Committing on every
+                        // refresh cycle (even idle ones) increments the generation and deletes the old
+                        // snapshot, which removes the reader from EngineReaderManager. Search then fails
+                        // because the latest snapshot has no reader associated with it.
+                        if (refreshed) {
```
mgodwan, line 588: We can short circuit before even engine's refresh is called.

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshot.java
```diff
@@ -277,5 +277,4 @@ public CatalogSnapshot cloneNoAcquire() {
      * @throws IOException in case of I/O error
      */
     public abstract Collection<String> getFiles(boolean includeSegmentsFile) throws IOException;
-
```
mgodwan, line 280: [nit] this change should be removed

sandbox/plugins/parquet-data-format/build.gradle
```diff
@@ -27,8 +27,13 @@ dependencies {
...
   implementation "com.google.flatbuffers:flatbuffers-java:${versions.flatbuffers}"
-  implementation "com.fasterxml.jackson.core:jackson-core:${versions.jackson}"
-  implementation "com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}"
+  implementation "org.slf4j:slf4j-api:${versions.slf4j}"
+  // jackson-core is on the server classpath; jackson-databind and jackson-annotations are not.
+  compileOnly "com.fasterxml.jackson.core:jackson-core:${versions.jackson}"
+  implementation("com.fasterxml.jackson.core:jackson-databind:${versions.jackson_databind}") {
+    exclude group: 'com.fasterxml.jackson.core', module: 'jackson-core'
+    exclude group: 'com.fasterxml.jackson.core', module: 'jackson-annotations'
+  }
```
mgodwan, line 36: Why are these dependencies needed at all?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -114,49 +122,9 @@ public EngineReaderManager<DatafusionReader> createReaderManager(ReaderManagerCo
...
-            }
-            if (dfReader == null) {
-                throw new IllegalStateException("No DatafusionReader available in the acquired reader");
-            }
-            DatafusionContext context = new DatafusionContext(ctx.getTask(), dfReader, dataFusionService.getNativeRuntime());
-            DatafusionSearchExecEngine engine = new DatafusionSearchExecEngine(context, dataFusionService::newChildAllocator);
-            engine.prepare(ctx);
-            return engine;
-        };
+    public List<String> getSupportedFormats() {
```
mgodwan, line 126: We should still return List of DataFormat here. 
String can be an internal representation for compile time, but we should use registry to get the actual format, and then return the supported formats from here

sandbox/plugins/analytics-backend-datafusion/build.gradle
mgodwan, line 11: I see you've added another plugin class but the same is not added for detection.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
bharath-techie: Yeah this makes sense , now that we have figured out how to initialize the plugin with data format registry, `SUPPORTED_FORMAT` can be string , but return type can be List<DataFormat>

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
vinaykpud: Its fixed in https://github.com/opensearch-project/OpenSearch/pull/21321

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
vinaykpud: We can revisit this later, i am going to add a TODO here. For now as discussed keeping it list of Strings.

sandbox/plugins/parquet-data-format/build.gradle
vinaykpud: 1. slf4j-api (implementation) : Arrow's BaseAllocator calls LoggerFactory.getLogger() in its static initializer. Without it: NoClassDefFoundError: org/slf4j/LoggerFactory
  2. jackson-databind (implementation, with exclusions) : Arrow's Schema creates new ObjectMapper() in its static initializer. Without it: NoClassDefFoundError: com/fasterxml/jackson/databind/util/ByteBufferBackedInputStream
  3. jackson-annotations (implementation, direct) : ObjectMapper initializes JacksonAnnotationIntrospector which loads JsonSerializeAs. Without it: NoClassDefFoundError: com/fasterxml/jackson/annotation/JsonSerializeAs
  4. jackson-core (excluded from jackson-databind transitive) — already on the OpenSearch server classpath, bundling it would cause jar hell with core
  5. jackson-annotations (excluded from jackson-databind transitive, added directly) : jackson-databind transitively pulls jackson-annotations which transitively pulls jackson-core. Excluding it from jackson-databind and adding it directly prevents jackson-core from being pulled in transitively.

sandbox/plugins/analytics-backend-datafusion/build.gradle
vinaykpud: If you mean `DataFusionAnalyticsBackendPlugin` is not a Plugin subclass and it implements `AnalyticsSearchBackendPlugin` and is discovered via META-INF/services SPI by the analytics-engine plugin's loadExtensions(). 

Only the main DataFusionPlugin (which extends Plugin) needs to be registered in classname.

mch2 (approved):

PR #21313, opened 2026-04-21
Add delete Index engine abstraction

thorkous:
### Description

Introduces the delete execution engine abstraction for `DataFormatAwareEngine` — enabling pluggable, format-specific delete handling in OpenSearch's multi-format engine. This builds on the `DataFormatAwareEngine` introduced in #21181, adding the delete-side counterpart to the existing `IndexingExecutionEngine`.

`DeleteExecutionEngine` follows the same plugin discovery and factory pattern as `CommitterFactory`. Each engine plugin provides a `DeleteExecutionEngineFactory` that is collected during index creation and stored on `EngineConfig`. When the shard initializes, `DataFormatAwareEngine` calls the factory with the already-created `Committer`, allowing the delete engine to be constructed with all required dependencies at the right time.

Each `Deleter` is paired 1:1 with a `Writer` sharing the same generation — enabling format-specific delete tracking (e.g., live-doc bitsets for Parquet files).

**Key additions:**

- `DeleteExecutionEngine<T>` — format-specific engine interface for delete operations, paired with writers via shared generation
- `DeleteExecutionEngineFactory` — factory interface (mirrors `CommitterFactory`) allowing deferred construction with the `Committer`
- `LuceneDeleteExecutionEngine` — Lucene-based implementation that tracks per-generation deleters and delegates to the committer's IndexWriter
- `LuceneDeleteExecutionEngineFactory` — concrete factory registered by `LucenePlugin` (mirrors `LuceneCommitterFactory`)
- `Deleter` — per-writer delete handler implementing `Lockable` for thread-safe pooling, mirrors `Writer` interface pattern
- `DeleterImpl` — Lucene-based deleter that delegates to `LuceneWriter.deleteDocument()`
- `DeleteResult` — sealed result type (Success/Failure) for delete operations, mirrors `WriteResult`
- `Writer.getWriterForFormat(String)` — default method on the `Writer` interface enabling format-specific writer resolution across plugin boundaries
- `EnginePlugin.getDeleteExecutionEngineFactory()` — SPI method for plugins to provide their delete engine factory

**Design decisions:**

1. **`Writer.getWriterForFormat(String)`** — Added because `CompositeWriter` (in the composite-engine plugin) wraps per-format writers including `LuceneWriter` (in the analytics-backend-lucene plugin). Since these reside in different plugin classloaders, `DeleterImpl` cannot directly cast or import `CompositeWriter`. The `getWriterForFormat` method on the `Writer` interface provides a plugin-boundary-safe way to resolve the underlying Lucene writer without cross-plugin dependencies. `CompositeWriter` overrides it to delegate to its internal format map; `LuceneWriter` returns itself for `"lucene"`.

2. **`LuceneDeleteExecutionEngine` naming** — Named to match the established convention where the Lucene plugin provides the single concrete implementation (same pattern as `LuceneCommitter`, `LuceneCommitterFactory`, `LuceneIndexingExecutionEngine`). All delete operations ultimately go through Lucene's IndexWriter for durable tracking, regardless of the indexing data format.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
@@ -1,0 +1,84 @@
+package org.opensearch.be.lucene.index;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.lucene.index.Term;
+import org.opensearch.be.lucene.LuceneDataFormat;
+import org.opensearch.index.engine.dataformat.DataFormat;
+import org.opensearch.index.engine.dataformat.DeleteExecutionEngine;
+import org.opensearch.index.engine.dataformat.DeleteInput;
+import org.opensearch.index.engine.dataformat.DeleteResult;
+import org.opensearch.index.engine.dataformat.Deleter;
+import org.opensearch.index.engine.dataformat.DeleterImpl;
+import org.opensearch.index.engine.dataformat.RefreshInput;
+import org.opensearch.index.engine.dataformat.RefreshResult;
+import org.opensearch.index.engine.dataformat.Writer;
+import org.opensearch.index.engine.exec.commit.Committer;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Lucene-based implementation of {@link DeleteExecutionEngine} that tracks per-generation
+ * deleters paired with their corresponding writers. Each deleter delegates document
+ * deletion to the underlying {@link LuceneWriter}.
+ *
+ * @opensearch.experimental
+ */
+public class LuceneDeleteExecutionEngine implements DeleteExecutionEngine<DataFormat> {
+
+    private static final Logger logger = LogManager.getLogger(LuceneDeleteExecutionEngine.class);
+
+    private final Map<Long, Deleter> generationToDeleterMap;
+    private final DataFormat dataFormat;
+    private final LuceneCommitter committer;
+
+    public LuceneDeleteExecutionEngine(DataFormat dataFormat, Committer committer) {
+        this.generationToDeleterMap = new ConcurrentHashMap<>();
+        this.dataFormat = dataFormat;
+        this.committer = (LuceneCommitter) committer;
+    }
+
+    @Override
+    public Deleter createDeleter(Writer<?> writer) {
+        LuceneWriter luceneWriter = writer.getWriterForFormat(LuceneDataFormat.LUCENE_FORMAT_NAME)
+            .map(w -> (LuceneWriter) w)
+            .orElseThrow(
+                () -> new IllegalArgumentException("Cannot create deleter: no Lucene writer found for generation=" + writer.generation())
+            );
+        Deleter deleter = new DeleterImpl<>(luceneWriter);
+        generationToDeleterMap.put(writer.generation(), deleter);
+        return deleter;
+    }
+
+    @Override
+    public RefreshResult refresh(RefreshInput refreshInput) throws IOException {
+        return null;
+    }
+
+    @Override
+    public DeleteResult deleteDocument(DeleteInput deleteInput) throws IOException {
+        Deleter deleter = generationToDeleterMap.get(deleteInput.generation());
+        if (deleter != null) {
+            return deleter.deleteDoc(deleteInput);
+        } else {
+            Term uid = new Term(deleteInput.fieldName(), deleteInput.value());
+            this.committer.getIndexWriter().deleteDocuments(uid);
+            return new DeleteResult.Success(1L, 1L, 1L);
+        }
+    }
+
+    @Override
+    public DataFormat getDataFormat() {
+        return this.dataFormat;
+    }
+
+    @Override
+    public void close() throws IOException {
+        for (Deleter deleter : generationToDeleterMap.values()) {
+            deleter.close();
+        }
+        generationToDeleterMap.clear();
+    }
+}
--- a/server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
+++ b/server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
@@ -34,6 +34,7 @@ import org.opensearch.core.index.shard.ShardId;
 import org.opensearch.index.VersionType;
 import org.opensearch.index.engine.dataformat.DataFormat;
 import org.opensearch.index.engine.dataformat.DataFormatRegistry;
+import org.opensearch.index.engine.dataformat.DeleteExecutionEngine;
 import org.opensearch.index.engine.dataformat.FileInfos;
 import org.opensearch.index.engine.dataformat.IndexingEngineConfig;
 import org.opensearch.index.engine.dataformat.IndexingExecutionEngine;
@@ -139,6 +140,7 @@ public class DataFormatAwareEngine implements Indexer {
     private final Store store;
 
     private final IndexingExecutionEngine indexingExecutionEngine;
+    private final DeleteExecutionEngine<?> deleteExecutionEngine;
     private final IndexingStrategyPlanner indexingStrategyPlanner;
     private final LockablePool<DefaultLockableHolder<Writer<?>>> writerPool;
     private final AtomicLong writerGenerationCounter;
@@ -276,6 +278,9 @@ public class DataFormatAwareEngine implements Indexer {
                 ),
                 registry.format(config().getIndexSettings().pluggableDataFormat())
             );
+
+            this.deleteExecutionEngine = registry.getDeleteExecutionEngine(committer);
+
             long maxGenFromCommit = 0L;
             try {
                 List<CatalogSnapshot> initSnapshots = committer.listCommittedSnapshots();
@@ -291,7 +296,9 @@ public class DataFormatAwareEngine implements Indexer {
             this.writerPool = new LockablePool<>(() -> {
                 long gen = writerGenerationCounter.incrementAndGet();
                 assert gen > 0 : "writer generation must be positive but was: " + gen;
-                return DefaultLockableHolder.of(new RowIdAwareWriter<>(indexingExecutionEngine.createWriter(new WriterConfig(gen))));
+                Writer<?> writer = indexingExecutionEngine.createWriter(new WriterConfig(gen));
+                deleteExecutionEngine.createDeleter(writer);
+                return DefaultLockableHolder.of(new RowIdAwareWriter<>(writer));
             }, LinkedList::new, Runtime.getRuntime().availableProcessors());
             // Create Reader managers
             // We will pass IndexStoreProvider to this, which would contain store
--- a/server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatRegistry.java
+++ b/server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatRegistry.java
@@ -14,11 +14,13 @@ import org.opensearch.common.CheckedFunction;
 import org.opensearch.common.annotation.ExperimentalApi;
 import org.opensearch.index.IndexSettings;
 import org.opensearch.index.engine.exec.EngineReaderManager;
+import org.opensearch.index.engine.exec.commit.Committer;
 import org.opensearch.index.store.FormatChecksumStrategy;
 import org.opensearch.plugins.PluginsService;
 import org.opensearch.plugins.SearchBackEndPlugin;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
@@ -278,5 +280,33 @@ public class DataFormatRegistry {
             );
         }
         return Map.of(readerManagerConfig.format(), readerManagerBuilders.get(readerManagerConfig.format()).apply(readerManagerConfig));
+    }
+
+    /**
+     * Returns the {@link DeleteExecutionEngine} by finding the single registered plugin that provides one.
+     * Iterates over all registered data format plugins and validates that exactly one returns a non-null
+     * result from {@link DataFormatPlugin#getDeleteExecutionEngine(Committer)}.
+     *
+     * @param committer the committer for durable delete tracking
+     * @return the delete execution engine
+     * @throws IllegalStateException if no plugin or multiple plugins provide a delete execution engine
+     */
+    public DeleteExecutionEngine<?> getDeleteExecutionEngine(Committer committer) {
+        List<DeleteExecutionEngine<?>> engines = new ArrayList<>();
+        for (DataFormatPlugin plugin : dataFormatPluginRegistry.values()) {
+            DeleteExecutionEngine<?> engine = plugin.getDeleteExecutionEngine(committer);
+            if (engine != null) {
+                engines.add(engine);
+            }
+        }
+        if (engines.size() > 1) {
+            throw new IllegalStateException(
+                "Multiple DataFormatPlugins provide a DeleteExecutionEngine, expected exactly one but found [" + engines.size() + "]"
+            );
+        }
+        if (engines.isEmpty()) {
+            throw new IllegalStateException("No DataFormatPlugin provides a DeleteExecutionEngine");
+        }
+        return engines.getFirst();
     }
 }
--- a/server/src/main/java/org/opensearch/index/engine/dataformat/Deleter.java
+++ b/server/src/main/java/org/opensearch/index/engine/dataformat/Deleter.java
@@ -1,0 +1,46 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.engine.dataformat;
+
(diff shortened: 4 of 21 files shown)


server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -220,9 +225,23 @@ public DataFormatAwareEngine(EngineConfig engineConfig) {
...
+
+            this.deleteExecutionEngine = registry.getDeleteEngine(registry.format(config().getIndexSettings().pluggableDataFormat()));
+            Queue<Deleter<?>> deleterQueue = new LinkedList<>();
             this.writerGenerationCounter = new AtomicLong(1L);// committer.getCommitStats().getGeneration());
             this.writerPool = new LockablePool<>(
-                () -> indexingExecutionEngine.createWriter(writerGenerationCounter.getAndIncrement()),
+                () -> {
+                    long gen = writerGenerationCounter.getAndIncrement();
+                    Writer<?> writer = indexingExecutionEngine.createWriter(gen);
+                    deleterQueue.add(deleteExecutionEngine.createDeleter(writer, gen));
```
RS146BIJAY: Can this be tracked in deleteExecutionEngine itself?

thorkous: It can be but we would still need to provide all the writers in deleteExecutionEngine when creating deleters. I wanted to provide one writer at a time while creating deleter .

server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatRegistry.java
```diff
@@ -43,21 +43,26 @@ public class DataFormatRegistry {
...
+     * and {@link SearchBackEndPlugin} implementations from the given {@link PluginsService}.
+     * Registers each data format with its indexing plugin, delete plugin, and reader manager factory.
      *
      * @param pluginsService the plugins service used to discover data format plugins and search back-end plugins
      * @throws IllegalArgumentException if a data format is registered by more than one plugin
      * @throws IllegalStateException if the set of formats with indexing plugins does not match the set with reader managers
      */
     public DataFormatRegistry(PluginsService pluginsService) {
         Map<DataFormat, DataFormatPlugin> dataFormatPlugiRegistry = new HashMap<>();
+        Map<DataFormat, DeleteDataFormatPlugin> deleteDataFormatPluginRegistry = new HashMap<>();
```
RS146BIJAY: Why do we need to keep delete format separately in map? DeleteFormat needs to be common across all format

thorkous: deleteDataFormatPluginRegistry provides DeleteExecutionEngine per format which would be different for each format, I can merge into one but then refresh/flush for all format need to follow same path which can create multiple if else blocks.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
RS146BIJAY: DataFormatAwareEngine should not be aware about any internals of DeleteExecutionEngine. deleterQueue is meant to be track writer and deleter mapping. DeleteExecutionEngine should be passed the Writer when creating a deleter where it should keep track of mapping Deleter and Writer.

thorkous: > deleterQueue is meant to be track writer and deleter mapping.

No, deleterQueue is used because writer are getting created one by one and we can't create deleterPool in the same loop.
DeleterQueue is not meant for tracking writer. All the deleter and writer mapping will be present in deleteExecutionEngine at the time of createDeleter.

server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatRegistry.java
```diff
@@ -78,6 +82,11 @@ public DataFormatRegistry(PluginsService pluginsService) {
                 }
             }
         }
+        List<DeleteDataFormatPlugin> deletePlugins = pluginsService.filterPlugins(DeleteDataFormatPlugin.class);
+        if (deletePlugins.size() > 1) {
+            throw new IllegalArgumentException("Only one DeleteDataFormatPlugin allowed, found " + deletePlugins.size());
+        }
+        this.deleteDataFormatPlugin = deletePlugins.isEmpty() ? null : deletePlugins.getFirst();
```
RS146BIJAY: How are we validating if deleteFormatPlugin gets set for parquet standalone plugin?

server/src/main/java/org/opensearch/index/engine/dataformat/DeleteDataFormatPlugin.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+@ExperimentalApi
+public interface DeleteDataFormatPlugin extends DataFormatPlugin{
+
+    /**
+     * Creates the delete execution engine for the data format. This should be instantiated per shard.
+     *
+     * @return the delete execution engine instance
+     */
+    DeleteExecutionEngine<?, ?> deleteEngine();
```
mgodwan: This can be a part of DataFormatPlugin only? Default impl can retuern an empty Optional?

server/src/main/java/org/opensearch/index/engine/dataformat/Deleter.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Handles document deletion for a specific data format. Each deleter is paired with a
+ * {@link Writer} and shares its generation. Implements {@link Lockable} for thread-safe
+ * pooling via {@link org.opensearch.common.queue.LockablePool}.
+ *
+ * @param <P> the document input type
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public interface Deleter<P extends DocumentInput<?>> extends Closeable, Lockable {
```
mgodwan: What does documentinput contain for a delete call? Shouldn't the input be always same?

server/src/main/java/org/opensearch/index/engine/dataformat/DeleteDataFormatPlugin.java
thorkous: Done.

server/src/main/java/org/opensearch/index/engine/dataformat/Deleter.java
thorkous: Updated.

server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatRegistry.java
thorkous: deleteFormatPlugin will get initialized for all the formats.

server/src/main/java/org/opensearch/index/engine/dataformat/DeleteDataFormatPlugin.java
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Plugin interface for providing custom delete data format implementations.
+ * Plugins implement this to register their delete data format
+ * with the {@link DataFormatRegistry} during node bootstrap.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public interface DeleteDataFormatPlugin extends DataFormatPlugin{
```
mgodwan: Is this still needed?

thorkous: Yes, it is required because we wanted delete as a separate entity and we wanted DeleteExecutionEngine to be discovered independently from the indexing engine.

Bukhtawar: Sorry why should deletes be a separate execution engine?

thorkous: I have updated the code, I have deleted DeleteDataFormatPlugin and we are using DataFormatPlugin to return deleteExecutionEngine.

gradle.properties
```diff
@@ -32,3 +32,6 @@ systemProp.jdk.tls.client.protocols=TLSv1.2,TLSv1.3
 
 # jvm args for faster test execution by default
 systemProp.tests.jvm.argline=-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m
+
+# Enable sandbox modules (parquet-data-format, delete-data-format, etc.)
```
RS146BIJAY: Elaborate why this is needed.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -227,6 +213,8 @@ public DataFormatAwareEngine(EngineConfig engineConfig) {
 
             // 1. Create Committer (uses translogPath for safe bootstrap trimming)
             this.committer = engineConfig.getCommitterFactory().getCommitter(new CommitterConfig(engineConfig));
+            this.deleteExecutionEngine = engineConfig.getDeleteExecutionEngineFactory()
```
RS146BIJAY: Why do we need a factory for this? Why can't we just return a DeleterExecutionEngine?

server/src/main/java/org/opensearch/index/engine/EngineConfigFactory.java
```diff
@@ -129,6 +134,7 @@ public EngineConfigFactory(PluginsService pluginsService, IndexSettings idxSetti
             enginePlugin.getAdditionalCodecs(idxSettings).ifPresent(codecRegistries::add);
 
             enginePlugin.getCommitterFactory(idxSettings).ifPresent(committerFactories::add);
+            enginePlugin.getDeleteExecutionEngineFactory(idxSettings).ifPresent(deleteExecutionEngineFactories::add);
```
RS146BIJAY: As discussed earlier, can we just get the DeleteExecutionEngine directly here.

gradle.properties
thorkous: I have removed this change.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: I wanted to pass the committer through the `DeleteExecutionEngine` constructor so it could use the parent index writer to delete documents from the parent index. However, the committer is only available after initialization, which happens at line 215. If I simply returned DeleteExecutionEngine directly, I wouldn’t have been able to provide the committer at construction time.

I preferred keeping the committer as a constructor dependency rather than using a setter because it’s required for parent deletes and should be treated as a mandatory dependency.

server/src/main/java/org/opensearch/index/engine/EngineConfigFactory.java
thorkous: Replied above

server/src/main/java/org/opensearch/plugins/EnginePlugin.java
```diff
@@ -132,4 +133,20 @@ default Optional<TranslogDeletionPolicyFactory> getCustomTranslogDeletionPolicyF
...
+     * whether or not to provide a {@link DeleteExecutionEngineFactory} for the given index. A plugin that does not provide a
+     * delete execution engine should return {@link Optional#empty()}.
+     * <p>
+     * The returned factory will be called with the shard's {@link org.opensearch.index.engine.exec.commit.Committer} once it
+     * is available, allowing the delete engine to be constructed with all required dependencies.
+     *
+     * @param indexSettings index settings to detect whether a delete execution engine factory should be provided.
+     * @return an optional delete execution engine factory
+     */
+    @ExperimentalApi
```
RS146BIJAY: Hey this need to happen from DataFormatPlugin instead of EnginePlugin.

thorkous: Kept it on `EnginePlugin` intentionally — unlike `IndexingExecutionEngine` which is format-specific, `DeleteExecutionEngine` is always `Lucene-backed` (deletes go through Lucene's IndexWriter regardless of indexing format). If placed on DataFormatPlugin, CompositeDataFormatPlugin and ParquetDataFormatPlugin can't provide it since `LuceneDeleteExecutionEngineFactory` lives in a different plugin/classloader. Same reasoning as `CommitterFactory` being on `EnginePlugin`.

RS146BIJAY: I think EnginePlugin manages scopes at Engine level. Committer is an Engine level abstraction. So this went in Engine plugin. Deleter are data level abstraction, so this needs to go to DataFormatPlugin.

thorkous: Updated the PR.

thorkous:
Build failure not because of my changes. 
> Task :test:fixtures:gcs-fixture:composeBuild FAILED

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private final Map<Long, Deleter> generationToDeleterMap;
+    private final DataFormat dataFormat;
+    private final LuceneCommitter committer;
+
+    public LuceneDeleteExecutionEngine(DataFormat dataFormat, Committer committer) {
+        this.generationToDeleterMap = new ConcurrentHashMap<>();
+        this.dataFormat = dataFormat;
+        this.committer = (LuceneCommitter) committer;
+    }
```
Bukhtawar, line 41: What is specific to a data format here?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -0,0 +1,83 @@
...
+            );
+        Deleter deleter = new DeleterImpl(luceneWriter);
+        generationToDeleterMap.put(writer.generation(), deleter);
+        return deleter;
+    }
+
+    @Override
+    public RefreshResult refresh(RefreshInput refreshInput) throws IOException {
+        return null;
+    }
```
Bukhtawar, line 58: Is this for a future intefgration

thorkous: Yes, this is a placeholder for future integration.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
thorkous: It's not required. I will remove this.

RS146BIJAY: So idea here is similar to how committer is implemented. Deleter is currently backed via Lucene's IndexWriter. So only LucenePlugin will only return a DeleteExecutionEngine implementation (LuceneDeleteExecutionEngine). This is then loaded inside DataFormatRegistry via dataFormatPluginRegistry (which gives view for loaded plugins). Later if we want to implement deleter via some other data format say via a DeleteVector, we can make the new format return the deleter implementation. This is almost parallel to how commiter is implemented via a Lucene IndexWriter: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/EngineConfigFactory.java#L131. Let me know if this makes sense.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/DeleterImpl.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Lucene-based implementation of {@link Deleter} that delegates document deletion
+ * to the paired {@link LuceneWriter}. Each instance shares the generation number
+ * of its associated writer. Constructs the {@link Term} uid from the field name
+ * and value provided in the {@link DeleteInput}.
+ *
+ * @opensearch.experimental
+ */
+public class DeleterImpl implements Deleter {
```
Bukhtawar: Does this need to be tied to Lucene?

thorkous: Initially the plan was to keep DeleterImpl at core but since right now its backed by lucene, we are keeping it at analytics-backend-lucene.

server/src/main/java/org/opensearch/index/engine/dataformat/DeleterImpl.java
```diff
@@ -0,0 +1,65 @@
...
+ * Lucene-based implementation of {@link Deleter} that delegates document deletion
+ * to the paired {@link Writer}. Each instance shares the generation number
+ * of its associated writer. Constructs the {@link Term} uid from the field name
+ * and value provided in the {@link DeleteInput}.
+ *
+ * @opensearch.experimental
+ */
+public class DeleterImpl<T extends Writer<?>> implements Deleter {
+
+    private final Writer<?> luceneWriter;
```
Bukhtawar: nit: just rename to `writer`

thorkous: Updated.

RS146BIJAY (approved):

Bukhtawar (approved):
LGTM

PR #21318, opened 2026-04-22
Add cancellation checkpoints in field data loading and aggregation paths

kaushalmahi12:
### Description

Cancelled search tasks continue consuming CPU indefinitely because key code paths in field data loading and global ordinals building have **zero cancellation checkpoints**. This causes "zombie threads" — tasks cancelled but consuming 100% CPU for days/weeks.

#### Root Cause

The call chain `AggregationPhase.preProcess → AggregatorFactories.createTopLevelAggregators → TermsAggregatorFactory → GlobalOrdinalsBuilder.build → AbstractIndexOrdinalsFieldData.loadGlobalDirect → PagedBytesIndexFieldData.loadDirect → OrdinalsBuilder.addDoc → GrowableWriter.get/set` has no cancellation checkpoints.

While `ExitableDirectoryReader` already wraps `TermsEnum.next()` with cancellation checks, it does **not** wrap `PostingsEnum`. The inner loop in `PagedBytesIndexFieldData.loadDirect()` iterates `docsEnum.nextDoc()` → `OrdinalsBuilder.addDoc()` in a tight CPU loop with no way to interrupt it once a task is cancelled.

#### Changes

This PR adds cancellation checks at three levels:

**1. `ExitablePostingsEnum` in `ExitableDirectoryReader`** (the critical fix)
- New `ExitablePostingsEnum` wraps `PostingsEnum.nextDoc()` and `advance()` with sampled cancellation checks (every 8191 calls, matching the existing `ExitableIntersectVisitor` pattern)
- Wired from `ExitableTermsEnum.postings()` so any code path that iterates postings through the exitableReader gets coverage automatically
- This directly covers the tight CPU loop: `docsEnum.nextDoc()` → `OrdinalsBuilder.addDoc()` → `GrowableWriter.get/set`
- The wrapping survives through `FrequencyFilter` and `RamAccountingTermsEnum` because `FilteredTermsEnum.postings()` delegates to the underlying `tenum`

**2. `GlobalOrdinalsBuilder.build()` — per-segment cancellation check**
- New overloaded `build()` accepts a `Runnable cancellationCheck`, called between segment iterations
- Old method delegates with `() -> {}` (no-op) — fully backward compatible
- `AbstractIndexOrdinalsFieldData.loadGlobalDirect(DirectoryReader, Runnable)` overload threads the check through

**3. `AggregatorFactories.createTopLevelAggregators()` — per-factory cancellation check**
- Checks `searchContext.isCancelled()` before each `factories[i].create()` call
- Allows deeply nested aggregation trees (4-8 levels observed in production) to be interrupted between levels

#### Why `PostingsEnum` is the only Lucene construct that needed wrapping

Other `LeafReader` iteration constructs (`NumericDocValues`, `SortedSetDocValues`, etc.) are used during aggregation **collection**, which is driven by the query scorer that already has cancellation checks. `PostingsEnum` is the only one used in an unbounded tight loop (`loadDirect`) that runs outside the query scorer's cancellation-aware iteration.

#### Backward Compatibility

- Zero interface changes (`IndexFieldData`, `IndexOrdinalsFieldData`, `IndexFieldDataCache`)
- Zero method signature changes on existing methods
- All new methods are overloads with backward-compatible defaults
- Warmers, recovery, and non-search paths are unaffected
- Plugins/modules that implement `IndexFieldData` don't need any changes
- When `lowLevelCancellation` is disabled, `ExitableDirectoryReader` doesn't wrap the reader, so there is zero overhead

#### Cache Safety

`IndicesFieldDataCache` uses `computeIfAbsent` — if `loadDirect()` or `loadGlobalDirect()` throws `TaskCancelledException`, the entry is simply not cached. No partial/corrupt cache entries.

### Testing

- `SearchCancellationTests.testExitablePostingsEnum` — verifies `ExitablePostingsEnum` throws `TaskCancelledException` on `nextDoc()` and `advance()` when cancelled
- `SearchCancellationTests.testExitablePostingsEnumNoOpWhenCancellationDisabled` — verifies no wrapping overhead when cancellation is disabled
- `GlobalOrdinalsBuilderTests` — verifies per-segment cancellation check fires, delayed cancellation works, and original method backward compat
- `AggregatorFactoriesCancellationTests` — verifies `createTopLevelAggregators` throws when `isCancelled()` is true

### Check List
- [x] API changes companion pull request - NA

For more information on following Developer Certificate of Origin and Signing Commits, please see the [Contribution Guidelines](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


kkhatua:
@jainankitk , @bowenlan-amzn , @sgup432 

Could you please help review this PR?

server/src/main/java/org/opensearch/search/internal/ExitableDirectoryReader.java
```diff
@@ -210,6 +211,85 @@ public BytesRef next() throws IOException {
             checkAndThrowWithSampling();
             return in.next();
         }
+
+        @Override
+        public PostingsEnum postings(PostingsEnum reuse, int flags) throws IOException {
```
bowenlan-amzn, line 216: Nice!
This should help with the text field path for building global ordinals from postings/inverted index.

I just learn this `ExitableDirectoryReader` from your change here. It looks like we might be able to extend the same pattern for the keyword field path that relies on DocValues, something like:

```
public SortedSetDocValues getSortedSetDocValues(String field) throws IOException {
            SortedSetDocValues values = in.getSortedSetDocValues(field);
            if (values == null) {
                return null;
            }
            return queryCancellation.isEnabled() ? new ExitableSortedSetDocValues(values, queryCancellation) : values;
        }
```

Please see if this makes sense.

Context:
SortedSetDocValues is used for building global ordinal here
https://github.com/opensearch-project/OpenSearch/blob/aca2e9de7daaadee77c33e21ff5215c3f1e8600f/server/src/main/java/org/opensearch/index/fielddata/ordinals/GlobalOrdinalsBuilder.java#L78-L83

kaushalmahi12: Yes, It will definitely help with high cardinality fields.

server/src/main/java/org/opensearch/index/fielddata/plain/AbstractIndexOrdinalsFieldData.java
```diff
@@ -169,6 +169,13 @@ public IndexOrdinalsFieldData loadGlobalDirect(DirectoryReader indexReader) thro
         return GlobalOrdinalsBuilder.build(indexReader, this, breakerService, logger, scriptFunction);
     }
 
+    /**
+     * Load global ordinals directly with periodic cancellation checks.
+     */
+    public IndexOrdinalsFieldData loadGlobalDirect(DirectoryReader indexReader, Runnable cancellationCheck) throws Exception {
+        return GlobalOrdinalsBuilder.build(indexReader, this, breakerService, logger, scriptFunction, cancellationCheck);
+    }
```
jainankitk: Where is `loadGlobalDirect(DirectoryReader, Runnable)` invoked from?

server/src/main/java/org/opensearch/search/internal/ExitableDirectoryReader.java
jainankitk: Can you address this gap when global ordinals are built via `SortedSetDocValues`
iteration, not `PostingsEnum` ?

jainankitk (commented):
Seems that gradle check is failing due to spotless issue. Can you address that as well along with review feedback?

server/src/main/java/org/opensearch/index/fielddata/plain/AbstractIndexOrdinalsFieldData.java
kaushalmahi12: Currently it is not called from anywhere but it will provide us the way to inject the cancellationCheck which GlobalOrdinalsBuilder can use to break out of IO loop. I think it makes sense to remove this now, since we are anyway have lower level checks in concrete lucene data structures

server/src/main/java/org/opensearch/search/internal/ExitableDirectoryReader.java
kaushalmahi12: Added it.

kaushalmahi12:
The gradle-check failure appears to be from pre-existing flaky tests tracked in:
- #21373 (SearchRestCancellationIT)
- #21385 (IndexFieldDataServiceTests)  
- #21034 (SubAggregationIT)

All tests related to this change pass locally. Requesting a re-run.

server/src/main/java/org/opensearch/index/fielddata/ordinals/GlobalOrdinalsBuilder.java
```diff
@@ -70,17 +73,41 @@ public static IndexOrdinalsFieldData build(
...
         }
-        final OrdinalMap ordinalMap = OrdinalMap.build(null, subs, PackedInts.DEFAULT);
+        // Wrap subs with cancellation-aware termsEnum for OrdinalMap.build().
+        // OrdinalMap.build() only calls termsEnum() and getValueCount() on subs,
+        // so this wrapper is safe. We don't wrap atomicFD (stored for later use)
+        // to preserve SingletonSortedSetDocValues type for unwrapSingleton().
+        final SortedSetDocValues[] cancellableSubs = new SortedSetDocValues[subs.length];
+        for (int i = 0; i < subs.length; ++i) {
+            cancellableSubs[i] = new CancellableTermsSortedSetDocValues(subs[i], cancellationCheck);
+        }
```
jainankitk: Does it make sense to wrap into `cancellableSubs` during `subs` initialization itself?

server/src/main/java/org/opensearch/index/fielddata/ordinals/GlobalOrdinalsBuilder.java
```diff
@@ -140,4 +167,84 @@ public void close() {}
...
+        CancellableTermsSortedSetDocValues(SortedSetDocValues in, Runnable cancellationCheck) {
+            this.in = in;
+            this.cancellationCheck = cancellationCheck;
+        }
+
+        @Override
+        public TermsEnum termsEnum() throws IOException {
+            TermsEnum te = in.termsEnum();
+            return new FilterLeafReader.FilterTermsEnum(te) {
+                private static final int CHECK_INTERVAL = (1 << 10) - 1; // 1023
```
jainankitk, line 188: I am wondering if `1023` is the right value, or should be likely higher

jainankitk (approved):
Minor comments, otherwise looks good to me!

server/src/main/java/org/opensearch/index/fielddata/ordinals/GlobalOrdinalsBuilder.java
kaushalmahi12: It is the common pattern in other exitable equivalent in OPenSearch so went with that.

server/src/main/java/org/opensearch/index/fielddata/ordinals/GlobalOrdinalsBuilder.java
kaushalmahi12: That is a fair call and I don't see a concern moving this inside the above loop.

PR #21326, opened 2026-04-23
Use atomic writes in FsBlobContainer

andrross:
I believe non-atomic writes are the causes of test failures [like this][1]. We could fix the logic to be resilient to non-atomic writes, but I don't believe the code will ever see a non-atomic write in real life. I think it's better just to make the FS repo always use atomic writes.

[1]: https://build.ci.opensearch.org/job/gradle-check/74598/testReport/org.opensearch.indices.replication/WarmIndexSegmentReplicationIT/testScrollCreatedOnReplica/

### Check List


andrross:
Closing as I haven't seen this test fail again

PR #21328, opened 2026-04-23
Add sandbox QA module for analytics engine REST integration tests

sandeshkr419:
### Description
Introduces `sandbox/qa/analytics-engine-rest`, a QA module for end-to-end REST integration tests covering the DataFusion and Parquet query paths.

Test classes:

- ParquetDataFusionIT:  parquet index creation, settings validation, bulk ingestion, basic search
- DslClickBenchIT: auto-discovers and runs ClickBench DSL queries via _search → DataFusion
- PplClickBenchI: auto-discovers and runs ClickBench PPL queries via /_analytics/ppl → test-ppl-frontend → DataFusion

Generic test runner infrastructure:

- `Dataset / DatasetProvisioner / DatasetQueryRunner`: dataset-agnostic. Add a new dataset by dropping mapping.json, bulk.json, and query files under resources/datasets/{name}/
- AnalyticsRestTestCase: base class for all analytics REST tests
- Auto-provisioned cluster via integTest with all required plugins and feature flags

Other changes:

- Fix TestPPLTransportAction to fork execution to SEARCH thread pool (was blocking on transport thread)
- ClickBench resources placed in resources/datasets/clickbench/ structure

### Related Issues

### Check List


sandbox/qa/analytics-engine-rest/build.gradle
```diff
@@ -0,0 +1,68 @@
...
+java { sourceCompatibility = JavaVersion.toVersion(25); targetCompatibility = JavaVersion.toVersion(25) }
+
+repositories {
+  maven {
+    name = 'OpenSearch Snapshots'
+    url = 'https://ci.opensearch.org/ci/dbc/snapshots/maven/'
+  }
+}
+
+dependencies {
```
mch2, line 23: we'll need to wire in a ppl front-end plugin - we can either wire in the [test plugin](https://github.com/mch2/OpenSearch/tree/ani/sandbox/qa/test-ppl-frontend) from our feat branch, or sql plugin directly.  i'm thinking to use our test plugin and include it here otherwise we'll hvae a weird circular dependency on sql plugin for core sandbox builds.

The integration in that test plugin is largely the same as what sql plugin does using the unified api.

sandeshkr419: The QA module's `restTest` task runs against an external cluster, so the SQL plugin is resolved from Maven at runtime, not at build time, so there's no circular dependency in the Gradle build graph. 

I removed wiring in `test-ppl-frontend` to avoid duplicating the SQL plugin's PPL integration and avoid extra maintenance burden.

Can include this in future if the need arises, what do you think?

mch2: hm, what i don't love about that is that a something in sql repo could theoretically break our builds here if we are depending on snapshot versions, which to me is a deal breaker.

sandeshkr419: Ack, I have added the test plugin, basically cloned it up from your branch and added up UTs for now.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/ClickBenchTestFixture.java
```diff
@@ -0,0 +1,137 @@
...
+ * Static helper that provisions the ClickBench test dataset into a live cluster
+ * with parquet as the primary data format.
+ * <p>
+ * Stateless utility — all methods are static and take a {@link RestClient} parameter.
+ * This makes it composable: any test class can use it without inheritance.
+ * <p>
+ * The ClickBench dataset uses the standard ClickBench schema (103 fields) with
+ * 100 rows stored as bulk JSON for OpenSearch indexing.
+ */
+public final class ClickBenchTestFixture {
```
vinaykpud: Can we rename this file better? `ClickBencTestHelper`? so that we know everything in this is helper methods or classes

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/DataFusionRestTestCase.java
```diff
@@ -0,0 +1,77 @@
...
+ * <p>
+ * Handles cluster-level concerns: preserving cluster/indices across test methods,
+ * loading classpath resources, JSON escaping, and common assertion helpers.
+ * <p>
+ * Test data provisioning (e.g. ClickBench dataset) is handled separately by
+ * {@link ClickBenchTestFixture} to keep cluster config orthogonal to test data.
+ * Future test classes (LuceneScanRestIT, etc.) can extend this without needing
+ * ClickBench data.
+ */
+public abstract class DataFusionRestTestCase extends OpenSearchRestTestCase {
```
vinaykpud: can it be named as `AnalyticsRestTestCase` ? lets not tie it to DF

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/ParquetDataFusionIT.java
```diff
@@ -0,0 +1,123 @@
...
+ * <p>
+ * Validates that a composite index with parquet as primary data format can be
+ * created, documents can be ingested, and the index settings are correctly persisted.
+ * <p>
+ * Requires plugins: analytics-engine, analytics-backend-datafusion, analytics-backend-lucene,
+ * dsl-query-executor, composite-engine, parquet-data-format.
+ * <p>
+ * Requires feature flag: {@code opensearch.experimental.feature.pluggable.dataformat.enabled=true}
+ */
+public class ParquetDataFusionIT extends DataFusionRestTestCase {
```
vinaykpud: why do we need this? we already have other dsl and ppl for clickbench test files right?

vinaykpud:
I think we are coupling the QA test with the dataset. I am imagining given any dataset, query set, response set QA should be able to run it. This should be possible with a generic test runner. 

- Auto-discover queries from `resources/{datasetName}/{queryLanguage}/`
- Provision dataset from resources/{datasetName}/mapping.json + bulk.json
- Asserts response against resources/{datasetName}/{queryLanguage}/expected/

```
  resources/
    datasets/
      clickbench/
        mapping.json
        bulk.json
        dsl/q1.json ... q43.json
        dsl/expected/q1.json ... q43.json 
        ppl/q1.ppl ... q43.ppl
        ppl/expected/q1.json ... q43.json
        sql/q1.sql ... q43.sql
        sql/expected/q1.json ... q43.json
```

For the responses, we may need to generate them first, but with a smaller data set size it will help us validate the correctness of any type of query easily. 

Also, right now it requires us to manually provision the cluster. Can we configure a single-node or multi-node cluster and auto-provision it before the tests begin?

vinaykpud:
Also, I noticed we already have a similar set of ClickBench tests under `sandbox/plugins/analytics-engine/src/internalClusterTest/ (ClickBenchUnifiedPipelineIT)` with its own copy of q1–q43 PPL files and index mapping. With this PR, we are adding a second parallel copy under `sandbox/qa/`. 

I understand the two serve different purposes ie IT runs at the transport level for end-to-end validation, while QA runs at the REST / cluster-wide level. 

But the underlying dataset is same. Queries, and expected responses are the same but may be in different format.

A generic runner backed by shared resources would let both layers consume the same test definitions, instead of duplicating them across modules.

sandbox/qa/test-ppl-frontend/src/main/java/org/opensearch/ppl/planner/PushDownPlanner.java
```diff
@@ -0,0 +1,95 @@
...
+ * {@code LogicalTableScan} with an {@code OpenSearchBoundaryTableScan}
+ * carrying the scan as its initial logical fragment.
+ *
+ * <p><b>Phase 2 (HepPlanner):</b> Runs absorb rules to push supported
+ * operators into the boundary node's logical fragment. Unsupported operators
+ * (e.g., projects containing functions not in the back-end's
+ * {@link SqlOperatorTable}) remain above the boundary node and execute
+ * in-process via Janino bytecode.
+ */
+public class PushDownPlanner {
```
mch2: we shouldnt need the PushDownPlanner anymore as there are no longer any pushdown rules - we can remove this and BoundaryTableScanShuttle etc.  The sql plugin will send us the full RelNode

sandbox/qa/analytics-engine-rest/build.gradle
sandeshkr419: Removed it back since it was added by your changes: https://github.com/opensearch-project/OpenSearch/pull/21356

sandbox/qa/test-ppl-frontend/src/main/java/org/opensearch/ppl/planner/PushDownPlanner.java
sandeshkr419: Outdated.

sandbox/plugins/test-ppl-frontend/src/main/java/org/opensearch/ppl/action/TestPPLTransportAction.java
```diff
@@ -18,45 +18,62 @@
...
-            listener.onResponse(response);
-        } catch (Exception e) {
-            logger.error("[UNIFIED_PPL] execution failed", e);
-            listener.onFailure(e);
-        }
+        // Fork to SEARCH thread pool — DefaultPlanExecutor.execute() blocks on a future
+        // internally, which is forbidden on the transport thread.
+        threadPool.executor(ThreadPool.Names.SEARCH).execute(() -> {
+            try {
+                PPLResponse response = unifiedQueryService.execute(request.getPplText());
```
mch2, line 72: lets update `UnifiedQueryService` to take a listener now that defaultPlanexecutor does. not critical as this is for ITs, but you wont need that threadpool fork.

mch2 (approved):

vinaykpud (approved):

sandbox/plugins/test-ppl-frontend/src/main/java/org/opensearch/ppl/action/TestPPLTransportAction.java
sandeshkr419: Added a todo for now to unblock the PR for now.

PR #21143, opened 2026-04-06
Refactor WLM settings to use Setting objects and rename field

dzane17:
**Description**

This PR refactors the workload group settings introduced in #20536. The `search_settings` field (added in 3.6 as `@ExperimentalApi`) is replaced with a `settings` field backed by the OpenSearch `Settings` framework instead of a raw `Map<String, String>`.

Changes:
- Rename the JSON field from `"search_settings"` to `"settings"`
- Rename the timeout key from `"timeout"` to `"search.default_search_timeout"` to match the cluster setting name
- Replace `Map<String, String>` with `Settings` objects internally
- Mark `WorkloadGroupSearchSettings` and `WorkloadGroup.getSettings()` as `@ExperimentalApi`

**Backward compatibility (3.6 → 3.7 rolling upgrade)**

- Settings configured on 3.6 workload groups do not carry forward to 3.7. Users must re-apply settings using the new `"settings"` field and `"search.default_search_timeout"` key after upgrade. This is permitted because `search_settings` was marked `@ExperimentalApi`.
- Serialization is not broken during rolling upgrade. A 3.7 node sends and receives empty/dummy values when communicating with 3.6 nodes to prevent deserialization errors. The wire format remains structurally valid in both directions.

**Testing**

**Create group with settings**
```bash
curl -s -X PUT localhost:9200/_wlm/workload_group -H 'Content-Type: application/json' -d '{
  "name": "test_group",
  "resiliency_mode": "enforced",
  "resource_limits": {"cpu": 0.3, "memory": 0.3},
  "settings": {"search.default_search_timeout": "30s"}
}'

{
    "_id": "Tt2a5BZhR7yvEyLmkBiLOg",
    "name": "test_group",
    "resiliency_mode": "enforced",
    "resource_limits": { "cpu": 0.3, "memory": 0.3 },
    "settings": { "search.default_search_timeout": "30s" },
    "updated_at": 1775177784440
}
```

**Update settings**
```bash
curl -s -X PUT localhost:9200/_wlm/workload_group/test_group -H 'Content-Type: application/json' -d '{
  "settings": {"search.default_search_timeout": "1m"}
}'

{
    "_id": "Tt2a5BZhR7yvEyLmkBiLOg",
    "name": "test_group",
    "resiliency_mode": "enforced",
    "resource_limits": { "cpu": 0.3, "memory": 0.3 },
    "settings": { "search.default_search_timeout": "1m" },
    "updated_at": 1775177797926
}
```

**Create group without settings**
```bash
[...]

PR #21343, opened 2026-04-24
Add balance.disk_usage and balance.mode settings for dimensionally-consistent shard rebalancing (#15520)

### LINKED ISSUE #15520: [Feature Request] Introduce a cluster allocation setting to account for disk usage

Sometimes, In a cluster where there are indices of varied sizes, it unbalances the disks of the nodes to a great extent as currently the allocation strategy is primarily based on indices and shard counts and it doesn't account for the shard sizes while rebalancing. 

It would be great to have a similar setting just like we have `cluster.routing.allocation.balance.index` and `cluster.routing.allocation.balance.shard`. Elasticsearch already has `cluster.routing.allocation.balance.disk_usage` cluster setting which accounts for disk usage stats while performing shard balancing. 

### Related component

ShardManagement:Placement

N/A

### Additional context

_No response_

AndreKurait:
## Summary

Resolves #15520. Adds two opt-in, default-preserving knobs to the shard allocator's weight function, plus a precision fix to the disk-usage term.

| Setting | Type | Default | What it does |
|---|---|---|---|
| `cluster.routing.allocation.balance.disk_usage` | float | `0.0f` | Adds a disk-byte balance term to the weight function. |
| `cluster.routing.allocation.balance.mode` | enum `count` \| `ratio` | `count` | Selects the scale on which the three balance terms are expressed. |

Both are dynamic, node-scope. Both defaults reproduce pre-PR behavior bit-for-bit (with one exception: the disk term now subtracts in `double` precision instead of `float` — a bug fix that only matters at very large cluster sizes; see "Precision fix" below).

## Background

`BalancedShardsAllocator.WeightFunction` scores each (node, index) pair so the allocator can pick moves that reduce imbalance. Before this PR it looked like:

```
weight = shardBalance * (node.numShards()      - avgShardsPerNode)
       + indexBalance * (node.numShards(index) - avgShardsPerNode(index))
```

### Problem 1 — no disk signal

Two nodes with identical shard counts can carry wildly different byte loads when shard sizes are heterogeneous. The allocator had no signal to correct this.

### Problem 2 — the three terms are not on the same scale

The existing shard and per-index terms are **raw shard-count deltas**. A node with 5 shards above average contributes `5.0`. A naively-added disk term expressed as a ratio (`(bytes - avgBytes) / avgBytes`) has magnitude `O(1)`. Setting `disk_usage ≈ shard` does not mean "equal influence" — the disk contribution is swamped by two orders of magnitude on any realistic cluster.

### 1. `balance.disk_usage` — new disk-byte balance factor

Adds a third term to the weight function:

```
+ diskUsageBalance * (node.diskUsageInBytes() - avgDiskUsage) / avgDiskUsage
```

When `avgDiskUsage == 0` (empty cluster) the disk term contributes zero. When the factor is `0.0f` (default), the term short-circuits and the allocator does not bother tracking per-shard bytes at all (see `trackDiskUsage` on `LocalShardsBalancer`). Upgrades are zero-overhead.

### 2. `balance.mode` — `count` (default) vs `ratio`

Resolves the units mismatch from Problem 2 opt-in, without changing any existing cluster's behavior.

**`count` mode (default):**
```
weight_shard(n)   = shardBalance    * (n.numShards    - avg)
weight_index(n,i) = indexBalance    * (n.numShards(i) - avg(i))
weight_disk(n)    = diskUsageBalance * (n.bytes       - avgBytes) / max(1, avgBytes)
```
Identical to pre-PR semantics. `threshold` (default `1.0`) keeps its shard-count-delta interpretation ("at least one shard of imbalance").

**`ratio` mode:**
```
weight_shard(n)   = shardBalance    * (n.numShards    - avg)      / max(1, avg)
weight_index(n,i) = indexBalance    * (n.numShards(i) - avg(i))   / max(1, avg(i))
weight_disk(n)    = diskUsageBalance * (n.bytes       - avgBytes) / max(1, avgBytes)
```
All three terms are relative deviations from the per-axis cluster average. The three balance factors operate on the **same dimensionless scale** and can be tuned directly relative to each other.

In ratio mode, `threshold` is interpreted as a relative-deviation fraction (e.g. `0.1` = "at least 10% imbalance"). Operators enabling ratio mode will typically also lower `threshold` from its default of `1.0`. This is documented on `BALANCE_MODE_SETTING`.

**Caveat — small-index amplification.** Ratio mode amplifies the weight of indices with few shards: a single misplaced shard of a 3-shard index is a 33% deviation, while the same misplacement on a 30-shard index is a 3% deviation. This is semantically correct (small indices genuinely are more imbalanced per misplaced shard) but changes prioritization relative to count mode. Documented on the setting.

### 3. Precision fix for the disk term

`ShardsBalancer.avgDiskUsageInBytesPerNode()` previously returned `float`. `float` has ~7 significant digits, so on clusters with total disk usage above ~10 TB the `(long bytes - float avgBytes)` subtraction lost sub-MB deltas entirely **before** the divide — the disk term silently stopped seeing imbalances below roughly MB-scale. The return type is now `double`; the subtraction is done in `double` (exact for `long` byte counts up to 2^53 ≈ 9 PB) and the divide is also in `double`, with a single cast to `float` for the final weight contribution. This applies to **both** modes; it is a bug fix, not a behavior change at any realistic cluster size.

No `@PublicApi` is affected — `ShardsBalancer` is `@opensearch.internal`.

## Key changes

- **`BalancedShardsAllocator`** — new `DISK_USAGE_BALANCE_FACTOR_SETTING` (dynamic, node-scope, finite-float validator rejecting NaN/±Inf) and new `BALANCE_MODE_SETTING` (enum). `WeightFunction.weight(...)` branches on `ratioMode` and normalizes shard/index terms in ratio mode. Javadoc on both settings and on `WeightFunction` describes the modes, the threshold interaction, and the small-index amplification caveat.
- **`BalanceMode`** (new) — small internal enum with a `parse(String)` that surfaces a clear error for unknown values.
- **`ModelNode` / `LocalShardsBalancer` / `ShardsBalancer`** — tracks `diskUsageInBytes` only when the disk factor is non-zero (see `trackDiskUsage`); `avgDiskUsageInBytesPerNode` is `double` and recomputed from live model state (not a stale snapshot).
- **`ClusterSettings`** — one-line registrations alongside existing `balance.*` settings.

## Testing

**Unit tests** (`DiskUsageBalanceTests`):
- weight-function algebra includes the normalized disk term iff factor > 0,
- `avgDiskUsageInBytesPerNode` stays consistent with per-node state across `allocateUnassigned` / `moveShards` / `tryRelocateShard` (regression guard),
- validator rejects NaN and ±Infinity,
- dynamic update path,
- **paired correctness tests** — `testDiskUsageBalanceMovesShardWhenCountBalanced` sets up a 3-node / 2-index cluster with counts balanced (3+3+3) but bytes skewed onto node-0 and asserts ≥1 large shard relocates once the factor is enabled; its twin `testDiskUsageBalanceNoOpWhenZero` runs the same scenario at factor=`0.0f` and asserts zero relocations,
- **ratio-mode unit tests (new)** — `testRatioModeNormalizesAllThreeTerms` asserts the expected algebra for all three terms, `testRatioModeHandlesZeroDenominators` guards the `max(1, avg)` divisor paths, `testDefaultModeKeepsRawCountDeltas` pins count-mode semantics.

**Integration tests** (`DiskUsageBalanceIT`, 4 specs, all pass locally in 14.4s):
- `testDiskUsageBalanceRebalancesByBytes` — end-to-end count-mode disk rebalance: byte spread does not grow when disk_usage balance is enabled.
- `testDefaultBehaviorUnchanged` — default `0.0f` is byte-spread-agnostic (allocator strictly count-balances).
- `testRatioModeRebalancesByBytes` (new) — with `balance.mode=ratio`, `disk_usage=1.0f`, `threshold=0.1f`: cluster stays green, all shards assigned, byte spread does not exceed count-mode baseline.
- `testBalanceModeDynamicToggle` (new) — toggles `balance.mode` between `ratio` and `count` four times at runtime; ensures the `addSettingsUpdateConsumer` path rebuilds the weight function without leaving shards unassigned. Also asserts that `balance.mode=bogus` is rejected with a clear error.

Suite resets all three settings via `@After` to prevent state leaking across specs on failure.

## Migration

None.
- `balance.disk_usage` defaults to `0.0f` (disabled).
- `balance.mode` defaults to `count` (pre-PR behavior).
- The disk-term precision fix is a no-op for any cluster under a few TB of data; above that, it makes the disk term more accurate (fewer below-threshold moves ignored).

All settings are dynamic — enable/tune/disable at runtime without restart.

## Unrelated included fix

Commit on this branch removes too-short `assertBusy` timeouts in `IndicesRequestCacheCleanupIT` — a known flaky tracked under #21397. The fix is also up standalone as #21494 so it can land on `main` independently. If #21494 merges first, that commit falls out of this branch on rebase.


AndreKurait:
The failing test is unrelated to this PR. The diff only touches `BalancedShardsAllocator`, `LocalShardsBalancer`, `ShardsBalancer`, `ClusterSettings`, `CHANGELOG.md`, and two new test files under `cluster/routing/allocation/**`. It does not touch `server/src/test/java/org/opensearch/action/search/QueryPhaseResultConsumerTests.java` or any of its dependencies.

The failure is in `QueryPhaseResultConsumerTests.testProgressListenerExceptionsAreCaught` (seed `283CB275127A6AC7`):

```
java.lang.NullPointerException: Cannot suppress a null exception.
    at java.lang.Throwable.addSuppressed(Throwable.java:1103)
    at QueryPhaseResultConsumerTests.lambda$testProgressListenerExceptionsAreCaught$1(QueryPhaseResultConsumerTests.java:136)
    at java.util.concurrent.atomic.AtomicReference.accumulateAndGet(AtomicReference.java:260)
```

The test's accumulator (`curr.addSuppressed(prev)`) does not null-guard `prev`, so when two partial-merge failures race the first `accumulateAndGet` call observes `prev == null` and `addSuppressed` rejects it. This is a race in the test, not in product code.

This is tracked by #19725 ([AUTOCUT] Gradle Check Flaky Test Report for QueryPhaseResultConsumerTests, open since 2025-10-22). I'll add a cross-reference there as well.

Pushing a follow-up commit with expanded manual-test evidence will re-trigger gradle-check.

AndreKurait:
## Manual-test matrix: does `disk_usage` beat tuning existing `shard` / `index` knobs?

Ran an expanded manual-test matrix on a real 3-node cluster (PR build @ 4cd59d56) to answer exactly this question. TL;DR: **no pre-existing setting can reach the byte-balance configurations that `disk_usage` reaches**; the existing weight function is byte-blind by construction.

### Workload A — 4 heterogeneous indices (huge/large/small/tiny), 2 shards×1 replica each

| # | scenario | shard | index | disk_usage | shards CV | **bytes CV** | **max/min** |
|---|---|---|---|---|---|---|---|
| A01 | baseline defaults | 0.45 | 0.55 | 0.0 | 8.8% | 26.3% | 1.79× |
| A02 | shard-heavy | **2.0** | 0.55 | 0.0 | 8.8% | 26.3% | 1.79× |
| A03 | index-heavy | 0.45 | **2.0** | 0.0 | 8.8% | 26.3% | 1.79× |
| A04 | both-heavy | **2.0** | **2.0** | 0.0 | 8.8% | 26.3% | 1.79× |
| A05 | shard-near-zero | **0.01** | 0.55 | 0.0 | 8.8% | 26.3% | 1.79× |
| **A06** | **disk_usage 1e-7** | 0.45 | 0.55 | **1e-7** | 8.8% | **7.3%** | **1.19×** |
| A07 | disk_usage 1e-6 | 0.45 | 0.55 | 1e-6 | 8.8% | 16.5% | 1.46× |
| A08 | disk_usage 1e-5 | 0.45 | 0.55 | 1e-5 | 8.8% | 12.9% | 1.37× |
| A09 | shard-heavy + du=1e-7 | 2.0 | 0.55 | 1e-7 | 8.8% | 8.0% | 1.20× |

**A01–A05 produce byte-identical layouts.** Tuning `shard` up 4× or down 45×, tuning `index` up 4×, or both, all land the cluster in the same (26.3% byte CV, 1.79× skew) configuration. The existing weight function has no byte signal, so once shard counts have stabilised nothing further moves. A06 cuts byte-CV by **72%** (26.3 → 7.3) at zero cost to shard count (still 5/6/5).

### Workload B — A + one fat index `idx-monster` (4 shards×1 replica, 8 kB docs)

This is the pathological case where shard counts are already at the theoretical optimum:

| # | scenario | shard | index | disk_usage | shards CV | **bytes CV** | **max/min** |
|---|---|---|---|---|---|---|---|
| B01 | baseline (shards already 10/10/10) | 0.45 | 0.55 | 0.0 | **0.0%** | 18.7% | 1.60× |
| B02 | shard-heavy | 2.0 | 0.55 | 0.0 | 0.0% | 18.7% | 1.60× |
| B03 | index-heavy | 0.45 | 2.0 | 0.0 | 0.0% | 18.7% | 1.60× |
| B04 | both-heavy | 2.0 | 2.0 | 0.0 | 0.0% | 18.7% | 1.60× |
| **B05** | **disk_usage 1e-7** | 0.45 | 0.55 | **1e-7** | 10.2% | **3.1%** | **1.08×** |
| B06 | disk_usage 1e-6 | 0.45 | 0.55 | 1e-6 | 10.2% | 8.6% | 1.24× |
| B07 | disk_usage 1e-5 | 0.45 | 0.55 | 1e-5 | 10.2% | 4.1% | 1.09× |
| B08 | shard-heavy + du=1e-7 | 2.0 | 0.55 | 1e-7 | 0.0% | 18.3% | 1.56× |

**B01–B04 are provably unfixable by stock knobs.** Shard counts are perfect (0.0% CV), so there is no count-imbalance pressure to move anything, yet one node carries ~1.6× the bytes of another. **B05 takes byte-CV from 18.7 → 3.1 (83% reduction) and max/min from 1.60× → 1.08×** in exchange for 9/10/11 shard count (10.2% shard CV). This is the tradeoff customers in #15520 have been asking for and it is unreachable via the existing weight function.

### Caveats from the matrix

- **Calibration matters.** Going higher than 1e-7 on this demo dataset (A07 16.5%, A08 12.9%) can actually worsen balance — the disk-usage term over-dominates the move gate and the allocator oscillates. Operators need `factor × typical_per_shard_bytes` to land around the `threshold` setting's sensitivity (~1.0). At TB-scale production shards this falls in the `1e-10` to `1e-9` range (matches the calibration table in the setting's Javadoc).
- **B08 is the cautionary row.** With `shard = 2.0`, even `disk_usage = 1e-7` is swamped (bytes CV regresses to 18.3%). Operators who raise `shard` above its default must proportionally raise `disk_usage` — this is an expected consequence of the non-normalised weight formulation documented in the Javadoc.

### Reproducer

```
./manual-evidence-v3.sh        # 17-scenario sweep, ~20 min
cat manual-test-run/evidence-v3/stats.txt
column -t -s $'\t' manual-test-run/evidence-v3/summary.tsv
```

Per-scenario `_cat/allocation` + `_cat/shards` outputs are saved per-file. Environment: 3 local nodes, same process lifetime across all scenarios, each scenario resets to `shard=0.45 / index=0.55 / disk_usage=0.0` and waits for green + `no_relocating_shards` before applying the scenario settings and snapshotting.

server/src/main/java/org/opensearch/cluster/routing/allocation/allocator/BalancedShardsAllocator.java
```diff
@@ -595,7 +684,13 @@ public float weightWithRebalanceConstraints(ShardsBalancer balancer, ModelNode n
         float weight(ShardsBalancer balancer, ModelNode node, String index) {
             final float weightShard = node.numShards() - balancer.avgShardsPerNode();
             final float weightIndex = node.numShards(index) - balancer.avgShardsPerNode(index);
-            return theta0 * weightShard + theta1 * weightIndex;
+            // The disk-usage deviation is normalized by the cluster average, so the disk term
+            // is a dimensionless ratio while weightShard / weightIndex are raw shard-count
+            // deviations. See DISK_USAGE_BALANCE_FACTOR_SETTING javadoc for calibration guidance.
+            // When the average is zero (empty cluster) the disk term contributes zero.
+            final float avgDiskUsage = balancer.avgDiskUsageInBytesPerNode();
```
andrross: Can you keep the avgDiskUsageInBytesPerNode() as a `long` so you can do integer arithmetic with the subtraction, then only convert to floating point with the final ratio? I'm not sure the values ever get large enough that the loss of precision will be significant, but it seems more natural to keep byte values as integers (even for an average).

andrross:
@shwetathareja FYI

AndreKurait:
The `detect-breaking-change` (japicmp) failure on this PR is not caused by anything in #21343. Since CI runs against the merged tree (`main` + this branch), the jar under test includes recent upstream commits \u2014 including #21143 (`a70e8ee0`, "Refactor WLM settings to use Setting objects and rename field"), which removed a `@PublicApi` method without deprecation.

Full japicmp output filtered for **binary-breaking** markers (`***!` / `---!`):

```
***! MODIFIED CLASS: PUBLIC org.opensearch.cluster.metadata.WorkloadGroup  (not serializable)
        ---! REMOVED METHOD: PUBLIC(-) java.util.Map<java.lang.String,java.lang.String> getSearchSettings()
```

That is the **only** incompatibility in the whole report. Everything else is `(compatible)` new-method/new-field additions \u2014 informational, not blocking.

Verification:
- Pre-merge `a1fd01e8` passed `Detect Breaking Changes` (run [25395196071](https://github.com/opensearch-project/OpenSearch/actions/runs/25395196071)) at 18:38 UTC.
- Post-merge (same PR SHA, but merge-base advanced to include `a70e8ee0`): fails (run [25401092289](https://github.com/opensearch-project/OpenSearch/actions/runs/25401092289)) at 20:40 UTC.
- Ran `./gradlew :server:japicmp` locally against `merge(origin/main, a1fd01e8)` and reproduced; `WorkloadGroup.getSearchSettings()` removal is the sole `---!` line.

PR #21343 itself introduces no `@PublicApi` removals or signature changes. `BalancedShardsAllocator`, `LocalShardsBalancer`, `ShardsBalancer`, and `ClusterSettings` changes are all additive (new setting, new concrete method with default return, new weight term) and all classes involved are either `@opensearch.internal` or unannotated.

Every in-flight PR against `main` right now will hit the same japicmp failure for the same reason. Unblocking options (up to maintainers):
1. Restore `WorkloadGroup.getSearchSettings()` as a `@Deprecated` shim delegating to the new accessor.
2. Apply `@DeprecatedApi` / `@InternalApi` on `WorkloadGroup` if it should no longer be `@PublicApi`.
3. Roll the `japicmp.compare.version` baseline forward.

cc @dzane17 @andrross

dzane17:
@AndreKurait Thanks for the analysis. I have a PR with the potential fix: https://github.com/opensearch-project/OpenSearch/pull/21500

ztomaszewska:
Hello @AndreKurait thank you for this PR. This feature is really needed! Do you think it is feasible to merge it and have it in 3.8 release?

kcpicot:
Hello @AndreKurait , yes thank you very much for this PR. I follow it since April as we have numerous issues with storage balancing which causes unassigned shard replicas (currently the storage balancing is by index number which is not really pertinent).

Your pull request would be of great use for us, thank you for your work !

PR #21346, opened 2026-04-24
Add infrastructure to convert analytics engine execution results to OpenSearch aggregation r…

varun-st:
## Description

Adds infrastructure to convert analytics engine execution results into OpenSearch `InternalAggregations` format, enabling proper aggregation responses in the DSL query executor.

## Changes

- **AggregationResponseBuilder**: Converts flat execution results to nested aggregation structure using granularity-based matching
- **ComparisonUtils**: Handles type coercion for numeric comparisons (Integer vs Double)
- **TransportDslExecuteAction**: Integrates response builder to construct aggregations from execution results
- **SearchResponseBuilder**: Updated to accept pre-built aggregations instead of raw results


bharath-techie:
Have we done some benchmarks ? Can you please add the performance regression due to this result conversion ? 

Lets do 10,100,10k results in hits type of query and in aggs type of query.

PR #21348, opened 2026-04-24
Added substrait converter for the fragments

vinaykpud:
### Description

Add Substrait-based fragment converter to the DataFusion backend plugin, enabling end-to-end query execution from DSL query → Calcite RelNode → Substrait plan → native DataFusion runtime → Arrow result batches → materialized rows. 

- **Fix classloader dependency issues**: `flatbuffers-java`, `arrow-format`, `jackson-databind`, `jackson-annotations` moved to bundled in `analytics-engine` (parent plugin). Arrow's `ArrowType` subclasses are loaded by the parent classloader and transitively reference these jars — child plugins can't provide them. Removed duplicates from `analytics-backend-datafusion` to avoid jar hell.

- **Fix Substrait TCCL issue**: Substrait uses Jackson polymorphic deserialization which resolves class names via the thread context classloader. On search threads the TCCL is the system CL and can't see plugin classes. Moved extension catalog loading to plugin bootstrap with a one-time TCCL swap so it's cached before any query runs.

- **Fix Arrow memory leak**: `batchesToRows()` iterated `VectorSchemaRoot` batches without closing them. Added `batch.close()` after materializing rows so the per-query allocator closes cleanly.

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/src/test/java/org/opensearch/be/datafusion/DataFusionFragmentConvertorTests.java
```diff
@@ -0,0 +1,276 @@
...
+import io.substrait.proto.PlanRel;
+import io.substrait.proto.ReadRel;
+import io.substrait.proto.Rel;
+
+/**
+ * Tests for {@link DataFusionFragmentConvertor}.
+ * Builds Calcite RelNode trees in-process, converts them via the convertor,
+ * then decodes the returned Substrait proto bytes and asserts on proto shape.
+ */
+public class DataFusionFragmentConvertorTests extends OpenSearchTestCase {
```
mch2, line 56: We should make a note to start updating and adding UTs using the actual df runtime, like `DataFusionQueryExecutionTests` to test the built substrait

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -97,6 +99,8 @@ public DefaultPlanExecutor(
     public Iterable<Object[]> execute(RelNode logicalFragment, Object context) {
         RelNode plan = PlannerImpl.createPlan(logicalFragment, new PlannerContext(capabilityRegistry, clusterService.state()));
         QueryDAG dag = DAGBuilder.build(plan, capabilityRegistry, clusterService);
+        PlanForker.forkAll(dag, capabilityRegistry);
+        FragmentConversionDriver.convertAll(dag, capabilityRegistry);
```
mch2, line 106: can be a follow up when we clean up the executor - we should measure time taken in all these steps, also i think we should wrap all these in a component that takes the fragment and returns the built dag ready for scheduling.

mch2 (approved):
overall looks like a good start here for substrait conversion, thanks @vinaykpud .

sandeshkr419 (approved):

sandeshkr419:
```
Execution failed for task ':sandbox:plugins:analytics-backend-datafusion:thirdPartyAudit'.
> Audit of third party dependencies failed
```

@vinaykpud lets fix this as well. Thanks!

vinaykpud:
@sandeshkr419 
Sure. 
The missing class `SqlDdlParserImpl` is Calcite's DDL parser,it handles SQL statements like CREATE TABLE, DROP TABLE, ALTER TABLE.  It lives in a separate calcite-server module and we don't need DDL parsing. So fixing by adding it to ignored list

sandbox/plugins/analytics-backend-datafusion/src/test/java/org/opensearch/be/datafusion/DataFusionFragmentConvertorTests.java
vinaykpud: Okay, i am adding TODO for it

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
vinaykpud: makes sense, will add a followup

PR #21350, opened 2026-04-24
Fix O(n^2) removeAll in remote translog metadata cleanup

gbbafna:
### Description

`ArrayList.removeAll(ArrayList)` in `RemoteFsTimestampAwareTranslog` is O(n*m) because `ArrayList.contains()` is a linear scan. When the number of translog metadata files grows large (e.g., 500+ files for long-lived system indices like `.opendistro-ism-config`), this causes sustained CPU spikes (500ms at 99.9% CPU) on the `remote_purge` thread.

Stack trace observed:
```
ArrayList.indexOfRange → ArrayList.indexOf → ArrayList.contains → ArrayList.batchRemove → ArrayList.removeAll
  → RemoteFsTimestampAwareTranslog$1.onResponse
```
Hot threads showing 100% cpu 

```
100.4% (502ms out of 500ms) cpu usage by thread 'opensearch[abc][remote_purge][T#1]'
     10/10 snapshots sharing following 20 elements
       java.base@21.0.9/java.util.ArrayList.indexOfRange(ArrayList.java:299)
       java.base@21.0.9/java.util.ArrayList.indexOf(ArrayList.java:286)
       java.base@21.0.9/java.util.ArrayList.contains(ArrayList.java:275)
       java.base@21.0.9/java.util.ArrayList.batchRemove(ArrayList.java:911)
       java.base@21.0.9/java.util.ArrayList.removeAll(ArrayList.java:873)
       app//org.opensearch.index.translog.RemoteFsTimestampAwareTranslog$1.onResponse(RemoteFsTimestampAwareTranslog.java:208)
       app//org.opensearch.index.translog.RemoteFsTimestampAwareTranslog$1.onResponse(RemoteFsTimestampAwareTranslog.java:173)
       app//org.opensearch.common.blobstore.EncryptedBlobContainer.lambda$listBlobsByPrefixInSortedOrder$8(EncryptedBlobContainer.java:266)
       app//org.opensearch.common.blobstore.EncryptedBlobContainer$$Lambda/0x0000000203720698.accept(Unknown Source)
```

### Solution

Wrap the `List` argument in `HashSet` before passing to `removeAll()`, reducing lookup complexity from O(n) to O(1). This fixes 3 occurrences in the file. A 4th `removeAll` call already uses a `Set` argument and is unaffected.

### Check List


server/src/main/java/org/opensearch/index/translog/RemoteFsTimestampAwareTranslog.java
```diff
@@ -202,12 +202,22 @@ public void onResponse(List<BlobMetadata> blobMetadata) {
...
+                    logger.debug(
+                        () -> "metadataFilesToBeDeleted count = "
+                            + metadataFilesToBeDeleted.size()
+                            + ", metadataFilesToBeDeleted = "
+                            + metadataFilesToBeDeleted
+                    );
                     // For all the files that we are keeping, fetch min and max generations
                     List<String> metadataFilesNotToBeDeleted = new ArrayList<>(metadataFiles);
-                    metadataFilesNotToBeDeleted.removeAll(metadataFilesToBeDeleted);
+                    metadataFilesNotToBeDeleted.removeAll(new HashSet<>(metadataFilesToBeDeleted));
```
cwperks, line 213: Does order matter for data structures like `metadataFilesToBeDeleted`? If not, then why not make the whole data structure a Set.

gbbafna: Good suggestion. Order doesn't matter for these lists, so Set would be semantically more correct. However, `metadataFilesToBeDeleted` originates from `RemoteStoreUtils.filterOutMetadataFilesBasedOnAge()` which is shared with the segment store path, and `deleteMetadataFilesAsync` downstream expects `List<String>` . 

Converting the full  chain would be a broader refactor touching `RemoteStoreUtils`, `RemoteSegmentStoreDirectory` , and multiple tests — better suited as a follow-up. The current fix already achieves O(n) by wrapping in HashSet for the `removeAll` calls.

sricharanvuppu (approved):

Bukhtawar (approved):
How do we ensure future consumers of metadataFiles handle this correctly. Can we open a TODO on refactoring the class structure broadly

PR #21351, opened 2026-04-24
Fix Lucene JUL logs appearing as WARN on stderr at startup

### LINKED ISSUE #21323: [BUG] warning log entries (stderr) from lucene on startup
### Describe the bug

```json
{"type": "server", "timestamp": "2026-04-22T13:53:25,584+0000", "level": "WARN", "component": "stderr", "cluster.name": "test-opensearch", "node.name": "test-opensearch-0", "message": "Apr 22, 2026 1:53:25 PM org.apache.lucene.internal.vectorization.PanamaVectorizationProvider <init>" }
{"type": "server", "timestamp": "2026-04-22T13:53:25,585+0000", "level": "WARN", "component": "stderr", "cluster.name": "test-opensearch", "node.name": "test-opensearch-0", "message": "INFO: Java vector incubator API enabled; uses preferredBitSize=256; FMA enabled" }
```

### Related component

_No response_

### To Reproduce

start a fresh cluster (i used the minimal distribution) and check the log

### Expected behavior

### Additional Details

**Plugins**
n/a

**Screenshots**
n/a

 - OS: linux
 - Version OpenSearch 3.6.0

**Additional context**
they also seem to be coming from lucene rather than OpenSearch; nevertheless i'm reporting this here since you're more familiar with how you integrate lucene and whether/how this can/should be fixed downstream.

Chocochip101:
## Description

Fixes #21323

Lucene's `PanamaVectorizationProvider` uses `java.util.logging` (JUL) to log an INFO message during initialization. Because JUL's `LogManager` is initialized by the JVM before OpenSearch sets the `java.util.logging.manager` system property, JUL falls back to its default `ConsoleHandler` which writes to `System.err`. OpenSearch redirects `System.err` to Log4j2 at WARN level, causing a harmless INFO message to appear as WARN with component `stderr`:

```
{"level": "WARN", "component": "stderr", "message": "INFO: Java vector incubator API enabled; ..."}
```

### Changes

Install `Log4jBridgeHandler` programmatically after Log4j2 is configured in `LogConfigurator.configure()`. This removes the default `ConsoleHandler` from the JUL root logger and routes all subsequent JUL logging through Log4j2 at the correct level.

The existing approach of setting the `java.util.logging.manager` system property alone is insufficient because the JVM initializes JUL's `LogManager` before this property is set.

## Issues Resolved

Closes #21323

## Testing

- Verified compilation succeeds
- Existing logging tests pass (`./gradlew :server:test --tests "org.opensearch.common.logging.*"`)


Chocochip101:
The latest CI failure appears to be an unrelated flaky test:

```
Tests with failures:
 - org.opensearch.storage.common.BlockTransferManagerTests.classMethod

java.lang.AssertionError: The test or suite printed 66252 bytes to stdout and stderr, 
even though the limit was set to 8192 bytes.
```

`BlockTransferManagerTests` is in the storage/block-transfer module and is unrelated to this PR's changes (which only modify JUL→Log4j2 bridge installation in `LogConfigurator`). This failure is a randomized-test seed issue where the test suite exceeded the stdout/stderr byte limit — a class of failures known to be flaky in the OpenSearch test environment.

Requesting a CI retry. /run-gradle-check

Chocochip101:
Closing in favor of #21359 which has a clean commit history.

PR #21355, opened 2026-04-24
Extend OpenSearchTestCase in BlockIndexInputTests

AndreKurait:
### Description

Two recently-added unit suites — `BlockTransferManagerTests` and `BlockIndexInputTests` — extend Lucene's `LuceneTestCase` directly instead of `OpenSearchTestCase`. Under `-Dtests.opensearch.logger.level=DEBUG`, `BlockTransferManagerTests` fails with:

```
java.lang.AssertionError: The test or suite printed 66252 bytes to stdout
and stderr, even though the limit was set to 8192 bytes. Increase the
limit with @Limit, ignore it completely with @SuppressSysoutChecks or
run with -Dtests.verbose=true
    at org.apache.lucene.tests.util.TestRuleLimitSysouts.afterIfSuccessful(...)
```

`BlockIndexInputTests` has the same latent exposure but today stays under the 8 KiB limit.

#### Root cause

Lucene's `TestRuleLimitSysouts` caps total `stdout+stderr` at **8192 bytes** per suite by default. Most OpenSearch tests extend `OpenSearchTestCase`, which carries an inherited relaxation of that cap:

```java
// test/framework/.../OpenSearchTestCase.java:213
@LuceneTestCase.SuppressSysoutChecks(bugUrl = "we log a lot on purpose")
public abstract class OpenSearchTestCase extends LuceneTestCase { ... }
```

Both affected suites skip that inheritance chain by extending `LuceneTestCase` directly, yet both exercise production OpenSearch infrastructure that logs substantially:

```
BlockTransferManagerTests (9 tests)        BlockIndexInputTests (8 tests)
────────────────────────────────────       ────────────────────────────────
@Before: new TestThreadPool()              @Before: FSDirectory.open(tempDir)
  → ~25 DEBUG lines per ThreadPool           → no server-infra logs
  → 9 tests × ~25 = ~225 lines             Production under test:
@Before: new IndexSettings()                 BlockIndexInput has 2
  → 3 INFO lines (MergeSchedulerConfig)      logger.debug() call sites
Production under test:                       → a few DEBUG lines per test
  BlockTransferManager has 4+ debug/
  error call sites exercised by failure-
  path tests
──────────────────────────────────────     ────────────────────────────
Measured XML under DEBUG (this branch):    Measured XML under DEBUG (this branch):
  110 KB, 366 DEBUG lines, 0 failures        43 KB, 99 DEBUG lines, 0 failures
Without inherited @SuppressSysoutChecks:   Without inherited @SuppressSysoutChecks:
  FAILS — 66,252 bytes > 8,192              passes today at ~1.2 KB but is
  cap (reproducer signature above)          latent; any future logger.debug
                                            or added @Before infra would trip
                                            the same cap
```

The assertion fires in the test-rule *afterIfSuccessful* phase, so individual `@Test` methods pass but the suite fails with `classMethod FAILED`. Because the triggering logs are produced by ordinary `ThreadPool` / `MergeSchedulerConfig` / `BlockTransferManager` code paths — not per-test anomalies — the failure is deterministic on any DEBUG run.

#### Why each test should extend `OpenSearchTestCase`

Both suites construct and test OpenSearch server infrastructure — not pure Lucene primitives — so both belong on the project-convention base class. A repo-wide audit (`rg 'extends\s+LuceneTestCase\b'` on the test tree) confirms the other direct-`LuceneTestCase` subclasses are either **pure Lucene primitive tests** (scorer, highlighter, hash-set), **container/IT tests** running in separate JVMs, or have documented reasons to avoid `OpenSearchTestCase`'s syscall filter (`SpawnerNoBootstrapTests`). These two are the only infrastructure-test outliers.

The sibling production test `TransferManagerTestCase` already uses this exact pattern, so the fix is a one-to-one alignment rather than a new convention:

```java
// server/src/test/java/org/opensearch/index/store/remote/utils/TransferManagerTestCase.java
@ThreadLeakFilters(filters = CleanerDaemonThreadLeakFilter.class)
public abstract class TransferManagerTestCase extends OpenSearchTestCase { ... }
```

Thanks to @mgodwan and @MayankHarsh03 for suggesting this approach over a per-class `@SuppressSysoutChecks` annotation.

#### Fix

For each of the two suites, swap the import and the `extends` clause — no other behavior change:

```
- import org.apache.lucene.tests.util.LuceneTestCase;
+ import org.opensearch.test.OpenSearchTestCase;

- public class BlockTransferManagerTests extends LuceneTestCase {
+ public class BlockTransferManagerTests extends OpenSearchTestCase {

- public class BlockIndexInputTests extends LuceneTestCase {
+ public class BlockIndexInputTests extends OpenSearchTestCase {
```

`@ThreadLeakFilters(filters = CleanerDaemonThreadLeakFilter.class)` is preserved on each class — it merges additively with the `OpenSearchTestCase`-level filter (`defaultFilters=true`), identical to the `TransferManagerTestCase` precedent.

#### Alternatives considered

- **`@LuceneTestCase.SuppressSysoutChecks(bugUrl = "we log a lot on purpose")` on each class** (v1 of this PR) — works, but duplicates an annotation that `OpenSearchTestCase` already carries and leaves these suites as outliers among siblings. Extending `OpenSearchTestCase` is the same fix one level up with no added cost.
- **`@Limit(bytes = …)`** — would need a hand-picked byte budget; brittle as tests or log lines evolve.
- **Lower log levels to `WARN` via a `log4j2-test.properties`** — would silence useful diagnostics for anyone debugging these classes and spreads concern across files for a local issue.
- **Stub/mute `ThreadPool` logger in `@BeforeClass`** — leakage risk into other suites running in the same JVM.

#### Testing

Locally on this branch (JDK 21, Linux), with the exact conditions that reproduce the failure against pristine main:

```
./gradlew :server:test \
    --tests org.opensearch.storage.common.BlockTransferManagerTests \
    --tests org.opensearch.storage.indexinput.BlockIndexInputTests \
    -Dtests.opensearch.logger.level=DEBUG --rerun
# BUILD SUCCESSFUL — 17 tests, 0 failures, 0 errors
# XML: 110 KB (BlockTransferManagerTests, 366 DEBUG lines) +
#       43 KB (BlockIndexInputTests,       99 DEBUG lines)
#      (inherited @SuppressSysoutChecks doing its job on both)

./gradlew :server:test \
    --tests org.opensearch.storage.common.BlockTransferManagerTests \
    -Dtests.iters=3 --rerun
# BUILD SUCCESSFUL — 27 tests, 0 failures (9 tests × 3 iters, random seeds)

./gradlew :server:spotlessJavaCheck
# BUILD SUCCESSFUL
```

### Related Issues

No prior issue filed — discovered via gradle-check failure report on `BlockTransferManagerTests`; `BlockIndexInputTests` pulled in during review as a same-commit-of-origin sibling with identical base-class drift.

### Check List

- [x] Functionality includes testing. *(Test-only change; the "testing" is the existing suites plus iteration stress run locally under `-Dtests.opensearch.logger.level=DEBUG`.)*
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable. *(N/A — no API change.)*
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable. *(N/A — test-infra only.)*


server/src/test/java/org/opensearch/storage/common/BlockTransferManagerTests.java
```diff
@@ -52,6 +52,7 @@
  * Tests cover single block downloads, failure scenarios, duplicate handling, and concurrent operations.
  */
 @ThreadLeakFilters(filters = CleanerDaemonThreadLeakFilter.class)
+@LuceneTestCase.SuppressSysoutChecks(bugUrl = "we log a lot on purpose")
```
mgodwan: Can this not extend from OpenSearchTestCase?

https://github.com/opensearch-project/OpenSearch/blob/21d3a8ac2585e5d7bb000624dec04734d0110fab/test/framework/src/main/java/org/opensearch/test/OpenSearchTestCase.java#L213

mgodwan: Tagging author of the original test @MayankHarsh03

MayankHarsh03: Thanks for raising this @AndreKurait . As @mgodwan suggested, the better fix here is to change extends LuceneTestCase to extends OpenSearchTestCase. It already includes @SuppressSysoutChecks and follows the project convention. Let me know if you have any suggestions or I'll push the changes.

AndreKurait: Thank you, I can tackle this today

MayankHarsh03:
Hi @AndreKurait , for now I have picked this change and the PR #21363 got merged.

AndreKurait:
@MayankHarsh03 i had made that same change and also added server/src/test/java/org/opensearch/storage/indexinput/BlockIndexInputTests.java. I can change this PR to just be BlockIndexInputTests since it also needs this

AndreKurait:
@cwperks Could you take a look at this since you reviewed https://github.com/opensearch-project/OpenSearch/pull/21363?

cwperks (approved):

cwperks:
@AndreKurait Approved. Do you think we need to change other instances? https://github.com/search?q=org%3Aopensearch-project+%22extends+LuceneTestCase%22&type=code

^ If so, is there anything we can do with checkstyle (forbiddenApis) to catch this in the precommit hook?

PR #21356, opened 2026-04-24
Coordinator side DataFusion reduce

### LINKED ISSUE #21403: Add Datafusion plugin provided sink for coordinator reduce tasks
Analytics engine will leverage backends to execute final reduce query fragments.  Update our DatFusion backend plugin to provide an implementation of ExchangeSinkProvider that merges results from child stages.

To start we will assume a single child stage, where multiple shards push rows into the sink.  Thinking to start with two implementations that we can test and benchmark 1. A MemoryTable implementation, where all batches are buffered until the child stage completes and 2. A StreamingTable implementation, where batches are incrementally pushed and reduced in the back-end.

mch2:
### Description

   Wires the **coordinator-side reduce** through DataFusion. Per-shard partials stream
  in over the existing exchange, get pushed into a DF session running in-process,
  and the merged batches stream back out.

     data nodes ──Arrow batches──►  DatafusionReduceSink (Java)
                                            │
                                         FFM push
                                            ▼
                                    PartitionStream (Rust) ──► DataFusion LocalExec
                                            │
                                         FFM pull (drain thread)
                                            ▼
                                     merged batches ──► coordinator output

  Two sink variants:
  - `DatafusionReduceSink` — streaming; partitions feed `PartitionStream` as they arrive.
  - `DatafusionMemtableReduceSink` — buffer-then-reduce via DF MemTable (reference path).

  Also in this PR:
  - `test-ppl-frontend` extracted to a top-level plugin so QA and analytics-engine ITs
    can share it; dead `ppl/planner` test trees removed.

  ### Tests
  `CoordinatorReduceIT`, `CoordinatorReduceMemtableIT` (3-node cluster, scalar sum
  across shards) + sink/native-bridge unit tests + Rust `local_exec_test.rs`.

  ### Related Issues
  Resolves https://github.com/opensearch-project/OpenSearch/issues/21403

### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/partition_stream.rs
```diff
@@ -0,0 +1,307 @@
...
+    /// thread to be a Tokio worker.
+    ///
+    /// Blocks while the channel is full (natural backpressure). Returns an
+    /// error only if the receiver has been dropped.
+    pub fn send_blocking(
+        &self,
+        batch: Result<RecordBatch, DataFusionError>,
+        handle: &Handle,
+    ) -> Result<(), DataFusionError> {
+        handle.block_on(self.tx.send(batch)).map_err(|_| {
```
bharath-techie, line 83: We are doing some experiments where we are evaluating spawn vs block_on performance in the data nodes [ today its block on ] . We can take a decision here based on that.

sandbox/plugins/analytics-backend-datafusion/rust/src/local_executor.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// plan. The returned stream is hot — polling it drives both the reduce
+    /// computation and the consumption of the streaming inputs.
+    pub async fn execute_substrait(
+        &self,
+        bytes: &[u8],
+    ) -> Result<SendableRecordBatchStream, DataFusionError> {
+        let plan = Plan::decode(bytes).map_err(|e| {
+            DataFusionError::Execution(format!("Failed to decode Substrait plan: {}", e))
+        })?;
+        let logical_plan = from_substrait_plan(&self.ctx.state(), &plan).await?;
```
bharath-techie, line 147: Trying to understand a bit here , will filter and scan be removed from this logical plan ? [ Since that is applicable only in the data node ] 

Will we do only final aggs on all the partial aggs which we receive as individual partitions ?

mch2: This executor won't manipulate the plan, it just takes the fragment generated during planning and executes.  During planning/dag building step when we generate the substrait fragments is where this would happen but it depends on the query.  The "Scan" here is on `OpenSearchStageInputScan`, where the input is arrow streams.  Where filters can be pushed down they will.

Bukhtawar (approved):
Overall LGTM

PR #21359, opened 2026-04-25
Fix Lucene JUL logs appearing as WARN on stderr at startup

### LINKED ISSUE #21323: [BUG] warning log entries (stderr) from lucene on startup

Chocochip101:
## Description

Fixes #21323

Lucene's `PanamaVectorizationProvider` uses `java.util.logging` (JUL) to log an INFO message during initialization. Because JUL's `LogManager` is initialized by the JVM before OpenSearch sets the `java.util.logging.manager` system property, JUL falls back to its default `ConsoleHandler` which writes to `System.err`. OpenSearch redirects `System.err` to Log4j2 at WARN level, causing a harmless INFO message to appear as WARN:

```
{"level": "WARN", "component": "stderr", "message": "INFO: Java vector incubator API enabled; ..."}
```

## Changes

Install `Log4jBridgeHandler` programmatically after Log4j2 is configured in `LogConfigurator.configure()`. This removes the default `ConsoleHandler` from the JUL root logger and routes all subsequent JUL logging through Log4j2 at the correct level.

An `AtomicBoolean` guard (`julBridgeInstalled`) prevents duplicate installation when `configure()` is called multiple times (e.g., in tests).

The existing approach of setting the `java.util.logging.manager` system property alone is insufficient because the JVM initializes JUL's `LogManager` before this property is set.

## Testing

- Added `JulBridgeTests` verifying:
  - `Log4jBridgeHandler` is installed on the JUL root logger
  - Non-bridge handlers (e.g., `ConsoleHandler`) are removed after installation
  - Multiple `install()` calls result in exactly one handler (idempotency)
- Existing logging tests pass: `./gradlew :server:test --tests "org.opensearch.common.logging.*"`

## Issues Resolved

Closes #21323


server/src/test/java/org/opensearch/common/logging/JulBridgeTests.java
```diff
@@ -0,0 +1,70 @@
...
+        boolean hasBridgeHandler = Arrays.stream(handlers).anyMatch(h -> h instanceof Log4jBridgeHandler);
+        boolean hasConsoleHandler = Arrays.stream(handlers).anyMatch(h -> h instanceof ConsoleHandler);
+
+        assertTrue("Log4jBridgeHandler should be installed on JUL root logger", hasBridgeHandler);
+        assertFalse("ConsoleHandler should be removed from JUL root logger after bridge installation", hasConsoleHandler);
+    }
+
+    public void testLog4jBridgeHandlerInstallationIsIdempotent() {
+        Log4jBridgeHandler.install(true, null, true);
+        Log4jBridgeHandler.install(true, null, true);
```
andrross: This is also testing the behavior of Log4jBridgeHandler. I'm also confused because you added an atomic boolean in LogConfigurator to guard against exactly this scenario, but then this unit test seems to be showing that it is fine to invoke the install method multiple times?

server/src/test/java/org/opensearch/common/logging/JulBridgeTests.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public void testLog4jBridgeHandlerInstallation() {
+        Log4jBridgeHandler.install(true, null, true);
+
+        Handler[] handlers = julRootLogger.getHandlers();
+        boolean hasBridgeHandler = Arrays.stream(handlers).anyMatch(h -> h instanceof Log4jBridgeHandler);
+        boolean hasConsoleHandler = Arrays.stream(handlers).anyMatch(h -> h instanceof ConsoleHandler);
+
+        assertTrue("Log4jBridgeHandler should be installed on JUL root logger", hasBridgeHandler);
```
andrross: This seems to be testing the behavior of Log4jBridgeHandler. Can you create a unit test that verifies the behavior of LogConfigurator instead?

Chocochip101: Replaced with testJulBridgeIsInstalledAfterConfigure() in EvilLoggerConfigurationTests, which calls logConfigurator.configure() and asserts that Log4jBridgeHandler is installed on the JUL root logger.

[5f57f14](https://github.com/opensearch-project/OpenSearch/pull/21359/commits/5f57f149579fb0f2a3f725474a8205fffa71a353)

server/src/test/java/org/opensearch/common/logging/JulBridgeTests.java
Chocochip101: Good catch. The test was inadvertently showing that Log4jBridgeHandler.install() is already idempotent, which made the julBridgeInstalled AtomicBoolean redundant.

 Removed the guard and replaced JulBridgeTests with a test in EvilLoggerConfigurationTests that exercises LogConfigurator.configure() directly.

[5f57f14](https://github.com/opensearch-project/OpenSearch/pull/21359/commits/5f57f149579fb0f2a3f725474a8205fffa71a353)

rursprung:
the test failure is due to the flaky tests in `EhCacheDiskCacheTests` (see #16159). it'd be great to have this land in 3.7.0

server/src/main/java/org/opensearch/common/logging/LogConfigurator.java
```diff
@@ -191,6 +193,14 @@ public FileVisitResult visitFile(final Path file, final BasicFileAttributes attr
 
         configureLoggerLevels(settings);
 
+        // Bridge java.util.logging (JUL) to Log4j2 so that libraries using JUL (e.g., Lucene) route
+        // through the configured logging framework instead of falling back to the default ConsoleHandler,
+        // which writes to stderr and would be captured at WARN level. Setting the system property
+        // "java.util.logging.manager" alone is insufficient because JUL's LogManager may already be
```
andrross, line 199: There are a number of caveats to using Log4jBridgeHandler: https://logging.apache.org/log4j/2.x/log4j-jul.html#bridge-handler

Why exactly is setting `java.util.logging.manager` not working here? Can we fix that instead of using the bridge handler? The bridge handler is intended to be used in situations where setting the system property is not possible, which is not the case here.

Chocochip101: @andrross 
AgentPolicy (from opensearch-agent.jar) calls Logger.getLogger() in its static initializer, which triggers JUL's LogManager to initialize before main() even starts. The agent's premain() runs before any OpenSearch code, so System.setProperty() is always too late — structurally the same situation the docs describe as "setting the system property is not possible."

Chocochip101: On the caveats: the main concern is overhead for disabled log statements, which we've addressed by passing propagateLevels=true. The other caveats (requires Log4j Core, no custom JUL handlers) don't apply here since OpenSearch already uses Log4j Core and doesn't configure custom JUL handlers. The Lucene JUL logging this fixes is a one-time startup message, not a hot path.

PR #21361, opened 2026-04-26
Add warm+format directory wiring with per-format tiered directory rou…

nishchay21:
---

### Description

Adds warm+format directory wiring so warm nodes with pluggable data formats (e.g., parquet) can route file operations through format-specific directories while Lucene files continue using the existing TieredDirectory (FileCache + remote metadata).

**Directory stack on warm+format:**
```
DataFormatAwareStoreDirectory (checksums, format metadata)
  → TieredSubdirectoryAwareDirectory (format routing)
    ├── TieredDirectory (Lucene: FileCache + remote)
    └── Map<String, Directory> (per-format directories from DataFormatPlugin.getTieredDirectory())
```

**Key changes:**

- `RemoteSyncAwareDirectory` — new interface for afterSyncToRemote callbacks. Implemented by CompositeDirectory, TieredSubdirectoryAwareDirectory, and DataFormatAwareStoreDirectory (pass-through).
- `TieredSubdirectoryAwareDirectory` — new directory that routes openInput, fileLength, createOutput, deleteFile, listAll, and afterSyncToRemote through format-specific directories when available, falling back to TieredDirectory for Lucene files.
- `TieredDataFormatAwareStoreDirectoryFactory` — new factory that builds the warm+format directory stack. Registered in Node.java under key "dataformat-tiered".
- `DataFormatPlugin.getTieredDirectory()` — new default method returning null. Format plugins override to provide a custom directory for warm nodes (e.g., ParquetTieredDirectory in a follow-up PR).
- `DataFormatRegistry.getTieredDirectories()` — calls the active format plugin's getTieredDirectory at shard open time.
- `DataFormatAwareStoreDirectory` — implements RemoteSyncAwareDirectory (pass-through to inner directory), adds direct-delegate constructor for warm stack.
- `DataFormatAwareStoreDirectoryFactory` — adds warm-aware overload with remoteDirectory, fileCache, threadPool params.
- `RemoteStoreUploaderService` — replaces hardcoded FilterDirectory unwrap + instanceof CompositeDirectory with generic chain walk finding RemoteSyncAwareDirectory.
- `DirectoryUtils.getFSDirectory()` — new chain walk helper replacing direct FSDirectory casts in CompositeDirectory, TieredDirectory, and DirectoryUtils.
- `IndexModule` — picks "dataformat-tiered" factory for warm+format indices.
- `IndexService.createShard()` — adds warm+format branch calling the warm-aware factory overload.

**Zero impact on hot nodes** — all new code is gated behind `isWarmIndex() && isPluggableDataFormatEnabled()` and the WRITABLE_WARM_INDEX feature flag.

**Zero impact on existing warm** — when no format plugin provides a tiered directory (null), all files route through TieredDirectory as before.

### Testing

- `TieredSubdirectoryAwareDirectoryTests` — 18 functional tests with real I/O: routing for Lucene and format files, listAll merging/dedup, deleteFile, afterSyncToRemote callbacks, createOutput, edge cases (non-existent files, close semantics, double-close prevention)
- `TieredDataFormatAwareStoreDirectoryFactoryTests` — factory stack verification, correct directory chain, no double wrapping
- `GracefulDegradationTests` — empty format directories, factory key constant, valid stack without format plugins
- `WarmShardDirectoryStackTests` — full stack creation and format directory routing
- All existing tests pass: TieredDirectoryTests, RemoteStoreUploaderServiceTests, DataFormatAwareStoreDirectoryTests, CompositeDirectoryTests

### Check List

- [x] New functionality has been documented in the PR description
- [x] Commits are signed off (DCO)
- [x] Public API changes companion pull request - N/A (experimental API)
- [x] No breaking changes


server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatPlugin.java
```diff
@@ -54,4 +56,33 @@ public interface DataFormatPlugin {
...
+     * Its {@code close()} method must NOT close {@code localDirectory} (e.g., must not call
+     * {@code super.close()} if it extends FilterDirectory wrapping localDirectory). Only clean up
+     * format-specific resources. The shared localDirectory is closed separately by TieredDirectory.
+     *
+     * @param localDirectory    the subdirectory-aware local directory for reading files from disk
+     * @param remoteDirectory   the remote segment store directory for reading files from remote
+     * @param indexSettings     the index settings for this shard
+     * @return a format-specific directory, or {@code null} to use the default TieredDirectory
+     */
+    default Directory getTieredDirectory(
```
Bukhtawar: Seems misplaced

nishchay21: The thinking here was that DataFormatPlugin is where each format declares its per-shard capabilities (indexingEngine(), getFormatDescriptors()), and a tiered directory felt like the same pattern — the format knows how to handle its files on warm nodes.

DataFormatDescriptor was kept lightweight intentionally (static metadata like format name + checksum strategy), while the directory factory needs runtime params (localDirectory, remoteDirectory) at shard open time.

That said, we could move this into the descriptor as a lazy TieredDirectoryFactory — the factory is just a lambda, so no resources are allocated at descriptor creation time. The actual directory would only be created when invoked at warm shard open. Happy to refactor this way if you think it's cleaner What do you prefer @Bukhtawar ?

server/src/main/java/org/opensearch/index/store/RemoteSyncListener.java
```diff
@@ -0,0 +1,44 @@
...
+     * Called after a file has been successfully uploaded to the remote store.
+     *
+     * <p>Implementations should use this callback to update internal state related to the
+     * file's remote availability — such as unpinning from a local cache, marking the file
+     * as remotely available in a registry, or forwarding the notification to a delegate.
+     *
+     * @param file the name of the file that was synced to remote
+     */
+    void afterSyncToRemote(String file);
+}
```
Bukhtawar, line 43: This repository doesn't make sense, can we instead model it as a listener

nishchay21: Its the previous logic reused. However will see to move composite logic also

server/src/main/java/org/opensearch/index/engine/dataformat/DataFormatRegistry.java
```diff
@@ -139,6 +142,39 @@ public Set<DataFormat> getRegisteredFormats() {
...
+     *         at most one format (the active {@code pluggable_dataformat}), but the Map return type
+     *         is forward-compatible for multi-format-per-shard support.
+     */
+    public Map<DataFormat, Directory> getTieredDirectories(
+        Directory localDirectory,
+        RemoteSegmentStoreDirectory remoteDirectory,
+        IndexSettings indexSettings
+    ) {
+        Map<DataFormat, Directory> result = new HashMap<>();
+        String dataformatName = indexSettings.getSettings().get(PLUGGABLE_DATAFORMAT_SETTING);
```
nishchay21: will change to correct settings

sandbox/plugins/native-repository-fs/build.gradle
```diff
@@ -9,7 +9,6 @@
 opensearchplugin {
   description = 'Native (Rust) ObjectStore backend for the FS repository plugin.'
   classname = 'org.opensearch.repositories.fs.native_store.FsNativeObjectStorePlugin'
-  extendedPlugins = ['repository-fs']
```
Bukhtawar, line 12: WHy is this not needed?

nishchay21: We do not hence removed it. FS repo is not a plugin in core.  FS is launched as a default plugin and mainly used for testing only

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/store/ParquetDataFormatStoreHandler.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    public ParquetDataFormatStoreHandler(ShardId shardId, boolean isWarm, NativeStoreRepository repo) {
+        if (isWarm) {
+            long remotePtr = (repo != null && repo.isLive()) ? repo.getPointer() : 0L;
+            long ptr = TieredStorageBridge.createTieredObjectStore(0L, remotePtr);
+            this.storeHandle = new NativeStoreHandle(ptr, TieredStorageBridge::destroyTieredObjectStore);
+            logger.debug("[{}] Created ParquetDataFormatStoreHandler with native store, ptr={}", shardId, ptr);
+        } else {
+            this.storeHandle = NativeStoreHandle.EMPTY;
+        }
```
Bukhtawar, line 64: Initialize as empty avoid else

mgodwan (approved):

mgodwan:
Tracking items: for next PR for the feature

```
1. Convert local/remote to enum  in DataformatStoreHandler
2. node.java <- use if condition for data format aware factories (based on feature flag)
3. Combine Common code for filter and tiered
3. StoreStrategy Abtsract class to ensure methods can't be overridden and marked final.
4. Tests with native handling of Parquet data format store handler. Can come along with ITs when we have full functionality working.
5. Fail when pointer is not live
```

PR #19650, opened 2025-10-16
fix issue 18446 NPE by making sure scoreSupplier.get() returns non-null scorer

haruki1243:
### Description
fix issue [18446](https://github.com/opensearch-project/OpenSearch/issues/18446) NPE by making sure scoreSupplier.get() returns non-null scorer

### Related Issues
Resolves [#[18446]](https://github.com/opensearch-project/OpenSearch/issues/18446)

### Check List
- [x] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21413, opened 2026-04-28
Fix NPE in FunctionScoreQuery.explain() when scorer is null

### LINKED ISSUE #21412: [BUG] NullPointerException in FunctionScoreQuery.explain() when scorer is null for leaf segment
### What is the bug?

`FunctionScoreQuery$CustomBoostFactorWeight.explain()` throws a `NullPointerException` when `functionScorer(context)` returns `null`. This happens when `explain=true` is used with a compound query (e.g. neural-search's HybridQuery) that delegates `explain` to sub-queries containing `function_score` with multiple scoring functions.

The root cause is a missing null-check at [FunctionScoreQuery.java#L475-L476](https://github.com/opensearch-project/OpenSearch/blob/3.5.0/server/src/main/java/org/opensearch/common/lucene/search/function/FunctionScoreQuery.java#L475-L476):

```java
FunctionFactorScorer scorer = functionScorer(context);
int actualDoc = scorer.iterator().advance(doc);  // NPE when scorer is null
```

`functionScorer()` returns `null` when `subQueryWeight.scorer(context)` yields no scorer for the leaf segment. The `explain` method checks the sub-query's doc-level explain (which reports a match), but does not guard against a null segment-level scorer before calling `scorer.iterator()`.

This code path is only reached when there are multiple scoring functions and more than one matches (`functionsExplanations.size() > 1 && !singleFunction`), so queries with a single function are not affected.

**Stack trace:**
```
java.lang.NullPointerException: Cannot invoke
  "org.opensearch.common.lucene.search.function.FunctionScoreQuery$FunctionFactorScorer.iterator()"
  because "scorer" is null
    at o.o.c.l.search.function.FunctionScoreQuery$CustomBoostFactorWeight.explain(FunctionScoreQuery.java:476)
    at org.apache.lucene.search.BooleanWeight.explain(BooleanWeight.java:74)
    at o.o.c.l.search.function.FunctionScoreQuery$CustomBoostFactorWeight.explain(FunctionScoreQuery.java:430)
    at o.o.neuralsearch.query.HybridQueryWeight.explain(HybridQueryWeight.java:138)
    at org.apache.lucene.search.IndexSearcher.explain(IndexSearcher.java:968)
    at org.apache.lucene.search.IndexSearcher.explain(IndexSearcher.java:944)
    at o.o.search.fetch.subphase.ExplainPhase$1.process(ExplainPhase.java:64)
    at o.o.search.fetch.FetchPhase.execute(FetchPhase.java:242)
```

### How can one reproduce the bug?

1. Create an index with multiple shards (e.g. 10)
2. Set up a search pipeline with RRF normalization
3. Run a hybrid query where at least one sub-query is a `function_score` with multiple weighted filter functions
4. Add `"explain": true` to the request

The same query without `explain` succeeds. The NPE occurs only on shards where the sub-query's scorer is null for the leaf segment containing the hit document.

Minimal query structure that triggers the bug:
```json
GET <index>/_search
{
  "explain": true,
  "query": {
    "hybrid": {
      "queries": [
        {
          "function_score": {
            "query": {
              "bool": {
                "should": [
                  {
                    "function_score": {
                      "query": { "multi_match": { "query": "test", "fields": ["title", "body"] } },
                      "functions": [
                        { "filter": { "match": { "title": "test" } }, "weight": 1.3 }
                      ]
                    }
                  }
                ],
                "filter": [
                  { "term": { "status": "active" } }
                ]
              }
            },
            "functions": [
              { "filter": { "term": { "tag": "a" } }, "weight": 10.0 },
              { "filter": { "term": { "tag": "b" } }, "weight": 5.0 }
            ],
            "score_mode": "multiply"
          }
        }
      ]
    }
  }
}
```

Search pipeline: RRF normalization (`rank_constant: 30`).

### What is the expected behavior?

`explain` should return a valid `Explanation` (match or noMatch) instead of throwing NPE. When the scorer is null for a segment, the explain path should handle this gracefully — the same way the scoring path (`scorerSupplier`, lines 417-425) already does.

### What is your host/environment?

- OpenSearch **3.5.0** (AWS Managed OpenSearch Service)
- neural-search plugin (bundled)
- Index: 10 primary shards
- Failure is partial: 2 of 10 shards fail, remaining shards return results

### Do you have any additional context?

The `scorerSupplier` path (normal scoring) already handles the null case correctly:

```java
public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
    Scorer scorer = functionScorer(context);
    if (scorer != null && minScore != null) {  // ← null-safe
        scorer = new MinScoreScorer(this, scorer, minScore);
    }
    if (scorer != null) {  // ← null-safe
        return new DefaultScorerSupplier(scorer);
    } else {
        return null;
    }
}
```

Only the `explain` path is missing this guard.

**Root cause analysis:** In a HybridQuery, `HybridQueryWeight.explain()` calls `wt.explain(context, doc)` for **every** sub-query weight — including those that didn't contribute the hit. The sub-query's `explain` is doc-based and may report a match, but `functionScorer()` creates a segment-level scorer which can be null if no documents in the leaf match. This explain/scorer mismatch causes the NPE.

Related (different bugs): opensearch-project/neural-search#1126, opensearch-project/neural-search#1415

derstrassi:
### Description

Adds a null-check for `functionScorer(context)` in `CustomBoostFactorWeight.explain()`. When the scorer is null (no matching documents in the leaf segment), the explain path now returns an `Explanation.noMatch` instead of throwing a `NullPointerException`.

This aligns the explain path with the scoring path (`scorerSupplier`), which already handles null scorers correctly.

### Issues Resolved

Resolves #21412

### Testing

Added `testExplainDoesNotThrowWhenScorerIsNull` to `FunctionScoreTests`:
- Creates a custom sub-query whose `explain()` reports a match but whose `scorerSupplier()` returns null (simulating the HybridQuery delegation scenario)
- Wraps it in a `function_score` with two filter functions to enter the multi-function code path
- Verifies that `explain()` returns a valid `Explanation` instead of throwing NPE
- **Without fix:** `NullPointerException` at `FunctionScoreQuery.java:476` (identical stack trace to the reported bug)
- **Without fix:** all existing `FunctionScoreTests` pass (no regressions)

### Check List
- [x] API changes companion pull request - n/a

### DIFF
--- a/server/src/main/java/org/opensearch/common/lucene/search/function/FunctionScoreQuery.java
+++ b/server/src/main/java/org/opensearch/common/lucene/search/function/FunctionScoreQuery.java
@@ -473,14 +473,26 @@ public class FunctionScoreQuery extends Query {
                     factorExplanation = functionsExplanations.get(0);
                 } else {
                     FunctionFactorScorer scorer = functionScorer(context);
-                    int actualDoc = scorer.iterator().advance(doc);
-                    assert (actualDoc == doc);
-                    double score = scorer.computeScore(doc, expl.getValue().floatValue());
-                    factorExplanation = Explanation.match(
-                        (float) score,
-                        "function score, score mode [" + scoreMode.toString().toLowerCase(Locale.ROOT) + "]",
-                        functionsExplanations
-                    );
+                    if (scorer != null) {
+                        int actualDoc = scorer.iterator().advance(doc);
+                        assert (actualDoc == doc);
+                        double score = scorer.computeScore(doc, expl.getValue().floatValue());
+                        factorExplanation = Explanation.match(
+                            (float) score,
+                            "function score, score mode [" + scoreMode.toString().toLowerCase(Locale.ROOT) + "]",
+                            functionsExplanations
+                        );
+                    } else {
+                        // Scorer can be null when explain is invoked from a compound query
+                        // (e.g. HybridQuery) for a doc whose sub-query reports a match via
+                        // explain but yields no scorer for this leaf segment.
+                        factorExplanation = Explanation.match(
+                            1.0f,
+                            "function score, score mode [" + scoreMode.toString().toLowerCase(Locale.ROOT)
+                                + "] (no scorer for this segment)",
+                            functionsExplanations
+                        );
+                    }
                 }
                 expl = combineFunction.explain(expl, factorExplanation, maxBoost);
             }
--- a/server/src/test/java/org/opensearch/index/query/functionscore/FunctionScoreTests.java
+++ b/server/src/test/java/org/opensearch/index/query/functionscore/FunctionScoreTests.java
@@ -46,6 +46,7 @@ import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.MatchAllDocsQuery;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.Scorer;
+import org.apache.lucene.search.ScorerSupplier;
 import org.apache.lucene.search.SortField;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.search.TopDocs;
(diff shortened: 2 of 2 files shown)


cwperks:
Similar bugfix: https://github.com/opensearch-project/OpenSearch/pull/19650

PR #21366, opened 2026-04-27
Update logic in FipsBuildParams.isInFipsApprovedOnlyMode to check for env param instead of org.bouncycastle.fips.approved_only

cwperks:
### Description

Currently, FipsBuildParam looks for the system prop (`org.bouncycastle.fips.approved_only`) to determine if FIPS is enforced at runtime. This system prop is specific to the bouncycastle library and implicitly set [here in opensearch-env](https://github.com/opensearch-project/OpenSearch/blob/main/distribution/src/bin/opensearch-env#L115-L128). This PR makes configuring FIPS more intentional by the cluster operator by looking for an OpenSearch env var instead of the bouncycastle system prop to determine if FIPS should be enforced at runtime.

### Related Issues

Related to https://github.com/opensearch-project/OpenSearch/issues/20738

### Check List
- [ ] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #20738, opened 2026-02-26
[META] Support for FIPS 140-3 enforced mode


nibix:
## Intro

The purpose of this meta issue is supporting the ongoing efforts for supporting FIPS 140-3 enforced mode in OpenSearch. It shall serve two core functions:

- to document fundamental architectural decisions 
- to document open to-do items and to keep track of them

This issue shall be a living document, we should try to keep the information here up to date. Discussions in the comments section are welcome - still, any fundamental decision in the comments section should be reflected here. 

*Note: Unfortunately, the GitHub issue permission model only allows me and maintainers to edit this issue. I would have hoped that there would have been some  way to allow more users editing, but that does not seem to be the case. I will try to do my best to keep the issue up to date. Still, if anyone has ideas on how to improve this, please speak up!*

*Note: This is just my first shot. There is a 200% chance that there are omissions and inaccuracies. Please comment or edit in order to complete this list.*

## Fundamental decisions 

- There will be only one OpenSearch distribution, which always bundles `bc-fips` but **does not** enable *FIPS approved only mode* by default.
- To enable *FIPS approved only mode*, these actions must be performed:
  - the script [`opensearch-fips-demo-installer`](https://github.com/opensearch-project/OpenSearch/blob/main/distribution/src/bin/opensearch-fips-demo-installer) needs to be executed 
  - the environment variable `OPENSEARCH_FIPS_MODE` needs to be set (this will be converted by the `opensearch-env` script into the JVM property `org.bouncycastle.fips.approved_only=true`). 
- The first version of this will be shipped with OpenSearch 3.6. However, enabling *FIPS approved only mode* will be only fully supported for the OpenSearch-min distribution (without any plugins, especially also without security plugin). Some further plugin functionality *might* work, but no guarantee is given.
- Full support, including security plugin and other plugins will be delivered in a later version of OpenSearch.

## Assumptions

*Please speak up if these assumptions are wrong.*

- All plugins in the OpenSearch distribution have been already adapted to be aware of the bundled `bc-fips` library *(This is only about the presence of the component, not about enabling FIPS approved only mode)*
- The OpenSearch min distribution already runs without issues in *FIPS approved only mode*. All necessary PRs for this have been already merged to `main`.

### Supporting References

https://github.com/opensearch-project/technical-steering/issues/77

### For OpenSearch 3.6.0:

- [x] Bundle bc-fips with distribution: https://github.com/opensearch-project/opensearch-build/pull/5995
[...]

PR #21415, opened 2026-04-28
Update logic in FIPS bootstrap check

### LINKED ISSUE #19702: [BUG] Security provider should not be forcibly removed due to a Bouncy Castle property
Security providers should not be removed due to a Bouncy Castle property.

For example, it is possible to configure Bouncy Castle FIPS libraries and the restricted Sun JCE in RHEL OpenJDK 21 in FIPS mode https://docs.redhat.com/en/documentation/red_hat_build_of_openjdk/21/html-single/configuring_red_hat_build_of_openjdk_21_on_rhel_with_fips/index#about-fips

This code can prevent the valid use of _org.bouncycastle.fips.approved_only_

https://github.com/opensearch-project/OpenSearch/blob/31aaff3f97bf176264221a5a4a22999c6adfdc7c/server/src/main/java/org/opensearch/bootstrap/Bootstrap.java#L200-L202

The check for _org.bouncycastle.fips.approved_only_ here should be removed. 

There is also an argument that the provider should not be removed at all, but honored if configured in the Java security file (rightly or wrongly). However, if an environment variable is desirable for enforcing strict requirements I propose adding a crypto standard of "FIPS-140-3-STRICT" or similar.

Initial discussion under https://github.com/opensearch-project/OpenSearch/pull/18921#discussion_r2432047789

terryquigleysas:
### Description
Currently, Bootstrap.java looks for the system prop (org.bouncycastle.fips.approved_only) to determine if FIPS is enforced at runtime. This system prop is specific to the bouncycastle library and implicitly set [here in opensearch-env](https://github.com/opensearch-project/OpenSearch/blob/main/distribution/src/bin/opensearch-env#L115-L128). This PR makes configuring FIPS more intentional by the cluster operator by looking for an OpenSearch env var instead of the bouncycastle system prop to determine if FIPS should be enforced at runtime.

Accompanies https://github.com/opensearch-project/OpenSearch/pull/21366

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/19702

Related to:
https://github.com/opensearch-project/OpenSearch/pull/21366
https://github.com/opensearch-project/OpenSearch/issues/20738

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/bootstrap/Bootstrap.java
+++ b/server/src/main/java/org/opensearch/bootstrap/Bootstrap.java
@@ -197,7 +197,9 @@ final class Bootstrap {
         );
 
         var cryptoStandard = System.getenv("OPENSEARCH_CRYPTO_STANDARD");
-        if ("FIPS-140-3".equals(cryptoStandard) || "true".equalsIgnoreCase(System.getProperty("org.bouncycastle.fips.approved_only"))) {
+        var fipsMode = System.getenv("OPENSEARCH_FIPS_MODE");
+
+        if ("FIPS-140-3".equals(cryptoStandard) || "true".equalsIgnoreCase(fipsMode)) {
             LogManager.getLogger(Bootstrap.class).info("running in FIPS-140-3 mode");
             SecurityProviderManager.removeNonCompliantFipsProviders();
             FipsTrustStoreValidator.validate();
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -31,7 +31,7 @@
 import groovy.xml.XmlParser
 
 plugins {
-  id('com.google.protobuf') version '0.9.6'
+  id('com.google.protobuf') version '0.10.0'
   id('opensearch.build')
   id('opensearch.publish')
   id('opensearch.internal-cluster-test')
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -31,7 +31,7 @@
 import groovy.xml.XmlParser
 
 plugins {
-  id('com.google.protobuf') version '0.10.0'
+  id('com.google.protobuf') version '0.9.6'
   id('opensearch.build')
   id('opensearch.publish')
   id('opensearch.internal-cluster-test')


cwperks:
@terryquigleysas thank you for this! I created https://github.com/opensearch-project/OpenSearch/pull/21366 yesterday, but missed this area.

server/src/main/java/org/opensearch/bootstrap/Bootstrap.java
```diff
@@ -197,7 +197,9 @@ private void setup(boolean addShutdownHook, Environment environment) throws Boot
         );
 
         var cryptoStandard = System.getenv("OPENSEARCH_CRYPTO_STANDARD");
-        if ("FIPS-140-3".equals(cryptoStandard) || "true".equalsIgnoreCase(System.getProperty("org.bouncycastle.fips.approved_only"))) {
+        var fipsMode = System.getenv("OPENSEARCH_FIPS_MODE");
+
+        if ("FIPS-140-3".equals(cryptoStandard) || "true".equalsIgnoreCase(fipsMode)) {
```
cwperks, line 202: Can we use `FipsBuildParams.isInFipsApprovedOnlyMode()` here?

terryquigleysas:
> @terryquigleysas thank you for this! I created #21366 yesterday, but missed this area.

Hey @cwperks , yes, I've only just seen your change from yesterday and was about to contact you with this accompanying, proposed PR but you beat me to it :-)

server/src/main/java/org/opensearch/bootstrap/Bootstrap.java
terryquigleysas: That is an AND check (which may be suitable for test needs) and this is an OR ? 
Longer term we may be able to remove the _cypto.standard_ check altogether but for this I was going for an initial small change in case I was missing something.

cwperks (approved):

server/src/main/java/org/opensearch/bootstrap/Bootstrap.java
cwperks: Marking the PR as approved, but we probably should just have a single flag here as you've built into this PR.

@beanuwave also would like to get your thoughts too.

- `-Pcrypto.standard=FIPS-140-3` is a build param, so it need to be supplied at runtime too?
- `OPENSEARCH_FIPS_MODE=true` is currently used to check if FIPS should be enforced at runtime. Should we require `-Pcrypto.standard=FIPS-140-3` too? I guess, is a boolean sufficient here or is that overly simplified?

PR #21417, opened 2026-04-28
Add failure handling for composite multi-format indexing engine

### LINKED ISSUE #21455: [Feature Request] Add failure handling for composite multi-format indexing engine

The composite indexing engine (DataFormatAwareEngine + CompositeIndexingExecutionEngine) needs deterministic failure handling when per-doc write failures occur across multiple data formats. Without it, a failure in one format (e.g., Lucene rejects an oversized term) can leave the other format (e.g., Parquet) with an extra row, breaking cross-format row-count parity.

A deterministic `rollbackTo(rowCount)` contract for multi-format write failure handling:

1. **`Writer.rollbackTo(long rowCount)`** — target-based rollback replacing the old `rollbackLastDoc()`. The caller declares the desired row count; the writer reaches it or throws. Supports exactly-1-doc rollback with a guard (production invariant).

2. **WriterState lifecycle** — `ACTIVE → PENDING_ROLLBACK → ACTIVE` (fully reversible, e.g. Parquet) or `ACTIVE → PENDING_ROLLBACK → RETIRED_FLUSHABLE → CLOSED` (counters can't reverse, e.g. Lucene). `PENDING_ROLLBACK` replaces the ambiguous `RETIRED` name.

3. **Failure paths handled:**
   - Per-doc `IllegalArgumentException` from Lucene (oversized term, missing binary value) → `WriteResult.Failure` → composite drives `rollbackTo` → writer retired with N-1 docs flushed.
   - Per-doc `MismatchedInputException` from Parquet (schema mismatch) → `WriteResult.Failure` → composite drives `rollbackTo` → writer stays ACTIVE (no row consumed).
   - Lucene `IOException` (Directory I/O fault) → tragic event → engine fails → shard failover → replica promoted.
   - Unmodeled throws → engine fails → translog recovery replays.

4. **`IndexingExecutionEngine.getTragicException()`** — tragic-exception detection moved from `Committer` to the indexing engine. Composite multiplexes across primary + secondary engines. DFAE consults the engine, not the committer.

5. **Cross-format atomicity** — `CompositeWriter.failureHandler.rollback(touched, acceptedRows)` drives `rollbackTo` on every format that touched the doc, ensuring both formats converge to the same row count.

6. **RowIdAwareWriter** — always decrements rowIdCounter on `WriteResult.Failure` (doc was never committed). `rollbackTo` throws `UnsupportedOperationException` (composite drives rollback internally on per-format delegates, not via the pool wrapper).

### Related component

Indexing:DataFormatAwareEngine

### Additional context

META: #20876
PR: #21417

alchemist51:
### Description

Adds failure handling to DataFormatAwareEngine and CompositeWriter to maintain cross-format data consistency when document write failures occur in the composite indexing engine.

 #### Testing
  
  - DataFormatAwareEngineTests — single-format failure handling (write failures, corruption, tragic events, flush failures, concurrent failures, recovery)
  - CompositeWriterFailureTests — CompositeWriter isolation (rollback, abort, flush failures)
  - CompositeEngineFailureTests — file-backed e2e with data integrity verification and writer pool lifecycle assertions
  - CompositeEngineFailureIT — cluster-level with single-format and multi-format configurations, cross-format consistency verification (CatalogSnapshot row counts, Lucene DirectoryReader numDocs, docId == rowId invariant, rowId set matching across formats)

### Related Issues
Resolves #21455

### Check List

### DIFF
--- a/libs/concurrent-queue/src/main/java/org/opensearch/common/queue/LockablePool.java
+++ b/libs/concurrent-queue/src/main/java/org/opensearch/common/queue/LockablePool.java
@@ -8,6 +8,8 @@
 
 package org.opensearch.common.queue;
 
+import org.opensearch.common.lease.Releasable;
+
 import java.io.Closeable;
 import java.io.IOException;
 import java.util.ArrayList;
@@ -134,6 +136,23 @@ public final class LockablePool<T extends Lockable> implements Iterable<T>, Clos
      */
     public synchronized boolean isRegistered(T item) {
         return items.contains(item);
+    }
+
+    /**
+     * Removes a single already-locked item from the pool. The returned {@link Releasable}
+     * unlocks the item when closed, intended for use with try-with-resources. The caller
+     * still owns the item itself and is responsible for closing any underlying resources.
+     *
+     * @param item the locked item to remove
+     * @return a releasable whose {@code close()} unlocks the item
+     * @throws IllegalArgumentException if the item is not registered in this pool
+     */
+    public synchronized Releasable checkout(T item) {
+        if (items.remove(item) == false) {
+            throw new IllegalArgumentException("Item is not registered in this pool");
+        }
+        availableItems.remove(item);
+        return item::unlock;
     }
 
     private void ensureOpen() {
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
@@ -216,6 +216,20 @@ public class LuceneCommitter extends SafeBootstrapCommitter {
         return fileName.startsWith(IndexFileNames.SEGMENTS) || fileName.equals(IndexWriter.WRITE_LOCK_NAME);
     }
 
+    @Override
+    public void markStoreCorrupted(IOException cause) {
+        if (store.tryIncRef() == false) {
+            return;
+        }
+        try {
+            store.markStoreCorrupted(cause);
+        } catch (IOException e) {
+            logger.warn("Couldn't mark store corrupted", e);
+        } finally {
+            store.decRef();
+        }
+    }
+
     /**
      * Builds upload-time Lucene {@code SegmentInfos} bytes from the {@link DirectoryReader}
      * registered for this {@code catalogSnapshot} (opened at that snapshot's refresh point).
--- a/sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeEngineHelper.java
+++ b/sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeEngineHelper.java
@@ -1,0 +1,194 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.composite;
+
+import org.apache.lucene.index.DirectoryReader;
+import org.opensearch.be.lucene.LuceneReader;
+import org.opensearch.cluster.ClusterState;
+import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.concurrent.GatedCloseable;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.index.IndexService;
+import org.opensearch.index.engine.DataFormatAwareEngine;
+import org.opensearch.index.engine.dataformat.DataFormat;
+import org.opensearch.index.engine.dataformat.DataFormatRegistry;
+import org.opensearch.index.engine.exec.IndexReaderProvider;
+import org.opensearch.index.engine.exec.Segment;
+import org.opensearch.index.engine.exec.WriterFileSet;
+import org.opensearch.index.engine.exec.coord.CatalogSnapshot;
+import org.opensearch.index.shard.IndexShard;
+import org.opensearch.index.shard.IndexShardTestCase;
+import org.opensearch.indices.IndicesService;
+import org.opensearch.plugins.PluginsService;
+import org.opensearch.test.InternalTestCluster;
+import org.opensearch.test.OpenSearchIntegTestCase;
+
+import java.io.IOException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+/**
+ * Shared utility methods for composite engine integration tests.
+ * Stateless — all methods take explicit parameters so any IT can use them.
+ */
+public final class CompositeEngineHelper {
+
+    private CompositeEngineHelper() {}
+
+    // -- Index creation --
+
+    public static void createCompositeIndex(
+        OpenSearchIntegTestCase testCase,
+        String indexName,
+        String primaryFormat,
+        String... secondaryFormats
+    ) {
+        Settings.Builder sb = Settings.builder()
+            .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
+            .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
+            .put("index.pluggable.dataformat.enabled", true)
+            .put("index.pluggable.dataformat", "composite")
+            .put("index.composite.primary_data_format", primaryFormat);
+        sb.putList("index.composite.secondary_data_formats", secondaryFormats);
+        testCase.createIndex(indexName, sb.build());
+        testCase.ensureGreen(indexName);
+    }
+
+    public static void createCompositeIndexWithMapping(
+        OpenSearchIntegTestCase testCase,
+        String indexName,
+        String primaryFormat,
+        Settings extraSettings,
+        String... secondaryFormats
+    ) {
+        Settings.Builder sb = Settings.builder()
+            .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
+            .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
+            .put("index.pluggable.dataformat.enabled", true)
+            .put("index.pluggable.dataformat", "composite")
+            .put("index.composite.primary_data_format", primaryFormat)
+            .put(extraSettings);
+        sb.putList("index.composite.secondary_data_formats", secondaryFormats);
+        OpenSearchIntegTestCase.client()
+            .admin()
+            .indices()
+            .prepareCreate(indexName)
+            .setSettings(sb)
+            .setMapping("field_keyword", "type=keyword", "field_number", "type=integer")
+            .get();
+        testCase.ensureGreen(indexName);
+    }
+
+    // -- Engine access --
+
+    public static IndexShard getPrimaryShard(ClusterService clusterService, InternalTestCluster cluster, String indexName) {
+        ClusterState state = clusterService.state();
+        String nodeId = state.routingTable().index(indexName).shard(0).primaryShard().currentNodeId();
+        String nodeName = state.nodes().get(nodeId).getName();
+        IndexService svc = cluster.getInstance(IndicesService.class, nodeName)
+            .indexServiceSafe(state.metadata().index(indexName).getIndex());
+        return svc.getShard(0);
+    }
+
+    public static DataFormatAwareEngine getEngine(ClusterService clusterService, InternalTestCluster cluster, String indexName) {
+        return (DataFormatAwareEngine) IndexShardTestCase.getIndexer(getPrimaryShard(clusterService, cluster, indexName));
+    }
+
+    // -- Row counts --
+
+    /**
+     * Returns the total live-doc count for a format by querying its actual reader.
+     *
+     * <p>For Lucene we use {@link DirectoryReader#numDocs()} which excludes soft-deleted
+     * docs (e.g., docs removed by {@code rollbackLastDoc}). For Parquet — which has no
+     * delete primitive in this codebase yet — we fall back to summing
+     * {@link WriterFileSet#numRows()} from the catalog snapshot, since the file footer's
+     * numRows IS the authoritative source for a Parquet file.
+     */
+    public static long getRowCount(ClusterService clusterService, InternalTestCluster cluster, String indexName, String formatName)
+        throws IOException {
+        DataFormatAwareEngine engine = getEngine(clusterService, cluster, indexName);
+        if ("lucene".equals(formatName)) {
+            String nodeId = clusterService.state().routingTable().index(indexName).shard(0).primaryShard().currentNodeId();
+            String nodeName = clusterService.state().nodes().get(nodeId).getName();
+            DataFormat luceneFormat = new DataFormatRegistry(cluster.getInstance(PluginsService.class, nodeName)).format(formatName);
+            try (GatedCloseable<IndexReaderProvider.Reader> ref = engine.acquireReader()) {
+                Object raw = ref.get().reader(luceneFormat);
+                if (raw == null) {
+                    return 0L;
+                }
+                if (raw instanceof LuceneReader == false) {
+                    throw new IllegalStateException(
+                        "Expected LuceneReader for format [" + formatName + "] but got " + raw.getClass().getName()
+                    );
+                }
+                return ((LuceneReader) raw).directoryReader().numDocs();
+            }
+        }
+        // Parquet path: file-footer numRows is the source of truth (no deletes for parquet today).
+        try (GatedCloseable<CatalogSnapshot> ref = engine.acquireSnapshot()) {
+            return ref.get().getSearchableFiles(formatName).stream().mapToLong(WriterFileSet::numRows).sum();
+        }
+    }
+
+    /**
+     * Engine-only overload — used when the caller already has the engine and only needs the
+     * Parquet/file-metadata path. The Lucene branch routes through
+     * {@link #getRowCount(ClusterService, InternalTestCluster, String, String)} which can
+     * resolve {@link DataFormat} via {@link DataFormatRegistry}.
+     */
+    public static long getRowCount(DataFormatAwareEngine engine, String formatName) throws IOException {
+        try (GatedCloseable<CatalogSnapshot> ref = engine.acquireSnapshot()) {
+            return ref.get().getSearchableFiles(formatName).stream().mapToLong(WriterFileSet::numRows).sum();
+        }
+    }
+
+    // -- Indexing --
+
+    public static void indexDocs(OpenSearchIntegTestCase testCase, String indexName, int count) {
+        for (int i = 0; i < count; i++) {
+            assertEquals(
+                RestStatus.CREATED,
+                OpenSearchIntegTestCase.client()
+                    .prepareIndex(indexName)
+                    .setSource(
+                        "field_keyword",
+                        OpenSearchIntegTestCase.randomAlphaOfLength(10),
+                        "field_number",
+                        OpenSearchIntegTestCase.randomIntBetween(0, 1000)
+                    )
+                    .get()
+                    .status()
+            );
+        }
+    }
+
+    // -- Flush --
+
+    public static void flush(OpenSearchIntegTestCase testCase, String indexName) {
+        OpenSearchIntegTestCase.client().admin().indices().prepareFlush(indexName).setForce(true).setWaitIfOngoing(true).get();
+    }
+
+    // -- Assertions --
+
+    public static void assertPerSegmentRowCountsMatch(DataFormatAwareEngine engine, String format1, String format2) throws IOException {
+        try (GatedCloseable<CatalogSnapshot> ref = engine.acquireSnapshot()) {
+            for (Segment seg : ref.get().getSegments()) {
+                WriterFileSet wfs1 = seg.dfGroupedSearchableFiles().get(format1);
+                WriterFileSet wfs2 = seg.dfGroupedSearchableFiles().get(format2);
+                assertNotNull("segment gen=" + seg.generation() + " missing " + format1, wfs1);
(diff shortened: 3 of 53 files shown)


sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeEngineHelper.java
```diff
@@ -0,0 +1,167 @@
...
+        return svc.getShard(0);
+    }
+
+    public static DataFormatAwareEngine getEngine(ClusterService clusterService, InternalTestCluster cluster, String indexName) {
+        return (DataFormatAwareEngine) IndexShardTestCase.getIndexer(getPrimaryShard(clusterService, cluster, indexName));
+    }
+
+    // -- Row counts --
+
+    public static long getRowCount(DataFormatAwareEngine engine, String formatName) throws IOException {
```
bharath-techie: Can we use the individual engine readers [ Lucene , parquet ] to get the row count for the catalog snapshot ?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -273,18 +280,23 @@ public DataFormatAwareEngine(EngineConfig engineConfig) {
...
-                if (initSnapshots.isEmpty() == false) {
-                    for (Segment seg : initSnapshots.getLast().getSegments()) {
+                Map<String, String> commitUserData = store.readLastCommittedSegmentsInfo().getUserData();
+                String snapshotData = commitUserData.get(CatalogSnapshot.CATALOG_SNAPSHOT_KEY);
+                if (snapshotData != null) {
+                    DataformatAwareCatalogSnapshot committed = DataformatAwareCatalogSnapshot.deserializeFromString(
+                        snapshotData,
+                        java.util.function.Function.identity()
+                    );
+                    for (Segment seg : committed.getSegments()) {
```
mgodwan: This should rely on committer

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -578,12 +592,48 @@ private Engine.IndexResult indexIntoEngine(Engine.Index index, IndexingStrategy
...
+                        }
+                    } finally {
+                        lockedWriter.unlock();
+                        IOUtils.closeWhileHandlingException(currentWriter);
+                    }
+                    assert writerPool.isRegistered(lockedWriter) == false : "retired writer must not be in pool";
+                }
+                if (treatDocumentFailureAsTragicError(index)) {
+                    failEngine("document failure on replica (id: " + index.id() + ")", f.cause());
+                }
```
mgodwan: Can we move this out to separate lifecycle management layer?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -273,18 +280,23 @@ public DataFormatAwareEngine(EngineConfig engineConfig) {
...
                 }
-            } catch (IOException e) {
+            } catch (Exception e) {
                 // Fall back to 0 on error
             }
-            this.writerGenerationCounter = new AtomicLong(maxGenFromCommit);
+            this.writerGenerationCounter = new AtomicLong(maxGenFromCommit + 1);
             this.writerPool = new LockablePool<>(() -> {
-                long gen = writerGenerationCounter.incrementAndGet();
+                long gen = writerGenerationCounter.getAndIncrement();
```
mgodwan: Why is this changed?

server/src/main/java/org/opensearch/index/engine/dataformat/Writer.java
```diff
@@ -52,4 +52,23 @@ public interface Writer<P extends DocumentInput<?>> extends Closeable {
...
+
+    /**
+     * Rolls back the last document added to this writer.
+     * Only valid immediately after {@link #addDoc} — cannot skip or rollback earlier docs.
+     * Called by composite writers when a sibling format fails after this writer already accepted the doc.
+     *
+     * @throws IOException if the rollback fails
+     * @throws IllegalStateException if no document has been added or last doc was already rolled back
+     */
+    default void rollbackLastDoc() throws IOException {}
```
mgodwan: Is this the only signal to abort? If yes, we should make it evident in the method contract

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
mgodwan: e.g. if lucene is primary, and fails, then we don't know whether doc id was consumed or not. hence, we should rollback here as well

mgodwan: essentially, writer should mark itself as aborted and then on failure, if writer marked itself as aborted, we should retire the writer.

mgodwan: This can be solved based on a strategy

libs/concurrent-queue/src/main/java/org/opensearch/common/queue/LockablePool.java
```diff
@@ -136,6 +136,20 @@ public synchronized boolean isRegistered(T item) {
...
     }
 
+    /**
+     * Removes a single already-locked item from the pool without unlocking it.
+     * The caller takes ownership and is responsible for closing the item.
+     *
+     * @param item the locked item to remove
+     * @throws IllegalArgumentException if the item is not registered in this pool
+     */
+    public synchronized void checkout(T item) {
```
mgodwan: Should this return a releasable then?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
```diff
@@ -210,6 +210,21 @@ public boolean isCommitManagedFile(String fileName) {
...
+     * returns {@link Exception}. Wraps non-Exception throwables in RuntimeException
+     * to preserve the root cause for {@code DataFormatAwareEngine.failOnTragicEvent}.
+     */
+    @Override
+    public Exception getTragicException() {
+        Throwable tragic = indexWriter.getTragicException();
+        if (tragic instanceof Exception) return (Exception) tragic;
+        if (tragic != null) return new RuntimeException(tragic);
+        return null;
+    }
```
mgodwan, line 271: This would be handled at refresh time, right? Should we be concerned about this within the write flow as this writer won't come into picture then

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdAwareWriter.java
```diff
@@ -63,7 +75,10 @@ public RowIdAwareWriter(Writer<P> delegate) {
      */
     @Override
     public WriteResult addDoc(P d) throws IOException {
-        d.setRowId(DocumentInput.ROW_ID_FIELD, rowIdCounter.getAndIncrement());
+        assert delegate.state() == WriterState.ACTIVE : "addDoc requires ACTIVE state but delegate was " + delegate.state();
+        if (d != null) {
+            d.setRowId(DocumentInput.ROW_ID_FIELD, rowIdCounter.getAndIncrement());
+        }
```
mgodwan: Should we mark this as non null? We should not get null docs ever.

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdAwareWriter.java
```diff
@@ -52,6 +52,18 @@ public RowIdAwareWriter(Writer<P> delegate) {
...
 
+    /**
+     * Returns the next row ID that would be assigned. After {@code N} successful addDoc
+     * calls this returns {@code N}. After a failed addDoc the counter is still
+     * incremented (the slot is consumed), so this returns {@code N + 1}.
+     * <p>
+     * Used by retirement-time invariant assertions to verify the leaf writers' final
+     * row counts agree with the row IDs assigned at this layer.
+     */
+    public long nextRowId() {
```
mgodwan: Why do we need to expose this?

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdAwareWriter.java
```diff
@@ -63,7 +75,10 @@ public RowIdAwareWriter(Writer<P> delegate) {
      */
     @Override
     public WriteResult addDoc(P d) throws IOException {
-        d.setRowId(DocumentInput.ROW_ID_FIELD, rowIdCounter.getAndIncrement());
+        assert delegate.state() == WriterState.ACTIVE : "addDoc requires ACTIVE state but delegate was " + delegate.state();
```
mgodwan: This should be a runtime check (like ensureOpen)

server/src/main/java/org/opensearch/index/engine/dataformat/WriterState.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Writer can no longer accept documents and its buffered content is inconsistent
+     * (e.g., rollback failed). The engine retires the writer without flushing —
+     * the buffered docs are dropped.
+     */
+    RETIRED_DROP,
+
+    /** Writer is closed. Terminal. */
+    CLOSED
```
mgodwan, line 68: Can you mention the recommended engine action in each case as well? (e.g. flush writer, fail shard)

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -203,6 +210,16 @@ public class DataFormatAwareEngine implements Indexer {
      * @param engineConfig the engine configuration
      */
     public DataFormatAwareEngine(EngineConfig engineConfig) {
+        // DataFormatAwareEngine is the primary-side engine in a segment-replication cluster.
+        // Replicas must use the read-only segment-consuming engine (e.g., NRTReplicationEngine).
+        // Fail fast at construction so a misconfiguration is caught before any indexing or recovery.
+        if (engineConfig.isReadOnlyReplica()) {
```
mgodwan: This won't be true for warm, right?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -761,6 +815,13 @@ public void refresh(String source) throws EngineException {
                         List<Segment> existingSegments = catalogSnapshot.get().getSegments();
                         List<Segment> newSegments = new ArrayList<>();
 
+                        // Drain segments from retired writers (flushed during failure handling)
+                        Segment retiredSeg;
+                        while ((retiredSeg = pendingRetiredSegments.poll()) != null) {
+                            newSegments.add(retiredSeg);
+                            refreshed = true;
+                        }
```
mgodwan: With the writer threads helping with flushes, can we flush directly here, and then refresh flow can pick from pendingRetiredSegments queue?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1333,6 +1394,23 @@ public void failEngine(String reason, @Nullable Exception failure) {
...
+                    // Mark store as corrupted before notifying listeners (same as Engine.java)
+                    if (failure != null && Lucene.isCorruptionException(failure)) {
+                        if (store.tryIncRef()) {
+                            try {
+                                store.markStoreCorrupted(
+                                    new IOException(
+                                        "failed engine (reason: [" + reason + "])",
+                                        org.opensearch.ExceptionsHelper.unwrapCorruption(failure)
+                                    )
+                                );
```
mgodwan: Should this be done via committer?

alchemist51: Moved it to committer.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
alchemist51: Added check for warmindex also

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdAwareWriter.java
alchemist51: Done

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdAwareWriter.java
alchemist51: Removed this redundant check

libs/concurrent-queue/src/main/java/org/opensearch/common/queue/LockablePool.java
alchemist51: Done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
alchemist51: I think we should still fail fast since anyway the refresh is going to fail and our translog build up will happen otherwise.

server/src/main/java/org/opensearch/index/engine/dataformat/WriterState.java
alchemist51: done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
alchemist51: Done

alchemist51:
```
 REPRODUCE WITH: ./gradlew ':server:internalClusterTest' --tests 'org.opensearch.storage.WarmIndexBasicIT.testWritableWarmPrimaryReplicaBoth' -Dtests.seed=95578561FE513EF9 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=kxv-Deva-IN -Dtests.timezone=Asia/Qatar -Druntime.java=25
  2> java.lang.AssertionError: Count is 0 hits but 1000 was expected.  Total shards: 1 Successful shards: 1 & 0 shard failures:
        at __randomizedtesting.SeedInfo.seed([95578561FE513EF9:926499785A260FD2]:0)
        at org.opensearch.test.hamcrest.OpenSearchAssertions.assertHitCount(OpenSearchAssertions.java:303)
        at org.opensearch.storage.WarmIndexBasicIT.testWritableWarmPrimaryReplicaBoth(WarmIndexBasicIT.java:309)
  2> NOTE: leaving temporary files on disk at: /var/jenkins/workspace/gradle-check/search/server/build/testrun/internalClusterTest/temp/org.opensearch.storage.WarmIndexBasicIT_95578561FE513EF9-001
  2> NOTE: test params are: codec=Asserting(Lucene104): {index_uuid=PostingsFormat(name=Asserting), type=PostingsFormat(name=Asserting)}, docValues:{}, maxPointsInLeafNode=210, maxMBSortInHeap=5.130104065135994, sim=Asserting(RandomSimilarity(queryNorm=true): {}), locale=kxv-Deva-IN, timezone=Asia/Qatar
```

unrelated failure, Test seems flaky

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/ParquetWriter.java
```diff
@@ -90,12 +93,66 @@ public ParquetWriter(
...
+        this.mappingVersion = mappingVersion;
+        this.dataFormat = dataFormat;
+        this.checksumStrategy = checksumStrategy;
+        this.vsrManager = vsrManager;
+    }
+
     @Override
     public WriteResult addDoc(ParquetDocumentInput d) throws IOException {
-        vsrManager.addDocument(d);
+        assert state == WriterState.ACTIVE : "addDoc requires ACTIVE state but was " + state;
```
mgodwan: Make this a runtime exception

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/ParquetWriter.java
```diff
@@ -90,12 +93,66 @@ public ParquetWriter(
...
+        // No state change on throw: VSR's getAcceptedRows tells rollbackLastDoc precisely
+        // whether the row was admitted, so the writer can stay ACTIVE.
+        try {
+            vsrManager.addDocument(d);
+        } catch (IOException | RuntimeException e) {
+            throw e;
+        }
```
mgodwan, line 115: not needed

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/ParquetWriter.java
```diff
@@ -90,12 +93,66 @@ public ParquetWriter(
...
     @Override
     public WriteResult addDoc(ParquetDocumentInput d) throws IOException {
-        vsrManager.addDocument(d);
+        assert state == WriterState.ACTIVE : "addDoc requires ACTIVE state but was " + state;
+        // No state change on throw: VSR's getAcceptedRows tells rollbackLastDoc precisely
+        // whether the row was admitted, so the writer can stay ACTIVE.
+        try {
+            vsrManager.addDocument(d);
+        } catch (IOException | RuntimeException e) {
+            throw e;
```
mgodwan: on any failure, 
1. should we ensure that setRowCount is set to the value earlier
2. Return a failure without updating the writer state.

caller/composite writers can then take a decision based on should rollback which checks for retired state

server/src/main/java/org/opensearch/index/engine/InternalEngine.java
```diff
@@ -1439,6 +1418,7 @@ private NoOpResult innerNoOp(final NoOp noOp) throws IOException {
                         throw ex;
                     }
                 }
+                // Importatnt
```
mgodwan: Remove this comment

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -624,21 +654,50 @@ private Engine.IndexResult indexIntoEngine(Engine.Index index, IndexingStrategy
...
+            if (lockedWriter != null) {
+                try {
+                    writerCheckedOut = retireWriterIfNeeded(lockedWriter);
+                } catch (IllegalStateException bufferLoss) {
+                    writerCheckedOut = true;
+                    failEngine("writer retirement could not preserve buffered acked docs", bufferLoss);
+                } catch (Exception retireEx) {
+                    e.addSuppressed(retireEx);
+                }
+            }
```
mgodwan: any exception should lead to fail engine on DFAE, catch nothing in Composite, always deal with writer states
Parquet auto-corrects always on failure
Given Lucene may never auto-correct -> move to retired

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/ParquetWriter.java
```diff
@@ -90,12 +93,66 @@ public ParquetWriter(
...
+        if (vsrManager.getAcceptedRows() == acceptedRows) {
+            // The in-flight doc never landed in the VSR — nothing to undo.
+            return;
+        }
+        try {
+            vsrManager.rollbackLastDoc();
+            acceptedRows--;
+        } catch (IOException | RuntimeException e) {
+            state = WriterState.RETIRED_DROP;
+            throw e;
```
mgodwan: 1. The impl can be owned by VSRManager. ParquetWriter should send it the last accepted doc?
2. Do we need to check for `vsrManager.getAcceptedRows() == acceptedRows`? Shouldn't we define by contract to have a valid state this needs to be in, and then writer handles whether it is doable?

mgodwan (changes requested):

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -300,16 +316,26 @@ public DataFormatAwareEngine(EngineConfig engineConfig) {
...
+                        java.util.function.Function.identity()
+                    );
+                    for (Segment seg : committed.getSegments()) {
                         maxGenFromCommit = Math.max(maxGenFromCommit, seg.generation());
                     }
                 }
-            } catch (IOException e) {
-                // Fall back to 0 on error
+            } catch (Exception e) {
+                // Fall back to 0 on error — engine still functions with a fresh counter.
```
mgodwan: This should already be handled. We don't need to modify this

sandbox/plugins/parquet-data-format/src/test/java/org/opensearch/parquet/engine/ParquetDataFormatAwareEngineTests.java
```diff
@@ -271,13 +274,39 @@ private Schema buildSchema() {
...
+
+    private static void addNumberField(List<Field> fields, String name, NumberFieldMapper.NumberType type) {
+        ParquetField pf = ArrowFieldRegistry.getParquetField(type.typeName());
+        fields.add(new Field(name, pf.getFieldType(), null));
+    }
+
+    private static void addByTypeName(List<Field> fields, String name, String typeName) {
+        ParquetField pf = ArrowFieldRegistry.getParquetField(typeName);
+        fields.add(new Field(name, pf.getFieldType(), null));
+    }
```
mgodwan: Why is this change needed?

alchemist51: Rolled back with the fix

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
alchemist51: Rolled back

mgodwan (approved):

PR #21418, opened 2026-04-28
Stream Arrow batches on the data-node fragment execution path

bowenlan-amzn:
### Description

Today's path drains DataFusion results into `Object[]` rows, sends one buffered response, and the coordinator converts back to Arrow. This PR replaces that with native Arrow batches over the stream transport from #21253 #21437 #21454.

When Arrow Flight stream transport is enabled, the data-node handler iterates `EngineResultStream` batches and sends each as an `ArrowBatchResponse` over the streaming channel. The coordinator receives batches directly as `VectorSchemaRoot` and feeds them into the stage's `ExchangeSink`. The row path is preserved as fallback when stream transport is not available.

#### Data flow

```mermaid
flowchart TD
    subgraph Coordinator
        HANDLER2[StreamingResponseListener]
        RESP[FragmentExecutionArrowResponse<br>extends ArrowBatchResponse]
        SFSE[ShardFragmentStageExecution]
        SINK[ExchangeSink]
    end

    subgraph Data Node
        HANDLER[StreamingFragmentHandler]
        SVC[AnalyticsSearchService]
        FR[FragmentResources<br>reader + engine + stream]
        DRS[DatafusionResultStream<br>fresh VSR per batch]
    end

    SFSE -- "dispatchFragmentStreaming<br>(STREAM type)" --> HANDLER
    HANDLER -- "executeFragmentStreaming" --> SVC
    SVC -- "startFragment" --> FR
    FR -. "iterate batches" .-> DRS
    DRS -- "FragmentExecutionArrowResponse<br>per batch" --> HANDLER
    HANDLER -- "sendResponseBatch" --> HANDLER2
    HANDLER2 -- "response.getRoot()" --> RESP
    RESP -- "VectorSchemaRoot" --> SINK
```

#### Key types

```mermaid
flowchart TD
    subgraph coordinator [Coordinator Node]
        direction TB
        DPE[DefaultPlanExecutor]
        QC[QueryContext<br>bufferAllocator per query]
        PW[PlanWalker]
        SFSE2[ShardFragmentStageExecution<br>extends AbstractStageExecution<br>implements DataProducer]
        LSE[LocalStageExecution<br>implements SinkProvidingStageExecution]
        DRS2[DatafusionReduceSink<br>implements ExchangeSink]
        ATXS[AnalyticsSearchTransportService<br>dispatchFragmentStreaming]
    end

    subgraph datanode [Data Node]
        direction TB
        ATXS2[AnalyticsSearchTransportService<br>handler]
        ASS[AnalyticsSearchService<br>service-level allocator]
        FR2[FragmentResources<br>reader + engine + stream]
        DFSE[DatafusionSearchExecEngine<br>implements SearchExecEngine]
        DFRS[DatafusionResultStream<br>fresh VSR per batch]
    end

    subgraph shared [Shared Root]
        AAP[ArrowAllocatorProvider<br>node-level RootAllocator]
    end

    DPE --> QC
    DPE --> PW
    PW --> SFSE2
    PW --> LSE
    SFSE2 --> ATXS
    LSE --> DRS2
    DRS2 -.-> QC

    ATXS --> ATXS2
    ATXS2 --> ASS
    ASS --> FR2
    FR2 --> DFSE
    DFSE --> DFRS

    AAP -.-> QC
    AAP -.-> ASS
```

#### Send side (data node)

- **Fresh VSR per batch.** `sendResponseBatch` is async — Flight's executor transfers buffers on a separate thread while the producer advances to the next batch. Reusing one VSR would race.
- **Producer does not close sent VSRs.** Flight closes both the producer VSR (after transfer into the channel's staging root) and the channel VSR (after gRPC write). Closing in the producer races the flight thread.
- **Service-level allocator.** Because sends are async, batches outlive the query. A per-query allocator would close while Flight still holds `ArrowBuf` references. `AnalyticsSearchService` owns one allocator for its lifetime, injected into engines via `ExecutionContext`.

#### Receive side (coordinator)

- **Per-query allocator from shared root.** `QueryContext.bufferAllocator()` lazily creates a child allocator (256 MB limit) used by `ExchangeSink` to hold received batches during the reduce. Closed by `DefaultPlanExecutor`'s terminal listener when the query completes or fails.
- **Shared allocator root across plugins (`ArrowAllocatorProvider`).** Arrow's `associate` check compares root identity (reference equality). Multi-shard streaming failed at `DatafusionReduceSink.feed` with *"A buffer can only be associated between two allocators that share the same root"* because `FlightTransport` and `QueryContext` had separate `RootAllocator` instances. Both now take children of a single node-level root exposed from `arrow-flight-rpc`.

#### Cross-plugin wiring

- **`extendedPlugins = ['arrow-flight-rpc']`** on analytics-engine so cross-plugin Arrow types share a classloader. Same FQN loaded by different classloaders = different Java classes; `instanceof` and casts fail silently. Overlapping jars (`arrow-*`, jackson, guava, slf4j, flatbuffers) moved to `compileOnly`.

#### `TransportService.java` change

Commit `a9606ffb` forwards `skipsDeserialization()` in the `sendRequestAsync` anonymous wrapper. PR #21454 threaded this marker through all named wrappers but missed this one. Without it, any `ArrowBatchResponseHandler` dispatched under a parent task receives a byte-serialized input and throws at `ArrowBatchResponse.<init>`.

#### Compatibility

With `transport.stream.enabled=false` (default), behavior is unchanged. `CoordinatorReduceIT` and `CoordinatorReduceMemtableIT` validate the row path. `StreamingCoordinatorReduceIT` (`@LockFeatureFlag(STREAM_TRANSPORT)`, 2-shard parquet, 20 rows) is the streaming regression gate.

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -100,58 +139,25 @@ public FragmentExecutionResponse executeFragment(FragmentExecutionRequest reques
...
-        } catch (Exception e) {
-            listener.onFragmentFailure(queryId, stageId, shardIdStr, e);
-            throw new RuntimeException("Failed to execute fragment on " + shard.shardId(), e);
-        }
+    private static ExecutionContext buildContext(
+        FragmentExecutionRequest request,
+        Reader reader,
+        FragmentExecutionRequest.PlanAlternative plan
+    ) {
+        SearchShardTask searchShardTask = null; // TODO: real task for cancellation
```
mch2: This task should be coming from transport

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
```diff
@@ -66,25 +66,20 @@ public AnalyticsSearchTransportService(
         AnalyticsSearchService searchService,
         IndicesService indicesService
     ) {
-        this.transportService = streamTransportService != null ? streamTransportService : transportService;
+        this.streamingEnabled = streamTransportService != null;
+        this.transportService = this.streamingEnabled ? streamTransportService : transportService;
```
mch2, line 69: i don't think that we need/want this fallback logic.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSearchExecEngine.java
```diff
@@ -34,6 +34,7 @@ public class DatafusionSearchExecEngine implements SearchExecEngine<ExecutionCon
 
     private final DatafusionContext datafusionContext;
     private final Supplier<BufferAllocator> allocatorFactory;
+    private BufferAllocator streamAllocator;
```
mch2: this assumes a per-query child allocator.  Are we seeing any leaks in tests? During initial testing @rishabhmaurya and I saw leak issues, we were also concerned with the frequent child allocation.

rishabhmaurya: There are 2 options after https://github.com/opensearch-project/OpenSearch/pull/21253 : 
1. Use channel allocator with each query.
2.  Create one allocator per Service and create VSR from it per query per batch. (Recommended)
Option 2 is cleaner as this way each service can track their allocator usage

bowenlan-amzn: I do see the leak during manual testing

Having AI implemented a solution for now, will review myself first

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
bowenlan-amzn: Keeping both paths for now so we can validate streaming against the row path behind the opensearch.experimental.feature.transport.stream.enabled flag. 

Once streaming is tested, these logics can go away — agreed that's the end state.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSearchExecEngine.java
bowenlan-amzn: Since the streaming send response path is async, the arrow batches may still belong to the SearchExecEngine after the engine invoke send for all the batches.

So a child allocator per SearchExecEngine won't be helpful to check the memory leak when the engine is closed.

I am going ahead with one single allocator sitting in AnalyticsSearchService. And pass this allocator to every query to account for all the memory used.

Later on, the metrics/monitoring should hook with this single allocator to monitor any memory leak.

Context of the async streaming send response path:
https://github.com/opensearch-project/OpenSearch/blob/dec02f901209d6b4b90bb4aeda76ad34f4ec13ba/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java#L138-L144

bowenlan-amzn: At the send time, streaming transport will create a new VSR using **the same allocator of the arrow data VSR that's being sent** (this allocator will be cached in the transport channel of this query).
Transfer the ownership, and then close the arrow data VSR. **So data producer side doesn't need to worry about closing it.**
Streaming transport then work with this new VSR, expecting flight/gRPC to release this VSR after wiring it off.

Context:
https://github.com/opensearch-project/OpenSearch/blob/dec02f901209d6b4b90bb4aeda76ad34f4ec13ba/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java#L156-L171

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
bowenlan-amzn: Pushed a commit https://github.com/opensearch-project/OpenSearch/pull/21418/changes/c5322096a577ea2c120553a6dd7ce3ac3e4f2527 It's small. 

I can also pull this out to another PR. I don't read in the logic of task thoroughly in this PR

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
mch2: sure, i would think a passing check here is good enough to yank the existing paths.

mch2 (approved):

sandbox/plugins/parquet-data-format/build.gradle
```diff
@@ -20,7 +20,12 @@ dependencies {
   // Apache Arrow dependencies
   implementation "org.apache.arrow:arrow-vector:${versions.arrow}"
   implementation "org.apache.arrow:arrow-memory-core:${versions.arrow}"
-  implementation "org.apache.arrow:arrow-memory-unsafe:${versions.arrow}"
+  // Arrow 18.1's default AllocationManager is Netty. arrow-memory-netty + its buffer-patch
```
rishabhmaurya, line 23: i'm guessing you will be replacing these once you have the arrow allocator plugin from https://github.com/opensearch-project/OpenSearch/pull/21465

rishabhmaurya (approved):

sandbox/plugins/parquet-data-format/build.gradle
bowenlan-amzn: yes will do

bowenlan-amzn:
Breaking change check failure is because of https://github.com/opensearch-project/OpenSearch/pull/21143#issuecomment-4382798336 
Not related to this PR

PR #21422, opened 2026-04-28
Adding Lucene Merge supports for pluggable dataformat support

darjisagar7:
# Add Lucene Merge Support for Pluggable Data Format (Composite Engine)

## Summary

This PR adds merge support for the pluggable data format framework, enabling composite indexes (Parquet primary + Lucene secondary) to merge segments while maintaining cross-format document consistency via a `RowIdMapping` protocol. The primary format (Parquet) drives the merge and produces a mapping that secondary formats (Lucene) consume to reorder their documents identically.

## Problem

In a composite index with Parquet as the primary data format and Lucene as the secondary, each format independently writes segments during indexing. Over time, segment count grows and must be reduced via merges. The challenge is:

1. Both formats must merge the **same set of segments** atomically.
2. After merge, documents in both formats must be in the **same physical order** so that a row ID in Parquet corresponds to the same document in Lucene.
3. The merge and refresh paths must not race — a reader must never see a partially-applied merge.

### Architecture

```
┌─────────────────────────────────────────────────────────────────────┐
│  DataFormatAwareEngine (server)                                     │
│  ┌─────────────┐   ┌──────────────────┐   ┌─────────────────────┐  │
│  │MergeScheduler│──▶│ CompositeMerger  │──▶│ CatalogSnapshotMgr  │  │
│  └─────────────┘   └──────────────────┘   └─────────────────────┘  │
│                            │                                        │
│              ┌─────────────┼─────────────┐                          │
│              ▼                            ▼                          │
│  ┌────────────────────┐      ┌─────────────────────────┐           │
│  │ Parquet Merger      │      │ Lucene Merger            │           │
│  │ (Primary — produces │      │ (Secondary — consumes    │           │
│  │  RowIdMapping)      │      │  RowIdMapping)           │           │
│  └────────────────────┘      └─────────────────────────┘           │
└─────────────────────────────────────────────────────────────────────┘
```

### RowIdMapping Protocol

The `RowIdMapping` is the contract between primary and secondary formats during merge:

```
┌──────────────────────────────────────────────────────────────┐
│  PackedRowIdMapping (memory-efficient via PackedLongValues)   │
│                                                              │
│  Structure:                                                  │
│    mapping[offset + oldRowId] = newRowId                     │
│                                                              │
│  Example: merge generations [5, 0, 3]                        │
│    gen 5 (2 rows): offset=0, mapping[0]=2, mapping[1]=3     │
│    gen 0 (3 rows): offset=2, mapping[2]=0, mapping[3]=4...  │
│    gen 3 (1 row):  offset=5, mapping[5]=5                   │
│                                                              │
│  Lookup: newRowId = mapping.get(genOffset[gen] + oldRowId)   │
└──────────────────────────────────────────────────────────────┘
```

### Merge Flow (Parquet Primary, Lucene Secondary)

```mermaid
sequenceDiagram
    participant MS as MergeScheduler
    participant CM as CompositeMerger
    participant PM as ParquetMerger (Primary)
    participant Rust as Rust/Native (FFM)
    participant LM as LuceneMerger (Secondary)
    participant IW as MergeIndexWriter
    participant CSM as CatalogSnapshotManager

    MS->>CM: triggerMerges()
    CM->>CM: selectSegmentsToMerge()
    
    Note over CM: Phase 1: Primary merge (produces RowIdMapping)
    CM->>PM: merge(MergeInput{segments, newGeneration})
    PM->>Rust: mergeParquetFilesInRust(inputFiles, outputFile)
    Rust->>Rust: Sorted merge with row-ID tracking
    Rust-->>PM: MergeFilesResult{metadata, RowIdMapping}
    PM-->>CM: MergeResult{parquet→WriterFileSet, RowIdMapping}

    Note over CM: Phase 2: Secondary merge (consumes RowIdMapping)
    CM->>LM: merge(MergeInput{segments, rowIdMapping, newGeneration})
    LM->>LM: Find matching Lucene segments by writer_generation attr
    LM->>LM: Set RowIdMapping on RowIdRemappingSortField
    LM->>IW: executeMerge(RowIdRemappingOneMerge, newGeneration)
    
    Note over IW: mergeInit → mergeMiddle → warmer(refreshLock) → commitMerge
    IW->>IW: wrapForMerge → RowIdRemappingCodecReader (remaps values)
    IW->>IW: MultiSorter uses IndexSort to reorder docs
    IW-->>LM: merged SegmentCommitInfo
    LM-->>CM: MergeResult{lucene→WriterFileSet, RowIdMapping}

    Note over CM: Phase 3: Atomic catalog update
    CM->>CSM: applyMergeResults(mergeResult, oneMerge)
    CSM->>CSM: Remove old segments, add merged segment
    CSM-->>CM: New CatalogSnapshot visible to readers
```

### Refresh-Lock Coordination (Race Condition Fix)

```mermaid
sequenceDiagram
    participant RT as Refresh Thread
    participant MT as Merge Thread
    participant RL as refreshLock (ReentrantLock)
    participant IW as IndexWriter Monitor
    participant CSM as CatalogSnapshotManager

    Note over RT,CSM: Normal refresh path ordering: refreshLock → IW monitor
    RT->>RL: lock()
    RT->>IW: addIndexes() [acquires IW monitor internally]
    RT->>CSM: update catalog
    RT->>RL: unlock()

    Note over MT,CSM: Merge path (with MergedSegmentWarmer hook)
    MT->>IW: executeMerge() → mergeMiddle (heavy I/O, no locks)
    MT->>RL: lock() [via MergedSegmentWarmer / preMergeCommitHook]
    MT->>IW: commitMerge() [acquires IW monitor internally]
    Note over MT: refreshLock → IW monitor (same order as refresh!)
    MT->>CSM: applyMergeChanges() [catalog update]
    MT->>RL: unlock()
```

### Lucene Secondary Merge — Document Reordering

```mermaid
sequenceDiagram
    participant LM as LuceneMerger
    participant OM as RowIdRemappingOneMerge
    participant CR as RowIdRemappingCodecReader
    participant DVP as RowIdRemappingDocValuesProducer
    participant SF as IndexSort (SortedNumericSortField)
    participant MS as MultiSorter

    LM->>OM: new RowIdRemappingOneMerge(segments, rowIdMapping)
    
    Note over OM,CR: wrapForMerge: remap VALUES written to output
    OM->>CR: new RowIdRemappingCodecReader(reader, mapping, gen, offset)
    CR->>DVP: getDocValuesReader() → remapped ___row_id values
    DVP->>DVP: For each doc: newRowId = mapping.getNewRowId(oldRowId, generation)

    Note over SF,MS: IndexSort: determine document ORDER
    MS->>SF: getIndexSorter() → reads original row IDs, remaps via RowIdMapping
    MS->>MS: Build DocMaps that reorder ALL data (stored fields, postings, doc values)
    
    Note over LM: Result: merged segment has docs in same order as Parquet
```

## Key Classes

| Class | Module | Role |
|-------|--------|------|
| `LuceneMerger` | analytics-backend-lucene | Lucene `Merger` impl — orchestrates merge via `MergeIndexWriter` |
| `LuceneMergeStrategy` | analytics-backend-lucene | Strategy interface (Primary vs Secondary behavior) |
| `SecondaryLuceneMergeStrategy` | analytics-backend-lucene | Creates `RowIdRemappingOneMerge`, passes through mapping |
| `PrimaryLuceneMergeStrategy` | analytics-backend-lucene | (Stub) For future Lucene-as-primary support |
| `RowIdRemappingOneMerge` | analytics-backend-lucene | Custom `OneMerge` that wraps readers with `RowIdRemappingCodecReader` |
| `RowIdRemappingCodecReader` | analytics-backend-lucene | `FilterCodecReader` that remaps `___row_id` doc values in output |
| `RowIdRemappingDocValuesProducer` | analytics-backend-lucene | Intercepts `___row_id` field reads and applies `RowIdMapping` |
| `PackedRowIdMapping` | server (dataformat) | Memory-efficient `RowIdMapping` impl using `PackedLongValues` |
| `MergeIndexWriter` | lucene-merge lib | Exposes Lucene's internal `merge(OneMerge)` path |
| `CommitterConfig` | server | Now carries `preMergeCommitHook` for refresh-lock coordination |
| `LuceneCommitter` | analytics-backend-lucene | Wires hook into `MergedSegmentWarmer` on `IndexWriterConfig` |
| `LuceneReaderManager` | analytics-backend-lucene | Validates reader/catalog consistency after refresh |

## Commit Breakdown

| # | Commit | Description |
|---|--------|-------------|
| 1 | `71e0632` | Core merge infrastructure: `LuceneMerger`, `RowIdRemappingOneMerge`, `RowIdRemappingCodecReader`, `RowIdRemappingDocValuesProducer`, `RowIdRemappingSortField`, `MergeIndexWriter`, `PackedRowIdMapping` |
| 2 | `85da2a8` | Refactor: Extract `LuceneMergeStrategy` interface with `PrimaryLuceneMergeStrategy` and `SecondaryLuceneMergeStrategy` implementations. Remove `RowIdRemappingSortField` (replaced by standard `SortedNumericSortField`) |
| 3 | `a00259a` | End-to-end RowIdMapping: Rust sorted merge produces mapping → FFM bridge passes it to Java → `PackedRowIdMapping` moved to server module → `CompositeMergeIT` integration test for Parquet+Lucene |
| 4 | `e7c9c90` | Race condition fix: `preMergeCommitHook` via `MergedSegmentWarmer` for refresh-lock coordination, `LuceneReaderManager` consistency assertions, `lucene-merge` extracted as shared library |

## How the RowIdMapping Flows Through the System

```
Parquet Rust Merge (sorted)
    │
    │  Tracks: for each row emitted, which file (generation) and
    │          which position within that file it came from
    │
    ▼
FFM Bridge (parquet_merge_files → Java)
    │
    │  Returns: long[] mapping, long[] genKeys, int[] genOffsets, int[] genSizes
    │
    ▼
PackedRowIdMapping (Java, memory-efficient)
    │
    │  API: getNewRowId(oldRowId, oldGeneration) → newRowId
    │
    ├──▶ RowIdRemappingCodecReader (remaps VALUES in merged output)
    │       └── RowIdRemappingDocValuesProducer
    │              └── For ___row_id field: return mapping.getNewRowId(originalValue, generation)
    │
    └──▶ IndexSort / SortedNumericSortField (determines document ORDER)
            └── MultiSorter reads remapped row IDs → builds DocMaps → reorders all data
```

## Testing

- **Unit tests**: `LuceneMergerTests`, `PackedRowIdMappingTests`
- **Integration test**: `CompositeMergeIT.testParquetPrimaryLuceneSecondaryMerge()`
  - Verifies segment count reduction
  - Validates both formats present in catalog
  - Checks Parquet row count == Lucene doc count
  - Asserts Lucene `___row_id` values are monotonically increasing (proves RowIdMapping applied)
  - Tests post-merge ingestion still works

## Risk & Limitations

- `PrimaryLuceneMergeStrategy` is a stub — Lucene-as-primary merge is not yet implemented
- `MergeIndexWriter` uses reflection to access `IndexWriter.segmentInfos` (package-private) — fragile across Lucene upgrades
- Merge is gated behind system property `opensearch.experimental.feature.pluggable.dataformat.merge.enabled=true`
- The `@AwaitsFix` annotation on the IT indicates known flakiness being stabilized


sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
```diff
@@ -219,13 +224,24 @@ private IndexWriterConfig createIndexWriterConfig(EngineConfig engineConfig) {
         }
         iwc.setRAMBufferSizeMB(engineConfig.getIndexingBufferSize().getMbFrac());
         iwc.setUseCompoundFile(engineConfig.useCompoundFile());
-        if (engineConfig.getIndexSort() != null) {
+
+        // Determine if Lucene is a secondary format in a composite setup.
+        // When secondary, use RowIdRemappingSortField so MultiSorter can reorder documents
+        // by remapped ___row_id during merge. When primary (or standalone), use the
+        // engine config's IndexSort (which may be user-configured).
+        List<String> secondaryFormats = engineConfig.getIndexSettings().getSettings().getAsList("index.composite.secondary_data_formats");
```
mgodwan, line 258: This shoudln't leak here

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
```diff
@@ -219,13 +224,24 @@ private IndexWriterConfig createIndexWriterConfig(EngineConfig engineConfig) {
...
+        if (isSecondary) {
+            iwc.setIndexSort(new Sort(new RowIdRemappingSortField("___row_id")));
+        } else if (engineConfig.getIndexSort() != null) {
             iwc.setIndexSort(engineConfig.getIndexSort());
         }
         iwc.setCommitOnClose(false);
         iwc.setOpenMode(IndexWriterConfig.OpenMode.CREATE_OR_APPEND);
         iwc.setIndexDeletionPolicy(deletionPolicy);
         iwc.setMergePolicy(NoMergePolicy.INSTANCE);
+        iwc.setMergeScheduler(new SerialMergeScheduler());
```
mgodwan, line 270: Why is this serial? This can slow down the merges a lot. We should multiple merges to happen concurrently.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -100,6 +109,9 @@ public LuceneWriter(long writerGeneration, LuceneDataFormat dataFormat, Path bas
         IndexWriterConfig iwc = analyzer != null ? new IndexWriterConfig(analyzer) : new IndexWriterConfig();
         iwc.setOpenMode(IndexWriterConfig.OpenMode.CREATE);
         iwc.setRAMBufferSizeMB(RAM_BUFFER_SIZE_MB);
+        if (indexSort != null) {
+            iwc.setIndexSort(indexSort);
```
mgodwan, line 110: In secondary mode, this should never be null

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -0,0 +1,185 @@
...
+ * </ol>
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class LuceneMerger implements Merger {
+
+    private static final Logger logger = LogManager.getLogger(LuceneMerger.class);
+
+    static final String ROW_ID_FIELD = "___row_id";
```
mgodwan: Use DocumentInput.ROW_ID_FIELD

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
```diff
@@ -0,0 +1,71 @@
...
+     * merged segment is cleaned up by IndexFileDeleter's reference counting.
+     *
+     * <p>Duplicate segment prevention is handled by the caller; this method does not
+     * validate against concurrent merges on the same segments.
+     *
+     * @param oneMerge       the merge to execute
+     * @param mergeGeneration the writer generation for the merged output segment
+     * @throws IOException if the merge fails
+     */
+    public void executeMerge(MergePolicy.OneMerge oneMerge, long mergeGeneration) throws IOException {
```
mgodwan, line 75: If this is not overriding any existing method, why do we need to implement a MergeIndexWriter?

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
```diff
@@ -0,0 +1,71 @@
...
+        synchronized (this) {
+            oneMerge.mergeGen = mergeGeneration;
+            oneMerge.isExternal = false;
+            oneMerge.maxNumSegments = -1;
+            oneMerge.registerDone = true;
+        }
+        // merge() must be called without holding the lock — mergeInit asserts !Thread.holdsLock(this)
+        merge(oneMerge);
```
mgodwan, line 86: We should override mergeSuccess as well to ensure that merged segments are only made available to index writer only if merge across all formats succeed. otherwise lucene can have a state where merge has succeeded but subsequent ones have failed, which may cause inconsistency in reader views.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -0,0 +1,185 @@
...
+            throw new IOException("Failed to access IndexWriter segmentInfos via reflection", e);
+        }
+
+        if (segmentInfos.size() == 0) {
+            logger.warn("No segments in IndexWriter — skipping merge");
+            return new MergeResult(Map.of());
+        }
+
+        List<SegmentCommitInfo> matchingSegments = new ArrayList<>();
+        for (SegmentCommitInfo sci : segmentInfos.asList()) {
```
mgodwan: No need to materialize as another list. It is already an iterable

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -0,0 +1,185 @@
...
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class LuceneMerger implements Merger {
+
+    private static final Logger logger = LogManager.getLogger(LuceneMerger.class);
+
+    static final String ROW_ID_FIELD = "___row_id";
+    static final String WRITER_GENERATION_ATTR = "writer_generation";
```
mgodwan: Use this from LuceneWriter

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -0,0 +1,185 @@
...
+                matchingSegments.add(sci);
+            }
+        }
+
+        if (matchingSegments.isEmpty()) {
+            logger.warn("No segments found matching writer generations {} — skipping merge", generationsToMerge);
+            return new MergeResult(Map.of());
+        }
+
+        logger.info(
```
mgodwan: Make this a debug log.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -0,0 +1,185 @@
...
+            "LuceneMerger: merging {} segments (generations {}) using merge(OneMerge) + IndexSort",
+            matchingSegments.size(),
+            generationsToMerge
+        );
+
+        // Set the RowIdMapping on the sort field so getIndexSorter() returns remapped
+        // values for MultiSorter to build DocMaps
+        if (rowIdMapping != null && sortField != null) {
+            sortField.setRowIdMapping(rowIdMapping);
```
mgodwan: This seems to be not thread safe. Multiple threads can modify this. 
1. We should maintain a map of OneMerge->RowIdMapping inside the sort field
2. Clean it on merge finish. (better to move to an executor layer where this init/destruction is taken care of as part of the lifecycle)
3. Add tests for concurrent merges. They should trigger multiple merges in parallel, and verify the data with new row id at the end.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -0,0 +1,185 @@
...
+
+            logger.info(
+                "LuceneMerger: completed merge of {} segments at generation {} ({} docs, {} files)",
+                matchingSegments.size(),
+                mergeInput.newWriterGeneration(),
+                mergedDocCount,
+                mergedInfo.files().size()
+            );
+
+            return new MergeResult(Map.of(dataFormat, mergedFileSet), rowIdMapping);
```
mgodwan: For lucene in primary mode, we should create a new row id mapping

mgodwan: For lucene in primary mode, we should create a new row id mapping.
this can be split into primary/secondary merger subclasses.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
darjisagar7: Done

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
darjisagar7: We are using `merge` function of IndexWriter which is protected method

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
mgodwan: This seems to be not thread safe. Multiple threads can modify this. 
1. We should maintain a map of OneMerge->RowIdMapping inside the sort field
2. Clean it on merge finish. (better to move to an executor layer where this init/destruction is taken care of as part of the lifecycle)
3. Add tests for concurrent merges. They should trigger multiple merges in parallel, and verify the data with new row id at the end.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
mgodwan: For lucene in primary mode, we should create a new row id mapping.
this can be split into primary/secondary merger subclasses.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
darjisagar7: The `SerialMergeScheduler` does **not** slow down merges or prevent concurrency in this architecture.

**The scheduler is never invoked for actual merges.** `MergeScheduler.merge()` is only called when Lucene auto-triggers merges via `MergePolicy`. Since we set `NoMergePolicy.INSTANCE`, Lucene never auto-triggers merges — `SerialMergeScheduler.merge()` is dead code.

**Our merges bypass the scheduler entirely.** `MergeIndexWriter.executeMerge()` calls `IndexWriter.merge(OneMerge)` directly, which goes through `mergeMiddle` without involving the scheduler's merge dispatch. Multiple `executeMerge()` calls can run concurrently from the engine's own thread pool.

**Why it's needed:** `mergeMiddle` calls `mergeScheduler.wrapForMerge(merge, directory)` internally. `ConcurrentMergeScheduler.wrapForMerge()` has a hard assertion (`throw new AssertionError`) that the calling thread must be a `ConcurrentMergeScheduler.MergeThread`. Since our merges run on the engine's thread pool (not Lucene's `MergeThread`), `ConcurrentMergeScheduler` would always fail. `SerialMergeScheduler.wrapForMerge()` simply returns the directory unchanged with no thread check.

**Concurrency is controlled by the engine's merge thread pool**, not by Lucene's scheduler. Parallel merges are fully supported.

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
darjisagar7: This part is already handle in the MergeHandler, where in any engine failure will rollback the merges happened in the previous engine

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
darjisagar7: added a TODO, will take this up in follow up PR

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
mgodwan: how? merged segments would have already added to parent writer.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/RowIdRemappingDocValuesProducer.java
```diff
@@ -0,0 +1,209 @@
...
+        public long cost() {
+            return delegate.cost();
+        }
+    }
+
+    /**
+     * Assigns sequential {@code ___row_id} = {@code rowIdOffset + docID}.
+     * Used when no RowIdMapping is provided.
+     */
+    private static class SequentialRowIdDocValues extends SortedNumericDocValues {
```
mgodwan: Remove this for now

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/RowIdRemappingDocValuesProducer.java
```diff
@@ -0,0 +1,209 @@
...
+        return delegate.getNumeric(field);
+    }
+
+    @Override
+    public SortedNumericDocValues getSortedNumeric(FieldInfo field) throws IOException {
+        if (DocumentInput.ROW_ID_FIELD.equals(field.name)) {
+            if (rowIdMapping != null) {
+                return new MappedRowIdDocValues(delegate.getSortedNumeric(field), rowIdMapping, generation);
+            }
+            return new SequentialRowIdDocValues(maxDoc, rowIdOffset);
```
mgodwan: throw here instead

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
darjisagar7: This is removed

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
darjisagar7: Created a Issue to make this changes when Lucene will be the primary engine
https://github.com/opensearch-project/OpenSearch/issues/21505

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: In Secondary mode, we wanted to keep child lucene index writer to get the sort as null, such that, rowmappings from parquet are going to be used for reordering the index (which expects the index to be created with no index sort).

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
```diff
@@ -59,6 +65,19 @@
  * The store reference is incremented on construction and decremented on {@link #close()}.
  * Closing the committer also closes the underlying IndexWriter.
  *
+ * <h2>Refresh-lock coordination</h2>
```
mgodwan, line 68: Move this comment to warmer

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneReaderManager.java
```diff
@@ -72,11 +77,60 @@ public void afterRefresh(boolean didRefresh, CatalogSnapshot catalogSnapshot) th
         }
         DirectoryReader refreshed = DirectoryReader.openIfChanged(currentReader);
         if (refreshed != null) {
+            // Guard against refresh/merge-apply races: a prior IT regression surfaced when
```
mgodwan, line 80: Please fix the comment to highlight the usefulness rather than the process with which it was discovered

mgodwan (approved):

PR #21426, opened 2026-04-29
Fix stream transport example plugin

bowenlan-amzn:
### Description
The example plugin used `api project(':plugins:arrow-flight-rpc')` which bundles flight's jars into the example plugin's zip. In a real cluster, each plugin gets its own classloader, so ArrowBatchResponse and ArrowFlightChannel loaded by the example plugin are different classes from those loaded by arrow-flight-rpc. This causes instanceof checks to fail at runtime.

https://github.com/opensearch-project/OpenSearch/blob/dec02f901209d6b4b90bb4aeda76ad34f4ec13ba/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/ArrowFlightChannel.java#L47

https://github.com/opensearch-project/OpenSearch/blob/dec02f901209d6b4b90bb4aeda76ad34f4ec13ba/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java#L156

```
java.lang.IllegalArgumentException: Channel is not backed by an ArrowFlightChannel: org.opensearch.transport.TaskTransportChannel
  at ArrowFlightChannel.from(ArrowFlightChannel.java:62)
  at TransportNativeArrowStreamDataAction.handleStreamRequest(TransportNativeArrowStreamDataAction.java:79)
```

Fix: use `extendedPlugins = ['arrow-flight-rpc']` with `compileOnly` so the example plugin delegates to flight's classloader. Both plugins see the same Arrow and flight classes.

Also adds:
- REST endpoint (/_native_arrow_test) to trigger the native Arrow streaming path from curl for manual verification
- run.gradle support for example-plugins so they can be installed via the standard -PinstalledPlugins mechanism

### Related Issues

### Check List

### DIFF
--- a/gradle/run.gradle
+++ b/gradle/run.gradle
@@ -63,6 +63,8 @@ testClusters {
           plugin(':plugins:' + p)
         } else if (project.findProject(':sandbox:plugins:' + p) != null) {
           plugin(':sandbox:plugins:' + p)
+        } else if (project.findProject(':example-plugins:' + p) != null) {
+          plugin(':example-plugins:' + p)
         } else {
           // attempt to fetch it from maven
           project.repositories.mavenLocal()
--- a/plugins/examples/stream-transport-example/build.gradle
+++ b/plugins/examples/stream-transport-example/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'opensearch.opensearchplugin'
 apply plugin: 'opensearch.internal-cluster-test'
+apply plugin: 'opensearch.yaml-rest-test'
 
 opensearchplugin {
   name = 'stream-transport-example'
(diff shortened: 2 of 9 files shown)


rishabhmaurya:
thanks @bowenlan-amzn! this makes sense and we may want to move ahead with new extended plugin lets say `arrow-opensearch`, which should have all arrow and netty allocators dependencies. Both `arrow-flight-rpc` and df plugin should be taking a dependency on it for shared classloader.

bowenlan-amzn:
> thanks @bowenlan-amzn! this makes sense and we may want to move ahead with new extended plugin lets say `arrow-opensearch`, which should have all arrow and netty allocators dependencies. Both `arrow-flight-rpc` and df plugin should be taking a dependency on it for shared classloader.

Will use this PR to investigate how to separate arrow deps and common transport API into a base plugin. So both arrow-flight-rpc and analytic-engine can base on.

For the anaytic-engine, will go ahead with `extendedPlugin=arrow-flight-rpc` approach for now to unblock if needed.

PR #21431, opened 2026-04-29
Add per-plugin native memory tracking via mimalloc heaps

himshikhagupta:
### Description
Adds per-plugin native memory tracking for Rust-based sandbox plugins (parquet, datafusion) using mimalloc heaps.

Architecture:

    HeapRoutingAllocator — global Rust allocator that routes allocations to per-plugin mimalloc heaps via thread-local state
    HeapGuard — RAII guard that scopes a thread to a plugin's heap for the duration of an FFM call
    Java-side NativeLibraryLoader API exposes per-heap heapUsed/heapCommitted and global globalCommitted stats

How it works:

    Each plugin (parquet, datafusion) creates a named mimalloc heap at startup via create_heap
    Every FFM entry point wraps its body in a scoped_thread_heap guard that sets the thread-local active heap
    All allocations within that scope route through the plugin's heap; allocations outside any scope use the default system allocator
    Java queries heap stats by plugin index — the Rust side walks the heap's pages via mi_heap_visit_blocks

### Check List


sandbox/libs/dataformat-native/rust/common/src/error.rs
```diff
@@ -24,12 +25,14 @@ pub fn into_error_ptr(msg: String) -> i64 {
 }
 
 /// Returns a pointer to the null-terminated error message.
+#[ffm_thread]
```
alchemist51, line 28: Can we add a linter or a test which makes sure going forward ffm_thread is used whereever applicable?

alchemist51: Checking online, I can see we can clippy linters or github workflows for this

sandbox/libs/dataformat-native/rust/common/src/heap_allocator.rs
```diff
@@ -0,0 +1,507 @@
...
+
+    #[test]
+    fn test_multiple_plugins_independent() {
+        let heap_a = create_heap("test-multi-A");
+        let heap_b = create_heap("test-multi-B");
+        let ptr_a = unsafe { heap_alloc(heap_a, 128 * 1024) };
+        let ptr_b = unsafe { heap_alloc(heap_b, 256 * 1024) };
+        let stats_a = heap_stats(heap_a);
+        let stats_b = heap_stats(heap_b);
+        assert!(stats_a.used >= 128 * 1024);
```
alchemist51, line 432: Why > ? Will it include some delta always?

sandbox/libs/dataformat-native/rust/common/src/heap_allocator.rs
```diff
@@ -0,0 +1,507 @@
...
+//!
+//! [`HeapRoutingAllocator`] is the `#[global_allocator]`. It checks a
+//! thread-local [`ACTIVE_HEAP`] on every allocation:
+//! - If set → `mi_heap_malloc_aligned(heap, ...)` (tracked by plugin)
+//! - If null → `System` allocator (libc malloc, safe on JVM threads)
+//!
+//! The System default path is critical for JVM interop: when the native
+//! library is loaded via dlopen/FFM, JVM threads call into Rust without
+//! mimalloc TLS being initialized. Using System for the default path
+//! avoids SIGSEGV in `mi_thread_init`. Once `#[ffm_safe]` spawns a Rust
```
alchemist51, line 24: Just for my understanding, only initially it will be called from System default right? Post the mimalloc TLS is initialized it's going to be directly through mimalloc?

sandbox/libs/dataformat-native/rust/common/src/heap_allocator.rs
```diff
@@ -0,0 +1,507 @@
...
+#[inline(always)]
+fn get_active_heap() -> *mut mi_heap_t {
+    ACTIVE_HEAP.try_with(|c| c.get()).unwrap_or(std::ptr::null_mut())
+}
+
+// ── Custom global allocator ─────────────────────────────────────────────────
+
+pub struct HeapRoutingAllocator;
+
+unsafe impl GlobalAlloc for HeapRoutingAllocator {
```
alchemist51, line 71: Do we have perf number of this GlobalAlloc?

sandbox/libs/dataformat-native/rust/common/src/heap_allocator.rs
```diff
@@ -0,0 +1,507 @@
...
+    if heap.ptr.is_null() {
+        return HeapStats::default();
+    }
+
+    struct Acc {
+        used: usize,
+        committed: usize,
+    }
+
+    unsafe extern "C" fn visitor(
```
alchemist51, line 240: Will this not require a ffm macro?

sandbox/libs/dataformat-native/rust/macros/src/lib.rs
```diff
@@ -28,7 +28,103 @@
...
+/// Wraps an `extern "C"` FFM function body in `std::thread::spawn(...).join()`
+/// so that all Rust code (including allocations) runs on a Rust-owned thread
+/// with properly initialized TLS. This prevents SIGSEGV in mimalloc when the
+/// native library is loaded via dlopen into a JVM process.
+///
+/// Raw pointer parameters and return types are transmitted as `usize` to
+/// satisfy `Send`. For void functions, panics in the spawned thread are
+/// silently ignored. For functions returning a value, panics produce 0.
+#[proc_macro_attribute]
+pub fn ffm_thread(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
alchemist51, line 42: Do we have tests for these?

sandbox/libs/dataformat-native/rust/Cargo.toml
```diff
@@ -46,6 +46,7 @@ log = "0.4"
 
 # Allocator
 mimalloc = { version = "0.1.48", default-features = false }
+libmimalloc-sys = { version = "0.1.44", features = ["extended"] }
```
alchemist51, line 49: Can we add a line above on what extended do?

sandbox/plugins/parquet-data-format/src/main/rust/src/ffm.rs
```diff
@@ -36,6 +38,7 @@ pub unsafe extern "C" fn parquet_create_writer(
     file_len: i64,
     schema_address: i64,
 ) -> i64 {
+    let _guard = heap_allocator::scoped_thread_heap(get_parquet_heap());
```
alchemist51, line 41: Do we want to do something like this:
```
 fn with_parquet_heap_guard<F, R>(f: F) -> R
  where
      F: FnOnce() -> R,
  {
      let _guard = heap_allocator::scoped_thread_heap(get_parquet_heap());
      f()
  }
```
  Then every functions becomes something like this:
```
  pub fn decode_page() -> Result<Page> {
      with_parquet_heap_guard(|| {
          // actual logic
      })
  }
```

We can make this general convention in the code to make sure every function goes through it!

Also I'm interested to know what happens when some function fails to add this? Will it bypass it or throw an exception from the memory allocator?

sandbox/plugins/parquet-data-format/src/main/rust/src/ffm.rs
```diff
@@ -133,11 +140,53 @@ pub unsafe extern "C" fn parquet_get_file_metadata(
...
+#[no_mangle]
+pub extern "C" fn parquet_init_heap() {
+    get_parquet_heap();
+}
+
+/// Set the calling thread's active heap to parquet's heap.
+#[ffm_thread]
+#[no_mangle]
+pub extern "C" fn parquet_set_thread_heap() {
+    heap_allocator::set_thread_heap(get_parquet_heap());
```
alchemist51, line 176: Why this is needed? How this we consumed by others?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetDataFormatPlugin.java
```diff
@@ -88,6 +89,7 @@ public Collection<Object> createComponents(
     ) {
         this.settings = clusterService.getSettings();
         this.threadPool = threadPool;
+        RustBridge.initHeap();
```
alchemist51, line 92: What about analytics-backend-datafusion? Why do we not have this?

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/HeapTrackingWithIndexingIT.java
```diff
@@ -0,0 +1,176 @@
...
+            .clear()
+            .setIndexing(true)
+            .get()
+            .getIndex(INDEX_NAME)
+            .getShards()[0];
+        assertEquals(50, shardStats.getStats().indexing.getTotal().getIndexCount());
+
+        long parquetAfter = NativeLibraryLoader.heapUsed(parquetIdx);
+        long growth = parquetAfter - parquetBefore;
+        assertTrue("parquet heap should grow by at least 512 bytes after indexing 50 docs, actual growth=" + growth, growth >= 512);
```
alchemist51, line 93: Can you add the calculation for 512?

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/HeapTrackingWithIndexingIT.java
```diff
@@ -0,0 +1,176 @@
...
+        assertTrue("global committed should be > 0", globalCommitted > 0);
+
+        long sumCommitted = 0;
+        for (int i = 0; i < NativeLibraryLoader.heapCount(); i++) {
+            sumCommitted += NativeLibraryLoader.heapCommitted(i);
+        }
+        assertTrue("sum committed (" + sumCommitted + ") <= global (" + globalCommitted + ")", sumCommitted <= globalCommitted);
+    }
+
+    public void testCrossPluginHeapIsolation() {
```
alchemist51, line 114: Why this test is here? Should it not lie in the commons?

sandbox/plugins/analytics-backend-datafusion/rust/src/runtime_manager.rs
```diff
@@ -39,6 +46,7 @@ impl RuntimeManager {
             .thread_name("datafusion-cpu")
             .enable_all()
             .on_thread_start(move || {
+                heap_allocator::set_thread_heap(df_heap);
```
alchemist51, line 49: Do we need to do this post the mimalloc TLS have been initialised as well?

alchemist51 (commented):
Thanks @himshikhagupta  for the PR! Still need to go over the heap_allocator, will do that in next go. Few comments I had:
1. I think we need more scrutized tests here, like the reported usage by OS vs our calculation comparisons. Edge cases handling, threads throwing error and how memory tracking is changing etc.
2. We need a way to enforce the memory tracking, it could be a CI testing for it or compile time failure whichever you think is better.
3. We need isolated micro benchmark for the new global allocator also to understand the impact of it in different cases. We will also need our integration benchmark for the allocator and overall memory tracking frameword delta on OS

himshikhagupta:
Closing this PR since we are moving to jemalloc.

PR #21433, opened 2026-04-29
Enable refresh listeners for DataFormatAwareEngine

mgodwan:
### Description
Enable refresh listeners for DataFormatAwareEngine

### Check List


server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -728,10 +738,11 @@ public void refresh(String source) throws EngineException {
                                     + existingSegments.size()
                                     + " but got "
                                     + result.refreshedSegments().size();
+                            notifyRefreshListenersBefore();
                             catalogSnapshotManager.commitNewSnapshot(result.refreshedSegments());
-
+                            notifyRefreshListenersRefreshed(true);
```
alchemist51: Can we use refreshed here? or let's use true in the refreshReaders(refreshed, newSnapshotRef.get());

mgodwan: Changed the way it is functioning so this should not apply anymore. Can you take a look at the updated changes if they take care of the concerns?

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshotManager.java
```diff
@@ -195,6 +198,9 @@ public synchronized void commitNewSnapshot(List<Segment> refreshedSegments) {
             latestCatalogSnapshot.getLastWriterGeneration() + 1,
             latestCatalogSnapshot.getUserData()
         );
+        for (CatalogSnapshotLifecycleListener listener : snapshotListeners) {
```
alchemist51, line 201: Do you think we should do it at the end of this function? Just in case this fails but we already notify the listeners for example the indexFileDeleter throwing an exception?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -729,12 +743,10 @@ public void refresh(String source) throws EngineException {
                                     + existingSegments.size()
                                     + " but got "
                                     + result.refreshedSegments().size();
-                            catalogSnapshotManager.commitNewSnapshot(result.refreshedSegments());
 
-                            try (GatedCloseable<CatalogSnapshot> newSnapshotRef = catalogSnapshotManager.acquireSnapshot()) {
```
alchemist51, line 734:
```
            List<CatalogSnapshotLifecycleListener> snapshotListeners = new ArrayList<>();
            for (Map.Entry<DataFormat, EngineReaderManager<?>> entry : readerManagers.entrySet()) {
                filesListeners.put(entry.getKey().name(), entry.getValue());
                snapshotListeners.add(entry.getValue());
            }
```

Will snapshotListeners always be a superset of refresh listeners? Do you think we should pass the engineConfig in the CatalogSnapshotManager otherwise to initialise all the listeners inside it?

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshotManager.java
mgodwan: this needs to happen before catalog snapshot is updated in the manager.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
mgodwan: both are divergent sets.

PR #21435, opened 2026-04-29
Add cluster-scope defaults for pluggable dataformat settings

### LINKED ISSUE #21439: [Feature Request] Resolve cluster-default pluggable dataformat settings from live ClusterSettings at index creation

The pluggable indexing introduces two `Property.Final` index-level settings:

- `index.pluggable.dataformat.enabled`
- `index.pluggable.dataformat`

Operators who want all newly created indices on a cluster to opt into pluggable dataformat by default must currently set these on every create-index request or maintain an index template. There is no cluster-scope knob that (a) expresses the intent once, and (b) honors `PUT /_cluster/settings` updates so the default can be changed at runtime.

The sandbox composite-engine plugin has the same gap for `index.composite.primary_data_format` and `index.composite.secondary_data_formats`.

Introduce cluster-scope sibling settings, declared `Property.NodeScope + Property.Dynamic` and registered in `ClusterSettings.BUILT_IN_CLUSTER_SETTINGS`:

Server (registered in IndicesService):
- cluster.default.index.pluggable.dataformat.enabled   (default: false)
- cluster.default.index.pluggable.dataformat           (default: "")

Composite-engine plugin (registered via Plugin#getSettings):
- cluster.default.index.composite.primary_data_format
- cluster.default.index.composite.secondary_data_formats

Resolution happens at index-creation time from the live `ClusterSettings` registry:

- Server: a helper in `MetadataCreateIndexService#aggregateIndexSettings` stamps the effective cluster default into `indexSettingsBuilder` when the index-level setting is not already set by the request or a template. Gated by the pluggable dataformat feature flag.
- Plugin: an `IndexSettingProvider` registered via `Plugin#getAdditionalIndexSettingProviders()` does the same for the two `index.composite.*` settings, reading live `ClusterSettings` through `ClusterService` captured in `createComponents`.

### Related component

Indexing

_No response_

### Additional context

Parent META: #20876

alchemist51:
### Description
The PR introduces cluster level settings for data format aware engine. For more details check the issue: #21439 

### Related Issues
Resolves #21439 
Part of meta: #20876 

### Check List

### DIFF
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
@@ -192,6 +192,9 @@ public class CompositeDataFormatPlugin extends Plugin implements DataFormatPlugi
         return Collections.singletonList(new IndexSettingProvider() {
             @Override
             public Settings getAdditionalIndexSettings(String indexName, boolean isDataStreamIndex, Settings templateAndRequestSettings) {
+                if (clusterService == null) {
+                    return Settings.EMPTY;
+                }
                 ClusterSettings clusterSettings = clusterService.getClusterSettings();
                 boolean restrict = clusterSettings.get(CLUSTER_INDEX_RESTRICT_COMPOSITE_DATAFORMAT_SETTING);
                 String clusterPrimary = clusterSettings.get(CLUSTER_DEFAULT_PRIMARY_DATA_FORMAT);
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
@@ -152,6 +152,7 @@ public class CompositeDataFormatPlugin extends Plugin implements DataFormatPlugi
         Setting.Property.NodeScope,
         Setting.Property.Dynamic
     );
+
     public CompositeDataFormatPlugin() {}
 
     @Override
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
@@ -61,6 +61,14 @@ import java.util.function.Supplier;
  *       formats (default empty)</li>
  * </ul>
  *
+ * <p>And three cluster settings:
+ * <ul>
+ *   <li>{@code cluster.composite.primary_data_format} — cluster-level default for the primary format</li>
+ *   <li>{@code cluster.composite.secondary_data_formats} — cluster-level default for secondary formats</li>
+ *   <li>{@code cluster.restrict.composite.dataformat} — when true, rejects index-level overrides that
+ *       differ from the cluster defaults</li>
+ * </ul>
+ *
  * <p>Format plugins (e.g., Parquet) extend this plugin by declaring
  * {@code extendedPlugins = ['composite-engine']} in their {@code build.gradle}
  * and implementing {@link DataFormatPlugin}.
@@ -134,7 +142,7 @@ public class CompositeDataFormatPlugin extends Plugin implements DataFormatPlugi
      * {@link #PRIMARY_DATA_FORMAT} and {@link #SECONDARY_DATA_FORMATS} settings.
      *
      * <p>This is scoped to the composite plugin so restriction can be toggled independently of the server-level
-     * {@code cluster.index.restrict.pluggable.dataformat} flag that governs the core
+     * {@code cluster.restrict.pluggable.dataformat} flag that governs the core
      * {@code index.pluggable.dataformat.*} settings.
      */
     public static final Setting<Boolean> CLUSTER_RESTRICT_COMPOSITE_DATAFORMAT_SETTING = Setting.boolSetting(
@@ -181,7 +189,7 @@ public class CompositeDataFormatPlugin extends Plugin implements DataFormatPlugi
      *
      * <p>Because both index settings are {@link Setting.Property#Final}, the effective value is
      * resolved once at index-creation time from the live {@link ClusterSettings} registry and
-     * frozen into the index metadata. Later updates to the {@code cluster.default.*} settings
+     * frozen into the index metadata. Later updates to the {@code cluster.composite.*} settings
      * affect only indices created after the update.
      *
      * <p>If {@link #createComponents} has not run yet (e.g. during early bootstrap), the provider
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
@@ -206,8 +206,8 @@ public class CompositeDataFormatPlugin extends Plugin implements DataFormatPlugi
                 }
                 ClusterSettings clusterSettings = clusterService.getClusterSettings();
 
-                List<String> skiplist = clusterSettings.get(IndicesService.CLUSTER_PLUGGABLE_DATAFORMAT_RESTRICT_SKIPLIST);
-                if (skiplist.stream().anyMatch(indexName::startsWith)) {
+                List<String> allowlist = clusterSettings.get(IndicesService.CLUSTER_PLUGGABLE_DATAFORMAT_RESTRICT_ALLOWLIST);
+                if (allowlist.stream().anyMatch(indexName::startsWith)) {
                     return Settings.EMPTY;
                 }
 
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
@@ -31,6 +31,7 @@ import org.opensearch.index.engine.dataformat.IndexingExecutionEngine;
 import org.opensearch.index.engine.dataformat.StoreStrategy;
 import org.opensearch.index.shard.IndexSettingProvider;
 import org.opensearch.indices.IndexCreationException;
+import org.opensearch.indices.IndicesService;
 import org.opensearch.plugins.ExtensiblePlugin;
 import org.opensearch.plugins.Plugin;
 import org.opensearch.repositories.RepositoriesService;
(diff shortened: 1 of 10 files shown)


Bukhtawar (commented):
How do we prevent users from overriding the index settings

server/src/main/java/org/opensearch/indices/IndicesService.java
```diff
@@ -304,6 +304,28 @@ public class IndicesService extends AbstractLifecycleComponent
...
+        Property.Dynamic
+    );
+
+    /**
+     * Cluster-level default for {@code index.pluggable.dataformat}.
+     * Applied at index creation time when the index setting is not explicitly provided.
+     */
+    public static final Setting<String> CLUSTER_PLUGGABLE_DATAFORMAT_VALUE_SETTING = Setting.simpleString(
+        "cluster.pluggable.dataformat",
+        "",
```
bharath-techie, line 328: I understand that we are keeping this as string to keep it string.

How are we validating the strings that they're part of the supported ones ?

server/src/main/java/org/opensearch/indices/IndicesService.java
```diff
@@ -366,6 +388,20 @@ public class IndicesService extends AbstractLifecycleComponent
         Property.Final
     );
 
+    /**
+     * If enabled, this setting enforces that indexes will be created with pluggable data-format settings matching the
+     * cluster-level defaults defined in {@code cluster.default.index.pluggable.dataformat.enabled} and
```
bharath-techie:
```
by rejecting any request that specifies an index-level value
     * that does not match
```
But i think we are simply rejecting the request if there is any specified value right , regardless if it matches or not ? https://github.com/opensearch-project/OpenSearch/pull/21435/changes#diff-ea2e29ab98d6120bb1c62077f504340d6465e5178b885631170f45d209838668R1849 ref

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
```diff
@@ -77,12 +103,144 @@ public class CompositeDataFormatPlugin extends Plugin implements DataFormatPlugi
         Setting.Property.Final
     );
 
+    /**
+     * Cluster-level default for {@code index.composite.primary_data_format}.
+     * When the index setting is not explicitly provided, this cluster setting is used as the fallback.
+     */
+    public static final Setting<String> CLUSTER_PRIMARY_DATA_FORMAT = Setting.simpleString(
```
bharath-techie, line 119: How are we validating these settings, primary and secondary settings. Can a user keep primary as lucene and secondary as parquet ?

server/src/main/java/org/opensearch/indices/IndicesService.java
bharath-techie: ALso pluggable data format can be either `composite`, `parquet` or `lucene` based on current code, 
do we need to restrict that somehow to `composite`.. do we see need of pluggable data format being just parquet or lucene ?

server/src/main/java/org/opensearch/indices/IndicesService.java
alchemist51: The code block adds validation error when the request has the index setting and it is not same as the cluster setting

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
alchemist51: Opened a issue for tracking: #21509

alchemist51:
```
curl -X PUT 'localhost:9200/test-index' -H 'Content-Type: application/json' -d '{
    "settings": {
      "index.number_of_shards": 1,
      "index.number_of_replicas": 0,
      "index.composite.secondary_data_formats": []
    }
  }'
{"error":{"root_cause":[{"type":"validation_exception","reason":"Validation Failed: 1: index setting [index.composite.secondary_data_formats] cannot differ from cluster default [lucene] when [cluster.restrict.composite.dataformat=true];"}],"type":"validation_exception","reason":"Validation Failed: 1: index setting [index.composite.secondary_data_formats] cannot differ from cluster default [lucene] when [cluster.restrict.composite.dataformat=true];"},"status":400}%
➜  ~ curl -X PUT "http://localhost:9200/_cluster/settings" \
-d '{
  "persistent": {
    "cluster.pluggable.dataformat.restrict.skiplist": ["test"],
    "cluster.pluggable.dataformat": "composite"
  }
}'
{"acknowledged":true,"persistent":{"cluster":{"pluggable":{"dataformat.restrict":{"skiplist":["test"]},"dataformat":"composite"}}},"transient":{}}%
➜  ~ curl -X PUT 'localhost:9200/test-index' -H 'Content-Type: application/json' -d '{
    "settings": {
      "index.number_of_shards": 1,
      "index.number_of_replicas": 0,
      "index.composite.secondary_data_formats": []
    }
  }'
{"acknowledged":true,"shards_acknowledged":true,"index":"test-index"}%
```

Verified the new skip list flow as well @Bukhtawar  @bharath-techie

server/src/main/java/org/opensearch/indices/IndicesService.java
alchemist51: Will cover in the same #21509

server/src/test/java/org/opensearch/cluster/metadata/MetadataCreateIndexServiceTests.java
```diff
@@ -2184,6 +2186,237 @@ public void testRefreshIntervalValidationFailureWithIndexSetting() {
...
+    }
+
+    // ---- skiplist tests ----
+
+    @LockFeatureFlag(PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG)
+    public void testUpdatePluggableDataFormatSettingsSkipsWhenIndexMatchesSkiplist() {
+        Settings clusterBag = Settings.builder()
+            .put(IndicesService.CLUSTER_PLUGGABLE_DATAFORMAT_ENABLED_SETTING.getKey(), true)
+            .put(IndicesService.CLUSTER_PLUGGABLE_DATAFORMAT_VALUE_SETTING.getKey(), "parquet")
+            .putList(IndicesService.CLUSTER_PLUGGABLE_DATAFORMAT_RESTRICT_SKIPLIST.getKey(), ".system", ".kibana")
```
mgodwan: Why .kibana is here? We should only need opensearch_dashboards?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
```diff
@@ -77,12 +104,154 @@ public class CompositeDataFormatPlugin extends Plugin implements DataFormatPlugi
...
+        "cluster.composite.primary_data_format",
+        "lucene",
```
mgodwan: Lets keep parquet as default for both primary index and cluster level setting

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -1419,6 +1423,41 @@ public static void updateRemoteStoreSettings(
...
+        }
+
+        final Settings current = settingsBuilder.build();
+
+        if (IndexSettings.PLUGGABLE_DATAFORMAT_ENABLED_SETTING.exists(current) == false) {
+            settingsBuilder.put(
+                IndexSettings.PLUGGABLE_DATAFORMAT_ENABLED_SETTING.getKey(),
+                clusterSettings.get(IndicesService.CLUSTER_PLUGGABLE_DATAFORMAT_ENABLED_SETTING)
+            );
+        }
```
mgodwan, line 1451: Should it not be the case that if setting exists at cluster level, we should skip using the index level, and only rely on cluster level?

server/src/test/java/org/opensearch/cluster/metadata/MetadataCreateIndexServiceTests.java
alchemist51: It's a list, we can add any number of prefix in it.

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeDataFormatPlugin.java
alchemist51: Done

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
alchemist51: In existing framework, we override for index settings over the cluster settings.

server/src/main/java/org/opensearch/indices/IndicesService.java
```diff
@@ -366,6 +388,33 @@ public class IndicesService extends AbstractLifecycleComponent
...
+    );
+
+    /**
+     * A list of index name prefixes that bypass the pluggable data-format restrict validation and
+     * cluster-default stamping. Indices whose name starts with any of these prefixes will not have
+     * cluster defaults applied and will not be rejected by the restrict setting.
+     */
+    public static final Setting<List<String>> CLUSTER_PLUGGABLE_DATAFORMAT_RESTRICT_SKIPLIST = Setting.listSetting(
+        "cluster.pluggable.dataformat.restrict.skiplist",
+        Collections.emptyList(),
```
Bukhtawar: nit: skiplist is a datastructure, lets use something like an `allow-list`

server/src/main/java/org/opensearch/indices/IndicesService.java
```diff
@@ -366,6 +388,33 @@ public class IndicesService extends AbstractLifecycleComponent
...
+        s -> s,
+        Property.NodeScope,
+        Property.Dynamic
```
Bukhtawar: should this be static instead?

server/src/main/java/org/opensearch/indices/IndicesService.java
alchemist51: Addressed

server/src/main/java/org/opensearch/indices/IndicesService.java
alchemist51: done

mgodwan (approved):

server/src/test/java/org/opensearch/cluster/metadata/MetadataCreateIndexServiceTests.java
```diff
@@ -2184,6 +2186,433 @@ public void testRefreshIntervalValidationFailureWithIndexSetting() {
...
+            .put(IndicesService.CLUSTER_PLUGGABLE_DATAFORMAT_VALUE_SETTING.getKey(), "parquet")
+            .build();
+        ClusterSettings cs = new ClusterSettings(clusterBag, ClusterSettings.BUILT_IN_CLUSTER_SETTINGS);
+
+        Settings.Builder indexSettingsBuilder = Settings.builder();
+        MetadataCreateIndexService.updatePluggableDataFormatSettings(indexSettingsBuilder, cs, "test-index");
+
+        Settings out = indexSettingsBuilder.build();
+        assertTrue(IndexSettings.PLUGGABLE_DATAFORMAT_ENABLED_SETTING.get(out));
+        assertEquals("parquet", IndexSettings.PLUGGABLE_DATAFORMAT_VALUE_SETTING.get(out));
```
bowenlan-amzn, line 2221: link

PR #21436, opened 2026-04-29
Add timestamp filter support for queries in analytics core

vinaykpud:
### Description
PPL filter queries on timestamp columns (e.g., `where created_at > "2024-01-01T00:00:00Z"`) failed during Substrait conversion because the SQL plugin emits a TIMESTAMP(varchar) function call that isthmus cannot convert becuase its a UDF defined in SQL plugin.

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/TimestampLiteralReduceRule.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Folds {@code TIMESTAMP(string_literal)} into a {@code TIMESTAMP(3)} literal in filter conditions.
+ *
+ * <p>TODO: Also handle TIMESTAMP literals in Project (e.g., {@code | eval ts = TIMESTAMP('2024-01-01')}).
+ * Add a second rule matching {@code Project.class} or generalize to match any RelNode.
+ *
+ * @opensearch.internal
+ */
+public class TimestampLiteralReduceRule extends RelOptRule {
```
expani: Can we look at triggering this from the FilterMarkingRule itself ? 

Since, it's already going through each RexNode within the FilterPredicate

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
```diff
@@ -0,0 +1,127 @@
...
+     * Rel-level visitor. Handles structural rewrites (table name stripping) and delegates
+     * expression rewrites to {@link PlanExpressionVisitor}.
+     */
+    private static class PlanRelVisitor extends RelCopyOnWriteVisitor<RuntimeException> {
+
+        private final PlanExpressionVisitor expressionVisitor = new PlanExpressionVisitor(this);
+
+        private static final String CATALOG_PREFIX = "opensearch";
+
+        // Strip "opensearch" catalog prefix: ["opensearch", "index_name"] -> ["index_name"]
```
expani: This can be centrally done in Scan Marking rule itself at Calcite level.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
```diff
@@ -0,0 +1,127 @@
...
+     * The base class handles recursion into function arguments, casts, if-then, etc.
+     */
+    private static class PlanExpressionVisitor extends ExpressionCopyOnWriteVisitor<RuntimeException> {
+
+        PlanExpressionVisitor(PlanRelVisitor relVisitor) {
+            super(relVisitor);
+        }
+
+        // Isthmus hardcodes timestamp literals to precision 6 (microseconds).
+        // Parquet stores Timestamp(MILLISECOND), so convert to precision 3.
```
expani, line 88: Shouldn't we get this from Index Mappings on what the precision is ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
```diff
@@ -0,0 +1,127 @@
...
+        @Override
+        public Optional<Expression> visit(Expression.PrecisionTimestampLiteral pts, EmptyVisitationContext ctx) {
+            if (pts.precision() != 3) {
+                return Optional.of(
+                    ImmutableExpression.PrecisionTimestampLiteral.builder()
```
expani, line 93: Looks like we are converting the Literal value in Predicate of the Filter operator.

If we are already examining the RexNode in Calcite, can this conversion get triggered centrally ? 

How to convert can still stay `toMillis`  in the DataFusion backend plugin

vinaykpud: I agree with you.
We convert to Millisecond precision at the calcite level in the planner. 
isthmus hardcodes precision 6 (microseconds) on all timestamp literals, regardless of what Calcite declares. We produce a correct TIMESTAMP(3) RexLiteral at the Calcite level, isthmus then ignores the precision and outputs microseconds.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
vinaykpud: Yes, but i havent considered it for now, will enhance.
The precision should come from the field's mapping type ie date -> 3, date_nanos ->9

At the Substrait rewriter level, I think we don't have access to the index mapping(I need to check).  The precision info would need to be carried through.

This was the reason why I used UDT approach earlier so that it stored in `OpenSearchTableScan` node's metadata.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
vinaykpud: Sure, we can move it

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/TimestampLiteralReduceRule.java
vinaykpud: Makes sense

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/CapabilityRegistry.java
```diff
@@ -62,6 +63,7 @@ public class CapabilityRegistry {
     private final Map<DelegationType, List<String>> delegationAcceptors = new HashMap<>();
     private final Map<FullTextParamKey, Set<String>> fullTextParamIndex = new HashMap<>();
 
+    private final List<RexNodeTransformer> rexTransformers = new ArrayList<>();
```
expani: I think we can get a backend by it's name and then extract the list of transformers. 

That would allow us to skip caching this.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTableScanRule.java
```diff
@@ -90,15 +100,102 @@ public void onMatch(RelOptRuleCall call) {
             );
         }
 
+        RelOptTable indexNameTable = new IndexNameTable(scan.getTable(), tableName);
```
expani, line 95: Thought over this a bit more, I think we can have this transformer in SQL/PPL plugin before they dispatch the relnode https://github.com/opensearch-project/sql/blob/feature/mustang-ppl-integration/plugin/src/main/java/org/opensearch/sql/plugin/rest/RestUnifiedQueryAction.java#L115 


So, this stays outside of AnalyticsPlugin and we always get only the IndexName as the table name

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/TimestampFunctionTransformer.java
```diff
@@ -0,0 +1,128 @@
...
+
+    @Override
+    public RexNode transform(RexNode node, RexBuilder rexBuilder, FieldMappingLookup fieldMappingTypes) {
+        int precision = resolveTimestampPrecision(node, fieldMappingTypes);
+        if (precision < 0) {
+            return node;
+        }
+        return node.accept(new RexShuttle() {
+            @Override
+            public RexNode visitCall(RexCall call) {
```
expani: Why are we visting the rexCall again if the marking rule has visited it and giving the exact timestamp function here ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTableScanRule.java
```diff
@@ -90,15 +100,102 @@ public void onMatch(RelOptRuleCall call) {
...
             )
         );
     }
+
+    /**
+     * Wraps a {@link RelOptTable} to override the qualified name with just the bare index name.
+     * Isthmus reads {@code getQualifiedName()} when creating {@code NamedScan} — this ensures
+     * the Substrait plan contains only the index name, not the Calcite catalog prefix.
+     */
+    private static class IndexNameTable implements RelOptTable {
```
expani: Does extending `extends RelOptAbstractTable` help resolve not overriding all these methods ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -111,22 +115,69 @@ public void onMatch(RelOptRuleCall call) {
...
-                annotatedOperands.add(annotateCondition(operand, fieldStorageInfos, childViableBackends));
+                annotatedOperands.add(
+                    annotateConditionRecursive(operand, fieldStorageInfos, childViableBackends, rexBuilder, transformers, lookup)
+                );
             }
             return rexCall.clone(rexCall.getType(), annotatedOperands);
         }
-        List<String> viableBackends = resolveViableBackends(rexCall, fieldStorageInfos, childViableBackends);
-        return new AnnotatedPredicate(rexCall.getType(), rexCall, viableBackends, context.nextAnnotationId());
+        RexCall transformed = applyRexTransformers(rexCall, transformers, rexBuilder, lookup);
```
expani: I think the transformation needs to be triggered above when we are annotating individual RexNode, that way we don't need a second traversal in the backends again

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/RexNodeTransformer.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Transforms a RexNode, potentially replacing function calls with simplified forms.
+     *
+     * @param node              the RexNode to transform
+     * @param rexBuilder        for creating new RexNodes
+     * @param fieldMappingTypes lookup for field mapping types by field index
+     * @return the transformed node, or the original if no transformation applies
+     */
+    RexNode transform(RexNode node, RexBuilder rexBuilder, FieldMappingLookup fieldMappingTypes);
```
mch2: do we need to send the whole node - or can we send `RexCall` 

```
    default RexCall handleProjectCall(RexCall call, RexBuilder rexBuilder) {
        return call;
    }
```

mch2: nvm, you need field metadata

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -111,22 +115,69 @@ public void onMatch(RelOptRuleCall call) {
...
     /**
      * Recursively walks the condition tree. Boolean connectives (AND, OR, NOT) are
-     * preserved — we recurse into their children. Leaf predicates are wrapped in
-     * {@link AnnotatedPredicate} with viable backends resolved from child's field storage.
+     * preserved — we recurse into their children. Leaf predicates are:
+     * 1. Transformed — backend-registered {@link RexNodeTransformer}s are applied.
+     * 2. Annotated — wrapped in {@link AnnotatedPredicate} with viable backends.
      */
-    private RexNode annotateCondition(RexNode condition, List<FieldStorageInfo> fieldStorageInfos, List<String> childViableBackends) {
+    private RexNode annotateCondition(
```
expani: Apologies for the late comment. 

Realized one thing that annotations here contain multiple backends so changing the RexNode means we are deciding which backend it is. But, that happens during Plan Forking. 

So, here if we add something extra to the annotation that can be read in stripAnnotation during fragment conversion and then converted to appropriate RexNode it would make sense. 

I am having to do the same for Filter Delegation

expani (approved):
Thanks @vinaykpud LGTM

mch2 (approved):

PR #21437, opened 2026-04-30
Transfer Flight batch to a response-owned root before returning

bowenlan-amzn:
### Description

FlightTransportResponse.nextResponse() returned a reference to the shared root held by Arrow's FlightStream. 
FlightStream reuses one root across batches — the next call to next() clears it and re-loads (FlightStream.java:273), and stream.close() releases its vectors. 

The streaming handler could dispatch onStreamResponse asynchronously to the search executor, so by the time the listener reads the VSR, the shared root has already been wiped. The coordinator saw valueCount=0 on every column despite a correct rowCount and schema.

### Related Issues

### Check List

### DIFF
--- a/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
+++ b/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
@@ -121,9 +121,10 @@ class FlightTransportResponse<T extends TransportResponse> implements StreamTran
             boolean hasNext = firstBatchConsumed ? flightStream.next() : (firstBatchConsumed = true);
             if (!hasNext) return null;
 
-            VectorSchemaRoot root = flightStream.getRoot();
-            currentBatchSize = FlightUtils.calculateVectorSchemaRootSize(root);
-            try (VectorStreamInput input = new VectorStreamInput(root, namedWriteableRegistry)) {
+            VectorSchemaRoot sharedRoot = flightStream.getRoot();
+            currentBatchSize = FlightUtils.calculateVectorSchemaRootSize(sharedRoot);
+            VectorSchemaRoot ownedRoot = transferToOwnedRoot(sharedRoot);
+            try (VectorStreamInput input = new VectorStreamInput(ownedRoot, namedWriteableRegistry)) {
                 input.setVersion(initialHeader.getVersion());
                 return handler.read(input);
             }
@@ -142,6 +143,21 @@ class FlightTransportResponse<T extends TransportResponse> implements StreamTran
 
     long getCurrentBatchSize() {
         return currentBatchSize;
+    }
+
+    /**
+     * Transfers the shared root's vectors into a response-owned root so the returned response
+     * is independent of FlightStream's lifecycle. The next call to {@code flightStream.next()}
+     * clears the shared root, and {@code stream.close()} releases its vectors — both would wipe
+     * the data out from under an async listener if we handed back a reference to the shared root.
+     */
+    private static VectorSchemaRoot transferToOwnedRoot(VectorSchemaRoot sharedRoot) {
+        VectorSchemaRoot ownedRoot = VectorSchemaRoot.create(
+            sharedRoot.getSchema(),
+            sharedRoot.getFieldVectors().getFirst().getAllocator()
+        );
+        FlightUtils.transferRoot(sharedRoot, ownedRoot);
+        return ownedRoot;
     }
 
     @Override
--- a/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightUtils.java
+++ b/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightUtils.java
@@ -8,7 +8,11 @@
 
 package org.opensearch.arrow.flight.transport;
 
+import org.apache.arrow.vector.FieldVector;
 import org.apache.arrow.vector.VectorSchemaRoot;
+import org.apache.arrow.vector.util.TransferPair;
+
+import java.util.List;
 
 class FlightUtils {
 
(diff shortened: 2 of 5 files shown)


plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightUtils.java
```diff
@@ -27,4 +31,19 @@ static long calculateVectorSchemaRootSize(VectorSchemaRoot root) {
...
+     * the buffers.
+     */
+    static void transferRoot(VectorSchemaRoot source, VectorSchemaRoot target) {
+        List<FieldVector> sources = source.getFieldVectors();
+        List<FieldVector> targets = target.getFieldVectors();
+        for (int i = 0; i < sources.size(); i++) {
+            TransferPair tp = sources.get(i).makeTransferPair(targets.get(i));
+            tp.transfer();
+        }
+        target.setRowCount(source.getRowCount());
```
alchemist51, line 46: Will we need to update the source row count since the data is now moved?

bowenlan-amzn: Thanks for catching this! addressed in d1ac3f8b04d4f7975859717e166e4b77eeb44bd9

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
```diff
@@ -121,9 +121,10 @@ public T nextResponse() {
             boolean hasNext = firstBatchConsumed ? flightStream.next() : (firstBatchConsumed = true);
             if (!hasNext) return null;
 
-            VectorSchemaRoot root = flightStream.getRoot();
```
rishabhmaurya, line 124: transfer of ownership isn't needed for non-arrow cases (streaming aggregation), so you may need a similar checks here

bowenlan-amzn: Pushed a change 4f985d7aedef8f9c740a515e8c204a290a6ee956
Decide the path based on whether is a single VarBinary field named 0.

bowenlan-amzn: This payload check approach is not good.

2 other approaches thinking:

**Handler marker interface** — `ArrowBatchResponseHandler extends
StreamTransportResponseHandler`; receive side dispatches via
`handler instanceof ArrowBatchResponseHandler<?>`.

Problem: the user's handler reaches FlightTransportResponse wrapped up
to 3 times:
- `MetricsTrackingResponseHandler` (FlightClientChannel:229, when
  callTracker != null)
- `ContextRestoreResponseHandler` (TransportService:1048, always on
  send)
- `TraceableTransportResponseHandler` (TransportService:977, when
  tracer is recording)

None expose their delegate, so the marker is invisible. Making it work
needs `default TransportResponseHandler<T> getDelegate()` on the core
interface plus overrides in each wrapper, then chain-walking in the
plugin. Mechanical cost is small (~4 lines) but it adds core-interface
surface motivated only by a plugin-specific dispatch need.

**Schema custom metadata** — sender stamps
`"opensearch.encoding" = "byte-serialized"` on the Arrow schema in
`VectorStreamOutput.ByteSerialized`; receiver reads
`schema.getCustomMetadata()` in `VectorStreamInput`. Immune to handler
wrapping, zero server-core changes, explicit wire-level contract.
Experimenting with this next.

bowenlan-amzn: I see pro and con for the 2 options above.
- handler brings change in core server
- schema metadata has some bwc concern

I think using header middleware may be the better way. But it's better to do those in a different PR.

bowenlan-amzn: https://github.com/opensearch-project/OpenSearch/pull/21452

rishabhmaurya: Handler marker interface is the way to go! Header based approach isn't needed here and making it overly complex. The client must know what response they are expecting and having marker interface to distinguish byte-serialized vs native-arrow is the right way to go about it.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
```diff
@@ -144,6 +145,21 @@ long getCurrentBatchSize() {
...
+    /**
+     * Transfers the shared root's vectors into a response-owned root so the returned response
+     * is independent of FlightStream's lifecycle. The next call to {@code flightStream.next()}
+     * clears the shared root, and {@code stream.close()} releases its vectors — both would wipe
+     * the data out from under an async listener if we handed back a reference to the shared root.
+     */
+    private static VectorSchemaRoot transferToOwnedRoot(VectorSchemaRoot sharedRoot) {
+        VectorSchemaRoot ownedRoot = VectorSchemaRoot.create(
+            sharedRoot.getSchema(),
+            sharedRoot.getFieldVectors().getFirst().getAllocator()
```
rishabhmaurya, line 157: this looks incorrect for use cases like DF, they may provide/use their dedicated allocator. I think the marker interface is the right way to go about it and get the allocator provided by response handler from their.

rishabhmaurya (approved):

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportResponse.java
rishabhmaurya: discussed offline, @bowenlan-amzn to address my concerns in the next PR.

PR #21438, opened 2026-04-30
Add BackendPlanAdapter SPI and Unify FilterOperator into ScalarFunction

expani:
Merges FilterOperator into ScalarFunction as a unified enum for all scalar functions (comparisons, full-text, math, string, conditional, datetime). Adds

### BackendPlanAdapter

A new planner component that applies per-function ScalarFunctionAdapter transforms between plan forking and fragment conversion. 

Anything that needs backend specific handling at the Calcite level will be triggered through this component.

  Changes:

  - Enum unification: FilterOperator deleted. Its entries (EQUALS, MATCHPHRASE, etc.) moved into `ScalarFunction` with a `Category` enum (COMPARISON, FULLTEXT, STRING,
  MATH, TYPE, CONDITIONAL, DATETIME). FilterCapability now references ScalarFunction via .function(). CapabilityRegistry uses ScalarKey for the filter index, removing the
  duplicate FilterKey.
  - SPI: New ScalarFunctionAdapter functional interface — backends register per-function adapters via BackendCapabilityProvider.scalarFunctionAdapters(). Takes (RexCall,
  List<FieldStorageInfo>), returns adapted RexNode.
  - BackendPlanAdapter: Walks resolved plan between forkAll and convertAll in DefaultPlanExecutor. Applies adapters to filter conditions and project expressions. Uses
  input's field storage for project (RexInputRef references input row type).
  - FieldStorageInfo moved from analytics.planner to analytics.spi (required by the SPI interface).
  - OpenSearchProjectRule fix: resolves ScalarFunction via fromSqlFunction fallback when fromSqlKind returns null (functions with SqlKind.OTHER).
  - 5 tests in BackendPlanAdapterTests: SIN CAST on integer field, SIN no-op on double field, SIN CAST in project expression, mixed adapted+non-adapted functions
  (SIN+ABS), no adapters registered.


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -8,56 +8,116 @@
...
+    public enum Category {
+        COMPARISON(false),
+        FULL_TEXT(true),
+        STRING(false),
+        MATH(false),
+        TYPE(false),
+        CONDITIONAL(false),
+        DATETIME(false);
+
+        private final boolean supportsParams;
```
vinaykpud: Does this mean say exg: MATH(false) , Match category supports parameters?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -8,56 +8,116 @@
...
-
-    // Conditional
-    CASE(SqlKind.CASE),
-    COALESCE(SqlKind.COALESCE),
-    NULLIF(SqlKind.NULLIF),
-
-    // Date/time
-    EXTRACT(SqlKind.EXTRACT);
 
+    // ── Comparisons ──────────────────────────────────────────────────
```
vinaykpud, line 28: So idea is any upstream functions/UDF we support , we will define them here rt?, its good. Because it will serve as registry for all functions Analytics core supports

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
```diff
@@ -0,0 +1,187 @@
...
+ * provided by the driving backend. Runs between plan forking and fragment conversion.
+ *
+ * <p>Each backend declares adapters keyed by {@link ScalarFunction} via
+ * {@link org.opensearch.analytics.spi.BackendCapabilityProvider#scalarFunctionAdapters()}.
+ * This component looks up the adapter for each scalar function RexCall in the plan
+ * and applies it if present.
+ *
+ * @opensearch.internal
+ */
+public class BackendPlanAdapter {
```
vinaykpud, line 44: Nice!

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
expani: Yes that's the idea.

This is the central place that should tie everything.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
expani: Actually this should have been per function level. Will remove it as it's unused

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
expani: 🚀

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
expani: Removed it. Will add with Filter Delegation PR when required

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunctionAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * Adapt the given expression for backend compatibility. Returns the adapted
+     * expression, or the original unchanged if no adaptation is needed.
+     *
+     * @param original     the backend-agnostic expression to adapt
+     * @param fieldStorage positional field storage info from the operator's child,
+     *                     indexed by {@link org.apache.calcite.rex.RexInputRef#getIndex()}
+     */
+    RexNode adapt(RexCall original, List<FieldStorageInfo> fieldStorage);
```
vinaykpud: Should ScalarFunctionAdapter.adapt() also receive a RexBuilder? Its needed to create to create new RexNodes in adapters like [this](https://github.com/opensearch-project/OpenSearch/pull/21436/changes#diff-c566b3b19844d57436148a10b1da10558a554665e7d76675091e48eb2dd9ce98R67)

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
```diff
@@ -0,0 +1,309 @@
...
+        MockDataFusionBackend dfWithAdapter = new MockDataFusionBackend() {
+            @Override
+            protected Map<ScalarFunction, ScalarFunctionAdapter> scalarFunctionAdapters() {
+                return Map.of(ScalarFunction.SIN, sinCastAdapter);
+            }
+        };
+
+        PlannerContext context = buildContext("parquet", 1, fields, List.of(dfWithAdapter));
+
+        RexNode sinCall = rexBuilder.makeCall(SIN_FUNCTION, rexBuilder.makeInputRef(typeFactory.createSqlType(operandType), 0));
```
vinaykpud, line 99: I think for test you might be getting `rexBuilder` from base class, So we need this in the adapter

expani: Yeah this was to create the valid RelNode but yes I am passing the whole RelOptCluster

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
```diff
@@ -0,0 +1,198 @@
...
+        // Recurse into operands first
+        List<RexNode> adaptedOperands = new ArrayList<>(call.getOperands().size());
+        boolean operandsChanged = false;
+        for (RexNode operand : call.getOperands()) {
+            RexNode adapted = adaptRex(operand, adapters, fieldStorage);
+            adaptedOperands.add(adapted);
+            if (adapted != operand) operandsChanged = true;
+        }
+
+        RexCall current = operandsChanged ? call.clone(call.getType(), adaptedOperands) : call;
```
mch2, line 184: We need to make an update here to prserve the wrapper subclass, without this RexCall.clone will strip annotations.

ex.:
```
        if (node instanceof AnnotatedProjectExpression annotated) {
            RexNode adaptedInner = adaptRex(annotated.getOriginal(), adapters, fieldStorage);
            return adaptedInner == annotated.getOriginal()
                ? annotated
                : new AnnotatedProjectExpression(annotated.getType(), adaptedInner, annotated.getViableBackends(), annotated.getAnnotationId());
        }
        if (node instanceof AnnotatedPredicate annotated) {
            RexNode adaptedInner = adaptRex(annotated.getOriginal(), adapters, fieldStorage);
            return adaptedInner == annotated.getOriginal()
                ? annotated
                : new AnnotatedPredicate(annotated.getType(), adaptedInner, annotated.getViableBackends(), annotated.getAnnotationId());
        }
```

expani: Great catch, will fix this.

expani: Added

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunctionAdapter.java
expani: Passed the RelOptCluster itself

expani:
@mch2 @vinaykpud Resolved all comments. Please take a final pass.

mch2 (approved):

vinaykpud (approved):

PR #21444, opened 2026-04-30
Dynamic Mapping support for Pluggable Data Formats

rayshrey:
### Description

**Why dynamic mapping support is needed for pluggable formats**
  
  OpenSearch supports dynamic mapping — when a document contains a field not in the index mapping, the field is automatically added. This works transparently with Lucene because Lucene is inherently schema-on-write: each document can have any fields regardless of what previous documents had. There's no upfront schema declaration.
  
  Parquet (and columnar formats in general) are fundamentally different. A Parquet file has a fixed schema declared at creation time. Every row in the file must conform to that schema. Once the native writer is initialized with a schema, it cannot accept rows with new columns. This means dynamic mapping — where new fields can appear at any time — requires explicit handling at
  the engine level.
  
  **How it works at the DataFormatAwareEngine level**
  
  The engine maintains a pool of writers. Each writer tracks:
  
  - Whether its schema can still evolve (isSchemaMutable)
  - What mapping version it was created with (mappingVersion)
  
  **Normal indexing flow (no new fields):**
  
  1. Document arrives → engine reads current mapping version
  2. Engine checks out a writer from the pool (predicate: writer is mutable OR its version matches)
  3. Document is written to the writer
  4. Writer is returned to the pool
  
  **When a new field arrives (dynamic mapping):**
  
  1. OpenSearch's bulk action detects the new field, sends mapping update to cluster manager
  2. Cluster state updates → mapping version increments → MapperService updated
  3. Document is retried with the new mapping in place
  4. Engine reads the new mapping version → checks out a writer:
    - If a mutable writer exists → it's selected (schema evolves dynamically inside)
    - If only immutable writers exist with old versions → they're evicted from the available queue, a new writer is created with the fresh schema
  5. Document is written successfully
  
  This design is format-agnostic. Lucene writers are always mutable (isSchemaMutable = true), so they always pass the predicate and handle any document regardless of schema changes. The version tracking and eviction logic is invisible to Lucene — it just works.
  
  **How Parquet handles it**
  
  On the Parquet side, the key insight is that the native writer initialization (which locks the schema) must be deferred until we're certain the schema is complete.
  
  Writer lifecycle:
  
  1. Writer is created with an initial Arrow schema from the current mapping → but the native Parquet writer is NOT initialized yet (isSchemaMutable = true)
  2. Documents arrive → if a document has a field not in the current Arrow schema, the field vector is dynamically added to the in-memory batch (VSR)
  3. When the batch is full (rotation) or flush is triggered:
    - The batch is frozen (no more writes)
    - The native writer is initialized with the frozen batch's schema (which includes all dynamically added fields)
    - isSchemaMutable becomes false
    - The batch is written to the native writer
  4. After initialization, any document with a newer mapping version will not match this writer → a new writer is created with the updated schema

### Related Issues
#21587 

### Check List

### DIFF
--- a/sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeDynamicMappingIT.java
+++ b/sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeDynamicMappingIT.java
@@ -8,6 +8,15 @@
 
 package org.opensearch.composite;
 
+import org.apache.lucene.index.DirectoryReader;
+import org.apache.lucene.index.DocValuesType;
+import org.apache.lucene.index.FieldInfo;
+import org.apache.lucene.index.LeafReader;
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.index.SortedNumericDocValues;
+import org.apache.lucene.index.SortedSetDocValues;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.NIOFSDirectory;
 import org.opensearch.action.admin.indices.create.CreateIndexResponse;
 import org.opensearch.action.admin.indices.flush.FlushResponse;
 import org.opensearch.action.admin.indices.mapping.get.GetMappingsResponse;
@@ -38,6 +47,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -85,14 +95,7 @@ public class CompositeDynamicMappingIT extends OpenSearchIntegTestCase {
      * 5. Verify all documents indexed, parquet files generated with correct segments
      */
     public void testDynamicMappingWithParquet() throws IOException {
-        Settings indexSettings = Settings.builder()
-            .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
-            .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
-            .put("index.pluggable.dataformat.enabled", true)
-            .put("index.pluggable.dataformat", "composite")
-            .put("index.composite.primary_data_format", "parquet")
-            .putList("index.composite.secondary_data_formats")
-            .build();
+        Settings indexSettings = parquetOnlySettings();
 
         // Create index with initial mapping
         CreateIndexResponse createResponse = client().admin()
@@ -124,22 +127,7 @@ public class CompositeDynamicMappingIT extends OpenSearchIntegTestCase {
         assertFalse("Mapping should NOT contain 'dynamic_long' yet", properties.containsKey("dynamic_long"));
 
         // Index documents with NEW dynamic fields
-        for (int i = 5; i < 10; i++) {
-            IndexResponse response = client().prepareIndex()
-                .setIndex(INDEX_NAME)
-                .setSource(
-                    "field_keyword",
-                    "value_" + i,
-                    "field_number",
-                    i,
-                    "dynamic_text",
-                    "dynamic_value_" + i,
-                    "dynamic_long",
-                    (long) i * 1000
-                )
-                .get();
-            assertEquals(RestStatus.CREATED, response.status());
-        }
+        indexDocsWithDynamicFields(INDEX_NAME, 5, 10);
 
         // Verify dynamic fields are now present in the mapping
         mappingsResponse = client().admin().indices().prepareGetMappings(INDEX_NAME).get();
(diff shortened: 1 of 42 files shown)


server/src/main/java/org/opensearch/index/engine/dataformat/IndexingExecutionEngine.java
```diff
@@ -33,9 +33,10 @@ public interface IndexingExecutionEngine<T extends DataFormat, P extends Documen
      * Creates a new writer for the given writer generation.
      *
      * @param writerGeneration the writer generation number
+     * @param mappingVersion the current mapping version
      * @return a new writer instance
      */
-    Writer<P> createWriter(long writerGeneration);
+    Writer<P> createWriter(long writerGeneration, long mappingVersion);
```
mgodwan: WriterConfig?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
```diff
@@ -221,6 +220,15 @@ public IndexStoreProvider getProvider() {
         return null;
     }
 
+    private Schema getOrBuildSchema(long mappingVersion) {
+        if (cachedSchemaVersion == mappingVersion && cachedSchema != null) {
+            return cachedSchema;
+        }
+        cachedSchema = ArrowSchemaBuilder.getSchema(mapperService);
```
mgodwan, line 266: We should use the verison based on schema returned rather than input to avoid a few flushes during concurrent mapping updates

mgodwan (commented):
1. Lets add concurrency tests.
2. Can we enable dynamic mapping indexing tests which already exists for DFAE?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -559,9 +560,13 @@ private Engine.IndexResult indexIntoEngine(Engine.Index index, IndexingStrategy
 
         // Convert ParsedDocument to DocumentInput and write via the execution engine's writer
         Writer currentWriter = null;
-        DefaultLockableHolder<Writer<?>> lockedWriter = writerPool.getAndLock();
+        long mappingVersion = currentMappingVersion();
+        DefaultLockableHolder<Writer<?>> lockedWriter = writerPool.getAndLock(
+            h -> h.get().isSchemaMutable() || h.get().mappingVersion() == mappingVersion
```
mgodwan: Should we check for `>=` to reuse writer where possible?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java
```diff
@@ -75,20 +77,27 @@ enum WriterState {
...
         this.primaryFormat = primaryDelegate.getDataFormat();
-        this.primaryWriter = (Writer<DocumentInput<?>>) primaryDelegate.createWriter(writerGeneration);
+        this.primaryWriter = (Writer<DocumentInput<?>>) primaryDelegate.createWriter(config);
+        this.mappingVersion = primaryWriter.mappingVersion();
 
         Map<DataFormat, Writer<DocumentInput<?>>> secondaries = new IdentityHashMap<>();
         for (IndexingExecutionEngine<?, ?> delegate : engine.getSecondaryDelegates()) {
-            secondaries.put(delegate.getDataFormat(), (Writer<DocumentInput<?>>) delegate.createWriter(writerGeneration));
+            Writer<DocumentInput<?>> secondary = (Writer<DocumentInput<?>>) delegate.createWriter(config);
+            assert secondary.isSchemaMutable() || secondary.mappingVersion() == this.mappingVersion : "Secondary writer mapping version ["
```
mgodwan: We should assert >=?

mgodwan: Also, an and check?

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeDynamicMappingIT.java
```diff
@@ -0,0 +1,386 @@
...
+     * 5. Verify all documents indexed, parquet files generated with correct segments
+     */
+    public void testDynamicMappingWithParquet() throws IOException {
+        Settings indexSettings = Settings.builder()
+            .put(IndexMetadata.SETTING_NUMBER_OF_SHARDS, 1)
+            .put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 0)
+            .put("index.pluggable.dataformat.enabled", true)
+            .put("index.pluggable.dataformat", "composite")
+            .put("index.composite.primary_data_format", "parquet")
+            .putList("index.composite.secondary_data_formats")
```
mgodwan: We should test with lucene as well

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/NativeParquetWriter.java
```diff
@@ -55,23 +79,31 @@ public NativeParquetWriter(String filePath, String indexName, long schemaAddress
      * @param arrayAddress  the native memory address of the Arrow array
      * @param schemaAddress the native memory address of the Arrow schema
      * @throws IOException if the write fails or the writer is flushed
+     * @throws IllegalStateException if the writer has not been initialized
      */
     public void write(long arrayAddress, long schemaAddress) throws IOException {
         if (writerFlushed.get()) {
             throw new IOException("Cannot write to flushed Parquet writer: " + filePath);
         }
+        if (initialized == false) {
```
mgodwan, line 88: We can instead use a memoized supplier? Can take up this in another PR but we should move away from initialized state mgmt.

mgodwan (approved):

PR #21447, opened 2026-04-30
Fix parent-bucket scoping in StreamStringTermsAggregator

harshavamsi:
When StreamStringTermsAggregator ran as a sub-aggregator under another terms aggregator, every parent bucket received the same inner bucket list. The collector wrote into a flat docCounts array keyed on segment ordinal, ignoring the owningBucketOrd passed into collect(), and selectTopBuckets walked that same global array once per parent ordinal.

Route collection through LongKeyedBucketOrds keyed on (owningBucketOrd, segmentOrdinal) so each parent bucket maintains its own doc-count state, and teach selectTopBuckets to enumerate buckets per owner via ordsEnum(owningBucketOrd). Thread CardinalityUpperBound into createStreamStringTermsAggregator so the factory can pick between FromSingle and FromMany implementations.

Add a deterministic reproducer in StreamStringTermsAggregatorTests that fails on the pre-fix code by asserting two parent buckets produce distinct inner results.

### Description

### Related Issues

### Check List


rishabhmaurya:
@harshavamsi I think we should add an IT to cover the sub aggregation case here.

harshavamsi:
@rishabhmaurya Added an IT for this case.

harshavamsi:
{"run-benchmark-test":"id_17"}

rishabhmaurya:
{"run-benchmark-test":"id_17"}

rishabhmaurya:
@harshavamsi do you mind taking a look at the failure

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/StreamStringTermsAggregator.java
```diff
@@ -261,92 +252,122 @@ private static class SelectionResult<B> {
...
+        private SelectionResult<B> selectTopBuckets(long owningBucketOrd, int segmentSize, BucketCountThresholds thresholds)
+            throws IOException {
+            long bucketsForOwner = bucketOrds.bucketsInOrd(owningBucketOrd);
+            if (bucketsForOwner == 0) {
+                return new SelectionResult<>(new ArrayList<>(), 0L);
+            }
+
+            // Pair up each candidate as (bucketOrd, segmentOrdinal). We keep them in two
+            // parallel arrays sized bucketsForOwner.
+            long[] candidateBucketOrds = new long[Math.toIntExact(bucketsForOwner)];
```
rishabhmaurya: we should not be creating a regular java long array of size with upper bound upto max ordinals, which seems to be the case here. Please make use of bigarrays or it will result into hard OOMs rather than CBE.

rishabhmaurya: It is fine to create java array of size of topN requested which is `effectiveSegmentSize` or `segmentSize` which is usually much smaller

rishabhmaurya: You can see how in previous logic we made use of reusableIndices bigArray until we knew the effective segment size and then we lazily used java int array of that size to perform quick select.

rishabhmaurya:
@harshavamsi added some comments, I highly suggest running some stress tests leading to CBEs for this part and validating its working as expected. 
So with this change, streaming string terms aggregation will be supporting the sub aggregations as well. How about numeric terms aggs? do we allow it at time of validation numeric term aggs?

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/StreamStringTermsAggregator.java
```diff
@@ -261,92 +252,122 @@ private static class SelectionResult<B> {
...
-            reusableIndices.fill(0, valueCount, 0);
-            for (int i = 0; i < segmentSize; i++) {
-                reusableIndices.set(selected[i], 1);
+            // Match the historical emit order (ascending by segment ordinal, i.e.
+            // alphabetical) so downstream reduce ordering assumptions hold.
+            Integer[] indices = new Integer[effectiveSegmentSize];
+            for (int i = 0; i < effectiveSegmentSize; i++) {
+                indices[i] = i;
             }
+            Arrays.sort(indices, (a, b) -> Long.compare(candidateSegmentOrds[a], candidateSegmentOrds[b]));
```
rishabhmaurya: is it possible in this new logic to retrieve in the ordinal order itself and avoid sorting here

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/StreamStringTermsAggregator.java
harshavamsi: thanks, this is indeed a valid concern. Using `BigArrays` now.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/StreamStringTermsAggregator.java
harshavamsi: We get each ordinal in insertion order and not segment ordinal order, which means we cannot avoid sorting since downstream reduce expects keys to be sorted

rishabhmaurya: but we can retrieve in ordinal order right? by running a loop to max ordinal and then checking which ones made to topN instead of sorting?

harshavamsi: So sorting here would be O(NlogN) where N is the topN value which is usually small. But if we want to loop through the entire set of unique values, those could be very high per batch. So it would be O(k) where K could be a much higher value. wdyt?

rishabhmaurya: since we already do O(K) throughout the logic, another O(K) still keeps complexity linear in my opinion and would avoid log factor. I'm not too adamant on it, so if you think sort here is simplifying the code, go ahead

harshavamsi:
{"run-benchmark-test":"id_17"}

server/src/test/java/org/opensearch/action/search/StreamSearchIntegrationTests.java
```diff
@@ -339,6 +351,82 @@ public void testStreamingAggregationTermsOnly() {
...
+        assertTrue(client().admin().indices().create(createIndexRequest).actionGet().isAcknowledged());
+        client().admin()
+            .cluster()
+            .prepareHealth(NESTED_TERMS_INDEX)
+            .setWaitForGreenStatus()
+            .setTimeout(TimeValue.timeValueSeconds(30))
+            .get();
+
+        BulkRequest bulkRequest = new BulkRequest();
+        for (int i = 0; i < 3; i++) {
```
rishabhmaurya, line 376: I think we can add more docs here maybe some field with cardinality of around 1000 and maybe higher topN, that should still be fast and more trustworthy test. Also, if we can validate sum other doc count logic here in IT.

harshavamsi:
> The benchmark job https://build.ci.opensearch.org/job/benchmark-pull-request/8633/ failed. Please see logs to debug.

The reason this is failing is because the benchmark repo is using @rishabh6788's branch instead of mainline. Will need to move it to use mainline

harshavamsi:
{"run-benchmark-test":"id_17"}

harshavamsi:
{"run-benchmark-test":"id_17"}

rishabhmaurya (approved):

PR #21448, opened 2026-05-01
Initial implementation of deployment APIs

msfroh:
### Description
This adds "drain" and "finish" APIs to start and finish deploying to nodes based on some node attribute(s).

When nodes are drained, they are not eligible to hold primary shards and they will not receive search traffic. That allows the nodes to be shut down with minimal impact to ongoing requests. When nodes return (or are replaced by new nodes with the same attribute(s)), they still do not receive search traffic until the deployment is marked as complete (which may wait for recovery to complete).

Future commits will add a "node dial-up" state (to add search traffic incrementally), as well as a "shadow" state (to send search traffic in parallel, to warm-up nodes and validate success).

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

The TASK.md file that was used to generate the code is available at https://gist.github.com/msfroh/c9d319fbd252975013f5aaca0dfc93dc

### Related Issues
Partial implementation of https://github.com/opensearch-project/OpenSearch/issues/21180

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/cluster/ClusterModule.java
+++ b/server/src/main/java/org/opensearch/cluster/ClusterModule.java
@@ -36,6 +36,8 @@ import org.opensearch.cluster.action.index.MappingUpdatedAction;
 import org.opensearch.cluster.action.index.NodeMappingRefreshAction;
 import org.opensearch.cluster.action.shard.ShardStateAction;
 import org.opensearch.cluster.decommission.DecommissionAttributeMetadata;
+import org.opensearch.cluster.deployment.DeploymentAllocationDecider;
+import org.opensearch.cluster.deployment.DeploymentMetadata;
 import org.opensearch.cluster.metadata.ComponentTemplateMetadata;
 import org.opensearch.cluster.metadata.ComposableIndexTemplateMetadata;
 import org.opensearch.cluster.metadata.DataStreamMetadata;
@@ -196,6 +198,7 @@ public class ClusterModule extends AbstractModule {
         registerMetadataCustom(entries, RepositoriesMetadata.TYPE, RepositoriesMetadata::new, RepositoriesMetadata::readDiffFrom);
         registerMetadataCustom(entries, IngestMetadata.TYPE, IngestMetadata::new, IngestMetadata::readDiffFrom);
         registerMetadataCustom(entries, SearchPipelineMetadata.TYPE, SearchPipelineMetadata::new, SearchPipelineMetadata::readDiffFrom);
+        registerMetadataCustom(entries, DeploymentMetadata.TYPE, DeploymentMetadata::new, DeploymentMetadata::readDiffFrom);
         registerMetadataCustom(entries, ScriptMetadata.TYPE, ScriptMetadata::new, ScriptMetadata::readDiffFrom);
         registerMetadataCustom(entries, IndexGraveyard.TYPE, IndexGraveyard::new, IndexGraveyard::readDiffFrom);
         registerMetadataCustom(
@@ -405,6 +408,7 @@ public class ClusterModule extends AbstractModule {
         addAllocationDecider(deciders, new AwarenessAllocationDecider(settings, clusterSettings));
         addAllocationDecider(deciders, new NodeLoadAwareAllocationDecider(settings, clusterSettings));
         addAllocationDecider(deciders, new TargetPoolAllocationDecider());
+        addAllocationDecider(deciders, new DeploymentAllocationDecider());
         addAllocationDecider(deciders, new RemoteStoreMigrationAllocationDecider(settings, clusterSettings));
 
         clusterPlugins.stream()
--- a/server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
+++ b/server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
@@ -1,0 +1,198 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.cluster.deployment;
+
+import org.opensearch.action.support.clustermanager.AcknowledgedResponse;
+import org.opensearch.cluster.AckedClusterStateUpdateTask;
+import org.opensearch.cluster.ClusterState;
+import org.opensearch.cluster.ack.AckedRequest;
+import org.opensearch.cluster.metadata.Metadata;
+import org.opensearch.cluster.routing.RerouteService;
+import org.opensearch.cluster.service.ClusterManagerTask;
+import org.opensearch.cluster.service.ClusterManagerTaskThrottler;
+import org.opensearch.cluster.service.ClusterService;
+import org.opensearch.common.Priority;
+import org.opensearch.core.action.ActionListener;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Service that manages deployment lifecycle by writing DeploymentMetadata to cluster state.
+ * Runs on the elected cluster manager node only.
+ *
+ * @opensearch.internal
+ */
+public class DeploymentManagerService {
+
+    private final ClusterService clusterService;
+    private final RerouteService rerouteService;
+    private final ClusterManagerTaskThrottler.ThrottlingKey startDeploymentTaskKey;
+    private final ClusterManagerTaskThrottler.ThrottlingKey finishDeploymentTaskKey;
+
+    public DeploymentManagerService(ClusterService clusterService, RerouteService rerouteService) {
+        this.clusterService = clusterService;
+        this.rerouteService = rerouteService;
+        this.startDeploymentTaskKey = clusterService.registerClusterManagerTask(ClusterManagerTask.START_DEPLOYMENT, true);
+        this.finishDeploymentTaskKey = clusterService.registerClusterManagerTask(ClusterManagerTask.FINISH_DEPLOYMENT, true);
+    }
+
+    public void startDeployment(
+        String deploymentId,
+        Map<String, String> nodeAttributes,
+        AckedRequest request,
+        ActionListener<AcknowledgedResponse> listener
+    ) {
+        validateDeploymentId(deploymentId);
+        validateAttributes(nodeAttributes);
+
+        clusterService.submitStateUpdateTask("start-deployment-" + deploymentId, new AckedClusterStateUpdateTask<>(request, listener) {
+            @Override
+            public ClusterState execute(ClusterState currentState) {
+                return innerStartDeployment(deploymentId, nodeAttributes, currentState);
+            }
+
+            @Override
+            public ClusterManagerTaskThrottler.ThrottlingKey getClusterManagerThrottlingKey() {
+                return startDeploymentTaskKey;
+            }
+
+            @Override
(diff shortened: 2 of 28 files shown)


msfroh:
> Other disruptive cluster-admin APIs in OpenSearch typically carry explicit privilege checks or are gated by security plugin role definitions; the absence here is an anomaly worth reviewing.

I've tried looking at a couple of other potentially disruptive cluster manager APIs (`TransportClusterRerouteAction`, `TransportAddWeightedRoutingAction`), but I'm not seeing these explicit privilege checks. If anyone (human or bot) can point me in the right direction for an example, I'm happy to add privilege checks.

msfroh:
I agree with all 4 suggestions in https://github.com/opensearch-project/OpenSearch/pull/21448#issuecomment-4374339925:
* Prevent empty shard routing table when all replicas are drained
* Route validation exceptions through ActionListener on failure
* Use path parameter instead of parsing raw URL path
* Validate attribute key consistency against all existing deployments

I'll submit a followup commit with those changes.

server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
```diff
@@ -0,0 +1,219 @@
...
+            }
+
+            @Override
+            public void onAllNodesAcked(Exception e) {
+                super.onAllNodesAcked(e);
+            }
+        });
+    }
+
+    public static ClusterState innerStartDeployment(String deploymentId, Map<String, String> nodeAttributes, ClusterState currentState) {
```
zhaih: Rename it to `addNewDeployment`? (A public static `innerXX` method doesn't make a lot of sense)

server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
```diff
@@ -0,0 +1,219 @@
...
+        }
+
+        return ClusterState.builder(currentState).metadata(metadataBuilder.build()).build();
+    }
+
+    private static void validateDeploymentId(String deploymentId) {
+        if (deploymentId == null || deploymentId.isEmpty()) {
+            throw new IllegalArgumentException("deployment ID must not be empty");
+        }
+        if (deploymentId.startsWith("_")) {
```
zhaih, line 158: Curious is there any specific reason for this?

server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
```diff
@@ -0,0 +1,219 @@
...
+                // We allow idempotent application of start deployment.
+                Deployment existingDeploymentWithSameId = entry.getValue();
+                if (existingDeploymentWithSameId.getState() != DeploymentState.DRAIN) {
+                    throw new IllegalArgumentException(
+                        "deployment [" + entry.getKey() + "] is already in state " + existingDeploymentWithSameId.getState()
+                    );
+                }
+                if (existingDeploymentWithSameId.getNodeAttributes().equals(newAttributes) == false) {
+                    throw new IllegalArgumentException(
+                        "deployment [" + entry.getKey() + "] already targets attributes " + existingDeploymentWithSameId.getNodeAttributes()
```
zhaih: Maybe some thing like `deployment [id] is not allowed to update targeting attributes to [new_attributes], already targeted [old_attributes]`?

server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
```diff
@@ -0,0 +1,219 @@
...
+        deployments.put(deploymentId, new Deployment(DeploymentState.DRAIN, nodeAttributes));
+
+        return ClusterState.builder(currentState)
+            .metadata(
+                Metadata.builder(currentState.getMetadata()).putCustom(DeploymentMetadata.TYPE, new DeploymentMetadata(deployments)).build()
+            )
+            .build();
+    }
+
+    public static ClusterState innerFinishDeployment(String deploymentId, ClusterState currentState) {
```
zhaih: rename to `removeDeployment`?

server/src/main/java/org/opensearch/cluster/deployment/DeploymentStateService.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+        return null;
+    }
+
+    private static boolean nodeMatchesAttributes(DiscoveryNode node, Map<String, String> targetAttributes) {
+        for (Map.Entry<String, String> entry : targetAttributes.entrySet()) {
+            String nodeValue = node.getAttributes().get(entry.getKey());
+            if (!entry.getValue().equals(nodeValue)) {
+                return false;
+            }
```
zhaih, line 91: Probably it worths adding a javadoc indicating the attributes set are worked as a conjunction rather than disjunction?

server/src/main/java/org/opensearch/cluster/routing/OperationRouting.java
```diff
@@ -260,11 +262,16 @@ public GroupShardsIterator<ShardIterator> searchShards(
         @Nullable SliceBuilder slice
     ) {
         final Set<IndexShardRoutingTable> shards = computeTargetedShards(clusterState, concreteIndices, routing);
+        final Map<String, DeploymentState> drainedNodes = DeploymentStateService.getNodeDeploymentStates(clusterState);
```
zhaih, line 265: Wonder whether we should pre-populate this when `clusterState.nodes()` or deployments are changed, otherwise calculating this per-search might be a slightly heavy operation? (Probably fine if we assume at max 1000 nodes x 4 or 5 in-flight deployment request?)

server/src/main/java/org/opensearch/cluster/routing/OperationRouting.java
```diff
@@ -325,6 +332,21 @@ public static ShardIterator getShards(ClusterState clusterState, ShardId shardId
...
+        IndexShardRoutingTable.Builder builder = new IndexShardRoutingTable.Builder(shard.shardId);
+        for (ShardRouting shardRouting : shard) {
+            if (!drainedNodes.containsKey(shardRouting.currentNodeId())) {
+                builder.addShard(shardRouting);
+            }
+        }
+        IndexShardRoutingTable filtered = builder.build();
+        if (filtered.size() == 0) {
+            // Fall back to original if all shards were filtered out
+            return shard;
```
zhaih, line 356: Is this the desired behavior? It should never happens but if this is really the case fallback to original means we are anyway routing to whatever is claimed 'drained', should we instead just error out?

server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
msfroh: It's a convention on OpenSearch that things starting with `_` are reserved for internal operations. While I currently have nothing in mind for internally-generated deployments, I figured adding this restriction would make sure we have the option in future. (We can always remove the restriction later, but it's hard to add a restriction after people start using the feature.)

server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
msfroh: Yeah... it's only public to let it be used from tests in other packages.

I was able to move some things around and make it package-private again.

server/src/main/java/org/opensearch/cluster/deployment/DeploymentManagerService.java
msfroh: Made package-private

server/src/main/java/org/opensearch/cluster/routing/OperationRouting.java
msfroh: Maybe. I would like to hold off to see if/when such precomputation is needed. 

In practice, I expect we would run 1 deployment at a time, with 2 as a rare occurrence (and most of the time, it's zero).

server/src/main/java/org/opensearch/cluster/routing/OperationRouting.java
msfroh: Yeah... I'm thinking of adding another safeguard to avoid starting a single deployment that targets all nodes (or maybe even more than 50% of nodes?), which would further make this less likely.

That said, if someone does kick off multiple deployments such that all nodes get drained, I feel like failing open is safer.

msfroh:
Hmm... once again, I like the current set of suggestions from the AI bot. I'll address those shortly.

server/src/main/java/org/opensearch/cluster/deployment/DeploymentMetadata.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Override
+    public String getWriteableName() {
+        return TYPE;
+    }
+
+    @Override
+    public Version getMinimalSupportedVersion() {
+        return Version.V_3_7_0;
```
varunbharadwaj: We have to now update this to 3.9.0?

server/src/main/java/org/opensearch/cluster/ClusterModule.java
```diff
@@ -196,6 +198,7 @@ public static List<Entry> getNamedWriteables() {
         registerMetadataCustom(entries, RepositoriesMetadata.TYPE, RepositoriesMetadata::new, RepositoriesMetadata::readDiffFrom);
         registerMetadataCustom(entries, IngestMetadata.TYPE, IngestMetadata::new, IngestMetadata::readDiffFrom);
         registerMetadataCustom(entries, SearchPipelineMetadata.TYPE, SearchPipelineMetadata::new, SearchPipelineMetadata::readDiffFrom);
+        registerMetadataCustom(entries, DeploymentMetadata.TYPE, DeploymentMetadata::new, DeploymentMetadata::readDiffFrom);
```
varunbharadwaj, line 201: On a full cluster restart, the gateway reads persisted cluster state as XContent. Since DeploymentMetadata isn't in the XContent registry, the parser doesn't recognize it and silently skips it. Any active deployment/drain state disappears.

We can explore if the following is needed here.

```
entries.add(
    new NamedXContentRegistry.Entry(
        Metadata.Custom.class,
        new ParseField(DeploymentMetadata.TYPE),
        DeploymentMetadata::fromXContent
    )
);
```

msfroh: Hmm... that's a point that I hadn't considered.

I *think* it's okay if a full cluster restart clears any active deployments, since you're trying to recover from something catastrophic anyway. You could always bring up the cluster managers, mark nodes for deployment using this API, then bring up data nodes in a drained state. 

When you create a deployment, you don't need any current cluster nodes to match the attribute filter. I did that because a blue-green deployment would want to launch the new nodes in a drained state. Similarly, if you're doing a rolling deployment on Kubernetes by replacing pods, you'd want to mark the new pod as drained before bringing it up.

If you have a counterargument, I'm happy to add the deployment state to the registry. Again, I didn't consider it one way or the other, so I don't have a strong opinion.

varunbharadwaj: Thinking more, I think it makes sense and clearing it in this scenario is okay.

zhaih (approved):
LGTM

msfroh:
Thank you for the feedback, @zhaih and @varunbharadwaj! 

I just updated the minimal supported version (and made a typo in the commit message 🤦 ), which I think is the one outstanding thing. Hopefully the Gradle check will pass.

varunbharadwaj (approved):

PR #21450, opened 2026-05-01
Planner changes to convert Predicates to Lucene QueryBuilder and insert Placeholder in Substrait

expani:
### Summary

   Enables the analytics plugin's coordinator-side planner to delegate filter predicates to a different backend than the one executing the operator. 

When a predicate like `MATCH(message, 'hello world')` can only be evaluated by Lucene but the filter operator runs on DataFusion, the framework now detects this mismatch during fragment conversion, serializes the predicate via the accepting backend, and replaces it with a delegated_predicate(annotationId) placeholder function in the Substrait plan sent to the data node.

This is the foundation for filter delegation — the mechanism that allows full-text search predicates (MATCH, MATCH_PHRASE, FUZZY, etc.) to coexist with columnar scan predicates in a single filter, each evaluated by the backend best suited for it.

### What this enables

   - Intra-operator delegation: predicates within a single filter operator can target different backends. The operator's backend is the delegator; predicates whose viable backend differs are delegated.
   - Per-function serialization: each backend registers serializers keyed by ScalarFunction. The Lucene backend's MATCH serializer produces MatchQueryBuilder bytes that the data node can deserialize and execute against the Lucene index.
   - Substrait-compatible placeholders: delegated_predicate(annotationId) is registered as a custom Substrait extension function, so Isthmus converts it cleanly. The boolean tree structure (AND/OR/NOT mixing native and delegated predicates) is preserved through to Substrait.
   - Delegated query transport: StagePlan carries a `Map<Integer, byte[]>` of serialized delegated queries alongside the Substrait plan bytes, keyed by annotation ID for data-node lookup. This makes the conversion easier on the 

###  Design decisions

   - Operators own their annotation removal: `stripAnnotations` accepts a `Function<OperatorAnnotation, RexNode>` resolver. Each operator walks its own internals (condition tree, project list, aggregate calls) and calls the resolver per annotation. The resolver decides native vs delegated — operators have zero delegation awareness.
   - Serializer anchoring: DelegatedPredicateSerializer is keyed by ScalarFunction on BackendCapabilityProvider, same pattern as ScalarFunctionAdapter. Adding a new delegatable function means registering one serializer — no framework changes.
   - Placeholder per annotation type: OperatorAnnotation.makePlaceholder() lets each annotation type produce the right placeholder shape (BOOLEAN for predicates, TBD for project/aggregate). Currently only AnnotatedPredicate implements it; others throw until needed.

###   What was tested

   - 6 delegation tagging tests covering single/mixed/AND/OR/NOT/interleaved boolean structures — verify placeholder replacement, delegated query counts, serialized function names
   - End-to-end test with real Lucene backend: MATCH predicate through marking → forking → fragment conversion → Substrait, with deserialized MatchQueryBuilder verified for field name and query text
   - Substrait structure tests: delegated_predicate function reference resolved from extension declarations, annotation IDs verified against delegated query map, AND/OR/NOT function names verified in proto

###   Not in this PR ( Will be Fast followups ) 

   - Transport serialization of delegatedQueries in FragmentExecutionRequest
   - Data node: deserialization, delegation bridge, Lucene query execution
   - Additional serializers (MATCH_PHRASE, FUZZY, WILDCARD, REGEXP, standard ops for performance delegation)
   - Same-backend combining (adjacent Lucene predicates → single BooleanQuery)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/FragmentExecutionRequest.java
```diff
@@ -98,7 +98,7 @@ public ActionRequestValidationException validate() {
 
     /**
      * A single plan alternative: a backend ID paired with its serialized fragment bytes.
-     * Produced by {@code FragmentConversionDriver.convert()} using the backend's
+     * Produced by {@code FragmentConversionDriver.convertAll()} using the backend's
      * {@code FragmentConvertor}.
      */
```
mch2, line 103: Are we adding the map here to send to data nodes?

mch2 (approved):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/FragmentExecutionRequest.java
expani: Yes in a follow up PR for data node execution.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
```diff
@@ -65,16 +80,83 @@ private static void convertStage(Stage stage, CapabilityRegistry registry) {
...
+                    annotation.getAnnotationId(),
+                    function,
+                    operatorBackend,
+                    annotationBackend,
+                    serialized.length
+                );
+                if (result == null) {
+                    result = new HashMap<>();
+                }
+                result.put(annotation.getAnnotationId(), serialized);
```
vinaykpud, line 146: nit: currently i think it cannot happen, but it will be safer here to assert for duplicate annotationId.

vinaykpud (approved):

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DelegatedPredicateFunction.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Placeholder function for delegated predicates in the Calcite plan.
+ *
+ * <p>When a predicate is delegated to another backend (e.g., a MATCH_PHRASE predicate
+ * delegated from DataFusion to Lucene), the original expression is serialized and sent
+ * as opaque bytes. In the Calcite plan that goes to Isthmus/Substrait, the original
+ * expression is replaced with {@code delegated_predicate(annotationId)} — a function
+ * that always evaluates to TRUE and carries the annotation ID so the driving backend
+ * can look up the delegated query at execution time.
```
vinaykpud, line 30: the javadoc is slightly misleading, nothing in the code actually makes it return `TRUE`. Currently it's a Calcite SqlFunction with `ReturnTypes.BOOLEAN` as its return type.

So can you help me to understand here how its envisioned to execute at data node?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
```diff
@@ -65,16 +80,83 @@ private static void convertStage(Stage stage, CapabilityRegistry registry) {
...
+        IntraOperatorDelegationBytes(CapabilityRegistry registry) {
+            this.registry = registry;
+        }
+
+        /**
+         * Creates an annotation resolver scoped to a specific operator. Compares each
+         * annotation's viable backend against the operator's backend: native annotations
+         * are unwrapped, delegated ones are serialized and replaced with a placeholder.
+         */
+        Function<OperatorAnnotation, RexNode> resolverFor(OpenSearchRelNode operator, RexBuilder rexBuilder) {
```
vinaykpud, line 107: All existing tests cover the happy path, can we cover these error path as well?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DelegatedPredicateFunction.java
expani: This is just adding the placeholder function to match schema. 

In DataNode PR, we register a UDF that is hardcoded to return true and add it. This acts as a marker for the same.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
expani: Added

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
expani: Yes Marking guarantees monotonically increasing IDs

PR #21457, opened 2026-05-02
[analytics-backend-datafusion] Implement partial/final aggregation mode for distributed execution

sandeshkr419:
Introduces end-to-end distributed aggregate execution for the analytics plugin. Shards compute partial state, the coordinator merges to final results, and the whole path works across a multi-node multi-shard cluster. 39 of 43 ClickBench PPL queries now pass in that topology.

   ### Architecture

   The control split is layered to avoid per-function branches. Java owns the logical plan; Rust compiles each half (PARTIAL or FINAL) of whatever substrait it receives:

   1. **HEP marking**: `OpenSearchAggregateRule`, `OpenSearchProjectRule` (with baseline scalar-op carve-out), and `OpenSearchAggregateReduceRule` (reduces AVG → SUM+COUNT+Project in a dedicated phase, before the Volcano split, using Calcite's testedmachinery).
   2. **Volcano CBO split**: `OpenSearchAggregateSplitRule` turns SINGLE → FINAL(Exchange(PARTIAL(…))) structurally — no per-function logic.
   3. **Decomposition resolver** (`AggregateDecompositionResolver`): single pass handling the three single-field shapes — pass-through (SUM/MIN/MAX), function-swap (COUNT → SUM at FINAL), engine-native merge (APPROX_COUNT_DISTINCT sketch). Rebuilds the exchange column row type with `ArrowCalciteTypes` (single-type bidirectional) and rewires parent projections.
   4. **DAG cut** (`DAGBuilder`): structural only — cuts at `OpenSearchExchangeReducer`.
   5. **Backend conversion** (`DataFusionFragmentConvertor`): mechanical Calcite → Substrait emission; APPROX_COUNT_DISTINCT routes to DataFusion's native HLL via the extension YAML + a subclassed `AggregateFunctionConverter`.
   6. **Instruction assembly** (`FragmentConversionDriver`).
   7. **Execution handlers** (`PartialAggregateInstructionHandler`, `FinalAggregateInstructionHandler`) → named FFI (`preparePartialPlan`, `prepareFinalPlan`, `executeLocalPreparedPlan`). No aggregate-mode enum on the wire.
   8. **Streaming** (`LocalStageScheduler`): one-line `arrowSchemaFromRowType(child.getFragment().getRowType())`.
   9. **Physical** (Rust `agg_mode.rs`): internal `Mode {Default, Partial, Final}`; `force_aggregate_mode` strips one half; `CombinePartialFinalAggregate` disabled on every SessionContext so `Final(Partial(…))` survives to the strip pass.

   ### Single source of truth for per-function decomposition

   `AggregateFunction` enum (SPI) carries `intermediateFields` — the only authority for what the PARTIAL emits and how the FINAL reduces it. `fromSqlAggFunction` / `toSqlAggFunction` bridge between Calcite operators and enum constants. `COUNT` and	`APPROX_COUNT_DISTINCT` declare their shapes there; SUM/MIN/MAX/AVG use null intermediates (pass-through post-reduction).

   ### Multi-shard unblocks

   - **Empty-batch schema preservation**: shards that match zero rows still emit one zero-row VSR carrying the declared schema, so the FINAL plan's first data frame carries types downstream even when WHERE filters out everything.
   - **Arrow IPC wire codec** replaces row-oriented `Object[]` in `FragmentExecutionResponse`. Arrow's own serializer/deserializer round-trips every type natively (temporal, view, dictionary) without hand-rolled per-type dispatch. Removes the `writeGenericValue(LocalDateTime)` crash class that blocked all TIMESTAMP queries.
   - **Java↔Rust type-coercion bridge** at `DatafusionReduceSink.feedToSender` reconciles DataFusion's physical emit types (e.g. `Utf8View` for string group keys) with Calcite's declared exchange schema (`Utf8`) before pushing into the FINAL plan's StreamingTable partition.

   ### TODOs to revisit (tracked independently)

The goal is to unblock distributed query execution framework and avoid spilling more changes in this already huge PR.

   1. **Type coercion at the right layer**: the `Utf8View → Utf8` shim in `coerceToDeclaredSchema` is a runtime bridge; the principled fix is a Rust `force_output_schema` cast at the PARTIAL root, a Substrait view-type extension, or verified Arrow Java ↔Rust FFI symmetry for view types.
   2. **Aggregate decomposition catalog → per-backend SPI**: `AggregateFunction.intermediateFields` hard-codes DataFusion's state shapes into the shared enum; the catalog should move to `BackendCapabilityProvider.aggregateDecompositions()` (engine keeps the structural resolver, backends declare their own metadata).
   3. **MultisearchCommandIT.testMultisearchCountEvalConditionalCount and AppendPipeCommandIT.testMultisearchCountEvalConditionalCount** - commented out for now - need to fix and revert this test back.
   
   
### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/local_executor.rs
```diff
@@ -162,6 +172,143 @@ impl LocalSession {
...
+impl PhysicalOptimizerRule for StripFinalAggregateRule {
+    fn optimize(
+        &self,
+        plan: Arc<dyn datafusion::physical_plan::ExecutionPlan>,
+        _config: &datafusion_common::config::ConfigOptions,
+    ) -> datafusion_common::Result<Arc<dyn datafusion::physical_plan::ExecutionPlan>> {
+        use datafusion::physical_plan::aggregates::{AggregateExec, AggregateMode};
+
+        if let Some(final_agg) = plan.as_any().downcast_ref::<AggregateExec>() {
+            if matches!(final_agg.mode(), AggregateMode::Final | AggregateMode::FinalPartitioned)
```
expani: I think we can explore iterating over the plan deterministically via an API from a DataNodeLocalPlanner using FragmentConversionAPI.

This way the backend can deterministically know whether the Substrait it's receiving is for partial or final

This is a good opportunity to introduce it.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/AvgDecomposition.java
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * <p>PARTIAL phase emits two columns: {@code [SUM(x): DOUBLE, COUNT(x): BIGINT]}.
+ * FINAL phase computes {@code SUM_ref / COUNT_ref} as a scalar expression.
+ */
+public class AvgDecomposition implements AggregateDecomposition {
+
+    public static final AvgDecomposition INSTANCE = new AvgDecomposition();
+
+    @Override
+    public List<AggregateCall> partialCalls(AggregateCall originalCall, org.apache.calcite.rel.RelNode input) {
```
expani: Decomposition is meant to represent the AggregateFunction of a backend that has a custom partial state that is not centralized.

AvgDecomposition should not exist unless a backend is doing something special for average aggregates.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/ArrowSchemaFromCalcite.java
```diff
@@ -40,12 +40,16 @@ private ArrowSchemaFromCalcite() {}
     public static Schema arrowSchemaFromRowType(RelDataType rowType) {
         List<Field> fields = new ArrayList<>();
         for (RelDataTypeField f : rowType.getFieldList()) {
-            ArrowType arrowType = toArrowType(f.getType().getSqlTypeName());
-            fields.add(new Field(f.getName(), new FieldType(true, arrowType, null), null));
+            fields.add(fieldFromCalcite(f));
```
expani: I think this entire class should be tied down to `FieldType` enum. 

It's a change outside the scope of this PR so maybe we can start with TODO

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/LocalStageScheduler.java
```diff
@@ -73,6 +77,50 @@ private static Schema deriveInputSchema(Stage stage) {
             + stage.getStageId()
             + " expected exactly one child stage, got "
             + children.size();
-        return ArrowSchemaFromCalcite.arrowSchemaFromRowType(children.getFirst().getFragment().getRowType());
+        Stage child = children.getFirst();
+        RelNode childFragment = child.getPlanAlternatives().isEmpty()
+            ? child.getFragment()
+            : child.getPlanAlternatives().getFirst().resolvedFragment();
+
+        // Find the aggregate in the child fragment to check for functions with
```
expani: This should be happening in the Planner as a Marking ( Validation via Capability API )

`AggregateFunction` is a part of the Capability that Datafusion publishes.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchAggregate.java
```diff
@@ -179,25 +182,31 @@ public RelNode copyResolved(String backend, List<RelNode> children, List<Operato
 
     @Override
     public RelNode stripAnnotations(List<RelNode> strippedChildren) {
+        RelNode strippedInput = strippedChildren.getFirst();
+        // Build a temporary no-op aggregate (empty aggCalls) just to serve as the
```
expani: This should be a part of `BackendPlanAdapter` 

Removing/Stripping annotation should only be about passing the cleanest possible RelNode to backends during FragmentConversion.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateRule.java
```diff
@@ -76,11 +77,21 @@ public void onMatch(RelOptRuleCall call) {
         // Annotate each AggregateCall with per-call viable backends
         List<AggregateCall> annotatedCalls = new ArrayList<>();
         for (AggregateCall aggCall : aggregate.getAggCallList()) {
-            List<String> callViable = resolveViableBackendsForCall(aggCall, childFieldStorage);
+            // Rewrite COUNT(DISTINCT x) → APPROX_COUNT_DISTINCT(x) for distributed execution.
```
expani: This can also be a part of `BackendPlanAdapter`

expani (commented):
Thanks for getting this going. @sandeshkr419 

Dropping some early comments on the approach.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/ArrowSchemaFromCalcite.java
sandeshkr419: Added a todo.

sandeshkr419:
[outdated]

@expani Thanks for the detailed review. Here's what I changed structurally based on your feedback:

- `AVG` no longer has a custom decomposition. It's handled by Calcite's `AggregateReduceFunctionsRule` in the pre-marking phase so backends only ever see `SUM` and `COUNT`. Similarly, `HLL` no longer has a decomposition either; the intermediate Arrow type (Binary) is now declared directly on AggregateCapability.intermediateArrowType, which LocalStageScheduler reads, no more ad-hoc `isApproximate() `check in the scheduler.

- The `COUNT(DISTINCT) → APPROX_COUNT_DISTINCT` rewrite moved from OpenSearchAggregateRule into `BackendPlanAdapter` via a new aggregateCallAdapters() hook on BackendCapabilityProvider, keeping backend-specific logic out of the generic planner rule. `stripAnnotations` was also simplified to only strip annotation wrappers.

- On the Rust side, the `StripFinalAggregateRule` and `StripPartialAggregateRule` physical optimizer rules are replaced by FragmentConvertor-driven mode forcing. `DataFusionFragmentConvertor` now prepends a 1-byte mode prefix to plan bytes (attachPartialAggOnTop →  partial, convertFinalAggFragment → final), and the Rust side calls force_aggregate_mode() directly after create_physical_plan, no plan inspection. **The FragmentConvertor API is the authoritative signal for execution mode.**

- **One open item:** DataFusion's Substrait consumer ignores `aggregation_phase`, which is why `force_aggregate_mode` exists. This could be a potential upstream contribution, if DataFusion adds `aggregation_phase support`, `force_aggregate_mode` can be removed entirely.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
```diff
@@ -89,31 +93,66 @@ public byte[] convertShardScanFragment(String tableName, RelNode fragment) {
         return convertToSubstrait(fragment);
     }
 
+    /** Mode byte prepended to plan bytes so the Rust side knows the aggregation phase. */
+    private static final byte AGG_MODE_PARTIAL = 0x01;
```
mch2: can we use AggregationPhase in substrait and read that rust side?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -82,7 +86,11 @@ static RelNode markAndOptimize(RelNode rawRelNode, PlannerContext context) {
         hepBuilder.addRuleCollection(
             List.of(
                 new ReduceExpressionsRule.FilterReduceExpressionsRule(Filter.class, RelBuilder.proto(Contexts.empty())),
-                new ReduceExpressionsRule.ProjectReduceExpressionsRule(Project.class, RelBuilder.proto(Contexts.empty()))
+                new ReduceExpressionsRule.ProjectReduceExpressionsRule(Project.class, RelBuilder.proto(Contexts.empty())),
+                // Rewrite AVG → SUM/COUNT + Project before marking so backends receive
+                // only primitive aggregates (SUM, COUNT) that they natively support.
+                new AggregateReduceFunctionsRule(Aggregate.class,
```
mch2: I am still on the fence about this decomposition.  This means we aren't using df's built in accumulators for these functions.  Are we going to have precision loss? correctness issues with other functions?

An alternative I have tried (and seems to work - cleaning up a diff to show) at the expense of calcite/planner level visibility is - we set data node to partial state,  then the issue becomes at coordinator registering the schema with the correct input types per stream. During table registration (streamingTableExec) we probe the partial state of the agg - during planning we can determine if the table should be registered with partial agg state or not.    We shouldn't have to manipulate the substrait plan.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
sandeshkr419: I tried this early on. DataFusion's Substrait consumer `aggregate_rel.rs` completely ignores `aggregation_phase`, it always creates a full `LogicalPlan::Aggregate` regardless of the phase. See last point on https://github.com/opensearch-project/OpenSearch/pull/21457#issuecomment-4365229520

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
sandeshkr419: Yours alternative would keep Substrait clean but should requires the Rust side to probe physical plan state at table registration time. The current would keep Rust side thin by having the Java `FragmentConvertor` API signal the mode explicitly. 

Happy to switch if your approach is cleaner overall. I think I was trying something similar in early iterations but the rust side was getting messier.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
mch2: right, we'd have to manually set partial

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
```diff
@@ -89,31 +93,66 @@ public byte[] convertShardScanFragment(String tableName, RelNode fragment) {
...
+    private static final byte AGG_MODE_FINAL   = 0x02;
+
+    /**
+     * Substrait function name → DataFusion function name remappings.
+     * Isthmus uses Substrait spec names; DataFusion may use different names.
+     */
+    private static final Map<String, String> FUNCTION_RENAMES = Map.of(
```
expani: Can this rename happen via RelNode adapter ? 

Does Isthumus respect the name when it appears in AggregateCall of LogicalAggregate ?

sandbox/plugins/analytics-backend-datafusion/rust/src/local_executor.rs
```diff
@@ -152,6 +157,38 @@ impl LocalSession {
...
+        bytes: &[u8],
+        mode: AggregateMode,
+    ) -> Result<SendableRecordBatchStream, DataFusionError> {
+        let plan = Plan::decode(bytes).map_err(|e| {
+            DataFusionError::Execution(format!("Failed to decode Substrait plan: {}", e))
+        })?;
+        let logical_plan = from_substrait_plan(&self.ctx.state(), &plan).await?;
+        let df = self.ctx.execute_logical_plan(logical_plan).await?;
+        let physical_plan = df.create_physical_plan().await?;
+        let physical_plan = force_aggregate_mode(physical_plan, mode)?;
```
expani: Can we think around returning the physical plan from here and keeping execution APIs unaware of the relational operators ( like Aggregate ) as much as possible ? 

Going forward, we might need manipulation for other relational operators as well

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -66,36 +75,92 @@ public boolean matches(RelOptRuleCall call) {
     public void onMatch(RelOptRuleCall call) {
         OpenSearchAggregate aggregate = call.rel(0);
         RelNode child = call.rel(1);
+        RexBuilder rexBuilder = aggregate.getCluster().getRexBuilder();
+
+        // All viable backends for a resolved plan alternative share the same decompositions
```
expani: Unless there is an aggregate delegation in the scenario like Painless.

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -563,7 +563,13 @@ pub async unsafe fn execute_local_plan(
     // `context_id` of 0 disables tracking (pool is not consulted).
     let query_context = QueryTrackingContext::new(context_id, session.memory_pool());
 
-    let df_stream = session.execute_substrait(substrait_bytes).await?;
+    let (mode_byte, plan_bytes) = crate::local_executor::strip_mode_prefix(substrait_bytes);
+
+    let df_stream = match mode_byte {
```
expani: Can't the mode be determined by the Fragment Conversion API instead ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/LocalStageScheduler.java
```diff
@@ -67,12 +83,63 @@ private static byte[] chosenBytes(Stage stage) {
      * fragment rowtype. Multi-child support (joins, set ops with heterogeneous
      * inputs) is deferred.
      */
-    private static Schema deriveInputSchema(Stage stage) {
+    private static Schema deriveInputSchema(Stage stage, CapabilityRegistry capabilityRegistry) {
```
expani: Can we add a TODO to move this upfront ? 

Unrelated to this PR, but scheduling shouldn't become another planner. This has carried forward due to the lack of proper Physical Execution Operators.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
sandeshkr419: The rename can't happen via RelNode adapter because isthmus looks up the function binding by name from its extension collection, which only has `approx_count_distinct`. Mapping to `approx_distinct` in the adapter would break the binding lookup. The proto-level rename after serialization is the only option until DataFusion aligns its function name with the Substrait spec.

sandbox/plugins/analytics-backend-datafusion/rust/src/local_executor.rs
sandeshkr419: Agreed that `force_aggregate_mode` shouldn't be in the execution path long-term. We can refactor this once DataFusion supports `aggregation_phase` in its Substrait consumer, at that point `force_aggregate_mode` goes away entirely.

Maybe a good time to contribute to upstream - https://github.com/opensearch-project/OpenSearch/pull/21457#issuecomment-4365229520 (last point)

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
sandeshkr419: Will clarify this in comment.

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
sandeshkr419: Refactoring the code now to keep planning and execution separate and orchestrate this from java entirely.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/PartialAggregateInstructionHandler.java
```diff
@@ -0,0 +1,32 @@
...
+class PartialAggregateInstructionHandler implements FragmentInstructionHandler<PartialAggregateInstructionNode> {
+
+    @Override
+    public BackendExecutionContext apply(
+        PartialAggregateInstructionNode node,
+        CommonExecutionContext commonContext,
+        BackendExecutionContext backendContext
+    ) {
+        DataFusionSessionState prev = (DataFusionSessionState) backendContext;
+        return new DataFusionSessionState(prev.sessionContextHandle(), AggregateExecutionMode.PARTIAL);
```
expani: The idea is that this goes into the Rust side and manipulates the physical plan. This becomes the caller for rust code essentially

sandeshkr419: That's exactly what it does. The handler sets `AggregateExecutionMode.PARTIAL` on the session state → DatafusionSearcher reads it → dispatches to `NativeBridge.executeWithContextAsync(..., PARTIAL, ...)` → calls `df_execute_with_context_partial` in Rust → which calls `force_aggregate_mode(plan, 1)` → manipulates the physical plan to emit partial state.

The handler IS the caller for the Rust code, it's just indirect (through the session state → searcher → NativeBridge → FFM → Rust).

expani: Then DatafusionSearcher is becoming a monolith handler.

For instance, the ShardScanInstructionHandler actually calls the FFM API and creates the SessionContext. It doesn't handoff to Searcher for that

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
```diff
@@ -139,7 +147,43 @@ public Set<AggregateCapability> aggregateCapabilities() {
                 Set<AggregateCapability> caps = new HashSet<>();
                 for (AggregateFunction func : AGG_FUNCTIONS) {
                     for (FieldType type : SUPPORTED_FIELD_TYPES) {
-                        caps.add(AggregateCapability.simple(func, Set.of(type), formats));
+                        if (func == AggregateFunction.APPROX_COUNT_DISTINCT) {
+                            caps.add(AggregateCapability.approximate(func, Set.of(type), formats, ArrowType.Binary.INSTANCE));
+                        } else if (func == AggregateFunction.AVG) {
+                            caps.add(
```
mch2: i think its time we pull some things out of DataFusionAnalyticsBackendPlugin - starting here.  

I like this better than the rule based decomp

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/PartialAggregateInstructionHandler.java
sandeshkr419: Got it, found the issue. Let me revisit this.

mch2 (approved):

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
```diff
@@ -365,6 +424,65 @@ private static byte[] serializePlan(Plan plan) {
...
+     * emission so the isthmus {@link AggregateFunctionConverter} resolves it through
+     * {@link #ADDITIONAL_AGGREGATE_SIGS} → {@code approx_distinct} (declared in
+     * {@code opensearch_aggregate_functions.yaml}) instead of Substrait's default
+     * {@code approx_count_distinct} under the standard URN.
+     *
+     * <p>A plain-swap preserves the call's return type (already {@code BIGINT NOT NULL}
+     * from stock inference); our custom operator uses the same {@code ReturnTypes.BIGINT}
+     * strategy so {@code Aggregate.typeMatchesInferred} continues to pass.
+     */
+    private static RelNode rewriteApproxCountDistinct(RelNode node) {
```
expani: We can write a BackendPlanAdapter for this. Also, import statemetns

expani (approved):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/AggregateDecompositionResolver.java
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * Rewrites PARTIAL/FINAL aggregate pairs so the exchange row type precisely
+ * describes what the engine emits. Uses {@link AggregateFunction#intermediateFields()}
+ * as the single source of truth — no downstream type overrides needed.
+ *
+ * <p>Runs after {@link BackendPlanAdapter} and before {@link FragmentConversionDriver}.
+ *
+ * @opensearch.internal
+ */
+public final class AggregateDecompositionResolver {
```
expani, line 44: This class is doing a lot of re-traversing the entire DAG for handling let's add a TODO to revisit later

mch2:
related - https://github.com/opensearch-project/OpenSearch/pull/21584

mch2 (approved):

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
sandeshkr419: Sorry, this crept in later - I have removed it entirely.
We control this via additional signature strategy where I register the custom operator via yaml.

PR #21460, opened 2026-05-02
[Pluggable Engine] Fixes for end to end ingestion in DataFormatAwareEngine

rayshrey:
### Description

This PR fixes multiple issues that we observed in our end to end testing during ingestion for DataFormatAwareEngine.

1. File deletion bug due to format name and relative path -  fixed by having single deleter instead of maps and resolving the relative path to absloute in the plugin
2. deleteCommit called on non-committed snapshot - fixed by making catalog snapshot aware of whether it is committed or not, and backed by assertions on the LuceneCommitter to throw when untracked snapshot id is called for delete
3. Merges naming convention bug - generation counter was not integrated in merge flow, fixed it
4. Reduce number of getVector calls for VectorSchemaRoot
5. Common Lockable entity and row-id aware generation
6.  Added more concrete assertions in multiple places                                         

### Related Issues
N/A

### Check List


sandbox/plugins/parquet-data-format/src/main/resources/plugin-security.policy
```diff
@@ -5,3 +5,7 @@
  * this file be licensed under the Apache-2.0 license or a
  * compatible open source license.
  */
+
+grant {
+  permission java.io.FilePermission "<<ALL FILES>>", "delete";
```
alchemist51: Why do we need this?

rayshrey: This is to ensure that parquet plugin has sufficient permissions for deleting stale files.
Since the call to delete files does not originate from the plugin (delegated from IndexFileDeleter -> CompositeEngine plugin -> ParquetPlugin), parquet plugin gets AccessDenied Error while deleting.

This security policy ensures that the plugin has sufficient permissions to delete the required files.

alchemist51: Are we doing this somewhere else as well? Not sure on the security angle of this. Maybe check with release/security team once if this is allowed.

mgodwan (approved):

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshotManager.java
alchemist51, line 109: Should we update the snapshotListeners with the lastCatalogSnapshot also here?

mgodwan: Lets handle this in a separate PR

PR #21463, opened 2026-05-03
Speed up Script Engine initialization

timlueg:
### Description
Small improvements to Painless Script Engine initialization to reduce OpenSearch startup time.
- Compile the whitespace pattern only once
- Reuse `PainlessLookup` instances for identical allowlist sequences
- Use sets to avoid repeated linear scans during lookup validation

| Version |Time to Node "started" using ./gradlew run (best of 5) |
|---------|------|
| Before  | 2012 ms |
| After   | 1815 ms |

Improvement: ~197 ms, about 9.8%.

### Check List

### DIFF
--- a/modules/lang-painless/src/main/java/org/opensearch/painless/PainlessScriptEngine.java
+++ b/modules/lang-painless/src/main/java/org/opensearch/painless/PainlessScriptEngine.java
@@ -95,11 +95,7 @@ public final class PainlessScriptEngine implements ScriptEngine {
         for (Map.Entry<ScriptContext<?>, List<Allowlist>> entry : contexts.entrySet()) {
             ScriptContext<?> context = entry.getKey();
             List<Allowlist> allowlists = List.copyOf(entry.getValue());
-            PainlessLookup lookup = allowlistsToLookups.get(allowlists);
-            if (lookup == null) {
-                lookup = PainlessLookupBuilder.buildFromAllowlists(allowlists);
-                allowlistsToLookups.put(allowlists, lookup);
-            }
+            PainlessLookup lookup = allowlistsToLookups.computeIfAbsent(allowlists, PainlessLookupBuilder::buildFromAllowlists);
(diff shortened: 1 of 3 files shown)


modules/lang-painless/src/main/java/org/opensearch/painless/PainlessScriptEngine.java
```diff
@@ -90,10 +90,16 @@ public PainlessScriptEngine(Settings settings, Map<ScriptContext<?>, List<Allowl
...
 
         for (Map.Entry<ScriptContext<?>, List<Allowlist>> entry : contexts.entrySet()) {
             ScriptContext<?> context = entry.getKey();
-            PainlessLookup lookup = PainlessLookupBuilder.buildFromAllowlists(entry.getValue());
+            List<Allowlist> allowlists = List.copyOf(entry.getValue());
+            PainlessLookup lookup = allowlistsToLookups.get(allowlists);
+            if (lookup == null) {
+                lookup = PainlessLookupBuilder.buildFromAllowlists(allowlists);
+                allowlistsToLookups.put(allowlists, lookup);
+            }
```
cwperks: nit: could be simplified to `allowlistsToLookups.computeIfAbsent(allowlists, PainlessLookupBuilder::buildFromAllowlists)`

cwperks:
@timlueg this is great! Thank you for looking into this. We've run into issues with bootstrap time on the CI checks of the security repo and this looks like it will help.

modules/lang-painless/src/main/java/org/opensearch/painless/PainlessScriptEngine.java
timlueg: Done!

timlueg:
@cwperks then the CI might like this PR even more :) but it's a bit less straightforward https://github.com/opensearch-project/OpenSearch/pull/21473

cwperks (approved):

PR #21464, opened 2026-05-04
Avoiding repeated encoding and compression operation for the sort column included writes

chaitanya588:
## Summary

When sort columns are configured, the writer now stages incoming batches as Arrow IPC instead of Parquet, avoiding redundant Parquet encoding and compression before the sort-on-close step. For large files, batches are individually sorted and written as Parquet chunks, then merged via the existing streaming k-way merge. The per-batch sort uses Arrow's `RowConverter` for compact byte-comparable row encoding. 

### Arrow IPC staging for sorted writes

Previously, all data was written as Parquet regardless of whether sort columns were configured. On finalize, the Parquet file was read back (decoded), sorted, and re-written as Parquet — a full encode-decode-encode cycle. Now, when sort columns are present, a `WriterVariant::Ipc` path writes batches as Arrow IPC (raw in-memory Arrow buffers with minimal framing). On finalize, the IPC file is read back with near-zero deserialization cost, sorted, and written as the final Parquet file — a single encode step.

When no sort columns are configured, the `WriterVariant::Parquet` path is used as before, and finalize simply renames the temp file to the final path.

### Small and large file sort strategies

The sort-on-close path (`sort_and_rewrite_parquet`) uses a size-based threshold (`sort_in_memory_threshold_bytes`, default 32 MB) to choose between two strategies:

- Small files (`sort_small_file`): All IPC batches are read into memory, concatenated, sorted, row IDs rewritten, and written as a single Parquet file.
- Large files (`sort_large_file`): Each IPC batch is sliced into `sort_batch_size` chunks (to bound memory since IPC batches can be arbitrarily large), each chunk is sorted individually and written as a temporary Parquet chunk file, then the existing streaming k-way merge produces the final globally-sorted Parquet output. Chunk files are cleaned up after the merge.

### RowConverter-based per-batch sort

The `sort_batch` function now uses Arrow's `RowConverter` instead of `lexsort_to_indices`. `RowConverter` converts sort columns into compact byte-comparable rows where sort direction and null ordering are baked into the encoding. Sorting then reduces to comparing opaque byte sequences via `sort_unstable_by`, followed by a `take` to reorder all columns. This approach handles all Arrow data types uniformly without a per-type dispatch in the sort comparator.

### CRC32 integration

The IPC staging path integrates with the existing `CrcWriter`-based CRC32 computation. Since IPC batches are not Parquet-encoded, the CRC is computed during the final Parquet write step (`write_final_file` for small files, `merge_sorted` for large files) rather than during the initial staging write. The `WriterState.crc_handle` field is `Option<CrcHandle>` — `Some` for the Parquet variant, `None` for IPC.

### Testing

Existing tests cover both sorted (IPC path) and unsorted (Parquet path) writer lifecycles, including concurrent writers, empty data, multi-batch sorting, ascending/descending sort, and mixed IPC+Parquet coexistence. New integration tests validate the IPC staging cleanup, sorted output correctness, and concurrent sorted writer behavior.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/parquet-data-format/src/main/rust/src/merge/heap.rs
+++ b/sandbox/plugins/parquet-data-format/src/main/rust/src/merge/heap.rs
@@ -14,7 +14,7 @@ use arrow::datatypes::{
     DataType as ArrowDataType, Date32Type, Date64Type, DurationMicrosecondType,
     DurationMillisecondType, DurationNanosecondType, DurationSecondType, Float32Type, Float64Type,
     Int16Type, Int32Type, Int64Type, Int8Type, TimestampMicrosecondType,
-    TimestampMillisecondType, TimestampNanosecondType, TimestampSecondType, UInt32Type,
+    TimestampMillisecondType, TimestampNanosecondType, TimestampSecondType,
 };
 
 use super::error::{MergeError, MergeResult};
@@ -138,7 +138,6 @@ pub fn get_sort_value(
         ArrowDataType::Int32 => SortKey::Int(col.as_primitive::<Int32Type>().value(row) as i64),
         ArrowDataType::Int16 => SortKey::Int(col.as_primitive::<Int16Type>().value(row) as i64),
         ArrowDataType::Int8 => SortKey::Int(col.as_primitive::<Int8Type>().value(row) as i64),
-        ArrowDataType::UInt32 => SortKey::Int(col.as_primitive::<UInt32Type>().value(row) as i64),
         ArrowDataType::Date32 => SortKey::Int(col.as_primitive::<Date32Type>().value(row) as i64),
         ArrowDataType::Date64 => SortKey::Int(col.as_primitive::<Date64Type>().value(row)),
         ArrowDataType::Timestamp(unit, _) => SortKey::Int(match unit {
--- a/sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
+++ b/sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
@@ -8,16 +8,19 @@
 
 use arrow::ffi::{FFI_ArrowArray, FFI_ArrowSchema};
 use arrow::record_batch::RecordBatch;
-use arrow::compute::{lexsort_to_indices, take, SortColumn};
+use arrow::compute::{concat_batches, take};
+use arrow::row::{RowConverter, SortField};
+use arrow_ipc::writer::FileWriter as IpcFileWriter;
+use arrow_ipc::reader::FileReader as IpcFileReader;
 use dashmap::DashMap;
 use lazy_static::lazy_static;
-use parquet::arrow::{arrow_reader::ParquetRecordBatchReaderBuilder, ArrowWriter};
+use parquet::arrow::ArrowWriter;
 use parquet::file::reader::{FileReader, SerializedFileReader};
 use std::fs::File;
 use std::path::Path;
 use std::sync::{Arc, Mutex};
 
-use crate::{log_error, log_debug};
+use crate::{log_error, log_debug, log_info};
 use crate::crc_writer::CrcWriter;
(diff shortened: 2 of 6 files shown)


sandbox/plugins/parquet-data-format/src/main/rust/src/merge/heap.rs
```diff
@@ -139,6 +140,9 @@ pub fn get_sort_value(
         ArrowDataType::Int16 => SortKey::Int(col.as_primitive::<Int16Type>().value(row) as i64),
         ArrowDataType::Int8 => SortKey::Int(col.as_primitive::<Int8Type>().value(row) as i64),
         ArrowDataType::UInt32 => SortKey::Int(col.as_primitive::<UInt32Type>().value(row) as i64),
+        ArrowDataType::UInt16 => SortKey::Int(col.as_primitive::<UInt16Type>().value(row) as i64),
+        ArrowDataType::UInt8 => SortKey::Int(col.as_primitive::<UInt8Type>().value(row) as i64),
+        ArrowDataType::UInt64 => SortKey::Int(col.as_primitive::<UInt64Type>().value(row) as i64),
```
mgodwan: Do we support these types in. opensearch?

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -30,16 +33,33 @@ pub struct FinalizeResult {
...
+/// When sort columns are configured, the IPC variant is used so that
+/// batches can be cheaply read back for sorting — Arrow IPC is a raw
+/// dump of in-memory Arrow buffers with minimal framing overhead.
+enum WriterVariant {
+    /// Direct Parquet writer — used when no sort columns are configured.
+    Parquet(Arc<Mutex<ArrowWriter<CrcWriter<File>>>>),
+    /// Arrow IPC staging writer — used when sort columns are configured.
+    /// Batches are written as raw Arrow IPC; on close they are read back,
+    /// sorted, and written as a final Parquet file.
+    Ipc(Arc<Mutex<IpcFileWriter<File>>>),
```
mgodwan, line 46: Can we verify that the crc during write is same as crc when reading back the file to be written as parquet?

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/heap.rs
chaitanya588: In Opensearch we supported UnsignedLong, so UInt64 is still relevant. The other ones are not supported in opensearch, so logically there will be no way to get them indexed through OpenSearch directly. However as a defensive check, keeping these additional one will help in reading the parquet files which are already written (may be for testing purpose or for independent benchmarking it will be helpful). Do you see any value in keeping this?

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
chaitanya588: In the sorted options scenario, there are 2 places where we can calculate the crc during the write step on-fly (1. Overall unsorted IPC file, 2. Chunked IPC Files), but for verifying the CRC during read requires an additional sequential read to calculate the CRC in the same order in which the file is written. This requires an additional effort to handle the calculation in optimal manner. Tracking this as a separate task now since it is not a blocker for this.

Tracking ticket: https://github.com/opensearch-project/OpenSearch/issues/21488

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/heap.rs
chaitanya588: I removed all the unsigned types which are not supported by OpenSearch at the first place. Hence deleted uinit32 as well.

mgodwan (approved):

mgodwan (approved):

PR #21454, opened 2026-05-01
Skip root transfer for byte-serialized via handler marker

bowenlan-amzn:
### Description

On the receiving side of stream transport, `FlightTransportResponse.nextResponse()` transfers the Flight stream root into a response-owned root on every batch — but byte-serialized responses copy bytes into Java fields via handler.read() and never touch the vectors. The transfer is wasted work on that path.

This PR splits VectorStreamInput into two subclasses and let the handler decide which path to take.

```mermaid
flowchart LR
    stream["FlightStream<br/>(stream root)"]
    dispatch{"handler.skips<br/>Deserialization()?"}
    byte["ByteSerialized<br/>reads bytes from stream root"]
    native["NativeArrow<br/>transfers into consumer root"]

    stream --> dispatch
    dispatch -- false --> byte
    dispatch -- true --> native
```

#### Handler marker

TransportResponseHandler gains a skipsDeserialization() bit (default false). ArrowBatchResponseHandler pins it to true. Wrappers forward their delegate's value:

```mermaid
flowchart LR
    M["MetricsTracking"] --> C["ContextRestore"] --> T["Traceable<br/>(optional)"] --> U["User handler"]

    M -. "skipsDeserialization()" .-> U
```

#### Ownership lifecycle of Arrow on the Receive End

```mermaid
sequenceDiagram
    participant FS as FlightStream
    participant FTR as FlightTransportResponse
    participant VSI as VectorStreamInput.NativeArrow
    participant ABR as ArrowBatchResponse

    FTR->>FS: getRoot()
    FS-->>FTR: stream root
    FTR->>VSI: newStreamInput(streamRoot)
    Note over VSI: transfer to consumer
    FTR->>ABR: handler.read(input)
    ABR->>VSI: getRoot() + claimOwnership()
    Note over ABR: response now owns the root
    FTR->>VSI: close()
    Note over VSI: no-op (ownership claimed)
```

#### Shared allocator

ArrowAllocatorProvider provides a single node-level RootAllocator so all Arrow plugins share the same root — required for zero-copy transfers to pass Arrow's AllocationManager associate check.

```mermaid
flowchart TD
    root["ArrowAllocatorProvider.ROOT"]
    flight["flight"]
    server["server"]
    client["client"]
    plugin["plugin allocators<br/>(e.g. analytics-search)"]

    root --> flight
    root --> plugin
    flight --> server
    flight --> client
```

### Who should close the Arrow resource

Send: arrow batch is transfered from producer to flight, so flight should close
[...]

PR #21465, opened 2026-05-04
Extract arrow-base plugin

bowenlan-amzn:
### Description

Apache Arrow's JVM classes have object identity tied to `(classloader, FQN)`. If two plugins each bundle their own copy of `arrow-vector`, a `VectorSchemaRoot` produced by one can't be consumed by the other — different classloaders, different classes, `ClassCastException` at the handoff. This PR introduces `arrow-base` as the single place Arrow lives: every Arrow-using plugin declares `extendedPlugins = ['arrow-base']` and shares its classloader.

```mermaid
flowchart TD
  base[arrow-base<br/>bundles Arrow + Netty]
  flight[arrow-flight-rpc]
  analytics[analytics-engine]
  parquet[parquet-data-format]
  dfusion[analytics-backend-datafusion]
  flight --> base
  analytics --> base
  parquet --> base
  dfusion --> analytics
```

### `ArrowAllocatorService`

The previous static `ArrowAllocatorProvider` is replaced with a proper node-level service. `ArrowBasePlugin` provides `DefaultArrowAllocatorService` — one root allocator per node, child allocators for each consumer.

### `PluginComponentRegistry`

Plugins are initialized in dependency order (topologically sorted by `extendedPlugins`). This PR introduces `PluginComponentRegistry` — a typed lookup that accumulates components as each plugin initializes. Later plugins discover services from their dependencies at `createComponents()` time without Guice or deferred resolution.

`FlightStreamPlugin` and `AnalyticsPlugin` obtain `ArrowAllocatorService` from the registry directly. `ArrowBasePlugin` also provides a Guice binding via `createGuiceModules()` for `@Inject` consumers (`DefaultPlanExecutor`, transport actions).

### Plugin close order

`Node.close()` now closes plugins in reverse dependency order, so child allocators are released before the root. This eliminates the need for defensive exception swallowing on shutdown.

### Testing

Integration tests using `FlightStreamPlugin` must declare it in `additionalNodePlugins()` with `extendedPlugins = [ArrowBasePlugin]` so the test framework sorts initialization correctly (the standard `nodePlugins()` API doesn't preserve dependency metadata).

### Non-goals

`arrow-c-data` stays in parquet (FFM-only). Flight-internal types stay in flight — only types used by ≥2 plugins move.

### Related Issues
N/A

### Check List


plugins/arrow-flight-rpc/build.gradle
```diff
@@ -13,36 +13,37 @@ apply plugin: 'opensearch.internal-cluster-test'
...
 }
 
 dependencies {
-  // all transitive dependencies exported to use arrow-vector and arrow-memory-core
-  api "org.apache.arrow:arrow-memory-netty:${versions.arrow}"
-  api "org.apache.arrow:arrow-memory-core:${versions.arrow}"
-  api "org.apache.arrow:arrow-memory-netty-buffer-patch:${versions.arrow}"
-  api "io.netty:netty-buffer:${versions.netty}"
-  api "io.netty:netty-common:${versions.netty}"
+  // arrow-base provides the Arrow + Netty core / Jackson / flatbuffers / slf4j / commons-codec
```
rishabhmaurya: these are already api dependencies in arrow-base, you can just use project dependency instead of adding them as compileOnly

plugins/arrow-base/src/main/java/org/opensearch/arrow/transport/ArrowAllocatorProvider.java
rishabhmaurya: this is not a right approach to expose factory for creating sub allocators. Check my comment here - https://github.com/opensearch-project/OpenSearch/pull/21454#discussion_r3184561844

sandbox/plugins/analytics-engine/build.gradle
```diff
@@ -20,10 +20,11 @@ def sqlUnifiedQueryVersion = '3.6.0.0-SNAPSHOT'
...
   description = 'Analytics engine hub: discovers and wires query extensions via ExtensiblePlugin SPI.'
   classname = 'org.opensearch.analytics.AnalyticsPlugin'
-  // Extend arrow-flight-rpc so analytics-engine and arrow-flight-rpc share one classloader.
-  // Cross-plugin Arrow types (VectorSchemaRoot, ArrowBatchResponse) only work when loaded
-  // by the same classloader, and zero-copy transfer requires identical class identity on both sides.
-  extendedPlugins = ['arrow-flight-rpc']
+  // Extend arrow-base so analytics-engine shares one Arrow classloader with any peer plugin
+  // (arrow-flight-rpc, analytics-backend-datafusion, parquet-data-format, …). Cross-plugin
+  // Arrow types (VectorSchemaRoot, ArrowBatchResponse) only work when loaded by the same
+  // classloader, and zero-copy transfer requires identical class identity on both sides.
```
rishabhmaurya: its not because zero copy requirements. This won't work fundamentally the way classloader resolution happens in java. Maybe fix the comment here

bowenlan-amzn: fixed

plugins/arrow-flight-rpc/build.gradle
bowenlan-amzn: Thanks! I update other places also.
Will do a dependency audit in another PR later.

rishabhmaurya:
@bowenlan-amzn 
I'm not fully convinced by the `ArrowAllocatorServiceBridge + ArrowAllocatorServiceHolder` pattern. It's duplicated in every consumer plugin, relies on implicit Guice ordering, and could masks real resource leaks. 

Suggestion: introduce a small plugin SPI in server instead.

Two interfaces (without Arrow deps in server):
```
// Generic service — T is BufferAllocator in practice
public interface NativeMemoryAllocatorService<T extends Closeable> extends Closeable {
   T newChildAllocator(String name, long limit);
   long getAllocatedMemory();
   long getPeakMemoryAllocation();
}

// Consumer plugins implement this to receive the service
public interface NativeMemoryAllocatorConsumer {
   void setNativeMemoryAllocatorService(NativeMemoryAllocatorService<?> service);
}
```
`arrow-base` provides the implementation, consumer plugins implement the consumer interface:

```
   public class FlightStreamPlugin extends Plugin
       implements NetworkPlugin, NativeMemoryAllocatorConsumer {

       private NativeMemoryAllocatorService<BufferAllocator> allocatorService;

       @Override
       public void setNativeMemoryAllocatorService(NativeMemoryAllocatorService<?> service) {
           this.allocatorService = (NativeMemoryAllocatorService<BufferAllocator>) service;
       }

       // getTransports() just uses this.allocatorService — no bridge needed
   }
```
Initialization order (in  Node.java):

   1. `createComponents()` runs in extendedPlugins order → arrow-base creates the service first
   2. Node finds the `NativeMemoryAllocatorService` in the collected components
   3. Node calls `setNativeMemoryAllocatorService()` on all `NativeMemoryAllocatorConsumer` plugins
   4. Node calls `getTransports()` — consumers already have the service
   5. Guice injector built — service also bound for `@Inject` in transport actions

Close order (in Node.java):

   1. Node stops transports → consumers close their child allocators
   2. Node closes plugins → any remaining children closed
   3. Node closes NativeMemoryAllocatorService last → root.close() validates no leaks

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -58,32 +58,42 @@ public class AnalyticsSearchService implements AutoCloseable {
...
         this.namedWriteableRegistry = namedWriteableRegistry;
     }
 
     @Override
     public void close() {
-        allocator.close();
+        try {
+            allocator.close();
+        } catch (IllegalStateException ignored) {
+            // Root may have already been closed by arrow-base's plugin shutdown, which cascades
```
rishabhmaurya: As proposed [before](https://github.com/opensearch-project/OpenSearch/pull/21465#issuecomment-4453080926), if the AllocatorService exist as a SPI in server then in Node.java we can close it at the end after all plugins are closed. I do not want to block this PR though, but this looks risky as it may swallow actual leaks. Maybe you can take it up as a follow up improvement later on

rishabhmaurya (approved):

PR #21468, opened 2026-05-04
Sort the documents according to the indexSort during the refresh time

chaitanya588:
## Summary

This PR enables physical document sorting by the index sort (e.g. timestamp) at refresh time across both Parquet (primary) and Lucene (secondary) data formats in the composite indexing engine. The Parquet writer sorts batches as they accumulate and emits a row-ID permutation on flush. That permutation is propagated to the Lucene secondary writer so its segment is reordered to match Parquet's new physical row order, keeping both formats aligned by __row_id__.

It also restructures the refresh flow with a cooperative flush queue so write threads share flush work with the refresh thread (backpressure), and cleans up the row-ID mapping API by unifying it into a single class for supporting both Merge and Refresh Flows.

## Why
Without sort-on-refresh:

Range queries on the index sort field (typically @timestamp) can't take advantage of segment ordering for early-termination or block-skip optimizations.
Parquet row groups can't prune by min/max effectively when rows are inserted in arrival order.
Lucene and Parquet would diverge in physical row order, breaking the __row_id__ 1:1 mapping the composite engine relies on for cross-format reads.
This change closes that gap by sorting eagerly during chunked Parquet writes and applying the same permutation to the Lucene segment via a custom MergePolicy + reorder pass + DocValues row-ID rewrite.

### 1. Sort-on-flush in the Parquet writer (Rust)

`sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs`
- New `SortingChunkedWriter` replaces the simple Arrow IPC staging path when `sort_columns` are configured.
- Strategy: write incoming batches to an Arrow IPC staging file; when accumulated bytes plus the next batch would exceed `memory_threshold_bytes`, read the IPC back, sort in memory, write a sorted Parquet chunk, then reopen a fresh IPC. Oversized single batches are sliced to fit the budget.
- On `finish()`: flush trailing IPC, return the list of sorted Parquet chunks. A k-way merge stitches them into the final Parquet file, producing a flat row-ID permutation `mapping[original_row_id] = new_row_id`.
- The chunk file rewrites `___row_id` to sequential `0..N` so each chunk is self-consistent; original row IDs are tracked separately for the permutation.

### 2. Permutation handoff Rust → Java

`sandbox/plugins/parquet-data-format/src/main/rust/src/ffm.rs`, `RustBridge.java`, `NativeParquetWriter.java`
- `parquet_finalize_writer` now returns two extra out-params: `sort_perm_ptr_out` and `sort_perm_len_out`, pointing to a heap-allocated `Box<[i64]>`.
- New `parquet_free_row_id_mapping` FFI to release the Rust-owned buffer after the Java side copies it.
- `RustBridge.finalizeWriter` now returns a `WriterFinalizeResult(metadata, rowIdMapping)` and constructs a `PackedRowIdMapping(mappingArray, true)` with reverse-lookup support, then frees the native buffer.

### 3. RowIdMapping API unification (server core)

`server/src/main/java/org/opensearch/index/engine/dataformat/`
- `RowIdMapping` interface gains:
  - `SINGLE_GEN = -1L` sentinel for single-generation flush-sort cases.
  - `getOldRowId(long newId)` reverse lookup.
  - `isNewToOldSupported()` capability check.
  - `size()`.
- `PackedRowIdMapping` now handles both multi-generation merge mappings and single-generation flush permutations through two constructors. The legacy `PackedSingleGenRowIdMapping` is removed.
- New `FlushInput` record carries the optional `RowIdMapping` into `Writer.flush(FlushInput)`. `Writer.flush()` is replaced with `flush(FlushInput)`; `FlushInput.EMPTY` is the no-op default.
- `FileInfos` carries `rowIdMapping` so flush results can hand the permutation to the parent writer.

### 4. Sort propagation in CompositeWriter

`sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeWriter.java`
- Flush order is now: primary (Parquet) first → capture its `RowIdMapping` → pass it via `FlushInput` to every secondary writer (Lucene).

### 5. Sorted flush in the Lucene writer

`sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/`
- `LuceneWriter.flush(FlushInput)`:
  - When a `RowIdMapping` is present, validates `mapping.size() == docCount`, configures a custom `ReorderingMergePolicy`, then `forceMerge(1)` triggers a single physical reorder of the segment.
  - After commit + close, rewrites the segment's `.si` to declare the IndexSort (`SortedNumericSortField` on `__row_id__`) so `addIndexes(Directory…)` on the shared writer sees matching sort metadata.
- `ReorderingMergePolicy` + `ReorderingOneMerge` use the mapping as a Lucene `Sorter.DocMap` (`oldToNew` / `newToOld`) to drive the physical reorder.
- `LuceneWriterCodec` gains `enableRowIdRewrite()`, which wraps the codec's `DocValuesFormat` with the new `LuceneWriterDocValuesFormat`. This intercepts writes to the `___row_id` field during the reorder merge and replaces values with sequential `0..N`, so reorder + ID rewrite happen in a single pass.

### 6. Cooperative refresh + write-side backpressure

`server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java`
- New shared `flushQueue` of writers pending flush, drained cooperatively by the refresh thread and write threads.
- A `CountDownLatch` (`activeFlushLatch`) tracks per-cycle completion; both refresh and write threads decrement it.
- New `preIndex()` (gated by `index.check_pending_flush.enabled`) lets a write thread pull a writer off `flushQueue` and flush it inline before its own indexing op. This provides backpressure on indexing while helping drain flush work.
- Inline-flushed segments are queued in `pendingSegments`; their writers are deferred-closed via `pendingWritersToClose` (since temp dirs must survive until refresh runs `addIndexes`).
- Refresh thread waits on the latch, drains both pending queues, then runs catalog commit. Added timing logs around flushAll, engine refresh, and catalog commit.

### 7. Tests

- `LuceneWriterSortedFlushTests` — validates sorted-flush with mapping handoff.
- `LuceneWriterDocValuesFormatTests` — verifies `___row_id` is rewritten to sequential values.
- `CompositeWriterSortPropagationTests` — verifies the primary→secondary `FlushInput` handoff.
- `CompositeRefreshSortedIT` — end-to-end IT covering sorted refresh across the composite engine.

### End to End Testing Scenarios:

CompositeRefreshSortedIT: A detailed list of correctness tests which ensures both primary and secondary formats correct end to end (with the final ordering .

1. testSortedRefreshProducesSortedParquet: (Parquet-only) Indexes 5 unsorted docs, flushes once. Asserts only parquet format is present, row count is 5, Parquet rows obey the sort order, and __row_id__ in Parquet is sequential starting at 0. 

3. testSortedRefreshWithLuceneSecondary: (Parquet + Lucene) Indexes 5 docs that also have a non-sort-key keyword field tag, flushes once. Asserts both formats are present, Parquet row count and sort order, Lucene doc count, sequential __row_id__ doc values in Lucene, and cross-format alignment: at every physical position, Parquet row i and Lucene doc i agree on name and tag (where tag confirms non-sort fields are also co-located by the row-id rewrite).

4. testMultipleSortedRefreshesProduceIndependentlySortedSegments: (Parquet-only) Indexes one batch, flushes; indexes a second batch, flushes. Asserts the snapshot has two segments and that each segment is independently sorted (sort is per-segment, not cross-segment).

5. testSortedRefreshWithNulls: (Parquet + Lucene) Mix of docs with and without age to 
exercise null handling. Asserts both formats are present, Parquet sort obeys nulls-first on age and nulls-last on name, and Lucene __row_id__ is sequential — confirming the row-id rewrite driven by the Parquet sort permutation works even when nulls participate in the key.

6. testSortedRefreshWithLargeBatch: (Parquet + Lucene) Indexes 200 docs to push past the chunked-sort path (default sort_batch_size = 65536 is much larger, so this mainly stresses non-trivial volume across both formats). Asserts row count, sort order, Lucene doc count, and sequential Lucene __row_id__. 

### Related Issues

### Check List


server/src/main/java/org/opensearch/index/engine/dataformat/RowIdMapping.java
```diff
@@ -23,7 +23,39 @@ public interface RowIdMapping {
...
+    /**
+     * Returns the old doc position for the given new doc position.
+     * Required by Lucene's Sorter.DocMap for physical reordering.
+     *
+     * @param newDocId the new document position
+     * @return the original document position, or the input if out of range
+     */
+    default int newToOld(int newDocId) {
+        return newDocId;
+    }
```
mgodwan: Would it be better to have a single signature only?
1We can convert RowIdMapping in merge flow to Map<Generation(Long), RowIdMapping>

mgodwan: Also, lets not implement these methods in the interface. We should implement these in each class.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -13,31 +13,46 @@
...
 
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Executor;
+import java.util.concurrent.locks.ReentrantLock;
```
mgodwan: Seems unused

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -147,27 +166,51 @@ public WriteResult addDoc(LuceneDocumentInput input) throws IOException {
...
-    public FileInfos flush() throws IOException {
+    public FileInfos flush(FlushInput flushInput) throws IOException {
         if (docCount == 0) {
             return FileInfos.empty();
         }
 
-        // Force merge to exactly 1 segment to maintain 1:1 mapping with other formats.
+        // If sort permutation is provided, configure the reorder merge policy
+        if (flushInput.hasRowIdMapping()) {
+            configureSortedMerge(flushInput.rowIdMapping());
```
mgodwan: It is possible that there are multiple segments before this step [as lucene may internally decide to flush]. We should ensure that there is only segment before this is kicked in? Or would reorder take care of the same?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriterCodec.java
```diff
@@ -68,4 +88,90 @@ public void write(Directory directory, SegmentInfo info, IOContext ioContext) th
...
+     * and replaces the values with sequential 0..N when {@link #enableRowIdRewrite()}
+     * has been called. This allows the reorder merge and the row ID rewrite to happen
+     * in a single pass.
+     */
+    @Override
+    public DocValuesFormat docValuesFormat() {
+        if (rewriteRowIds == false) {
+            return delegate.docValuesFormat();
+        }
+        DocValuesFormat delegateFormat = delegate.docValuesFormat();
```
mgodwan: Can we move this into a separate class?

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdMapping.java
chaitanya588: Changed the interface to meet merge reordered use case (i.e including both newToOld and oldToNew). Also, changed the the RowMapping to capture the mapping details only for one generation.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriterCodec.java
chaitanya588: Refactored the changes.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: Current reorder logic is going to take care a cases where more than 1 segment is present within this flow. I added a test case to cover this (testSortedFlushWithMultipleInternalSegments).

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -194,6 +202,47 @@ public class DataFormatAwareEngine implements Indexer {
...
+    );
+
+    /**
+     * Node-level shared thread pool for parallel writer flush during refresh.
+     * Shared across all indices/shards on this node to avoid thread bloat.
+     */
+    private static volatile ExecutorService writerFlushExecutor;
+    private static final Object FLUSH_EXECUTOR_LOCK = new Object();
+
+    private static ExecutorService getOrCreateFlushExecutor(int maxThreads, ThreadContext threadContext) {
```
mgodwan: This should be in ThreadPool?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -756,9 +806,39 @@ public void refresh(String source) throws EngineException {
...
+                            try {
+                                childWriterFlushResult = future.get();
+                            } catch (ExecutionException e) {
+                                throw new IOException("Writer flush failed", e.getCause());
+                            } catch (InterruptedException e) {
+                                Thread.currentThread().interrupt();
+                                throw new IOException("Writer flush interrupted", e);
+                            }
+                            Writer<?> writer = childWriterFlushResult.writer;
+                            FileInfos fileInfos = childWriterFlushResult.fileInfos;
```
mgodwan: Lets use opensearch convention for this.

e.g. https://github.com/opensearch-project/OpenSearch/blob/75b6e8226bc6694bc0f496aa36069878313f3dfd/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L762-L819

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -794,16 +888,30 @@ public void refresh(String source) throws EngineException {
...
                             assert result.refreshedSegments().size() >= existingSegments.size()
                                 : "refresh must not lose existing segments; had "
                                     + existingSegments.size()
                                     + " but got "
                                     + result.refreshedSegments().size();
 
+                            final long commitStartNanos = System.nanoTime();
                             catalogSnapshotManager.commitNewSnapshot(result.refreshedSegments());
+                            final long commitElapsedMs = TimeValue.nsecToMSec(System.nanoTime() - commitStartNanos);
+                            logger.debug("refresh[{}]: catalogSnapshot commit took [{}ms]", source, commitElapsedMs);
```
mgodwan: This can be a trace log

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -829,6 +937,8 @@ public void refresh(String source) throws EngineException {
             }
             throw new RefreshFailedEngineException(shardId, ex);
         }
+        final long totalRefreshElapsedMs = TimeValue.nsecToMSec(System.nanoTime() - refreshStartNanos);
+        logger.info("refresh[{}]: total time to make documents searchable [{}ms] refreshed={}", source, totalRefreshElapsedMs, refreshed);
```
mgodwan: Lets keep this as a debug log

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdMapping.java
```diff
@@ -11,19 +11,44 @@
...
      */
     long getNewRowId(long oldId, long oldGeneration);
+
+    /**
+     * Reverse lookup: returns the old row ID corresponding to the given new row ID.
+     *
+     * @param newId the new row ID
+     * @return the original row ID, or -1 if not found
+     */
+    long getOldRowId(long newId);
```
mgodwan, line 39: These contracts don't seem to be bi-directional. Given we will maintain one rowidmapping per generation, why can we just not do `long getNewRowId(long oldId)`

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: No longer needed. We are reutilizing Flush Threadpool itself to parallelize the operations.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: Added CountDownLatch based approach for waiting for all the flushes to complete.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: Added.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: Added.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -756,9 +807,49 @@ public void refresh(String source) throws EngineException {
...
+                        try {
+                            flushLatch.await();
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
+                            throw new IOException("Writer flush interrupted", e);
+                        }
+
+                        // Propagate first failure if any
+                        if (firstFlushFailure.get() != null) {
+                            throw new IOException("Child writer flush failed", firstFlushFailure.get());
```
mgodwan: This should fail the shard

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -756,9 +807,49 @@ public void refresh(String source) throws EngineException {
...
+                                    firstFlushFailure.compareAndSet(null, e);
+                                } finally {
+                                    flushLatch.countDown();
+                                }
+                            });
+                        }
+
+                        // Wait for all flush tasks to complete
+                        try {
+                            flushLatch.await();
```
mgodwan: This seems risky as it may block refresh operations for all the other shards.

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdMapping.java
mgodwan: Lets take this in a subsequent PR

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: This is no longer needed.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: Not applicable in the latest revision.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -756,28 +787,81 @@ public void refresh(String source) throws EngineException {
...
+                            flushLatch.countDown();
                         }
-                        logger.debug("Produced {} new segments from flush", newSegments.size());
+
+                        // Wait for any writers that write threads picked up to finish
+                        try {
+                            flushLatch.await();
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
+                            throw new IOException("Refresh interrupted waiting for flush completion", e);
```
mgodwan: Fail shard here

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdMapping.java
chaitanya588: https://github.com/opensearch-project/OpenSearch/pull/21728

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1506,6 +1591,61 @@ private long currentMappingVersion() {
...
+            pendingWritersToClose.add(writerToFlush);
+
+            logger.debug(
+                "preIndex: write thread flushed writer gen={} in [{}ms]",
+                writerToFlush.generation(),
+                flushElapsedMs
+            );
+        } catch (Exception e) {
+            logger.warn("preIndex: flush failed for writer gen={}", writerToFlush.generation(), e);
+            IOUtils.closeWhileHandlingException(writerToFlush);
```
mgodwan: will this fail the shard?

sandbox/plugins/analytics-backend-lucene/src/test/java/org/opensearch/be/lucene/index/LuceneWriterDocValuesFormatTests.java
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    public void testNonRowIdFieldsAreUnchangedDuringRewrite() throws IOException {
+        Path tempDir = createTempDir();
+        try (Directory directory = new MMapDirectory(tempDir)) {
+            Codec baseCodec = Codec.getDefault();
+            LuceneWriterCodec codec = new LuceneWriterCodec(baseCodec, 1L);
+
+            IndexWriterConfig iwc = new IndexWriterConfig();
```
mgodwan: This test should be against LuceneWriter

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -114,6 +128,10 @@ public LuceneWriter(
         IndexWriterConfig iwc = analyzer != null ? new IndexWriterConfig(analyzer) : new IndexWriterConfig();
         iwc.setOpenMode(IndexWriterConfig.OpenMode.CREATE);
         iwc.setRAMBufferSizeMB(RAM_BUFFER_SIZE_MB);
```
mgodwan: Because of the race conditions which may happen while we update the merge policy during reorder, lets keep a no merge policy here, update the force merge policy later and then trigger force merge.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
```diff
@@ -160,6 +161,7 @@ public FormatStore getStore(DataFormat dataFormat) {
     public Writer<LuceneDocumentInput> createWriter(WriterConfig config) {
         assert sharedWriter.isOpen() : "Cannot create writer — shared IndexWriter is closed";
         try {
+            Sort indexSort = isIndexNeedToBeSortedInternally() ? sharedWriter.getConfig().getIndexSort() : null;
```
mgodwan: This will break for not-sorted cases always, as row ids will not be aligned in the force merge step later.

mgodwan: based on isIndexNeedToBeSortedInternally returning false

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -155,19 +173,67 @@ public WriteResult addDoc(LuceneDocumentInput input) throws IOException {
...
+        long flushStartNanos = System.nanoTime();
+        logger.info(
+            "flush: START generation={}, docCount={}, hasRowIdMapping={}",
+            writerGeneration,
+            docCount,
+            flushInput.hasRowIdMapping()
+        );
+
+        // If sort permutation is provided, configure the reorder merge policy
+        if (flushInput.hasRowIdMapping()) {
```
mgodwan, line 250: throw if it is true when index writer already has a sort field set

mgodwan: Add test for failure as well in this case

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -155,19 +173,67 @@ public WriteResult addDoc(LuceneDocumentInput input) throws IOException {
...
+        long forceMergeStartNanos = System.nanoTime();
         indexWriter.forceMerge(1, true);
+        long forceMergeDurationMs = (System.nanoTime() - forceMergeStartNanos) / 1_000_000;
+        logger.info(
+            "flush: forceMerge complete: generation={}, docCount={}, duration={}ms",
+            writerGeneration,
+            docCount,
+            forceMergeDurationMs
+        );
+
```
mgodwan, line 297: assert that doc values for row id field are in increasing order

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -189,12 +264,179 @@ public FileInfos flush() throws IOException {
...
+            segmentCommitInfo.getDelCount(),
+            segmentCommitInfo.getSoftDelCount(),
+            segmentCommitInfo.getDelGen(),
+            segmentCommitInfo.getFieldInfosGen(),
+            segmentCommitInfo.getDocValuesGen(),
+            segmentCommitInfo.getId()
+        );
+        segmentInfos.clear();
+        segmentInfos.add(newCommitInfo);
+        segmentInfos.commit(directory);
```
mgodwan, line 489: Can something like below work:

```
segmentInfos.remove(segmentInfos.info(0));
        segmentInfos.add(newCommitInfo);
        segmentInfos.commit(directory);
```

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/RustBridge.java
```diff
@@ -240,18 +257,37 @@ static ParquetFileMetadata finalizeWriter(String file) throws IOException {
...
+            if (permAddr != 0 && permLen > 0) {
+                try {
+                    long[] mappingArray = MemorySegment.ofAddress(permAddr)
+                        .reinterpret(permLen * ValueLayout.JAVA_LONG.byteSize())
+                        .toArray(ValueLayout.JAVA_LONG);
+                    rowIdMapping = new PackedRowIdMapping(mappingArray, true);
+                } finally {
+                    NativeCall.invokeVoid(FREE_ROW_ID_MAPPING, permAddr, permLen);
+                }
+            }
```
mgodwan, line 289: Can we use the direct memory here? [can be covered later]

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: Added fix to fail this shard.

sandbox/plugins/analytics-backend-lucene/src/test/java/org/opensearch/be/lucene/index/LuceneWriterDocValuesFormatTests.java
chaitanya588: Refactored some of the tests which are doing end to end kind of scenarios into a different class and modeled those tests with LuceneWriter. Keeping this test class only meant for coded readers, formatters, docvalue producers etc.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: Done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
chaitanya588: Instead of taking isIndexNeedToBeSortedInternally always as false, relying on the sort field name (ROW_ID) itself to set null for the child writer.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: Throwing an error and added a test case.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: Done.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: Yes, it is possible. Added this change.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/bridge/RustBridge.java
chaitanya588: Going to pick this up in subsequent PR related to rowID mapping simplification. https://github.com/opensearch-project/OpenSearch/pull/21728

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: Here failure is going to be caught by the overall try-catch block.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
```diff
@@ -163,6 +165,17 @@ public FormatStore getStore(DataFormat dataFormat) {
...
         try {
+            Sort indexSort = sharedWriter.getConfig().getIndexSort();
+            if ( indexSort != null
+                && indexSort.getSort().length == 1
+                && DocumentInput.ROW_ID_FIELD.equals(indexSort.getSort()[0].getField())) {
+                // When Lucene is primary, apply the customer's IndexSort so segments
+                // are natively sorted and compatible with the shared writer's IndexSort.
+                // When Lucene is secondary, no IndexSort — reorder is done via
+                // ReorderingOneMerge.reorder() in configureSortedMerge().
+                indexSort = null;
```
mgodwan: We should honor the incoming index sort in this case, right?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -765,28 +786,81 @@ public void refresh(String source) throws EngineException {
...
                                 newSegments.add(segmentBuilder.build());
                             }
                             refreshed |= hasFiles;
+                            flushLatch.countDown();
                         }
-                        logger.debug("Produced {} new segments from flush", newSegments.size());
+
+                        // Wait for any writers that write threads picked up to finish
+                        try {
+                            flushLatch.await();
```
mgodwan:
```
while (flushLatch.getCount() > 0) {
      if (isClosed.get() || failedEngine.get() != null) {
          throw new AlreadyClosedException("engine closed during refresh flush");
      }
      if (flushLatch.await(1, TimeUnit.SECONDS) == false) {
          continue; // re-check isClosed
      }
  }
```

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -185,6 +187,23 @@ public class DataFormatAwareEngine implements Indexer {
...
+    private final ConcurrentLinkedQueue<Writer<?>> pendingWritersToClose = new ConcurrentLinkedQueue<>();
+
+    // Shared queue of writers pending flush. Populated by refresh (checkoutAll),
+    // drained cooperatively by both the refresh thread and write threads (backpressure).
+    private final ConcurrentLinkedQueue<Writer<?>> flushQueue = new ConcurrentLinkedQueue<>();
+
+    // Latch for the current refresh cycle — decremented by any thread that flushes a writer.
+    // Refresh thread awaits this before proceeding with catalog commit.
+    private volatile CountDownLatch activeFlushLatch;
+
```
mgodwan, line 209: Lets take a TODO to refactor this out in a FlushManager as part of a separate PR

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
mgodwan: This is mostly to short circuit if multiple write threads are doing refreshes and we can just exit faster here if engine is already closed or some other thread has failed flush for the writer.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -155,19 +188,78 @@ public WriteResult addDoc(LuceneDocumentInput input) throws IOException {
...
 
-        // Force merge to exactly 1 segment to maintain 1:1 mapping with other formats.
+        long flushStartNanos = System.nanoTime();
+        logger.info(
+            "flush: START generation={}, docCount={}, hasRowIdMapping={}",
+            writerGeneration,
+            docCount,
+            flushInput.hasRowIdMapping()
+        );
+
```
mgodwan, line 248: Lets do an `indexwriter.flush()` to ensure all docs are out of buffer and into segments.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -176,6 +268,27 @@ public FileInfos flush() throws IOException {
...
+        // either naturally (docs added in row ID order) or via OneMerge.reorder() + row ID
+        // rewrite on the codec. This guards the 1:1 offset correspondence with the
+        // Parquet file for the same writer generation. When Lucene is primary with a
+        // user IndexSort on a non-row-id field, row IDs follow the docs and are not
+        // sequential, so the invariant is skipped. Wrapped in `assert` so the I/O cost
+        // is paid only when assertions are enabled.
+        assert segmentInfo.info.getIndexSort() != null
+            || assertRowIdsSequential(directory)
+            : "___row_id__ doc values not sequential after forceMerge for writer generation [" + writerGeneration + "]";
+
```
mgodwan: This should be `&&`?

mgodwan: 1. row id should always be sequential, despite the sort order
2. We should always have a sort order (either user supplied in primary, or row id set for secondary post reorder/set at creation)

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -113,15 +131,30 @@ public LuceneWriter(
...
+     * Tests that need to force aggressive intra-writer segment creation
+     * (to exercise multi-segment merge paths) can subclass and return a small
+     * value such as {@code 0.1}.
+     *
+     * <p>This method is invoked from the constructor, so overrides must be
+     * stateless (return a constant) — they cannot depend on any instance fields.
+     */
+    double ramBufferSizeMB() {
+        return RAM_BUFFER_SIZE_MB;
+    }
```
mgodwan, line 172: Given this is for testing, would it be more predictable to use `setMaxBufferedDocs`
Defaults to DISABLE_AUTO_FLUSH

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -155,19 +188,78 @@ public WriteResult addDoc(LuceneDocumentInput input) throws IOException {
...
+        logger.info(
+            "flush: forceMerge complete: generation={}, docCount={}, duration={}ms",
+            writerGeneration,
+            docCount,
+            forceMergeDurationMs
+        );
+
+        long commitStartNanos = System.nanoTime();
         indexWriter.commit();
+        long commitDurationMs = (System.nanoTime() - commitStartNanos) / 1_000_000;
```
mgodwan: `TimeValue.nsecToMSec`

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -189,12 +302,237 @@ public FileInfos flush() throws IOException {
...
+                    return mapping.size();
+                }
+            };
+        }
+
+        @Override
+        public void setMergeInfo(SegmentCommitInfo info) {
+            super.setMergeInfo(info);
+            if (info != null) {
+                info.info.putAttribute(WRITER_GENERATION_ATTRIBUTE, String.valueOf(0));
```
mgodwan, line 598: Why is this 0? Should this not be equal to the actual writer generation?

mgodwan: LuceneReaderManager also has an assertion which checks for writer generations to be intact which should have failed. Wondering as to why that succeeded.

```
private Collection<Long> collectReferencedGenerations(DirectoryReader reader) {
        return reader.leaves().stream().map(lrc -> {
            SegmentReader segmentReader = (SegmentReader) lrc.reader();
            SegmentCommitInfo sci = segmentReader.getSegmentInfo();
            return Long.parseLong(sci.info.getAttribute(WRITER_GENERATION_ATTRIBUTE));
        }).sorted().toList();
    }
```

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -765,28 +786,81 @@ public void refresh(String source) throws EngineException {
...
-                            Writer<?> writer = lockable.get();
-                            FileInfos fileInfos = writer.flush();
-                            Segment.Builder segmentBuilder = Segment.builder(writer.generation());
+                            flushQueue.add(lockable.get());
+                        }
+
+                        // Refresh thread drains the queue itself (it's not idle — it does work)
+                        Writer<?> writerToFlush;
+                        while ((writerToFlush = flushQueue.poll()) != null) {
+                            final long writerFlushStartNanos = System.nanoTime();
```
mgodwan, line 808: ensureOpen here to short circuit if engine has failed/closed?

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -208,13 +500,25 @@ impl NativeParquetWriter {
...
+                            );
+
+                            let (crc32, row_id_mapping) = Self::finalize_sorted_chunks(
+                                &chunk_paths, &chunk_row_ids, &chunk_crcs, &filename, index_name,
+                                &settings.sort_columns, &settings.reverse_sorts, &settings.nulls_first,
+                                writer_generation, schema.clone(),
+                            )?;
+
+                            // Clean up sorted chunk files
+                            for path in &chunk_paths {
```
mgodwan, line 520: Can we do a `let result + cleanup` kind of pattern to ensure this is always cleaned

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -275,143 +579,70 @@ impl NativeParquetWriter {
         }
     }
 
-    fn sort_and_rewrite_parquet(
-        temp_filename: &str,
+    /// Finalize pre-sorted Parquet chunks: k-way merge them into the final file.
+    /// Chunks are already sorted (done eagerly during write), so no sort needed here.
+    /// For single chunk, just rename. For empty, write empty Parquet.
+    fn finalize_sorted_chunks(
```
mgodwan, line 586: Can we verify in all tests that for chunk.len == 1 and larger cases, writer properties provided in input are always honored?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: Within the IndexWriter's forcemerge(), we already have flush(). I am adding here as well for more defensive check.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: Agree, removed this segmentInfo.info.getIndexSort() check altogether.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: This writer_generation value is just a temporary value. We are check this assertion as part of catalog snapshot's commitNewSnapshot creation, and by that time we are already updating the correct generation (as part of composite writer's refresh) flow. So, we are good here.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
chaitanya588: We can clean those files, but for some reason if finalize writer failed in the middle before even creating the merged files etc, then having those old files will be meaningful right. Here ? in the above line will make sure the cleanup happens post the success, but not during a failure scenario.

sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
chaitanya588: Added tests to verify this across all scenarios.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
chaitanya588: done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
chaitanya588: Added both extension points. In the production we are going to utilize max buffer size where as for testing we will be using buffer docs for deterministic testing.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
```diff
@@ -171,13 +175,25 @@ public Writer<LuceneDocumentInput> createWriter(WriterConfig config) {
...
+                getChildWriterSortConfiguration()
             );
         } catch (IOException e) {
             throw new RuntimeException("Failed to create LuceneWriter for generation " + config.writerGeneration(), e);
         }
     }
 
+    private Sort getChildWriterSortConfiguration() {
+        // When Lucene is secondary, then clear child writer's sort configuration and restamp
+        // it at the flush end. In all other cases, propagate same sort configuration as it is.
```
mgodwan, line 187: when lucene is secondary, and userprovided sort is null, we should honor the sort config then as well, right?

mgodwan: I think code below follows this. was confused by the wording

sandbox/plugins/analytics-backend-lucene/src/test/java/org/opensearch/be/lucene/LuceneMergerTests.java
```diff
@@ -101,19 +101,19 @@ public void testMergeWithEmptyInput() throws IOException {
...
+//
+//        LuceneMerger merger = new LuceneMerger(writer, new LuceneDataFormat(), dataPath);
+//
+//        Segment segment = Segment.builder(99L).build();
+//        MergeInput input = MergeInput.builder().addSegment(segment).newWriterGeneration(100L).build();
+//
+//        MergeResult result = merger.merge(input);
+//        assertNotNull(result);
+//        assertTrue(result.getMergedWriterFileSet().isEmpty());
+//    }
```
mgodwan: We can fully remove this?

mgodwan (approved):

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
chaitanya588: handled this case.

sandbox/plugins/analytics-backend-lucene/src/test/java/org/opensearch/be/lucene/LuceneMergerTests.java
chaitanya588: done

PR #21473, opened 2026-05-04
Speed up java agent type matching

timlueg:
### Description
From what I understand it should be possible to narrow the java agent bytebuddy matchers to JDK own classes. This avoids running the expensive `isSubTypeOf` check for all classes. Additionally using `named()` instead of the `is()` type check.
As I understand it any subclass of e.g. `FileSystemProvider` would now not be instrumented at the class level but since this subclass would normally uses JDK classes to access the file system these would still be instrumented.

Time to Node "started" best of 5:

|  | Before | After | diff. |
|---|---:|---:|---:|
| `./gradlew run` | 2013 ms | 1544 ms | -469 ms (-23.3%) |
| `./gradlew run` with plugins  | 4744 ms | 3450 ms | -1294 ms (-27.3%) |

<details>

<summary>./gradlew run -PinstalledPlugins=...</summary>

`./gradlew run -PinstalledPlugins="['opensearch-job-scheduler', 'opensearch-notifications-core', 'notifications', 'opensearch-ml-plugin', 'opensearch-knn', 'alerting', 'opensearch-anomaly-detection', 'asynchronous-search', 'opensearch-cross-cluster-replication', 'opensearch-custom-codecs', 'opensearch-flow-framework', 'geospatial', 'opensearch-index-management', 'opensearch-skills', 'neural-search', 'opensearch-observability', 'opensearch-reports-scheduler', 'opensearch-sql-plugin', 'query-insights', 'opensearch-system-templates', 'opensearch-ubi', 'opensearch-search-relevance', 'workload-management']"`

</details>

### Related Issues

### Check List


libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
```diff
@@ -77,11 +77,17 @@ public static void agentmain(String agentArguments, Instrumentation instrumentat
...
-            .or(ElementMatchers.isSubTypeOf(Socket.class));
-        final Junction<TypeDescription> pathType = ElementMatchers.isSubTypeOf(Files.class);
-        final Junction<TypeDescription> fileChannelType = ElementMatchers.isSubTypeOf(FileChannel.class);
-        final Junction<TypeDescription> fileSystemProviderType = ElementMatchers.isSubTypeOf(FileSystemProvider.class);
+        final Junction<TypeDescription> socketType = ElementMatchers.nameStartsWith("java.net.")
+            .or(ElementMatchers.nameStartsWith("sun.nio.ch."))
+            .and(ElementMatchers.isSubTypeOf(SocketChannel.class).or(ElementMatchers.isSubTypeOf(Socket.class)));
+        final Junction<TypeDescription> pathType = ElementMatchers.named(Files.class.getName());
+        final Junction<TypeDescription> fileChannelType = ElementMatchers.nameStartsWith("java.nio.channels.")
+            .or(ElementMatchers.nameStartsWith("sun.nio.ch."))
```
cwperks, line 83: Not sure if it makes sense, but I believe we could extract the package names. i.e. `FileChannel.class.getPackage().getName()`

timlueg: Good idea. I did it for `Socket` and `FileChannel`. Other packages such as `sun.nio.ch.` are JDK internal. We can't get any of the `FileChannel` implementations package name this way.

cwperks (approved):
Thank you @timlueg. The change looks good to me. @reta wdyt?

prudhvigodithi (approved):

libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
```diff
@@ -74,11 +74,17 @@ public static void agentmain(String agentArguments, Instrumentation instrumentat
     }
 
     private static AgentBuilder createAgentBuilder() throws Exception {
-        final Junction<TypeDescription> socketType = ElementMatchers.isSubTypeOf(SocketChannel.class)
-            .or(ElementMatchers.isSubTypeOf(Socket.class));
-        final Junction<TypeDescription> pathType = ElementMatchers.isSubTypeOf(Files.class);
-        final Junction<TypeDescription> fileChannelType = ElementMatchers.isSubTypeOf(FileChannel.class);
-        final Junction<TypeDescription> fileSystemProviderType = ElementMatchers.isSubTypeOf(FileSystemProvider.class);
+        final Junction<TypeDescription> socketType = ElementMatchers.nameStartsWith(Socket.class.getPackageName() + ".")
```
reta, line 77: @timlueg @cwperks I am afraid this slippery road: the JDK evolves, the assumptions invalidate fast. Where the decision which package to take into consideration are coming from? 

Please, correct me if I am missing something, but clearly we are excluding `sun.security.ssl.SSLSocketImpl` (JDK) and any user class that implement `SocketChannel` or `Socket` by package filters?

timlueg: Agree the PR suggestion is not an ideal solution. I selected the JDK packages where I found the relevant Subclasses.
`SSLSocketImpl` is excluded. But `SSLSocketImpl.connect()` delegates to `java.net.Socket.connect()` which is matched. But it is better to still include the package explicitly.

Right, user classes that extend e.g. `Socket` would no longer be matched unless they delegate like above. 

(My understanding is that these Agent matcher(s) cover the most common cases. Because I can imagine a arbitrary user class could for example open a connection without using a Socket subclass.)

reta: > (My understanding is that these Agent matcher(s) cover the most common cases. Because I can imagine a arbitrary user class could for example open a connection without using a Socket subclass.)

Thanks @timlueg , I think we should be covering any possible scenario (API set is very limited), otherwise the agent is not really serving the purpose

PR #21474, opened 2026-05-04
[Analytics Engine] PPL append: mark Union + multi-child stage runtime

ahkcs:
## Summary

Brings PPL `append` end-to-end through the analytics-engine route by
marking `LogicalUnion` and lifting the planner's "single child stage
per coordinator-reduce stage" simplification. Without this PR, every
`source=… | … | append [ … ]` query failed at the marking phase with
*Filter / Sort rule encountered unmarked child [LogicalUnion]*.

This is the architectural follow-up the codebase was already deferring
in three places — `ExchangeSinkContext.java`, `LocalStageScheduler.java`
("expected exactly one child stage"), and
`AbstractDatafusionReduceSink.java` (`INPUT_ID = "input-0"` hardcoded with
*"This will change to represent child ID"*). With multi-input support
in place, both `UNION` (this PR) and a future `JOIN` PR can share the
same per-child sink-routing infrastructure.

## Changes (4 commits)

1. **[Analytics Engine] Mark LogicalUnion for backend assignment** —
   `OpenSearchUnion` rel + `OpenSearchUnionRule` (drops empty-Values
   inputs, collapses single-input unions, wraps every input in an
   ExchangeReducer for per-branch stage isolation), `EngineCapability.UNION`
   registered on DataFusion, and the supporting wiring in `PlannerImpl`
   / `RelNodeUtils.copyToCluster`.
2. **[Analytics Engine + Backend / DataFusion] Multi-child stage
   runtime for Union** — `ExchangeSinkContext` now carries
   `List<ChildInput>`; new `MultiInputExchangeSink` SPI;
   `LocalStageScheduler` / `LocalStageExecution` route each child to its
   own native input partition via `sinkForChild(int)`;
   `DatafusionReduceSink` registers one partition per child up front and
   exposes per-child `ChildSink` wrappers; `DataFusionFragmentConvertor`
   uses per-stage input ids (`input-<childStageId>`) and rewrites
   stage-input scans inside `attachFragmentOnTop`; `replaceInput` learns
   about `Fetch` (the rel emitted for SystemLimit / `head N` above
   Union); `FragmentConversionDriver` recognises the multi-input "all
   children are ER" boundary; `PlanForker` handles multi-input
   resolution.
3. **[Analytics Engine] Filter rule fallback for derived columns** —
   when a column has no physical storage (e.g. a Union output where
   branches disagree on a positional column's storage), resolve viable
   filter backends via a new `CapabilityRegistry.filterBackendsAnyFormat`
   helper instead of throwing.
4. **[QA] AppendCommandIT** — self-contained QA integration test
   under `sandbox/qa/analytics-engine-rest`. Mirrors the SQL plugin's
   `CalcitePPLAppendCommandIT` six-test surface using the existing
   `calcs` dataset (provisioned twice — into `calcs` and `calcs_alt` —
   for the cross-index test). All eight test methods (the four
   `EmptySearchCommand*` variants are split into separate methods)
   pass against the QA module's own testcluster, no SQL plugin or
   force-routing required.

## Test results

`CalcitePPLAppendCommandIT` (in `opensearch-project/sql`,
`integ-test/src/test/java/org/opensearch/sql/calcite/remote/`) run via
`./gradlew :integ-test:analyticsCompatibilityTest --tests
"…CalcitePPLAppendCommandIT"` against a force-routed analytics-engine
cluster: **6 of 8 tests pass**.

| Test | Status |
|---|---|
| `testAppend` | ✅ |
| `testAppendDifferentIndex` | ✅ |
| `testAppendWithMergedColumn` | ✅ |
| `testAppendEmptySearchCommand` | ✅ |
| `testAppendEmptySearchWithJoin` | ✅ |
| `testAppendWithConflictTypeColumn` | ✅ |
| `testAppendSchemaMergeWithTimestampUDT` | ❌ — see below |
| `testAppendSchemaMergeWithIpUDT` | ❌ — see below |

QA-side `AppendCommandIT` (this PR, run via
`./gradlew :sandbox:qa:analytics-engine-rest:integTest --tests
"*AppendCommandIT"`): **8 of 8 pass** (the four empty-subsearch
variants are split into one test method each — same coverage as the
SQL-plugin's single multi-variant `testAppendEmptySearchCommand`).

### Deferred to follow-up PRs (pre-existing analytics-engine type/function gaps, unrelated to Append)

- **`testAppendSchemaMergeWithTimestampUDT`** — *Execution error:
  Panic: primitive array* from native DataFusion when reading the
  `birthdate` column (`TIMESTAMP`). The Java-side schema mapping
  (`ArrowSchemaFromCalcite`) needs a TIMESTAMP entry **and** the
  Rust DataFusion bridge needs to handle the timestamp Arrow encoding
  without panicking. Reproduces on **any** non-Append query referencing
  a `TIMESTAMP` column on the analytics-engine route.
- **`testAppendSchemaMergeWithIpUDT`** — *No enum constant
  `org.opensearch.analytics.spi.ScalarFunction.CIDRMATCH`*. Requires
  registering CIDRMATCH in the `ScalarFunction` enum, declaring it as a
  filter / project capability on DataFusion, and wiring a Substrait
  extension + native UDF. Also needs `FieldType.IP` in DataFusion's
  scan capability set. Reproduces on any non-Append `where cidrmatch(…)`
  query.

Both are general type/function support gaps that warrant their own
PRs per the routing doc's *"non-S0 planner gap → file a separate
issue"* guidance, and won't block the Append-side work landing.


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionMemtableReduceSink.java
```diff
@@ -81,7 +90,8 @@ protected Throwable closeUnderLock() {
                 arrayPtrs[i] = arrays.get(i).memoryAddress();
                 schemaPtrs[i] = schemas.get(i).memoryAddress();
             }
-            NativeBridge.registerMemtable(session.getPointer(), INPUT_ID, schemaIpc, arrayPtrs, schemaPtrs);
+            int singleChildStageId = childInputs.keySet().iterator().next();
```
mch2, line 123: this means memTable won't be supported for multi - child stages, lets add a TODO here or throw early, i think we already default to streaming

mch2:
thanks @ahkcs for adding the runtime for multi-child - i had a draft in flight with joins that looks very similar. have linked it above - theres a few fixes in there on top of this i've found running join ITs, but its specific to joins.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionMemtableReduceSink.java
ahkcs: Added TODO and error handling

mch2 (approved):

PR #21472, opened 2026-05-04
Enable PPL fillnull on the analytics-engine route via DataFusion COALESCE

ahkcs:
## Description

Enables PPL `fillnull` on the analytics-engine query route by wiring `COALESCE` (and `CEIL`, used inside one fillnull-with-function test) into the DataFusion backend's project-side scalar capabilities. Bundled with one analytics-engine planner fix that's a prerequisite for any nested project call to round-trip through Substrait.

This is the proof-of-concept template for the remaining ~99 Bucket-1 PPL functions: every direct DataFusion mapping is a one-line addition to `DataFusionAnalyticsBackendPlugin.STANDARD_PROJECT_OPS` (no Rust-side, convertor, or planner changes).

### How it failed before

`CalciteRelNodeVisitor.visitFillNull` already lowers `fillnull` into a clean `LogicalProject(COALESCE(field, replacement))` — verified by `CalcitePPLFillnullTest`. But running `CalciteFillNullCommandIT` against a force-routed analytics-engine cluster yielded:

```
IllegalStateException: No backend supports scalar function [COALESCE] among [datafusion]
  at OpenSearchProjectRule.annotateExpr(OpenSearchProjectRule.java:123)
```

The DataFusion backend declared `filterCapabilities`, `scanCapabilities`, and `aggregateCapabilities`, but `projectCapabilities()` defaulted to empty — so `OpenSearchProjectRule` rejected every `COALESCE` it saw. Once COALESCE was declared, the fillnull-with-function test (`fillnull with ceil(num1) in num0`) hit a second issue:

```
IllegalArgumentException: Unable to convert call ANNOTATED_PROJECT_EXPR(fp64?)
```

`OpenSearchProjectRule.annotateExpr` recurses into operands and wraps every sub-call. `OpenSearchProject.stripAnnotations` only unwrapped the top-level wrapper, leaving an inner `AnnotatedProjectExpression(CEIL)` that Substrait isthmus has no converter for.

### Changes

1. **`OpenSearchProject.stripAnnotations`** — replace single-level unwrap with a `RexShuttle` that recursively strips `AnnotatedProjectExpression` at every depth before handing the plan to the backend `FragmentConvertor`. Adds `ProjectRuleTests#testStripAnnotationsRecursivelyUnwrapsNestedExpressions` covering the nested-call shape. This is a general fix — any project containing a nested scalar call would hit it once both operators are project-capable; fillnull-with-function just happened to expose it first.
[...]

PR #21476, opened 2026-05-04
Add 3 different types of PPL scalar functions to analytics-engine - prove wiring based on DataFusion capabilities

mch2:
### Description
Ships the analytics-engine's PPL-function wiring foundation on top of PR #21436. Demonstrates one function per decision-tree category so follow-up PRs have a reference for every pattern.

This PR demonstrates 3 different ways to add functions, based on where each function's semantics live and whether the Calcite call shape needs rewriting:

 - **YAML signature only** — `date_part`. DataFusion has it natively; substrait core doesn't declare a signature. One YAML entry.

  - **Adapter rewrite (cat 3)** — `YEAR(ts)` → `date_part('year', ts)` via `YearAdapter`. `UNIX_TIMESTAMP(ts)` → `to_unixtime(ts)` via `UnixTimestampAdapter`. Both use the reusable `AbstractNameMappingAdapter` — a concrete subclass per function. Follow-up date-part functions (MonthAdapter, DayAdapter, ...) use the same shape.

  - **Rust UDF (cat 4)** — `convert_tz`. DataFusion has no native timezone-shift; the UDF implements it via chrono-tz. A `ConvertTzAdapter` bridges PPL's bespoke operator to a locally-declared `SqlFunction` that isthmus can resolve to the UDF.

  ### Supporting changes

  - `AbstractNameMappingAdapter` — reusable base for rename + prepend/append literal-arg injection. Preserves the original call's return type so the enclosing `Project`'s cached `rowType` stays consistent with the adapted expression.
  - `ScalarFunction` enum additions: `YEAR`, `CONVERT_TZ`, `UNIX_TIMESTAMP`. `fromSqlFunction` now returns null on unknown rather than throwing (matches its javadoc contract).
  - `DataFusionAnalyticsBackendPlugin` declares `projectCapabilities()` for the new functions and wires the three adapters through `scalarFunctionAdapters()`.
  - `DataFusionFragmentConvertor` wires a custom `ScalarFunctionConverter` through the 6-arg `ConverterProvider` constructor so `ADDITIONAL_SCALAR_SIGS` actually takes effect.
  - Ported `OpenSearchProject.stripAnnotations` from #21472: recursive RexShuttle unwrap so nested `AnnotatedProjectExpression` wrappers are stripped at every depth before fragment conversion. Required for any call with literal operands.

### Check List


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -119,10 +122,17 @@ public static ScalarFunction fromSqlKind(SqlKind kind) {
...
 
-    /** Maps a Calcite SqlFunction to a ScalarFunction by name, or throws if not recognized. */
+    /** Maps a Calcite SqlFunction to a ScalarFunction by name, or null if not recognized. */
     public static ScalarFunction fromSqlFunction(SqlFunction function) {
         // TODO: Add an explicit functionName field per enum constant instead of relying on
         // valueOf(toUpperCase). This couples enum constant naming to SQL function naming convention.
-        return ScalarFunction.valueOf(function.getName().toUpperCase(Locale.ROOT));
+        try {
+            return ScalarFunction.valueOf(function.getName().toUpperCase(Locale.ROOT));
+        } catch (IllegalArgumentException e) {
```
expani: Then the caller should handle the exception and fallback ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/AbstractNameMappingAdapter.java
```diff
@@ -0,0 +1,98 @@
...
+ *       }
+ *   }
+ * </pre>
+ * rewrites {@code YEAR(ts)} to {@code date_part('year', ts)}. Paired with the
+ * {@code date_part} signature in {@code opensearch_scalar.yaml} so isthmus
+ * resolves it against DataFusion's native date_part.
+ *
+ * @opensearch.internal
+ */
+abstract class AbstractNameMappingAdapter implements ScalarFunctionAdapter {
```
expani: This can be in analytics-framework.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConvertTzAdapter.java
```diff
@@ -0,0 +1,64 @@
...
+ * bespoke {@code SqlUserDefinedFunction} built in {@code sql/core}'s
+ * {@code PPLBuiltinOperators} — isthmus's {@code ScalarFunctionConverter} binds
+ * Sigs by operator reference, so the Sig's target must be an operator instance
+ * we can import. Adding a {@code sql/core} dependency on this module is not
+ * desirable; instead we declare a minimal {@link SqlFunction} here whose sole
+ * purpose is to be the referent of our Sig.
+ *
+ * @opensearch.internal
+ */
+class ConvertTzAdapter extends AbstractNameMappingAdapter {
```
expani: Should we create one file and add all the small classes inside it ?

mch2: yeah i think we can consolidate by func type

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/convert_tz.rs
```diff
@@ -0,0 +1,434 @@
...
+
+fn parse_tz(s: &str) -> Option<TzSpec> {
+    if let Some(off) = parse_offset_seconds(s) {
+        return Some(TzSpec::Offset(off));
+    }
+    s.parse::<Tz>().ok().map(TzSpec::Iana)
+}
+
+/// Parse `±HH:MM` → seconds east of UTC; None if not an offset literal.
+fn parse_offset_seconds(s: &str) -> Option<i32> {
```
expani, line 240: Given that a lot of these functions are ALSO manipulating the literal value, is there any opportunity of reducing the code and moving some of it to a BackendPlanAdapter ?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
mch2: i think we need to either throw a checked exception here or null and flag this as nullable - theres a lot of callers here relying on a rte that will bubble through uncaught.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/convert_tz.rs
mch2: we can probably move validation / conversion to java side, but we won't be able to replace this parsing.

sandeshkr419 (approved):

mch2:
GC stalled , but passed on prior run - no changes outside of sandbox and sandbox checked passes...merging this.

PR #21477, opened 2026-05-05
Fix resize request to return 4xx for missing write block

### LINKED ISSUE #19026: [BUG] Resize request should throw 4xx as part of validation failure for write block
### Describe the bug

Currently as part of the resize request validation while creating a target index using MetadataCreateIndexService, it throws IllegalStateException that gets translated to 5xx. Here it should instead throw IllegalArgumentException which will be translated to 4xx. 5xx is usually treated as server side failure/error whereas 4xx will be wrong user input which fits well in this case.

Ref: https://github.com/opensearch-project/OpenSearch/blame/1a5331727b8037b4bcaf3e7c2a46dfb467259d13/server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java#L1655

#### Response: 
```
HTTP/2 500 Internal Server Error
Date: Mon, 07 Jul 2025 15:48:17 GMT
Content-Type: application/json; charset=UTF-8
Content-Length: 315
Access-Control-Allow-Origin: *

{"error":{"root_cause":[{"type":"illegal_state_exception","reason":"index test-01 must block write operations to resize index. use \"index.blocks.write=true\""}],"type":"illegal_state_exception","reason":"index test-01 must block write operations to resize index. use \"index.blocks.write=true\""},"status":500}
```

### Related component

_No response_

### To Reproduce

* Submit request to shrink an existing index (test-01) without write block on it.

```
curl localhost:9200/test-01/_shrink/test-shrink-01
```

### Expected behavior

OpenSearch should throw 4xx error instead of 5xx

### Additional Details

**Plugins**

**Screenshots**

**Additional context**

serelli:
## Summary
- Changes `IllegalStateException` to `IllegalArgumentException` in `MetadataCreateIndexService.validateResize()` so that resize requests (shrink/split/clone) without a write block return 400 instead of 500
- Updates corresponding test assertions in `MetadataCreateIndexServiceTests`

Fixes #19026

## Test plan
- [x] `MetadataCreateIndexServiceTests.testValidateShrinkIndex` passes
- [x] `MetadataCreateIndexServiceTests.testValidateSplitIndex` passes
- [x] `MetadataCreateIndexServiceTests.testValidateCloneIndex` passes


cwperks:
Is there a way to write a test for this to validate the status code?

serelli:
> Is there a way to write a test for this to validate the status code?

@cwperks 
  Good suggestion — added REST YAML tests in the latest commit.                                                                                              
                                                                                                                                                             
  The YAML test framework supports catch: bad_request which asserts the HTTP response code is exactly 400. Added a "Returns 400 when write block is missing" 
  scenario to each of the three operation test files (indices.shrink/10_basic.yml, indices.split/10_basic.yml, indices.clone/10_basic.yml):                  
                                                                                                                                                             
  ---                                                                                                                                                        
  "Returns 400 when write block is missing on shrink":                                                                                                     
    # Attempt to shrink without setting index.blocks.write=true. The server should                                                                           
    # reject this with a 400 Bad Request (not a 500 Internal Server Error).                                                                                  
    - do:                                                                                                                                                    
        catch: bad_request                                                                                                                                   
        indices.shrink:                                                                                                                                      
          index: "source"                                                                                                                                    
          target: "target_no_block"
          body:                                                                                                                                              
            settings:                                                                                                                                        
              index.number_of_replicas: 0
              index.number_of_shards: 1                                                                                                                      
    - match: { status: 400 }                                                                                                                               
                                                                                                                                                             
  This directly validates the HTTP status code, not just the exception type, so it would have caught the original 500 regression.

PR #21479, opened 2026-05-05
Introduce Instructions for the Backend to use during execution

expani:
## Backend Execution Metadata Framework

Replaces the opaque data-node execution flow with a chained instruction-based system. The planner now produces typed, ordered instructions per stage that tell the backend how to configure its execution environment before running the query.

### Why:

The old flow bundled SessionContext creation, table registration, and query execution into one indivisible call. This made it hard to inject work between steps — like registering UDFs for filter delegation or configuring optimizer settings for partial aggregates without making backends COMPLEX. 

### What it does:

The planner analyzes the query shape and produces instructions like "set up a shard scan", "configure filter delegation", or "prepare for final aggregation". These travel with the plan to the execution site. At execution time, each instruction's handler runs in order, composably building up the native execution environment. Only after all handlers have run does the query execute against the fully-configured environment.

For the Rust/DataFusion side, the previous execute_query is decomposed into create_session_context → register_table_provider → execute_with_context, each callable independently via FFM. This lets Java instruction handlers configure the SessionContext incrementally.

What this enables next: Filter delegation (Core bridges Lucene and DataFusion without either knowing about the other), partial aggregate optimizer configuration, and any future execution concern — all as additional instruction handlers without touching existing code.


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSearchExecEngine.java
```diff
@@ -28,32 +26,25 @@
...
+    public void prepare(ShardScanExecutionContext requestContext) {
         byte[] substraitBytes = requestContext.getFragmentBytes();
         long contextId = datafusionContext.task() != null ? datafusionContext.task().getId() : 0L;
         datafusionContext.setDatafusionQuery(new DatafusionQuery(requestContext.getTableName(), substraitBytes, contextId));
     }
 
     @Override
-    public EngineResultStream execute(ExecutionContext requestContext) throws IOException {
+    public EngineResultStream execute(ShardScanExecutionContext requestContext) throws IOException {
         DatafusionSearcher searcher = datafusionContext.getSearcher();
```
bharath-techie, line 49: Will `ShardScanExecutionContext requestContext` be used in this block ?

expani: Diff is because of rename. 

From the looks of it, was never getting used. Maybe we can cleanup after delegation has been implemented to ensure it's not required ?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FinalAggregateInstructionNode.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Instruction node for final aggregate in coordinator reduce — ExchangeSink path,
+ * remove partial agg, preserve final-only for the driving backend's reduce execution.
+ *
+ * <p>TODO: add backend-specific config fields as final aggregate implementation is built out.
+ *
+ * @opensearch.internal
+ */
+public class FinalAggregateInstructionNode implements InstructionNode {
```
expani, line 24: Intentionally empty. 

Will be part of a fast followup.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ShardScanInstructionHandler.java
```diff
@@ -0,0 +1,60 @@
...
+        }
+        if (dfReader == null) {
+            throw new IllegalStateException("No DatafusionReader available in the acquired reader");
+        }
+
+        long readerPtr = dfReader.getReaderHandle().getPointer();
+        long runtimePtr = dataFusionService.getNativeRuntime().get();
+        long contextId = context.getTask() != null ? context.getTask().getId() : 0L;
+
+        SessionContextHandle sessionCtxHandle = NativeBridge.createSessionContext(readerPtr, runtimePtr, contextId);
```
bharath-techie: Can't we merge this , do we need two different calls via FFM, one for session context and one for table provider ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/LocalStageScheduler.java
```diff
@@ -44,6 +48,17 @@ public StageExecution createExecution(Stage stage, ExchangeSink sink, QueryConte
             deriveInputSchema(stage),
             sink
         );
+
+        // Apply instruction handlers for the reduce stage
+        FragmentInstructionHandlerFactory factory = stage.getInstructionHandlerFactory();
+        if (factory != null) {
+            BackendExecutionContext backendContext = null;
+            for (InstructionNode node : stage.getPlanAlternatives().getFirst().instructions()) {
```
bharath-techie, line 64: Are we assuming , plan alternatives size will be always 1 ? do we need assertion here

expani: This is only for the root stage so it should always have one alternative.

There is a TODO to never keep PlanAlternatives in the RootStage but we can refactor that later.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ShardScanInstructionHandler.java
expani: Merged

sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
```diff
@@ -29,9 +29,14 @@ use substrait::proto::Plan;
 use crate::cross_rt_stream::CrossRtStream;
 use crate::executor::DedicatedExecutor;
 use crate::api::DataFusionRuntime;
+use crate::session_context::SessionContextHandle;
 
 /// Execute a vanilla parquet query: substrait plan → DataFusion → CrossRtStream.
 /// File access goes through DataFusion's registered object store.
+///
+/// Deprecated: Production now uses the decomposed `create_session_context` +
```
bharath-techie, line 37: Should we remove this since we're in same release cycle ?

sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
```diff
@@ -0,0 +1,138 @@
...
+        runtime_env_builder = runtime_env_builder.with_memory_pool(pool);
+    }
+
+    let runtime_env = runtime_env_builder.build().map_err(|e| {
+        error!("create_session_context: failed to build runtime env: {}", e);
+        e
+    })?;
+
+    let mut config = SessionConfig::new();
+    config.options_mut().execution.parquet.pushdown_filters = false;
```
bharath-techie: Can we try using the DatafusionQueryConfig ? its wired with defaults already - we will have a different PR to wire with cluster settings.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/SessionContextHandle.java
```diff
@@ -0,0 +1,34 @@
...
+ * registry so that {@link NativeHandle#validatePointer} passes for FFM calls.
+ */
+public class SessionContextHandle extends NativeHandle {
+
+    public SessionContextHandle(long ptr) {
+        super(ptr);
+    }
+
+    @Override
+    protected void doClose() {
```
bharath-techie, line 29: Lets handle this similar to 
```
 @Override
    protected void doClose() {
        NativeBridge.closeGlobalRuntime(ptr);
    }
```

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSearcher.java
```diff
@@ -45,6 +46,41 @@ public DatafusionSearcher(ReaderHandle readerHandle) {
 
     @Override
     public void search(DatafusionContext context) throws IOException {
+        SessionContextHandle sessionCtx = context.getSessionContextHandle();
+        if (sessionCtx != null) {
+            searchWithSessionContext(context, sessionCtx);
+        } else {
+            searchVanilla(context);
```
bharath-techie, line 53: if this cleanup is followup lets add todo with deprecated warning.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FilterTreeShapeDeriver.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param filter              the OpenSearchFilter with annotations intact
+     * @param drivingBackendId    the filter operator's resolved backend
+     * @return the tree shape, or {@code null} if no delegated annotations exist
+     */
+    static FilterTreeShape derive(OpenSearchFilter filter, String drivingBackendId) {
+        Result result = walk(filter.getCondition(), drivingBackendId, false);
+        if (!result.hasDelegated) {
+            return null;
+        }
+        return result.hasMixed ? FilterTreeShape.MIXED_BOOLEAN : FilterTreeShape.SINGLE_AND;
```
bharath-techie: Can result represent an ENUM directly somehow , we are seeing cases where we might need to avoid the whole RG by RG stream and choose a different fast path such as `GetByID` call. 

So there will be more filterTreeShapes other than these two.

Curious if we can represent `NotDelegated` also as a filter shape [`Plain`] instead of returning null ?

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/FilterTreeShapeDeriverTests.java
```diff
@@ -0,0 +1,115 @@
...
+    public void testNoDelegation() {
+        // Single native predicate — no delegation
+        RexNode nativePred = annotated(DRIVING);
+        OpenSearchFilter filter = buildFilter(nativePred);
+
+        FilterTreeShape shape = FilterTreeShapeDeriver.derive(filter, DRIVING);
+        assertNull("No delegation should return null", shape);
+    }
+
+    public void testSingleDelegatedPredicate() {
```
bharath-techie, line 39: The cases should be extended further and also if possible we should have mix of randomized generators + hardcoded tests to assert filter tree shapes.

Also lucene queries rewrites/ boolean query rewrites etc am hoping will be covered in delegates PR or follow up PR.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/LocalStageScheduler.java
```diff
@@ -44,6 +48,17 @@ public StageExecution createExecution(Stage stage, ExchangeSink sink, QueryConte
...
+                FragmentInstructionHandler handler = factory.createHandler(node);
+                backendContext = handler.apply(node, context, backendContext);
```
mch2, line 66: where is this used on coord reduce?

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/FilterTreeShapeDeriverTests.java
expani: Randomized testing is a broader issue with entire E2E. 

Have some idea of how to structure it better, will start an issue.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FilterTreeShapeDeriver.java
expani: The internal walking requires more granular data than the external Enum so we need those 3 flags unfortunately. Will revisit this later if there is a chance to simplify.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSearcher.java
expani: Added TODO for now.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/SessionContextHandle.java
expani: Added a TODO for now given this was getting complex with `QueryTrackingContext`

sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
expani: Done

sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
expani: Added ToDo as cleanup required refactoring a lot of query benchmarks.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FilterTreeShapeDeriver.java
expani: Have added the PLAIN shape to enum instead of using `null`

bharath-techie (approved):
LGTM

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -108,14 +112,26 @@ public FragmentExecutionResponse executeFragment(FragmentExecutionRequest reques
...
             ctx.setFragmentBytes(selectedPlan.getFragmentBytes());
 
             AnalyticsSearchBackendPlugin backend = backends.get(selectedPlan.getBackendId());
 
+            // Apply instruction handlers in order — each builds upon the previous handler's backend context
+            FragmentInstructionHandlerFactory factory = backend.getInstructionHandlerFactory();
+            BackendExecutionContext backendContext = null;
+            for (InstructionNode node : selectedPlan.getInstructions()) {
+                FragmentInstructionHandler handler = factory.createHandler(node);
+                backendContext = handler.apply(node, ctx, backendContext);
```
mch2: handlers here create contexts that are owning sessionContext ptrs, if one throws this will leave open contexts in df?  think we need a try/catch here that walks and closes whatever was created.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/BackendExecutionState.java
```diff
@@ -0,0 +1,22 @@
...
+ * successive instruction handler calls. The first handler in the chain receives
+ * {@code null} and bootstraps the state; subsequent handlers receive and build
+ * upon the previous handler's output.
+ *
+ * <p>Each backend defines its own concrete implementation (e.g.,
+ * {@code DataFusionSessionState} holding a native SessionContext handle).
+ *
+ * @opensearch.internal
+ */
+public interface BackendExecutionState {}
```
mch2, line 22: this is identical to BackendExecutionContext with identical javadoc.

expani: Good catch. It's unused will remove it

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
expani: Yes closing in case of an error is a leak right now. 

I tried a bunch of approaches but it was taking a significant time. 

Have added a TODO and synced offline with @bharath-techie to pick this up as a fast followup.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/LocalStageScheduler.java
expani: The removal of partial mode from aggregate so only Final is performed will be done under the implementation of this handler. 

Currently, added an empty implementation for @sandeshkr419 to follow up in his PR.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/BackendExecutionState.java
expani: Removed

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/InstructionType.java
```diff
@@ -0,0 +1,48 @@
...
+import java.io.IOException;
+
+/**
+ * Types of instructions that the planner can produce for backend execution.
+ * Each type corresponds to a specific execution concern that the backend
+ * must handle during the prepare phase on the data node.
+ *
+ * @opensearch.internal
+ */
+public enum InstructionType {
```
sandeshkr419, line 22: nitpick on making the types imperative:
```
INITIALIZE_SHARD_SCAN,
DELEGATE_FILTER,
CONFIGURE_PARTIAL_AGGREGATE,
CONFIGURE_FINAL_AGGREGATE;
```

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FilterTreeShape.java
```diff
@@ -0,0 +1,32 @@
...
+package org.opensearch.analytics.spi;
+
+/**
+ * Backend-agnostic description of the boolean tree shape when filter delegation is active.
+ * Provided by the planner so backends can choose their execution strategy without
+ * re-inspecting the Substrait plan.
+ *
+ * @opensearch.internal
+ */
+public enum FilterTreeShape {
```
sandeshkr419, line 18: better naming suggestion:

```
public enum FilterTreeShape {
       /** No delegation — all predicates handled natively by the driving backend. */
       NO_DELEGATION,
       /** Delegated and native predicates are combined only by AND — bitsets can be AND-ed independently. */
       CONJUNCTIVE,
       /** Delegated and native predicates are interleaved under OR/NOT — requires a tree evaluator. */
       INTERLEAVED
   }
```

sandeshkr419 (approved):
Thanks for working on this. The abstraction LGTM with a few nit-picks: happy to pick them up myself while implementing in my follow-up PR https://github.com/opensearch-project/OpenSearch/pull/21457 as well if they make more sense.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FilterTreeShape.java
expani: Changed with slight modifications

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/InstructionType.java
expani: Changed to add a Setup prefix again all names.

mch2 (approved):

PR #21480, opened 2026-05-05
Add coordinator side hash joins to analytics engine

mch2:
## Summary

  Brings PPL join coverage (inner, left outer, cross, lookup, lookup+rename) online through the analytics-engine route, plus fixes uncovered during the merge from `main`.

  - **`JoinCommandIT`** — 6 ITs covering inner, left outer, cross, lookup, lookup+rename (all green). appendcol skipped with `@AwaitsFix` — depends on window-function capability track (separate PR).
  - **Cross join marking** — `OpenSearchJoinRule.matches` relaxed from `!leftKeys.isEmpty()` to `info.isEqui()`.
  - **`CAST` project capability** — added `Scalar(CAST, ...)` to `DataFusionAnalyticsBackendPlugin.projectCapabilities()`. LEFT OUTER JOIN injects CAST for nullable coercion; without this the project rule rejected the expression.
  - **Substrait names-list emission fix** (general correctness — not just joins) — `DataFusionFragmentConvertor.rewire(Plan, Rel)` previously reused `innerRoot.getNames()` for the wrapper's output names. When the wrapper's output schema differs from the inner plan's (e.g. `Aggregate` over `Join` shrinks 4 columns → 1), this emitted a mismatched `Plan.Root.names` list and DataFusion rejected with `"Names list must match exactly to nested schema, but found N uses for M names"`. Fixed
  by taking the wrapper's own names.
  - **`RowProducingSink.close()` no longer drops buffered batches** — `close()` used to release and clear the buffer. `ShardFragmentStageExecution.onShardTerminated` invokes `close()` before the SUCCEEDED transition; the PlanWalker's completion listener reads results during that transition → saw 0 rows. Terminal root sink is now a no-op on close; the downstream consumer (`DefaultPlanExecutor#batchesToRows`) already owns per-batch cleanup. Error path keeps leak safety via new
  `releaseUnread()`.

### Related Issues

### Check List

### DIFF
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/EngineCapability.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/EngineCapability.java
@@ -21,5 +21,6 @@ package org.opensearch.analytics.spi;
  */
 public enum EngineCapability {
     SORT,
-    UNION
+    UNION,
+    JOIN
 }
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
@@ -266,8 +266,9 @@ public class FragmentConversionDriver {
         IntraOperatorDelegationBytes delegationBytes
     ) {
         if (node instanceof OpenSearchExchangeReducer) {
-            // Strip ExchangeReducer — StageInputScan below it is the schema source
-            // This should never be reached directly; handled by the parent (final agg)
+            // Root ExchangeReducer — the coordinator fragment is just the reducer over
+            // the data-node child stage's output. Convert the inner subtree (with
+            // StageInputScan as the schema-carrying leaf) as the final-agg fragment.
             return convertor.convertFinalAggFragment(strip(node.getInputs().getFirst(), delegationBytes));
         }
         if (node instanceof OpenSearchRelNode openSearchNode) {
@@ -275,25 +276,18 @@ public class FragmentConversionDriver {
             Function<OperatorAnnotation, RexNode> resolver = delegationBytes.resolverFor(openSearchNode, node.getCluster().getRexBuilder());
             RelNode strippedNode = openSearchNode.stripAnnotations(strippedInputs, resolver);
 
-            if (!finalAggConverted) {
-                // First OpenSearchRelNode whose ALL inputs are ExchangeReducers is treated as the
-                // boundary between the coordinator-side fragment and the data-node child stages.
-                // For single-input shapes (Sort/Project/Aggregate over a partial agg) this is the
-                // final-aggregate operator; for multi-input shapes (Union) every branch is itself
-                // an ER → StageInputScan, and the entire Union+ER subtree is converted as one
-                // fragment so all branches end up in the same Substrait plan reading from their
-                // respective input partitions.
-                boolean allChildrenAreExchangeReducer = !node.getInputs().isEmpty()
-                    && node.getInputs().stream().allMatch(input -> input instanceof OpenSearchExchangeReducer);
-                if (allChildrenAreExchangeReducer) {
-                    List<RelNode> finalAggInputs = new ArrayList<>(node.getInputs().size());
-                    for (RelNode input : node.getInputs()) {
-                        // Skip the ER, keep StageInputScan below it as the leaf for schema inference.
-                        finalAggInputs.add(strip(input.getInputs().getFirst(), delegationBytes));
-                    }
-                    RelNode finalAggFragment = openSearchNode.stripAnnotations(finalAggInputs, resolver);
-                    return convertor.convertFinalAggFragment(finalAggFragment);
-                }
+            if (!finalAggConverted && isFinalAggBoundary(node)) {
+                // Boundary between coordinator-side fragment and data-node child stages. For
+                // single-input shapes (Sort/Project/Aggregate over a partial agg) this is the
+                // final-aggregate operator; for multi-input shapes (Union, coord-side Join,
+                // future set-ops) every branch is an ER → StageInputScan and the entire
+                // operator+branches subtree converts as one fragment so all branches end up in
+                // the same Substrait plan reading from their respective input partitions.
+                //
+                // strippedInputs already resolves each ER to the StageInputScan below it via
+                // the strip() helper, so strippedNode is the final-agg fragment ready for
+                // conversion — no further per-branch rewrite is needed.
+                return convertor.convertFinalAggFragment(strippedNode);
             }
 
             // Operator above the final-fragment boundary — convert child first, then attach.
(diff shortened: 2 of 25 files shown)


sandeshkr419 (commented):
Have we considered making the join's distribution strategy a pluggable interface (e.g., JoinStrategy that provides per-side ExchangeInfo) rather than hardcoding singleton() in cutSingleton? Would make adding shuffle/broadcast later additive instead of requiring changes across DAGBuilder and stage wiring.

Something that would look like:

```
 public interface JoinStrategy {

       /** How to distribute the left input to the join executor(s). */
       ExchangeInfo leftExchange(JoinContext ctx);

       /** How to distribute the right input to the join executor(s). */
       ExchangeInfo rightExchange(JoinContext ctx);

       /** Where the join itself executes. */
       StageExecutionType executionType();
   }
```

Implementations:

```
   class CoordinatorHashJoin implements JoinStrategy {
       ExchangeInfo leftExchange(ctx)  → singleton()
       ExchangeInfo rightExchange(ctx) → singleton()
       StageExecutionType executionType() → COORDINATOR_REDUCE
   }

   class ShuffleHashJoin implements JoinStrategy {
       ExchangeInfo leftExchange(ctx)  → hash(ctx.leftKeys(), ctx.numPartitions())
       ExchangeInfo rightExchange(ctx) → hash(ctx.rightKeys(), ctx.numPartitions())
       StageExecutionType executionType() → PARTITION_LOCAL  // new type
   }

   class BroadcastJoin implements JoinStrategy {
       ExchangeInfo leftExchange(ctx)  → random()  // stays on shards
       ExchangeInfo rightExchange(ctx) → broadcast()
       StageExecutionType executionType() → SHARD_LOCAL  // join runs on each shard
   }
```

   How this avoids refactoring later:

   - DAGBuilder doesn't need to know about join types — it asks the strategy for exchange info and stage type
   - OpenSearchJoinRule picks the strategy (based on stats, hints, or defaults; could be user later on as well) and attaches it to OpenSearchJoin
   - Adding a new strategy is purely additive — implement the interface, register it
   - Each side gets its own ExchangeInfo independently (critical — broadcast join has asymmetric distribution)

We don't have to expose a user-facing execution hint right now, we can decide that later if this needs a user-facing hint later on, but certainly want to avoid a major refactor later.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/EngineCapability.java
```diff
@@ -21,5 +21,6 @@
  */
 public enum EngineCapability {
     SORT,
-    UNION
+    UNION,
+    JOIN
```
expani, line 25: This can be a DedicatedCapability like Aggregate, Filter and not a generic one like Sort/Union. The details that we can capture from the backend as a capability are JoinKind support ( LEFT/RIGHT/OUTER/CROSS) 
Some like
```
   public record JoinCapability(
       Set<JoinKind> kinds,
       Set<String> formats
   ) {

       public enum JoinKind {
           INNER, LEFT, RIGHT, FULL, SEMI, ANTI, CROSS
       }
   }

```

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchJoinRule.java
```diff
@@ -0,0 +1,157 @@
...
+        RelTraitSet singletonTraits = join.getTraitSet()
+            .replace(OpenSearchConvention.INSTANCE)
+            .replace(context.getDistributionTraitDef().singleton());
+
+        // Wrap each input in an OpenSearchExchangeReducer to gather it to the coord.
+        // The DAG builder cuts at every reducer, producing one child stage per join input.
+        // The join itself is SINGLETON since both gathered inputs are SINGLETON — declaring
+        // this on the join's trait set lets any operator above it (e.g. Project) inherit
+        // SINGLETON without Volcano inserting a redundant top-level reducer.
+        RelNode left = wrapInExchange(join.getLeft(), singletonTraits, viableBackends);
```
expani, line 113: This needs to happen via Volcano CBO similar to AggregateSplitting to detect trait mismatch and insert exchanges only for the case of multi shard indices.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
```diff
@@ -303,6 +297,19 @@ private static byte[] convertReduceNode(
         throw new IllegalStateException("Unexpected reduce stage node: " + node.getClass().getSimpleName());
     }
 
+    /**
+     * True when {@code node} sits on the boundary between the coordinator-side fragment and
+     * the data-node child stages — either a multi-input shape with every branch in its own
+     * ExchangeReducer (Union, coord-side Join) or a single-input operator directly above
+     * one ExchangeReducer (final agg over a partial-agg child stage).
+     */
+    private static boolean isFinalAggBoundary(RelNode node) {
```
expani, line 306: We know the number of childs a stage has, we can look at reusing that to detect whether there are multiple inputs or not. 
`stage.getChildStages().size()`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanForker.java
```diff
@@ -73,39 +73,56 @@ private static List<Resolved> resolve(RelNode node, CapabilityRegistry registry)
...
-                }
         }
-        return resolveOperator(node, resolvedChildren, agreedBackend);
+        crossProduct(childAlternativeSets, 0, new ArrayList<>(), null, (chosenBackend, combo) -> {
+            results.addAll(resolveOperator(node, combo, chosenBackend));
+        });
+        return results;
+    }
+
+    private interface ComboHandler {
```
expani, line 97: nit: Can be a `BiConsumer`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -163,26 +162,27 @@ private List<String> resolveViableBackends(
...
-                // non-derived columns above.
-                // TODO: for FULL_TEXT operators, extract required params from RexCall
-                fieldViable = new HashSet<>(registry.filterBackendsAnyFormat(function, fieldType));
-            } else {
-                // Format-aware: backends that can access this field's storage (doc values + index).
-                // A backend is viable only if it has the field in its own storage formats — ensuring
-                // delegation targets are also field-storage-aware (e.g. Lucene is viable for a keyword
-                // field only when the field has indexFormats=[lucene] set in the mapping).
-                fieldViable = new HashSet<>(registry.filterBackendsForField(function, storageInfo));
+                // Derived columns (post-Union, post-Project, HAVING on aggregate outputs) have no
```
expani, line 167: Candidate backend for a derived column should be the same as the underlying operators childViableBackends. 

Then they can be further trimmed down based on the functions present in post aggregate/join filter.

Delegation only comes into the picture if the function on post filter is supported by a different backend. 

Can you elaborate more on the type of query this was causing to fail ?

PR #21497, opened 2026-05-05
Replace mimalloc with jemalloc as global allocator for native sandbox plugins

himshikhagupta:
### Description
Switches the native sandbox plugin layer from mimalloc to jemalloc (tikv-jemallocator 0.6) as the global allocator. jemalloc provides better memory observability, and avoids the TLS/dlopen compatibility issues encountered with mimalloc in the plugin loading context.

### Related Issues

### Check List


sandbox/libs/dataformat-native/rust/common/src/metrics.rs
```diff
@@ -0,0 +1,92 @@
...
+    let m = mib();
+    if m.epoch.advance().is_err() {
+        return -1;
+    }
+    m.resident.read().map(|v| v as i64).unwrap_or(-1)
+}
+
+/// FFI: Returns current jemalloc allocated bytes, or -1 on error.
+#[no_mangle]
+pub extern "C" fn native_jemalloc_allocated_bytes() -> i64 {
```
alchemist51: Does it tell the error type?

sandbox/libs/dataformat-native/rust/common/src/metrics.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+/// Returns current jemalloc allocated bytes, or -1 on error.
+/// This reflects the total bytes actively in use by application code (live malloc'd objects).
+/// Useful for application-level memory accounting and DataFusion memory pool budgeting.
+///
+/// TODO: integrate with node/stats
+pub fn allocated_bytes() -> i64 {
+    let m = mib();
+    if m.epoch.advance().is_err() {
```
alchemist51: epoch.advance is an expensive call, in what frequency we will be calling it? we are doing it from the resident_bytes also. Should we think to centralize this and make sure <1s can use the same mib() values? wdyt? @himshikhagupta

sandbox/libs/dataformat-native/rust/lib/src/lib.rs
```diff
@@ -9,20 +9,19 @@
...
 //   3. All #[no_mangle] extern "C" functions from the plugin crates are
 //      automatically available for dlsym/SymbolLookup
 // ═══════════════════════════════════════════════════════════════════════════════
 
-//TODO: AwaitsFix: Fix mimalloc lifecycle issue
-// #[global_allocator]
-// static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
+/// jemalloc tuning: reduce dirty page decay (30s), muzzy decay (30s), and tcache max (64KB).
+#[export_name = "malloc_conf"]
+pub static MALLOC_CONF: &[u8] = b"dirty_decay_ms:30000,muzzy_decay_ms:30000,lg_tcache_max:16\0";
```
alchemist51, line 27: Is there a way to set it up at runtime? Or this will need a code change always?

sandbox/libs/dataformat-native/src/test/java/org/opensearch/nativebridge/spi/NativeMemoryMetricsTests.java
```diff
@@ -0,0 +1,45 @@
...
+import java.lang.foreign.SymbolLookup;
+import java.lang.foreign.ValueLayout;
+import java.lang.invoke.MethodHandle;
+
+/**
+ * Tests that the native jemalloc metrics functions are available and return valid data.
+ */
+public class NativeMemoryMetricsTests extends OpenSearchTestCase {
+
+    public void testAllocatedBytesIsPositive() throws Throwable {
```
alchemist51, line 24: Can we add tests for -1 case here?

sandbox/plugins/analytics-backend-datafusion/rust/Cargo.toml
```diff
@@ -36,12 +36,9 @@ once_cell = { workspace = true }
 dashmap = { workspace = true }
 log = { workspace = true }
 num_cpus = { workspace = true }
-mimalloc = { workspace = true }
```
alchemist51, line 39: where are we adding jemalloc in this plugin?

sandbox/libs/dataformat-native/rust/Cargo.toml
```diff
@@ -46,7 +46,8 @@ serde_json = "1.0"
 log = "0.4"
 
 # Allocator
-mimalloc = { version = "0.1.48", default-features = false }
+tikv-jemallocator = { version = "0.6", features = ["disable_initial_exec_tls"] }
```
alchemist51, line 53: Why are we adding disable_initial_exec_tls? We were able to run without it right?

sandbox/plugins/analytics-backend-datafusion/rust/Cargo.toml
bharath-techie: not needed right, usage is only in df-native

sandbox/libs/dataformat-native/rust/Cargo.toml
bharath-techie, line 53: kindly add documentation for `disable_initial_exec_tls`

sandbox/libs/dataformat-native/src/test/java/org/opensearch/nativebridge/spi/NativeMemoryMetricsTests.java
himshikhagupta: -1 only occurs if jemalloc isn't the active allocator (which can't happen in production since we set #[global_allocator]), and the Java test already validates the positive path. This is just a defensive mechanism to ensure there is no crash for some unhandled case.

sandbox/libs/dataformat-native/rust/Cargo.toml
himshikhagupta: The disable_initial_exec_tls feature is required for correctness when loading the native library via dlopen/JVM FFM. Without it, jemalloc uses initial-exec TLS which fails with "cannot allocate memory in static TLS block" on Linux (aarch64). This worked on macOS during development because macOS uses a different TLS model, but failed on EC2 (aarch64 Linux) where the library is loaded at runtime into an already-initialized process. With this feature enabled, jemalloc uses the global-dynamic TLS model which is compatible with runtime loading — confirmed working on both macOS (local dev) and Linux (EC2 cluster nodes).

sandbox/libs/dataformat-native/rust/lib/src/lib.rs
himshikhagupta: Added cluster setting for dirty_decay_ms and muzzy_decay_ms. This is still required though otherwise jemalloc gets started with default values and applies the cluster setting after they are available through module.

sandbox/libs/dataformat-native/rust/common/src/metrics.rs
himshikhagupta: Changed interface to send error

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
```diff
@@ -0,0 +1,163 @@
...
+        epoch: epoch::mib().unwrap(),
+        allocated: stats::allocated::mib().unwrap(),
+        resident: stats::resident::mib().unwrap(),
+    })
+}
+
+/// Advances the jemalloc epoch and reads both stats atomically.
+fn refresh_stats() -> Result<(i64, i64), String> {
+    let m = mib();
+    m.epoch.advance().map_err(|e| format!("jemalloc epoch advance failed: {}", e))?;
```
alchemist51, line 38: Will this error cause panic?

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
```diff
@@ -0,0 +1,163 @@
...
+pub fn resident_bytes() -> i64 {
+    match refresh_stats() {
+        Ok((_, res)) => res,
+        Err(msg) => into_error_ptr(msg),
+    }
+}
+
+/// FFI: Returns current jemalloc allocated bytes, or negative error pointer.
+#[no_mangle]
+pub extern "C" fn native_jemalloc_allocated_bytes() -> i64 {
```
alchemist51, line 70: Do we need to use #[ffm_safe] for handling error?

sandbox/modules/native-bridge/build.gradle
```diff
@@ -0,0 +1,20 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+opensearchplugin {
+  description = 'Native bridge module: manages jemalloc allocator settings for the native (Rust/FFM) layer.'
```
bharath-techie: Nit : maybe lets keep this a bit generic for anything that needs df-native wide tuning

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
himshikhagupta: This should ideally not panic, but wrapped the FFM method with `std::panic::catch_unwind` to avoid any unhandled exceptions.

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
himshikhagupta: #[ffm_safe] cant be used directly, since Rust doesn't allow a crate to reference itself by name from within. Since these functions live inside native-bridge-common itself, we apply the same logic manually. The error contract is identical to all other #[ffm_safe] functions

sandbox/libs/dataformat-native/rust/common/src/error.rs
```diff
@@ -23,6 +23,19 @@ pub fn into_error_ptr(msg: String) -> i64 {
     -(ptr as i64)
 }
 
+/// Wraps a closure with `catch_unwind` and error-pointer conversion.
+/// Same contract as `#[ffm_safe]` but usable within this crate.
+pub fn ffm_wrap<F>(name: &str, f: F) -> i64
+where
+    F: FnOnce() -> Result<i64, String> + std::panic::UnwindSafe,
+{
+    match std::panic::catch_unwind(f) {
```
alchemist51, line 33: Can we make this common function here:
```
match ::std::panic::catch_unwind(::std::panic::AssertUnwindSafe(
                || -> ::std::result::Result<i64, ::std::string::String> #body
            )) {
                Ok(Ok(v)) => v,
                Ok(Err(msg)) => native_bridge_common::error::into_error_ptr(msg),
                Err(panic) => {
                    let msg = if let Some(s) = panic.downcast_ref::<String>() {
                        s.clone()
                    } else if let Some(s) = panic.downcast_ref::<&str>() {
                        s.to_string()
                    } else {
                        "unknown panic".to_string()
                    };
                    native_bridge_common::error::into_error_ptr(msg)
                }
            }
```

Will help in centralizing the code and one place to change

alchemist51 (approved):

bharath-techie (approved):

himshikhagupta:
> ❌ Gradle check result for [310bfb2](https://github.com/opensearch-project/OpenSearch/commit/310bfb235225f5a7fee4880fc43415f78d195a75): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/76038/)
> 

Flaky test https://github.com/opensearch-project/OpenSearch/issues/21373

PR #21498, opened 2026-05-05
Enable PPL eval string concat on the analytics-engine route via DataFusion CONCAT/CAST

ahkcs:
## Description

Drives the analytics-engine route to parity for PPL `eval` with string concatenation and `CAST(... AS STRING)`. Pairs with the [routing playbook](https://github.com/opensearch-project/sql/blob/feature/mustang-ppl-integration/docs/dev/ppl-analytics-engine-routing.md) (developed alongside the [fillnull PR #21472](https://github.com/opensearch-project/OpenSearch/pull/21472)) — same pattern, one new resolver helper, three new project capabilities, one Bucket-2 adapter, one QA IT.

### Bucket classification

| Surface | Bucket | Mechanism |
|---|---|---|
| `'lit' + str_field`, `str + str` | **S0** capability | `STANDARD_PROJECT_OPS` += `CONCAT` |
| `CAST(x AS STRING)` (non-null source) | **S0** capability | `STANDARD_PROJECT_OPS` += `CAST` |
| `CAST(x AS STRING)` (nullable source) | **S0** capability | `STANDARD_PROJECT_OPS` += `SAFE_CAST` |
| Null-propagation through `\|\|` | **S1** adapter | `ConcatFunctionAdapter` rewrites `\|\|(a,b)` → `CASE WHEN IS_NULL(a) OR IS_NULL(b) THEN NULL ELSE \|\|(a,b) END` |
| Symbolic-operator resolution (`\|\|`) | Planner fix | `ScalarFunction.fromSqlOperator(SqlOperator)` |

### Failure modes addressed

PPL string `+` lowers to Calcite's `SqlStdOperatorTable.CONCAT` — a `SqlBinaryOperator` named `\|\|` with `SqlKind.OTHER`. Before this PR, both `OpenSearchProjectRule` and `OpenSearchFilterRule` failed to resolve it because:

- `ScalarFunction.fromSqlKind(OTHER)` returns null (OTHER is shared by many ops).
- `instanceof SqlFunction` is false for `SqlBinaryOperator`, so the SqlFunction-cast branch was skipped.

```
java.lang.IllegalStateException: No backend supports scalar function [null] among [datafusion]
  at OpenSearchProjectRule.annotateExpr(OpenSearchProjectRule.java:123)
```

After wiring `CONCAT`/`CAST`/`SAFE_CAST` into `STANDARD_PROJECT_OPS`, the next failure was DataFusion-side null semantics: `'Age: ' + CAST(null AS STRING)` returned `'Age: '` instead of `null`. Calcite's `\|\|` follows SQL standard (any NULL → NULL); Substrait's default extension catalog documents the same; but DataFusion's substrait reader maps to its native `concat()` function which treats NULL as empty string. `ConcatFunctionAdapter` short-circuits via a CASE/IS_NULL wrapper.

### Test results

|  | Before | After |
|---|---|---|
| `CalciteEvalCommandIT` (SQL-plugin v2-side, analytics-engine route) | 0/4 | **4/4** |
| `EvalCommandIT` (this PR, QA-side, `POST /_analytics/ppl`) | n/a | **4/4** |
| `FillNullCommandIT` (regression check) | 13/13 | **13/13** |
| `AppendCommandIT` (regression check) | passing | **passing** |
| `ScalarFunctionTests` (new unit coverage) | n/a | **7/7** |
| `./gradlew check -p sandbox` | passing | **passing** |

### Design trade-off worth flagging

`ConcatFunctionAdapter` uses a CASE/IS_NULL plan rewrite rather than a custom DataFusion null-propagating concat UDF. The rewrite is surgical, reuses the existing Substrait conversion path (CASE serializes to `if_then`, no new extension), and avoids cross-language work. The cost is a per-`\|\|`-call CASE wrapper that double-references each operand in IS_NULL checks — fine for `RexInputRef`/`RexLiteral` (the only shapes PPL emits for string concat today), proportional in nested concats (`'A=' + str0 + ', B=' + str2` becomes three CASE-wrapped `\|\|` calls — each evaluating its direct operands twice, not the whole expression). A Bucket-3 UDF is the alternative if reviewers prefer; happy to move the adapter out if so.

### Commit shape

| # | Commit | Files | Why split |
|---|---|---|---|
| 1 | `[Analytics Framework] Resolve symbolic operators and add SAFE_CAST` | `ScalarFunction.java`, `ScalarFunctionTests.java` | Generic resolver fix, useful even outside this PR (any future symbolic operator). |
| 2 | `[Analytics Engine] Migrate rules and adapter dispatch to fromSqlOperator` | `OpenSearchProjectRule`, `OpenSearchFilterRule`, `BackendPlanAdapter` | Mechanical migration of 3 call sites; behavior change only for previously-unresolvable operators. |
| 3 | `[Analytics Backend / DataFusion] Wire CONCAT/CAST/SAFE_CAST + concat null adapter` | `DataFusionAnalyticsBackendPlugin.java`, `ConcatFunctionAdapter.java` | Eval-specific capability + Bucket-2 adapter. |
| 4 | `[QA] Add EvalCommandIT for the analytics-engine REST path` | `EvalCommandIT.java` | QA-side IT in `sandbox/qa/analytics-engine-rest`, reuses `calcs` dataset. |

### Forward pattern

After this lands, future Bucket-1 PPL `eval` functions for which DataFusion has the substrait mapping are one-line additions to `STANDARD_PROJECT_OPS` plus an enum entry in `ScalarFunction` if missing — no resolver work needed.

### By submitting this pull request

- [x] My code follows [the OpenSearch style guidelines](https://github.com/opensearch-project/OpenSearch/blob/main/DEVELOPER_GUIDE.md#java-language-formatting-guidelines).
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] I have signed off my commit using DCO.

### DIFF
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
@@ -185,7 +185,9 @@ public enum ScalarFunction {
                 byOperator.put(func.referenceOperator, func);
             }
         }
-        BY_REFERENCE_OPERATOR = Map.copyOf(byOperator);
+        // The HashMap is private static final and never exposed beyond the get() in the resolver
+        // below — wrapping it in Map.copyOf adds an allocation without any external safety guarantee.
+        BY_REFERENCE_OPERATOR = byOperator;
     }
 
     /**
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
@@ -10,8 +10,10 @@ package org.opensearch.analytics.spi;
 
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlOperator;
 
 import java.util.Locale;
+import java.util.Map;
 
 /**
  * All scalar functions a backend may support — comparisons, full-text search,
@@ -68,6 +70,14 @@ public enum ScalarFunction {
 
     // ── Cast / type ──────────────────────────────────────────────────
     CAST(Category.SCALAR, SqlKind.CAST),
+    /**
+     * Calcite's {@code SAFE_CAST} — emitted by PPL's explicit {@code CAST(... AS ...)} when the
+     * source value may be NULL or the conversion may fail; returns NULL on failure rather than
+     * throwing. Resolves through {@link SqlKind#SAFE_CAST}, distinct from {@link #CAST} which
+     * uses {@link SqlKind#CAST}. DataFusion's native cast already returns NULL on conversion
+     * failure, so SAFE_CAST and CAST share the same backend semantics.
+     */
+    SAFE_CAST(Category.SCALAR, SqlKind.SAFE_CAST),
 
     // ── Conditional ──────────────────────────────────────────────────
     CASE(Category.SCALAR, SqlKind.CASE),
@@ -133,5 +143,39 @@ public enum ScalarFunction {
         // TODO: Add an explicit functionName field per enum constant instead of relying on
         // valueOf(toUpperCase). This couples enum constant naming to SQL function naming convention.
         return ScalarFunction.valueOf(function.getName().toUpperCase(Locale.ROOT));
+    }
+
+    /**
+     * Symbolic operator names (e.g. {@code "||"}) whose enum constant cannot be resolved by
+     * {@link #valueOf(String)}. Calcite emits these as {@link org.apache.calcite.sql.SqlBinaryOperator}
+     * instances with {@link SqlKind#OTHER}, so neither {@link #fromSqlKind(SqlKind)} nor
+     * {@link #fromSqlFunction(SqlFunction)} resolves them.
+     */
+    private static final Map<String, ScalarFunction> SYMBOLIC_OPERATOR_NAMES = Map.of("||", CONCAT);
+
+    /**
+     * Maps any Calcite {@link SqlOperator} to a {@link ScalarFunction}, or returns null if
+     * unrecognized. Resolution order: {@link SqlKind} match, then symbolic-name lookup
+     * (handles {@code ||}), then identifier-name {@link #valueOf(String)} match.
+     *
+     * <p>Prefer this entry point over {@link #fromSqlKind(SqlKind)} /
+     * {@link #fromSqlFunction(SqlFunction)} when resolving an arbitrary {@code RexCall}'s
+     * operator: a {@code RexCall} may be backed by a {@code SqlBinaryOperator} (e.g. {@code ||})
+     * which is neither covered by {@code OTHER} {@code SqlKind} nor by {@code SqlFunction}.
+     */
+    public static ScalarFunction fromSqlOperator(SqlOperator operator) {
+        ScalarFunction byKind = fromSqlKind(operator.getKind());
+        if (byKind != null) {
+            return byKind;
+        }
+        ScalarFunction bySymbolicName = SYMBOLIC_OPERATOR_NAMES.get(operator.getName());
+        if (bySymbolicName != null) {
+            return bySymbolicName;
+        }
+        try {
+            return ScalarFunction.valueOf(operator.getName().toUpperCase(Locale.ROOT));
+        } catch (IllegalArgumentException ignored) {
+            return null;
+        }
     }
 }
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
@@ -163,7 +163,7 @@ public enum ScalarFunction {
      * operator: a {@code RexCall} may be backed by a {@code SqlBinaryOperator} (e.g. {@code ||})
      * which is neither covered by {@code OTHER} {@code SqlKind} nor by {@code SqlFunction}.
      */
-    public static ScalarFunction fromSqlOperator(SqlOperator operator) {
+    public static ScalarFunction fromSqlOperatorWithFallback(SqlOperator operator) {
         ScalarFunction byKind = fromSqlKind(operator.getKind());
         if (byKind != null) {
             return byKind;
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
@@ -11,6 +11,7 @@ package org.opensearch.analytics.spi;
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
 
 import java.util.HashMap;
 import java.util.Locale;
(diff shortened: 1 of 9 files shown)


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -125,4 +135,38 @@ public static ScalarFunction fromSqlFunction(SqlFunction function) {
...
+     * Maps any Calcite {@link SqlOperator} to a {@link ScalarFunction}, or returns null if
+     * unrecognized. Resolution order: {@link SqlKind} match, then symbolic-name lookup
+     * (handles {@code ||}), then identifier-name {@link #valueOf(String)} match.
+     *
+     * <p>Prefer this entry point over {@link #fromSqlKind(SqlKind)} /
+     * {@link #fromSqlFunction(SqlFunction)} when resolving an arbitrary {@code RexCall}'s
+     * operator: a {@code RexCall} may be backed by a {@code SqlBinaryOperator} (e.g. {@code ||})
+     * which is neither covered by {@code OTHER} {@code SqlKind} nor by {@code SqlFunction}.
+     */
+    public static ScalarFunction fromSqlOperator(SqlOperator operator) {
```
expani: We can rename this to `fromSqlOperatorWithFallback` or something similar to indicate there is an explicit fallback inside. 

I see that the API docs cover it, but always good to have it in the name as well.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -125,4 +135,38 @@ public static ScalarFunction fromSqlFunction(SqlFunction function) {
...
         return ScalarFunction.valueOf(function.getName().toUpperCase(Locale.ROOT));
     }
+
+    /**
+     * Symbolic operator names (e.g. {@code "||"}) whose enum constant cannot be resolved by
+     * {@link #valueOf(String)}. Calcite emits these as {@link org.apache.calcite.sql.SqlBinaryOperator}
+     * instances with {@link SqlKind#OTHER}, so neither {@link #fromSqlKind(SqlKind)} nor
+     * {@link #fromSqlFunction(SqlFunction)} resolves them.
+     */
+    private static final Map<String, ScalarFunction> SYMBOLIC_OPERATOR_NAMES = Map.of("||", CONCAT);
```
expani: How big will this grow as we add more functions ? 

Should we think around adding it as a new Nullable field called `symbolicOperatorName` ? 

Also, how many more such customizations per ScalarFunction are we expecting ? in other words, how many other special maps do we need ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConcatFunctionAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+        RexBuilder rexBuilder = cluster.getRexBuilder();
+        List<RexNode> nullChecks = new ArrayList<>(operands.size());
+        for (RexNode operand : operands) {
+            // RexLiteral.isAlwaysTrue is false for non-NULL literals, but the dedicated
+            // RexBuilder.makeNullLiteral path is unnecessary — IS_NULL on a non-null literal
+            // becomes a constant-false predicate, so OR(constant_false, ...) reduces cleanly.
+            nullChecks.add(rexBuilder.makeCall(SqlStdOperatorTable.IS_NULL, operand));
+        }
+        RexNode anyNull = nullChecks.get(0);
+        for (int i = 1; i < nullChecks.size(); i++) {
```
expani: Can be simplified to
```
RexNode anyNull = rexBuilder.makeCall(SqlStdOperatorTable.IS_NULL, operands.get(0));
   for (int i = 1; i < operands.size(); i++) {
       anyNull = rexBuilder.makeCall(SqlStdOperatorTable.OR, anyNull,
           rexBuilder.makeCall(SqlStdOperatorTable.IS_NULL, operands.get(i)));
   }
```

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
```diff
@@ -141,7 +149,7 @@ public Set<AggregateCapability> aggregateCapabilities() {
 
             @Override
             public Map<ScalarFunction, ScalarFunctionAdapter> scalarFunctionAdapters() {
-                return Map.of(ScalarFunction.TIMESTAMP, new TimestampFunctionAdapter());
+                return Map.of(ScalarFunction.TIMESTAMP, new TimestampFunctionAdapter(), ScalarFunction.CONCAT, new ConcatFunctionAdapter());
```
expani: nit : Let's use a formatting where every function is in one line unless spotless complains

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
expani: Can we look at using `SqlStdOperatorTable.CONCAT` for comparison as it won't require us to maintain these strings ? 

Or is there a case where even SqlStdOperatorTable would not work ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
ahkcs: Spotless rejects per-pair Map.of formatting; kept single-line and added a comment + alphabetical convention for future entries

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConcatFunctionAdapter.java
ahkcs: Updated

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
ahkcs: Renamed

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
ahkcs: Replaced `String symbolicOperatorName = "||"` with `SqlOperator referenceOperator = SqlStdOperatorTable.CONCAT` on the enum

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -125,4 +159,50 @@ public static ScalarFunction fromSqlFunction(SqlFunction function) {
...
+    private static final Map<SqlOperator, ScalarFunction> BY_REFERENCE_OPERATOR;
+
+    static {
+        Map<SqlOperator, ScalarFunction> byOperator = new HashMap<>();
+        for (ScalarFunction func : values()) {
+            if (func.referenceOperator != null) {
+                byOperator.put(func.referenceOperator, func);
+            }
+        }
+        BY_REFERENCE_OPERATOR = Map.copyOf(byOperator);
```
expani: nit: Don't think we need another copy here.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConcatFunctionAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * The inner {@code ||} is left intact and serializes through the same Substrait conversion path,
+ * but with the surrounding CASE/IS_NULL the DataFusion {@code concat()} call is short-circuited
+ * for any input that contains a NULL — restoring SQL-standard null-propagation without requiring
+ * a custom DataFusion UDF.
+ *
+ * <p>Single-operand calls fall through unchanged (the result equals the operand, so no
+ * null-handling rewrite is needed).
+ */
+class ConcatFunctionAdapter implements ScalarFunctionAdapter {
```
expani, line 45: Can we add some UTs for this ?

expani (approved):
Thanks @ahkcs for the iteration. 

Left some minor comments, but approving it nonetheless.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
ahkcs: dropped

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConcatFunctionAdapter.java
ahkcs: added `ConcatFunctionAdapterTests`

mch2 (approved):

mch2:
@ahkcs can u pls rebase to pick up sandbox check fix.

ahkcs:
> @ahkcs can u pls rebase to pick up sandbox check fix.

Rebased

mch2:
going ahead all sandbox code and sandbox check passes:

gc again:
```
WORKFLOW_URL
time passed: 3540
```

PR #21499, opened 2026-05-05
Add support for PPL SEARCH function

mch2:
### Description
Add support for PPL SEARCH function using our scalar adapter pattern.

Calcite folds IN-lists, BETWEEN, and unions of ranges into a compact SEARCH(field, Sarg[...]) RexCall that DataFusion's substrait consumer can't decode — Sarg has no wire representation. SearchAdapter runs during the backend-adapt pass and calls RexUtil.expandSearch to invert the fold before substrait serialization.

Adds ScalarFunction.SEARCH (SqlKind.SEARCH) to the SPI enum, registers SearchAdapter with the DataFusion backend, and declares SEARCH as a filter capability so the filter rule accepts the predicate and lets the adapter handle it.

SearchAdapterTests covers IN-list, BETWEEN, and pass-through for non-SEARCH. SearchOperatorIT exercises the full PPL -> Calcite -> adapter -> substrait -> DataFusion path against the calcs dataset with three Sarg shapes.

### Check List


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -37,6 +37,7 @@ public enum ScalarFunction {
     IN(Category.COMPARISON, SqlKind.IN),
     LIKE(Category.COMPARISON, SqlKind.LIKE),
     PREFIX(Category.COMPARISON, SqlKind.OTHER_FUNCTION),
+    SEARCH(Category.COMPARISON, SqlKind.SEARCH),
```
sandeshkr419: I don't like having a super generic name `SEARCH` as I this implicitly implies that its a full text search which this is not. 

How about we call it `SARG` or `SARG_PREDICATE` - something that narrows it down and doesn't creates up confusion when we introduce other similar support for other functions.

----

Accordingly, this will overflow to SearchAdapter -> SargAdapter as well.

----

Do we need a better category for this as well?

ahkcs:
The QA IT covers `IN / BETWEEN / range-union` but not `NOT IN`. Maybe we can have one additional `testNotInListFoldsToSearch`

ahkcs:
Filter side looks complete and the adapter is the right shape. One thing I'd ask before merging: project-side seems not included, can we have IT tests that verify queries like 
`source=calcs | eval is_match = int0 in (1, 8, 10) | fields str2, int0, is_match` 
would pass without throwing `No backend supports scalar function [SEARCH].` Probably one-line fix (add to `STANDARD_PROJECT_OPS`) since the same adapter handles both.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
mch2: sure, we can rename and fallback to SqlKind given no other function maps to SEARCH.

sandeshkr419 (approved):

PR #21501, opened 2026-05-05
[Sandbox] Add calcite-patched module: Gradle-driven patched Calcite build

bowenlan-amzn:
Depends on #21549 publish patched calcite [here](https://ci.opensearch.org/ci/dbc/snapshots/maven/org/apache/calcite/)

## What

Wires the `analytics-engine` sandbox plugin to the **pre-published patched Calcite** at `org.apache.calcite:calcite-core:1.41.0-opensearch-1` (and `calcite-linq4j`) from OpenSearch's S3 Maven snapshots repo (version composed from the `calcite` + `calcite_os_rev` keys in `gradle/libs.versions.toml`). No local Calcite build machinery required — it's just a Maven coordinate.

## Why

Fixes [CALCITE-3745](https://issues.apache.org/jira/browse/CALCITE-3745) — Calcite hardcodes Janino's parent classloader, which breaks UDF resolution in OpenSearch's plugin classloader hierarchy. The patch introduces a `TcclChainedClassLoader` so Janino-compiled code can see classes in child plugin classloaders. Only runtime classloader behavior changes; API surface and bytecode references are identical to vanilla.

## Design

- **`sandbox/build.gradle`** declares the OpenSearch Snapshots Maven repo and a `resolutionStrategy.force` on `calcite-core` / `calcite-linq4j` inside its existing `subprojects { }` block. Every sandbox subproject inherits both — so there's no need for per-plugin duplication. The force uses `"${libVersions['calcite']}-opensearch-${libVersions['calcite_os_rev']}"`, reading the keys from `gradle/libs.versions.toml` as the single source of truth (bump `calcite_os_rev` there and every sandbox plugin follows). The force also pins `unified-query-*`'s transitive vanilla `1.41.0` (via `calcite-babel`) to the patched revision.
- **`analytics-framework`** declares `api "org.apache.calcite:calcite-core:${calciteVersion}"` (and `calcite-linq4j`), with `calciteVersion` composed from the same two `libs.versions.toml` keys. No local repo or force block — it inherits from the sandbox parent. The patched jars flow through `runtimeClasspath` of every consumer; the zip naturally ships the patched coordinate.
- **License SHA-1 files** under `sandbox/libs/analytics-framework/licenses/` carry the `-opensearch-1` coordinate:
  - `calcite-core-1.41.0-opensearch-1.jar.sha1` → `d4ac2aff0c76b2ea15f47940542999fa42e17d75`
  - `calcite-linq4j-1.41.0-opensearch-1.jar.sha1` → `e9bcb0ec7ca38a4bff84283b39d4a736c5217645`
- **No post-processing of `bundlePlugin`.** `compileJava`, `test`, `forbiddenApis`, `thirdPartyAudit`, `dependencyLicenses`, and the `bundlePlugin` Zip task all see the same calcite jars — the patched ones.

## Companion

SQL plugin side: [sql#5302](https://github.com/opensearch-project/sql/pull/5302) (sets the TCCL before invoking Calcite).

## E2e verification

Reproduced the failure with vanilla Calcite, then confirmed the patched build fixes it. Query under test: `source=t | eval u = upper(name) | fields u`, which triggers `SqlFunctions.<clinit>` → Janino code-gen → parent-CL lookup — the exact path CALCITE-3745 is about.

<details>
<summary>Before / after (click to expand)</summary>

Companion branch for this run: [sql#5403](https://github.com/opensearch-project/sql/pull/5403) (`analytics-optional`), rebased on latest `feature/mustang-ppl-integration`.

### Before — vanilla Calcite in the zip

Zip ships with vanilla:
```
8811320 calcite-core-1.41.0.jar
 523019 calcite-linq4j-1.41.0.jar
```

Query returns HTTP 500:
```json
{"error":{"reason":"There was internal problem at backend",
  "details":"Line 3, Column 3: Cannot determine simple type name \"org\"",
  "type":"CompileException"},"status":500}
```

Node log shows Janino failing to resolve the fully-qualified `org.apache.calcite.runtime.SqlFunctions` reference in its own generated code — classic CALCITE-3745: Janino's parent CL is pinned to a classloader that can't see the Calcite package it was generated from.

```
// Janino-generated body (excerpt)
return inputEnumerator.current() == null || inputEnumerator.current().toString() == null
    ? null
    : org.apache.calcite.runtime.SqlFunctions.upper(
        inputEnumerator.current() == null ? null : inputEnumerator.current().toString());
```

```
Caused by: org.codehaus.commons.compiler.CompileException:
    Line 3, Column 3: Cannot determine simple type name "org"
    at org.codehaus.janino.UnitCompiler.compileError(UnitCompiler.java:13228)
    at org.codehaus.janino.UnitCompiler.getRawReferenceType(UnitCompiler.java:7321)
    …
    at org.codehaus.janino.SimpleCompiler.cook(SimpleCompiler.java:219)
    at org.apache.calcite.adapter.enumerable.EnumerableInterpretable
        .compileToBindable(EnumerableInterpretable.java:176)
```

### After — patched Calcite in the zip

Zip ships with patched:
```
8889483 calcite-core-1.41.0-opensearch-1.jar
 528940 calcite-linq4j-1.41.0-opensearch-1.jar
```

Same query, same cluster:
```json
{"schema":[{"name":"u","type":"string"}],
 "datarows":[["ITEM1"],["ITEM2"]],
 "total":2,"size":2}
```

Log scan for `CompileException` / `ClassNotFoundException.*calcite` / `Janino.*classloader` across both nodes: **0 hits**.

`TcclChainedClassLoader` (introduced by the patch) makes Janino's parent CL consult the thread context classloader first — which in the plugin CL hierarchy sees the Calcite package — before falling back to the original CL for anything else.

</details>

## Check List


peterzhuamazon:
Put to draft to give a review.

mch2 (approved):

PR #21504, opened 2026-05-06
Add PPL operator support: 19 comparison / arithmetic / logical operators

mch2:
### Description
Add PPL operator support: 19 comparison / arithmetic / logical operators

  Without this, PPL queries using these operators in eval expressions fail
  at plan time with "No backend supports scalar function [...] among
  [datafusion]" — the analytics-engine planner requires each backend to
  declare per-operator project capability, with no implicit default for
  built-ins.

  Covered on both filter- and project-side:
    Comparison: =, !=, <, <=, >, >=, between, in, like, ilike
    Logical: and, or, not, xor
    Arithmetic: +, -, *, /, %

  Adds project-side capability declarations so eval can emit these operators,
  a LikeAdapter to rewrite PPL's default ILIKE (3-arg with escape) to LIKE
  (2-arg) for the Substrait `like` signature, and a StdOperatorRewriteAdapter
  for PPL's custom DIVIDE/MOD UDF wrappers so Isthmus's default SCALAR_SIGS
  resolves them. AND/OR/NOT are handled structurally by the filter rule and
  folded to Sarg/inverse-comparison by Calcite before reaching the project
  rule — no enum entries needed.

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/StdOperatorRewriteAdapter.java
```diff
@@ -0,0 +1,76 @@
...
+ *   <li>PPL's {@code MOD} UDF → {@code SqlStdOperatorTable.MOD} → substrait {@code modulus}.</li>
+ * </ul>
+ *
+ * <p>Adapter-level rewriting (rather than extending Isthmus's {@code ADDITIONAL_SCALAR_SIGS})
+ * keeps the rewrite local to the DataFusion backend and avoids cross-cutting changes to the
+ * framework or SPI. The rewrite preserves operand order and result type.
+ *
+ * @opensearch.internal
+ */
+class StdOperatorRewriteAdapter implements ScalarFunctionAdapter {
```
expani: analytics-engine :)

sandbox/plugins/analytics-engine/licenses/jackson-databind-2.21.3.jar.sha1
```diff
@@ -1 +0,0 @@
-aa7ccec161c275f3e6332666ab758916f3120714
```
expani, line 1: is this intentional ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/LikeAdapter.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+
+/**
+ * Adapts PPL-emitted LIKE-family calls so Isthmus can serialize them via the Substrait default
+ * extension catalog's {@code like} function. Two rewrites happen in sequence:
+ * <ol>
+ *   <li>If the call's operator is {@code SqlLibraryOperators.ILIKE} (name "ILIKE"), substitute
+ *       {@link SqlStdOperatorTable#LIKE}. PPL defaults to case-insensitive LIKE and emits ILIKE
+ *       with {@code SqlKind.LIKE}; Substrait's default catalog has no {@code ilike} entry, so
+ *       the call has to become a plain LIKE to serialize. This loses case-insensitivity —
```
expani: Datafusion also has the case sensitive and insensitive variations. 

If we add a custom function signature in Isthumus, I think we can cover all the cases.

sandbox/plugins/analytics-engine/licenses/jackson-databind-2.21.3.jar.sha1
mch2: yeah, its provided by arrow-flight-rpc now at runtime, with it the license check throws unused sha file error.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/StdOperatorRewriteAdapter.java
mch2: moved

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/LikeAdapter.java
mch2: ah yeah, updating this to use a yaml extension to map direclty

expani (approved):
Thanks for this @mch2 🚀

bharath-techie (approved):

PR #21515, opened 2026-05-06
Grpc client poc

OVI3D0:
### Description
Adds reverse converters for Bulk and a subset of Search queries alongside the existing forward parsers in `modules/transport-grpc`

BulkProtoConverter handles INDEX, CREATE, UPDATE, and DELETE, along with cluster level fields. The BulkResponseProtoConverter converts the proto bulk responses back into BulkResponse objects. 

SearchProtoConverter does the same but for search requests. Supports 4 query types (match_all, match_none, term, terms) + TermsLookup. Other query types should throw UnsupportedOperationException

Tested using round trip tests (BulkRoundTrips.java) which calls BulkProtoConverter.toProto -> BulkRequestProtoUtils.prepareRequest and asserts that the fields are equivalent across the board + unit tests.

### Related Issues

### Check List

### DIFF
--- a/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkProtoConverter.java
+++ b/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkProtoConverter.java
@@ -1,0 +1,274 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.transport.grpc.proto.request.document.bulk;
+
+import com.google.protobuf.ByteString;
+import org.opensearch.action.DocWriteRequest;
+import org.opensearch.action.bulk.BulkRequest;
+import org.opensearch.action.delete.DeleteRequest;
+import org.opensearch.action.index.IndexRequest;
+import org.opensearch.action.support.ActiveShardCount;
+import org.opensearch.action.support.WriteRequest.RefreshPolicy;
+import org.opensearch.action.update.UpdateRequest;
+import org.opensearch.core.common.bytes.BytesArray;
+import org.opensearch.core.common.bytes.BytesReference;
+import org.opensearch.index.VersionType;
+import org.opensearch.index.seqno.SequenceNumbers;
+
+/**
+ * Converts a server-internal {@link BulkRequest} into a {@code protobufs.BulkRequest}.
+ *
+ * <p>This is the inverse of {@link BulkRequestProtoUtils#prepareRequest}, which lives next door and translates the same
+ * proto type back into a server-internal {@code BulkRequest}. Co-locating both directions enables round-trip parity tests
+ * that catch divergence as the schema evolves.
+ *
+ * <p>This converter exists so client libraries (HLRC's API model, opensearch-java in the future) can take a server-internal
+ * request object and emit a proto for transmission over gRPC. The cluster-side code already in this module then consumes
+ * that proto and reconstructs the server-internal object.
+ *
+ * <p><b>Canonicalization rules</b> (chosen so {@code prepareRequest(toProto(s))} reconstructs an equivalent
+ * {@code BulkRequest}):
+ * <ul>
+ *   <li>{@link RefreshPolicy#NONE} maps to proto unset (default {@code REFRESH_UNSPECIFIED}); the forward path treats both
+ *       {@code REFRESH_FALSE} and {@code REFRESH_UNSPECIFIED} as {@code RefreshPolicy.NONE}, so we pick the unset form.</li>
+ *   <li>{@link VersionType#INTERNAL} (server default) maps to proto unset.</li>
+ *   <li>{@link ActiveShardCount#DEFAULT} maps to proto unset.</li>
+ *   <li>Unassigned {@code ifSeqNo} / {@code ifPrimaryTerm} are not emitted.</li>
+ *   <li>{@code Versions.MATCH_ANY} is not emitted as a {@code version} field.</li>
+ *   <li>Server {@code globalIndex} / {@code globalRouting} / {@code globalPipeline} / {@code globalRequireAlias} are not
+ *       carried — the forward path inlines top-level proto defaults into per-row requests at parse time and never reads
+ *       these slots back, so the round-trippable form is per-row only.</li>
+ *   <li>A {@code CREATE} op type with version control is emitted as {@code IndexOperation} with {@code op_type=CREATE},
+ *       since {@code WriteOperation} (the dedicated CREATE proto) lacks those fields in the current schema.</li>
+ * </ul>
+ */
+public final class BulkProtoConverter {
+
+    private BulkProtoConverter() {}
+
+    /**
+     * Converts a server-internal {@link BulkRequest} to its protobuf equivalent.
+     *
+     * @param serverBulk the request to convert
+     * @return the equivalent {@code protobufs.BulkRequest}
+     */
+    public static org.opensearch.protobufs.BulkRequest toProto(BulkRequest serverBulk) {
+        org.opensearch.protobufs.BulkRequest.Builder builder = org.opensearch.protobufs.BulkRequest.newBuilder();
+
+        if (serverBulk.timeout() != null) {
+            builder.setTimeout(serverBulk.timeout().getStringRep());
+        }
+
+        org.opensearch.protobufs.Refresh refresh = toProtoRefresh(serverBulk.getRefreshPolicy());
+        if (refresh != null) {
+            builder.setRefresh(refresh);
+        }
+
+        org.opensearch.protobufs.WaitForActiveShards was = toProtoWaitForActiveShards(serverBulk.waitForActiveShards());
+        if (was != null) {
+            builder.setWaitForActiveShards(was);
+        }
+
+        for (DocWriteRequest<?> req : serverBulk.requests()) {
+            builder.addBulkRequestBody(toRequestBody(req));
+        }
+
+        return builder.build();
+    }
+
+    private static org.opensearch.protobufs.BulkRequestBody toRequestBody(DocWriteRequest<?> req) {
+        org.opensearch.protobufs.BulkRequestBody.Builder body = org.opensearch.protobufs.BulkRequestBody.newBuilder();
+        org.opensearch.protobufs.OperationContainer.Builder container = org.opensearch.protobufs.OperationContainer.newBuilder();
+
+        if (req instanceof IndexRequest) {
+            IndexRequest indexReq = (IndexRequest) req;
+            if (indexReq.opType() == DocWriteRequest.OpType.CREATE) {
+                if (hasVersionControl(indexReq)) {
+                    container.setIndex(toIndexOperation(indexReq, true));
+                } else {
+                    container.setCreate(toCreateWriteOperation(indexReq));
+                }
+            } else {
+                container.setIndex(toIndexOperation(indexReq, false));
+            }
+            body.setOperationContainer(container);
+            attachSource(body, indexReq.source());
+        } else if (req instanceof UpdateRequest) {
+            UpdateRequest updateReq = (UpdateRequest) req;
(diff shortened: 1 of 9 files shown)


msfroh:
@OVI3D0 -- This is a need idea! 

I'm wondering if we want to tackle it for the REST high-level client or if it makes more sense to add this capability to the OpenSearch Java client (https://github.com/opensearch-project/opensearch-java). In particular, I believe we've been saying that the REST high-level client is on its way out and the OpenSearch Java client is preferred. (Specifically, the REST high-level client adds a compile-time dependency on the whole server, which tends to defeat some of the purpose of client/server applications.)

I would love if the OpenSearch Java client had an endpoint config so you could build a client in "REST mode" or "gRPC mode", but the rest of your code would "just work" with no other changes.

OVI3D0:
@msfroh thank you!! packaged the changes alongside the existing grpc plugin now to use as a reference for an intern project. Actually working on adding this to the OpenSearch Java client as well now :) 
@bowenlan-amzn suggested we go with a `HybridTransport` implementation that routes gRPC where supported and falls back to REST where not.

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/search/SearchProtoConverter.java
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Converts an arbitrary value to a proto {@code FieldValue}. Server-side query builders carry values as
+     * {@link Object}; concrete types are {@code String}, {@code Long}, {@code Integer}, {@code Double}, {@code Float},
+     * {@code Boolean}, or {@link BytesRef}. {@code BytesRef} is unwrapped to its UTF-8 string form, since the proto
+     * forward parser does not accept it directly.
+     *
+     * @throws IllegalArgumentException if {@code value} is none of the supported runtime types
+     */
+    private static org.opensearch.protobufs.FieldValue toFieldValue(Object value) {
```
finnegancarroll, line 164: I think this is already implemented in [FieldValueProtoUtils](https://github.com/opensearch-project/OpenSearch/blob/main/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/common/FieldValueProtoUtils.java#L38).

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkProtoConverter.java
```diff
@@ -0,0 +1,274 @@
...
+                container.setIndex(toIndexOperation(indexReq, false));
+            }
+            body.setOperationContainer(container);
+            attachSource(body, indexReq.source());
+        } else if (req instanceof UpdateRequest) {
+            UpdateRequest updateReq = (UpdateRequest) req;
+            container.setUpdate(toUpdateOperation(updateReq));
+            body.setOperationContainer(container);
+            body.setUpdateAction(buildUpdateAction(updateReq));
+            // The forward parser reads update doc bytes from BulkRequestBody.object and ignores
```
finnegancarroll: Do we want to prefer the update action doc here? It seems like the bulk request body is currently just a fallback and [may be removed at some point.](https://github.com/opensearch-project/OpenSearch/blob/main/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkRequestParserProtoUtils.java#L205-L206)

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/search/SearchProtoConverter.java
```diff
@@ -0,0 +1,182 @@
...
+        org.opensearch.protobufs.TermQuery.Builder b = org.opensearch.protobufs.TermQuery.newBuilder()
+            .setField(tq.fieldName())
+            .setValue(toFieldValue(tq.value()));
+        if (tq.boost() != 1.0f) b.setBoost(tq.boost());
+        if (tq.queryName() != null) b.setXName(tq.queryName());
+        if (tq.caseInsensitive()) b.setCaseInsensitive(true);
+        return b.build();
+    }
+
+    private static org.opensearch.protobufs.TermsQuery toTerms(TermsQueryBuilder tsq) {
```
finnegancarroll, line 123: Do we have access to the terms value type here? It's handled on the server side but we don't write it to the protobuf here.

https://github.com/opensearch-project/OpenSearch/blob/main/modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/search/query/TermsQueryBuilderProtoUtils.java#L57-L59

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/response/document/bulk/BulkResponseProtoConverter.java
```diff
@@ -0,0 +1,172 @@
...
+     * @param proto the proto response received over gRPC
+     * @return the equivalent {@link BulkResponse}
+     */
+    public static BulkResponse fromProto(org.opensearch.protobufs.BulkResponse proto) {
+        int n = proto.getItemsCount();
+        BulkItemResponse[] items = new BulkItemResponse[n];
+        for (int i = 0; i < n; i++) {
+            items[i] = fromProto(i, proto.getItems(i));
+        }
+        long ingestTook = proto.hasIngestTook() ? proto.getIngestTook() : -1L;
```
finnegancarroll: nit: replace hard coded with constant [NO_INGEST_TOOK](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/bulk/BulkResponse.java#L70).

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkProtoConverter.java
```diff
@@ -0,0 +1,274 @@
...
+        org.opensearch.protobufs.UpdateAction.Builder action = org.opensearch.protobufs.UpdateAction.newBuilder();
+
+        // Note: doc bytes go on BulkRequestBody.object, not UpdateAction.doc; see toRequestBody.
+        if (req.upsertRequest() != null && req.upsertRequest().source() != null) {
+            action.setUpsert(toByteString(req.upsertRequest().source()));
+        }
+        if (req.docAsUpsert()) action.setDocAsUpsert(true);
+        if (req.scriptedUpsert()) action.setScriptedUpsert(true);
+        if (req.detectNoop() == false) action.setDetectNoop(false); // server default is true
+        // Script support deferred — the Script proto is a non-trivial subgraph.
```
finnegancarroll: if we aren't handling scripts yet lets throw an exception if we see one.

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkProtoConverter.java
```diff
@@ -0,0 +1,274 @@
...
+
+        return action.build();
+    }
+
+    private static ByteString toByteString(BytesReference bytes) {
+        BytesArray ba = (bytes instanceof BytesArray) ? (BytesArray) bytes : new BytesArray(bytes.toBytesRef());
```
finnegancarroll: Are these cases different? I see that `new BytesArray()` ends up just creating a deep copy as well, so I think they end up both just copying the bytes over.

```
    public static BytesRef deepCopyOf(BytesRef other) {
        return new BytesRef(ArrayUtil.copyOfSubArray(other.bytes, other.offset, other.offset + other.length), 0, other.length);
    }
```

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/search/SearchProtoConverter.java
OVI3D0: replaced with a thin wrapper around `FieldValueProtoUtils.toProto()`. Only adds null→NULL_VALUE_NULL and BytesRef→unwrap-to-string on top.

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkProtoConverter.java
OVI3D0: yeah, changed to prefer `UpdateAction.doc` over `BulkRequestBody.object`

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkProtoConverter.java
OVI3D0: should throw `UnsupportedOperationException` on scripts for now

modules/transport-grpc/src/main/java/org/opensearch/transport/grpc/proto/request/document/bulk/BulkProtoConverter.java
OVI3D0: i see, yeah seems redundant, simplified to `bytes.toBytesRef() → ByteString.copyFrom(ref.bytes, ref.offset, ref.length)`

karenyrx:
Thanks @OVI3D0 for putting this together 

Before we merge a permanent module of this size (which looks to be roughly 2x the amount of code in the transport-grpc plugin) to maintain, could we ensure we arent introducing a long-term maintainence burden

For example, could you share via a preliminary RFC/issue to align on the intended usage / clear roadmap of these reverse converters, taht help us understand better: 
1. Target consumers: Who are the intended consumers of this 
- For example is it just https://docs.opensearch.org/latest/clients/java-rest-high-level/ ? This is explicitly not recommended for use so wondering if we have other use cases in mind  
2. Long term roadmap: I realize we have equivalent reverse converters on the REST side as well, but is it planned for us to keep these or deprecate them in the future? If so this proto reverse converter can be treated the same. 
- For example @msfroh proposed/floated the idea the spec could be the source of truth long term and clients should just generate their own serializers / deserializers from the API spec
- Since we're starting afresh, could we have the high level REST client to own this code rather than the core repo? 

If we do decide to move forward with this PR, could we at least do the following to make future maintenance/moves/deletion more manageable: e.g. 
*  put all these in a separate `/clients` folder for clearer separation so we can do a simple move/deletion in the future
* add a README explaining the experimental nature/intended usage of these 
* perhaps  have`@ExperimentalApi` tags added to mark them as "deletable at any time"

msfroh:
@OVI3D0 -- I'm kind of wondering how this PR actually helps us build a good gRPC client. 

In particular, we end up adding the client-side serialization of requests and the deserialization of responses to server code, just so that clients can take a dependency. At that point, though, the client has taken a compile-time and runtime dependency on server code, which is the exact design flaw of the high-level REST client. ("You want to call OpenSearch? Here is ALL OF OPENSEARCH in your client.")

The correct thing to do is define the APIs in a specification and use the specification to generate both client and server APIs. For gRPC, we have server-side generated code and client-side generated code. IMO, the client-side generated code should be directly connected to a client's POJOs as part of a client -- not involving server POJOs.

finnegancarroll:
Hi @msfroh @karenyrx , looking to pick this up again.

> Since we're starting afresh, could we have the high level REST client to own this code rather than the core repo?

What do you think of this compromise? Our goal here is to help facilitate low effort migration for users who may not want to refactor their client side logic.

msfroh:
> Hi @msfroh @karenyrx , looking to pick this up again.
> 
> > Since we're starting afresh, could we have the high level REST client to own this code rather than the core repo?
> 
> What do you think of this compromise? Our goal here is to help facilitate low effort migration for users who may not want to refactor their client side logic.

What exactly is the compromise?

Are we talking about users of the deprecated high-level REST client? Or are we talking about users of the opensearch-java client?

IMO, we should not make any more changes to the high-level REST client and should probably have removed it in OpenSearch 3.0.

finnegancarroll:
Sorry yea i meant moving this code into the high level rest client to reduce maintenance burden but i'm remembering high level rest client is directly in core and not a separate repo, so any change in opensearch-protobufs version would automatically end up impacting HLRC as well.

PR #21516, opened 2026-05-06
refactor(hunspell): rename package terminology to generic ref_path

shayush622:
## Description
  
  Renames package-specific terminology in the hunspell dictionary loading API to generic ref_path terminology, and splits the single input validator into type-specific validators with appropriate character
  allowlists.
  
Motivation: The original API was named around "packages" (getDictionaryFromPackage, buildPackageCacheKey, validatePackageIdentifier) but the feature is a generic directory-based dictionary loading mechanism.
  This rename makes the API self-documenting and prepares for follow-up work that allows nested paths (e.g., analyzers/my-dict).
  
 ## Changes
  
  HunspellService.java — renames getDictionaryFromPackage → getDictionaryFromRefPath, loadDictionaryFromPackage → loadDictionaryFromRefPath, buildPackageCacheKey → buildRefPathCacheKey; extracts
  CACHE_KEY_SEPARATOR constant; updates path resolution to config/{ref_path}/hunspell/{locale}/; updates all javadoc and comments
  
  HunspellTokenFilterFactory.java — splits validatePackageIdentifier into validateRefPath (allows / for nested paths) and validateLocale (allows dots, disallows /); adds validateAgainstPattern helper with
  per-param error messages; updates call sites and comments
  
  HunspellServiceTests.java — renames all test methods from Package to RefPath terminology; updates API calls and assertions
  
  HunspellTokenFilterFactoryTests.java — renames validator tests; updates API calls to use validateRefPath/validateLocale
  
  No behavioral change for existing functionality. All 43 tests pass.
  
 ## Related
  
  - Follow-up to: #20840
  - RFC: #20712
  
 ## Check List
  
  For more information on following Developer Certificate of Origin and signing off your commits, please check here
  (https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -124,37 +124,68 @@ public boolean longestOnly() {
...
+     * @param value the locale to validate
      * @throws IllegalArgumentException if validation fails
      */
-    static void validatePackageIdentifier(String value, String paramName) {
+    static void validateLocale(String value) {
+        validateAgainstPattern(
+            value,
+            "locale",
+            SAFE_LOCALE_PATTERN,
+            "Only alphanumeric characters, hyphens, underscores, and dots (not leading/trailing) are allowed."
```
cwperks: Can we forbid dots entirely and use the same pattern matching as ref_path here? Better yet, if there's an Enum that captures all allowed Locales can we use that for validation?

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -354,7 +330,6 @@ private Dictionary loadDictionary(String locale, Settings nodeSettings, Environm
             throw new OpenSearchException(String.format(Locale.ROOT, "Too many affix files exist for hunspell dictionary [%s]", locale));
         }
 
-        // Load dictionary files and create Dictionary object
```
cwperks, line 357: nit: some of these comments can remain

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: AFAIK, 
OpenSearch has no enum or fixed list of allowed locales. The locale is simply a directory name under config/hunspell/ — whatever folder the user creates becomes a valid locale.

shayush622: ACK, 
I'll make the validation for locale same as ref_path. I couldn't find any existing locale with a dot in between. so doesnot make sense to allow it.

cwperks:
The changes LGTM. I had the same 2 comments as the top on https://github.com/opensearch-project/OpenSearch/pull/21516#issuecomment-4390763275, but these are fairly minor.

cwperks (approved):

PR #21519, opened 2026-05-06
Changes to handle session context handle close

bharath-techie:
### Description
Session context lifecycle is as follows
1. Create session context during shard scan
2. Execute query - session context gets converted to rust native object during execute query

This PR handles 2 problems
1. Any failures between 1 & 2 are handled by calling backendContext close
2. The session context that gets absorbed in rust heap is marked as consumed in the `ConsumedNativeHandle`

### Related Issues
Resolves TODO in https://github.com/opensearch-project/OpenSearch/pull/21479

### Check List

### DIFF
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/BackendExecutionContext.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/BackendExecutionContext.java
@@ -8,15 +8,31 @@
 
 package org.opensearch.analytics.spi;
 
+import java.io.IOException;
+
 /**
- * Marker interface for backend-specific execution context that flows between
- * successive instruction handler calls. The first handler in the chain receives
- * {@code null} and bootstraps the context; subsequent handlers receive and build
- * upon the previous handler's output.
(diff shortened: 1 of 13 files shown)


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/BackendExecutionContext.java
```diff
@@ -8,15 +8,31 @@
...
+ * no-op {@link #close()}. {@code close()} must be idempotent; in particular it must
+ * tolerate being called after the resources have already been handed off to a
+ * successfully-constructed engine.
+ *
  * @opensearch.internal
  */
-public interface BackendExecutionContext {}
+public interface BackendExecutionContext extends AutoCloseable {
+    @Override
+    default void close() throws IOException {
```
expani, line 35: Is this to avoid bigger changes ? Ideally we should not give a default for close to FORCE child classes to implement and then they no-op on their end.

Bukhtawar (approved):
LGTM

alchemist51 (approved):

PR #21523, opened 2026-05-06
feat(wlm): Add additional search settings and override_request_values to WLM groups

dzane17:
### Description

Adds additional WLM search settings and introduces `override_request_values` to control how workload group settings interact with request-level parameters.

### New search settings

Workload groups can now configure the following settings in addition to the existing `search.default_search_timeout`:

- `search.cancel_after_time_interval` — time after which a search request is cancelled
- `search.max_concurrent_shard_requests` — max concurrent shard requests per node (>= 1)
- `search.batched_reduce_size` — number of shard results to reduce at once (>= 2)
- `override_request_values` — boolean (default `false`) controlling whether WLM settings override explicitly set request-level values

### Why `override_request_values` is needed

OpenSearch convention is that nothing should override request-level parameters — they represent the caller's explicit intent. However, WLM groups are configured by system administrators to protect the cluster. There should be a mechanism for WLM group settings to take precedence over a potentially malicious or uninformed individual who sets aggressive request parameters (e.g., unbounded timeouts or excessive concurrency).

At the same time, WLM group settings should not immediately override request values by default because that can cause unpredictable response behavior which breaks existing clients. Therefore `override_request_values` defaults to `false` — admins must explicitly opt in to the override behavior after understanding the impact on their workloads.

### Settings update semantics

Settings use merge semantics on update:

- **Settings key absent from request** — existing settings preserved
- **`"settings": {}`** or **`"settings": null`** — clears all search settings and sets `override_request_values` to false
- **`"settings": {"key": "value"}`** — merges with existing (adds/updates the key)
- **`"settings": {"key": null}`** — removes that specific key from existing settings

The `override_request_values` field is always present in the GET response (defaults to `"false"` when not explicitly set).

### Related Issues
Part of #20555 

### Check List

### DIFF
--- a/plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
+++ b/plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
@@ -148,14 +148,17 @@ public class WorkloadManagementRestIT extends OpenSearchRestTestCase {
     }
 
     public void testSearchSettings() throws Exception {
-        // Create with settings
+        // Create with all search settings
         String createJson = """
             {
                 "name": "search_test",
                 "resiliency_mode": "enforced",
                 "resource_limits": {"cpu": 0.3, "memory": 0.3},
                 "settings": {
-                    "search.default_search_timeout": "30s"
+                    "search.default_search_timeout": "30s",
+                    "search.cancel_after_time_interval": "1m",
+                    "search.max_concurrent_shard_requests": "5",
+                    "search.batched_reduce_size": "512"
                 }
             }""";
         Response response = performOperation("PUT", "_wlm/workload_group", createJson);
@@ -165,13 +168,20 @@ public class WorkloadManagementRestIT extends OpenSearchRestTestCase {
         Response getResponse = performOperation("GET", "_wlm/workload_group/search_test", null);
         String responseBody = EntityUtils.toString(getResponse.getEntity());
         assertTrue(responseBody.contains("\"settings\""));
+        assertTrue(responseBody.contains("\"override_request_values\":\"false\""));
         assertTrue(responseBody.contains("\"search.default_search_timeout\":\"30s\""));
+        assertTrue(responseBody.contains("\"search.cancel_after_time_interval\":\"1m\""));
+        assertTrue(responseBody.contains("\"search.max_concurrent_shard_requests\":\"5\""));
+        assertTrue(responseBody.contains("\"search.batched_reduce_size\":\"512\""));
 
-        // Update settings
+        // Update search settings
         String updateJson = """
             {
                 "settings": {
-                    "search.default_search_timeout": "1m"
+                    "search.default_search_timeout": "1m",
+                    "search.cancel_after_time_interval": "5m",
+                    "search.max_concurrent_shard_requests": "10",
+                    "search.batched_reduce_size": "256"
                 }
             }""";
         Response updateResponse = performOperation("PUT", "_wlm/workload_group/search_test", updateJson);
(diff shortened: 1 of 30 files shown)


plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
```diff
@@ -181,10 +191,137 @@ public void testSearchSettings() throws Exception {
...
         String responseBody2 = EntityUtils.toString(getResponse2.getEntity());
         assertTrue(responseBody2.contains("\"search.default_search_timeout\":\"1m\""));
+        assertTrue(responseBody2.contains("\"search.cancel_after_time_interval\":\"5m\""));
+        assertTrue(responseBody2.contains("\"search.max_concurrent_shard_requests\":\"10\""));
+        assertTrue(responseBody2.contains("\"search.batched_reduce_size\":\"256\""));
 
         performOperation("DELETE", "_wlm/workload_group/search_test", null);
     }
 
+    public void testSearchSettingsOverrideRequestValues() throws Exception {
```
cwperks, line 201: How exactly is this test provide that override is working?

plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
```diff
@@ -181,10 +191,137 @@ public void testSearchSettings() throws Exception {
...
+        String invalidBatchJson = """
+            {
+                "name": "invalid_test",
+                "resiliency_mode": "enforced",
+                "resource_limits": {"cpu": 0.3, "memory": 0.3},
+                "settings": {
+                    "search.batched_reduce_size": "1"
+                }
+            }""";
+        assertThrows(ResponseException.class, () -> performOperation("PUT", "_wlm/workload_group", invalidBatchJson));
```
cwperks: Can we check the message here too?

server/src/main/java/org/opensearch/cluster/metadata/WorkloadGroup.java
```diff
@@ -75,11 +76,21 @@ public WorkloadGroup(String name, String _id, MutableWorkloadGroupFragment mutab
         }
 
         // Normalize null settings to empty Settings for storage
-        if (mutableWorkloadGroupFragment.getSettings() == null) {
+        // Ensure override_request_values is always present
+        Settings existingSettings = mutableWorkloadGroupFragment.getSettings();
+        if (existingSettings == null
+            || existingSettings.hasValue(WorkloadGroupSearchSettings.WLM_OVERRIDE_REQUEST_VALUES.getKey()) == false) {
```
cwperks: Why must this be present exactly? Can't we treat absence as default like the normal cluster and index settings APIs?

plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
cwperks: oh I see its testing that it can be toggled to true.

cwperks (commented):
Thank you @dzane17. Overall the change looks good to me, but is it possible to add any integ tests to demonstrate that the overrides are taking when a request is explicitly given values but maps to a workload group that specifies to override?

idk we can can do anything contrived and cancel a request after a few seconds even if the timeout passed to a request is 1m?

plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
dzane17: Yes this test was a bit light. I have added more settings and assertions to ensure the full path executes without error.

plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
dzane17: Added

server/src/main/java/org/opensearch/cluster/metadata/WorkloadGroup.java
dzane17: Good point, the setting should only appear if the user has made an explicit definition

dzane17:
@cwperks A true end-to-end functional test is tricky.

The integ test `testSearchSettingsOverrideRequestValues` now covers the full request flow but does not assert the WLM settings are actually injected. For the timeout example, doing so would require controlling request duration deterministically, which isn't really possible — CI machine variability plus the fact that painless scripts can't use `Thread.sleep` for security reasons means we'd need a test-only plugin exposing a blocking query type just to make one assertion reliable.

Override behavior is already thoroughly covered in unit tests `WorkloadGroupRequestOperationListenerTests` with both positive and negative cases per setting. Each test creates a real `SearchRequest`, invokes the real listener, and asserts on the post-listener state of the request. That tells us WLM properly injects the setting values into the `SearchRequest` object. From that point it's just a matter of OpenSearch honoring attributes on the request, which is an existing workflow.

cwperks (approved):

PR #21434, opened 2026-04-29
[Backport 3.5] [Bugfix] Remove X-Request-Id format restrictions and make size configurable

finnegancarroll:
### Description
Backport https://github.com/opensearch-project/OpenSearch/commit/80ce21cd13ec5ce10b64a0d0a7967685a8532243 from https://github.com/opensearch-project/OpenSearch/pull/21048.

### Check List
- [x] Functionality includes testing.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21529, opened 2026-05-07
Fix Detect Breaking Changes version selection logic for previous released version

cwperks:
### Description

Currently the logic for Detect Breaking Changes simply selects the current released version of OpenSearch as the version to test against for breaking changes. 

This PR updates the logic to be more intelligent and selects the last minor version.

This will resolve issues like https://github.com/opensearch-project/OpenSearch/pull/21434#issuecomment-4391886832 where backports against the 3.5 branch are choosing the wrong version (i.e. 3.6 instead of 3.4) to test against with the Detect Breaking Changes CI check.

### Related Issues

Fixes issues as seen in https://github.com/opensearch-project/OpenSearch/pull/21434#issuecomment-4391886832

### Check List


server/build.gradle
```diff
@@ -186,18 +186,15 @@ tasks.named('forbiddenApisInternalClusterTest').configure { forbidSleep() }
 
 // Set to current version by default
 def japicmpCompareTarget = System.getProperty("japicmp.compare.version")
-if (japicmpCompareTarget == null) { /* use latest released version */
+if (japicmpCompareTarget == null) { /* use latest released version from the previous minor line */
     // Read the list from maven central.
-    // Fetch the metadata and parse the xml into Version instances, pick the latest one
+    // Fetch the metadata and parse the xml into Version instances, pick the latest previous minor.
     japicmpCompareTarget = new URL('https://repo1.maven.org/maven2/org/opensearch/opensearch/maven-metadata.xml').openStream().withStream { s ->
-        new XmlParser().parse(s)
```
reta, line 193: I think we could just filter up to last version that is before the current one, wdyt?

```
        new XmlParser().parse(s)
           .versioning.versions.version
           .collect { it.text() }.findAll { it ==~ /\d+\.\d+\.\d+/ }
           .collect { org.opensearch.gradle.Version.fromString(it) }
           .filter { it.onOrBefore(org.opensearch.gradle.VersionProperties.getOpenSearch()) }
           .toSorted()
           .last()
           .toString()
```

cwperks: Would that work in all cases? i.e. for 3.5.1 should we be testing against 3.4.0 or 3.5.0? This would have it testing with 3.4.0.

reta: > i.e. for 3.5.1 should we be testing against 3.4.0 or 3.5.0?

I think it should work, for 3.5.1, it should be 3.5.0.

cwperks: ok so the logic _should_ be as follows?

- Skip Detect Breaking Changes on initial major release (i.e. when main is bumped to 4.0.0, skip entirely until either 4.1.0 or 4.0.1 is next release)
- For subsequent versions take the immediate released prior version

reta: Yeah, I think that sounds right, major versions may do breaking changes

reta: The VersionSelection  is looking unnecessary (to me), we could do this check in place (since build tools are shareable, adding some ad-hoc classes could be avoided)

reta (approved):
Thanks @cwperks !

server/build.gradle
cwperks: Agreed. Kept it all in build.gradle now.

PR #21530, opened 2026-05-07
Feature/warm cache spi and stats and wiring foyer with tiered object store

vishwasgarg18:
---

### Description

This PR introduces two related pieces of warm-node infrastructure:

**1. NodeCacheOrchestrator** — the single authority for all SSD caches on a warm node. Manages lifecycle, validates the SSD budget across the file cache and any registered block-cache plugins, and aggregates stats for `_nodes/stats`.

**2. Foyer block cache wired into the warm-node read path** — `TieredObjectStore` now consults the Foyer block cache on every byte-range read (`get_range` / `get_ranges`), populates it on remote-store misses, and evicts entries deterministically on shard and index deletion. Before this PR, Foyer was initialized and allocated SSD capacity but never used for actual reads.

**What changed:**

*Core abstractions (`server/`)*
- `NodeCacheOrchestrator` — validates budget, manages lifecycle, implements `BlockCacheRegistry`, aggregates stats across FileCache and block caches
- `NodeCacheOrchestratorCleaner` — `IndexStoreListener` that evicts both FileCache and BlockCache entries on shard/index deletion
- `BlockCache` — added `evictPrefix(String)` and `nativeCacheHandle()` (returns `NativeCacheHandle` — a borrowed, non-owning raw pointer to native memory)
- `NativeCacheHandle` — borrowed, non-closeable handle to a native block-cache instance; prevents callers from accidentally freeing node-scoped memory
- `BlockCacheRegistry` + `BuiltInBlockCaches` — allows format handlers to resolve the preferred cache by name at shard open
- `BlockCacheProvider` — added `cacheName()` for named map registration; `List<BlockCacheProvider>` → `Map<String, BlockCacheProvider>` with fail-fast duplicate detection at startup

*Foyer plugin (`sandbox/plugins/block-cache-foyer/`)*
- `FoyerBlockCache.evictPrefix()` → `FoyerBridge.evictPrefix()` → `foyer_evict_prefix` FFM export
- `foyer/ffm.rs` — `foyer_evict_prefix` export; `foyer_create_cache` returns `Box<Arc<dyn BlockCache>>` fat pointer directly

*Tiered storage library (`sandbox/libs/tiered-storage/`)*
- `TieredObjectStore` — cache-first routing in `get_ranges` (probe → fetch misses → populate), `delete_stream` evicts cache on deletion, `evict_path()` called from `ts_remove_file`
- `ffm.rs` — `ts_create_tiered_object_store` accepts `cache_box_ptr` and wires cache via `with_cache()`

*Parquet format handler (`sandbox/plugins/parquet-data-format/`)*
- `ParquetDataFormatStoreHandler` — resolves `BuiltInBlockCaches.FOYER` from `BlockCacheRegistry` and passes the handle to `TieredStorageBridge.createTieredObjectStore`

**Stats** — `GET _nodes/stats aggregate_file_cache` now includes block cache counters merged into totals (`hit_count`, `evictions_in_bytes`, `used_in_bytes`, `total_in_bytes`).

**TODO (follow-up PRs)**
- Separate `_nodes/stats` sections for file cache vs block cache
- `key_index` persistence across restarts (currently evict_prefix is best-effort after restart)

### Tests

- `NodeCacheOrchestratorCleanerTests`, `NodeCacheOrchestratorTests`
- `tiered_object_store_tests.rs` — 19 cache-specific tests
- `foyer/tests.rs`, `ffm_tests.rs`, `range_cache.rs` — 55 Rust unit tests pass
- `ParquetDataFormatStoreHandlerTests`

### Related Issues
Resolves #[issue]

### Check List
- [ ] API changes companion PR created, if applicable.

Signed-off-by: Vishwas Garg <vishwasgarg18@gmail.com>


sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBlockCacheSettings.java
```diff
@@ -0,0 +1,143 @@
...
+     * <p>Default: {@code 25%}. Set to {@code 0%} to disable the block cache.
+     * Accepts a percentage (e.g. {@code 25%}) or a ratio (e.g. {@code 0.25}).
+     *
+     * <p>Configure in {@code opensearch.yml}:
+     * <pre>{@code
+     * block_cache.size: 25%
+     * }</pre>
+     */
+    public static final Setting<String> CACHE_SIZE_SETTING = new Setting<>(
+        "block_cache.size",
```
vishwasgarg18: Rename.

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/ffm.rs
mgodwan, line 68: foyer_create_cache is duplicated. How is the resolution done in these cases for symbol lookup?

mgodwan:
Compilation is failing for the changes. could you please check?

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBlockCacheSettings.java
vishwasgarg18: done

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/ffm.rs
vishwasgarg18: Sorry, didn't get this fully.

Rust rejects duplicate `#[no_mangle]` symbols at compile time(`error: symbol already defined`). Is guaranteed that  exactly one `foyer_create_cache` in its symbol table.

vishwasgarg18:
> Compilation is failing for the changes. could you please check?

Yes, updated. Build is successful now.

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerAggregatedStats.java
```diff
@@ -0,0 +1,96 @@
...
+        MISS_BYTES,
+        EVICTION_COUNT,
+        EVICTION_BYTES,
+        USED_BYTES;
+
+        static final int COUNT = values().length;
+    }
+
+    /** Buffer size for {@link FoyerBridge#snapshotStats}: 2 sections × Field.COUNT values. */
+    static final int STATS_BUFFER_SIZE = Field.COUNT * 2;
```
nishchay21, line 49: where is this used?

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBridge.java
```diff
@@ -119,5 +135,49 @@ public static void destroyCache(long ptr) {
...
+     * {@code FoyerAggregatedStats.Field}: HIT_COUNT, HIT_BYTES, MISS_COUNT,
+     * MISS_BYTES, EVICTION_COUNT, EVICTION_BYTES, USED_BYTES.
+     *
+     * <p>The buffer size (14 = 7 fields × 2 sections) must stay in sync with
+     * the Rust {@code foyer_snapshot_stats} implementation.
+     */
+    public static long[] snapshotStats(long ptr) {
+        // 7 counters per section (must match FoyerAggregatedStats.Field ordinals)
+        // × 2 sections (overall + block_level) = 14 longs total.
+        final int bufferSize = 14;
```
nishchay21: Use the constant

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerAggregatedStats.java
vishwasgarg18: Used below `final int bufferSize = 14;` . Resolving.

sandbox/plugins/block-cache-foyer/src/main/java/org/opensearch/blockcache/foyer/FoyerBridge.java
vishwasgarg18: Done

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
```diff
@@ -296,16 +312,97 @@ impl ObjectStore for TieredObjectStore {
         result
     }
 
-    /// Delete stream: remove each path from registry only, NO local delete.
-    /// Local file deletion is handled by the Java layer.
+    /// Multi-range read with cache-first routing.
+    /// Probes cache per range; fetches only misses; populates cache on success.
+    async fn get_ranges(&self, location: &Path, ranges: &[Range<u64>]) -> OsResult<Vec<Bytes>> {
```
vishwasgarg18, line 427: We have overriden `get_ranges`as that is the only function called to get the data bytes


How the flow would work for retrieval.

**Complete picture:**

| Call | Route | Foyer? |
|------|-------|--------|
| `store.head()` (file size) | `TieredObjectStore::get_opts` head fast-path | No |
| `ArrowReaderMetadata::load_async` (footer) | `ParquetObjectReader::get_opts` → `TieredObjectStore::get_opts` | No |
| `CachedMetadataReader.get_metadata()` | Returns `Arc<ParquetMetaData>` from memory | N/A (in-memory) |
| `CachedMetadataReader.get_bytes(range)` | `get_range` → `get_ranges` → Foyer | ✅ |
| `CachedMetadataReader.get_byte_ranges(ranges)` | `get_ranges` → Foyer | ✅ |

The footer (~few KB) is read once via `get_opts` (bypasses Foyer), then kept in memory. All column data reads go through Foyer. To also cache the footer in Foyer, we'd need to intercept `get_opts` for footer-range reads in `TieredObjectStore` — a future optimization, not needed for correctness.

sandbox/libs/tiered-storage/src/main/rust/src/ffm.rs
```diff
@@ -117,7 +132,7 @@ pub extern "C" fn ts_get_object_store_box_ptr(tiered_store_ptr: i64) -> i64 {
     // Coerce to trait object and box it
     let boxed: Box<Arc<dyn ObjectStore>> = Box::new(arc as Arc<dyn ObjectStore>);
     let ptr = Box::into_raw(boxed) as i64;
-    native_bridge_common::log_info!("ffm: ts_get_object_store_box_ptr input={}, output={}", tiered_store_ptr, ptr);
+    native_bridge_common::log_info!("ffm: ts_get_object_store_box_ptr: ok");
```
nishchay21, line 135: lets have correct logs

sandbox/libs/tiered-storage/src/main/rust/src/ffm.rs
```diff
@@ -96,8 +102,17 @@ pub extern "C" fn ts_create_tiered_object_store(
         store.set_remote(remote_arc);
     }
 
-    let ptr = Arc::into_raw(store) as i64;
```
nishchay21, line 99: why we change thie and create mut?

sandbox/libs/tiered-storage/src/main/rust/src/ffm.rs
```diff
@@ -86,7 +92,7 @@ pub extern "C" fn ts_create_tiered_object_store(
         Arc::new(object_store::local::LocalFileSystem::new())
     };
 
-    let store = Arc::new(TieredObjectStore::new(file_registry, local));
+    let mut store = TieredObjectStore::new(file_registry, local);
```
nishchay21, line 95: why is this required ?

sandbox/libs/tiered-storage/src/main/rust/src/ffm.rs
```diff
@@ -130,7 +145,7 @@ pub extern "C" fn ts_destroy_object_store_box_ptr(ptr: i64) -> i64 {
         return Err("ts_destroy_object_store_box_ptr: null pointer (0)".to_string());
     }
     let _boxed = unsafe { Box::from_raw(ptr as *mut Arc<dyn ObjectStore>) };
-    native_bridge_common::log_info!("ffm: ts_destroy_object_store_box_ptr ptr={}", ptr);
+    native_bridge_common::log_info!("ffm: ts_destroy_object_store_box_ptr: ok");
```
nishchay21, line 148: same as above

server/src/main/java/org/opensearch/plugins/NativeCacheHandle.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * <p>Callers must not attempt to free the pointer. The handle is valid for as
+ * long as the owning {@link BlockCache} is alive.
+ *
+ * <p>An empty sentinel ({@link #EMPTY}) is provided for the no-cache case.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public final class NativeCacheHandle {
```
nishchay21: Can we see how NtiveStoreHandle is and use same where it has set and extra checks

server/src/main/java/org/opensearch/plugins/BlockCache.java
```diff
@@ -52,4 +52,52 @@ public interface BlockCache extends Closeable {
...
+     * is alive. Callers that hold a reference past that lifetime will have a
+     * dangling pointer.
+     *
+     * <p>Pure-Java implementations return {@link NativeCacheHandle#EMPTY}.
+     * Native storage components check {@link NativeCacheHandle#isLive()} and
+     * fall back to uncached behaviour when the handle is empty.
+     *
+     * @return a borrowed handle to the native cache; never {@code null}
+     */
+    default NativeCacheHandle nativeCacheHandle() {
```
nishchay21: BlockCacheHandle rename

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -2550,10 +2503,12 @@ private static String validateFileCacheSize(String capacityRaw) {
...
     /**
-     * Returns the {@link FileCache} instance for remote warm node
+     * Returns the {@link FileCache} instance for remote warm nodes, or {@code null} on non-warm nodes.
+     * Delegates to {@link NodeCacheOrchestrator} which owns the FileCache lifecycle.
      * Note: Visible for testing
      */
+    @Nullable
     public FileCache fileCache() {
-        return this.fileCache;
+        return nodeCacheOrchestrator != null ? nodeCacheOrchestrator.fileCache() : null;
```
nishchay21, line 2523: Can we create a NoOp cache instead of passing null?

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -817,7 +815,14 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
                 settingsModule.getClusterSettings()
             );
 
-            initializeFileCache(settings);
+            // Block-cache providers discovered once, reused for budget, registration, and capacity.
+            final List<org.opensearch.plugins.BlockCacheProvider> blockCacheProviders = pluginsService.filterPlugins(
```
nishchay21: We should have a map if name , BlockCacheProvider map

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
nishchay21: Do object store still have get_ranges and get_range is get_opts not handling this?

nishchay21: For me get_opts was helping can we not use the same method?

sandbox/libs/tiered-storage/src/main/rust/src/ffm.rs
vishwasgarg18: We shouldn't log ptr address. Removed it intentionally.

sandbox/libs/tiered-storage/src/main/rust/src/ffm.rs
vishwasgarg18: One wrapped in Arc, I can't attach cache. Post associating cache with the object, wrapping in Arc again.

sandbox/libs/tiered-storage/src/main/rust/src/ffm.rs
vishwasgarg18: We shouldn't log ptr address. Removed it intentionally.

server/src/main/java/org/opensearch/plugins/NativeCacheHandle.java
vishwasgarg18: The name was too coupled to store. Hence created a different one.

server/src/main/java/org/opensearch/plugins/BlockCache.java
vishwasgarg18: the `Native` prefix signals that this holds a raw pointer into Rust heap memory, not a Java object. I'd suggest to keep it the same.

server/src/main/java/org/opensearch/node/Node.java
vishwasgarg18: I agree but for now keeping it the same as the existing implementation.

Created an asana task. Will update once the P0s are done.

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
vishwasgarg18: We have to override get_ranges.

> `get_opts` returns `GetResult` which wraps a lazy `BoxStream<Bytes>`, caching a stream is not possible without consuming it, which would break the response contract for the caller. Additionally, without a range hint, `get_opts` reads the entire file, making it too coarse-grained to be useful as a byte-range cache.
>
> `get_ranges` returns `Vec<Bytes>` — each range is fully materialized, making it trivially storable in Foyer keyed by `(path, start, end)`. More importantly, this is the actual hot path: DataFusion's `CachedMetadataReaderFactory` calls `get_byte_ranges` for column-chunk reads, which delegates directly to `get_ranges`. Intercepting here gives us per-column-chunk granularity, meaning a cache miss on one column does not evict others.
>
> The Parquet footer read (one-time per file) goes through `get_opts` → `ParquetObjectReader`, and is already cached in-memory by DataFusion's `CachedMetadataReaderFactory`. There is no benefit to also caching it in Foyer.

sandbox/libs/tiered-storage/src/main/rust/src/ffm.rs
vishwasgarg18: One wrapped in Arc, I can't attach cache. Post associating cache with the object, wrapping in Arc again.

server/src/main/java/org/opensearch/node/Node.java
vishwasgarg18: Let me know if you think otherwise

server/src/main/java/org/opensearch/node/Node.java
vishwasgarg18: Agreed, updating

server/src/main/java/org/opensearch/plugins/NativeCacheHandle.java
nishchay21: I am talking in terms of functionality handling

sandbox/libs/tiered-storage/src/main/rust/src/tiered_object_store.rs
nishchay21: Move to smaller functions

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -233,39 +315,65 @@ impl FoyerCache {
...
 
-impl PageCache for FoyerCache {
-    async fn get(&self, key: &CacheKey) -> Option<Bytes> {
+impl BlockCache for FoyerCache {
+    fn as_any(&self) -> &dyn std::any::Any { self }
+    fn get<'a>(&'a self, key: &'a CacheKey)
+        -> std::pin::Pin<Box<dyn std::future::Future<Output = Option<Bytes>> + Send + 'a>>
+    {
+        Box::pin(async move {
         match self.inner.get(&key.as_str().to_string()).await {
```
nishchay21, line 319: We should handle error here. If we fail in get lets handle the same

nishchay21: Also Lets add a ref counter that increments and decrements ref count during and after read in registry. We need a clean er which is API based and periodic to clear the cache.

server/src/main/java/org/opensearch/plugins/BuiltInBlockCaches.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public final class BuiltInBlockCaches {
+
+    /** Name of the Foyer-backed disk block cache. */
+    public static final String FOYER = "foyer";
+
+    private BuiltInBlockCaches() {}
```
nishchay21: why is this required?

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -817,7 +815,25 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
                 settingsModule.getClusterSettings()
             );
 
-            initializeFileCache(settings);
```
nishchay21, line 820: This is initialised on hot as well ? will removing this be breaking

nishchay21: lets keep filecache separate and orchestrator to take as input

vishwasgarg18: > This is initialised on hot as well ? 

Nope

```
private void initializeFileCache(Settings settings) throws IOException {
        if (DiscoveryNode.isWarmNode(settings) == false) {
            return;
        }
```

vishwasgarg18: It's the same behaviour as hot now as well

nishchay21 (approved):

mgodwan (approved):

PR #21540, opened 2026-05-07
Scope jdk.incubator.vector module to libs:common tests

andrross:
The root build.gradle adds --add-modules=jdk.incubator.vector to every test JVM in the build, which causes the JVM to emit a "WARNING: Using incubator modules: jdk.incubator.vector" line for each forked test worker.

The only production code that imports jdk.incubator.vector is BtreeSearcher in libs/common, gated behind the
`simd.rounding.experimental` feature flag. This change moves the --add-modules flag to the libs:common test task and drops it from the global test JVM args so the warning is only emitted where the module is actually used.

### Check List


reta:
> The only production code that imports jdk.incubator.vector is BtreeSearcher in libs/common, gated behind the
simd.rounding.experimental feature flag. This change moves the --add-modules flag to the libs:common test task and drops it from the global test JVM args so the warning is only emitted where the module is actually used.

@andrross I think this is not strictly correct, Apache Lucene heavily rely on "jdk.incubator.vector" as well [1], so it would be good to have it enabled for all tests, wdyt?

[1] https://github.com/apache/lucene/blob/main/lucene/core/src/java25/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java

andrross:
> > The only production code that imports jdk.incubator.vector is BtreeSearcher in libs/common, gated behind the
> > simd.rounding.experimental feature flag. This change moves the --add-modules flag to the libs:common test task and drops it from the global test JVM args so the warning is only emitted where the module is actually used.
> 
> @andrross I think this is not strictly correct, Apache Lucene heavily rely on "jdk.incubator.vector" as well [1], so it would be good to have it enabled for all tests, wdyt?
> 
> [1] [apache/lucene@`main`/lucene/core/src/java25/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java](https://github.com/apache/lucene/blob/main/lucene/core/src/java25/org/apache/lucene/internal/vectorization/PanamaVectorConstants.java)

@reta Yeah, you're right. I'll discard this.

PR #21542, opened 2026-05-07
Remove Unsafe class injection from Java agent

andrross:
The Java agent previously used ClassInjector.UsingUnsafe at premain time to inject AgentPolicy, SubjectInterceptor, and the two stack caller chain extractors into the boot classloader. This relied on sun.misc.Unsafe, which is deprecated for removal on recent JDKs and emits JVM warnings.

This change relocates StackCallerClassChainExtractor, StackCallerProtectionDomainChainExtractor, and SubjectInterceptor from :libs:agent-sm:agent into :libs:agent-sm:bootstrap, under a new org.opensearch.javaagent.bootstrap.internal subpackage to distinguish them from AgentPolicy, which is the shared API between the server and the agent. The bootstrap jar is listed on the agent jar's Boot-Class-Path manifest attribute, so the JVM loads these classes into the boot classloader natively with no runtime injection needed.

Also set -Dnet.bytebuddy.safe=true on test JVMs to disable the Unsafe-based dispatchers inside ByteBuddy's ClassInjector. ByteBuddy falls back to MethodHandles.Lookup / reflection injection, which works on JDK 9+.

### Check List

### DIFF
--- a/libs/agent-sm/bootstrap/src/main/java/org/opensearch/javaagent/bootstrap/internal/package-info.java
+++ b/libs/agent-sm/bootstrap/src/main/java/org/opensearch/javaagent/bootstrap/internal/package-info.java
@@ -1,0 +1,16 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/**
+ * Internal agent support classes that must be loaded by the boot classloader
+ * so that bytecode woven into JDK classes (either inlined ByteBuddy Advice or
+ * MethodDelegation stubs) can resolve them. These classes are implementation
+ * details of the Java agent and are not part of any public API; do not depend
+ * on them from outside {@code :libs:agent-sm:agent}.
+ */
+package org.opensearch.javaagent.bootstrap.internal;
--- a/build.gradle
+++ b/build.gradle
@@ -421,7 +421,11 @@ gradle.projectsEvaluated {
         task.jvmArgs += [
             "--add-modules=jdk.incubator.vector",
             "--add-exports=java.base/com.sun.crypto.provider=ALL-UNNAMED",
-            "--enable-native-access=ALL-UNNAMED"
+            "--enable-native-access=ALL-UNNAMED",
+            // Disable ByteBuddy's Unsafe-based class injection path to avoid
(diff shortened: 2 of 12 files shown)


libs/agent-sm/bootstrap/src/main/java/org/opensearch/javaagent/bootstrap/internal/package-info.java
```diff
@@ -0,0 +1,16 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+/**
+ * Internal agent support classes that must be loaded by the boot classloader
+ * so that bytecode woven into JDK classes (either inlined ByteBuddy Advice or
+ * MethodDelegation stubs) can resolve them. These classes are implementation
+ * details of the Java agent and are not part of any public API; do not depend
+ * on them from outside {@code :libs:agent-sm:agent}.
```
andrross, line 14: @reta I'm curious about your thoughts here. This does somewhat violate the original spirit of the bootstrap jar, which was intended to be the shared API between the server and the agent, delivered via the boot classloader so both sides see the same class. The classes moved here have no need to be used by the server, but they do need to be on the boot classloader. Creating an "internal" package in this jar seems reasonable to me. We could always refactor it out into another "agent-boot-internals" jar in the future if needed, but that seems like overkill at the moment.

reta: Thanks @andrross 

> The classes moved here have no need to be used by the server, but they do need to be on the boot classloader. Creating an "internal" package in this jar seems reasonable to me.

+1, it is certainly reasonable price to pay to get rid of unsafe. However, it looks to me the goal would not be reachable with the work happening around Apache Arrow / Apache DataFusion / etc integration - the Java/JNI wrappers rely very heavily on unsafe and I am not sure we would be able to eliminate that entirely. So I am a bit sceptical the change will deliver ...

andrross: That's fair, but I think it's worthwhile to remove the direct invocation of unsafe in our code, since this refactoring seems simple enough.

reta (approved):

andrross:
@reta About 3000 fewer warnings in the build log! For now anyway...

```
% grep 'Unsafe' pre-agent-update.log | wc -l
    3372
% grep 'Unsafe' post-agent-update.log | wc -l
     231
```

PR #21543, opened 2026-05-07
Wire PPL string scalar functions with analytics-backend-datafusion

tanik98:
### Description
Wires PPL string scalar functions through the analytics-backend-datafusion plugin, enabling native DataFusion execution for string operations.

### Functions added

| PPL Function | Strategy | Return Type |
|---|---|---|
| `ascii(str)` | Direct name match → DataFusion built-in | INTEGER |
| `concat(str1, ..., strN)` | Direct name match → DataFusion built-in | KEYWORD |
| `concat_ws(sep, str1, str2)` | Direct name match → DataFusion built-in | KEYWORD |
| `left(str, n)` | Direct name match → DataFusion built-in | KEYWORD |
| `length(str)` | Name alias → `char_length` | INTEGER |
| `locate(substr, str)` | Adapter (PositionAdapter) | INTEGER |
| `lower(str)` | Direct name match → DataFusion built-in | KEYWORD |
| `ltrim(str)` | Direct name match → DataFusion built-in | KEYWORD |
| `position(substr, str)` | PositionAdapter | INTEGER |
| `reverse(str)` | Direct name match → DataFusion built-in | KEYWORD |
| `right(str, n)` | Direct name match → DataFusion built-in | KEYWORD |
| `rtrim(str)` | Direct name match → DataFusion built-in | KEYWORD |
| `strcmp(str1, str2)` | Adapter (StrcmpFunctionAdapter) | INTEGER |
| `substr(str, start[, len])` | Name alias → `substring` | KEYWORD |
| `tostring(value[, format])` | Adapter (ToStringFunctionAdapter) + Rust UDF | KEYWORD |
| `tonumber(string[, base])` | Adapter (ToNumberFunctionAdapter) + Rust UDF | DOUBLE |
| `trim(str)` | Direct name match → DataFusion built-in (`btrim`) | KEYWORD |
| `upper(str)` | Direct name match → DataFusion built-in | KEYWORD |

### Key changes

- **`ToStringFunctionAdapter`** — Routes 1-arg calls through `CAST(x AS VARCHAR)`, boolean values through an explicit `CASE WHEN ... THEN 'TRUE' ... THEN 'FALSE' END` (DataFusion's CAST produces lowercase), and 2-arg format calls (`hex`/`binary`/`commas`/`duration`/`duration_millis`) through the `tostring` Rust UDF.

- **`ToNumberFunctionAdapter`** — 1-arg calls use SAFE_CAST to DOUBLE (substrait's null-on-failure cast, which DataFusion maps to `try_cast`) so unparseable values like `tonumber('abc')` return NULL instead of hard-erroring in the `simplify_expressions` optimizer. Handles both fractional (`'4598.678'`) and integer input. 2-arg calls route through the `tonumber` Rust UDF for base-N integer parsing (`i64::from_str_radix`); the UDF returns NULL on parse failure.

- **`StrcmpFunctionAdapter`** — The PPL frontend reverses `strcmp` arguments; the adapter swaps them back. The comparison itself is decomposed into a SIMD-vectorized `CASE WHEN lhs IS NULL OR rhs IS NULL THEN NULL WHEN lhs < rhs THEN -1 WHEN lhs = rhs THEN 0 ELSE 1 END` so DataFusion evaluates it through its native arrow-rs compare kernels — no Rust UDF needed.

- **`PositionAdapter`** — PPL's `POSITION(substr IN str)` and `locate(substr, str[, start])` has reversed argument order vs DataFusion's `strpos(str, substr)` and supports an optional `start` position that `strpos` doesn't. The adapter swaps arguments for the 2-arg form; for the 3-arg form it decomposes to `substring(str, start) + strpos(substring(str, start), substr) - 1`, preserving 1-based semantics.

### Related Issues

### Check List

### DIFF
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
@@ -56,6 +56,7 @@ public enum ScalarFunction {
     UPPER(Category.STRING, SqlKind.OTHER_FUNCTION),
     LOWER(Category.STRING, SqlKind.OTHER_FUNCTION),
     TRIM(Category.STRING, SqlKind.TRIM),
+    SUBSTR(Category.STRING, SqlKind.OTHER_FUNCTION),
     SUBSTRING(Category.STRING, SqlKind.OTHER_FUNCTION),
     /**
      * String concatenation. Calcite's {@code SqlStdOperatorTable.CONCAT} is a
@@ -66,9 +67,23 @@ public enum ScalarFunction {
      * rename surfaces as a compile error rather than as a silent string mismatch at runtime.
      */
     CONCAT(Category.STRING, SqlKind.OTHER_FUNCTION, SqlStdOperatorTable.CONCAT),
+    CONCAT_WS(Category.STRING, SqlKind.OTHER_FUNCTION),
     CHAR_LENGTH(Category.STRING, SqlKind.OTHER_FUNCTION),
     REPLACE(Category.STRING, SqlKind.OTHER_FUNCTION),
     REGEXP_REPLACE(Category.STRING, SqlKind.OTHER_FUNCTION),
+    ASCII(Category.STRING, SqlKind.OTHER_FUNCTION),
+    LEFT(Category.STRING, SqlKind.OTHER_FUNCTION),
+    LENGTH(Category.STRING, SqlKind.OTHER_FUNCTION),
+    LOCATE(Category.STRING, SqlKind.OTHER_FUNCTION),
+    POSITION(Category.STRING, SqlKind.POSITION),
+    LTRIM(Category.STRING, SqlKind.OTHER_FUNCTION),
+    RTRIM(Category.STRING, SqlKind.OTHER_FUNCTION),
+    REVERSE(Category.STRING, SqlKind.OTHER_FUNCTION),
+    RIGHT(Category.STRING, SqlKind.OTHER_FUNCTION),
+    TOSTRING(Category.STRING, SqlKind.OTHER_FUNCTION),
+    NUMBER_TO_STRING(Category.STRING, SqlKind.OTHER_FUNCTION), // Alias for TOSTRING
+    TONUMBER(Category.STRING, SqlKind.OTHER_FUNCTION),
+    STRCMP(Category.STRING, SqlKind.OTHER_FUNCTION),
 
     // ── Math ─────────────────────────────────────────────────────────
     PLUS(Category.MATH, SqlKind.PLUS),
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
@@ -111,6 +111,7 @@ pub async unsafe fn create_session_context(
         .build();
 
     let ctx = SessionContext::new_with_state(state);
+    crate::udf::register_all(&ctx);
 
     // Register default ListingTable for parquet scans
     let listing_options = ListingOptions::new(Arc::new(ParquetFormat::new()))
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/udf/mod.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/udf/mod.rs
@@ -113,10 +113,14 @@ pub(crate) fn coerce_args(
 }
 
 pub mod convert_tz;
+pub mod tonumber;
+pub mod tostring;
 
 pub fn register_all(ctx: &SessionContext) {
     convert_tz::register_all(ctx);
-    log::info!("OpenSearch UDF register_all: convert_tz registered");
+    tonumber::register_all(ctx);
+    tostring::register_all(ctx);
+    log::info!("OpenSearch UDF register_all: convert_tz, tonumber, tostring registered");
 }
 
 #[cfg(test)]
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/udf/tonumber.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/udf/tonumber.rs
@@ -1,0 +1,396 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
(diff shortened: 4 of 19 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/io/substrait/isthmus/expression/NameBasedScalarFunctionConverter.java
```diff
@@ -0,0 +1,187 @@
...
+ *       {@link io.substrait.expression.Expression.ScalarFunctionInvocation} directly.</li>
+ *
+ *   <li><b>Wildcard YAML signatures don't direct-match.</b> Even when a {@code Sig} exists,
+ *       {@code attemptMatch} keys its directMap by runtime operand types (e.g. {@code ascii:str}),
+ *       which don't match YAML wildcard keys (e.g. {@code ascii:any1}). The name-keyed fallback
+ *       here sidesteps that: arity is the only structural check, argument types are left for the
+ *       substrait consumer (DataFusion) to validate.</li>
+ * </ol>
+ */
+public class NameBasedScalarFunctionConverter extends ScalarFunctionConverter {
```
mch2: we should be able to get around this by leveraging our adapter pattern to map the names and not have to re-write at substrait time - see  using https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AbstractNameMappingAdapter.java and UnixTimestampAdapter

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/strcmp.rs
```diff
@@ -0,0 +1,298 @@
+/*
```
mch2: I think we can actually get rid of this UDF and decompose in the adapter to regular string comparison.

```java
**
 * Adapts PPL {@code strcmp(a, b)} into a pure Substrait/DataFusion CASE expression,
 * avoiding a native UDF. Result is {@code INTEGER}:
 * <ul>
 *   <li>{@code -1} when {@code a < b}</li>
 *   <li>{@code  0} when {@code a = b}</li>
 *   <li>{@code  1} when {@code a > b}</li>
 *   <li>{@code NULL} when either operand is {@code NULL}</li>
 * </ul>
 *
 * <p>Rewrite:
 * <pre>{@code
 *   strcmp(a, b)
 *     →
 *   CASE
 *     WHEN a IS NULL OR b IS NULL THEN NULL
 *     WHEN a < b THEN -1
 *     WHEN a = b THEN  0
 *     ELSE              1
 *   END
 * }</pre>
 *
 * <p>Why the adapter beats a row-by-row Rust UDF: the {@code <} and {@code =}
 * comparisons between {@code StringArray} operands lower to arrow-rs compute
 * kernels ({@code arrow::compute::lt}, {@code arrow::compute::eq}) which are
 * SIMD-vectorized on x86_64 (AVX2) and arm64 (NEON). The CASE ({@code ifelse})
 * is also an arrow vectorized kernel. A UDF that loops
 * {@code for i in 0..n { str::cmp(...) }} per row is strictly slower — it
 * amortizes FFI over the batch but the inner compare is scalar.
 *
 * <p>PPL's frontend reverses {@code strcmp}'s args vs. user order. This adapter
 * preserves that — we consume {@code original.getOperands()} in the order the
 * planner handed them to us; the resulting {@code a < b} / {@code a = b} maps
 * 1:1 to the user-intended {@code -1 / 0 / 1} convention.
 */
class StrcmpFunctionAdapter implements ScalarFunctionAdapter {

    @Override
    public RexNode adapt(RexCall original, List<FieldStorageInfo> fieldStorage, RelOptCluster cluster) {
        List<RexNode> operands = original.getOperands();
        if (operands.size() != 2) {
            return original;
        }
        RexBuilder rexBuilder = cluster.getRexBuilder();
        RexNode a = operands.get(0);
        RexNode b = operands.get(1);

        RelDataType intType = cluster.getTypeFactory()
            .createTypeWithNullability(cluster.getTypeFactory().createSqlType(SqlTypeName.INTEGER), true);
        RexNode neg1 = rexBuilder.makeExactLiteral(BigDecimal.valueOf(-1), intType);
        RexNode zero = rexBuilder.makeExactLiteral(BigDecimal.ZERO, intType);
        RexNode one = rexBuilder.makeExactLiteral(BigDecimal.ONE, intType);
        RexNode nullLit = rexBuilder.makeNullLiteral(intType);

        // NULL propagation must be explicit — SQL comparators on NULL return NULL, but
        // the CASE below needs to short-circuit them so we don't fall through to the
        // `ELSE 1` branch when either operand is NULL.
        RexNode aIsNull = rexBuilder.makeCall(SqlStdOperatorTable.IS_NULL, a);
        RexNode bIsNull = rexBuilder.makeCall(SqlStdOperatorTable.IS_NULL, b);
        RexNode anyNull = rexBuilder.makeCall(SqlStdOperatorTable.OR, aIsNull, bIsNull);

        RexNode lessThan = rexBuilder.makeCall(SqlStdOperatorTable.LESS_THAN, a, b);
        RexNode equalTo = rexBuilder.makeCall(SqlStdOperatorTable.EQUALS, a, b);

        // CASE WHEN anyNull THEN NULL WHEN a<b THEN -1 WHEN a=b THEN 0 ELSE 1 END
        return rexBuilder.makeCall(
            intType,
            SqlStdOperatorTable.CASE,
            List.of(anyNull, nullLit, lessThan, neg1, equalTo, zero, one)
        );
    }
}
```

mch2:
pls add a test in our qa pkg - similar tests https://github.com/opensearch-project/OpenSearch/tree/main/sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa

sandbox/plugins/analytics-backend-datafusion/src/main/java/io/substrait/isthmus/expression/NameBasedScalarFunctionConverter.java
tanik98: Removed and adapted to Adapter based pattern.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/strcmp.rs
tanik98: Great suggestion, made the changes and removed the UDF.

mch2 (approved):

PR #21547, opened 2026-05-07
Make parquet/datafusion memory settings configurable

gaurav-amz:
## Summary

  Adds configurable memory settings for the DataFusion and parquet sandbox
  plugins,
  replacing hardcoded heap-fraction defaults with non-heap percentages and
  floors,
  and wires the cluster-settings dynamic-update path through to the live native
  runtimes where the underlying APIs allow it.

  Mirrors the `indices.memory.index_buffer_size` / `min_index_buffer_size`
  pattern
  in `IndexingMemoryController` — the existing OpenSearch precedent for non-heap
  percentage memory settings — for the static resolution logic.

  ## Settings

  All eight settings register on the cluster and appear under
  `GET /_cluster/settings?include_defaults=true`. The "main" string settings
  accept either a percentage of non-heap memory
  (`totalPhysicalMemory − configuredMaxHeap`, e.g. `"25%"`) **or** an absolute
  byte
  size (e.g. `"10gb"`). When a percentage is supplied, the resolved value is
  floored by the `.min` companion. Validators reject malformed values
  (`"abc%"`, `"10xb"`) at PUT time.

  | Setting | Default | Dynamic? |
  | --- | --- | --- |
  | `datafusion.memory_pool_limit_bytes` | `15%` | ✅ |
  | `datafusion.memory_pool_limit.min` | `512mb` | ✅ | 
  | `datafusion.spill_memory_limit_bytes` | `50%` | ❌ static (NodeScope) |
  | `datafusion.spill_memory_limit.min` | `1gb` | ❌ static (NodeScope) |
  | `parquet.max_native_allocation` | `10%` | ✅ |
  | `parquet.min_native_allocation` | `512mb` | ✅ | 
  | `parquet.max_native_allocation_ceiling` | `8gb` (`-1` disables) | ✅ | 
  | `parquet.arrow.child_allocation` | `512mb` | ✅ | 

  ### Resolution logic (shared parser)

  if configured ends with "%":
      base = totalPhysicalMemory − configuredMaxHeap   // non-heap
      resolved = base × (percentage / 100)
      resolved = max(resolved, min)                     // floor
      if ceiling != -1: resolved = min(resolved, ceiling)
  else:
      resolved = parseBytes(configured)                 // floor / ceiling
  ignored

  ## How runtime updates flow

  ### DataFusion memory pool

  PUT /_cluster/settings { datafusion.memory_pool_limit_bytes: "40%" }
          │
          ▼
  ClusterSettings update consumer in DataFusionPlugin
          │
          ▼
  re-resolve against current min  →  resolveBytes(rawValue, floor)
          │
          ▼
  DataFusionService.setMemoryPoolLimit(newBytes)
          │
          ▼
  NativeBridge.df_set_memory_pool_limit  (FFM)
          │
          ▼
  DynamicLimitPool.set_limit (Rust, atomic)

  A second listener on `…limit.min` re-resolves the current raw limit against
  the
  new floor and goes through the same path.

  ### Parquet Arrow allocator

  PUT /_cluster/settings { parquet.max_native_allocation: "32mb" }
          │
          ▼
  ArrowBufferPoolRegistry  (one per node, owned by ParquetDataFormatPlugin)
          │     ── caches the four allocation values in volatile fields
          │        because clusterSettings.get(...) inside a listener still
          │        reads the previous lastSettingsApplied
          ▼
  applyToAllPools()  →  iterates a synchronized WeakHashMap of pools
          │
          ▼  (per pool)
  ArrowBufferPool.applyLimits(Settings)
          │
          ├──  RootAllocator.setLimit(newRoot)
          ├──  this.maxChildAllocation = newChild           // volatile
          └──  for each child in rootAllocator.getChildAllocators():
                  child.setLimit(newChild)

  Per-shard pools come and go (`ParquetIndexingEngine` is per-shard); the
  registry
  holds them weakly so closed pools drop out as the GC reclaims them. There is
  no
  `removeSettingsUpdateConsumer` API on `ClusterSettings`, so the listeners are
  registered once at the plugin level rather than per-pool.

  **Lowering a limit below current usage** is a no-op for already-allocated
  memory:
  Arrow's `setLimit` only rejects future allocations until in-flight buffers
  drain. Same model as the DataFusion memory pool. Documented on each setting.

  ## Why spill is static

  I traced the DataFusion `DiskManager` API: `max_temp_directory_size: u64` is a
  plain field, and the only setter (`set_arc_max_temp_directory_size`) requires
  `Arc::get_mut`, which fails the moment any query holds a strong reference to
  the runtime. There is no thread-safe runtime setter upstream. Marking it
  `Dynamic` would silently no-op for any non-trivial cluster, so I kept it
  NodeScope-only and documented why in the setting javadoc and in
  `DataFusionPluginSettingsTests.testSpillMemoryLimit{,Min}IsStatic`. If/when
  upstream offers an atomic setter we can flip it on with a small change.

  Dynamic — change at runtime:

  PUT /_cluster/settings
  {
    "persistent": {
      "datafusion.memory_pool_limit_bytes": "40%",
      "parquet.max_native_allocation": "4gb",
      "parquet.arrow.child_allocation": "512mb"
    }
  }

## Where the current defaults came from

  Setting: datafusion.memory_pool_limit_bytes
  Default: 15%
  Reasoning at the time: Mirrors IndexingMemoryController.INDEX_BUFFER_SIZE
    default (10% of heap there; 15% of non-heap here ≈ similar fraction of total RAM).
  ────────────────────────────────────────
  Setting: datafusion.memory_pool_limit.min
  Default: 512mb
  Reasoning at the time: Matches MIN_INDEX_BUFFER_SIZE_SETTING floor 48mb × ~10× (DataFusion needs more headroom for hash tables / sorts).
  ────────────────────────────────────────
  Setting: datafusion.spill_memory_limit_bytes
  Default: 50%
  Reasoning at the time: DataFusion's own example config uses 0.5 × memory_pool
  for spill. We exposed it as 50% of non-heap (= ~2× the pool default, since spill is additional on-disk budget).
  ────────────────────────────────────────
  Setting: parquet.max_native_allocation
  Default: 10%
  Reasoning at the time: Original hardcoded value before this PR — kept the same to avoid behavior change for existing deployments.
  ────────────────────────────────────────
  Setting: parquet.min_native_allocation
  Default: 512mb
  Reasoning at the time: Same reasoning as DF min.
  ────────────────────────────────────────
  Setting: parquet.max_native_allocation_ceiling
  Default: 8gb
  Reasoning at the time: Reasonable cap for Arrow buffers — past this, you're better off raising the Lucene heap.
  ────────────────────────────────────────
  Setting: parquet.arrow.child_allocation
  Default: 512mb
  Reasoning at the time: Each VSR ≈ 50k rows × ~10kb avg row = 250 MB. One round number above that.

### Description

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
@@ -70,23 +70,17 @@ public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<Data
     /** Default floor for {@link #DATAFUSION_MEMORY_POOL_LIMIT_MIN}. */
     public static final ByteSizeValue DEFAULT_MEMORY_POOL_LIMIT_MIN = new ByteSizeValue(512, ByteSizeUnit.MB);
 
-    /** Default ceiling for {@link #DATAFUSION_MEMORY_POOL_LIMIT_MAX}. */
-    public static final ByteSizeValue DEFAULT_MEMORY_POOL_LIMIT_MAX = new ByteSizeValue(30, ByteSizeUnit.GB);
-
     /** Default value for {@link #DATAFUSION_SPILL_MEMORY_LIMIT}. */
     public static final String DEFAULT_SPILL_MEMORY_LIMIT = "50%";
 
     /** Default floor for {@link #DATAFUSION_SPILL_MEMORY_LIMIT_MIN}. */
     public static final ByteSizeValue DEFAULT_SPILL_MEMORY_LIMIT_MIN = new ByteSizeValue(1, ByteSizeUnit.GB);
 
-    /** Default ceiling for {@link #DATAFUSION_SPILL_MEMORY_LIMIT_MAX}. */
-    public static final ByteSizeValue DEFAULT_SPILL_MEMORY_LIMIT_MAX = new ByteSizeValue(100, ByteSizeUnit.GB);
-
     /**
      * Memory pool limit for the DataFusion runtime. Accepts a percentage of non-heap memory
      * ({@code totalPhysicalMemory - configuredMaxHeap}, e.g. {@code "25%"}) or an absolute byte
-     * size (e.g. {@code "10gb"}). When a percentage is supplied, the resolved value is clamped
-     * by {@link #DATAFUSION_MEMORY_POOL_LIMIT_MIN} / {@link #DATAFUSION_MEMORY_POOL_LIMIT_MAX}.
+     * size (e.g. {@code "10gb"}). When a percentage is supplied, the resolved value is floored
+     * by {@link #DATAFUSION_MEMORY_POOL_LIMIT_MIN}.
      * <p>
      * Dynamic: changes take effect for new allocations only. Existing reservations that exceed
      * the new limit are not reclaimed — they drain naturally as queries complete.
@@ -105,31 +99,25 @@ public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<Data
         DEFAULT_MEMORY_POOL_LIMIT_MIN,
         new ByteSizeValue(0, ByteSizeUnit.BYTES),
         new ByteSizeValue(Long.MAX_VALUE, ByteSizeUnit.BYTES),
-        Setting.Property.NodeScope
-    );
-
-    /**
-     * Ceiling applied when {@link #DATAFUSION_MEMORY_POOL_LIMIT} is a percentage. {@code -1}
-     * disables the ceiling.
-     */
-    public static final Setting<ByteSizeValue> DATAFUSION_MEMORY_POOL_LIMIT_MAX = Setting.byteSizeSetting(
-        "datafusion.memory_pool_limit.max",
-        DEFAULT_MEMORY_POOL_LIMIT_MAX,
-        new ByteSizeValue(-1),
-        new ByteSizeValue(Long.MAX_VALUE, ByteSizeUnit.BYTES),
-        Setting.Property.NodeScope
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
     );
 
     /**
      * Spill memory limit — when exceeded, DataFusion spills to disk. Accepts a percentage of
      * non-heap memory or an absolute byte size. When a percentage is supplied, the resolved value
-     * is clamped by {@link #DATAFUSION_SPILL_MEMORY_LIMIT_MIN} / {@link #DATAFUSION_SPILL_MEMORY_LIMIT_MAX}.
+     * is floored by {@link #DATAFUSION_SPILL_MEMORY_LIMIT_MIN}.
+     * <p>
+     * Dynamic in the cluster-settings sense: PUTs are accepted and persisted, but the spill limit
+     * is fixed in the native runtime at construction time (no setter). The new value takes effect
+     * after the next node restart.
      */
     public static final Setting<String> DATAFUSION_SPILL_MEMORY_LIMIT = Setting.simpleString(
         "datafusion.spill_memory_limit_bytes",
         DEFAULT_SPILL_MEMORY_LIMIT,
         DataFusionPlugin::validateMemorySizeOrPercentage,
-        Setting.Property.NodeScope
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
     );
 
     /** Floor applied when {@link #DATAFUSION_SPILL_MEMORY_LIMIT} is a percentage. */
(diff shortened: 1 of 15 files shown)


himshikhagupta:
Do we have any data around how we came up with default values? Is it based on benchmark/performance comparison with different settings?

Code changes mostly look good.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -59,24 +64,77 @@ public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<Data
...
     /**
-     * Memory pool limit for the DataFusion runtime.
+     * Default value for {@link #DATAFUSION_MEMORY_POOL_LIMIT}. {@code 15%} of non-heap memory
+     * comes from the OFAT sweep on r7g.2xlarge documented in
+     * <a href="https://quip-amazon.com/baxtAdGoq6S9/Mustang-Memory-Tuning-Recommended-Defaults">
+     * Mustang Memory Tuning - Recommended Defaults</a>: doubling 4&nbsp;GiB → 8&nbsp;GiB buys
+     * ~15% lower p95 ClickBench latency, but 8&nbsp;GiB → 16&nbsp;GiB only ~5%. 15% is the
+     * inflection point that scales cleanly across instance sizes.
+     */
+    public static final String DEFAULT_MEMORY_POOL_LIMIT = "15%";
```
bharath-techie: This is way less - we need 50% of RAM to be allocated to datafusion.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -59,24 +64,77 @@ public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<Data
...
+
+    /** Default floor for {@link #DATAFUSION_MEMORY_POOL_LIMIT_MIN}. */
+    public static final ByteSizeValue DEFAULT_MEMORY_POOL_LIMIT_MIN = new ByteSizeValue(512, ByteSizeUnit.MB);
+
+    /** Default value for {@link #DATAFUSION_SPILL_MEMORY_LIMIT}. */
+    public static final String DEFAULT_SPILL_MEMORY_LIMIT = "50%";
```
bharath-techie: Why do we need 2 different settings ? Also whether datafusion spills or not depends on the operator. If we hit memory limit and hit rejection, the operator chooses whether to fail the query or spill and continue with the query.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -59,24 +64,77 @@ public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<Data
...
+
+    /** Default floor for {@link #DATAFUSION_SPILL_MEMORY_LIMIT_MIN}. */
+    public static final ByteSizeValue DEFAULT_SPILL_MEMORY_LIMIT_MIN = new ByteSizeValue(1, ByteSizeUnit.GB);
```
bharath-techie: This also is quite less - we also need to ensure we spill on data volume - lets give a percentage of disk to start with ? For example, if you run clickbench dataset on 4 GB ram, group by URL field takes 32 GB worth of disk.

Its a bit of an extreme case but you have concurrent queries, limited memory - then we will atleast need decent spill space.

bharath-techie:
`datafusion.spill_memory_limit_bytes` - this also has to be configurable @gaurav-amz - we might need to reduce it in case of disk watermark etc and do we need to do any investigation to apply this on spill similar to how this is applied on memory ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
bharath-techie: Can we also please rename this to DISK SPILL LIMIT. Maybe we need to include this as part of watermark calculation ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -60,24 +64,50 @@ public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<Data
...
+     * Static (NodeScope only). DataFusion's {@code DiskManager} stores the spill cap as a plain
+     * {@code u64} and only exposes a setter behind {@code Arc::get_mut}, which fails as soon as
+     * any query holds a strong reference to the runtime. Until upstream offers a thread-safe
+     * setter we treat this as a startup setting; change it in {@code opensearch.yml} and restart.
+     */
+    public static final Setting<String> DATAFUSION_SPILL_MEMORY_LIMIT = Setting.simpleString(
         "datafusion.spill_memory_limit_bytes",
-        Runtime.getRuntime().maxMemory() / 8,
-        0L,
+        DEFAULT_SPILL_MEMORY_LIMIT,
```
bharath-techie: Yeah throughout lets update the java docs and rename to reflect disk spill limit please

gaurav-amz:
> `datafusion.spill_memory_limit_bytes` - this also has to be configurable @gaurav-amz - we might need to reduce it in case of disk watermark etc and do we need to do any investigation to apply this on spill similar to how this is applied on memory ?

The approach is currently under discussion, and the changes will be addressed in a follow-up PR.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/memory/ArrowBufferPool.java
```diff
@@ -63,14 +75,68 @@ public long getTotalAllocatedBytes() {
...
+     * {@code totalPhysicalMemory - configuredMaxHeap}; absolute byte sizes are used as-is.
+     * <p>
+     * Throws {@link IllegalStateException} if a percentage is supplied but non-heap memory is
+     * unmeasurable — operators on misconfigured boxes should specify an absolute byte size.
+     */
+    static long resolveMaxAllocationBytes(Settings settings) {
+        String configured = ParquetSettings.MAX_NATIVE_ALLOCATION.get(settings);
+
+        if (configured.endsWith("%")) {
+            long totalAvailableMemory = OsProbe.getInstance().getTotalPhysicalMemorySize() - JvmInfo.jvmInfo().getConfiguredMaxHeapSize();
```
Bukhtawar, line 127: I don't think we should make this a function of available physical memory, this in most case would be capped by a cgroup. I would prefer taking this as an input from yaml

PR #21553, opened 2026-05-07
Add fix for local recovery from flush

alchemist51:
## Description
When a DataFormatAwareEngine restarts with committed data and no pending translog operations, the CatalogSnapshotManager would restore the committed snapshot but never notify listeners (reader managers). This meant the engine appeared healthy but couldn't serve search requests — the reader managers had no open readers.

Root cause: CatalogSnapshotManager's constructor restored the committed CatalogSnapshot from Lucene commit data but did not call afterRefresh() on registered listeners. Reader managers rely on this callback to open readers for the current snapshot.

Fix: After restoring the committed snapshot in CatalogSnapshotManager's constructor, iterate over all registered CatalogSnapshotLifecycleListeners and invoke afterRefresh(true, latestCatalogSnapshot) so reader managers are initialized immediately on engine open.

 ## Test plan

  - DataFormatAwareEngineRecoveryTests — 14 unit tests covering translog replay, partial recovery, concurrent indexing, double-restart, committed data restoration, and snapshot generation correctness
  - CompositeLocalRecoveryIT — 5 integration tests with full cluster restart using parquet + lucene composite engine
  - LocalRecoveryIT — 2 REST-level E2E tests verifying PPL query results are identical after index close/reopen (including force-merge scenario)

### DIFF
--- a/sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/LocalRecoveryIT.java
+++ b/sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/LocalRecoveryIT.java
@@ -229,7 +229,7 @@ public class LocalRecoveryIT extends AnalyticsRestTestCase {
         client().performRequest(flushRequest);
 
         // Query before restart
-        String ppl = "source=" + indexName + " | sort age | fields name, age, score";
+        String ppl = "source=" + indexName + " | sort age, name | fields name, age, score";
         List<List<Object>> beforeRows = executePplRows(ppl);
         assertEquals("Should have " + totalDocs + " rows before restart", totalDocs, beforeRows.size());
 
--- a/sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/LocalRecoveryIT.java
+++ b/sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/LocalRecoveryIT.java
@@ -143,6 +143,129 @@ public class LocalRecoveryIT extends AnalyticsRestTestCase {
         }
     }
 
+    public void testQueryResultsIdenticalAfterForceMergeAndRestart() throws IOException {
+        String indexName = "local_recovery_forcemerge_test";
+
+        // Delete if exists from a previous run
+        try {
+            client().performRequest(new Request("DELETE", "/" + indexName));
+        } catch (Exception e) {
+            // index may not exist
+        }
+
+        // Create index with parquet primary + lucene secondary
+        String body = "{"
+            + "\"settings\": {"
+            + "  \"number_of_shards\": 1,"
+            + "  \"number_of_replicas\": 0,"
+            + "  \"index.pluggable.dataformat.enabled\": true,"
+            + "  \"index.pluggable.dataformat\": \"composite\","
+            + "  \"index.composite.primary_data_format\": \"parquet\","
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshotManager.java
```diff
@@ -118,6 +118,12 @@ public CatalogSnapshotManager(
             shardPath,
             commitFileManager
         );
+
+        // Notify listeners about the committed snapshot so reader managers
+        // are initialized on engine open.
+        for (CatalogSnapshotLifecycleListener listener : snapshotListeners) {
+            listener.afterRefresh(true, latestCatalogSnapshot);
```
mgodwan, line 125: Can you verify from LuceneReaderManager perspective as well i.e. directory reader is as expected?

server/src/test/java/org/opensearch/index/engine/DataFormatAwareEngineRecoveryTests.java
```diff
@@ -0,0 +1,689 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Recovery tests for {@link DataFormatAwareEngine}.
+ * Mirrors the recovery test patterns from {@code InternalEngineTests} (L1024-1133)
+ * to verify translog replay, checkpoint advancement, and segment creation
+ * after engine close/reopen.
+ */
+public class DataFormatAwareEngineRecoveryTests extends OpenSearchTestCase {
```
mgodwan, line 83: Can we add this in test framework, and add a test in composite Plugin as well which extends this and uses all the actual plugins?

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/LocalRecoveryIT.java
```diff
@@ -0,0 +1,174 @@
...
+        assertEquals(200, openResponse.getStatusLine().getStatusCode());
+
+        // Wait for recovery
+        Request healthRequest = new Request("GET", "/_cluster/health/" + INDEX_NAME);
+        healthRequest.addParameter("wait_for_status", "green");
+        healthRequest.addParameter("timeout", "60s");
+        client().performRequest(healthRequest);
+
+        // Query after restart — must return identical results
+        List<List<Object>> afterRows = executePplRows(ppl);
```
mgodwan, line 128: Did the assertion not fail? due to missing generation stamp https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneReaderManager.java#L130


If not, lets ensure to ingest enough for triggering merges

alchemist51: Added a test specifically to trigger force merge after doing flush couple of time.

server/src/test/java/org/opensearch/index/engine/DataFormatAwareEngineRecoveryTests.java
alchemist51: Have add a AbstractCompositEngineIT for that. I see in your other PR we are adding it for the test framework, skipping it for now.

mgodwan (approved):

Issue #5424, opened 2022-12-01
Deprecate REST client


dblock:
**Is your feature request related to a problem? Please describe.**

In https://github.com/opensearch-project/opensearch-java/issues/262 we are removing the dependency on OpenSearch (this repo). I propose to loudly deprecate REST client code in this repo and require users to migrate to a dependency on opensearch-java. This would significantly reduce the amount of projects that depend on OpenSearch core by all the ones that depend on the REST APIs.

**Describe the solution you'd like**

1. Deprecate a large portion of/all `org.opensearch.client` namespace in 2.x. 
2. Remove it in 3.0.

reta:
@dblock Regarding the suggested plan, it might be a bit aggressive:
 - Deprecate a large portion of/all org.opensearch.client namespace in 2.x
 - Keep it in 3.0.0 but make the respective transport optional in `opensearch-java`  (giving a time to migrate since it could be a short notice, might be a good start for 2.0 but we are at 2.4 already)
 - Remove it in 4.0.0

wbeckler:
I've proposed removing documentation of the high-level rest client from the organization documentation. There hasn't been any feedback on the proposal: https://github.com/opensearch-project/documentation-website/issues/2010

manirudh78:
Hello all,
Hope all is well.
Based on the comments in this ticket
**The Opensearch-Hadoop will support the scala client for opensearch? Yes**

Response is highly appreciated.
Thanks

wbeckler:
@manirudh78 I'm not sure I understand your comment. Is it possibly in the wrong repo?

dblock:
I think it was meant for https://github.com/opensearch-project/opensearch-hadoop. I don't see why that wouldn't work in Scala.

rursprung:
nice @reta!

should we have a follow-up issue to track the deprecation of `RestHighLevelClient`? i couldn't find an issue for that but i see a bunch of option issues which still ask for new features in it. having a deprecation issue for it would allow closing the others as "not planned" and pointing them at [`opensearch-java`](https://github.com/opensearch-project/opensearch-java/)

reta:
> nice [@reta](https://github.com/reta)!
> 
[...]

Issue #5421, opened 2022-12-01
[Proposal] Upgrade Testing Framework UX


chelma:
## Objective

The purpose of this doc is to outline key user experiences for the Upgrade Testing Framework so the design is driven from a customer-first perspective.  The designs represented in the doc are speculative and open to change; think of them as proposals rather than assertions.

## Terminology

* **Application Developer:** individual that uses OpenSearch as a search/logs engine in their system, but doesn't necessarily manage it
* **Cluster Admin:** individual responsible for the maintenance and/or support of one or more ElasticSearch/OpenSearch Clusters
* **Committed OpenSearch Developer:** individual who contributes code to the OpenSearch Project, either its core repositories or the constellation of repositories connected to them
* **Casual OpenSearch Developer:** individual who might be willing to contribute code to the OpenSearch Project, assuming that the process is easy
* **User:** Either a Cluster Admin or a Committed/Casual OpenSearch Developer; an Application Developer might wear any of those hats as well
* **Upgrade:** moving from one version of ElasticSearch/OpenSearch to another without necessarily changing the underlying hardware hosting the cluster
* **Migration:** moving an ElasticSearch/OpenSearch cluster from one underlying hosting solution to another without necessarily changing the version of the cluster

## Background

The Upgrade Testing Framework is a proposed project ([see here](https://github.com/opensearch-project/OpenSearch/issues/5167) (https://github.com/opensearch-project/OpenSearch/issues/5167)) for the OpenSearch Project.  It has the following goals:

* Provide OpenSearch Developers a way to easily test the outcomes of more complex, real-world upgrades than the existing backwards compatibility (BWC) tests allow and ideally replace those existing tests.  It intends to address upgrades to the core engine and plugins, along with accompanying data, configuration and metadata, and handle N+M version upgrades instead of just N+1.
* Perform upgrades during testing in the same manner as Users execute them “in the wild”.
* Provide Cluster Admins mechanisms to predict the outcome of an upgrade before executing one and enable them to make a representative simulacra of their cluster on their laptop or other development machine in order to test out an upgrade in a safe environment.
* Promote and encapsulate validation mechanisms that Cluster Admins can use to determine whether a upgraded candidate cluster is ready for production traffic.
* Serve as a central repo to iteratively capture the community’s full understanding of differences between versions and continuously assert that each difference remains as expected.  This will enable the community to formally understand, document, and ideally resolve these differences.

## Setup

[...]

PR #21554, opened 2026-05-07
[Analytics Backend / DataFusion] Onboard PPL array constructor and 8 multivalue (mv) functions to analytics-engine route

ahkcs:
### Description

Onboards eight PPL collection functions to the analytics-engine route, plus the row-codec / decimal handling fixes, the `concat()` Sig bridge, and a Rust-side fix that's load-bearing for any of these UDFs to actually work end-to-end:

| Function | Calcite operator | Substrait → DataFusion | Adapter / Rust UDF |
|---|---|---|---|
| `array(a, b, …)` | `PPLBuiltinOperators.ARRAY` | `make_array` | `MakeArrayAdapter` (rename + variadic operand widening) |
| `array_length(arr)` | `SqlLibraryOperators.ARRAY_LENGTH` | `array_length` | none |
| `array_slice` / `mvindex(arr, from, to)` | `SqlLibraryOperators.ARRAY_SLICE` | `array_slice` | `ArraySliceAdapter` (BIGINT index coerce + 0-based-`(start, length)` → 1-based-`(start, end)`) |
| `mvdedup` | `SqlLibraryOperators.ARRAY_DISTINCT` | `array_distinct` | none |
| `mvjoin(arr, sep)` | `SqlLibraryOperators.ARRAY_JOIN` | `array_to_string` | `ArrayToStringAdapter` (rename) |
| `mvindex(arr, N)` single-element | `SqlStdOperatorTable.ITEM` | `array_element` | `ArrayElementAdapter` (rename + BIGINT index coerce) |
| `mvzip(left, right [, sep])` | `PPLBuiltinOperators.MVZIP` (UDF) | **custom Rust UDF** `udf::mvzip` | `MvzipAdapter` |
| `mvfind(arr, regex)` | `PPLBuiltinOperators.MVFIND` (UDF) | **custom Rust UDF** `udf::mvfind` | `MvfindAdapter` |
| `mvappend(arg1, arg2, …)` | `PPLBuiltinOperators.MVAPPEND` (UDF) | **custom Rust UDF** `udf::mvappend` | `MvappendAdapter` (uniform-list reshape) |

`ARRAY_RETURNING_PROJECT_OPS` is a separate set from `STANDARD_PROJECT_OPS` so `FieldType.ARRAY` doesn't pollute filter / aggregate capabilities.

### Critical Rust-side fix (`session_context.rs`)

`create_session_context` (the Rust-side builder behind `df_create_session_context`) built a fresh DataFusion `SessionContext` but never called `udf::register_all` on it. Every fragment query routed through `df_execute_with_context` reused that handle's `ctx` via `query_executor::execute_with_context`, so substrait function references to `mvappend` / `mvfind` / `mvzip` / `convert_tz` failed planning with `This feature is not implemented: Unsupported function name`. The matching `register_all` call exists on `execute_query` / `local_executor` / `indexed_executor` — this just brings the FFM session-context path to parity. Without this, none of the custom UDFs in this PR actually worked at runtime.

### Custom Rust UDFs

DataFusion has no stdlib equivalent for `mvzip` (element-wise zip into strings), `mvfind` (regex match → first index), or `mvappend` (mixed array+scalar flatten with null filtering). All three onboard as Rust ScalarUDFs on the analytics-backend-datafusion plugin's session context, mirroring the existing `convert_tz` pattern. Each has unit tests (mvzip 7, mvfind 7, mvappend 6).

### Substrait extension catalog

New `opensearch_array_functions.yaml` declares `make_array`, `array_length`, `array_slice`, `array_distinct`, `array_to_string`, `array_element`, `mvzip`, `mvfind`, `mvappend`. Loaded via `SimpleExtension.load("/opensearch_array_functions.yaml")` and merged in `DataFusionPlugin.loadSubstraitExtensions()`.

### Operand-normalization adapters (5)

* **`MakeArrayAdapter`** — implements `ScalarFunctionAdapter` directly. PPL's `ArrayFunctionImpl` infers `ARRAY<commonElementType>` for the call's return type but does NOT widen individual operand types. So `array(1, 1.5)` produces a RexCall whose operands are `(INTEGER, DECIMAL(2,1))` but whose return type is `ARRAY<DOUBLE>`. Substrait's variadic-`any1` consistency validator throws a fatal `AssertionError` in that case. The adapter extracts the call's component type and CASTs each non-matching operand.
* **`ArrayToStringAdapter`** — declares a local `array_to_string` op and name-maps `SqlLibraryOperators.ARRAY_JOIN` → it.
* **`ArraySliceAdapter`** — two transforms: BIGINT index coercion and `(0-based start, length)` → `(1-based start, 1-based end inclusive)`. Without the latter, `mvindex(arr=[1..5], 1, 3)` returns `[1, 2, 3]` instead of the expected `[2, 3, 4]`.
* **`ArrayElementAdapter`** — renames `SqlStdOperatorTable.ITEM` → `array_element` and coerces the index operand to BIGINT.
* **`MvappendAdapter`** — wraps each scalar operand in a singleton `make_array(scalar)` call so substrait's variadic-`any1` sees a uniform `list<componentType>` shape across all variadic positions.

### Sig bridges

* **`SqlLibraryOperators.CONCAT_FUNCTION` → `concat`** — PPL's `concat()` is a function-form CONCAT (operator name `"CONCAT"`), distinct from `||` (`SqlStdOperatorTable.CONCAT`) which isthmus' default Sig table already binds to substrait `concat`. Without this bridge, `concat(a, b)` flowing into any analytics-engine query (e.g. `mvfind(arr, concat('ban', '.*'))`) fails substrait conversion with `Unable to convert call CONCAT(string, string)`.

### `RowResponseCodec` list support + decimal handling

The row-oriented fragment-execution wire format ships each cell through OpenSearch's `writeGenericValue` / `readGenericValue` (preserves `List` values as `ArrayList<Object>`), then re-materializes them into a `VectorSchemaRoot` on the coordinator. Three bugs ate array values:

1. `inferArrowType` walked rows for the first non-null cell and matched against `{Long, Integer, …, CharSequence, byte[], Number}`. `List` wasn't in the chain → fell through to `Utf8` — every array column became VARCHAR.
2. `setVectorValue` for `VarCharVector` called `value.toString()`. For a `JsonStringArrayList` that returns the JSON form `[2,3,4]`, which then serialized as a JSON string in the final response.
3. The `Number` fallback in `scalarArrowType` matched `BigDecimal` (extends `Number`) before `Double`/`Float` — encoding decimal cells as `Int(64)` and silently truncating fractional digits.

Fixes:
* Replace `inferArrowType` with `inferField` (returns a full `Field`, builds `List<inner>` for list cells).
* Add a `ListVector` arm to `setVectorValue` that writes directly to the list's offset / validity buffers and the inner data vector — bypassing `UnionListWriter`'s tricky per-element `ArrowBuf` lifecycle.
* Promote `BigDecimal` to `FloatingPoint(DOUBLE)` before the `Number` fallback — fixes `testArray` mixed-numeric `array(1, -1.5, 2, 1.0)` returning `[1, -1, 2, 1]`.

Plus a related Arrow gotcha: `ListVector.getObject` for a `VarCharVector` child returns elements typed as `org.apache.arrow.vector.util.Text` (not `String`). `ExprValueUtils.fromObjectValue` rejected those as "unsupported object class". `ArrowValues.toJavaValue` now normalizes `Text → String` for list cells.

### Other plumbing

* **`ArrowSchemaFromCalcite.toArrowField`** — recurses into the component type for ARRAY → builds the matching `List<inner>` field.
* **`commons-text`** to analytics-engine — Calcite's `SqlFunctions.<clinit>` references `LevenshteinDistance`.
* **`jackson-datatype-jsr310`** to **`arrow-flight-rpc`** — `arrow-vector`'s `JsonStringArrayList` eagerly registers `JavaTimeModule` on its ObjectMapper in `<clinit>`.
* **`regex` 1.10** Cargo dep on the analytics-backend-datafusion crate — backs the `udf::mvfind` per-row regex match.

---

## Current Status (empirically verified, post-rebase)

Tested per the [Mustang + SQL plugin SOP](https://github.com/ahkcs/sql/commit/29339c9c0316d5bad43b5bcb13cc91627b54b1d9) — `:integ-test:integTestRemote` against an externally-managed cluster with the full sandbox plugin set (`opensearch-job-scheduler`, `arrow-flight-rpc`, `analytics-engine`, `parquet-data-format`, `composite-engine`, `analytics-backend-lucene`, `analytics-backend-datafusion`, `opensearch-sql-plugin`), `tests.analytics.parquet_indices=true`, `tests.analytics.force_routing=true`, with companion #5421 (merged) + #5424 applied on top of `feature/ppl-coverage-bundle`.

### `CalciteMVAppendFunctionIT` (force-routed)

| Status | Before | After |
|---|---|---|
| Passed | 0/15 | **10/15** |
| Failed | 15/15 | 5/15 |

Newly passing: `testMvappendWith{MultipleElements, SingleElement, ArrayFlattening, MixedArrayAndScalar, StringValues, NestedArrays, RealFields, NumericArrays, ComplexExpression, IntAndDouble}`.

### `CalciteArrayFunctionIT` (force-routed)

| Status | Before | After |
|---|---|---|
| Passed | 1/60 | **43/60** |
| Failed | 59/60 | 17/60 |

Newly passing across this PR's commits:
- `testArray*` (4) — array constructors incl. mixed-numeric decimals
- `testMvjoinWith*` (8) — mvjoin over arrays + empty array
- `testMvindex*` (9) — mvindex single-element + range, with semantic + JSON formatting fixes
- `testMvdedupWith*` (5)
- `testSplitWith*` (3)
- `testMvzip{Basic, WithCustomDelimiter, Nested}` (3)
- `testMvfindWith*` (9) — all 9 mvfind variants

---

## What's left (follow-ups)

The 22 remaining failures (5 in MVAppend + 17 in ArrayFunction) split into clear buckets — none are regressions, all are pre-existing limits or out-of-scope work:

| # | Bucket | Tests | Status | Owner / next step |
|---|---|---|---|---|
| 1 | **Lambda functions** (`transform`, `mvmap`, `reduce`, `forall`, `exists`, `filter`) | 16 in `CalciteArrayFunctionIT` (`testTransform*`, `testMvmap*`, `testReduce*`, `testForAll`, `testExists`, `testFilter`) | **Out of scope** | Substrait extension YAML doesn't support declaring `func<…>` lambda-typed arguments. Confirmed: capability registration alone gets past the planner, isthmus emits `Expression.Lambda` correctly, but no Sig binding exists for `transform(list<…>, func<… -> …>)`. Multi-PR architectural work — needs either a substrait-spec change or per-shape higher-order Rust UDFs. **Track separately.** |
| 2 | **Genuinely heterogeneous `mvappend`** (no common element type) | 3 in `CalciteMVAppendFunctionIT`: `testMvappendWith{MixedTypes, FieldsAndLiterals, Null}` | **Architectural limit** | Calcite legitimately widens to `ARRAY<ANY>` when `leastRestrictive` returns null (INT+VARCHAR, etc.). Substrait can't encode `ANY`; Arrow's Union arrays exist but `datafusion-functions-array` doesn't operate on them. Resolving requires either changing PPL's mvappend contract from heterogeneous-`Object[]` to a uniform stringified type (breaks user expectations) or shipping Arrow Union arrays through the wire format (multi-PR). **Document as known limit.** |
| 3 | **Empty-`array()` operand carried via column ref** | 1 in MVAppend (`testMvappendWithEmptyArray`); also surfaces as 4 in ArrayFunctionIT (`testMvjoinWithEmptyArray`, `testMvdedupWithEmptyArray`, `testMvzip{WithEmptyArray, WithBothEmptyArrays}`, `testMvfindWithEmptyArray`) — all those 4 actually now pass with #5421 merged on the bundle branch | **Architectural limit (mvappend)** / **passes elsewhere** | `eval empty_arr = array(), result = mvappend(empty_arr, 1, 2)` — at MVAppend's type-inference site we see a `RexInputRef` to a column whose declared type is `ARRAY<VARCHAR>` (#5421 default), not the literal `array()`. Per-call detection at type-inference can't reach back through the project chain. Fixing on the analytics-engine route would need either data-flow analysis or changing `array()`'s element-type default to `NULL`/`UNKNOWN` (which substrait can't serialize either). **Document.** |
| 4 | **Filter predicate on ARRAY field** | 1 in MVAppend (`testMvappendInWhereClause`) | **Filter-rule capability gap** | `OpenSearchFilterRule.resolveViableBackends` extracts only the predicate's TOP operator (EQUALS) and checks it against the field's storage type (ARRAY); it doesn't walk into nested calls, so `array_length(arr) = 2` is rejected as "EQUALS on ARRAY field". Adding `ARRAY_LENGTH` to `STANDARD_FILTER_OPS` doesn't help — the rule's coarse-grained check is the actual blocker. **Tracked separately as a planner refactor; not specific to mvappend.** |
| 5 | **`testMvdedupPreservesOrder`** | 1 in ArrayFunctionIT | **Likely DataFusion stdlib behavior gap** | DataFusion's `array_distinct` likely doesn't preserve insertion order. Needs a custom Rust UDF or upstream contribution. **Track separately.** |
| 6 | **Self-contained QA IT in `sandbox/qa/analytics-engine-rest`** | — | **Open follow-up** | Mirror the SQL-plugin ITs for the functions onboarded here so the analytics-engine sister plugins exercise the same surface independently of the SQL plugin's IT runner. |

### Companion changes

- **opensearch-project/sql#5421** (merged) — defaults empty `array()` element type to `ARRAY<VARCHAR>` in PPL's `ArrayFunctionImpl` so the empty-call return type is substrait-serializable. Required for `testMvjoinWithEmptyArray`, `testMvdedupWithEmptyArray`, and the empty-array mvzip / mvfind variants to pass on the analytics-engine route.

- **opensearch-project/sql#5424** — `MVAppendFunctionImpl.updateMostGeneralType` widens via `RelDataTypeFactory.leastRestrictive` (with DECIMAL → DOUBLE promotion) instead of strict `Object.equals`, plus pre-casts each scalar operand to the call's element Java class in `MVAppendImplementor` so Avatica's `AbstractCursor.ArrayAccessor.DoubleAccessor.getDouble` (`(Double) value`) succeeds. Required for `testMvappendWith{MixedArrayAndScalar, ComplexExpression, IntAndDouble}` to pass on the analytics-engine route.

### Test plan

* [x] Rebuilt analytics-framework, analytics-backend-datafusion (Java + Rust), analytics-engine, arrow-flight-rpc bundles via `./gradlew publishToMavenLocal -Dsandbox.enabled=true`.
* [x] `./gradlew :sandbox:plugins:analytics-backend-datafusion:test -Dsandbox.enabled=true` — green after rebase.
* [x] `cargo test --lib udf::mvzip` — 7/7. `cargo test --lib udf::mvfind` — 7/7. `cargo test --lib udf::mvappend` — 6/6.
* [x] Started cluster per [Mustang + SQL plugin SOP](https://github.com/ahkcs/sql/commit/29339c9c0316d5bad43b5bcb13cc91627b54b1d9): full plugin set + `-Dtests.jvm.argline="-Djava.library.path=… -Dopensearch.experimental.feature.pluggable.dataformat.enabled=true"`. Verified all 8 plugins picked up the rebuilt jars and the rebuilt native dylib.
* [x] `:integ-test:integTestRemote --tests "org.opensearch.sql.calcite.remote.CalciteMVAppendFunctionIT"` against running cluster with `tests.analytics.{parquet_indices,force_routing}=true` — **10/15 passing** (was 0/15).
* [x] `:integ-test:integTestRemote --tests "org.opensearch.sql.calcite.remote.CalciteArrayFunctionIT"` — **43/60 passing** (was 1/60).
* [x] Verified `mvindex(arr=[1..5], 1, 3)` produces `[2,3,4]` (was `[1,2,3]`); negative `mvindex(arr, -3, -1)` produces `[3,4,5]`.
* [x] Verified `mvjoin(array(), '-')` returns `""` and `mvdedup(array())` returns `[]` (companion #5421 default).
* [x] Verified `mvzip(['a','b'], ['1','2'])` returns `['a,1', 'b,2']`; `mvfind(['apple','banana','apricot'], 'ban.*')` returns `1`.
* [x] Verified `array(1, -1.5, 2, 1.0)` returns `[1.0, -1.5, 2.0, 1.0]` (was `[1, -1, 2, 1]`).
* [x] Verified `mvfind(arr, concat('ban', '.*'))` returns `1` (was failing on `Unable to convert call CONCAT`).
* [x] Verified `mvappend(arr=array(1,2), 3, 4)` returns `[1, 2, 3, 4]` (was failing with `Unable to convert call mvappend(list<i32?>, i32?, i32?)` until uniform-list reshape).
* [x] Verified `mvappend(1, 2.5)` returns `[1, 2.5]` end-to-end on the analytics-engine route (with #5424 widening + operand pre-cast).
* [ ] Self-contained QA IT in `sandbox/qa/analytics-engine-rest` — **follow-up #6 above**.


mch2:
@ahkcs Thanks Kai, can you pls fix this one:

```
* What went wrong:
Execution failed for task ':sandbox:plugins:analytics-engine:dependencyLicenses'.
> Missing SHA for commons-text-1.11.0.jar. Run "gradle updateSHAs" to create them
```

looks like we need to add this dependency's license files.

ahkcs:
Fixed in aa2fb0a9127. Added the SHA + LICENSE + NOTICE for `commons-text-1.11.0` in analytics-engine, and the SHA for `jackson-datatype-jsr310-2.21.3` in arrow-flight-rpc (the shared `jackson-LICENSE` / `jackson-NOTICE` already cover it via the existing `/jackson-.*/` → `jackson` mapping). Verified all three plugins' `:precommit` passes locally.

mch2 (approved):

PR #21559, opened 2026-05-08
feat(hunspell): Add reload_cached_resources in TokenFilterFactory to support dictionary hot-reload

shayush622:
## Description
  
  Adds cache reload support to enable hot-reload of hunspell dictionaries via the existing _refresh_search_analyzers API without requiring a node restart.
  
  Problem: When hunspell dictionary files are updated on disk, calling _refresh_search_analyzers rebuilds the analyzer factories but reuses the cached Dictionary object from HunspellService's node-level cache. The
  only way to pick up new dictionary content was to restart the node.
  
  Solution: Add a boolean reloadCachedResources parameter to MapperService.reloadSearchAnalyzers(). When true, the method walks each ReloadableCustomAnalyzer's token filters and calls reloadCachedResources() — a
  new default method on TokenFilterFactory. HunspellTokenFilterFactory overrides it to atomically reload its dictionary from disk into the cache (no eviction window). The rebuilt factory then picks up the fresh
  dictionary.
  
 ## Changes:
  
  - TokenFilterFactory.java — adds default void reloadCachedResources() {} interface method
  - MapperService.java — new overload reloadSearchAnalyzers(registry, boolean); existing single-arg method delegates with false (backward compatible). Uses pattern matching and Arrays.stream().forEach() for the
  reload walk.
  - HunspellTokenFilterFactory.java — stores hunspellService/refPath/locale fields (assigned after validation); adds updateable flag support (AnalysisMode.SEARCH_TIME); overrides reloadCachedResources()
  - HunspellService.java — adds reloadDictionaryFromRefPath(refPath, locale) and reloadDictionary(locale) methods that atomically load fresh from disk and replace the cached entry
  
  ## Related
  
  - Refactoring PR: #21516
  - ISM companion PR (adds reload_cache query param to REST API): upcoming
  - RFC: #20712
  
 ## Check List
  
  For more information on following Developer Certificate of Origin and signing off your commits, please check here
  (https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


shayush622:
The RaceCondition indicated by the [comment](https://github.com/opensearch-project/OpenSearch/pull/21559#issuecomment-4404576571) above it theoretically possible but not a real concern in practice, because: 
1. Who else would load the dictionary between invalidate and reload? Only a search request hitting the same analyzer on the same node. But reloadSearchAnalyzers is synchronized — it holds the lock on the
  MapperService instance. While it's running, no new analyzer can be built from this MapperService.
  2. The race would be: Thread A invalidates cache → Thread B does a search, triggers getDictionaryFromRefPath via the old still-active analyzer → caches the old dictionary → Thread A's reload builds a new
  factory that gets the re-cached old dictionary.
  But this can't happen because the old analyzer's dictionary field is already loaded (it's a final field set at construction). Searches using the old analyzer don't call getDictionaryFromRefPath again — they
  use the already-loaded Dictionary object directly. The only code that calls getDictionaryFromRefPath is the HunspellTokenFilterFactory constructor during reload().
  4. What about a concurrent reloadSearchAnalyzers on another index? Different MapperService instance, different lock — but they'd only invalidate their own filter's cache key, not ours.
  
  So the window doesn't exist in practice. The dictionary is only re-fetched from cache during factory construction (inside reload()), which happens immediately after invalidation within the same synchronized
  block.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -123,6 +134,29 @@ public boolean longestOnly() {
...
 
+    /**
+     * Returns the analysis mode for this filter.
+     * When {@code updateable: true} is set, returns {@code SEARCH_TIME} which enables hot-reload
+     * via the _refresh_search_analyzers API.
+     */
+    @Override
+    public AnalysisMode getAnalysisMode() {
+        return this.analysisMode;
+    }
```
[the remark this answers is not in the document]
shayush622: the _refresh_search_analyzers API exists in ISM repo

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -856,6 +868,13 @@ public synchronized List<String> reloadSearchAnalyzers(AnalysisRegistry registry
         for (NamedAnalyzer namedAnalyzer : indexAnalyzers.getAnalyzers().values()) {
             if (namedAnalyzer.analyzer() instanceof ReloadableCustomAnalyzer) {
                 ReloadableCustomAnalyzer analyzer = (ReloadableCustomAnalyzer) namedAnalyzer.analyzer();
+                if (invalidateCache) {
```
RajatGupta02: By the time invalidateCache() runs, the new HunspellTokenFilterFactory has already been constructed and has captured a reference to the stale Dictionary. so basically the hot-reload silently does nothing. invalidateCache() only clears the lookup table. It cannot reach into a factory that already cached its own pointer to the dictionary. So evicting after the new factory has been constructed is too late: the new factory has already copied the stale pointer into a final field, and that field is what every search uses. 

Fix is invalidate before `buildTokenFilterFactories`, so that the cache miss forces a disk reload

shayush622: ack, thanks for pointing out. the intent was to invalidate before building tokenFilterFactories, but it got disordered while refactoring the code, thanks for pointing out.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -123,6 +134,29 @@ public boolean longestOnly() {
...
+     * reloaded from disk on the next access. Called by {@code MapperService.reloadSearchAnalyzers}
+     * when the caller requests cache invalidation.
+     */
+    public void invalidateCache() {
+        if (refPath != null) {
+            hunspellService.invalidateDictionary(HunspellService.buildRefPathCacheKey(refPath, locale));
+        } else if (locale != null) {
+            hunspellService.invalidateDictionary(locale);
+        }
+    }
```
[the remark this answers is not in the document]
shayush622: updated the javadoc

server/src/main/java/org/opensearch/indices/analysis/HunspellService.java
```diff
@@ -389,4 +389,20 @@ public static String buildRefPathCacheKey(String refPath, String locale) {
...
+     */
+    public boolean invalidateDictionary(String cacheKey) {
+        Dictionary removed = dictionaries.remove(cacheKey);
+        if (removed != null) {
+            logger.info("Invalidated hunspell dictionary cache for key: {}", cacheKey);
+            return true;
+        }
+        logger.debug("No cached dictionary found for key: {}", cacheKey);
+        return false;
+    }
```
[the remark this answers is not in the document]
shayush622: ack, updated to log.debug from log.info

server/src/test/java/org/opensearch/index/analysis/HunspellTokenFilterFactoryTests.java
```diff
@@ -368,4 +377,152 @@ public void testLanguageAliasForLocale() throws IOException {
...
+        Dictionary dict2 = hunspellService.getDictionary("en_US");
+        assertNotNull(dict2);
+        assertNotSame("Expected fresh Dictionary instance after invalidateCache()", dict1, dict2);
+    }
+
+    /**
+     * Simulates the MapperService.reloadSearchAnalyzers(registry, true) invalidation walk.
+     * Verifies that iterating token filters and calling invalidateCache() on HunspellTokenFilterFactory
+     * correctly evicts the cached dictionary.
+     */
```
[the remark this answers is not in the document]
shayush622: The simulated walk test covers the instanceof + invalidateCache() logic.
  The real E2E test will be covered in index-management repo.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -82,6 +86,13 @@ public HunspellTokenFilterFactory(IndexSettings indexSettings, String name, Sett
         String refPath = settings.get("ref_path");
         String locale = settings.get("locale", settings.get("language", settings.get("lang", null)));
 
+        // Check for updateable flag
+        boolean updateable = settings.getAsBoolean("updateable", false);
+        this.analysisMode = updateable ? AnalysisMode.SEARCH_TIME : AnalysisMode.ALL;
+        this.hunspellService = hunspellService;
+        this.refPath = refPath;
+        this.locale = locale;
```
shwetathareja: initialize them after validation is successful

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -846,8 +847,34 @@ protected Analyzer getWrappedAnalyzer(String fieldName) {
...
+
+        if (invalidateCache) {
+            // Invalidate caches BEFORE building new factories so they load fresh from disk
+            for (NamedAnalyzer namedAnalyzer : indexAnalyzers.getAnalyzers().values()) {
+                if (namedAnalyzer.analyzer() instanceof ReloadableCustomAnalyzer) {
+                    ReloadableCustomAnalyzer analyzer = (ReloadableCustomAnalyzer) namedAnalyzer.analyzer();
+                    for (TokenFilterFactory filter : analyzer.getComponents().getTokenFilters()) {
+                        if (filter instanceof HunspellTokenFilterFactory) {
+                            ((HunspellTokenFilterFactory) filter).invalidateCache();
+                        }
```
shwetathareja: this looks like a hack. you can add default method in the TokenFilterFactory interface for invalidating caches so that you don't need to cast to HunspellTokenFilterFactory

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: ack, moved `this.hunspellService`, `this.refPath`, and `this.locale` assignments to after the validation/loading block. No state is stored if construction fails.

server/src/main/java/org/opensearch/index/mapper/MapperService.java
shayush622: ack, added a default void invalidateCache() {} method to TokenFilterFactory interface. MapperService now just calls filter.invalidateCache() without instanceof/cast. HunspellTokenFilterFactory overrides it with the actual eviction logic.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
```diff
@@ -123,6 +135,29 @@ public boolean longestOnly() {
...
+    @Override
+    public AnalysisMode getAnalysisMode() {
+        return this.analysisMode;
+    }
+
+    /**
+     * Evicts this filter's hunspell dictionary from the node-level {@code HunspellService} cache.
+     * Must be called <em>before</em> rebuilding analyzer factories so that the newly constructed
+     * factory loads a fresh dictionary from disk.
+     */
```
shwetathareja, line 151: you would need to Override

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -846,8 +846,32 @@ protected Analyzer getWrappedAnalyzer(String fieldName) {
...
+    public synchronized List<String> reloadSearchAnalyzers(AnalysisRegistry registry, boolean invalidateCache) throws IOException {
+        logger.info("reloading search analyzers (invalidateCache={})", invalidateCache);
+
+        if (invalidateCache) {
+            // Invalidate caches BEFORE building new factories so they load fresh from disk
+            for (NamedAnalyzer namedAnalyzer : indexAnalyzers.getAnalyzers().values()) {
+                if (namedAnalyzer.analyzer() instanceof ReloadableCustomAnalyzer) {
+                    ReloadableCustomAnalyzer analyzer = (ReloadableCustomAnalyzer) namedAnalyzer.analyzer();
+                    for (TokenFilterFactory filter : analyzer.getComponents().getTokenFilters()) {
+                        filter.invalidateCache();
```
shwetathareja: if you invalidate the cache in advance what will happen if any search request during this time before reload it?

shayush622: AFAIK, in-flight searches are unaffected. The active analyzer's HunspellTokenFilterFactory holds a final Dictionary reference loaded at constructio, it never re-reads from the HunspellService cache during
  analysis. The cache is only consulted when building a new factory instance (during reload()). So the window between invalidation and reload is safe: existing searches use the old dictionary, and the rebuilt
  factory picks up the fresh one.

server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java
shayush622: ack, added

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -846,8 +846,32 @@ protected Analyzer getWrappedAnalyzer(String fieldName) {
...
+     * @param invalidateCache If true, evicts cached hunspell dictionaries before rebuilding analyzers
+     * @return List of reloaded analyzer names
+     */
+    public synchronized List<String> reloadSearchAnalyzers(AnalysisRegistry registry, boolean invalidateCache) throws IOException {
+        logger.info("reloading search analyzers (invalidateCache={})", invalidateCache);
+
+        if (invalidateCache) {
+            // Invalidate caches BEFORE building new factories so they load fresh from disk
+            for (NamedAnalyzer namedAnalyzer : indexAnalyzers.getAnalyzers().values()) {
+                if (namedAnalyzer.analyzer() instanceof ReloadableCustomAnalyzer) {
```
cwperks:
```suggestion
                if (namedAnalyzer.analyzer() instanceof ReloadableCustomAnalyzer analyzer) {
```

cwperks: Let's use [PatternMatching](https://docs.oracle.com/en/java/javase/17/language/pattern-matching.html) here.

server/src/main/java/org/opensearch/index/analysis/TokenFilterFactory.java
```diff
@@ -149,4 +149,11 @@ public TokenStream create(TokenStream tokenStream) {
             return tokenStream;
         }
     };
+
+    /**
+     * Invalidates any cached resources held by this filter factory.
+     * Called during analyzer reload when cache invalidation is requested.
+     * Default implementation is a no-op.
+     */
+    default void invalidateCache() {}
```
cwperks: I still don't think this captures the semantics well. I think we should consider something like `default void reloadCachedResources() {}`

and in the implementation do an overwrite so there's no time where the cache key is empty:

```
public Dictionary reloadDictionaryFromRefPath(String refPath, String locale) {
    if (Strings.isNullOrEmpty(refPath)) {
        throw new IllegalArgumentException("refPath cannot be null or empty");
    }
    if (Strings.isNullOrEmpty(locale)) {
        throw new IllegalArgumentException("locale cannot be null or empty");
    }

    String cacheKey = buildRefPathCacheKey(refPath, locale);

    final Dictionary freshDictionary;
    try {
        freshDictionary = loadDictionaryFromRefPath(refPath, locale);
    } catch (Exception e) {
        throw new IllegalStateException(
            String.format(Locale.ROOT, "Failed to reload hunspell dictionary for ref_path [%s] locale [%s]", refPath, locale),
            e
        );
    }

    dictionaries.put(cacheKey, freshDictionary);
    logger.debug("Reloaded hunspell dictionary cache for key [{}]", cacheKey);
    return freshDictionary;
}
```

shayush622: I really liked and apprectiate this suggestion! This would technically swap the value of the cacheKey which removes our painPoint regarding the time gap between cache Eviction and freshDictionary read from fs. Thanks for this suggestion!

server/src/main/java/org/opensearch/index/mapper/MapperService.java
shayush622: ack, improved

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -846,16 +846,38 @@ protected Analyzer getWrappedAnalyzer(String fieldName) {
...
-        logger.info("reloading search analyzers");
-        // refresh indexAnalyzers and search analyzers
+        return reloadSearchAnalyzers(registry, false);
+    }
+
+    /**
+     * Reloads search analyzers, optionally invalidating hunspell dictionary caches first.
+     *
+     * @param registry The analysis registry
+     * @param invalidateCache If true, evicts cached hunspell dictionaries before rebuilding analyzers
```
cwperks: This param should be renamed to reflect the change in the interface.

server/src/main/java/org/opensearch/index/mapper/MapperService.java
```diff
@@ -846,16 +846,38 @@ protected Analyzer getWrappedAnalyzer(String fieldName) {
...
+    public synchronized List<String> reloadSearchAnalyzers(AnalysisRegistry registry, boolean invalidateCache) throws IOException {
+        logger.info("reloading search analyzers (invalidateCache={})", invalidateCache);
+
+        if (invalidateCache) {
+            // Invalidate caches BEFORE building new factories so they load fresh from disk
+            for (NamedAnalyzer namedAnalyzer : indexAnalyzers.getAnalyzers().values()) {
+                if (namedAnalyzer.analyzer() instanceof ReloadableCustomAnalyzer analyzer) {
+                    for (TokenFilterFactory filter : analyzer.getComponents().getTokenFilters()) {
+                        filter.reloadCachedResources();
+                    }
```
cwperks:
```suggestion
                    Arrays.stream(analyzer.getComponents().getTokenFilters())
    .forEach(TokenFilterFactory::reloadCachedResources);
```

server/src/main/java/org/opensearch/index/mapper/MapperService.java
shayush622: ack

server/src/test/java/org/opensearch/indices/analyze/HunspellServiceTests.java
```diff
@@ -388,4 +388,49 @@ private void createHunspellFiles(Path directory, String locale) throws java.io.I
...
+    public void testReloadTraditionalDictionary() throws Exception {
+        Settings settings = Settings.builder()
+            .put(Environment.PATH_HOME_SETTING.getKey(), createTempDir())
+            .put(HUNSPELL_LAZY_LOAD.getKey(), true)
+            .build();
+        Environment environment = new Environment(settings, getDataPath("/indices/analyze/conf_dir"));
+        HunspellService hunspellService = new HunspellService(settings, environment, emptyMap());
+
+        hunspellService.getDictionary("en_US");
+        hunspellService.reloadDictionary("en_US");
```
cwperks: What does this test assert exactly?

server/src/test/java/org/opensearch/index/analysis/HunspellTokenFilterFactoryTests.java
```diff
@@ -368,4 +377,150 @@ public void testLanguageAliasForLocale() throws IOException {
...
+
+        HunspellTokenFilterFactory hunspellFactory = new HunspellTokenFilterFactory(idx, "my_hunspell", filterSettings, hunspellService);
+
+        // Pre-load into cache
+        Dictionary dict1 = hunspellService.getDictionaryFromRefPath("analyzers/test-dict", "en_US");
+        assertNotNull(dict1);
+
+        // Simulate MapperService cacheReload walk (same code as MapperService.reloadSearchAnalyzers)
+        TokenFilterFactory[] filters = new TokenFilterFactory[] { hunspellFactory };
+        for (TokenFilterFactory filter : filters) {
```
cwperks: Why loop here if there's a single factory?

shayush622: ack, will fix this

server/src/test/java/org/opensearch/indices/analyze/HunspellServiceTests.java
shayush622: this test meant to test that traditional dictionaries should be refreshed, missing assert statements, will fix this. thanks for pointing out

cwperks (approved):

shwetathareja (approved):

PR #21531, opened 2026-05-07
Add IO task cancellation support for DataFusion queries

aravindsagar:
### Description
Add cancellation support for IO runtime tasks in DataFusion. This adds a cancellation token for tasks in the Rust-side query registry, and introduces cancellation wrappers that race the future that is doing the work against the cancellation token - if the token is fired, the task is discarded and the function returns early.

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [x] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21560, opened 2026-05-08
Add CPU task cancellation for DataFusion queries + wiring

aravindsagar:
### Description

This PR adds CPU executor task cancellation to the DataFusion query engine, complementing the IO task cancellation added in #21531. When a query is cancelled (via `_tasks/_cancel`, HTTP disconnect, or timeout), both the IO-level async work and the CPU-bound DataFusion operator task are now terminated promptly.

#### Changes

**Rust — CPU task abort infrastructure:**
- `executor.rs`: Add `spawn_with_abort_handle()` which returns a tokio `AbortHandle` alongside the join future. Refactor `spawn()` to delegate to it.
- `cross_rt_stream.rs`: Add `new_with_df_error_stream_cancellable()` that eagerly spawns the CPU task and returns the `AbortHandle`. The original `new_with_df_error_stream()` remains unchanged for callers that don't need cancellation.
- `query_executor.rs`: Both `execute_query` and `execute_with_context` use the cancellable variant and store the abort handle via `query_tracker::set_abort_handle()`.
- `indexed_executor.rs`: Same — uses the cancellable variant.
- `query_tracker.rs`: Add `abort_handle: OnceLock<AbortHandle>` to `QueryTracker`. `cancel_query()` now fires both `token.cancel()` (IO) and `handle.abort()` (CPU).
- However, if the Rust executor is blocked in a synchronous FFM upcall to Java (e.g., collectDocs for Lucene filter delegation), the token cannot interrupt the call — cancellation must be checked cooperatively inside the Java callback.

**Java — cancellation wiring:**
- `DatafusionSearchExecEngine`: Registers a cancellation listener on `SearchShardTask` before query execution, clears it after. This ensures `cancelQuery()` fires immediately when the task is cancelled, even while the Java thread is blocked in an FFM call.
- `DatafusionContext`: Adds `isCancelled()` and `getContextId()` helpers. `close()` fires `cancelQuery()` as a safety net.

**Java — FFM upcalls on the indexed query path**

If the Rust executor is blocked in a synchronous FFM upcall to Java (e.g., collectDocs for Lucene filter delegation), the token cannot interrupt the call — cancellation must be checked cooperatively inside the Java callback.

- Add `default boolean isCancelled()` to `FilterDelegationHandle` interface
- Check `handle.isCancelled()` in `FilterTreeCallbacks.collectDocs()` before delegating — returns `-1` (error) to Rust on cancellation
- `LuceneFilterDelegationHandle` implements `isCancelled()` by checking the task's cancellation state via a `BooleanSupplier` wired from the `ShardScanExecutionContext`
- `LuceneAnalyticsBackendPlugin` passes the cancellation supplier when constructing the handle

**Server — `SearchShardTask` (minimal change):**
- Adds `setCancellationListener(Runnable)` / `clearCancellationListener()` and an `onCancelled()` override.
- **Rationale:** The DataFusion FFM bridge blocks the calling Java thread during query execution (`block_on`). When `_tasks/_cancel` fires, the framework cannot call `DatafusionContext.close()` until the blocked thread returns — creating a deadlock where cancellation needs `close()` but `close()` needs the FFM call to return. The `onCancelled()` hook breaks this cycle by firing the Rust cancellation token from the transport thread immediately, without waiting for the blocked query thread. This is a generic mechanism (not DataFusion-specific) that any execution engine with blocking native calls can use.

### Testing

- Existing unit tests pass (26 tests, 0 failures)
- Manual testing with artificial delays confirms both IO and CPU cancellation fire within the expected window

### Related Issues

### Check List


aravindsagar:
gradle-check failing
```
* What went wrong:
Execution failed for task ':test:fixtures:azure-fixture:composeBuild'.
> Exit-code 1 when calling /usr/bin/docker-compose, stdout: Sending build context to Docker daemon  26.62kB

  Step 1/5 : FROM ubuntu:22.04
  22.04: Pulling from library/ubuntu
```

server/src/main/java/org/opensearch/action/search/SearchShardTask.java
```diff
@@ -84,4 +93,30 @@ public boolean supportsResourceTracking() {
...
     }
+
+    /**
+     * Registers a listener that is called immediately when this task is cancelled.
+     * Only one listener is supported; registering a second one replaces the first.
+     * If the task is already cancelled when the listener is registered, it fires immediately.
+     */
+    public void setCancellationListener(Runnable listener) {
+        cancellationListener.set(listener);
+        if (isCancelled() && listener != null) {
```
himshikhagupta: Is there a possible race condition here? If onCancelled() fires on the transport thread between these two operations, onCancelled() reads the listener (sees it), runs it — but then the isCancelled() check also sees true and runs it again. The listener fires twice.
Should we use `cancellationListener.getAndSet(null) != null` ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionContext.java
```diff
@@ -48,6 +50,10 @@ public DatafusionContext(Task task, DatafusionReader reader, NativeRuntimeHandle
 
     @Override
     public void close() throws IOException {
+        // Signal cancellation to any in-flight Rust tasks for this query.
+        if (datafusionQuery != null && datafusionQuery.getContextId() != 0) {
+            NativeBridge.cancelQuery(datafusionQuery.getContextId());
```
himshikhagupta, line 55: Lets add a test for this

himshikhagupta (commented):
Minor comments, mostly looks good.

server/src/main/java/org/opensearch/action/search/SearchShardTask.java
aravindsagar: Good catch, updated

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionContext.java
aravindsagar: Done

himshikhagupta (approved):

Bukhtawar (commented):
```
try {
            DatafusionSearcher searcher = datafusionContext.getSearcher();
            searcher.search(datafusionContext);
            StreamHandle handle = datafusionContext.takeStreamHandle();
            return new DatafusionResultStream(handle, allocator);
        } finally {
            // Clear the listener so the task doesn't hold a reference after the query.
            if (shardTask != null) {
                shardTask.clearCancellationListener();
            }
```
At this point I doubt if the shard task finished or even the stream is fully consumed

aravindsagar:
> ```
> try {
>             DatafusionSearcher searcher = datafusionContext.getSearcher();
>             searcher.search(datafusionContext);
>             StreamHandle handle = datafusionContext.takeStreamHandle();
>             return new DatafusionResultStream(handle, allocator);
>         } finally {
>             // Clear the listener so the task doesn't hold a reference after the query.
>             if (shardTask != null) {
>                 shardTask.clearCancellationListener();
>             }
> ```
> 
> At this point I doubt if the shard task finished or even the stream is fully consumed

Thanks for catching. Made changes so that the listener stays registered for the task's lifetime.

Bukhtawar (approved):
Thanks LGTM

PR #21561, opened 2026-05-08
Add multi-partition consumer and partition-aware pointer for flexible shard-to-partition mapping

### LINKED ISSUE #21260: [Feature Request] Add multi-partition consumer factory methods and consumer implementation

`IngestionConsumerFactory` only supports creating single-partition consumers via `createShardConsumer(clientId, shardId)`, where shard ID is used directly as partition ID. There is no way to discover source partition count or create a consumer that reads from multiple partitions. The `KafkaOffset` pointer type has no partition awareness, and the `IngestionShardConsumer` interface has no method for seeking multiple partitions during recovery.

Add two default methods to `IngestionConsumerFactory` (`@PublicApi`): `getSourcePartitionCount()` and `createMultiPartitionShardConsumer()`. Add `seekToPartitionOffsets()` default method to `IngestionShardConsumer`. Introduce `KafkaPartitionOffset` (extends `KafkaOffset` with partition ID, `"partition:offset"` string format), `KafkaMultiPartitionConsumer` (composition-based, single `KafkaConsumer` assigned to multiple `TopicPartition`s), and `PartitionAwarePointer` marker interface. Update `KafkaConsumerFactory` with backward-compatible pointer parsing and widened type parameter to support both consumer types.

### Related component

_No response_

_No response_

### Additional context

_No response_

imRishN:
### Description
Introduces the consumer layer for flexible shard-to-partition mapping, enabling a single OpenSearch shard to consume from multiple Kafka partitions. 

### Related Issues
Resolves #21260

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/SourcePartitionAwarePointer.java
+++ b/server/src/main/java/org/opensearch/index/SourcePartitionAwarePointer.java
@@ -8,18 +8,21 @@
 
 package org.opensearch.index;
 
+import org.opensearch.common.annotation.PublicApi;
+
 /**
  * Marker interface for {@link IngestionShardPointer} implementations that carry source partition
  * information. Used by the checkpoint tracking layer to maintain per-partition progress when a
  * single shard consumes from multiple source partitions.
  * <p>
  * Implementations should return the source partition ID (e.g., Kafka partition number) from
- * {@link #getPartition()}.
+ * {@link #getSourcePartition()}.
  */
-public interface PartitionAwarePointer {
+@PublicApi(since = "3.7.0")
+public interface SourcePartitionAwarePointer {
     /**
      * Returns the source partition ID that this pointer belongs to.
      * @return the partition ID
      */
-    int getPartition();
+    int getSourcePartition();
 }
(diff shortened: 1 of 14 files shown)


server/src/main/java/org/opensearch/index/SourcePartitionAwarePointer.java
```diff
@@ -0,0 +1,28 @@
...
+ * Implementations should return the source partition ID (e.g., Kafka partition number) from
+ * {@link #getSourcePartition()}.
+ */
+@PublicApi(since = "3.7.0")
+public interface SourcePartitionAwarePointer {
+    /**
+     * Returns the source partition ID that this pointer belongs to.
+     * @return the partition ID
+     */
+    int getSourcePartition();
```
msfroh, line 27: Will this work for Kinesis? 

A Kinesis stream shard has a string identifier (see the `ShardId` property [here](https://docs.aws.amazon.com/kinesis/latest/APIReference/API_Shard.html)).

A Kinesis stream in steady state will have a set of "active" shards that have a well-defined order based on the `HashKeyRange` property, which can be used to map between integers and shards. Unfortunately, since a Kinesis shard can be split, it's possible for a shard to stop existing and get replaced by multiple shards, which would change the numbers for all shards covering higher ranges.

I don't know if it makes sense to create an interface to identify source partitions, so we have more flexibility than using an integer.

IMO, if you have a design that works through the lifecycle of a Kinesis shard split, that would be ideal. For that it would be something like:
1. OpenSearch index and Kinesis stream each have N shards in a 1-1 mapping.
2. Kinesis shards get split, resulting in 2N shards. Each OpenSearch shard automatically adjusts to read from 2 child shards (ideally without user intervention). We've just double stream throughput (and cost).

PR #21563, opened 2026-05-08
 [Analytics Engine] Add data node row ID emission and fetch-by-row-ids for QTF late materialization

alchemist51:
Adds the data-node half of Query-Then-Fetch late materialization:

  - **Row ID emission (indexed path):** Compute `__row_id__` from position via `PositionMap` in `finalize_batch` for all filter classifications (SingleCollector, Tree, predicate-only without sort)
  - **Row ID emission (ListingTable path):** `ShardTableProvider` with `row_base` partition column + `ProjectRowIdOptimizer` for queries requiring data-node sort without a collector
  - **Fetch-by-row-ids:** Rust execution path that resolves global IDs → per-file positions → `ParquetAccessPlan` with row-level `RowSelection`, skipping entire row groups with no target rows
  - **Reader context:** `ReaderContextStore` holds shard readers open across query→fetch phases with configurable keep-alive (`analytics.qtf.reader_context.keep_alive`, default 5m) and reaper
  - **Shared helpers:** Extracted `build_query_runtime_env`, `build_shard_file_infos`, `store_url_from_table_path` for reuse across query and fetch paths


server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshotManager.java
```diff
@@ -110,6 +110,11 @@ public CatalogSnapshotManager(
         for (CatalogSnapshot cs : committedSnapshots) {
             catalogSnapshotMap.put(cs.getGeneration(), cs);
         }
+
```
alchemist51, line 113: Will remove it once the #21553 get's merged

alchemist51:
Closing it, have added my changed in #21836

PR #21404, opened 2026-04-27
Fix deadlock between engineMutex and writeLock during index close and engine reset (#11869)

sean-:
During `resetEngineToGlobalCheckpoint` on a segment replication replica, the `ReadOnlyEngine` delegates (`acquireLastIndexCommit`, `acquireSafeIndexCommit`, `getSegmentInfosSnapshot`) synchronized on `engineMutex`.  This creates a deadlock cycle when close runs concurrently: the close thread holds `engineMutex` and waits for `writeLock`, while the recovery thread holds `readLock` (via `recoverFromTranslog`) and a refresh listener calls a delegate that waits for `engineMutex`.

Replace `synchronized(engineMutex)` in the delegates with an `AtomicReference` that provides happens-before visibility to unsynchronized readers. Mutations remain serialized by `engineMutex`; the engine's own `writeLock` in `Engine.close()` ensures in-flight delegate calls complete before teardown.

Fixes: #11869
May fix: #12114

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21572, opened 2026-05-08
Fix TOCTOU race in newEngineReference null checks during engine reset (#21404)

sean-:
Store newEngineReference.get() in a local variable before the null check to prevent a race where another thread nulls the reference between the check and subsequent use, which would cause a NullPointerException.

### DIFF
--- a/server/src/main/java/org/opensearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/opensearch/index/shard/IndexShard.java
@@ -5544,28 +5544,31 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
             ) {
                 @Override
                 public GatedCloseable<IndexCommit> acquireLastIndexCommit(boolean flushFirst) {
-                    if (newEngineReference.get() == null) {
+                    Indexer indexer = newEngineReference.get();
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -5409,27 +5409,30 @@ void resetEngineToGlobalCheckpoint() throws IOException {
             ) {
                 @Override
                 public GatedCloseable<IndexCommit> acquireLastIndexCommit(boolean flushFirst) {
-                    if (newEngineReference.get() == null) {
+                    Indexer indexer = newEngineReference.get();
```
andrross, line 5547: I'm fine with this change because it indeed does _look_ like a TOCTOU race, but with the semantics of SetOnce it's actually impossible for the reference to change if `newEngineReference.get()` is not null.

andrross (approved):

sean-:
The checks continue to fail with error messages like:
```
toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
Service 'azure-fixture' failed to build : Build failed
```

PR #21248, opened 2026-04-16
Adding fieldtypecapabilities changes

darjisagar7:
# Add FieldTypeCapabilities-based routing for pluggable data formats

## Description

This PR introduces a capability-based field routing mechanism that allows the composite indexing engine to determine which data format (e.g., Parquet, Lucene) is responsible for storing each field's data. At mapping creation time, each `MappedFieldType` is assigned an immutable capability map (`Map<DataFormat, Set<Capability>>`) that records which format owns which capabilities (e.g., `COLUMNAR_STORAGE`, `FULL_TEXT_SEARCH`, `POINT_RANGE`) for that field. At indexing time, each format's `DocumentInput` uses this map to accept only the fields it owns.

## Problem

Previously, when multiple data formats were registered in a composite engine, there was no mechanism to decide which format should handle which aspect of a field. Every format received every field, leading to redundant storage and no clear ownership of capabilities like full-text search vs. columnar storage.

### Capability assignment algorithm

For each field type, the `DataFormatRegistry.computeCapabilityMap()` method:

1. Determines which capabilities the field requests (from its mapping configuration: `index`, `doc_values`, `store`)
2. For each requested capability, finds all formats that declare support for it
3. The highest-priority format (lowest priority value) wins ownership of that capability
4. For metadata fields (e.g., `_id`, `_source`) whose type names aren't declared by any format, `defaultCapabilities()` provides a fallback assigned to the primary format

### Key changes

- **`MappedFieldType`**: Added `capabilityMap`, `requestedCapabilities()`, and `defaultCapabilities()` methods
- **`DataFormatRegistry`**: Added `computeCapabilityMap()` with 1-arg, 2-arg, and 3-arg overloads
- **`Mapper.BuilderContext`**: Carries `DataFormatRegistry` and provides `assignCapabilities()` helper
- **`DocumentMapper.Builder`**: Recursively assigns capability maps to all field types (user fields + metadata) at mapping build time
- **`ObjectMapper`**: Validates at parse time that each field type is supported by at least one registered format
- **Field type overrides**: `NumberFieldType`, `DateFieldType`, `IpFieldType`, `RangeFieldType`, `ScaledFloatFieldType` override `requestedCapabilities()` to use `POINT_RANGE` instead of `FULL_TEXT_SEARCH`
- **Metadata field overrides**: `_id`, `_routing`, `_seq_no`, `_source`, `_version` override `defaultCapabilities()` to declare their intrinsic storage needs
[...]

PR #21585, opened 2026-05-10
Add indexing support for metadata fields.

mgodwan:
Improve Lucene field parser logic to use BytesRef directly

### Description
Add field support for metadata fields.
Id, seq no, version, primary term are the non-nullable fields which should be present in every document.
Currently, the code is with assumption around parquet as primary, with lucene containing indices for seq no and id.

### Check List

### DIFF
--- a/modules/parent-join/src/main/java/org/opensearch/join/mapper/ParentJoinFieldMapper.java
+++ b/modules/parent-join/src/main/java/org/opensearch/join/mapper/ParentJoinFieldMapper.java
@@ -378,6 +378,9 @@ public final class ParentJoinFieldMapper extends FieldMapper {
 
     @Override
     public void parse(ParseContext context) throws IOException {
+        if (isPluggableDataFormatFeatureEnabled(context)) {
+            throw new UnsupportedOperationException("Field type: " + CONTENT_TYPE + " is not supported for pluggable formats");
+        }
         context.path().add(simpleName());
         XContentParser.Token token = context.parser().currentToken();
         String name = null;
@@ -435,13 +438,8 @@ public final class ParentJoinFieldMapper extends FieldMapper {
         }
 
         BytesRef binaryValue = new BytesRef(name);
-        if (isPluggableDataFormatFeatureEnabled(context)) {
-            context.documentInput().addField(fieldType(), binaryValue);
-        } else {
-            context.doc().add(new Field(fieldType().name(), binaryValue, fieldType));
-            context.doc().add(new SortedDocValuesField(fieldType().name(), binaryValue));
-        }
-
+        context.doc().add(new Field(fieldType().name(), binaryValue, fieldType));
+        context.doc().add(new SortedDocValuesField(fieldType().name(), binaryValue));
         context.path().remove();
     }
 
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDocumentInput.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDocumentInput.java
@@ -9,7 +9,9 @@
 package org.opensearch.be.lucene.index;
 
 import org.apache.lucene.document.Document;
+import org.apache.lucene.document.FieldType;
 import org.apache.lucene.document.SortedNumericDocValuesField;
+import org.apache.lucene.index.DocValuesType;
 import org.opensearch.be.lucene.LuceneFieldFactory;
 import org.opensearch.be.lucene.LuceneFieldFactoryRegistry;
 import org.opensearch.common.annotation.ExperimentalApi;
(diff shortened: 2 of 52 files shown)


server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1157,7 +1166,13 @@ public DocsStats docStats() {
             long count = snapshot.get()
                 .getSegments()
                 .stream()
-                .flatMap(segment -> segment.dfGroupedSearchableFiles().values().stream())
+                .map(
+                    segment -> segment.dfGroupedSearchableFiles()
+                        .values()
+                        .stream()
+                        .findFirst()
+                        .orElseGet(() -> new WriterFileSet("", -1L, Set.of(), 0))
```
mgodwan, line 1178: This fixes multi format, and uses numRows from first one.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1419,7 +1434,7 @@ private void applyMergeChanges(MergeResult mergeResult, OneMerge oneMerge) {
     }
 
     private void triggerPossibleMerges() {
-        if (Booleans.parseBoolean(System.getProperty(MERGE_ENABLED_PROPERTY, Boolean.FALSE.toString())) == false) {
+        if (Booleans.parseBoolean(System.getProperty(MERGE_ENABLED_PROPERTY, Boolean.TRUE.toString())) == false) {
```
mgodwan, line 1441: This is intended. Enabling merges now since both lucene/parquet are in place

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDocumentInput.java
alchemist51: let's remove this also as part of this PR? We can make all the code use DocumentInput.ROW_ID_FIELD

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/fields/ArrowSchemaBuilder.java
```diff
@@ -59,6 +60,7 @@ public static Schema getSchema(MapperService mapperService) {
         // Add row ID field (long)
         LongParquetField longField = new LongParquetField();
         fields.add(new Field(DocumentInput.ROW_ID_FIELD, longField.getFieldType(), null));
```
alchemist51, line 62: Should we make RowID also nonNullable?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/fields/ArrowSchemaBuilder.java
alchemist51, line 43: Will it make sense to have few invariants in the comsumer ParquetWriter/LuceneWriter. To check for the essential metadata fields in the supplied Schema?

alchemist51: Or we can do it here as well:

```
    @Override
    public List<FieldValuePair> getFinalInput() {
        assert rowId >= 0 : "Row ID must be set before calling getFinalInput";
        return collectedFields;
    }
```

server/src/main/java/org/opensearch/index/engine/dataformat/DocumentInput.java
```diff
@@ -46,4 +46,6 @@ public interface DocumentInput<T> extends AutoCloseable {
      * @param rowId the row ID value
      */
     void setRowId(String rowIdFieldName, long rowId);
+
+    long getFieldCount(String fieldName);
```
alchemist51, line 55: Can we add documentation?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/fields/ArrowSchemaBuilder.java
mgodwan: Yes, let me add the verification here

Bukhtawar (commented):
If Lucene is used standalone, those factories may produce a different field configuration (e.g., indexed-only, no stored) than what a standalone
  Lucene engine needs (indexed + stored, since there's no Parquet to hold the source). Do we need handling in composite engine?

mgodwan:
>  Lucene engine needs (indexed + stored, since there's no Parquet to hold the source). Do we need handling in composite engine?

Tracking this already in the meta issue. For now, the changes support parquet as primary.
Once, the field capabilities are determined per format during indexing, we will have the field caps passed to each format as part of MappedFieldType and honor that in composite engine.

modules/parent-join/src/main/java/org/opensearch/join/mapper/ParentJoinFieldMapper.java
```diff
@@ -378,6 +378,9 @@ protected void parseCreateFieldForPluggableFormat(ParseContext context) throws I
 
     @Override
     public void parse(ParseContext context) throws IOException {
+        if (isPluggableDataFormatFeatureEnabled(context)) {
+            throw new UnsupportedOperationException("Field type: " + CONTENT_TYPE + " is not supported for pluggable formats");
+        }
```
Bukhtawar, line 383: Should this be done at index creation time?

Bukhtawar (approved):
Left a comment, mostly looks good

modules/parent-join/src/main/java/org/opensearch/join/mapper/ParentJoinFieldMapper.java
mgodwan: https://github.com/opensearch-project/OpenSearch/pull/21248 This PR covers the validation for create index and mapping update flows

PR #21593, opened 2026-05-10
[Analytics Engine] Wire window-function support for PPL top / rare

ahkcs:
## Description

PPL `top` and `rare` lower (via `CalciteRelNodeVisitor#visitRareTopN` in opensearch-project/sql) to a `LogicalProject` containing
`ROW_NUMBER() OVER (PARTITION BY ... ORDER BY ...)`. On the analytics-engine route, `OpenSearchProjectRule.annotateExpr` treated the `RexOver` as an ordinary `RexCall` and fell into the scalar-function viability path. That path keys on `ScalarFunction.fromSqlOperatorWithFallback`, which returns null for window-only aggregates like `ROW_NUMBER`, so every `top` / `rare` query died with `No backend supports scalar function [ROW_NUMBER] among [datafusion]` before substrait emission.

This PR adds `EngineCapability.WINDOW` and detects `RexOver` ahead of the standard `RexCall` branch in `OpenSearchProjectRule.annotateExpr`. When the child's viable backends include a WINDOW-capable backend, the call is wrapped in an `AnnotatedProjectExpression` exactly like other annotated calls — the existing strip path's `OperatorAnnotation::unwrap` returns the original `RexOver` unchanged, so isthmus's `RexExpressionConverter#visitOver` emits an inline substrait `WindowFunctionInvocation`. The substrait standard catalog already binds `ROW_NUMBER`, so DataFusion's substrait consumer decodes it natively — no separate substrait Window rel needed.

`WINDOW` is intentionally coarse (one boolean per backend rather than a per-window-function `WindowCapability`). The substrait standard catalog already constrains which window aggregates the backend's substrait consumer can decode; a runtime decode failure is a clearer error than a duplicated registry split between SPI and backend. The existing `AggregateFunction` / `AggregateCapability` model remains right for ordinary aggregates; once a second backend with different window support lands, or a window-only function carries a non-standard call shape, WINDOW can be promoted to a per-function class without disturbing the planner-side annotation flow.

## Pass-rate impact

`CalciteTopCommandIT` / `CalciteRareCommandIT` against `tests.analytics.force_routing=true -Dtests.analytics.parquet_indices=true` on the SQL plugin's `integTestRemote`:

| IT | Before | After (this PR alone) | After (this PR + companion [opensearch-project/sql#5433](https://github.com/opensearch-project/sql/pull/5433)) |
|---|---|---|---|
| `CalciteTopCommandIT` | 0 / 6 | 5 / 6 | **6 / 6** |
| `CalciteRareCommandIT` | 0 / 5 | 3 / 5 | **5 / 5** |
| **combined** | **0 / 11** | **8 / 11** | **11 / 11** |

The 3-failure delta in the middle column comes entirely from pre-existing analytics-route gaps unrelated to window-function wiring (see below); they are closed by the companion SQL-plugin PR.

## Out-of-scope follow-ups (resolved by companion PR)

The 3 failures left after this PR are not specific to `top` / `rare` — they are pre-existing analytics-route gaps that already affect other commands. All 3 pass on the in-process Calcite path with the same query strings. They are closed by the companion SQL-plugin PR [opensearch-project/sql#5433](https://github.com/opensearch-project/sql/pull/5433):

1. **`testTopCommandLegacyFalse` / `testRareCommandLegacyFalse`** — `RestUnifiedQueryAction.applyClusterOverrides` (in opensearch-project/sql) only forwarded `PPL_REX_MAX_MATCH_LIMIT` to `UnifiedQueryContext`. `withSettings(PPL_SYNTAX_LEGACY_PREFERRED, "false")` updated the cluster setting but the analytics-engine route's `PPLQueryParser` reads it from the unified context, which the override builder never populated. The companion PR refactors the override builder into a `forwardClusterSetting` helper and forwards `PPL_SYNTAX_LEGACY_PREFERRED` alongside the rex-limit key.
2. **`testRareWithGroup`** — multiple states tied at the bucket count and `ROW_NUMBER` picked one based on insertion order, while the test expected a specific tie-broken value (same class as `testStatsSortOnMeasure`). The companion PR appends the rare/top field columns as secondary ASC keys to the `ROW_NUMBER` ORDER BY in `CalciteRelNodeVisitor#visitRareTopN`, so ties resolve deterministically across backends. This matches the OpenSearch terms-aggregation pushdown's existing `_key:asc` tie-break — wire payload unchanged.

## Tests

`ProjectRuleTests` adds three regressions on the new code path:
- `testRowNumberOverPartitionByOrderByMarksAsWindow` — happy path
- `testRowNumberWithoutWindowCapabilityErrors` — capability-gap error message names the window function
- `testStripAnnotationsPreservesRexOver` — strip path returns plain `LogicalProject(RexOver)` so isthmus can decode it

`MockDataFusionBackend` declares `EngineCapability.WINDOW` so existing project-rule tests continue to exercise the same backend gate.

## Verification

```bash
# 1. publishToMavenLocal in this checkout (sandbox.enabled=true)
# 2. publishToMavenLocal in opensearch-project/sql
# 3. ./gradlew :run -Dsandbox.enabled=true -PinstalledPlugins="['opensearch-job-scheduler:3.7.0.0-SNAPSHOT', 'arrow-flight-rpc', 'analytics-engine', 'parquet-data-format', 'analytics-backend-datafusion', 'analytics-backend-lucene', 'composite-engine', 'opensearch-sql-plugin:3.7.0.0-SNAPSHOT']"
# 4. ./gradlew :integ-test:integTestRemote \
#      -Dtests.rest.cluster=localhost:9200 \
#      -Dtests.cluster=localhost:9300 \
#      -Dtests.clustername=runTask \
#      -Dtests.analytics.force_routing=true \
#      -Dtests.analytics.parquet_indices=true \
#      --tests "org.opensearch.sql.calcite.remote.CalciteTopCommandIT" \
#      --tests "org.opensearch.sql.calcite.remote.CalciteRareCommandIT"
```

Also added self-contained QA mirrors under `sandbox/qa/analytics-engine-rest` exercising the analytics-engine route via `POST /_analytics/ppl` (no SQL plugin needed):

| QA IT | Tests |
|---|---|
| `TopCommandIT` | 5 / 5 — `top`, `top N`, `top N … by …`, `top` (default usenull), `top usenull=false` |
| `RareCommandIT` | 4 / 4 — `rare`, `rare … by …`, `rare` (default usenull), `rare usenull=false` |

Run via:

```bash
./gradlew :sandbox:qa:analytics-engine-rest:integTest -Dsandbox.enabled=true   --tests "*TopCommandIT" --tests "*RareCommandIT"
```

Sandbox-wide `./gradlew check -p sandbox -Dsandbox.enabled=true` is green.

### Issues Resolved
N/A — partial parity with PPL command coverage on the analytics-engine route. Closes the remaining 3 out-of-scope `top`/`rare` gaps together with the companion SQL-plugin PR [opensearch-project/sql#5433](https://github.com/opensearch-project/sql/pull/5433).

### Check List

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license. For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/MockDataFusionBackend.java
+++ b/sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/MockDataFusionBackend.java
@@ -140,7 +140,16 @@ public class MockDataFusionBackend extends MockBackend implements SearchBackEndP
     protected Set<WindowCapability> windowCapabilities() {
         return Set.of(
             new WindowCapability(
-                Set.of(WindowFunction.SUM, WindowFunction.AVG, WindowFunction.COUNT, WindowFunction.MIN, WindowFunction.MAX),
+                Set.of(
+                    WindowFunction.SUM,
+                    WindowFunction.AVG,
+                    WindowFunction.COUNT,
+                    WindowFunction.MIN,
+                    WindowFunction.MAX,
+                    // ROW_NUMBER backs PPL `top` / `rare` / `dedup` lowering
+                    // (ROW_NUMBER() OVER (PARTITION BY ... ORDER BY ...)).
+                    WindowFunction.ROW_NUMBER
+                ),
                 Set.of(PARQUET_DATA_FORMAT)
             )
         );
--- a/sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/ProjectRuleTests.java
+++ b/sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/ProjectRuleTests.java
@@ -8,12 +8,18 @@
 
 package org.opensearch.analytics.planner;
 
+import com.google.common.collect.ImmutableList;
 import org.apache.calcite.plan.RelOptTable;
 import org.apache.calcite.plan.RelOptUtil;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexFieldCollation;
+import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexOver;
+import org.apache.calcite.rex.RexWindowBounds;
+import org.apache.calcite.rex.RexWindowExclusion;
 import org.apache.calcite.sql.SqlFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
 import org.apache.calcite.sql.SqlKind;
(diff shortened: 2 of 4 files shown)


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/EngineCapability.java
```diff
@@ -21,5 +21,16 @@
...
+     * Window-function evaluation inside a Project expression. A backend that declares this
+     * capability commits to evaluating any {@link org.apache.calcite.rex.RexOver} encountered in a
+     * project list — i.e. window aggregates such as {@code ROW_NUMBER() OVER (PARTITION BY ...
+     * ORDER BY ...)} that PPL's {@code top}/{@code rare}/{@code streamstats} commands lower to.
+     * The capability is intentionally coarse (no per-window-function granularity): the substrait
+     * standard catalog already constrains which window aggregates a backend's substrait consumer
+     * can decode, and failing the runtime decode is a clearer error than spreading a per-function
+     * registry across SPI plus backend.
+     */
+    WINDOW
```
mch2: This declaration should not be required anymore - can u please rebase from main and pick up the window support changes.

ahkcs:
Rebased onto current `main` (`44836306aa6`) and dropped `EngineCapability.WINDOW` per your comment, @mch2. The PR now uses the upstream `WindowCapability` / `WindowFunction` SPI:

- Removed `EngineCapability.WINDOW` enum entry + the matching declaration in `DataFusionAnalyticsBackendPlugin`.
- Removed `annotateWindowCall` from `OpenSearchProjectRule` — `RexOver` is now left unannotated; the existing `narrowByWindowCapability` pass handles backend selection.
- Added `WindowFunction.ROW_NUMBER` to `MockDataFusionBackend.windowCapabilities()` so the top/rare unit tests still resolve a viable backend.
- Updated 3 ProjectRule tests to assert the new flow (RexOver unannotated, error message format change, `liftNestedRexOver` hoisting in stripAnnotations).

Net diff vs main is now ~424 LOC across 4 files. Running `./gradlew check -p sandbox -Dsandbox.enabled=true` locally to verify before CI.

sandeshkr419:
@ahkcs Quick clarification - we are just adding ITs now, right?

ahkcs:
> @ahkcs Quick clarification - we are just adding ITs now, right?

yes

mch2 (approved):

PR #21597, opened 2026-05-11
deletion handling during merge

sachin-27:
### Description
Adds live-docs filtering during Parquet merges so that hard-deleted rows are physically removed from the merged output file.

### What this does:
- Adds getLiveDocsForSegments(List<Segment>) to DeleteExecutionEngine — returns a per-segment bitmap of live rows using an NRT reader over the shared Lucene IndexWriter
- MergeHandler.doMerge calls this at merge start and passes the bitmap through MergeInput to the Rust merger
- The Rust merger (cursor.rs, sorted.rs, unsorted.rs) skips dead rows during the k-way merge, producing a smaller output file
- Fixes LuceneDeleteExecutionEngine.createDeleter to return null for parquet-only writers instead of throwing (was breaking indexing on parquet-only indices)

### Testing
- Added relevant unit / integ tests wherever required
- End to End tests for merge to follow once delete API's are plugged int
- Manual verification: Ran a local single-node cluster with -Dopensearch.pluggable.dataformat.merge.enabled=true and hardcoded map to drop first 2 nodes in each segment. Indexed 50 docs across multiple segments, triggered force-merge, confirmed merged Parquet file had the expected reduced row count and no errors in logs. This exercises the full Java → FFM → Rust live-docs path end-to-end.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
@@ -12,13 +12,14 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.lucene.index.IndexWriter;
 import org.apache.lucene.index.MergeIndexWriter;
-import org.apache.lucene.index.MergePolicy;
+import org.apache.lucene.index.PreparableOneMerge;
 import org.apache.lucene.index.SegmentCommitInfo;
 import org.apache.lucene.index.SegmentInfos;
 import org.opensearch.be.lucene.stats.LuceneShardStatsTracker;
 import org.opensearch.common.SuppressForbidden;
 import org.opensearch.common.annotation.ExperimentalApi;
 import org.opensearch.index.engine.dataformat.DataFormat;
+import org.opensearch.index.engine.dataformat.LiveDocs;
 import org.opensearch.index.engine.dataformat.MergeInput;
 import org.opensearch.index.engine.dataformat.MergeResult;
 import org.opensearch.index.engine.dataformat.Merger;
@@ -41,6 +42,9 @@ import static org.opensearch.be.lucene.index.LuceneWriter.WRITER_GENERATION_ATTR
 /**
  * Lucene-specific {@link Merger} that merges segments using Lucene's internal
  * {@code merge(OneMerge)} path with IndexSort-based document reordering.
+ *
+ * <p>Supports the two-phase prepare/execute flow so a primary-format merger can read
+ * Lucene's frozen live-docs snapshot before {@code mergeMiddle} runs.
  *
  * <h2>How it works</h2>
  *
@@ -93,6 +97,43 @@ public class LuceneMerger implements Merger {
         this.stats = stats;
         // TODO implement primary and integrate the same here
         this.strategy = new SecondaryLuceneMergeStrategy();
+    }
+
+    @Override
+    public LiveDocs prepareMerge(MergeInput mergeInput) throws IOException {
+        List<Segment> segments = mergeInput.segments();
+        if (segments.isEmpty()) {
+            return LiveDocs.ALL_ALIVE;
+        }
+        Set<Long> generationsToMerge = new HashSet<>();
+        for (Segment segment : segments) {
+            generationsToMerge.add(segment.generation());
+        }
+
+        SegmentInfos segmentInfos;
+        try {
+            segmentInfos = (SegmentInfos) SEGMENT_INFOS_FIELD.get(indexWriter);
+        } catch (IllegalAccessException e) {
+            throw new IOException("Failed to access IndexWriter segmentInfos via reflection", e);
+        }
+        if (segmentInfos.size() == 0) {
+            return LiveDocs.ALL_ALIVE;
+        }
+        List<SegmentCommitInfo> matchingSegments = findMatchingSegments(segmentInfos, generationsToMerge);
+        if (matchingSegments.isEmpty()) {
+            return LiveDocs.ALL_ALIVE;
+        }
+
+        // Build with null mapping; merge() sets it once the primary has produced it.
+        PreparableOneMerge oneMerge = strategy.createOneMerge(matchingSegments, null, mergeInput.newWriterGeneration());
+        indexWriter.prepareMerge(oneMerge, mergeInput.newWriterGeneration());
+
+        return LiveDocs.fromPackedBits(packLiveDocsFromMergeReaders(oneMerge));
+    }
+
+    @Override
+    public void abortPreparedMerge(MergeInput mergeInput) throws IOException {
+        indexWriter.abortPreparedMerge(mergeInput.newWriterGeneration());
     }
 
     @Override
@@ -158,16 +199,35 @@ public class LuceneMerger implements Merger {
                 generationsToMerge
             );
 
-            // Delegate OneMerge creation to the strategy (primary vs secondary behavior).
-            // For the secondary path, the returned RowIdRemappingOneMerge stamps the
-            // writer_generation attribute onto the merged SegmentInfo via setMergeInfo, which
-            // Lucene invokes immediately before codec.segmentInfoFormat().write(...) — so the
-            // attribute is persisted to the .si file and survives a writer reopen.
-            MergePolicy.OneMerge oneMerge = strategy.createOneMerge(matchingSegments, rowIdMapping, mergeInput.newWriterGeneration());
+            // Prefer the prepared OneMerge from prepareMerge (its snapshot is what the primary's
+            // bitmap was derived from). Fall back to a fresh OneMerge for code paths that did not
+            // go through the two-phase flow (e.g. tests). For the secondary path, the returned
+            // RowIdRemappingOneMerge stamps the writer_generation attribute onto the merged
+            // SegmentInfo via setMergeInfo, which Lucene invokes immediately before
+            // codec.segmentInfoFormat().write(...) — so the attribute is persisted to the .si
+            // file and survives a writer reopen.
+            PreparableOneMerge prepared = indexWriter.takePreparedMerge(mergeInput.newWriterGeneration());
+            PreparableOneMerge oneMerge;
+            if (prepared != null) {
+                if (prepared instanceof RowIdRemappingOneMerge rowIdMerge) {
+                    rowIdMerge.setRowIdMapping(rowIdMapping);
+                }
+                oneMerge = prepared;
+            } else {
+                oneMerge = strategy.createOneMerge(matchingSegments, rowIdMapping, mergeInput.newWriterGeneration());
+            }
             indexWriter.executeMerge(oneMerge, mergeInput.newWriterGeneration());
 
             // Build the merged WriterFileSet from the output segment info
             SegmentCommitInfo mergedInfo = oneMerge.getMergeInfo();
+
+            // If carryOverHardDeletes produced in-memory deletes for the merged segment
+            // (concurrent deletes that arrived during the merge), force the .liv file to be
+            // written now so mergedInfo.files() includes it. Otherwise the catalog records
+            // a file set without .liv, but a later flush writes the .liv onto disk — and
+            // the next refresh fails with "Catalog segment ... has no matching Lucene leaf".
+            indexWriter.flushLiveDocsForMergedSegment(mergedInfo);
+
             WriterFileSet mergedFileSet = buildMergedFileSet(mergedInfo, mergeInput.newWriterGeneration());
 
             // Delegate RowIdMapping production to the strategy
@@ -189,6 +249,19 @@ public class LuceneMerger implements Merger {
         } finally {
             stats.addMergeTimeMillis(TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start));
         }
+    }
+
+    /**
+     * Builds the per-segment live-docs map from a {@link PreparableOneMerge} whose merge
+     * readers have been populated by {@link org.apache.lucene.index.MergeIndexWriter#prepareMerge}.
+     * Reads the frozen {@code hardLiveDocs} via {@link PreparableOneMerge#packLiveDocsByGeneration}
+     * so the bitmap matches the snapshot Lucene's {@code mergeMiddle} will use for physical drops.
+     */
+    private static Map<Long, long[]> packLiveDocsFromMergeReaders(PreparableOneMerge oneMerge) {
+        return oneMerge.packLiveDocsByGeneration(sci -> {
+            String genAttr = sci.info.getAttribute(WRITER_GENERATION_ATTRIBUTE);
+            return genAttr == null ? -1L : Long.parseLong(genAttr);
+        });
     }
 
     /**
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMergeExecutor.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMergeExecutor.java
@@ -41,6 +41,13 @@ public class CompositeMergeExecutor {
     }
 
     /**
+     * Returns the merger registered for the given data format, or {@code null} if absent.
+     */
+    public Merger getMerger(DataFormat format) {
+        return mergers.get(format);
+    }
+
+    /**
      * Executes the merge described by the plan.
      *
      * @param plan the pre-validated merge plan
@@ -105,7 +112,14 @@ public class CompositeMergeExecutor {
         for (WriterFileSet wfs : files) {
             segments.add(Segment.builder(wfs.writerGeneration()).addSearchableFiles(format, wfs).build());
         }
-        MergeResult result = merger.merge(new MergeInput(segments, mapping, plan.mergedWriterGeneration()));
+        MergeResult result = merger.merge(
+            MergeInput.builder()
+                .segments(segments)
+                .rowIdMapping(mapping)
+                .newWriterGeneration(plan.mergedWriterGeneration())
+                .liveDocs(plan.liveDocs())
+                .build()
+        );
         return new FormatMergeResult(format, result.getMergedWriterFileSetForDataformat(format), result.rowIdMapping().orElse(null));
     }
 
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMerger.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMerger.java
@@ -14,6 +14,7 @@ import org.opensearch.composite.CompositeIndexingExecutionEngine;
 import org.opensearch.composite.stats.CompositeShardStatsTracker;
 import org.opensearch.index.engine.dataformat.DataFormat;
 import org.opensearch.index.engine.dataformat.IndexingExecutionEngine;
+import org.opensearch.index.engine.dataformat.LiveDocs;
 import org.opensearch.index.engine.dataformat.MergeInput;
 import org.opensearch.index.engine.dataformat.MergeResult;
 import org.opensearch.index.engine.dataformat.Merger;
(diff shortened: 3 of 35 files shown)


sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMergeExecutor.java
```diff
@@ -76,7 +76,14 @@ private FormatMergeResult mergeFormat(MergePlan plan, DataFormat format, RowIdMa
         for (WriterFileSet wfs : files) {
             segments.add(Segment.builder(wfs.writerGeneration()).addSearchableFiles(format, wfs).build());
         }
-        MergeResult result = merger.merge(new MergeInput(segments, mapping, plan.mergedWriterGeneration()));
+        MergeResult result = merger.merge(
+            MergeInput.builder()
```
RS146BIJAY, line 116: Do not change anything in the abstract.

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
```diff
@@ -219,7 +229,13 @@ public MergeResult doMerge(OneMerge oneMerge) throws IOException {
         assert oneMerge.getSegmentsToMerge().isEmpty() == false : "merge must have at least one segment";
         long generation = generationProvider.get();
         assert generation > 0 : "merge writer generation must be positive but was: " + generation;
-        MergeInput mergeInput = MergeInput.builder().segments(oneMerge.getSegmentsToMerge()).newWriterGeneration(generation).build();
+        Map<Long, long[]> liveDocs = deleter == null ? Map.of() : deleter.getLiveDocs(oneMerge.getSegmentsToMerge());
```
RS146BIJAY: Maybe this will become more clear once deleter integration is complete.

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMergeExecutor.java
sachin-27: Its just switched from constructor to builder since its available. I can revert this

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
sachin-27: Yes, basically all the downstream codechanges in rust rely on this Map<Long, long[]> for live bitset. Long -> Segment, long[] -> bitset depicting if row is alive. each long stores 64 bits

Bukhtawar (commented):
Lets add an integ test for this, include assertions. overall looks good

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/unsorted.rs
```diff
@@ -143,3 +170,16 @@ pub fn merge_unsorted(
...
+fn is_alive(bits: &[u64], abs_row_id: u64, num_rows: u64) -> bool {
+    if abs_row_id >= num_rows {
+        return true;
+    }
+    let word = (abs_row_id / 64) as usize;
+    let bit = (abs_row_id % 64) as u64;
+    match bits.get(word) {
+        Some(&w) => (w & (1u64 << bit)) != 0,
+        None => true,
+    }
```
Bukhtawar: This logic is being duplicated with cursor

sachin-27:
> Lets add an integ test for this, include assertions. overall looks good

Actually its not possible to write an integ test for this since the deletetionengine changes are not wired yet. I can introduce them in a follow up PR once the wiring changes ared added

server/src/main/java/org/opensearch/index/engine/dataformat/LiveDocs.java
```diff
@@ -0,0 +1,111 @@
...
+ * (to drop dead rows from the output).
+ *
+ * <p>The internal representation uses Lucene's {@code FixedBitSet#getBits()} layout
+ * (packed {@code long[]} where bit {@code i} set = row {@code i} alive), but consumers
+ * interact through the typed API without knowing the wire format.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public final class LiveDocs {
```
RS146BIJAY, line 31: Can you elaborate why Format-neutral live-docs is needed here?

RS146BIJAY:
So merge does not necessarly mean will be called via a scheduler, but we can directly get a merger from Engine and call merger over it. I don't think current approach for maintaining context for COW handles this? Can you compare the abstractions with Lucene and check how it handle this scenario: https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L5190 , and frame an approach on this.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -91,6 +95,43 @@ public LuceneMerger(MergeIndexWriter indexWriter, DataFormat dataFormat, Path st
...
+        // Build with null mapping; merge() sets it once the primary has produced it.
+        PreparableOneMerge oneMerge = strategy.createOneMerge(matchingSegments, null, mergeInput.newWriterGeneration());
+        indexWriter.prepareMerge(oneMerge, mergeInput.newWriterGeneration());
+
+        return LiveDocs.fromPackedBits(packLiveDocsFromMergeReaders(oneMerge));
+    }
+
+    @Override
+    public void abortPreparedMerge(MergeInput mergeInput) throws IOException {
+        indexWriter.abortPreparedMerge(mergeInput.newWriterGeneration());
```
RS146BIJAY, line 136: Why this override is needed here?

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMerger.java
```diff
@@ -51,16 +52,101 @@ public CompositeMerger(CompositeIndexingExecutionEngine engine, CompositeDataFor
...
+    private static long totalRowsForSegment(Segment seg) {
+        long total = 0L;
+        for (var wfs : seg.dfGroupedSearchableFiles().values()) {
+            total = Math.max(total, wfs.numRows());
+        }
+        return total;
+    }
+
+    @Override
+    public LiveDocs prepareMerge(MergeInput mergeInput) throws IOException {
```
RS146BIJAY, line 119: So since Deleter changes are merged now, can we add an end to end test cases? I would suggest you to see if we can pick up few test case scenarios from IndexWriter merges: https://github.com/apache/lucene/tree/main/lucene/core/src/test/org/apache/lucene/index. I mean similar test case scenarios can be written for DataFormatAwareEngine to see if merges are correctly excluding deleted documents.

mgodwan:
Lucene handles deletes arriving between snapshot-freeze and merge commit via carryOverHardDeletes. But the Parquet primary merger only seems to have the frozen bitmap. Due to this, a doc deleted while the merge runs will end up surviving in the merged Parquet file with no visible mechanism to apply the delta. @sachin-27 Could you confirm this?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/NativeParquetMergeStrategy.java
```diff
@@ -80,24 +80,48 @@ public MergeResult mergeParquetFiles(MergeInput mergeInput) {
...
+                filePaths,
+                liveBitsArg,
+                mergedFilePath.toString(),
+                indexName,
+                writerGeneration
+            );
             ParquetFileMetadata mergeMetadata = merged.metadata();
             RowIdMapping rowIdMapping = merged.rowIdMapping();
 
             assert mergeMetadata.numRows() > 0 : "Merged file should contain at least one row";
```
mgodwan: Will this not fail for all deleted?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -95,6 +99,43 @@ public LuceneMerger(MergeIndexWriter indexWriter, DataFormat dataFormat, Path st
...
+            return LiveDocs.ALL_ALIVE;
+        }
+        Set<Long> generationsToMerge = new HashSet<>();
+        for (Segment segment : segments) {
+            generationsToMerge.add(segment.generation());
+        }
+
+        SegmentInfos segmentInfos;
+        try {
+            segmentInfos = (SegmentInfos) SEGMENT_INFOS_FIELD.get(indexWriter);
```
mgodwan: Can this have any races? Do we need this to be synchronized?

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
```diff
@@ -41,14 +44,169 @@
...
+     * a file set without {@code .liv}; later when a flush triggers the {@code .liv}
+     * write, the leaf has one extra file and the catalog/leaf mismatch surfaces as
+     * {@code "Catalog segment ... has no matching Lucene leaf"} on the next refresh.
+     */
+    public synchronized void flushLiveDocsForMergedSegment(SegmentCommitInfo mergedInfo) throws IOException {
+        ReadersAndUpdates rld = getPooledInstance(mergedInfo, false);
+        if (rld == null) {
+            return;
+        }
+        rld.writeLiveDocs(getDirectory());
```
mgodwan, line 133: Does this write need to be synchronized? What is the overhead here as this may stall refresh flow as well for larger time duration? Trying to see how we can reduce this

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/sorted.rs
```diff
@@ -202,15 +218,22 @@ pub fn merge_sorted_with_pool(
...
-                    for _ in 0..remaining {
-                        mapping[file_offset + rows_emitted_per_file[file_id]] = new_row_id;
-                        rows_emitted_per_file[file_id] += 1;
-                        new_row_id += 1;
+                    let slice = cursor.take_slice(start_idx, remaining, reservation)?;
+                    // Build mapping for surviving rows (indexed by absolute source row id).
+                    for i in 0..remaining {
+                        let abs = base_row_id + (start_idx + i) as u64;
+                        if cursor.is_row_id_alive(abs) {
+                            mapping[file_offset + abs as usize] = new_row_id;
```
mgodwan: Can we add guard rails for array index (add an assertion maybe)?

server/src/main/java/org/opensearch/index/engine/dataformat/LiveDocs.java
sachin-27: this was a suggestion by @Bukhtawar to extract the logic into a class

sachin-27:
> Lucene handles deletes arriving between snapshot-freeze and merge commit via carryOverHardDeletes. But the Parquet primary merger only seems to have the frozen bitmap. Due to this, a doc deleted while the merge runs will end up surviving in the merged Parquet file with no visible mechanism to apply the delta. @sachin-27 Could you confirm this?

You are right about lucene handling deletes via carryOverHardDeletes. However the same CarryOverHardDeletes flushes these deleted rows in liv docs file. The deletes present in this liv docs file will be cleaned up in the following merge based on changes in this PR. 

TLDR: We do not need to specifrically handle this scenario since we are relying on lucene liv docs for source of truth and that is already handling this

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
sachin-27: yes, it is possible. Fixed it now

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
sachin-27: It doesn't — removed the `synchronized`, so the `.liv` disk write can no longer
stall a refresh/flush waiting on the writer monitor. Just checked that the writelivedocs synchronizes on the per segment instances

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/sorted.rs
sachin-27: added

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/NativeParquetMergeStrategy.java
sachin-27: Correct. Changed the assertion now to - `anyLiveDocs || numRows > 0`. One gap here though is that when all docs are deleted, lucene detects and drops the segment whereas parquet writes an empty file. Since this PR is primarily focussed on handling and applying the liv docs during merge, I can raise a separate issue to track this. Let me know your thoughts.

server/src/main/java/org/apache/lucene/index/MergeIndexWriter.java
```diff
@@ -41,14 +47,208 @@
...
+        } catch (Throwable t) {
+            // Ride Lucene's own abort path: setAborted + merge fast-fails inside mergeMiddle's
+            // checkAborted and the surrounding finally runs closeMergeReaders, which closes any
+            // readers we managed to open, decRefs their files (balancing our per-reader incRef),
+            // and clears the setIsMerging flags. We swallow the expected MergeAbortedException
+            // and suppress any other cleanup failure into the original throwable.
+            try {
+                oneMerge.setAborted();
+                merge(oneMerge);
+            } catch (MergePolicy.MergeAbortedException expected) {
```
Shailesh-Kumar-Singh, line 207: This catch is currently unreachable. `IndexWriter#handleMergeException` only rethrows
`MergeAbortedException` when `merge.isExternal` is true.
Here `prepareMerge` sets `oneMerge.isExternal = false`.

We can consider to keep this as is because Lucene has a [TODO](https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/IndexWriter.java#L4929) to always throw, we can just mention the same reasoning here as a comment.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/merge/LuceneMerger.java
```diff
@@ -158,16 +172,35 @@ public MergeResult merge(MergeInput mergeInput) throws IOException {
...
+            // file and survives a writer reopen.
+            PreparableOneMerge prepared = indexWriter.takePreparedMerge(mergeInput.newWriterGeneration());
+            PreparableOneMerge oneMerge;
+            if (prepared != null) {
+                if (prepared instanceof RowIdRemappingOneMerge rowIdMerge) {
+                    rowIdMerge.setRowIdMapping(rowIdMapping);
+                }
+                oneMerge = prepared;
+            } else {
+                oneMerge = strategy.createOneMerge(matchingSegments, rowIdMapping, mergeInput.newWriterGeneration());
```
Shailesh-Kumar-Singh, line 190: The fallback re-reads live docs in `mergeMiddle`, so Lucene would filter against a different snapshot
than Parquet used. Its not reachable today (`CompositeMerger` always prepares; `CompositeMergeExecutor`), but since `executeMerge` already detects this via
`p.readersInitialized()`, can we throw here instead of silently recomputing?

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/live_docs.rs
```diff
@@ -0,0 +1,169 @@
...
+        None => true,
+    }
+}
+
+/// Builds a boolean filter mask for a batch of `batch_len` rows starting at `base_row_id`.
+///
+/// Returns `None` if all rows in the batch are alive (caller should skip filtering).
+/// Otherwise returns `(mask, alive_count)`.
+#[inline]
+pub fn build_filter_mask(
```
Shailesh-Kumar-Singh, line 51: unused function, please remove.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/live_docs.rs
```diff
@@ -0,0 +1,169 @@
...
+/// Per-file live-docs filter. `None` = all rows alive (no filtering needed).
+pub type LiveBits = Option<Arc<Vec<u64>>>;
+
+/// Returns `true` if the row at `abs_row_id` is alive.
+///
+/// - `None` bits → all alive
+/// - Row beyond `num_rows` → alive (defensive)
+/// - Word beyond bitset length → alive (defensive)
+#[inline]
+pub fn is_alive(bits: &LiveBits, num_rows: u64, abs_row_id: u64) -> bool {
```
Shailesh-Kumar-Singh, line 25: unused function, please remove.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/live_docs.rs
```diff
@@ -0,0 +1,169 @@
...
+            } else {
+                Some((mask, alive_count))
+            }
+        }
+    }
+}
+
+/// Counts total alive rows in `[0, num_rows)`.
+/// Returns `num_rows` if bits is `None`.
+pub fn count_alive(bits: &LiveBits, num_rows: u64) -> u64 {
```
Shailesh-Kumar-Singh, line 80: unused function, please remove.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/cursor.rs
```diff
@@ -47,9 +53,81 @@ pub struct FileCursor {
...
+    base_row_id: u64,
+    batch_upper: usize,
+) -> usize {
+    match live_bits {
+        None => batch_upper,
+        Some(_) => {
+            let mut i = batch_upper;
+            while i > 0 {
+                let idx = i - 1;
+                if is_row_id_alive(live_bits, num_rows, base_row_id + idx as u64) {
```
Shailesh-Kumar-Singh, line 94: Instead of checking one bit at a time, we can look at a whole 64-bit word and use `leading_zeros()` to find the highest live row in it. This will be faster when there's a long run of dead rows at the end of the batch.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/merge/NativeParquetMergeStrategy.java
```diff
@@ -80,24 +80,56 @@ public MergeResult mergeParquetFiles(MergeInput mergeInput) {
...
+                    "Merge at generation [{}] produced empty file [{}] — every input row was deleted",
+                    writerGeneration,
+                    mergedFileName
+                );
+            }
 
             long expectedRows = files.stream().mapToLong(MonoFileWriterSet::numRows).sum();
-            assert mergeMetadata.numRows() == expectedRows : "Merged row count ["
+            // With live-docs, output is bounded by input sum. Without, strict equality.
+            assert mergeMetadata.numRows() <= expectedRows : "Merged row count ["
```
Shailesh-Kumar-Singh, line 123: Are we not weakening the check here in the sense that if we are losing rows, this will silently pass. 
Can we check the exact expected row count by summing the set bits per input?

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
rayshrey, line 1: Unintended changes ?

server/src/main/java/org/opensearch/index/engine/dataformat/Merger.java
```diff
@@ -26,4 +26,27 @@ public interface Merger {
...
+        return LiveDocs.ALL_ALIVE;
+    }
+
+    /**
+     * Releases resources acquired by {@link #prepareMerge} when the subsequent
+     * {@link #merge} call won't happen. Default no-op.
+     */
+    default void abortPreparedMerge(MergeInput mergeInput) throws IOException {
+        // no-op
+    }
```
rayshrey, line 51: Seems like a leaky interface where it becomes the responsibility of the callers to cleanup the resources acquired by prepareMerge.

Also prepareMerge does not seem the right name, if we are returning LiveDocs we can directly have something delete related. Maybe if only one plugin will be implementing this merge preparation, then we can have prepareMerge in a separate interface altogether and have only the LuceneMerger implement this.

server/src/main/java/org/opensearch/index/engine/dataformat/Merger.java
rayshrey, line 1: In continuation with the previous comment, was thinking of something like this.

```
  public interface Merger {
  
      /**
       * Whether this format owns delete state and can freeze it for the duration of a merge.
       * At most one format in a composite may return true; validated when the composite is built.
       */
      default boolean providesMergeLiveDocs() { return false; }
  
      /**
       * Phase 1. Freezes this format's delete state for the merge inputs and publishes the frozen
       * view, so the primary drops exactly the rows this format will drop. Implementations must
       * derive the view from the merge's frozen readers, not from a fresh reader.
       */
      default MergePreparation prepareMerge(MergeInput in) throws IOException {
          return MergePreparation.EMPTY;
      }
  
      MergeResult merge(MergeInput in) throws IOException;   // unchanged
  }
  
  /**
   * Mirrors {@link MergeResult}: what phase 1 produces. Closeable because phase 1 pins merge
   * readers; close() releases anything the subsequent merge did not consume and is a no-op
   * once merge() has taken the prepared state.
   */
  public interface MergePreparation extends Closeable {
      Optional<LiveDocs> liveDocs();
      @Override void close() throws IOException;      // idempotent
      MergePreparation EMPTY = /* Optional.empty(), no-op close */;
  }
  
```

And in Composite something like this

```
  public CompositeMerger(CompositeIndexingExecutionEngine engine, CompositeDataFormat format) {
      ...
      this.liveDocsProducer = resolveLiveDocsProducer(secondaryFormats, executor);   // nullable
  }
  
  /** Mirrors DataFormatRegistry#getDeleteExecutionEngine: opt-in providers, at most one, fail fast. */
  private static DataFormat resolveLiveDocsProducer(List<DataFormat> secondaries, CompositeMergeExecutor executor) {
      List<DataFormat> producers = secondaries.stream()
          .filter(f -> { Merger m = executor.getMerger(f); return m != null && m.providesMergeLiveDocs(); })
          .toList();
      if (producers.size() > 1) {
          throw new IllegalStateException("Multiple secondary formats provide merge live-docs, expected at most one but found "
              + producers.stream().map(DataFormat::name).toList());
      }
      return producers.isEmpty() ? null : producers.getFirst();
  }
  
  @Override
  public MergeResult merge(MergeInput in) throws IOException {
      Merger producer = liveDocsProducer == null ? null : executor.getMerger(liveDocsProducer);
      try (MergePreparation prep = producer == null ? MergePreparation.EMPTY : producer.prepareMerge(in)) {
          MergePlan plan = new MergePlan(in.newWriterGeneration(), primaryFormat, secondaryFormats,
                                         extractFilesByFormat(in.segments()),
                                         prep.liveDocs().orElse(LiveDocs.ALL_ALIVE));
          return executor.execute(plan);
      }
  }

```

sachin-27:
## Delete/update-during-merge benchmark (composite Parquet engine)

Churn run on a live composite (Parquet + Lucene) node at ~12% updates + ~12% deletes, with a force-merge as the delete-during-merge event. **Correctness: PASS, 0 errors.**

Purpose-built correctness harness, not a stock workload: no stock OSB workload has a delete-by-percentage op, and proving an exact final count needs known ids. Corpus uses explicit `_id`s; OSB drove create+update via a custom `nyc_taxis`-schema procedure; deletes + force-merge driven by id.

100k creates + 12k updates (`doc-0..11999`) + 12k deletes (`doc-88000..99999`, disjoint). Expected final live docs = 100,000 − 12,000 = **88,000**.

The engine defers same-`_id` dedup and delete application to **merge time**. Pre-merge the primary physically holds **112,000 Parquet rows** (100k unique + 12k stale update copies), deletes tracked logically in the Lucene `.liv` — this is the physical/stored row count, not a live-doc count (this engine has no `_search`/`_count` path yet; counts read from `_stats/docs`). The force-merge reconciled the duplicates **and** applied the deletes in one pass → exactly **88,000** in one clean segment, `docs.deleted=0`, no drift/races.

<details><summary>Config &amp; results</summary>

- Composite index, 1 shard / 0 replicas, `index.append_only.enabled=false` (required for custom-id updates/deletes).
- OSB 2.4.0 for create+update (bulk, 4 clients); by-id `_bulk` deletes; `POST /_forcemerge?max_num_segments=1`.

| Phase | Median throughput | p50 | p90 |
|---|---|---|---|
| creates (100k) | 17,878 docs/s | 300 ms | 572 ms |
| updates (12k) | 24,485 docs/s | 142 ms | 165 ms |
| deletes (12k) | ~29,600/s (0.41 s) | — | — |

Force-merge 5 → 1 segment: 376 ms.

| Checkpoint | primary rows | docs.deleted | segments |
|---|---|---|---|
| after creates+updates | 112,000 | 0 | 5 |
| after 12k deletes + refresh | 112,000 | 0 | 5 |
| after force-merge to 1 | 88,000 | 0 | 1 |

Spot checks: `doc-5` (updated) → v2, new value; `doc-90000` (deleted) → not found; `doc-50000` → found.
</details>

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMerger.java
```diff
@@ -43,24 +45,114 @@ public class CompositeMerger implements Merger {
     private final List<DataFormat> secondaryFormats;
     private final CompositeMergeExecutor executor;
     private final CompositeShardStatsTracker statsTracker;
+    /** The one secondary whose {@link Merger#providesMergeLiveDocs()} is true, or {@code null}. */
+    private final DataFormat liveDocsProducer;
```
rayshrey: Instead of DataFormat type can we have it simply as the Merger for that dataformat ?

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/MergesWithDeletesIT.java
rayshrey, line 1: Let's rename this to MergesWithDeletesIT.

Since the search support for deleted docs is now added, can we add basic search support in the ITs as well ? Basic sanity would be enough.

Also can we have a concurrent ingestion test as well (multiple threads sending traffic to same index with multiple shards in parellel, each thread doing ingestion of few docs, refresh, deletion of few documents and refresh again) and then asserting number of doc present through search results.

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMerger.java
sachin-27: changed

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/MergesWithDeletesIT.java
sachin-27: Renamed and added a concurrent ingestion test.

On the search assertion: search on a composite index is only plugged in via DataFusion, and for it to filter deletes the Lucene backend must be loaded too. There's no easy way to load both in a cluster IT today

rayshrey (approved):
Thanks for the changes, LGTM.
Approved

PR #21599, opened 2026-05-11
[sandbox] feat: Add on-demand jemalloc heap profiling support via FFM

5u5hma:
Add on-demand jemalloc heap profiling via JMX MBean and a `jcmd`-equivalent CLI tool (`bin/opensearch-heap-prof`). Connects to the running OpenSearch process via JVM Attach API — no REST API, no cluster settings, no cluster state dependency.

### Architecture

```
bin/opensearch-heap-prof CLI                 OpenSearch JVM
┌──────────────────────────┐                 ┌─────────────────────────────┐
│ opensearch-heap-prof     │── JMX Attach ───│ NativeHeapProfilerMBean     │
│   start                  │   (same as      │   activate() → FFM → Rust   │
│   dump /tmp/x.heap       │    jcmd)        │   dump(path) → FFM → Rust   │
│   stop                   │                 │   deactivate()              │
│   status                 │◄── response ────│   isActive()                │
└──────────────────────────┘                 └─────────────────────────────┘
                                                        │ FFM downcall
                                                        ▼
                                             ┌─────────────────────────────┐
                                             │ allocator.rs (Rust)         │
                                             │   tikv_jemalloc_ctl::       │
                                             │     prof.active / prof.dump │
                                             └─────────────────────────────┘
```

```
distribution/tools/heap-prof-cli/src/main/java/org/opensearch/tools/cli/
├── jmx/
│   └── JmxCommand.java          ← Generic: JVM Attach + JMX connection (reusable)
└── heapprof/
    ├── HeapProfCli.java          ← CLI entry point (MultiCommand)
    ├── HeapProfCommand.java      ← Heap-specific: MBean lookup + validation
    ├── StartCommand.java
    ├── StopCommand.java
    ├── DumpCommand.java
    ├── ResetCommand.java
    └── StatusCommand.java
```

### Changes

- **.cargo/config.toml**: `JEMALLOC_SYS_WITH_MALLOC_CONF="prof:true,prof_active:false,lg_prof_sample:17"`
- **Cargo.toml**: `tikv-jemallocator = "=0.6.1"` with `profiling` feature
- **allocator.rs**: 4 `extern "C"` FFI functions (activate/deactivate/dump/reset) + tests
- **NativeHeapProfilerMBean.java** (new): MBean interface
- **NativeHeapProfiler.java**: MBean implementation — bridges JMX to Rust FFI via FFM
- **NativeBridgeModule.java**: Registers MBean at module init
- **distribution/tools/heap-prof-cli/** (new): CLI tool using JVM Attach API + JMX
- **distribution/src/bin/opensearch-heap-prof** (new): Launcher script

### CLI Usage

```bash
# Auto-discovers running OpenSearch process (like jcmd)
bin/opensearch-heap-prof start
bin/opensearch-heap-prof dump /tmp/before.heap
# ... run workload ...
bin/opensearch-heap-prof dump /tmp/after.heap
bin/opensearch-heap-prof stop
bin/opensearch-heap-prof status
bin/opensearch-heap-prof reset 15    # higher accuracy sampling

# Specify PID explicitly (if multiple JVMs running)
bin/opensearch-heap-prof start --pid 12345

# Analyze locally
jeprof --text --lines --base=/tmp/before.heap libopensearch_native.so /tmp/after.heap | rustfilt
```

### Performance Impact

| State | CPU Overhead | Memory Overhead |
|-------|-------------|-----------------|
| `prof:true, prof_active:false` (default) | 0% | ~1-3 MB larger binary |
| `prof_active:true, lg_prof_sample:17` | ~1-2% | ~2 words per sampled allocation |
| `prof_active:true, lg_prof_sample:19` | <1% | minimal |
| Dump operation | momentary pause ~10-100ms | dump file 50KB-5MB |

### Security

- JMX Attach requires same-user access (OS-level security, same as jcmd)
- No network ports opened (local JMX only)
- No cluster-wide API surface — single node only
- Heap dumps contain only allocation stack traces and byte counts (no customer data)

### Why JMX over other approaches

| Approach | Issue |
|----------|-------|
| Cluster settings | Affects all nodes, requires cluster state, exposes customer-facing API |
| REST endpoint | Requires HTTP server healthy + cluster joined |
| Unix domain socket | Custom protocol, needs socat for testing, non-standard |
| **JMX (chosen)** | Standard JVM mechanism, same as jcmd, auto-discovers PID, no dependencies |

### Description

On-demand jemalloc heap profiling for the native (Rust/DataFusion) layer of OpenSearch. Provides a `jcmd`-equivalent CLI tool that connects to the running process via JMX to activate profiling, capture heap dumps, and change sample rates — all node-local with no cluster dependency.

### Check List


5u5hma:
> ## PR Code Analyzer ❗
> **AI-powered '_Code-Diff-Analyzer_' found issues on commit [d3e95af](https://github.com/opensearch-project/OpenSearch/commit/d3e95affacd94e63725f62f20e0d4ff2c2dc4ef0).**
> 
> Path	Line	Severity	Description
> sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java	110	medium	Heap profile dumps are restricted to the node's data directory, but the path validation uses startsWith() on normalized Path objects. On case-insensitive filesystems (e.g., macOS HFS+), this could be bypassed with mixed-case paths. The intent appears legitimate (diagnostic profiling), but the restriction mechanism has a known edge-case weakness.
> sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java	30	medium	The Javadoc explicitly acknowledges that heap dumps may contain sensitive in-memory data (credentials, index documents, keys). While the dump path is gated by OperatorDynamic privilege and restricted to the data directory, this feature introduces a new privileged code path that allows operator-triggered file writes containing potentially sensitive memory contents. This is consistent with the stated purpose but warrants careful access control review.
> The table above displays the top 10 most important findings. **Total: 2 | Critical: 0 | High: 0 | Medium: 2 | Low: 0**
> 
> **Pull Requests Author(s)**: Please update your Pull Request according to the report above.
> 
> **Repository Maintainer(s)**: You can `bypass diff analyzer` by adding label `skip-diff-analyzer` after reviewing the changes carefully, then `re-run failed actions`. To re-enable the analyzer, remove the label, then `re-run all actions`.
> 
> ⚠️ Note: The _Code-Diff-Analyzer_ helps protect against potentially harmful code patterns. Please ensure you have thoroughly reviewed the changes beforehand.
> 
> Thanks.

Path validation startsWith() on case-insensitive filesystems (medium)

"Valid edge case, though OpenSearch's supported production environments are Linux-only (ext4/xfs — case-sensitive). macOS is dev/test only. The Path.normalize().startsWith() approach is the same pattern used by OpenSearch's Environment.validateFileSetting() and Elasticsearch's path security checks. For additional hardening, we could resolve symlinks via Path.toRealPath() — happy to add that if reviewers want it, though it would fail on non-existent parent directories. In practice, the data directory is always a real path on Linux with no case ambiguity."

Operator-triggered sensitive memory dump (medium)

"Acknowledged. This is by design — the feature's explicit purpose is to enable operators to capture native heap state for memory leak diagnosis. The same capability already exists via OS-level tools (gcore, /proc/pid/mem, jmap -dump:live) available to any user with the same privilege level. This PR adds no new capability that operators don't already have — it adds a convenient, auditable, sandboxed path to exercise it. Access controls in place: OperatorDynamic privilege gate, path restricted to data dir, chmod 600 on output, null-guard during init window."

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
```diff
@@ -0,0 +1,138 @@
...
+        }
+        if (allowedDumpDir == null) {
+            logger.warn("Heap dump rejected: module not yet initialized");
+            return;
+        }
+        // Path validation: prevent arbitrary file writes
+        java.nio.file.Path resolved;
+        try {
+            resolved = java.nio.file.Path.of(path).toAbsolutePath().normalize();
+            java.nio.file.Path allowed = java.nio.file.Path.of(allowedDumpDir).toAbsolutePath().normalize();
```
himshikhagupta: 1. Should we have some standard path which would always get these dumps? Building tolling on it will be easier with standard path.
2. Are we also planning to upload to S3 or will the operator have to copy from nodes? What if the nodes go down before someone copies?

sandbox/libs/dataformat-native/rust/lib/src/lib.rs
```diff
@@ -23,8 +23,13 @@
...
 ///   only used until that point.
 /// - lg_tcache_max: NOT dynamically tunable by jemalloc — init-time only, requires process restart to change.
+/// - prof:true,prof_active:false: Enables profiling infrastructure at startup but keeps sampling
+///   inactive (zero overhead). This is required so that operators can activate/deactivate profiling
+///   at runtime via cluster settings without restarting the process. jemalloc does not allow
+///   enabling prof after initialization — it must be set at startup or not at all.
+/// - lg_prof_sample:17: sample every ~128KB of allocation when profiling is active.
 #[export_name = "malloc_conf"]
-pub static MALLOC_CONF: &[u8] = b"dirty_decay_ms:30000,muzzy_decay_ms:30000,lg_tcache_max:16\0";
+pub static MALLOC_CONF: &[u8] = b"dirty_decay_ms:30000,muzzy_decay_ms:30000,lg_tcache_max:16,prof:true,prof_active:false,lg_prof_sample:17\0";
```
himshikhagupta: can you add performance benchmarks in details on if there is some impact of always having prof true, and when prof_active is enabled?

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
5u5hma: 1. Should we have some standard path which would always get these dumps? Building tolling on it will be easier with standard path.
- It is currently using the data path, had initially though of setting it to /tmp and had written SOP accordingly to set it to /tmp but there was a PR bot comment saying sensitive domain data should not be added outside data path. If there is a example I can leverage I can check that out and can update accordingly. 
2. Are we also planning to upload to S3 or will the operator have to copy from nodes? What if the nodes go down before someone copies?
Copy from node and upload and is in synonymous to existing JFR thread dump tooling. https://quip-amazon.com/M1f9AaAc1hIt/SOP-Getting-JFR-from-OpenSearch-Node. Here also you get it from the host and put object to s3 or from mechanic host. Can check on what would it take to upload to s3 from node will check on how asprof is doing it.

sandbox/libs/dataformat-native/rust/lib/src/lib.rs
5u5hma: sure. Will add expected is 
prof_active: true with lg_prof_sample:17 | ~1-2% CPU overhead, minimal memory overhead

5u5hma:
> ## PR Code Analyzer ❗
> **AI-powered '_Code-Diff-Analyzer_' found issues on commit [fb85403](https://github.com/opensearch-project/OpenSearch/commit/fb85403a7bc0970b658c92cf4562ecb6eb38ff42).**
> 
> Path	Line	Severity	Description
> sandbox/libs/dataformat-native/rust/Cargo.toml	55	high	Dependency feature modification: 'profiling' feature added to tikv-jemallocator. Per mandatory supply chain rule, any dependency change must be flagged. Adding the 'profiling' feature changes the compiled artifact and enables jemalloc heap profiling APIs in the final binary. Maintainers should verify the tikv-jemallocator crate at version 0.6 with this feature is the expected artifact.
> sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java	148	medium	Heap dump triggered by dynamic cluster setting (native.jemalloc.heap_prof_dump_path) writes raw JVM heap contents — including in-memory credentials, index documents, and cryptographic keys — to disk. The code itself acknowledges this risk. An attacker with cluster operator privileges can trigger dumps on demand. The allowlist provides some containment but the path is dynamically configurable and executes immediately on setting change.
> sandbox/libs/dataformat-native/build.gradle	82	medium	JEMALLOC_SYS_WITH_MALLOC_CONF with 'prof:true' is compiled into all release production binaries. While prof_active:false suppresses runtime overhead, the profiling infrastructure is permanently embedded and activatable at runtime via cluster settings without a process restart. This expands the production attack surface for any operator-level access that can flip native.jemalloc.heap_prof_active to true.
> The table above displays the top 10 most important findings. **Total: 3 | Critical: 0 | High: 1 | Medium: 2 | Low: 0**
> 
> **Pull Requests Author(s)**: Please update your Pull Request according to the report above.
> 
> **Repository Maintainer(s)**: You can `bypass diff analyzer` by adding label `skip-diff-analyzer` after reviewing the changes carefully, then `re-run failed actions`. To re-enable the analyzer, remove the label, then `re-run all actions`.
> 
> ⚠️ Note: The _Code-Diff-Analyzer_ helps protect against potentially harmful code patterns. Please ensure you have thoroughly reviewed the changes beforehand.
> 
> Thanks.

The access is controlled through internal tool command 2PR and customer consent and the allowlisted paths are controlled through yml config file and just the data path for the opensource side so this should protect the attack surface area and also the dump recorded being safe enough.

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
5u5hma: Updated the paths and did add the tooling accordingly.

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
```diff
@@ -111,6 +111,67 @@ fn set_all_arenas(suffix: &[u8], ms: i64) -> Result<i64, String> {
...
+
+/// FFI: Resets profiling state and sets a new sample interval.
+/// Discards all accumulated profiling data and applies the new lg_prof_sample value
+/// for future allocations. Returns 0 on success, negative error pointer on failure.
+///
+/// Common values: 15 (~32KB, high accuracy), 17 (~128KB, default), 19 (~512KB, low overhead).
+#[no_mangle]
+pub extern "C" fn native_jemalloc_heap_prof_reset(lg_sample: usize) -> i64 {
+    ffm_wrap("native_jemalloc_heap_prof_reset", || {
+        unsafe { tikv_jemalloc_ctl::raw::write(b"prof.reset\0", lg_sample) }
```
himshikhagupta, line 169: Is this for changing lg_sample field? What is being reset? 
For my understanding is lg_prof_sample and lg_sample different?

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
```diff
@@ -0,0 +1,199 @@
...
+            if (!allowed) {
+                logger.warn("Heap dump path [{}] (resolved: [{}]) rejected: must be under one of {}",
+                    path, realFilePath, allowedDumpDirs);
+                return;
+            }
+
+            // Pre-create the file with restrictive permissions (600) BEFORE the native write.
+            // This eliminates the TOCTOU window where the file could be world-readable.
+            // jemalloc's prof.dump will overwrite/truncate this file.
+            Files.deleteIfExists(realFilePath);
```
himshikhagupta: So effectively only 1 dump can exist at a time?

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
```diff
@@ -0,0 +1,199 @@
...
+            Files.createFile(realFilePath, PosixFilePermissions.asFileAttribute(OWNER_ONLY_PERMS));
+
+            String resolvedPath = realFilePath.toString();
+            try (Arena arena = Arena.ofConfined()) {
+                MemorySegment cPath = arena.allocateFrom(resolvedPath);
+                long rc = (long) DUMP.invokeExact(cPath);
+                NativeLibraryLoader.checkResult(rc);
+                logger.info("jemalloc heap profile dumped to {} (permissions: 600)", resolvedPath);
```
himshikhagupta: nit: do we need to mention permission here?

sandbox/modules/native-bridge/src/main/java/org/opensearch/nativebridge/NativeBridgeModule.java
```diff
@@ -54,6 +56,53 @@ public class NativeBridgeModule extends Plugin {
...
+     * Additional directories where heap profile dumps are permitted, beyond the node's data directory.
+     * Useful for operational tooling that mounts external volumes or uses dedicated profiler paths.
+     * Configured in opensearch.yml; not dynamically updatable.
+     */
+    public static final Setting<List<String>> JEMALLOC_HEAP_PROF_ALLOWED_PATHS = Setting.listSetting(
+        "native.jemalloc.heap_prof_allowed_paths",
+        Collections.emptyList(),
+        s -> s,
+        Setting.Property.NodeScope
+    );
```
himshikhagupta: If its in only yml, then any tooling would need to restart the node. Do we really need this?

5u5hma: No you dont need to restart node right it will get packaged through internal config and come in the file in the first place. Its like how all other settings are being handled today and you will need this if you want the the internal tool to attach ebs -> save the dump in that mount -> upload to s3. This is how the current tooling works so I will have to have allowed data paths for the dump creation by jemalloc. 

The use case: operational tooling that attaches an external EBS volume at a custom mount point (e.g., /abc) and wants to dump there instead of the data dir. These paths are configured once in config file at deployment time — they don't change during the node's lifetime.

If we made this Dynamic, an attacker with cluster settings API access could add arbitrary paths at runtime, bypassing the path sandboxing security. NodeScope ensures only the node operator (who controls opensearch.yml) can configure allowed paths.

For the common case (open-source user, no special tooling), this setting is never needed — dumps go to the data dir which is always permitted. The setting exists purely for managed deployments with custom volume mounts.

We could remove it entirely and hardcode only the data dir — but that would make the feature unusable for any deployment that needs dumps on a separate volume (to avoid filling the data disk). The NodeScope setting is the minimal-privilege way to support both use cases.

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
5u5hma: This is to ensure no one changed any permissions to from what was expected, in the logs if we say it should be 600 but during the troubleshooting you found it to be reduced or increased it will help catch the issues. But can remove as well if you think not needed.

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
5u5hma: prof.reset is a jemalloc mallctl that does two things:

Discards all accumulated profiling data (clears the internal backtrace hash table)
Sets a new sample interval for future allocations (the lg_sample argument becomes the new effective lg_prof_sample)
The parameter name lg_sample here is the same as lg_prof_sample — it's the log2 of the sample interval in bytes (e.g., 17 = sample every ~128KB). I used the shorter name to match jemalloc's mallctl documentation which calls the argument to prof.reset simply "lg_sample".

From [jemalloc docs](https://jemalloc.net/jemalloc.3.html#opt.lg_prof_sample):

prof.reset (size_t) rw: Reset all memory profile statistics, and optionally update the sample rate.

This is the only way to change the effective sample rate at runtime without restarting the process. The tradeoff is that it discards accumulated data — which is why the tool command dumps before resetting.

I can rename the parameter to lg_prof_sample for clarity if preferred:

```
pub extern "C" fn native_jemalloc_heap_prof_reset(lg_prof_sample: usize) -> i64 {
    ffm_wrap("native_jemalloc_heap_prof_reset", || {
        unsafe { tikv_jemalloc_ctl::raw::write(b"prof.reset\0", lg_prof_sample) }
```

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
5u5hma: No — multiple dumps can exist simultaneously. The deleteIfExists only applies to the specific path being written to in that call. Each dump goes to a different path (e.g., /data/before.heap, /data/after.heap), so they don't overwrite each other.

The deleteIfExists is only needed to handle the case where the same path is reused (e.g., operator triggers a dump to the same file twice). Without it, Files.createFile() would throw FileAlreadyExistsException. The sequence is:

Delete the file at this specific path if it already exists
Re-create it with 600 permissions
jemalloc writes the dump content into it
A typical workflow produces multiple files:

```
PUT {"native.jemalloc.heap_prof_dump_path": "/data/before.heap"}  → creates /data/before.heap
PUT {"native.jemalloc.heap_prof_dump_path": "/data/after.heap"}   → creates /data/after.heap
```

Both files coexist. Only if you dump to the same path again does the old file get replaced.

himshikhagupta (approved):

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
```diff
@@ -111,6 +111,67 @@ fn set_all_arenas(suffix: &[u8], ms: i64) -> Result<i64, String> {
...
+    })
+}
+
+/// FFI: Dumps a heap profile to the given path. Path must be a null-terminated C string.
+/// Returns 0 on success, negative error pointer on failure.
+/// Called from Java when the cluster setting `native.jemalloc.heap_prof_dump_path` is updated.
+#[no_mangle]
+pub unsafe extern "C" fn native_jemalloc_heap_prof_dump(path: *const std::ffi::c_char) -> i64 {
+    ffm_wrap("native_jemalloc_heap_prof_dump", || {
+        if path.is_null() {
```
alchemist51, line 149: We should see if we can use: `str_from_raw` which we have contructed for converting string

alchemist51: Also on passing a path, I'm not sure if we need this. We should have it fixed and should not provide a way to configure I think in realtime.

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
```diff
@@ -0,0 +1,199 @@
...
+ * The cluster setting requires operator-level privileges to modify.
+ */
+public final class NativeHeapProfiler {
+
+    private static final Logger logger = LogManager.getLogger(NativeHeapProfiler.class);
+
+    /** Allowed dump directories — empty until explicitly set by createComponents(). */
+    private static volatile List<String> allowedDumpDirs = List.of();
+
+    private static final Set<PosixFilePermission> OWNER_ONLY_PERMS = Set.of(
```
alchemist51: I think we can simplify the logic and try to dump it in log folder always?

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
5u5hma: The str_from_raw helper is for the (ptr, len). Our function uses null-terminated C strings which is the correct approach for passing a file path to jemalloc's prof.dump mallctl (which expects a const char*). Using str_from_raw would add unnecessary conversion since jemalloc needs the C string directly.

We should keep path configurable to be able to use in the tooling and on demand heap dump cases.

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeHeapProfiler.java
5u5hma: log folders can get filled up and we will not be able to save in the additional volume usecases like ebs etc if we want.

5u5hma:
> ## PR Code Analyzer ❗
> **AI-powered '_Code-Diff-Analyzer_' found issues on commit [f98f0bd](https://github.com/opensearch-project/OpenSearch/commit/f98f0bdf0d90c5e7775dfbb412ba5417f8baaaf0).**
> 
> Path	Line	Severity	Description
> sandbox/libs/dataformat-native/rust/Cargo.toml	55	high	Dependency changes: tikv-jemallocator and tikv-jemalloc-ctl versions pinned to =0.6.1 and 'profiling' feature added to tikv-jemallocator. Mandatory flag — maintainers must verify artifact authenticity and that pinned versions are the intended upstream releases.

tikv-jemallocator and tikv-jemalloc-ctl are well-established crates maintained by TiKV (PingCAP), with 50M+ downloads on crates.io. Version 0.6.1 is the latest stable release.
Source: https://github.com/tikv/jemallocator
crates.io: https://crates.io/crates/tikv-jemallocator/0.6.1
The profiling feature simply passes --enable-prof to jemalloc's configure script during build — it does not introduce any external code paths or data collection. It enables jemalloc's built-in profiling infrastructure (stack sampling + heap dump generation).
Pinning to =0.6.1 matches the version already used on main branch for tikv-jemalloc-ctl.

> distribution/tools/heap-prof-cli/src/main/java/org/opensearch/tools/cli/jmx/JmxCommand.java	60	medium	JMX MBean operations (activate, deactivate, dump, reset) are accessible to any local OS user who can run the opensearch-heap-prof CLI. No authentication or authorization check is performed before attaching to the OpenSearch JVM and invoking profiling operations, enabling any local user to trigger heap dumps of the running server process.

This is by design — the same security model as jcmd and jmap:
* JVM Attach API requires same OS user — only the user running the OpenSearch process can attach. A different local user gets AttachNotSupportedException or permission denied.
* This is enforced by the JVM itself (checks /proc/<pid>/status UID match on Linux).
* No additional authentication is needed because OS-level user isolation is the security boundary — same as jcmd <pid> GC.heap_dump, jmap -dump:format=b, or kill -SIGQUIT <pid>.
* Adding JMX authentication would require configuring JMX passwords/SSL, which is unnecessary for local-only attach (no network port is opened).

If a local user has same-user access to the OpenSearch process, they already have equivalent capabilities via gdb -p, jcmd, or direct file access to the data directory.

> sandbox/modules/native-bridge/src/main/java/org/opensearch/nativebridge/NativeBridgeModule.java	84	low	The allowed heap dump directories are extended by the system property 'native.heap_prof.allowed_paths' (comma-separated). If an attacker can influence JVM startup options (e.g., via jvm.options or environment manipulation), they could direct heap dumps to arbitrary filesystem locations.

Intentional — the system property is set in jvm.options which is a root-owned configuration file on the node (permissions 640, owned by root:opensearch). Any party that can modify jvm.options already has root access to the node and can write files anywhere directly. The system property doesn't grant any privilege escalation beyond what the config file editor already has.

This is the same trust model as path.data, path.logs, and all other path-based settings in OpenSearch — they're all configured via files that require root/admin access to modify.

> distribution/tools/heap-prof-cli/src/main/java/org/opensearch/tools/cli/jmx/JmxCommand.java	72	low	vm.loadAgent() dynamically loads management-agent.jar whose path is derived from the target JVM's java.home system property. If java.home is tampered with in the target process, this could cause an attacker-controlled JAR to be loaded into the running OpenSearch JVM.

This is the standard jcmd/jmap pattern used across the Java ecosystem. The java.home system property is set by the JVM itself at startup from the actual JDK installation path — it cannot be modified at runtime (it's a read-only system property).

For java.home to be tampered with, an attacker would need to either:

* Replace the JDK installation on disk (requires root access)
* Start the JVM with a spoofed -Djava.home (requires controlling the process launch)

In both cases, the attacker already has full control of the process and doesn't need this tool. This is the same risk accepted by jcmd, jmap, jstack, VisualVM, and every other JMX-based tool in the JDK ecosystem.

> Reply to comment 3 (MBean with no additional access control):

This is the same security model as all existing JMX MBeans in OpenSearch (GarbageCollectorMXBean, MemoryMXBean, HotSpotDiagnosticMXBean) and the JDK's built-in com.sun.management:type=DiagnosticCommand which powers jcmd. None of these have cluster-level authorization — they all rely on OS-level user isolation via the JVM Attach API.

Adding cluster-level authorization to a JMX MBean would defeat the purpose — the tool needs to work on nodes that can't join the cluster (the primary use case). JMX local attach is intentionally outside the cluster security model, same as jcmd GC.heap_dump or jcmd VM.native_memory.

> Reply to comment 4 (auto-discovery + any OS user with execute access):

JVM Attach API enforces same-user access at the kernel level — ptrace_may_access() on Linux checks that the attaching process has the same UID as the target. A different OS user attempting to attach gets IOException: well-known file is not secure or permission denied, regardless of whether they can read /proc.

The --pid flag is available for environments where explicit targeting is preferred. Auto-discovery is a convenience that matches jcmd's behavior (which also lists and attaches to JVMs by display name without requiring a PID).

Bukhtawar (commented):
Thanks for revamping the PR, the JMX route looks decent. Can you add an IT?

5u5hma:
> Thanks for revamping the PR, the JMX route looks decent. Can you add an IT?

Added IT and output recorded

```
find ~/OpenSearch/sandbox/libs/dataformat-native/build/test-results -name "*.xml" | xargs grep "NativeHeapProfiler"
<testsuite name="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" tests="9" skipped="0" failures="0" errors="0" timestamp="2026-05-18T05:58:39.006Z" hostname="ip-172-31-81-173.ec2.internal" time="1.796">
  <testcase name="testDumpRejectsNullPath" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.337"/>
  <testcase name="testDumpRejectsDisallowedPath" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.005"/>
  <testcase name="testResetWithValidLgSample" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.005"/>
  <testcase name="testDumpRejectsEmptyPath" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.003"/>
  <testcase name="testDumpRejectsRelativePath" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.003"/>
  <testcase name="testActivateAndDeactivate" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.004"/>
  <testcase name="testResetRejectsInvalidLgSample" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.004"/>
  <testcase name="testDumpRejectsPathTraversal" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.004"/>
  <testcase name="testDumpCreatesNonEmptyFile" classname="org.opensearch.nativebridge.spi.NativeHeapProfilerTests" time="0.01"/>
  <system-out><![CDATA[[2026-05-18T05:58:40,442][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsNullPath] before test
[2026-05-18T05:58:40,743][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsNullPath] after test
[2026-05-18T05:58:40,765][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsDisallowedPath] before test
[2026-05-18T05:58:40,767][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsDisallowedPath] after test
[2026-05-18T05:58:40,770][INFO ][o.o.n.s.NativeHeapProfilerTests] [testResetWithValidLgSample] before test
[2026-05-18T05:58:40,771][INFO ][o.o.n.s.NativeHeapProfiler] [testResetWithValidLgSample] jemalloc heap profiling activated
[2026-05-18T05:58:40,772][INFO ][o.o.n.s.NativeHeapProfiler] [testResetWithValidLgSample] jemalloc profiling reset with lg_prof_sample=17
[2026-05-18T05:58:40,772][INFO ][o.o.n.s.NativeHeapProfiler] [testResetWithValidLgSample] jemalloc heap profiling deactivated
[2026-05-18T05:58:40,773][INFO ][o.o.n.s.NativeHeapProfilerTests] [testResetWithValidLgSample] after test
[2026-05-18T05:58:40,776][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsEmptyPath] before test
[2026-05-18T05:58:40,777][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsEmptyPath] after test
[2026-05-18T05:58:40,780][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsRelativePath] before test
[2026-05-18T05:58:40,781][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsRelativePath] after test
[2026-05-18T05:58:40,783][INFO ][o.o.n.s.NativeHeapProfilerTests] [testActivateAndDeactivate] before test
[2026-05-18T05:58:40,784][INFO ][o.o.n.s.NativeHeapProfiler] [testActivateAndDeactivate] jemalloc heap profiling activated
[2026-05-18T05:58:40,784][INFO ][o.o.n.s.NativeHeapProfiler] [testActivateAndDeactivate] jemalloc heap profiling deactivated
[2026-05-18T05:58:40,784][INFO ][o.o.n.s.NativeHeapProfilerTests] [testActivateAndDeactivate] after test
[2026-05-18T05:58:40,787][INFO ][o.o.n.s.NativeHeapProfilerTests] [testResetRejectsInvalidLgSample] before test
[2026-05-18T05:58:40,788][INFO ][o.o.n.s.NativeHeapProfilerTests] [testResetRejectsInvalidLgSample] after test
[2026-05-18T05:58:40,791][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsPathTraversal] before test
[2026-05-18T05:58:40,792][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpRejectsPathTraversal] after test
[2026-05-18T05:58:40,795][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpCreatesNonEmptyFile] before test
[2026-05-18T05:58:40,796][INFO ][o.o.n.s.NativeHeapProfiler] [testDumpCreatesNonEmptyFile] jemalloc heap profiling activated
[2026-05-18T05:58:40,801][INFO ][o.o.n.s.NativeHeapProfiler] [testDumpCreatesNonEmptyFile] jemalloc heap profile dumped to /home/ec2-user/OpenSearch/sandbox/libs/dataformat-native/build/testrun/test/temp/org.opensearch.nativebridge.spi.NativeHeapProfilerTests_397CCD874904C85E-001/tempDir-002/test.heap
[2026-05-18T05:58:40,802][INFO ][o.o.n.s.NativeHeapProfiler] [testDumpCreatesNonEmptyFile] jemalloc heap profiling deactivated
[2026-05-18T05:58:40,802][INFO ][o.o.n.s.NativeHeapProfilerTests] [testDumpCreatesNonEmptyFile] after test
```

5u5hma:
Tests with failures:
 - org.opensearch.action.admin.cluster.stats.ClusterStatsIT.testClusterStatsWithMappingsAndAnalysisStatsIndexMetricsFilter
 - org.opensearch.search.profile.aggregation.AggregationProfilerIT.testTopHitsAggregationFetchProfiling {p0={"search.concurrent_segment_search.enabled":"true"}}

Gradle check which is failing for these ITs. All ITs build related to this code change has passed.

Bukhtawar:
```
REPRODUCE WITH: ./gradlew ':sandbox:modules:native-bridge:internalClusterTest' --tests 'org.opensearch.nativebridge.NativeHeapProfilerIT.testMBeanIsRegistered' -Dtests.seed=6E8CA0864280A00D -Dtests.security.manager=false -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=kn-IN -Dtests.timezone=America/Grand_Turk -Druntime.java=25
NativeHeapProfilerIT > testMBeanIsRegistered FAILED
    java.lang.ExceptionInInitializerError
        at __randomizedtesting.SeedInfo.seed([6E8CA0864280A00D:6284CA55FCE65A06]:0)
        at org.opensearch.nativebridge.spi.NativeLibraryLoader.symbolLookup(NativeLibraryLoader.java:73)
        at org.opensearch.nativebridge.spi.NativeHeapProfiler.<clinit>(NativeHeapProfiler.java:46)
        at org.opensearch.nativebridge.NativeBridgeModule.createComponents(NativeBridgeModule.java:92)
        at org.opensearch.plugins.Plugin.createComponents(Plugin.java:200)
        at org.opensearch.node.Node.<init>(Node.java:1155)
        at org.opensearch.node.MockNode.<init>(MockNode.java:103)
        at org.opensearch.node.MockNode.<init>(MockNode.java:113)
        at org.opensearch.test.InternalTestCluster.buildNode(InternalTestCluster.java:837)
        at org.opensearch.test.InternalTestCluster.reset(InternalTestCluster.java:1263)
        at org.opensearch.test.InternalTestCluster.beforeTest(InternalTestCluster.java:1154)
        at org.opensearch.test.OpenSearchTestClusterRule.lambda$beforeInternal$0(OpenSearchTestClusterRule.java:148)
        at org.opensearch.test.OpenSearchTestClusterRule.beforeInternal(OpenSearchTestClusterRule.java:161)
        at org.opensearch.test.OpenSearchTestClusterRule.before(OpenSearchTestClusterRule.java:172)
        at org.opensearch.test.OpenSearchTestClusterRule$1.evaluate(OpenSearchTestClusterRule.java:365)
```

alchemist51 (approved):

Bukhtawar (approved):
LGTM

PR #21613, opened 2026-05-11
Fix ExceptionsHelper.status() to handle fasterxml Jackson exceptions

### LINKED ISSUE #21612: [BUG] ExceptionsHelper.status() no longer maps com.fasterxml.jackson.core.JsonParseException to BAD_REQUEST
### Describe the bug

PR #21029 ("Support Jackson 3.x release line") changed `ExceptionsHelper.status()` to use `org.opensearch.tools.jackson.core.JsonParseException` in the switch statement. However, `com.fasterxml.jackson.core.JsonParseException` (the fasterxml variant) is still on the classpath and used by plugins.

Previously:
- `com.fasterxml.jackson.core.JsonParseException` → `RestStatus.BAD_REQUEST`

Now:
- `org.opensearch.tools.jackson.core.JsonParseException` → `RestStatus.BAD_REQUEST`
- `com.fasterxml.jackson.core.JsonParseException` → falls to `default` → `RestStatus.INTERNAL_SERVER_ERROR`

This is a silent behavioral regression. Any plugin that throws or encounters a fasterxml `JsonParseException` will now get a 500 instead of 400, and exception-based retry logic will treat it as retryable when it should not be.

### Impact

Discovered via anomaly-detection 3.7.0 version bump (opensearch-project/anomaly-detection#1714): a test that throws `com.fasterxml.jackson.core.JsonParseException` as a non-retryable exception now causes infinite retries → StackOverflowError because the exception is classified as `INTERNAL_SERVER_ERROR` (retryable).

This likely affects any plugin using fasterxml Jackson directly.

### Proposed fix

Add the fasterxml variant back to the switch in `ExceptionsHelper.status()`:

```java
case org.opensearch.tools.jackson.core.JsonParseException ignored -> RestStatus.BAD_REQUEST;
case com.fasterxml.jackson.core.JsonParseException ignored -> RestStatus.BAD_REQUEST;
```

Same for `InputCoercionException` if applicable.

### Related
- Introduced by: #21029
- Downstream impact: opensearch-project/anomaly-detection#1714

cc @reta

dbwiddis:
### Description

Fixes #21612.

PR #21029 introduced Jackson 3.x support and changed `ExceptionsHelper.status()` to match `org.opensearch.tools.jackson.core.JsonParseException`. However, `com.fasterxml.jackson.core.JsonParseException` (Jackson 2) is still on the classpath and actively used by plugins. It now falls through to the default case (`INTERNAL_SERVER_ERROR`) instead of `BAD_REQUEST`.

### Impact

- Plugins throwing fasterxml `JsonParseException` get 500 instead of 400
- Exception-based retry logic treats parse errors as retryable (causes infinite retries in anomaly-detection: opensearch-project/anomaly-detection#1714)
- Affects any plugin using fasterxml Jackson directly

### Fix

Since `libs/core` cannot depend on `com.fasterxml.jackson.core` directly, this fix matches fasterxml exceptions by class name in the default case of both `status()` and `summaryMessage()`.

### Testing

- Added tests for `com.fasterxml.jackson.core.JsonParseException` in `ExceptionsHelperTests`
- `./gradlew :server:test --tests org.opensearch.ExceptionsHelperTests` passes

cc @reta

### DIFF
--- a/libs/core/src/main/java/org/opensearch/ExceptionsHelper.java
+++ b/libs/core/src/main/java/org/opensearch/ExceptionsHelper.java
@@ -132,7 +132,17 @@ public final class ExceptionsHelper {
             case JsonParseException ignored -> RestStatus.BAD_REQUEST;
             case NotXContentException ignored -> RestStatus.BAD_REQUEST;
             case OpenSearchRejectedExecutionException ignored -> RestStatus.TOO_MANY_REQUESTS;
-            case null, default -> RestStatus.INTERNAL_SERVER_ERROR;
+            case null -> RestStatus.INTERNAL_SERVER_ERROR;
+            default -> {
+                // Handle fasterxml Jackson exceptions by class name since libs/core
+                // does not depend on com.fasterxml.jackson.core directly
+                String className = t.getClass().getName();
+                if (className.equals("com.fasterxml.jackson.core.JsonParseException")
+                    || className.equals("com.fasterxml.jackson.core.exc.InputCoercionException")) {
+                    yield RestStatus.BAD_REQUEST;
+                }
+                yield RestStatus.INTERNAL_SERVER_ERROR;
+            }
         };
     }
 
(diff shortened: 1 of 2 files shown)


reta (changes requested):
@dbwiddis the core only relies on Jackson 3.x, all plugins have been migrated (or pull requests are waiting for review), the Jackson 2.x should not be observed (see please https://github.com/opensearch-project/anomaly-detection/pull/1705 for the issue in question).

PR #21584, opened 2026-05-10
[Analytics Backend / DataFusion] Wire SPAN scalar through analytics-engine route

ahkcs:
## Summary

Wires the PPL `span(...)` scalar through the analytics-engine route to DataFusion. Repurposed per @sandeshkr419's [review](https://github.com/opensearch-project/OpenSearch/pull/21584#pullrequestreview-) — the AVG / STDDEV / VAR custom-Sig collection, percentile_approx UDAF onboarding, and defensive try/catch wrappers have all been dropped. AVG / STDDEV_POP / STDDEV_SAMP / VAR_POP / VAR_SAMP now flow through Calcite's `OpenSearchAggregateReduceRule` decomposition (landed in [#21645](https://github.com/opensearch-project/OpenSearch/pull/21645)) — no hand-holding needed.

The PR now contains two commits:

1. **`[Analytics Backend / DataFusion] Wire SPAN scalar for stats`** — adds the bucketing primitive used by `stats … by span(...)`. Numeric span lowers to `FLOOR(field/interval)*interval`; time span with `interval=1` lowers to `DATE_TRUNC(<unit>, field)`. Multi-unit time spans (`12h`, …) fall through unchanged — separate follow-up (DataFusion `date_bin`).
2. **`[QA] Add self-contained StatsCommandIT mirror`** — 16 self-contained tests under `sandbox/qa/analytics-engine-rest` exercising the `stats` operators reachable from this PR (SPAN, plus the broader stats family that already flows through the analytics route via #21645's decomposition).

The previous `[Rebase tax] Add missing FieldType import` commit is dropped — [#21663](https://github.com/opensearch-project/OpenSearch/pull/21663) fixed it upstream.

## What changed since the last review

| @sandeshkr419 comment | Disposition |
|---|---|
| `DataFusionFragmentConvertor.java:479` — *"We are decomposing the methods via Calcite decomposition. I think all these cases/functions here are covered."* | **Done.** `collectCustomAggregateSigs` + `substraitNameForCustomAgg` removed. |
| `SubstraitPlanRewriter.java:94` — *"These changes should go away as well as they seem related to support the above decomposed methods."* | **Done.** The whole `percentile_approx` UDAF onboarding commit was dropped (including the `SubstraitPlanRewriter` Aggregate visitor literal-lift). |
[...]

PR #21621, opened 2026-05-12
[analytics-engine]Add PPL bucketing scalars (span_bucket / width_bucket / minspan_bucket / range_bucket) to analytics-engine route

sandeshkr419:
Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>
  Signed-off-by: Marc Handalian <marc.handalian@gmail.com>
  Signed-off-by: Sandesh Kumar <sandeshkr419@gmail.com>

   ## Summary

  Wires PPL's four bucketing label functions end-to-end on the analytics-engine route, exercised by the `bin <field> {span,bins,minspan,start..end}=N` command surfaces. Each returns a VARCHAR bucket label (e.g. `"0-100"`) via the OpenSearch nice-number algorithm — these are the PPL-specific variants, not ISO-SQL `width_bucket` (which returns an integer index).

  User-facing surface delivered:
  - `bin <f> span=N`           → `span_bucket(f, N)`
  - `bin <f> bins=N`           → `width_bucket(f, N, MAX(f) OVER (), …)`
  - `bin <f> minspan=N`        → `minspan_bucket(f, N, MAX(f) OVER (), …)`
  - `bin <f> start=X end=Y`    → `range_bucket(f, MIN(f) OVER (), MAX(f) OVER (), X, Y)`

  The `span()` scalar (used by `stats … by span(…)`) was wired separately in #21584 and is reused here.

  ## Implementation

  - **4 Rust UDFs** under `analytics-backend-datafusion/rust/src/udf/` ({span,width,minspan,range}_bucket.rs) ported from `sql/core`'s Java  reference implementations. Per-UDF unit tests cover algorithm correctness, null propagation, arity, and type coercion.
  - **4 Java adapters** (rename adapters extending `AbstractNameMappingAdapter`)  bind PPL's `SqlUserDefinedFunction` operators to locally-declared  SqlFunctions whose `FunctionMappings.Sig` resolves to the Rust UDFs.
  - **`ScalarFunction` enum entries** + `STANDARD_PROJECT_OPS` capability  registration + `scalarFunctionAdapters()` map wiring.
  - **YAML signatures** in `opensearch_scalar_functions.yaml` with distinct `any1` … `anyN` per slot so substrait can bind mixed-type calls.
  - **`OpenSearchProject.liftNestedRexOver`**: pre-substrait pass that  hoists nested `RexOver` out of project expressions into a child`LogicalProject`. Required because `bin bins=N` / `minspan=N` /  `start=… end=…` lower to scalar calls whose *operands* embed`MAX/MIN OVER ()`. DataFusion's substrait consumer auto-lifts  *top-level* WindowFunction project expressions into a `LogicalWindow` but doesn't touch nested RexOvers — without this lift the data-node  fragment errors with "Physical plan does not support logicalexpression WindowFunction(…)". Builds on the window support added by #21668.

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchProjectRule.java
```diff
@@ -219,6 +241,76 @@ private RexNode annotateExpr(RexNode expr, List<String> childViableBackends) {
...
+     * OVER ()}) are recognized — the only shape PPL emits via its {@code bin}
+     * command lowerings. Non-empty {@code PARTITION BY} clauses or non-default
+     * frames are rejected by returning {@code List.of()}.
+     *
+     * <p>Lookup is by {@code (AggregateFunction, operand-FieldType)} via
+     * {@link CapabilityRegistry#emptyPartitionWindowBackends}. Capability is
+     * declared per {@link org.opensearch.analytics.spi.ProjectCapability.Window}
+     * registered by each backend.
+     */
+    private List<String> resolveWindowViableBackends(RexOver rexOver, List<String> childViableBackends) {
```
expani: Using the ProjectToWindow rule of calcite, I understand the problem is coming in conversion from Substrait to DF Logical Plan. Can we explore if a simple extension of DF Substrait Consumer will help us going around the problem.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -245,7 +245,49 @@ public enum ScalarFunction {
...
      * custom Rust UDF ({@code udf::mvappend}) registered on its session context.
      */
-    MVAPPEND(Category.SCALAR, SqlKind.OTHER_FUNCTION);
+    MVAPPEND(Category.SCALAR, SqlKind.OTHER_FUNCTION),
+
+    // ── Bucketing (PPL-specific; return VARCHAR bucket labels) ──────
+    /** PPL span_bucket(value, span). Resolved to the span_bucket Rust UDF.
+     *  NOT ISO SQL width_bucket — returns a VARCHAR bucket label like "10-20",
+     *  not an integer bucket index. */
+    SPAN_BUCKET(Category.SCALAR, SqlKind.OTHER_FUNCTION),
```
expani, line 289: Window function are not ScalarFunction. 
We might need to categorize them separately as WindowFunction. 
Even if they are coming inside a Project expression.

sandeshkr419: Hi, these four are per-row scalars (RexCall+ScalarUDF+scalar substrait sigs, no frame/accumulator), not window functions and the window aggregates appear as their operands via PPL bin, and WindowFunction already models those correctly. LEt us revisit this to get a better SPI structure?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchProjectRule.java
sandeshkr419: Not exactly, liftNestedRexOver in OpenSearchProject solves it Java-side.

mch2 (approved):

PR #21625, opened 2026-05-12
[analytics-engine] Fold scalar PPL earliest()/latest() predicates at plan time

songkant-aws:
### Description

Adds an analytics-engine plan-time fold for the scalar PPL boolean predicates
`where earliest(expr, ts)` and `where latest(expr, ts)` so they execute on the
analytics-engine route without needing per-backend `EARLIEST`/`LATEST` ScalarFunction
registrations.

The first argument is required to be a string literal — either the relative-time DSL
(`-7d`, `+1mon`, `-1h@d`, …) or an absolute timestamp (`2023-01-05 12:00:00`). The
fold parses the literal once at plan time, computes the instant relative to query-start,
and rewrites the call into a plain `ts >= LITERAL` (earliest) / `ts <= LITERAL` (latest)
comparison. After the fold, the standard `OpenSearchFilterRule` carries the comparison
through DataFusion.

Pipeline:
```
PPL parser → Calcite RexCall (operator name "EARLIEST" / "LATEST")
  → PlannerImpl Phase 0: EarliestLatestAdapter.foldRelativeTimePredicates
  → standard OpenSearchFilterRule + DataFusion comparison execution
```

Semantics match `EarliestFunction` / `LatestFunction` in the SQL plugin
(at-or-after / at-or-before the folded instant).

### Files

- `EarliestLatestAdapter.java` — the relative-time DSL parser + `RelShuttle`/`RexShuttle`
  pair that rewrites `Filter` and `Project` nodes.
- `PlannerImpl.java` — invokes the fold as Phase 0, before the existing
  pre-marking / aggregate-reduction / marking phases.
- `EarliestLatestAdapterTests.java` — unit coverage for the DSL parser and
  the rewrite (296 lines, hits all relative-time units, snap-to-day, absolute-
  literal short-circuit, and the no-op cases).
- `EarliestLatestCommandIT.java` (qa REST) — end-to-end coverage on the
  analytics-engine route, mirroring the `ReverseCommandIT` pattern from #21548.

### Related Issues

N/A — sandbox feature work.

### Check List

- [x] Functionality includes testing (unit + qa REST IT).
- [ ] API changes companion pull request created — N/A (no public API surface).
- [x] Public documentation issue/PR created — N/A (sandbox).


mch2:
Thanks @songkant-aws if we need to make calcite level updates to the RexNodes, like in EarliestLatestAdapter, we can do so with a [BackendPlanAdapter](https://github.com/opensearch-project/OpenSearch/blob/446a1c9013f2067ebda2cc96aa8f8c45b23b251a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java#L50) implementation

songkant-aws:
@mch2 Thanks for pointing that out. I wired them via BackendPlanAdapter now with optimized logic.

mch2 (approved):

mch2:
sorry @songkant-aws, can u pls rebase

PR #21628, opened 2026-05-12
Wire PPL conversion scalar functions with analytics-backend-datafusion

tanik98:
### Functions added

| PPL Function | Return | Strategy |
|---|---|---|
| `num(x)` | DOUBLE | Unary adapter → Rust UDF. Strict numeric parse with comma-strip + leading-number-plus-unit-suffix fallback. |
| `auto(x)` | DOUBLE | Unary adapter → Rust UDF. `memk` first, falls back to `num`. |
| `memk(x)` | DOUBLE | Unary adapter → Rust UDF. Memory-size parser (`k` / `m` / `g` → KB). |
| `rmcomma(x)` | DOUBLE | Unary adapter → Rust UDF. Strip commas, parse; NULL if the input contains any letter. |
| `rmunit(x)` | DOUBLE | Unary adapter → Rust UDF. Extract and parse the leading numeric token. |
| `dur2sec(x)` | DOUBLE | Unary adapter → Rust UDF. Parse `[D+]HH:MM:SS` duration into total seconds. |
| `mstime(x)` | DOUBLE | Unary adapter → Rust UDF. Parse `[MM:]SS.SSS` time-of-day into total seconds (millisecond precision). |
| `ctime(value[, format])` | VARCHAR | Binary adapter → Rust UDF. Render UNIX epoch seconds as a formatted time string. |
| `mktime(value[, format])` | DOUBLE | Binary adapter → Rust UDF. Parse a formatted time string back to UNIX epoch seconds. |
| `none(x)` | Same as x | No-op function that preserves the original field value. |

### Changes

Two Calcite adapters cover all nine functions 

1. **`ConversionFunctionAdapter` — unary DOUBLE** (`num` / `auto` / `memk` / `rmcomma` / `rmunit` / `dur2sec` / `mstime`). Parameterised by the target `SqlFunction` constant. Coerces the value operand to VARCHAR (numeric columns become `CAST(x AS VARCHAR)` — DataFusion renders integers/doubles in their canonical string form, which the UDF's parse path accepts). Rebuilds the call with the synthetic target operator; the matching `Sig` entry in `DataFusionFragmentConvertor.ADDITIONAL_SCALAR_SIGS` pairs it with the YAML extension name so isthmus serialises it under the name the Rust UDF is registered as (`rust/src/udf/conversion.rs`).

2. **`TimeConversionFunctionAdapter` — binary (value, format)** (`ctime` / `mktime`). Value operand coerced to VARCHAR same as above. Format operand filled in with PPL's documented default `"%m/%d/%Y %H:%M:%S"` when the PPL call arrives as unary, so the Rust UDF always sees a stable 2-arg signature. Return type is per-target (VARCHAR for `ctime`, DOUBLE for `mktime`).

3. **Two Rust UDF modules**:
   - `rust/src/udf/conversion.rs` — shared `ConversionUdf` keyed by an internal `ConversionFn` enum. Seven variants, one dispatch table, shared scalar/array paths.
   - `rust/src/udf/time_conversion.rs` — separate `TimeConversionUdf` for `ctime` / `mktime` because the return type varies per variant. Uses `chrono::Utc.timestamp_opt(whole, nanos)` for `ctime` so fractional epoch seconds survive round-trip. `mktime` uses a single `NaiveDateTime::parse_from_str` call.

### Testing

- **Rust unit tests** — `udf/conversion.rs` and `udf/time_conversion.rs` each has unit tests for all functions with various conversion formats.

- **Java adapter unit tests** — `ConversionFunctionAdapterTests` covers per-function rewrite, VARCHAR coercion on numeric input, return-type preservation, arity-mismatch pass-through. 

- **`ConversionFunctionsIT`** — End to end coverage with each conversion functions and each possible formats.

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConversionFunctionAdapter.java
```diff
@@ -0,0 +1,107 @@
...
+ * {@code num} / {@code auto} / {@code memk} / {@code rmcomma} / {@code rmunit}
+ *
+ * <p>All five share the same Calcite shape — unary call whose single operand is of any type,
+ * returning a nullable DOUBLE. The adapter normalizes the operand to VARCHAR, then rebuilds
+ * the call with a per-function synthetic {@link SqlFunction}.
+ * Each synthetic operator is paired with its Substrait extension name.
+ *
+ * @opensearch.internal
+ */
+class ConversionFunctionAdapter implements ScalarFunctionAdapter {
```
expani: If this is not datafusion specific, then it should  be in `analytics-engine`

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
```diff
@@ -89,6 +89,15 @@ public enum ScalarFunction {
     NUMBER_TO_STRING(Category.STRING, SqlKind.OTHER_FUNCTION), // Alias for TOSTRING
     TONUMBER(Category.STRING, SqlKind.OTHER_FUNCTION),
     STRCMP(Category.STRING, SqlKind.OTHER_FUNCTION),
+    NUM(Category.SCALAR, SqlKind.OTHER_FUNCTION),
+    AUTO(Category.SCALAR, SqlKind.OTHER_FUNCTION),
+    MEMK(Category.SCALAR, SqlKind.OTHER_FUNCTION),
+    RMCOMMA(Category.SCALAR, SqlKind.OTHER_FUNCTION),
```
expani, line 113: We need to think in the long term around letting Frontend/Lang specific functions getting supplied by them at startup. 

Can you add a TODO for this as it's out of scope of this PR.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/conversion/numeric_conversion.rs
```diff
@@ -0,0 +1,534 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+//! conversion functions: `num`, `auto`, `memk`, `rmcomma`, `rmunit`, `dur2sec`, `mstime`
```
expani, line 9: The file name can be better and I think we should classify the package better. 

Ideally these could have come as a Rust crate from PPL plugin that we consumed in the runtime.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConversionFunctionAdapter.java
expani: Can we name this Regex/Text or use a better classification like TimeConversion ? 

Also, the package needs to be nested `org.opensearch.be.datafusion.planner.adapter`  We can move the rest of classes later.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/conversion/numeric_conversion.rs
tanik98: Renamed this to numeric_conversion, and moved all conversion related UDFs like numeric_conversion, time_conversion, tostring, tonumber to conversion directory.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConversionFunctionAdapter.java
tanik98: Updated and moved under org.opensearch.be.datafusion.planner.adapter.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
tanik98: Make sense, added TODO for now.

expani (approved):

mch2 (approved):

PR #21630, opened 2026-05-12
[analytics-engine] Fix TIME-operand date_part and enable 6 datetime scalars

mengweieric:
## Summary

Datetime-function fixes on the analytics-engine (DataFusion) route. All changes are engine-side (`analytics-framework` / `analytics-backend-datafusion` / `analytics-engine`); no SQL-plugin PPL behavior is touched.

- **VARCHAR→TIMESTAMP coercion** for datetime scalars (`DATE_PART`, `TO_TIMESTAMP`, `CONVERT_TZ`, `TIMESTAMPADD`, `TIMESTAMPDIFF`) so PPL literal forms like `YEAR('2003-12-31')` resolve through the Substrait signature instead of failing with `Unable to convert call <fn>(string, ...)`. Applied at two sites: `DatetimeOperandCoerceShuttle` (pre-planning) and `BackendPlanAdapter` (post-adapter, for internal Substrait-bound names).
- **6 additional scalar functions** enabled: `UTC_DATE`, `UTC_TIME`, `UTC_TIMESTAMP`, `DAYNAME`, `MONTHNAME`, `MINUTE_OF_DAY`.
- **TIME-operand synthesis** for `hour`/`minute`/`second`/`minute_of_day`/`day_of_week`: bypass the `CAST(time AS VARCHAR) AS TIMESTAMP` chain (which the DataFusion simplifier folds to a sub-10-char string Arrow rejects) by emitting `TIMESTAMP '1970-01-01 HH:MM:SS'` directly at the adapter site, matching PPL reference semantics.
- **`CONVERT_TZ` nullability-correct identity + NULL-on-out-of-range offsets**: same-tz identity fold stays NOT-NULL-consistent for the parent `Project` rowType; out-of-range offsets (e.g. `+15:00`) pass through to the runtime UDF and return NULL, matching `DateTimeFunctionIT#testConvertTZ`.
- **Standalone `TIMESTAMPADD`/`TIMESTAMPDIFF` rewrites + literal folds**.
- **`DATE_FORMAT`** gains yaml impls for `VARCHAR<L>` format args.
- **`mysql_format`** back-solves month names for `%b`/`%M` tokens.

## Test plan

- [x] `CalciteDateTimeFunctionIT`: 7 TIME-operand tests (`testHour`, `testHour_of_day`, `testMinute`, `testMinute_of_hour`, `testMinute_of_day`, `testSecond`, `testSecond_of_minute`) FAIL → PASS; no regressions in the other 58.
- [x] Unit tests green for the touched classes (`ConvertTzAdapterTests`, `DatetimeOperandCoerceShuttleTests`, `DatePartAdapters`, `TimestampAdd/DiffAdapter`, `mysql_format`).
- [x] Local CI gates green for `analytics-framework` / `analytics-engine` / `analytics-backend-datafusion`: `spotlessJavaCheck`, `forbiddenApisMain`, `compileJava`.


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -94,6 +95,11 @@ public static RelNode runAllOptimizations(RelNode rawRelNode, PlannerContext con
         RuleProfilingListener listener = context.isProfilingEnabled() ? new RuleProfilingListener() : null;
 
         RelNode modifiedRelNode = rawRelNode;
+        // Phase 0: VARCHAR→TIMESTAMP operand coercion for datetime scalars. Must precede
+        // reduceExpressions so constant folding sees post-CAST types.
+        // Fresh shuttle per call — RelShuttleImpl maintains a stateful traversal
+        // stack and cannot be shared across concurrent planning runs.
+        modifiedRelNode = modifiedRelNode.accept(new DatetimeOperandCoerceShuttle());
```
mch2: why do we need a shuttle here?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
```diff
@@ -243,7 +244,15 @@ private static RexNode adaptRex(RexNode node, Adapters adapters, List<FieldStora
         if (function != null) {
             ScalarFunctionAdapter adapter = adapters.scalar().get(function);
             if (adapter != null) {
-                return adapter.adapt(current, fieldStorage, cluster);
+                RexNode adapted = adapter.adapt(current, fieldStorage, cluster);
+                // Adapters (e.g. DatePartAdapters, DatetimeAdapter) rewrite calls to internal
+                // Substrait-bound names like date_part / to_timestamp with VARCHAR literal
+                // operands. Coerce those literals to TIMESTAMP so the Substrait signature
+                // resolves. Post-adapter is the only site that sees these internal names.
+                if (adapted instanceof RexCall adaptedCall) {
```
vinaykpud: Could it produce a coerced version directly at emit time in adapter, instead of relying on a post-pass?

mch2 (approved):

PR #21631, opened 2026-05-12
Added scan support for indexed fieldTypes

vinaykpud:
### Description
End-to-end support for `ip` and `binary` field types on the analytics-engine read path, plus the planner-side adaptations needed to compare a `VARCHAR` literal against the underlying `VARBINARY` columns.

Related change in opensearch-project/sql#5443, which extended the SQL plugin to emit a `BINARY(varchar)` placeholder whenever a VARCHAR literal is compared to a VARBINARY column. This PR tells the analytics-engine planner and the DataFusion backend to consume that placeholder and execute against the parquet-backed columns end-to-end.

 ### Scope

  Three layers, in the order the data flows through them:

  1. **Schema / scan** — `ip` and `binary` mappings now bind to a VARBINARY Calcite column. The Rust executor's substrait consumer normalizes the parquet `BinaryView` schema back to `Binary` before plan execution (`schema_coerce::coerce_inferred_schema`), since the DataFusion substrait consumer rejects the `BinaryView` shape that the parquet reader infers. DataFusion's `DefaultPhysicalPlanner` then inserts the runtime cast between the coerced schema and the underlying parquet `BinaryView` storage.

  2. **Predicates** — `BinaryFunctionAdapter` rewrites the SQL plugin's BINARY(varchar)` placeholder into a typed VARBINARY literal whose bytes match the on-disk encoding (`InetAddressPoint`-style 16-byte IPv6-mapped for `ip`, base64-decoded for `binary`). Wired into the filter / `IN` / `BETWEEN` / `cidrmatch` paths via `STANDARD_FILTER_OPS`, `OpenSearchTypeFactory.leastRestrictive`, and the existing CIDR expansion.

  3. **Project expressions** — `BINARY` is now in `STANDARD_PROJECT_OPS`, so PPL `eval if(col=lit, …)`, `eval case(col=lit, …)`, and `stats count(eval(col=lit))` no longer get rejected at the project gate. To keep the planner's per-slot type-validity assertion happy, the adapter now wraps its literal in `makeAbstractCast(nullableVarbinary, …, false)`  — `makeLiteral(ByteString, …)` always reports `BINARY(N) NOT NULL`, which mismatches the column slot's `nullable VARBINARY` type and crashes the node inside `LogicalProject.create` when the literal lands in  projection.

 `unsigned_long`

OpenSearch's unsigned_long is a 64-bit unsigned integer. Substrait integers are signed-only and Calcite has no UNSIGNED BIGINT, so the type doesn't survive through the plan as-is.

  1. Java side (OpenSearchSchemaBuilder.mapFieldType): map unsigned_long → BIGINT. The plan now treats the column as a signed 64-bit integer.
  2. Rust side (schema_coerce::coerce_inferred_schema): rewrite the parquet-inferred UInt64 schema to Int64 before handing it to the substrait consumer. The parquet reader's SchemaAdapter reinterprets each batch from UInt64 → Int64 — this
  is a bit-pattern relabel (zero copy, no value transform).

`half_float`
OpenSearch's half_float is a 16-bit IEEE float that lands as Arrow Float16 in parquet. Calcite has no fp16 type, and Substrait emits fp32, so the column doesn't bind end-to-end

 1. Java side (OpenSearchSchemaBuilder.mapFieldType): map half_float → REAL (Calcite's 32-bit float). The planner sees the same shape as a regular float column.
  2. Rust side (schema_coerce::coerce_inferred_schema): rewrite the parquet-inferred Float16 schema to Float32 before handing it to the substrait consumer. The parquet reader's SchemaAdapter widens each batch with a real IEEE half→single conversion.

### Related Issues
https://github.com/opensearch-project/sql/pull/5443

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
@@ -156,7 +156,7 @@ pub async fn execute_indexed_query(
     let resolved_schema = listing_options
         .infer_schema(&ctx.state(), &shard_view.table_path)
         .await?;
-    let resolved_schema = crate::schema_coerce::coerce_for_substrait(resolved_schema);
+    let resolved_schema = crate::schema_coerce::coerce_inferred_schema(resolved_schema);
     let table_config = datafusion::datasource::listing::ListingTableConfig::new(shard_view.table_path.clone())
         .with_listing_options(listing_options)
         .with_schema(resolved_schema);
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
@@ -156,6 +156,7 @@ pub async fn execute_indexed_query(
     let resolved_schema = listing_options
         .infer_schema(&ctx.state(), &shard_view.table_path)
         .await?;
+    let resolved_schema = crate::schema_coerce::coerce_for_substrait(resolved_schema);
     let table_config = datafusion::datasource::listing::ListingTableConfig::new(shard_view.table_path.clone())
         .with_listing_options(listing_options)
         .with_schema(resolved_schema);
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchFilter.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchFilter.java
@@ -19,6 +19,7 @@ import org.apache.calcite.rel.logical.LogicalFilter;
 import org.apache.calcite.rel.metadata.RelMetadataQuery;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexUtil;
 import org.opensearch.analytics.planner.RelNodeUtils;
 import org.opensearch.analytics.spi.FieldStorageInfo;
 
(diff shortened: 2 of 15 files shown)


sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
```diff
@@ -148,6 +148,7 @@ pub async fn execute_indexed_query(
     let resolved_schema = listing_options
         .infer_schema(&ctx.state(), &shard_view.table_path)
         .await?;
+    let resolved_schema = crate::schema_coerce::coerce_for_substrait(resolved_schema);
```
expani: I think we should explore using a custom DataFusion Substrait Consumer to do this https://github.com/datafusion-contrib/datafusion-substrait/blob/main/src/consumer.rs because such incompatibilities might arise more in the future and it will be a Good Long term investment towards generating DataFusion Physical Plan directly at the Coordinator.

vinaykpud: This is longer-term direction
1. a custom logical-plan substrait consumer that owns these type rewrites natively
2.  physical-plan-at-coordinator
We will track this separatly

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchFilter.java
```diff
@@ -102,7 +103,13 @@ public RelNode stripAnnotations(List<RelNode> strippedChildren) {
 
     @Override
     public RelNode stripAnnotations(List<RelNode> strippedChildren, Function<OperatorAnnotation, RexNode> annotationResolver) {
-        return LogicalFilter.create(strippedChildren.getFirst(), resolveCondition(getCondition(), annotationResolver));
+        RexNode resolved = resolveCondition(getCondition(), annotationResolver);
+        // LogicalFilter.<init> asserts isFlat on the condition: an AND must not contain an
+        // AND child, and an OR must not contain an OR child. Adapter substitutions in
+        // BackendPlanAdapter.adaptRex can introduce that nesting (e.g. SargAdapter expands
+        // SEARCH into AND(>=, <=) under a parent AND). Flatten canonicalizes the tree.
+        RexNode flattened = RexUtil.flatten(getCluster().getRexBuilder(), resolved);
```
mch2, line 111: why is this required?

mch2 (approved):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchFilter.java
vinaykpud: when `SargAdapter` replaces SEARCH with ANDs which causes AND or OR nesting. Without this LogicalFilter create fails (they have assertion not to have nesting)

PR #19396, opened 2025-09-24
Harden the circuit breaker and failure handle logic in query result consumer

bowenlan-amzn:
### Description

This PR is meant for harden the circuit breaker logic in query result consumer. The main logic change is only at one place, `addEstimateAndMaybeBreak(aggsSize)` in `consumeResult` before actually perform any real "consume logic".

On the other hand, considering this is a important but spaghetti code, and possibly need to be improved in the future, some refactoring works are also done.

#### Circuit Breaker Change

Circuit breaker `addEstimateBytesAndMaybeBreak` are used twice
1. before `consume`: query result received at coordinator transport layer, and will be handled by query result consumer. We estimate the heap size of the query result, and check if it breaks the REQUEST circuit breaker.
2. during `tryExecuteNext`: before doing partial reduce on the buffered query results, we estimate the extra heap size that will be used, and check if it breaks the REQUEST circuit breaker.

Some context about the partial merge logic:
- `PendingMerges` are the core logic to buffer and reduce shard results in a batched manner without waiting for all results arrived.
- `consume` buffer the shard result and check if the buffer size reaches threshold, and create merge task from buffer.
- `tryExecuteNext` poll merge task from queue and perform the partial reduce.

#### Refactoring around Failure Handling

The rule I followed is to use `onMergeFailure` to handle any failure we captured. It could be a circuit breaker exception, a task cancellation exception or any exception caught during the partial reduce.

`onMergeFailure` stores the failure, resets the circuit breaker, and clears the merge task queue. It also sends the cancel search task request to other nodes so they could try stop processing the shard query at their best.

image

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [x] Functionality includes testing.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #19340, opened 2025-09-18
Add logging and check for negative search stats

dzane17:
### Description
[Describe what this change achieves]

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #13553, opened 2024-05-06
Fix negative requestStats memory_size issue

sgup432:
### Description
- This solves the bug where RequestStats memory_size metric was going negative in certain scenarios as reported in the issue. 
#### Why so?
- It turns out that the issue occurs when an indexShard is deleted and then reallocated on the same node. So whenever stale entries from older shard are deleted, those are accounted for the new shard which has the same shardId. 
#### Why did it happen now?
We made changes recently as part of 2.12 to make IndicesRequestCache key serializable. Earlier we were storing an actual reference to indexShard in the cache. But with the new change we extract indexShard from shardId which eventually caused the above issue.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/13343

### Check List
- [x] New functionality includes testing.
  - [x] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
- [x] Failing checks are inspected and point to the corresponding known issue(s) (See: [Troubleshooting Failing Builds](../blob/main/CONTRIBUTING.md#troubleshooting-failing-builds))
- [x] Commits are signed per the DCO using --signoff
- [x] Commit changes are listed out in CHANGELOG.md file (See: [Changelog](../blob/main/CONTRIBUTING.md#changelog))
- ~[ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose)~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21635, opened 2026-05-13
Handle negative circuit breaker estimated bytes to prevent /_nodes/stats API failure

### LINKED ISSUE #21634: [BUG] Negative circuit breaker estimated_size crashes /_nodes/stats API for the entire cluster
### Describe the bug

When a circuit breaker's `estimated_size` becomes negative due to race conditions in `ChildMemoryCircuitBreaker.addWithoutBreaking()`, the `/_nodes/stats` API crashes with HTTP 400 for the **entire cluster**, even though only a single node is affected.

**Error:**
```
{
  "error": {
    "root_cause": [{
      "reason": "Values less than -1 bytes are not supported: -59452b"
    }]
  },
  "status": 400
}
```

### Root Cause

The `ChildMemoryCircuitBreaker.used` field is an `AtomicLong` that can go negative when concurrent callers invoke `addWithoutBreaking(-bytes)` to release memory. The only guard is a Java `assert` statement (disabled in production JVMs):

```java
// ChildMemoryCircuitBreaker.java:218
assert u >= 0 : "Used bytes: [" + u + "] must be >= 0";
```

When `CircuitBreakerStats.toXContent()` serializes the negative value via `new ByteSizeValue(estimated)`, the `ByteSizeValue` constructor throws `IllegalArgumentException` for values less than -1, crashing the entire `/_nodes/stats` response.

There are 20+ callers of `addWithoutBreaking()` across the codebase (RestController, InboundAggregator, AggregatorBase, BigArrays, QueryPhaseResultConsumer, etc.) that can all independently trigger this underflow.

### Impact

- **Datadog/monitoring integration completely breaks** — Datadog's elastic check calls `/_nodes/stats` and gets HTTP 400, reporting 0 metrics for the entire cluster
- **One bad node breaks cluster-wide stats** — The `/_nodes/stats` endpoint aggregates all nodes; a single node with negative breaker stats causes the entire API to fail
- **Misleading error messages** — Monitoring tools interpret the 400 as "credentials incorrect" rather than a stats serialization failure

### Affected Versions

- OpenSearch 2.19.5 (confirmed in production)
- Likely affects all versions, as the code path has existed since the Elasticsearch fork

### Steps to Reproduce

This is a race condition that occurs under production load. It can be triggered when:
1. High concurrent search/aggregation traffic causes frequent circuit breaker adjustments
2. Multiple threads call `addWithoutBreaking(-bytes)` concurrently
3. The `AtomicLong` counter underflows to a negative value
4. Any call to `/_nodes/stats` (including from monitoring agents) crashes

### Workaround

Restart the affected data node to reset circuit breaker counters

### Proposed Fix

Clamp negative `estimated` values to 0 in `CircuitBreakerStats.toXContent()` and `toString()`, with a warning log. This is a defensive fix at the serialization layer that protects against ALL callers causing underflow, following the same pattern used in:
- `RequestCacheStats` (PR #13553): `Math.max(0, totalMetric.count())`
- `SearchStats` (PR #19340): `if (current < 0) { out.writeVLong(0); }`

The raw negative value is preserved in `estimated_size_in_bytes` for debugging.

### Related Issues

- #20363 (indices stats API fails with negative values)
- #19464 (Node stats failing with illegal_argument_exception)
- #5442 (API endpoints failing with illegal_argument_exception)
- #4474 (Negative memory causes _cat APIs to fail)

### Environment

- OpenSearch version: 2.19.5
- JVM version: 21.0.10
- OS: Linux (AKS)
- Cluster size: 63 nodes (55 data, 3 master, 5 ML)

cuonghm2809:
### Description

Fixes #21634

Circuit breaker accounting can underflow due to caller-level bugs — for example, the old `QueryPhaseResultConsumer.PendingMerges.resetCircuitBreakerForCurrentRequest()` was not synchronized but was called from `checkCancellation()` on multiple transport threads concurrently. The `circuitBreakerBytes` field was volatile, but the compound read-check-release-zero sequence was not atomic: multiple threads could read the same non-zero value and each call `addWithoutBreaking(-circuitBreakerBytes)`, double-releasing and driving the breaker's `used` AtomicLong negative. The `assert u >= 0` guard in `addWithoutBreaking()` only fires in test JVMs (`-ea`); in production it is disabled.

This was fixed for `QueryPhaseResultConsumer` by #19396 (synchronized + idempotent `onFailure()`). However, there are still 19 call sites of `addWithoutBreaking()` across 8 files that could independently have similar accounting bugs.

When `CircuitBreakerStats.toXContent()` serializes a negative `estimated` value via `new ByteSizeValue(estimated)`, it throws `IllegalArgumentException` ("Values less than -1 bytes are not supported"), crashing the entire `/_nodes/stats` API for the cluster.

This fix clamps negative `estimated` values to 0 during serialization in `toXContent()` and `toString()`, with a warning log to surface the accounting bug without breaking the stats API. The raw negative value is still preserved in `estimated_size_in_bytes` for debugging.

### Changes

**`CircuitBreakerStats.java`**:
- Added `sanitizeEstimated()` method that clamps negative values to 0 with a warning log
- Updated `toXContent()` and `toString()` to use `sanitizeEstimated()` when creating `ByteSizeValue`
- Added logger for the warning

**`CircuitBreakerStatsTests.java`** (new):
- Tests for `toXContent()` with negative, valid, and -1 estimated values
- Tests for `toString()` with negative and valid values
- Tests for getters

**`HierarchyCircuitBreakerServiceTests.java`**:
- Added `testUnsynchronizedResetCausesNegativeBreaker()`: reproduces the race condition from the old `resetCircuitBreakerForCurrentRequest()` — 10 threads × 1000 iterations, uses `CyclicBarrier` to maximize contention, proves the breaker goes negative via `AssertionError: Used bytes: [-1000] must be >= 0`

### Why this approach

`addWithoutBreaking()` itself is thread-safe (`AtomicLong.addAndGet`), but callers can have accounting bugs that pass incorrect negative values. Fixing at the serialization layer in `CircuitBreakerStats` is the most defensive approach — it protects against ALL current and future callers without altering breaker semantics.

This follows the same pattern used in:
- `RequestCacheStats` (PR #13553): `Math.max(0, totalMetric.count())`
- `SearchStats` (PR #19340): `if (current < 0) { out.writeVLong(0); }`

### Issues Resolved

Fixes #21634

### Check List

### DIFF
--- a/libs/core/src/main/java/org/opensearch/core/indices/breaker/CircuitBreakerStats.java
+++ b/libs/core/src/main/java/org/opensearch/core/indices/breaker/CircuitBreakerStats.java
@@ -32,6 +32,8 @@
 
 package org.opensearch.core.indices.breaker;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.opensearch.common.annotation.PublicApi;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
@@ -50,6 +52,8 @@ import java.util.Locale;
  */
 @PublicApi(since = "1.0.0")
 public class CircuitBreakerStats implements Writeable, ToXContentObject {
+
+    private static final Logger logger = LogManager.getLogger(CircuitBreakerStats.class);
 
     /** The name of the circuit breaker */
     private final String name;
@@ -158,7 +162,7 @@ public class CircuitBreakerStats implements Writeable, ToXContentObject {
         builder.field(Fields.LIMIT, limit);
         builder.field(Fields.LIMIT_HUMAN, new ByteSizeValue(limit));
         builder.field(Fields.ESTIMATED, estimated);
-        builder.field(Fields.ESTIMATED_HUMAN, new ByteSizeValue(estimated));
+        builder.field(Fields.ESTIMATED_HUMAN, new ByteSizeValue(sanitizeEstimated()));
         builder.field(Fields.OVERHEAD, overhead);
         builder.field(Fields.TRIPPED_COUNT, trippedCount);
         builder.endObject();
@@ -180,12 +184,31 @@ public class CircuitBreakerStats implements Writeable, ToXContentObject {
             + ",estimated="
             + this.estimated
             + "/"
-            + new ByteSizeValue(this.estimated)
+            + new ByteSizeValue(sanitizeEstimated())
             + ",overhead="
             + this.overhead
             + ",tripped="
             + this.trippedCount
             + "]";
+    }
+
+    /**
+     * Returns the estimated value clamped to 0 if negative. Circuit breaker accounting can
+     * underflow due to race conditions in concurrent addWithoutBreaking() calls, producing
+     * negative values that crash ByteSizeValue serialization and break /_nodes/stats API
+     * for the entire cluster.
+     */
+    private long sanitizeEstimated() {
+        if (estimated < 0) {
+            logger.warn(
+                "Circuit breaker [{}] has negative estimated bytes [{}], clamping to 0. "
+                    + "This indicates a bug in circuit breaker accounting.",
+                name,
+                estimated
+            );
+            return 0;
+        }
+        return estimated;
     }
 
     /**
--- a/libs/core/src/main/java/org/opensearch/core/indices/breaker/CircuitBreakerStats.java
+++ b/libs/core/src/main/java/org/opensearch/core/indices/breaker/CircuitBreakerStats.java
@@ -32,8 +32,6 @@
 
 package org.opensearch.core.indices.breaker;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import org.opensearch.common.annotation.PublicApi;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
@@ -52,8 +50,6 @@ import java.util.Locale;
  */
 @PublicApi(since = "1.0.0")
 public class CircuitBreakerStats implements Writeable, ToXContentObject {
-
-    private static final Logger logger = LogManager.getLogger(CircuitBreakerStats.class);
 
     /** The name of the circuit breaker */
     private final String name;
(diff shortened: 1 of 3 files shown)


cuonghm2809:
@andrross @mch2 Could you please take a look at this PR?

andrross:
> Circuit breaker accounting can underflow due to race conditions in concurrent addWithoutBreaking() calls

What are the race conditions? Can we fix it at that level?

cuonghm2809:
> What are the race conditions? Can we fix it at that level?

Sorry for the inaccurate initial description — I've updated it with the correct root cause.

The race was in `QueryPhaseResultConsumer.PendingMerges` (fixed by #19396). `checkCancellation()` called `resetCircuitBreakerForCurrentRequest()` — neither was synchronized. `checkCancellation()` ran from `consumeResult()` on multiple transport threads concurrently. When the task got cancelled, multiple threads entered `resetCircuitBreakerForCurrentRequest()` at the same time. `circuitBreakerBytes` was volatile, but the compound read-check-release-zero was not atomic — two threads read the same non-zero value and both called `addWithoutBreaking(-circuitBreakerBytes)`, double-releasing. The `assert u >= 0` in `addWithoutBreaking()` only fires in test JVMs (`-ea`); production silently underflows.

I added a reproduction test (`testUnsynchronizedResetCausesNegativeBreaker`) — 10 threads × 1000 iterations, the assertion fires with `Used bytes: [-1000] must be >= 0`.

#19396 fixed this specific caller. But there are 19 other call sites of `addWithoutBreaking()` across 8 files. We could clamp in `ChildMemoryCircuitBreaker.addWithoutBreaking()`, but that would silently mask accounting bugs for all callers. This fix at the serialization layer keeps the stats API stable without altering breaker semantics — same approach as `RequestCacheStats` (#13553) and `SearchStats` (#19340).

andrross:
> but the compound read-check-release-zero was not atomic

@cuonghm2809 Sounds like that is the race condition. Can we fix that?

cuonghm2809:
> > but the compound read-check-release-zero was not atomic
> 
> @cuonghm2809 Sounds like that is the race condition. Can we fix that?

Sorry for the confusion, that race was already fixed by #19396. This PR is a defensive layer at the serialization side same pattern as RequestCacheStats (#13553) and SearchStats (#19340). There are 19 other addWithoutBreaking() call sites, clamping in ByteSizeValue serialization keeps /_nodes/stats stable without masking accounting bugs at the breaker level.

andrross:
> clamping in ByteSizeValue serialization keeps /_nodes/stats stable without masking accounting bugs

I guess I'm a little concerned that this will mask bugs because it will just silently succeed despite the errors. See #21667 for a PR that looks like it addresses a similar root cause. Is there a way to address the root cause here? If we don't know of any specific outstanding race conditions, I'm a bit leery of introducing a purely defensive change.

cuonghm2809:
> > clamping in ByteSizeValue serialization keeps /_nodes/stats stable without masking accounting bugs
> 
> I guess I'm a little concerned that this will mask bugs because it will just silently succeed despite the errors. See #21667 for a PR that looks like it addresses a similar root cause. Is there a way to address the root cause here? If we don't know of any specific outstanding race conditions, I'm a bit leery of introducing a purely defensive change.

You're right, I looked at #21667 and that's the better approach. We've seen this error a few times on production (estimated_size going negative on circuit breaker stats), but I haven't identified the specific caller causing it yet. Let me investigate the root cause and come back with a proper fix instead of a defensive clamp.

PR #21639, opened 2026-05-13
analytics-engine: add distributed join planning and execution

mch2:
### Description
 Adds distributed planning and execution for Join and Union in the analytics engine. Multi-input shapes now plan through Volcano's cost model — OpenSearchExchangeReducer (ER) insertion is no longer hand-rolled at HEP marking time.

  How ER insertion works now

  1. HEP marker pass rewrites LogicalJoin/LogicalUnion → OpenSearchJoin/OpenSearchUnion. No ERs are inserted here, the marker only attaches viable backends and the required output distribution (SINGLETON).
  2. Per-operator split rules (OpenSearchJoinSplitRule, OpenSearchUnionSplitRule) emit alternative plans for the optimizer:
    - COORDINATOR-local alternative: gather both sides to coord (always valid).
    - SHARD-local alternative: when inputs co-locate (same tableId, single shard), keep the operator shard-local with no gather — strictly cheaper, so Volcano picks it.
  3. OpenSearchDistribution gained a Locality dimension (SHARD vs COORDINATOR) and carries tableId/shardCount on SHARD distributions; this is what the cost gate compares against.
  4. DAGBuilder now recurses with sever() into ER child fragments, nested ERs (a Join's per-side ERs sitting below a top-level gather ER) get cut into their own child stages. Each Stage reads its ExchangeInfo directly off the ER instead of hard-coding SINGLETON.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchJoinRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchJoinRule.java
@@ -1,0 +1,135 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.analytics.planner.rules;
+
+import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.plan.RelTraitSet;
+import org.apache.calcite.plan.hep.HepRelVertex;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.JoinInfo;
+import org.apache.calcite.rel.core.JoinRelType;
+import org.apache.calcite.rel.logical.LogicalJoin;
+import org.opensearch.analytics.planner.PlannerContext;
+import org.opensearch.analytics.planner.RelNodeUtils;
+import org.opensearch.analytics.planner.rel.OpenSearchDistributionTraitDef;
+import org.opensearch.analytics.planner.rel.OpenSearchExchangeReducer;
+import org.opensearch.analytics.planner.rel.OpenSearchJoin;
+import org.opensearch.analytics.planner.rel.OpenSearchRelNode;
+import org.opensearch.analytics.spi.JoinCapability;
+
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * HEP marker rewriting {@link LogicalJoin} → {@link OpenSearchJoin}. Both inputs are
+ * gathered to the coordinator (enforced by the join's cost gate, which only accepts
+ * SINGLETON inputs — Volcano inserts an {@link OpenSearchExchangeReducer} per side).
+ *
+ * <p>Accepts INNER / LEFT / RIGHT / FULL / SEMI / ANTI equi-joins. Cross joins match
+ * via {@link JoinInfo#isEqui()}. Pure non-equi predicates are rejected.
+ *
+ * @opensearch.internal
+ */
+public class OpenSearchJoinRule extends RelOptRule {
+
+    private final PlannerContext context;
+
+    public OpenSearchJoinRule(PlannerContext context) {
+        super(operand(LogicalJoin.class, any()), "OpenSearchJoinRule");
+        this.context = context;
+    }
+
+    @Override
+    public boolean matches(RelOptRuleCall call) {
+        LogicalJoin join = call.rel(0);
+        JoinRelType joinType = join.getJoinType();
+        // Accept INNER / LEFT / RIGHT / FULL / SEMI / ANTI equi-joins. FULL is needed
+        // by PPL's `appendcol` lowering (ROW_NUMBER pairing via a full outer join on the
+        // row numbers). Pure non-equi joins are rejected below via JoinInfo.isEqui().
+        if (joinType != JoinRelType.INNER
+            && joinType != JoinRelType.LEFT
+            && joinType != JoinRelType.RIGHT
+            && joinType != JoinRelType.FULL
+            && joinType != JoinRelType.SEMI
+            && joinType != JoinRelType.ANTI) {
+            return false;
+        }
+        // Accept equi-joins and cross joins (both satisfy JoinInfo.isEqui() — empty
+        // nonEquiConditions). A pure non-equi predicate (e.g. t1.a < t2.b) yields
+        // isEqui()=false and stays rejected — DataFusion would need a non-equi
+        // NestedLoopJoin path we don't enable yet.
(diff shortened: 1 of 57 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchJoinRule.java
```diff
@@ -0,0 +1,166 @@
...
+        List<String> rightBackends = viableBackendsOf(right);
+
+        Set<String> intersection = new LinkedHashSet<>(leftBackends);
+        intersection.retainAll(rightBackends);
+        return new ArrayList<>(intersection);
+    }
+
+    private static List<String> viableBackendsOf(RelNode rel) {
+        RelNode unwrapped = rel;
+        if (unwrapped instanceof HepRelVertex vertex) {
```
expani: nit: `RelNodeUtils.unwrapHep(RelNode node)` can be used like other places.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchJoinRule.java
```diff
@@ -0,0 +1,166 @@
...
+        JoinCapability.JoinKind requiredKind = toJoinKind(join.getJoinType());
+        viableBackends.removeIf(backend -> {
+            var caps = context.getCapabilityRegistry().getBackend(backend).getCapabilityProvider();
+            for (JoinCapability cap : caps.joinCapabilities()) {
+                if (cap.kinds().contains(requiredKind)) return false;
+            }
+            return true;
+        });
+        if (viableBackends.isEmpty()) {
+            throw new IllegalStateException("No backend supports join kind [" + requiredKind + "] among viable backends");
```
expani: nit: Can we also log the viableBackends in the message like `requireKind`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchJoinRule.java
```diff
@@ -0,0 +1,166 @@
...
+            gatheredRight,
+            join.getCondition(),
+            join.getJoinType(),
+            viableBackends,
+            strategy
+        );
+        call.transformTo(osJoin);
+    }
+
+    private static JoinCapability.JoinKind toJoinKind(JoinRelType joinType) {
```
expani: nit: Can be in the JoinCapability as a static method

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchJoinRule.java
```diff
@@ -0,0 +1,166 @@
...
+        // a separate child stage per side — same pattern as OpenSearchUnionRule. If the
+        // input already delivers SINGLETON (e.g. a FINAL aggregate's output), the
+        // ConverterImpl-based ER dedupes into the same RelSet subset.
+        OpenSearchDistributionTraitDef distTraitDef = context.getDistributionTraitDef();
+        List<String> reduceViable = CapabilityResolutionUtils.filterByReduceCapability(context.getCapabilityRegistry(), viableBackends);
+        RelNode leftUnwrapped = RelNodeUtils.unwrapHep(join.getLeft());
+        RelNode rightUnwrapped = RelNodeUtils.unwrapHep(join.getRight());
+        RelTraitSet leftSingletonTraits = leftUnwrapped.getTraitSet().replace(distTraitDef.singleton());
+        RelTraitSet rightSingletonTraits = rightUnwrapped.getTraitSet().replace(distTraitDef.singleton());
+        RelNode gatheredLeft = new OpenSearchExchangeReducer(join.getCluster(), leftSingletonTraits, leftUnwrapped, reduceViable);
```
expani: This means Exchanges don't represent data movement anymore. 

In case of a Join on the same index/table with one shard, will we push the entire Plan to Datafusion ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortSplitRule.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.calcite.rel.RelDistribution;
+import org.apache.calcite.rel.RelNode;
+import org.opensearch.analytics.planner.PlannerContext;
+import org.opensearch.analytics.planner.rel.OpenSearchDistribution;
+import org.opensearch.analytics.planner.rel.OpenSearchDistributionTraitDef;
+import org.opensearch.analytics.planner.rel.OpenSearchSort;
+
+/**
+ * For a collated {@link OpenSearchSort}, requests SINGLETON input so the plan becomes
+ * {@code Sort ← ER ← scan} — gather first, then global sort. Our ExchangeReducer is a
```
expani, line 24: Does this break the single shard case as there is no exchange required for such cases ?

expani (commented):
Thanks @mch2 for getting this started. 

I think we should explore Calcite's Volcano Trait mismatch to insert exchanges instead of deterministically doing it via HEP rule.

mch2:
@expani Thanks for the review - i've swapped this to be pure CBO, no hardcoded HEP wrapping - covers single shard case and multi nicely with two traits.   Still polishing a few things and will open back up

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchJoinRule.java
mch2: removed this.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/JoinCommandIT.java
```diff
@@ -0,0 +1,356 @@
...
+     * {@code ROW_NUMBER() OVER (ORDER BY …)} for pairing rows. The
+     * analytics-framework has no window-function capability track — no
+     * {@code WindowFunction} enum, no {@code windowFunctionCapabilities()} on
+     * {@link org.opensearch.analytics.spi.BackendCapabilityProvider}, and
+     * {@link org.opensearch.analytics.spi.ScalarFunction} has no ROW_NUMBER
+     * entry either (scalar functions are per-row; window functions are per-row
+     * over an aggregation frame — modelling them as scalars breaks the
+     * capability-type contract). Wiring the full track is a follow-up.
+     */
+    @AwaitsFix(bugUrl = "Task #113: appendcol plans correctly (ROW_NUMBER supported) but hits the same AggregateSplit-under-per-side-ER issue surfacing a runtime schema coercion mismatch.")
```
mch2, line 211: window required for this test.

LantaoJin (approved):
Basically LGTM, broadcast join (M1) was done in local and I will submit in tomorrow after merging and rebasing with this PR.

sandeshkr419 (approved):

PR #21643, opened 2026-05-13
Wire PPL conditional functions with analytics-backend-datafusion

tanik98:
### Description

Wires PPL [conditional functions](https://docs.opensearch.org/latest/sql-and-ppl/ppl/functions/condition/) through the `analytics-backend-datafusion` plugin.

1. **`coalesce(...)`** — upstream registers to `PPLBuiltinOperators.ENHANCED_COALESCE`, a UDF that performs evaluation-time type coercion to the operands' `leastRestrictive` common type. Stock `SqlStdOperatorTable.COALESCE` doesn't coerce; its `SameOperandTypeChecker` requires pre-unified operand types. New `CoalesceAdapter` moves the UDF's runtime coercion to plan time by computing `leastRestrictive` and inserting CASTs, then rewrites to stock COALESCE. Parity with `EnhancedCoalesceFunction` on return type, null handling, mixed-type fallback to VARCHAR, and "first non-null wins" semantics.
2. **`isempty(x)` / `isblank(x)`** — PPL lowers both to structural shapes containing `SqlStdOperatorTable.IS_EMPTY`, which DataFusion doesn't have a native binding for. New `IsEmptyAdapter` rewrites `IS_EMPTY(x) → CHAR_LENGTH(x) = 0`, decomposing to primitives DataFusion already supports.

Project-side capability also needed to be broadened so `x IS NULL` / `x IS NOT NULL` / `ispresent(x)` work in `eval` expressions — they were already filter-capable but not project-capable.

### Functions covered

| PPL function | Return | Routing strategy |
|---|---|---|
| `case(cond1, val1, …, [else default])` | ANY | Native — PPL parses to `SqlStdOperatorTable.CASE`, isthmus serialises as Substrait `IfThen`. Already in `STANDARD_PROJECT_OPS`. |
| `if(cond, a, b)` | ANY | Native — PPL registers to `SqlStdOperatorTable.CASE` with a BOOL+ANY+ANY type checker. Same path as `case`. |
| `ifnull(a, b)` | ANY | Native — PPL registers to plain `SqlStdOperatorTable.COALESCE`. Serialises via isthmus' default catalog. Already in `STANDARD_PROJECT_OPS`. |
| `x IS NULL` / `x IS NOT NULL` | BOOLEAN | Native in filter; this PR adds them to `STANDARD_PROJECT_OPS` so they work in `eval` too. |
| `ispresent(x)` | BOOLEAN | Native — PPL aliases to `SqlStdOperatorTable.IS_NOT_NULL`. Same path as `x IS NOT NULL`. |
| `nullif(a, b)` | ANY | Native — PPL lowers to `CASE(EQUALS(a,b), NULL, a)`; backend sees CASE + EQUALS + a typed NULL literal (already handled by `UntypedNullPreprocessor`). Enum entry + capability declaration added for symmetry. |
| `coalesce(a, b, …)` | ANY | New `CoalesceAdapter` rewrites `ENHANCED_COALESCE(args…) → SqlStdOperatorTable.COALESCE(args…)`. |
| `isempty(x)` | BOOLEAN | PPL lowers to `OR(IS_NULL(x), IS_EMPTY(x))`; new `IsEmptyAdapter` rewrites the IS_EMPTY arm to `CHAR_LENGTH(x) = 0`. |
| `isblank(x)` | BOOLEAN | PPL lowers to `OR(IS_NULL(x), IS_EMPTY(TRIM(x)))`; same adapter handles the IS_EMPTY arm. |
| `regexp_match(x, pattern)` | BOOLEAN | Native — PPL aliases to `SqlLibraryOperators.REGEXP_CONTAINS`, already routed by `FunctionMappings.s(REGEXP_CONTAINS, "regex_match")` in our substrait catalog. |

### Changes

**`CoalesceAdapter`** — preserves `EnhancedCoalesceFunction` semantics by moving the UDF's evaluation-time type coercion to plan time. For each call:

1. Compute `typeFactory.leastRestrictive(operandTypes)` — the same source of truth `EnhancedCoalesceFunction.getReturnTypeInference` uses. Falls back to nullable VARCHAR when unification returns `null`, mirroring the UDF's same fallback.
2. Insert a plain CAST on any operand whose type differs from the common type. Calcite's CAST covers the same numeric / boolean / character / date / time / timestamp conversions that the UDF's `coerceToType` covers at runtime via `ExprValueUtils`.
3. Rebuild the call with `SqlStdOperatorTable.COALESCE` so isthmus' default catalog serialises to Substrait `coalesce`.

For the all-same-type happy path (every `ifnull(a, b)` via PPL's stock-COALESCE registration, plus uniform-typed user `coalesce(...)` calls) no CAST is added and the rewrite is a pure operator swap. The mixed-type path — `coalesce(str, int, "fallback")` — unifies to VARCHAR at plan time, inserts a CAST on the numeric operand, and DataFusion evaluates the call natively. End-to-end the user-visible result matches the SQL plugin's UDF: same return type, same "first non-null, coerced to common type" behavior, same VARCHAR fallback on unification failure.

**`IsEmptyAdapter`** — unary rewrite `IS_EMPTY(x) → CHAR_LENGTH(x) = 0`. CHAR_LENGTH is already in isthmus' default catalog as `length`, EQUALS is a stock binary operator. Null propagation rides through naturally: `CHAR_LENGTH(NULL) = NULL` and `NULL = 0 → NULL`, and the outer `OR(IS_NULL(x), …)` shape that PPL wraps around it already compensates.

**`ScalarFunction.IS_EMPTY`** — Calcite's `SqlStdOperatorTable.IS_EMPTY` is named `"IS EMPTY"` (with a space) and is of `SqlKind.OTHER`, so neither `fromSqlKind` nor identifier-name `valueOf` resolves it. Uses the `referenceOperator` singleton-identity hook — same pattern `CONCAT` uses for `"||"`.

**`STANDARD_PROJECT_OPS` additions** — `IS_NULL`, `IS_NOT_NULL`, `NULLIF`, `IS_EMPTY`. Native predicates were previously filter-only; conditional expressions in `eval` projections now route natively. `NULLIF` is cosmetic (PPL lowers to CASE+EQUALS before the backend sees it) but declared for capability-set completeness and to unblock any future planner pass that inspects `SqlKind.NULLIF` by enum.

**`BASELINE_SCALAR_OPS` additions** — `AND`, `OR`, `NOT` added to `OpenSearchProjectRule`'s baseline allow-list. These boolean combinators were previously only recursed into on the filter side (comment in `STANDARD_FILTER_OPS` says "`AND/OR/NOT` are recursed into by `OpenSearchFilterRule` structurally and never looked up here"). `isempty(x)` / `isblank(x)` surface `OR` in project position for the first time (they lower to `OR(IS_NULL(x), IS_EMPTY(…))`), so the project-side rule needs to treat them the same way — structural recursion into operands, no per-backend capability lookup on the combinator itself. Every backend already supports these primitives natively through isthmus' default substrait catalog.

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/CoalesceAdapter.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/CoalesceAdapter.java
@@ -1,0 +1,80 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.be.datafusion;
+
+import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.opensearch.analytics.spi.FieldStorageInfo;
+import org.opensearch.analytics.spi.ScalarFunctionAdapter;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Adapter for coalesce(field1, field2, ...),
+ * returns the first non-null, non-missing value in the parameter list.
+ *
+ * <p>Stock {@link SqlStdOperatorTable#COALESCE} requires operands that already share a common type.
+ * So naively rewriting the call to stock {@code COALESCE} with the original operands would reject
+ * mixed-type inputs like {@code coalesce(VARCHAR, INTEGER, "fallback")} at plan validation.
+ *
+ * <p>This adapter moves the coercion step to plan time by:
+ * <ol>
+ *   <li>Computing {@code leastRestrictive} across all operand types</li>
+ *   <li>Falling back to nullable VARCHAR when the unification returns {@code null}</li>
+ *   <li>Inserting a plain CAST on each operand whose type differs from the common type.
+ *       Calcite's CAST covers numeric / boolean / character / date / time / timestamp
+ *       / decimal conversions</li>
+ *   <li>Rebuilding the call with {@link SqlStdOperatorTable#COALESCE} so isthmus serialises
+ *       it through the default Substrait catalog</li>
+ * </ol>
+ *
+ * @opensearch.internal
+ */
+class CoalesceAdapter implements ScalarFunctionAdapter {
+
+    @Override
+    public RexNode adapt(RexCall original, List<FieldStorageInfo> fieldStorage, RelOptCluster cluster) {
+        RexBuilder rexBuilder = cluster.getRexBuilder();
+        RelDataTypeFactory typeFactory = cluster.getTypeFactory();
+        List<RexNode> operands = original.getOperands();
+        if (operands.isEmpty()) {
+            return original;
+        }
+
+        // leastRestrictive([T]) is T, so single-operand coalesce degenerates to that operand's
+        // type and no CAST is needed. leastRestrictive across mixed numeric types returns the
+        // widest; across string + numeric returns VARCHAR; across incompatible type families
+        // (e.g. DATE + BIGINT) returns null, which we backstop with VARCHAR
+        List<RelDataType> operandTypes = operands.stream().map(RexNode::getType).toList();
+        RelDataType commonType = typeFactory.leastRestrictive(operandTypes);
+        if (commonType == null) {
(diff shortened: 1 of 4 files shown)


songkant-aws:
I have simplified the is_empty and is_blank functions, basically syntax sugar from PPL side. The `IS_EMPTY` is designed for collections instead of strings in Calcite builtin function. So `IS_EMPTY` s a misused, non SQL standard function(not supported in most of common dialects like MySQL, PostgreSQL, Spark, etc.) for checking string emptiness. This PPL side PR(https://github.com/opensearch-project/sql/pull/5439) improves this behavior so we don't need a isEmptyAdapter anymore.

tanik98:
> I have simplified the is_empty and is_blank functions, basically syntax sugar from PPL side. The `IS_EMPTY` is designed for collections instead of strings in Calcite builtin function. So `IS_EMPTY` s a misused, non SQL standard function(not supported in most of common dialects like MySQL, PostgreSQL, Spark, etc.) for checking string emptiness. This PPL side PR([opensearch-project/sql#5439](https://github.com/opensearch-project/sql/pull/5439)) improves this behavior so we don't need a isEmptyAdapter anymore.

Makes sense, this now simplifies and optimizes the behaviour.

songkant-aws (approved):

mch2 (approved):

Issue #3040, opened 2022-04-22
Upgrade EC2 discovery plugin to use IMDSv2


Bukhtawar:
**Is your feature request related to a problem? Please describe.**
The current implementation uses a less secure IMDSv1 API 

**Describe the solution you'd like**
Upgrade to use IMDSv2 API
**Describe alternatives you've considered**
N/A

**Additional context**

Bukhtawar:
Closing in favour of https://github.com/opensearch-project/OpenSearch/issues/7916

dblock:
Let's reopen this one since it was here earlier!

shurkus:
Hello! Any news?

dblock:
> Hello! Any news?

AFAIK nobody is working on this Want to help @shurkus?

eliskovets:
I believe it's already fixed. At least ec2 discovery works for me on ec2 instances with IMDSv2 required configured.

msfroh:
I'm surprised that it works for you @eliskovets -- are you sure that the instances are not set to IMDSv2 optional?

From what I can see, the ec2-discovery plugin isn't setting a token header when fetching from IMDS. It's just opening a basic URLConnection ([here](https://github.com/opensearch-project/OpenSearch/blob/0ee69c6a7e1f293a2389ee24a7f3971fd6a52afe/plugins/discovery-ec2/src/main/java/org/opensearch/discovery/ec2/Ec2DiscoveryPlugin.java#L160) and [here](https://github.com/opensearch-project/OpenSearch/blob/0ee69c6a7e1f293a2389ee24a7f3971fd6a52afe/plugins/discovery-ec2/src/main/java/org/opensearch/discovery/ec2/Ec2NameResolver.java#L114)).

I see [this comment](https://github.com/opensearch-project/OpenSearch/blob/0ee69c6a7e1f293a2389ee24a7f3971fd6a52afe/plugins/discovery-ec2/src/main/java/org/opensearch/discovery/ec2/Ec2DiscoveryPlugin.java#L158) saying that we should use `EC2MetadataUtils#getAvailabilityZone`, which would indeed use IMDSv2. It looks like that hasn't been done yet, though.

sbourzai:
Hello, any update about the imdsv2 support ?

jawon-benchling:
Hi,
I have some interest in fixing this - it looks like this TODO
[...]

PR #21644, opened 2026-05-13
Migrate discovery-ec2 plugin availability zone fetch to use IMDSv2

jawon-benchling:
### Description

Migrate Ec2DiscoveryPlugin.java and Ec2eNameResolver.java to Ec2MetadataClient (which uses IMDSv2).
Previous implementation used IMDSv1.

Adds the AWS SDK IMDS package as a plugin dependency; license, notice, sha1 added.

Signed-off-by: Jawon Breed <jawon@benchling.com>

### Related Issues
Partial fix for #3040

### Check List
  - Not applicable
  - Not applicable


plugins/discovery-ec2/src/main/java/org/opensearch/discovery/ec2/Ec2DiscoveryPlugin.java
```diff
@@ -127,56 +120,46 @@ public List<Setting<?>> getSettings() {
...
-        }
 
         try (
-            InputStream in = AccessController.doPrivilegedChecked(urlConnection::getInputStream);
-            BufferedReader urlReader = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8))
+            Ec2MetadataClient client = AccessController.doPrivilegedChecked(
+                () -> Ec2MetadataClient.builder()
+                    .httpClient(ApacheHttpClient.builder().connectionTimeout(Duration.ofSeconds(2)).socketTimeout(Duration.ofSeconds(2)))
+                    .build()
+            )
```
andrross: Why not use the default configuration via `Ec2MetadataClient.create()` that the javadoc suggests? Why do we need the custom timeouts?

plugins/discovery-ec2/src/main/java/org/opensearch/discovery/ec2/Ec2DiscoveryPlugin.java
```diff
@@ -127,56 +120,46 @@ public List<Setting<?>> getSettings() {
...
-            urlConnection.setConnectTimeout(2000);
-        } catch (final IOException e) {
-            // should not happen, we know the url is not malformed, and openConnection does not actually hit network
-            throw new UncheckedIOException(e);
-        }
 
         try (
-            InputStream in = AccessController.doPrivilegedChecked(urlConnection::getInputStream);
-            BufferedReader urlReader = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8))
+            Ec2MetadataClient client = AccessController.doPrivilegedChecked(
```
andrross: The [javadoc](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/imds/Ec2MetadataClient.html) for this client says:

> Note: A single client instance should be reused for multiple requests when possible.

Should the plugin create one instance of this client and reuse it throughout its lifetime?

plugins/discovery-ec2/src/main/java/org/opensearch/discovery/ec2/Ec2DiscoveryPlugin.java
jawon-benchling: The existing IMDSv1 code sets connection timeout at 2 seconds (line 162), so I just copied that - happy to omit if you don't think it's necessary.

plugins/discovery-ec2/src/main/java/org/opensearch/discovery/ec2/Ec2DiscoveryPlugin.java
jawon-benchling: My understanding is that this code runs only once at startup, so I don't think this should be that expensive as it is - but it looks like we do build and re-use the Ec2Client, in [AwsEc2ServiceImpl](https://github.com/opensearch-project/OpenSearch/blob/main/plugins/discovery-ec2/src/main/java/org/opensearch/discovery/ec2/AwsEc2ServiceImpl.java) , so I can make a similar implementation for the Ec2MetadataClient. Should I do that?

jawon-benchling:
@andrross I went ahead and implemented `Ec2MetadataClient` lifecycle handling, into the existing `AwsEc2Service`. (This also removed the need for any additions to `plugin-security.policy`, so that has been removed.)

Please take a look! Thank you :)

andrross (approved):

jawon-benchling:
@andrross Thank you for the review and merge!

I have some questions about how this code gets released.

We are currently on OpenSearch 2.19, using the ECR public Docker image from [here](https://gallery.ecr.aws/opensearchproject/opensearch) and then using `bin/opensearch-plugin install discovery-ec2` in a Dockerfile to install the plugin.

My question is, will this change be included in the next `2.19.x` patch version that gets added to ECR public gallery? Or will there be an updated plugin version that `bin/opensearch-plugin` can pull? Or will this change not be included in the `2.19.x` line at all, and we would need to look into running at least `3.7`? I read through [RELEASING.md](https://github.com/opensearch-project/.github/blob/main/RELEASING.md) but I couldn't find a clear answer from it.

Thank you,
Jawon Breed

andrross:
@jawon-benchling Version 2.19 is in maintenance mode, so only critical bug fixes or security vulnerabilities get backported. A change like this is generally not a candidate for backporting.

The 3.7 release is already underway, so this change will be released in 3.8.

jawon-benchling:
@andrross Understood - thank you.

Just to be clear, there is no way to install a `3.8` `discovery-ec2` plugin into a `2.19.x` OpenSearch docker image, correct?

andrross:
@jawon-benchling You cannot install the 3.8 binaries in a 2.19.x server. However, you could create a local branch off of 2.19 and backport your changes, resolving any conflicts. From there you could create a build of the plugin and you should be able to install the resulting zip file.

andrross:
For example, the command

```
./gradlew :plugins:discovery-ec2:assemble -Dbuild.snapshot=false
```

will create the file

```
plugins/discovery-ec2/build/distributions/discovery-ec2-<VERSION>.zip
```

PR #21645, opened 2026-05-13
[analytics-engine] Collapse AggregateDecompositionResolver into BackendPlanAdapter's adaptNode chain

sandeshkr419:
## Summary

   Replaces the monolithic `AggregateDecompositionResolver` (438 lines) with a small in-tree rewriter that runs as part of `BackendPlanAdapter.adaptNode` — the same tree-walk path that handles scalar-function adaptation for `OpenSearchFilter` and  `OpenSearchProject`. Removes the pre-existing "extra post-adaptation pass" and the `DefaultPlanExecutor.resolveAll` call, with no functional change to query results.

   The prior resolver was documented as "backend-agnostic" but in practice was coupled to DataFusion's intermediate-state conventions via `AggregateFunction.intermediateFields`. The new rewriter reads the enum directly (no adapter indirection) and lives where scalar adaptation already lives — one tree walk, one aggregate-call construction site, one responsibility per adapter branch.

### Volcano split rule (unchanged intent, clearer responsibility)

   `OpenSearchAggregateSplitRule` produces `FINAL(ExchangeReducer(PARTIAL))` with identical aggCalls on both halves. No rewriting inside Volcano — the planner's `typeMatchesInferred` and row-type equivalence checks are strict about nullability, which the decomposition rewrites inevitably perturb (SUM of a NOT-NULL count column infers nullable BIGINT; original COUNT's output was NOT NULL).

### BackendPlanAdapter.adaptNode — single tree-walk, per-stage, post-Volcano

OpenSearchFilter                       → adaptFilter   (existing, scalar-function adaptation)
OpenSearchProject                    → adaptProject  (existing + row-type-shift rebinding)
OpenSearchAggregate(FINAL)  → DistributedAggregateRewriter.rewrite else → pass through adapted children

   `adaptProject` now rebinds its `RexInputRef`s + CASTs projections back to the declared column type when its child's row type shifts (happens when a FINAL  aggregate below rewrites SUM's output to nullable BIGINT). This is a general  correctness invariant — any adapter that shifts types is handled safely.

   ### DistributedAggregateRewriter

   Takes a FINAL `OpenSearchAggregate`, returns a rewritten FINAL. For each aggCall, classifies via `AggregateFunction.fromSqlAggFunction(call).intermediateFields()`:

   | Classification | Action |
   |---|---|
   | `null` (pass-through: SUM/MIN/MAX and post-reduction primitives) | rebase arg to partial output column, keep function, Calcite-infer type |
   | `size == 1, reducer != self` (function-swap: COUNT→SUM) | swap function to reducer, rebase arg, Calcite-infer type, exchange column = Arrow-derived |
   | `size == 1, reducer == self` (engine-native: APPROX_COUNT_DISTINCT sketch merge) | keep function, rebase arg, explicit return type (Calcite can't infer `approx_count_distinct(Binary)`), exchange column = Arrow-derived |
   | `size > 1` | IllegalStateException (multi-field should have been reduced by `OpenSearchAggregateReduceRule` in HEP) |

   Single `AggregateCall.create` site. Direct structural access to the `StageInputScan` below — **no recursive tree walks**.

   ### DefaultPlanExecutor

   BackendPlanAdapter.adaptAll(dag, capabilityRegistry);   // aggregate rewrite lives here now
   FragmentConversionDriver.convertAll(dag, capabilityRegistry);

   No separate `AggregateDecompositionResolver.resolveAll` pass.

### Related Issues

### Check List


mch2 (approved):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
```diff
@@ -197,4 +221,41 @@ private static RexNode adaptRex(
...
+                if (ref.getType().equals(actual)) return ref;
+                return new RexInputRef(ref.getIndex(), actual);
+            }
+        };
+        ArrayList<RexNode> rebound = new ArrayList<>(projects.size());
+        for (int i = 0; i < projects.size(); i++) {
+            RexNode expr = projects.get(i).accept(rebind);
+            RelDataType targetType = originalProject.getRowType().getFieldList().get(i).getType();
+            if (!expr.getType().equals(targetType)) {
+                expr = rexBuilder.makeCast(targetType, expr);
```
expani, line 255: This might hamper performance. 
Why do we need an explicit cast ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
```diff
@@ -197,4 +221,41 @@ private static RexNode adaptRex(
...
     }
+
+    /**
+     * Rebind a Project's expressions against a new input whose row type has shifted (typically
+     * in nullability — e.g. FINAL aggregate's rewriter turned a NOT-NULL count into a nullable
+     * BIGINT). RexInputRefs get retyped to the new column types; projections that diverge from
+     * the Project's declared column type get wrapped in a CAST so the outer-visible schema is
+     * preserved.
+     */
+    private static ArrayList<RexNode> rebindProjectsAgainstInput(
```
expani, line 232: I think we might have to move DAG building after PlanForker and BackendPlanAdaption.

expani (approved):
Thanks @sandeshkr419 for working on this. 

This helps in cleaning things up, but I think we can follow up with moving DAGBuilder after Plan forking and Backend plan adaption as a fast follow up.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
sandeshkr419: Ack, will revisit as a fast follow-up.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
sandeshkr419: Let me revisit this in a follow-up.
Thanks.

PR #21647, opened 2026-05-13
Native memory search backpressure for off-heap query cancellation

pradeep-L:
## Description
Adds a Native Memory Usage Tracker to `SearchBackpressureService` so off-heap
allocations from native backends (DataFusion's pool, direct buffers, etc.)
can drive task cancellation. Existing CPU/heap/elapsed-time trackers can't
see native memory.

Per-task threshold setting follows the same shape as `heap_percent_threshold`:
`search_backpressure.{search_task,search_shard_task}.native_memory_percent_threshold`,
type `Double` in `[0.0, 1.0]`. Effective per-task byte threshold is
`budget * fraction`, where `budget` comes from a backend-installed supplier
(`NativeMemoryUsageTracker.setNativeMemoryBudgetSupplier`).

## Changes
- New `NativeMemoryUsageTracker` and `ResourceType.NATIVE_MEMORY`.
- `OsProbe`: `getProcessRssAnonBytes`, `getFreePhysicalMemorySizeFromProcMeminfo`,
  `getProcessNativeMemoryBytes` (Linux only, reading `/proc/meminfo` and
  `/proc/self/status`).
- `NodeDuressTrackers`: native-memory duress signal.
- DataFusion plugin installs the per-task snapshot supplier; `AnalyticsShardTask`
  now extends `SearchShardTask` so SBP observes it.
- Top-N FFI snapshot via `df_query_registry_top_n_by_current` replaces the
  full-snapshot path.


sandbox/libs/dataformat-native/rust/Cargo.toml
```diff
@@ -81,7 +80,7 @@ opensearch-repository-azure = { path = "../../../plugins/native-repository-azure
 opensearch-repository-fs = { path = "../../../plugins/native-repository-fs/src/main/rust" }
 
 [profile.release]
-lto = true
+lto = "thin"
```
himshikhagupta: Can we revert any changes made for testing only?

sandbox/plugins/analytics-backend-datafusion/rust/src/ffm.rs
```diff
@@ -202,6 +203,52 @@ pub extern "C" fn df_cancel_query(context_id: i64) {
...
+        info!("[nativemem-bp] ffm.df_query_registry_top_n_by_current: capacity=0, nothing to write");
+        return Ok(0);
+    }
+    if out_ptr.is_null() {
+        return Err("null snapshot buffer".to_string());
+    }
+    let out: &mut [WireQueryMetric] =
+        slice::from_raw_parts_mut(out_ptr as *mut WireQueryMetric, cap_entries as usize);
+    let written = snapshot_top_n_by_current(out);
+    info!(
```
himshikhagupta: nit: Can this be debug?

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
```diff
@@ -133,6 +133,21 @@ impl QueryTracker {
         }
     }
 
+    /// Wall-clock duration in nanoseconds, as an `i64` for FFM transport.
+    /// Returns the frozen snapshot if completed, otherwise live elapsed time.
+    /// Elapsed nanos is `u128` internally; saturates at `i64::MAX` (~292 years)
+    /// so it can always be represented as an `i64`.
+    pub fn elapsed_nanos(&self) -> i64 {
```
himshikhagupta: Can you rebase current changes? we are no longer storing queries after completion. For SBP, you need only active queries right?

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
```diff
@@ -159,6 +174,163 @@ pub fn drain_completed_query(context_id: i64) -> Option<Arc<QueryTracker>> {
...
+/// | wall_nanos    | live elapsed or frozen wall time (see `elapsed_nanos()`)  |
+/// | completed     | 1 if the tracker has been marked completed, else 0        |
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
+pub struct WireQueryMetric {
+    pub context_id: i64,
+    pub current_bytes: i64,
+    pub peak_bytes: i64,
+    pub wall_nanos: i64,
+    pub completed: i64,
```
himshikhagupta: Why do you need completed queries to be checked?

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
```diff
@@ -159,6 +174,163 @@ pub fn drain_completed_query(context_id: i64) -> Option<Arc<QueryTracker>> {
...
+}
+
+const _: () = assert!(std::mem::size_of::<WireQueryMetric>() == 5 * 8);
```
himshikhagupta: Why is this check required?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
```diff
@@ -675,6 +660,61 @@ public static DataFusionStats stats() {
...
+    public static Map<Long, QueryExecutionMetrics> queryRegistryTopN(int n) {
+        if (n < 0) {
+            throw new IllegalArgumentException("n must be non-negative: " + n);
+        }
+        if (n == 0) {
+            return Collections.emptyMap();
+        }
+        long bytes = (long) n * QueryRegistryLayout.ENTRY_BYTES;
+        if (bytes > Integer.MAX_VALUE) {
+            throw new IllegalArgumentException("n too large for a single snapshot: " + n);
```
himshikhagupta, line 881: Will this cause panic on the node?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
```diff
@@ -675,6 +660,61 @@ public static DataFusionStats stats() {
...
+            // Defensive: Rust contract is `written <= cap_entries`. A higher value
+            // would mean the buffer was overrun, so refuse to decode past `n`.
+            int rows = (int) Math.min(written, (long) n);
+            // LinkedHashMap preserves the heap-drain order Rust used so logs and
+            // tests see stable iteration; values are O(1) regardless.
+            Map<Long, QueryExecutionMetrics> out = new LinkedHashMap<>(rows);
+            for (int i = 0; i < rows; i++) {
+                long ctxId = QueryRegistryLayout.readContextId(seg, i);
+                // putIfAbsent guards against context_id reuse, which is a Rust-side
+                // caller bug but cheap to defend against here.
```
himshikhagupta: in what cases can this happen? Should this be handled at rust side?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/task/AnalyticsShardTask.java
```diff
@@ -8,26 +8,28 @@
...
+public class AnalyticsShardTask extends SearchShardTask {
 
     public AnalyticsShardTask(long id, String type, String action, String description, TaskId parentTaskId, Map<String, String> headers) {
         super(id, type, action, description, parentTaskId, headers);
     }
-
-    @Override
-    public boolean shouldCancelChildrenOnCancellation() {
-        return false;
-    }
```
himshikhagupta, line 35: Why is this removed? Also should this be true so all child tasks are cancelled if this is cancelled?

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -236,6 +237,180 @@ String readProcLoadavg() throws IOException {
...
+     * Reads the contents of {@code /proc/meminfo} as a list of lines, one per key.
+     * Package-private so tests can override with a synthetic file layout.
+     *
+     * @return the lines from {@code /proc/meminfo}
+     * @throws IOException if an I/O exception occurs reading the file
+     */
+    @SuppressForbidden(reason = "access /proc/meminfo")
+    List<String> readProcMeminfo() throws IOException {
+        return Files.readAllLines(PathUtils.get("/proc/meminfo"));
+    }
```
himshikhagupta: Why do we need this?

server/src/main/java/org/opensearch/search/backpressure/settings/SearchShardTaskSettings.java
```diff
@@ -34,6 +34,12 @@ private static class Defaults {
         private static final double HEAP_PERCENT_THRESHOLD = 0.005;
         private static final double HEAP_VARIANCE_THRESHOLD = 2.0;
         private static final int HEAP_MOVING_AVERAGE_WINDOW_SIZE = 100;
+        // See SearchTaskSettings.Defaults for rationale: 0 / 0.0 keep the native-memory tracker
+        // inert until an operator or backend plugin opts in by setting a non-zero threshold.
+        // The per-task threshold is a fraction of the backend-installed native memory budget,
+        // mirroring HEAP_PERCENT_THRESHOLD.
+        private static final long TOTAL_NATIVE_MEMORY_BYTES_THRESHOLD = 0L;
+        private static final double NATIVE_MEMORY_PERCENT_THRESHOLD = 0.0;
```
himshikhagupta: Does this mean by default it will be disabled? Should we keep it enabled and put some sensible default values?

server/src/main/java/org/opensearch/search/backpressure/trackers/NodeDuressTrackers.java
```diff
@@ -63,8 +72,14 @@ public boolean isNodeInDuress() {
 
     private void updateCache() {
         if (nodeDuressCacheExpiryChecker.getAsBoolean()) {
-            for (ResourceType resourceType : ResourceType.values())
-                resourceDuressCache.put(resourceType, duressTrackers.get(resourceType).test());
+            // Callers may register duress trackers for only a subset of {@link ResourceType}
+            // values (e.g. WLM's WorkloadGroupService registers CPU + MEMORY only, leaving
+            // NATIVE_MEMORY unregistered). Treat absence as "not in duress" so unrelated
+            // call sites aren't forced to register a no-op tracker for every new enum value.
```
himshikhagupta: Should it be registered for WLM as well?

server/src/main/java/org/opensearch/search/backpressure/SearchBackpressureService.java
```diff
@@ -452,6 +541,19 @@ protected void doClose() throws IOException {}
...
+        logger.info(
+            "[nativemem-bp] nodeStats: refreshing all trackers — searchTasks={}, searchShardTasks={}",
+            searchTasks.size(),
+            searchShardTasks.size()
+        );
+        for (TaskResourceUsageTrackers trackers : taskTrackers.values()) {
+            for (TaskResourceUsageTracker tracker : trackers.all()) {
+                tracker.refresh();
+            }
+        }
```
himshikhagupta: Why do we need a refresh now if we didnt need it before?

himshikhagupta (commented):
Can you cleanup code added for testing, check verbose logging and remove/change to debug and add sensible defaults wherever applicable rather than 0.

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
pradeep-L: the completed queries will be filtered out in snapshot_top_n_by_current

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/task/AnalyticsShardTask.java
pradeep-L: have added it back.

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
pradeep-L: this is a check if WireQueryMetric is 40 bytes long , because Java side (QueryRegistryLayout) hardcodes the layout. this helps is a compile time validation so that if any fields are changed .

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
pradeep-L: No, this will not cause the node to panic , This exception will be caught at searchbackPressureService Exception block.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
pradeep-L: have changed the approch for FFM call.

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
pradeep-L: i used this to check the native memory nodeDuress. This will come from native memory admission control CR. 
will remove this once that CR is merged

server/src/main/java/org/opensearch/search/backpressure/SearchBackpressureService.java
pradeep-L: The Search Backpressure service refreshes task resource metrics through taskResourceTrackingService.refreshResourceStats, where the resource utilization values are currently derived from Java thread statistics.

For native memory usage, however, the metrics are sourced from the DataFusion QueryRegistry. Therefore, I need to invoke the native memory tracker explicitly to retrieve the latest memory consumption values.

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
pradeep-L: i have removed it.

server/src/main/java/org/opensearch/search/backpressure/trackers/NodeDuressTrackers.java
pradeep-L: i have introduced a new Resource type called NativeMemory , but i did not want to make WLM related changes , so i am not registering it with WLM.

server/src/main/java/org/opensearch/search/backpressure/settings/SearchShardTaskSettings.java
pradeep-L: yes, by default this will not be enabled. this need to get enabled ponly after we set these values.

server/src/main/java/org/opensearch/search/backpressure/SearchBackpressureService.java
```diff
@@ -114,14 +129,38 @@ public SearchBackpressureService(
...
                 );
+                put(ResourceType.NATIVE_MEMORY, new NodeDuressTracker(() -> {
+                    // Native-memory duress probe. OsProbe.getProcessNativeMemoryFraction()
+                    // returns nativeBytes / NATIVE_MEMORY_DENOMINATOR_BYTES on Linux, or -1.0
+                    // when the underlying RssAnon signal is unavailable (non-Linux or
+                    // kernels < 4.5). The denominator is a POC constant in OsProbe — replace
+                    // with total physical memory or a cgroup limit before shipping.
+                    double used = OsProbe.getInstance().getProcessNativeMemoryBytes();
+                    double totalNative = settings.getSearchTaskSettings().getTotalNativeMemoryBytesThreshold();
+                    double usedFraction = used / totalNative;
```
himshikhagupta, line 145: can totalNative be 0 for any case?

server/src/main/java/org/opensearch/search/backpressure/SearchBackpressureService.java
himshikhagupta: But this will call refresh on all trackers not just native tracker right?

server/src/main/java/org/opensearch/search/backpressure/SearchBackpressureService.java
```diff
@@ -114,14 +129,38 @@ public SearchBackpressureService(
...
+                        "[nativemem-bp] duress-probe: usedFraction={}, threshold={}, breached={}",
+                        String.format(java.util.Locale.ROOT, "%.4f", usedFraction),
+                        String.format(java.util.Locale.ROOT, "%.4f", threshold),
+                        breached
+                    );
+                    return breached;
+                }, () -> settings.getNodeDuressSettings().getNumSuccessiveBreaches()));
             }
-        }),
+        }, new TimeBasedExpiryTracker(System::nanoTime)),
```
himshikhagupta: Why is this change required?

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
```diff
@@ -150,11 +151,192 @@ impl QueryTracker {
...
+            continue;
+        }
+        let bytes = usize_to_i64_saturating(bytes_raw);
+        if bytes == 0 {
+            continue;
+        }
+        if heap.len() < n {
+            heap.push(Reverse(HeapEntry {
+                bytes,
+                tracker: Arc::clone(tracker),
```
himshikhagupta, line 273: Will Arc cloning for all trackers in query registry cause contention under high query concurrency?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
```diff
@@ -579,6 +580,14 @@ public void configureFilterDelegation(FilterDelegationHandle handle, BackendExec
         FilterTreeCallbacks.setHandle(handle);
     }
 
+    @Override
+    public Map<Long, QueryExecutionMetrics> getActiveQueryMetrics() {
+        // Delegate to the plugin that owns the DataFusionService and native runtime.
+        // Keeping the real implementation on DataFusionPlugin lets operators call it
+        // directly (e.g., from a REST action) without going through the SPI.
+        return plugin.getActiveQueryMetrics();
```
himshikhagupta: nit: this fetches only top N not all, rename according to avoid confusion

himshikhagupta (commented):
Please clean up the code

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
pradeep-L: yes , this might cause congestion , so i instead of cloning i have created a new heap struct which just has , contextId, current and peak bytes.

sandbox/plugins/analytics-backend-datafusion/rust/src/lib.rs
```diff
@@ -35,4 +35,4 @@ pub mod session_context;
 pub mod statistics_cache;
 pub mod udf;
 pub mod stats;
-pub mod task_monitors;
\ No newline at end of file
+pub mod task_monitors;
```
himshikhagupta, line 45: nit:revert

sandbox/plugins/analytics-backend-datafusion/rust/src/query_tracker.rs
```diff
@@ -220,8 +351,8 @@ impl Drop for QueryTrackingContext {
     fn drop(&mut self) {
         if let Some(tracker) = &self.tracker {
             tracker.mark_completed();
-            debug!(
-                "Query completed ctx={}: wall={:.3}s, mem_current={}B, mem_peak={}B",
+           debug!(
+               "Query completed ctx={}: wall={:.3}s, mem_current={}B, mem_peak={}B",
```
himshikhagupta: nit:revert

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
```diff
@@ -709,6 +719,61 @@ public static DataFusionStats stats() {
...
+                out.putIfAbsent(ctxId, QueryRegistryLayout.readMetrics(seg, i));
+            }
```
himshikhagupta: Do we need this only for tests?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/task/AnalyticsQueryTask.java
```diff
@@ -49,19 +51,16 @@ public AnalyticsQueryTask(
             id,
             type,
             action,
-            "queryId[" + queryId + "]",
+            (Supplier<String>) () -> "queryId[" + queryId + "] indices[" + String.join(",", indices) + "]",
```
himshikhagupta: Why is this change required?

server/src/main/java/org/opensearch/search/backpressure/SearchBackpressureService.java
```diff
@@ -195,27 +241,34 @@ void doRun() {
...
-            isHeapUsageDominatedBySearchTasks ? searchTasks : Collections.emptyList(),
-            SearchShardTask.class,
-            isHeapUsageDominatedBySearchShardTasks ? searchShardTasks : Collections.emptyList()
-        );
+        final boolean isNativeMemoryInDuress = nodeDuressTrackers.isNativeMemoryInDuress();
+
+        final Map<Class<? extends SearchBackpressureTask>, List<CancellableTask>> cancellableTasks;
+        if (isNativeMemoryInDuress) {
+            cancellableTasks = Map.of(SearchTask.class, searchTasks, SearchShardTask.class, searchShardTasks);
+            nativeMemoryUsageService.refresh();
```
himshikhagupta, line 252: Should we refresh on every SBP tick rather than only when node is in duress?

server/src/main/java/org/opensearch/search/backpressure/settings/NodeDuressSettings.java
```diff
@@ -62,6 +67,42 @@ private static class Defaults {
...
+     * duress probe. Independent from the node-level resource-tracker's
+     * {@code node.native_memory.limit} so the two features can be tuned separately. When this
+     * value is {@link ByteSizeValue#ZERO} (default), the duress probe treats the budget as
+     * unconfigured and stays inert.
+     */
+    private volatile ByteSizeValue nodeNativeMemory;
+    public static final Setting<ByteSizeValue> NODE_NATIVE_MEMORY_LIMIT_SETTING = Setting.byteSizeSetting(
+        "search_backpressure.node_duress.native_memory_limit",
+        ByteSizeValue.ZERO,
+        Setting.Property.Dynamic,
```
himshikhagupta, line 102: Why is this default to 0? DO we need some benchmarking to find the right values?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -211,7 +218,37 @@ public TaskId getParentTask() {
...
+            return new AnalyticsQueryTask(id, type, action, queryId, indices, parentTaskId, headers, cancelAfterTimeInterval);
+        }
+    }
+
+    /**
+     * Walks the DAG and returns the distinct index names referenced by all
+     * {@link OpenSearchTableScan} leaves, preserving discovery order. Used to
+     * populate the coordinator task's description for the Tasks API and slow logs.
+     */
+    private static List<String> collectIndices(QueryDAG dag) {
```
himshikhagupta: This looks unrelated. Did we miss something in rebase?

server/src/main/java/org/opensearch/wlm/tracker/WorkloadGroupResourceUsageTrackerService.java
```diff
@@ -24,7 +24,12 @@
  * This class tracks resource usage per WorkloadGroup
  */
 public class WorkloadGroupResourceUsageTrackerService {
-    public static final EnumSet<ResourceType> TRACKED_RESOURCES = EnumSet.allOf(ResourceType.class);
+    // Explicit set (CPU + MEMORY) rather than {@code EnumSet.allOf(ResourceType.class)}. The
+    // NATIVE_MEMORY entry was added to {@link ResourceType} only to flow duress signals
+    // through NodeDuressTrackers for search backpressure; WLM does not account off-heap
+    // memory per workload group and iterating it here would run the no-op calculator,
+    // produce zero usage, and clutter stats output.
+    public static final EnumSet<ResourceType> TRACKED_RESOURCES = EnumSet.of(ResourceType.CPU, ResourceType.MEMORY);
```
himshikhagupta, line 33: Should we leave a TODO note to eventually integrate WLM as well for Native memory?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/task/AnalyticsQueryTask.java
pradeep-L: for API GET _tasks?detailed=true , Analytics tasks would only have QueryID , So i have added indices . it mirrors what SearchTask does

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
pradeep-L: will use only hashMap and put .

server/src/main/java/org/opensearch/search/backpressure/settings/NodeDuressSettings.java
pradeep-L: This needs to be set explicitly, This is the max native memory assigned.

server/src/main/java/org/opensearch/search/backpressure/SearchBackpressureService.java
pradeep-L: have added check.

server/src/main/java/org/opensearch/search/backpressure/SearchBackpressureService.java
pradeep-L: we should be fetching the memory usage only when node is in native duress, else we will be making FFM call every sec.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
pradeep-L: I added this to collect the list of indices for which the query running.

server/src/main/java/org/opensearch/search/backpressure/settings/SearchShardTaskSettings.java
himshikhagupta: @pradeep-L That was the question, why are we not enabling by default with some sensible values?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
himshikhagupta: Why do we need that?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/task/AnalyticsQueryTask.java
himshikhagupta: Lets follow the pattern in SearchRequest and pass the description, rather than passing params which are not used here
 https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/search/SearchRequest.java#L766

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -153,11 +166,47 @@ public Collection<Object> createComponents(
...
+            return Collections.emptyMap();
+        }
+        Map<Long, QueryExecutionMetrics> metrics = getTopQueriesByMemory();
+        if (metrics.isEmpty()) {
+            return Collections.emptyMap();
+        }
+        Map<Long, Long> out = new HashMap<>(metrics.size());
+        for (Map.Entry<Long, QueryExecutionMetrics> e : metrics.entrySet()) {
+            out.put(e.getKey(), e.getValue().currentBytes());
+        }
```
himshikhagupta, line 369: We are converting from rust map to  `Map<Long, QueryExecutionMetrics>` in NativeBridge and then converting again here. Do we need both conversions? Even if map is small, lets try to limit unnecessary conversions

Bukhtawar (commented):
Will we follow this up with native cpu + non-native cpu checks too for cancellation?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
pradeep-L: getTopQueriesByMemory() returns the backend's full SPI type (QueryExecutionMetrics) since the plugin owns that contract and may grow new fields over time. setSnapshotSupplier in server only needs taskId -> bytes, and server can't depend on the analytics SPI. This method like a adaptor.

server/src/main/java/org/opensearch/monitor/os/OsProbe.java
```diff
@@ -301,6 +302,15 @@ long readRssAnonFromProcSelfStatus() throws IOException {
...
 
+    public long getProcessNativeMemoryBytes() {
+        long rssAnon = getProcessRssAnon();
+        if (rssAnon < 0L) {
+            return -1L;
+        }
+        long jvmHeapMax = JvmStats.jvmStats().getMem().getHeapMax().getBytes();
+        return Math.max(0L, rssAnon - jvmHeapMax);
+    }
+
```
Bukhtawar, line 314: lets substract non-heap committed as well

server/src/main/java/org/opensearch/search/backpressure/settings/NodeDuressSettings.java
```diff
@@ -62,6 +67,42 @@ private static class Defaults {
...
+        Setting.Property.NodeScope
+    );
+
```
Bukhtawar, line 105: how is this different from native_memory_limit used in admission control. Can we ensure we align with values and settings used in the admission control so that we are consistent

pradeep-L: I will take this in a separate CR to keep one settings to get native memory for both admission control and search back pressure

Bukhtawar (approved):

PR #21649, opened 2026-05-13
[QA] Add UnionCommandIT for analytics-engine route

ahkcs:
## Description

Adds a self-contained `UnionCommandIT` under `sandbox/qa/analytics-engine-rest`, mirroring the union surface from `CalciteUnionCommandIT` in `opensearch-project/sql` so the analytics-engine path can be verified inside core without cross-plugin dependencies on the SQL plugin.

### Scope change after rebase

This PR was originally a two-commit change: (1) a Java-side `Timestamp → Timestamp` coercion patch in `DatafusionReduceSink.coerceToDeclaredSchema`, plus (2) the new IT. PR #21690 (merged 2026-05-16) [moved schema coercion handling from Java to Rust](https://github.com/opensearch-project/OpenSearch/pull/21690) — `DatafusionReduceSink.coerceToDeclaredSchema` was deleted in favour of a `typesMatch` tripwire that treats any two `Timestamp` types as matching, and the native side picks up the lowered precision from substrait at registration time via `derive_schema_from_partial_plan`.

The Java-side commit is therefore obsolete and has been **dropped on rebase**. Only the IT remains.

### Validation

`UnionCommandIT` is run against `upstream/main` (post-#21690) with **no Java-side coercion patch**:

```
testBasicUnionTwoSubsearches              ✅
testUnionThreeSubsearches                 ✅
testUnionMidPipelineSingleExplicitDataset ✅
testUnionPreservesDuplicates              ✅
testUnionWithEmptySubsearch               ✅
testUnionWithAllEmptyDatasets             ✅
testUnionWithSingleSubsearchThrowsError   ✅
```

| Test class | Before | After |
|---|---|---|
| `UnionCommandIT` (new) | n/a | **7 / 7** |
| `DatafusionReduceSinkTests` (existing, post-#21690) | 8 / 8 | **8 / 8** |

This validates @sandeshkr419's claim that the Java-side patch is no longer needed: the union surface, including the `Timestamp` cross-precision scenario that originally motivated the patch, works end-to-end on the Rust-side coercion alone.

### What the IT covers

Seven tests against `POST /_analytics/ppl` on the `calcs` dataset:

- `testBasicUnionTwoSubsearches` — two-subsearch shape
- `testUnionThreeSubsearches` — three-subsearch shape
- `testUnionMidPipelineSingleExplicitDataset` — `search ... | union [search ...]`
- `testUnionPreservesDuplicates` — three identical branches → 3× row count
- `testUnionWithEmptySubsearch` — one branch returns 0 rows
- `testUnionWithAllEmptyDatasets` — both branches empty
- `testUnionWithSingleSubsearchThrowsError` — validation error surfaces

Assertions are dataset-independent (sums of branch counts, not hardcoded totals) so the suite stays resilient to `calcs` bulk-data changes.

`calcs` carries `date` / `time` / `datetime` columns, so each branch's partial output schema includes `Timestamp` types — the reduce sink's loose-match tripwire on `Timestamp` is exercised on every test here even when the final projection is `| stats count()`, making this IT the end-to-end regression net for the `Timestamp` handling that #21690 introduced.


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionReduceSink.java
sandeshkr419, line 1: I had moved the schema coercion handling on java side entirely: https://github.com/opensearch-project/OpenSearch/pull/21690 and moved it to rust to eliminate the need of this - I think your change should not be needed now. Can you please validate this with your tests.

Can you please still add up the test-cases for these cases just so that we can ensure cases covered by you are part of tests.

sandeshkr419 (changes requested):

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionReduceSink.java
ahkcs: Confirmed — after rebasing on `main` (which now includes #21690), the Java-side `coerceToDeclaredSchema` is gone and the union surface works end-to-end without my Timestamp patch. The Java commit is dropped on rebase.

Validation on `upstream/main` + this branch (no Java-side coercion):

- `UnionCommandIT` (this PR): **7 / 7** ✅
- `DatafusionReduceSinkTests` (post-#21690): **8 / 8** ✅

`calcs` has `date` / `time` / `datetime` columns, so each branch's partial output schema carries `Timestamp` types through to the reduce sink's `typesMatch` tripwire. That means the IT does exercise the `Timestamp → Timestamp` loose-match path on every test even when the final projection is `| stats count()` — so the case my unit tests used to cover is now covered end-to-end here. The original four unit-pair coercion paths (SEC/MILLI/MICRO/NANO rescaling, overflow, timezone metadata) are now Rust-side concerns under `derive_schema_from_partial_plan` and not directly callable from Java, so I haven't re-added equivalents to the Java test surface — happy to add Rust-side coverage if you'd like, just point me at the right test module.

PR title / description updated to reflect the reduced scope.

ahkcs:
### `sandbox-check` failures are pre-existing flakes on `upstream/main`, not caused by this PR

Both `sandbox-check` runs on this branch failed with the same set of `analytics-engine-coordinator:internalClusterTest` and `composite-engine:internalClusterTest` failures. I reproduced them locally against the **exact base commit** `debf6edcad0` (i.e. `upstream/main` without my IT cherry-picked):

| Run | Module | Failures |
|---|---|---|
| Rebased branch (this PR, run 26114136971) | `analytics-engine-coordinator:internalClusterTest` | 8 (3× `SearchCancellationIT`, 5× `CoordinatorTransportStressIT`) + a few more |
| Rebased branch (retrigger, run 26119419897) | same | same 8 + `CoordinatorSingleNodeTopologyIT`, `CoordinatorTwoNodeTopologyIT`, `WindowSqlIT`, `ValuesSqlIT`, `CompositeLocalRecoveryIT` |
| **Base commit `debf6edcad0` with no IT cherry-picked (local)** | same | **same 8 failures, identical test names** |
| `UnionCommandIT` (this PR, the only file changed) | `analytics-engine-rest:integTest` | **7 / 7 pass** ✅ |

All failures share the same root cause: `IllegalStateException: Allocator[ROOT] closed with outstanding child allocators` — an Arrow allocator lifecycle issue at JVM-test teardown. The failures vary slightly across runs (which is the hallmark of a flake), but the same suites surface consistently.

This PR only adds one file under `sandbox/qa/analytics-engine-rest`, a different gradle module from the failing one, and touches no allocator / coordinator code. There is no causal path from the IT change to these failures.

`sandbox-check` on the same base commit `debf6edcad0` was green in upstream CI ([run](https://github.com/opensearch-project/OpenSearch/actions/runs/26114135355)), so these tests pass intermittently on `main` too.

Happy to retry CI again if helpful, but absent an upstream fix for the allocator-lifecycle flake the PR has no path to a clean `sandbox-check`. Could you take a look or rerun the job on the merge queue? `UnionCommandIT` itself passes 7/7 both locally and in CI.

PR #21650, opened 2026-05-13
[analytics-engine] Rewrite engine-output CAST(TIMESTAMP AS VARCHAR) to to_char

### LINKED ISSUE #5420: Datetime output cast format diverges between Calcite and DataFusion engines
### Description

The `DatetimeOutputCastRule` introduced in #5408 wraps datetime
output columns in `CAST(... AS VARCHAR)` so PPL responses are
rendered as ISO strings. The string produced by that cast is
engine-dependent:

- **PPL Calcite path** (`UnifiedQueryCompiler`) — emits ANSI SQL
  format `2024-01-15 12:00:00` (space separator, no `T`),
  consistent with SparkSQL, PostgreSQL, MySQL, Oracle, and
  SQL Server.
- **Analytics-engine path** (DataFusion native runtime) — emits
  ISO 8601 format `2024-01-15T12:00:00` (with `T` separator).

This means the same PPL query against the same data returns
two different string formats depending on which execution
engine handled it, breaking the wire-format contract that
#5408 set out to enforce.

### Reproduction

```
source=events | fields created_at
```

- Calcite engine → `"2024-01-15 12:00:00"`
- DataFusion engine → `"2024-01-15T12:00:00"`

### Context

Called out as Note 2 of #5408. Filed as a follow-up so the
divergence is tracked outside that already-merged PR.

### Possible directions

1. Replace the implicit `CAST(... AS VARCHAR)` with an explicit
   `DATE_FORMAT(..., '<pattern>')` (or `to_char` /
   `format_datetime`) that pins the wire format on both engines.
2. Push the formatting into the response formatter rather than
   the logical plan, so engine-side cast semantics never leak
   to the wire.
3. Add a DataFusion-side cast adapter that emits the ANSI SQL
   string format to match Calcite.

### Related

- #5408 — introduced `DatetimeOutputCastRule`
- #5250 — parent issue tracking unified-query datetime work

mengweieric:
Fixes [opensearch-project/sql#5420](https://github.com/opensearch-project/sql/issues/5420).

## Problem

PPL's documented timestamp wire format uses a space separator (`"2024-01-15 12:00:00"`). On the analytics-engine route, `DatetimeOutputCastRule` wraps every output datetime field in `CAST(<datetime> AS VARCHAR)` so the unified planner stays backend-agnostic. Calcite renders that cast as the documented space-separated form; DataFusion's Arrow CAST kernel renders it ISO-8601 (`"2024-01-15T12:00:00"`). The `datafusion.format.timestamp_format` session config only affects DataFusion's CLI display pipeline, not the cast kernel.

## Approach

A small pre-Substrait rewrite at the analytics-engine boundary swaps the engine-output cast for an explicit format call:

```
CAST(<TIMESTAMP> AS VARCHAR)  →  to_char(<TIMESTAMP>, '%Y-%m-%d %H:%M:%S')
```

Routed to DataFusion's native `to_char` scalar via the existing Substrait simple-extension mechanism — no new UDFs.

Scope is intentionally narrow:

- **Output Project only.** Accepts any `Project` subclass (`LogicalProject` or `OpenSearchProject`) and descends through the unified planner's single `LogicalSort` system-limit wrapper when present. Inner projections round-trip verbatim. Not recursive.
- **Direct slots only.** Casts buried inside `CASE`/`COALESCE`/UDF arguments round-trip verbatim.
- **`TIMESTAMP → VARCHAR` only.** `DATE`/`TIME` cast cleanly through Arrow already; `CHAR(n)` and `TIMESTAMP_WITH_LOCAL_TIME_ZONE` are out of scope.
- **Seconds-only format.** Matches the contract from `DatetimeOutputCastRule`'s introducing PR ([sql#5408](https://github.com/opensearch-project/sql/pull/5408)): *"PPL Calcite produces ANSI SQL format `2024-01-15 12:00:00`."* `TIMESTAMP(9)` sources resolve to the same seconds-only output.

## Performance

Negligible. Plan-time: a single pass over the root projection's slot list, bounded by SELECT-list width. Runtime: replaces one row-level formatter with another in the same projection — no extra physical stage, same node count, on timestamp output columns only.

## Test plan

- [x] `DatetimeOutputCastRewriterTests` — direct cast rewrite, `Sort(Project)` wrapper handling, inner projection untouched, nested cast inside `CASE` untouched, `CHAR(n)` untouched, `DATE`/`TIME`/integer sources untouched, plain field reference untouched, `TIMESTAMP(9)` source resolves to seconds-only format.
- [x] `DataFusionFragmentConvertorTests#testProjectTimestampOutputCastEmitsToCharExtension` — end-to-end Substrait serialization asserts the formatted-render extension function (not a raw cast) is emitted.
- [x] `./gradlew check -p sandbox -Dsandbox.enabled=true`.
- [x] PPL IT smoke on a force-routed analytics-engine cluster: pre-fix, the canonical `now()` / `current_timestamp()` regression class failed with `T`-separator output; post-fix the same calls return space-separated strings and pass.


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatetimeOutputCastRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+    static final String PPL_TIMESTAMP_FORMAT = "%Y-%m-%d %H:%M:%S";
+
+    private DatetimeOutputCastRewriter() {}
+
+    /**
+     * Walk the RelNode tree and rewrite direct-projection-slot
+     * {@code CAST(<TIMESTAMP> AS VARCHAR)} calls.
+     */
+    static RelNode rewrite(RelNode root) {
+        return root.accept(new RelHomogeneousShuttle() {
```
dai-chen: Could you confirm only the outermost / final Projection will be rewritten?

mch2 (approved):

PR #21652, opened 2026-05-14
Update pull-based ingestion consumer factory to be stateless

### LINKED ISSUE #21648: [BUG] Fix shared state in pull-based ingestion consumer factory
### Describe the bug

The consumer factory maintains state today and is a singleton. This can result in race conditions when a single node hosts multiple shards across index. When there is a race, incorrect ingestion source settings can be used resulting in incorrect behavior.

Factory is supposed to be stateless here. Fix this by removing shared state from the factory.

### Related component

Indexing

varunbharadwaj:
### Description
Pull-based ingestion factory classes need to be stateless for correct behavior to prevent race conditions. The two step initialize and createConsumer can result in race conditions as the factory is a singleton. This PR updates the factory to receive the ingestion source when createConsumer is called.

### Related Issues
Resolves #21648


server/src/main/java/org/opensearch/index/IngestionConsumerFactory.java
```diff
@@ -20,23 +20,31 @@
...
-     *  is passed to the factory through the {@link #initialize(IngestionSource)} method.
+     * Create a consumer to ingest messages from a shard of the streams.
      *
      * @param clientId the client id assigned to the consumer
      * @param shardId the id of the shard
      * @return the created consumer
+     * @deprecated Use {@link #createShardConsumer(String, int, IngestionSource)} instead.
      */
-    T createShardConsumer(String clientId, int shardId);
+    @Deprecated
```
msfroh: Can you mark this (and `initialize()`) as `@Deprecated(forRemoval = true)`? 

That will make it easier for us to make sure to remove it in 4.0.

msfroh (approved):

PR #5441, opened 2022-12-03
[Feature/Identity] Fixes test failures during gradle check

DarshitChanpura:
### Description
As a follow-up for #4798, this PR fixes test failures that were noticed in CI gradle check.

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [x] New functionality has been documented.
  - [x] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff
- [ ] Commit changes are listed out in CHANGELOG.md file (See: [Changelog](../blob/main/CONTRIBUTING.md#changelog))

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21664, opened 2026-05-14
[analytics-engine] Wire PPL spath end-to-end through the analytics-engine route

ahkcs:
## Summary

Closes the analytics-engine gap for the PPL `spath` command. The path-mode variant (`spath path=...`) already worked via the existing `json_extract` wiring; this PR adds the auto-extract mode (`spath input=doc` → `JSON_EXTRACT_ALL` returning `MAP<VARCHAR, VARCHAR>`) and its downstream operators (ITEM lookup, WHERE on extracted values), plus a self-contained QA IT.

## Pass rate

| IT | Before | After |
|---|---|---|
| `sql/integ-test/.../CalcitePPLSpathCommandIT` (analytics-engine route) | 0 / 16 | **16 / 16** |
| `sql/integ-test/.../CalcitePPLSpathCommandIT` (default v2 / Calcite route) | 16 / 16 | 16 / 16 (no regression) |
| `sandbox/qa/analytics-engine-rest/.../SpathCommandIT` (new) | — | **16 / 16** |

Analytics-engine route uses `-Dtests.analytics.force_routing=true -Dtests.analytics.parquet_indices=true` on `:integ-test:integTestRemote`.

Baseline failure modes (before this PR):

- 15 tests: `OpenSearchProjectRule.annotateExpr` → `IllegalStateException: No backend supports scalar function [JSON_EXTRACT_ALL] among [datafusion]`.
- 1 test (`testSimpleSpath`): `EngineBackedIndexer.acquireReader` → `UnsupportedOperationException` — a test-infra issue, fixed on the SQL plugin side in [opensearch-project/sql#5441](https://github.com/opensearch-project/sql/pull/5441).

### 1. `json_extract_all` Rust UDF

`sandbox/plugins/analytics-backend-datafusion/rust/src/udf/json_extract_all.rs` — ~550 lines plus 16 unit tests. Returns Arrow `Map<Utf8, Utf8>`; mirrors `JsonExtractAllFunctionImpl`'s legacy contract:

- Nested objects flatten to dot-paths (`user.name`).
- Arrays append a `{}` segment (`tags{}`).
- Duplicate logical keys merge into a Java-`List.toString()` rendering (`[a, b, c]`).
- JSON nulls render as the literal string `"null"`.
- Top-level scalar / null / empty / whitespace input → NULL map.
- Malformed JSON → empty map (legacy swallowed the parse error).

### 2. SPI enum additions in `analytics-framework`

- `ScalarFunction.JSON_EXTRACT_ALL` enum constant.
- `FieldType.MAP` enum constant + `case MAP -> FieldType.MAP` in `fromSqlTypeName`.

### 3. Capability registrations in `DataFusionAnalyticsBackendPlugin`

- New `MAP_RETURNING_PROJECT_OPS` set (mirrors `ARRAY_RETURNING_PROJECT_OPS`) registered with `FieldType.MAP`. Required because `OpenSearchProjectRule.resolveScalarViableBackends` keys on the call's return type, and JSON_EXTRACT_ALL's `MAP<VARCHAR, VARCHAR>` return wouldn't match `SUPPORTED_FIELD_TYPES`.
- `STANDARD_FILTER_OPS` registered against `FieldType.MAP` so `where doc.user.name = 'John'` (which references the underlying MAP column through ITEM) survives the filter-rule's field-index-keyed viability check.
- Adapter binding `ScalarFunction.JSON_EXTRACT_ALL → JsonExtractAllAdapter`.

### 4. Substrait wiring

- `opensearch_scalar_functions.yaml` — entries for `json_extract_all` and `map_extract`.
- `DataFusionFragmentConvertor.ADDITIONAL_SCALAR_SIGS` — function mappings for both names.
- `JsonFunctionAdapters.JsonExtractAllAdapter` — name-mapping adapter.

### 5. `ITEM(Map, key)` dispatch in `ArrayElementAdapter`

PPL's `result.user.name` lowers to `ITEM(JSON_EXTRACT_ALL(doc), 'user.name')`. Two transforms apply for the MAP-input branch:

- Route to `map_extract` (DataFusion's native map accessor) instead of `array_element`. `map_extract` returns `List<value>` because maps semantically permit duplicate keys, so we wrap the call in `array_element(..., 1)` to project the singleton list back to a scalar.
- Coerce the lookup key (CHAR(N) literal) to VARCHAR before emission so it unifies with the substrait `any1` type-variable binding the YAML declares.

### 6. `ArrowValues.MapVector` flattening in `analytics-engine`

Arrow `MapVector` is laid out as `List<Struct{key, value}>`, so `MapVector.getObject(i)` returns a `JsonStringArrayList` of entry structs rather than a proper map. This change reassembles entries into a `LinkedHashMap<String, Object>` with `Text → String` normalization on both keys and values, so the SQL-plugin response marshaller sees the same shape as a legacy v2 `Map<String, Object>` column.

### 7. QA-side `SpathCommandIT`

Under `sandbox/qa/analytics-engine-rest/`. Mirrors `CalcitePPLSpathCommandIT` one test method to one, sends queries via `POST /_analytics/ppl`, no SQL-plugin dependency. Verifies the full spath surface end-to-end:

- Both modes (path / auto-extract).
- ITEM-on-MAP through eval / where / stats / sort.
- Edge cases (empty / malformed JSON, duplicate keys, array suffix, null input).

Four small datasets under `resources/datasets/spath_{simple,auto,cmd,null}/`.

## Knock-on coverage

Every piece in this PR is reusable beyond `spath`:

- `MAP_RETURNING_PROJECT_OPS` + MAP filter capability are generic for any future PPL function emitting a Calcite MAP RelDataType.
- `ArrayElementAdapter`'s ITEM-on-MAP branch + the `map_extract` YAML entry handle every `result['key']` / `result.field` access on a map column, not just spath's.
- `ArrowValues.MapVector` flattening unblocks any UDF returning `Map<Utf8, Utf8>` from the analytics-engine route.

## Paired SQL plugin PR

[opensearch-project/sql#5441](https://github.com/opensearch-project/sql/pull/5441) — test-infra fix that routes the v2 / Calcite IT's test indices through `TestUtils.createIndexByRestClient`, so the `tests.analytics.parquet_indices=true` toggle has a chance to inject parquet/composite settings before any doc PUTs land.

## How to verify

QA-side IT (no SQL plugin needed):

```bash
./gradlew :sandbox:qa:analytics-engine-rest:integTest \
  -Dsandbox.enabled=true --tests "*SpathCommandIT"
```

End-to-end through the SQL plugin (requires #5441 applied):

```bash
# Start the cluster with all sandbox plugins. The transport.stream feature
# gate is required when analytics-engine and the SQL plugin co-install.
./gradlew :run -Dsandbox.enabled=true \
  -Dopensearch.experimental.feature.transport.stream.enabled=true \
  -PinstalledPlugins="['opensearch-job-scheduler:3.7.0.0-SNAPSHOT', \
    'arrow-flight-rpc', 'analytics-engine', 'parquet-data-format', \
    'analytics-backend-datafusion', 'analytics-backend-lucene', \
    'composite-engine', 'opensearch-sql-plugin:3.7.0.0-SNAPSHOT']"

# In the SQL plugin checkout
./gradlew :integ-test:integTestRemote \
  -Dtests.rest.cluster=localhost:9200 \
  -Dtests.cluster=localhost:9300 \
  -Dtests.clustername=runTask \
  -Dtests.analytics.force_routing=true \
  -Dtests.analytics.parquet_indices=true \
  --tests "org.opensearch.sql.calcite.remote.CalcitePPLSpathCommandIT"
```

## Also addressed

Rebased onto current `main`. Per @mch2's review, no gradle build files are modified by this PR — the streaming-transport feature gate needed for the analytics-engine + SQL-plugin co-install must be passed at run time as `-Dopensearch.experimental.feature.transport.stream.enabled=true`.


gradle/run.gradle
```diff
@@ -89,6 +89,10 @@ testClusters {
           systemProperty 'io.netty.noUnsafe', 'false'
           systemProperty 'io.netty.tryUnsafe', 'true'
           systemProperty 'io.netty.tryReflectionSetAccessible', 'true'
+          // analytics-engine / SQL-plugin co-install needs the stream-transport feature gate
+          // so the SQL plugin's PPL transport handler doesn't double-register against the
+          // analytics-engine's handler (Guice CreationException at boot otherwise).
+          systemProperty 'opensearch.experimental.feature.transport.stream.enabled', 'true'
```
mch2: lets move this as an override in analytics-engine - we shouldn't enable it for regular runs.

ahkcs: Updated — `gradle/run.gradle` is no longer modified by this PR. The override now lives in `sandbox/plugins/analytics-engine/build.gradle` as an `afterEvaluate` block that reaches into `rootProject.testClusters.runTask` and only attaches `opensearch.experimental.feature.transport.stream.enabled=true` when `analytics-engine` is in the installed-plugin list. Regular runs that don't pull analytics-engine see no change.

mch2 (approved):

ahkcs:
Done @mch2 — dropped the `afterEvaluate` block in `sandbox/plugins/analytics-engine/build.gradle`. PR no longer touches any gradle file. The `opensearch.experimental.feature.transport.stream.enabled=true` flag is now expected as a CLI `-D` arg when running the co-install (docs updated).

Issue #5442, opened 2022-12-03
[BUG] API endpoints failing with illegal_argument_exception


lpiob:
**Describe the bug**
Ocassionaly, and without an established cause, following endpoints:
- `/_stats`
- `/_cat/indices`
- `/_cluster/stats`

return 400 Bad Request with following error:
```
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Values less than -1 bytes are not supported: -9223372036854028697b"}],"type":"illegal_argument_exception","reason":"Values less than -1 bytes are not supported: -9223372036854028697b"},"status":400}
```

Other API methods - especially related to searching documents, or checking overall cluster health - work fine.

The negative value is not constant. Other values that we've encountered: `-9223372036854704795b`, `-9223372036853398293b`.

The error goes away after restarting the nodes, and sometimes the error goes away on it's own.

OpenSearch nodes do not log any information when such error is reported. 

**To Reproduce**
Reproduction steps - unknown.

Please advise how to debug further if the error returns.

**Expected behavior**
More verbose errors.

**Plugins**
opensearch-alerting opensearch-anomaly-detection opensearch-asynchronous-search opensearch-cross-cluster-replication opensearch-geospatial opensearch-index-management opensearch-job-scheduler opensearch-knn opensearch-ml opensearch-notifications opensearch-notifications-core opensearch-observability opensearch-performance-analyzer opensearch-reports-scheduler opensearch-sql

**Host/Environment (please complete the following information):**
 - OpenSearch 2.3.0
 - 3 nodes, each with master,ingest,data,remote_cluster_client roles, 4GB of memory for each of them, java memory limits set to 2GB, 100GB disks with about 1GB used for each node, 9 indices, each with 2 shards and 1 replica, about 2 000 000 documents total.
 - Kubernetes installation with https://opensearch-project.github.io/helm-charts/ helm chart

**Additional context**
Test requests and responses recorded in https://gist.github.com/lpiob/56d28fca8378bb94ead9a30893a5dfe8

bowenlan-amzn:
#4474 looks similar

d3vzer0:
Facing the same issue atm with 5 nodes on a vanilla (ie. non-k8s) deployment. Sometimes the issue just disappears by itself 🤔

jgough:
I have been getting this issue too.  I ran:

[...]

PR #21667, opened 2026-05-14
Prevent negative fielddata stats by guarding against stale removals after shard reallocation

### LINKED ISSUE #20363: [BUG] indices stats API fails with negative values
### Describe the bug

On **OpenSearch 3.4.0**, indices stats–related APIs intermittently fail due to **negative values being produced during shard stats serialization**.

When calling APIs such as `/_stats`, `/_cat/indices`, or any request that triggers `indices:monitor/stats`, the request fails with an `illegal_argument_exception`, indicating that values less than `-1` bytes are not supported.

This failure originates from specific shards on certain data nodes and causes partial or complete failure of stats APIs at the cluster level.

### Related component

Storage

### To Reproduce

Exact reproduction steps are **not fully known**, but the issue was observed under the following conditions:

1. Multiple indices (dozens) were subjected to `force_merge` operations **concurrently**.

2. This caused **excessive system resource usage** (CPU and I/O pressure observed).

3. During this period, the following log message was **continuously and repeatedly emitted** on the coordinating node:

```text
[2026-01-01T11:21:04,731][INFO ][o.o.a.u.d.DestinationMigrationCoordinator] [coordinating-02] Cancelling the migration process.
[2026-01-01T11:21:04,908][INFO ][o.o.a.u.d.DestinationMigrationCoordinator] [coordinating-02] Cancelling the migration process.
[2026-01-01T11:21:07,855][INFO ][o.o.j.s.JobSweeper       ] [coordinating-02] Running full sweep
[2026-01-01T11:23:58,598][INFO ][o.o.a.u.d.DestinationMigrationCoordinator] [coordinating-02] Cancelling the migration process.
[2026-01-01T11:23:59,010][INFO ][o.o.a.u.d.DestinationMigrationCoordinator] [coordinating-02] Cancelling the migration process.
[2026-01-01T11:23:59,399][INFO ][o.o.a.u.d.DestinationMigrationCoordinator] [coordinating-02] Cancelling the migration process.
```

4. At some point, the above log **stopped appearing**, and **immediately after that**, stats-related API calls began failing.
   This was detected via the **[OpenSearch Prometheus Exporter plugin](https://github.com/opensearch-project/opensearch-prometheus-exporter)**, which calls indices stats APIs internally.

   Example error observed right after the log stopped:

```text
[2026-01-01T11:24:11,854][WARN ][r.suppressed             ] [coordinating-02] path: /_prometheus/metrics, params: {}
java.lang.NullPointerException: Cannot invoke "org.opensearch.index.shard.DocsStats.getCount()" because the return value of "org.opensearch.action.admin.indices.stats.CommonStats.getDocs()" is null
```

5. From the moment the error started occurring, Grafana panels showed abnormal behavior corresponding to index-level metrics (see attached screenshot).

Image

**Reproducibility:** not deterministic
**Trigger condition:** heavy concurrent `force_merge` under resource pressure (observed, not guaranteed)
**Workaround:** restarting affected data nodes clears the issue temporarily

### Expected behavior

* Shard and index stats APIs should **never emit negative values** for:

  * store size
  * fielddata memory usage
  * document counts
* Even if shard stats are inconsistent or unavailable, the APIs should:

  * sanitize invalid values, or
  * skip the affected shard without failing the entire request

Stats APIs should remain robust and safe for monitoring systems.

### Error examples

From API response:

```json
{
  "error": {
    "root_cause": [
      {
        "reason": "Values less than -1 bytes are not supported: -14624b"
      }
    ],
    "reason": "Values less than -1 bytes are not supported: -14624b"
  },
  "status": 400
}
```

From data node logs:

```text
java.lang.IllegalStateException: Negative longs unsupported, use writeLong or writeZLong for negative numbers [-14624]
    at org.opensearch.core.common.io.stream.StreamOutput.writeVLong
    at org.opensearch.index.fielddata.FieldDataStats.writeTo
    at org.opensearch.action.admin.indices.stats.CommonStats.writeTo
    at org.opensearch.action.admin.indices.stats.ShardStats.writeTo
```

### Affected index / shards

* Failed shards observed on multiple data nodes
* Other shards of the same index succeed

### Impact

* `_cat/indices` fails entirely (HTTP 400)
* Monitoring systems (Grafana / Prometheus exporters) break
* Cluster visibility is significantly degraded
* opensearch prometheus-exporter is not working

### maybe releated issue

https://github.com/opensearch-project/opensearch-prometheus-exporter/issues/266

dzane17:
### Summary
Fixes negative `FieldDataStats.memorySize` causing API serialization error `IllegalStateException: Negative longs unsupported`.

### Root Cause
`FieldDataStats.memorySize` is tracked by a per-shard counter on each node. The counter is incremented when field data is inserted into the node's cache and decremented when field data is removed from the cache. However, cache invalidation and stats decrement runs lazily — fired by a periodic `CacheCleaner` (default 1 min). This timing gap allows field data stats to become corrupt when an `IndexShard` is destroyed and a new one is allocated on the same node with the same `shardId` while the index's `IndexService` (and its `FieldDataCacheListener`) stays alive on that node.

The `IndexService` survives an `IndexShard` recreation when **another shard of the same index keeps the `IndexService` alive on the same node**. Closing the entire index (`_close`/`_open`) destroys the `IndexService` and so does NOT trigger this bug — the stale decrement is dispatched to a listener whose `IndexService` is gone, and is silently dropped.

The trigger conditions on long-running clusters are:

- **Cluster rebalancing of multi-shard indices** — shards naturally migrate between nodes while another shard of the same index keeps the `IndexService` alive on each node.
- **Operator-driven shard moves** — allocation filter changes for node maintenance, zone evacuation, or capacity rebalancing. Repeated ops cause shards to revisit nodes they've previously lived on while sibling shards anchor the `IndexService`.
- **Tier-aware allocation routing** — allocation filter changes driven by hot/warm/cold setups (including ISM-driven routing) move shards between nodes while sibling shards keep the `IndexService` alive.
- **Heavy cluster activity** — concurrent force-merge, indexing surges, or other resource pressure can destabilize allocation and accelerate the conditions above. Segment churn from these workloads also changes the byte size of cached entries, so a stale decrement can be larger than the new shard's freshly-loaded counter.

A single misrouted decrement is often masked by the shard's own positive bytes, so the corruption accumulates slowly and the visible failure only appears after enough stale decrements have piled up. Restarting the affected node temporarily clears the bad state.

If shard A is destroyed and a new shard B is allocated on the same node with the same shard id before that decrement fires (and the `IndexService` survives because another shard anchors it), the removal listener resolves the shard by id, finds B, and decrements B's counter — even though B never cached the entry. This is how B's counter can become negative.

### Local Reproduction
I reproduced one of the error scenarios on a locally running 2-node cluster.

1. Create an index with two shards, pinned to one node.
2. Index documents across multiple flushes so each shard has multiple segments (fielddata cache is bypassed for single-segment loads).
3. Run a sort search to populate the fielddata cache.
4. Three times in a row: widen the allocation filter so the second shard moves to the other node, then re-narrow so it returns. The first shard never moves and keeps the `IndexService` alive on the original node throughout. Re-run the sort search after each round-trip so the new shard instance has cached bytes.
5. Wait long enough for the periodic `CacheCleaner` to fire (~90s, spans at least one default 60s tick).
6. Query `/_nodes/stats/indices/fielddata` — without the fix, the affected node reports a negative `memory_size_in_bytes` and the API returns HTTP 400 with `"Values less than -1 bytes are not supported"`.

The same procedure on a build with this fix applied returns HTTP 200 with non-negative bytes on every node. The reproduction was run 5 times on each branch (10 total) with consistent results: bug reproduced 5/5 on `main`, fix held 5/5 on the fix branch.

### Integration Test Reproduction
Added a new integ test `FieldDataStatsNegativeIT.testReproduceNegativeFieldDataStats` to reproduce negative stats using a similar procedure as above.

Test fails without this PR's fix
```
% ./gradlew :server:internalClusterTest --tests "org.opensearch.index.fielddata.FieldDataStatsNegativeIT"
=======================================
OpenSearch Build Hamster says Hello!
  Gradle Version        : 9.4.1
  OS Info               : Mac OS X 26.5 (aarch64)
  JDK Version           : 25 (Homebrew JDK 25 (25))
  JAVA_HOME             : /opt/homebrew/Cellar/openjdk/25/libexec/openjdk.jdk/Contents/Home
  Random Testing Seed   : EBF5A0D333EA8F82
  Crypto Standard       : any-supported
=======================================

> Task :server:internalClusterTest
WARNING: Using incubator modules: jdk.incubator.vector

REPRODUCE WITH: ./gradlew ':server:internalClusterTest' --tests 'org.opensearch.index.fielddata.FieldDataStatsNegativeIT.testReproduceNegativeFieldDataStats' -Dtests.seed=EBF5A0D333EA8F82 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=kam -Dtests.timezone=Asia/Jerusalem -Druntime.java=25

FieldDataStatsNegativeIT > testReproduceNegativeFieldDataStats FAILED
    java.lang.AssertionError: shard [test_fd_stats][1] bytes: -2676
    Expected: a value equal to or greater than <0L>
         but: <-2676L> was less than <0L>
        at __randomizedtesting.SeedInfo.seed([EBF5A0D333EA8F82:ED8072CEB4AA408D]:0)
        at org.apache.lucene.tests.util.LuceneTestCase.assertThat(LuceneTestCase.java:2123)
        at org.opensearch.index.fielddata.FieldDataStatsNegativeIT.testReproduceNegativeFieldDataStats(FieldDataStatsNegativeIT.java:95)

...
```

### Fix
When a cache entry is inserted, store a unique identifier of the shard that inserted it (using `System.identityHashCode` of the `IndexShard`). When a cache entry is removed, compare the current shard identity to the stored one. If they don't match, the original shard has been replaced — skip the decrement so the new shard's counter stays correct.

### Related Issues
Resolves #20363 & #5442

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/IndexService.java
+++ b/server/src/main/java/org/opensearch/index/IndexService.java
@@ -320,6 +320,10 @@ public class IndexService extends AbstractIndexComponent implements IndicesClust
                 this.indexSortSupplier = () -> null;
             }
             indexFieldData.setListener(new FieldDataCacheListener(this));
+            indexFieldData.setShardIdentityResolver(shardId -> {
+                final IndexShard shard = getShardOrNull(shardId.id());
+                return shard == null ? 0 : System.identityHashCode(shard);
+            });
             this.bitsetFilterCache = new BitsetFilterCache(indexSettings, indicesBitsetFilterCache, new BitsetCacheListener(this));
             this.warmer = new IndexWarmer(
                 threadPool,
@@ -1248,6 +1252,23 @@ public class IndexService extends AbstractIndexComponent implements IndicesClust
                     shard.fieldData().onRemoval(shardId, fieldName, wasEvicted, sizeInBytes);
                 }
             }
+        }
+
+        // Skip the decrement if the shard has been replaced since this entry was cached.
+        // Prevents negative FieldDataStats.memorySize after shard reallocation (#20363).
+        @Override
+        public void onRemoval(ShardId shardId, String fieldName, boolean wasEvicted, long sizeInBytes, int shardIdentity) {
+            if (shardId == null) {
+                return;
+            }
+            final IndexShard shard = indexService.getShardOrNull(shardId.id());
+            if (shard == null) {
+                return;
+            }
+            if (shardIdentity != 0 && shardIdentity != System.identityHashCode(shard)) {
+                return;
+            }
+            shard.fieldData().onRemoval(shardId, fieldName, wasEvicted, sizeInBytes);
         }
     }
 
--- a/server/src/main/java/org/opensearch/index/IndexService.java
+++ b/server/src/main/java/org/opensearch/index/IndexService.java
@@ -1253,23 +1253,6 @@ public class IndexService extends AbstractIndexComponent implements IndicesClust
                 }
             }
         }
-
-        // Skip the decrement if the shard has been replaced since this entry was cached.
-        // Prevents negative FieldDataStats.memorySize after shard reallocation (#20363).
-        @Override
-        public void onRemoval(ShardId shardId, String fieldName, boolean wasEvicted, long sizeInBytes, int shardIdentity) {
-            if (shardId == null) {
-                return;
-            }
-            final IndexShard shard = indexService.getShardOrNull(shardId.id());
-            if (shard == null) {
-                return;
-            }
-            if (shardIdentity != 0 && shardIdentity != System.identityHashCode(shard)) {
-                return;
-            }
-            shard.fieldData().onRemoval(shardId, fieldName, wasEvicted, sizeInBytes);
-        }
     }
 
     public IndexMetadata getMetadata() {
--- a/server/src/main/java/org/opensearch/index/IndexService.java
+++ b/server/src/main/java/org/opensearch/index/IndexService.java
@@ -322,7 +322,7 @@ public class IndexService extends AbstractIndexComponent implements IndicesClust
             indexFieldData.setListener(new FieldDataCacheListener(this));
             indexFieldData.setShardIdentityResolver(shardId -> {
                 final IndexShard shard = getShardOrNull(shardId.id());
-                return shard == null ? 0 : System.identityHashCode(shard);
+                return shard == null ? IndicesFieldDataCache.Key.NO_SHARD_IDENTITY : System.identityHashCode(shard);
             });
             this.bitsetFilterCache = new BitsetFilterCache(indexSettings, indicesBitsetFilterCache, new BitsetCacheListener(this));
             this.warmer = new IndexWarmer(
--- a/server/src/internalClusterTest/java/org/opensearch/index/fielddata/FieldDataStatsNegativeIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/index/fielddata/FieldDataStatsNegativeIT.java
@@ -1,0 +1,128 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.fielddata;
+
+import org.opensearch.action.admin.cluster.node.stats.NodesStatsResponse;
+import org.opensearch.action.admin.indices.stats.IndicesStatsResponse;
+import org.opensearch.action.admin.indices.stats.ShardStats;
+import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.index.query.MatchAllQueryBuilder;
+import org.opensearch.indices.IndicesService;
+import org.opensearch.search.sort.SortOrder;
+import org.opensearch.test.OpenSearchIntegTestCase;
+
+import static org.opensearch.test.hamcrest.OpenSearchAssertions.assertAcked;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+
+/**
+ * Verifies fielddata byte counters do not go negative when a shard is relocated
+ * off and back to the same node while a sibling shard keeps the IndexService
+ * alive. Without the PR #21667 fix, stale CacheCleaner removals are routed by
+ * shardId to the current shard, decrementing a counter on bytes it never cached.
+ *
+ * Two timing levers:
+ * 1. Restore the production default cleanup interval (1m, vs the 1s set by
+ *    OpenSearchIntegTestCase) so the periodic cleaner cannot fire during the
+ *    relocation round-trips and prematurely drain stale entries.
+ * 2. After the round-trips are complete, drive cleanup synchronously via
+ *    IndicesFieldDataCache.clear() instead of waiting on the periodic timer.
+ *    Avoids Thread.sleep and runs deterministically.
+ */
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.TEST, numDataNodes = 0)
+public class FieldDataStatsNegativeIT extends OpenSearchIntegTestCase {
+
+    private static final String INDEX = "test_fd_stats";
+    private static final String FIELD = "name";
+
+    @Override
+    protected Settings nodeSettings(int nodeOrdinal) {
+        return Settings.builder()
+            .put(super.nodeSettings(nodeOrdinal))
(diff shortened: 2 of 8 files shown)


kkhatua:
@jainankitk , @sgup432 
Can you guys please review this PR ?

server/src/main/java/org/opensearch/index/IndexService.java
```diff
@@ -320,6 +320,10 @@ public IndexService(
                 this.indexSortSupplier = () -> null;
             }
             indexFieldData.setListener(new FieldDataCacheListener(this));
+            indexFieldData.setShardIdentityResolver(shardId -> {
+                final IndexShard shard = getShardOrNull(shardId.id());
+                return shard == null ? 0 : System.identityHashCode(shard);
```
cwperks: ~~What is the purpose of using identityHashCode exactly? Why would shardId itself not be useful here?~~

Edit: Need to spend a little more time to understand the bug to review in detail.

dzane17: shardId is the current identifier that is used. However, a shard can be closed then quickly created with the same shardId on the same node (common during ISM transitions, force-merge to UltraWarm, shard relocation). That is the root of this stats issue.

server/src/main/java/org/opensearch/index/IndexService.java
```diff
@@ -1249,6 +1253,23 @@ public void onRemoval(ShardId shardId, String fieldName, boolean wasEvicted, lon
                 }
             }
         }
+
+        // Skip the decrement if the shard has been replaced since this entry was cached.
+        // Prevents negative FieldDataStats.memorySize after shard reallocation (#20363).
+        @Override
+        public void onRemoval(ShardId shardId, String fieldName, boolean wasEvicted, long sizeInBytes, int shardIdentity) {
```
sgup432: Why not move this field data cache specific logic to FieldDataCache class itself?  The guard logic for identity should sit inside the cache itself, not here.

dzane17: This method is already inside the `FieldDataCacheListener` class so it is field data specific.

It also needs to call `indexService.getShardOrNull()` to find the live shard for the comparison, which is per-index state. `IndicesFieldDataCache` is node-level and shared across indices, so it doesn't have access to it.

sgup432: I meant we don't need to expose shardIdentity logic inside the listener logic altogether.

We should instead introduce the staleness logic inside `IndicesFieldDataCache.onRemoval()` directly, and not invoke the `listener.onRemoval()` logic if the key is stale.

By doing this, we don't need to touch the listener interface or leak the identity logic inside the listener. Imagine someone in the future wants to listen to fieldDataCache events, they shouldn't be responsible to implement this redundant shardIdentity logic or they might miss that as well. This is the responsibility of the cache, not the listener implementor.

dzane17: I refactored the identity check to `IndicesFieldDataCache`. However it was not that straightforward because we cannot blindly skip `listener.onRemoval()` for all listeners when a stale shard is found. 

Currently there are 2 known subscribers. First is listener which updates per shard field data stats. Once a shard is "gone" (deleted, moved to diff node, etc.) it is safe to skip `onRemoval()` because we no longer care about that shard's stats. The second listener updates a node level circuit breaker counter. Since this is per-node, we still need to call `onRemoval()` for stale shards.

Since there are two classes of listeners - shard-level & node-level, I separated them so they can be applied accordingly.

server/src/main/java/org/opensearch/index/IndexService.java
sgup432: Lets not use 0 as the fallback here(and elsewhere) to denote that tracking is disabled. As there is still an extremely rare chance that `System.identityHashCode` returns 0 for a shardId, and which can break the logic. 
Its also confusing to read as 0 can be a valid shardId.
Lets use -1 as the fallback everywhere else?

sgup432:
@dzane17 I have one more comment. Can you please check the code coverage and other failures?

server/src/main/java/org/opensearch/index/IndexService.java
dzane17: Done

sgup432:
Thanks @dzane17 for the fix and addressing comments! Code looks much cleaner now

sgup432 (approved):

jainankitk (approved):

PR #21674, opened 2026-05-15
 [Analytics Backend] Source FINAL exchange schema from data-node prepared plan; remove coercion at Java/Rust 

sandeshkr419:
Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### Description
 
 Source the coordinator's exchange wire schema from the data-node's prepared physical plan instead of bridging a Java↔Rust mismatch with a per-cell coercion loop.

**Schema handshake:** coordinator asks the backend for the partial plan's output schema instead of guessing; drops coerceToDeclaredSchema.
**String-variant UDFs**: json-family UDFs accept Utf8/LargeUtf8/Utf8View directly; drops cast inserts on parquet reads. (silent optimization)

  ## Follow-up

Explore lazy schema binding: register the coordinator's streaming table from the first batch's IPC schema, removing the empty-table lowering and the `partialAggOutputSchema` SPI. 
 Not implementing this to avoid taking latency on multi-input cases for joins - need to review that design before going towards that approach.

### Related Issues

### Check List

### DIFF
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ExchangeSinkProvider.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ExchangeSinkProvider.java
@@ -8,6 +8,8 @@
 
 package org.opensearch.analytics.spi;
 
+import org.apache.arrow.vector.types.pojo.Schema;
+
 /**
  * Factory for creating a coordinator-side {@link ExchangeSink}.
  *
@@ -33,4 +35,15 @@ public interface ExchangeSinkProvider {
      *        {@code FinalAggregateInstructionHandler}), or {@code null} when no handler ran
      */
     ExchangeSink createSink(ExchangeSinkContext context, BackendExecutionContext backendContext);
+
+    /**
+     * Returns the Arrow schema the data-node prepared physical plan will emit
+     * for the given partial-aggregate plan bytes, or {@code null} if the backend
+     * has no opinion (caller falls back to a row-type-based derivation).
+     *
+     * @param partialAggBytes bytes from a prior {@link FragmentConvertor#attachPartialAggOnTop}
+     */
+    default Schema partialAggOutputSchema(byte[] partialAggBytes) {
+        return null;
+    }
 }
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FragmentConvertor.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FragmentConvertor.java
@@ -93,18 +93,11 @@ public interface FragmentConvertor {
     }
 
     /**
-     * Returns the Arrow schema the data-node prepared physical plan will emit for
-     * the given partial-aggregate plan bytes. The coordinator uses this to declare
-     * its streaming-input wire schema in agreement with what the producer actually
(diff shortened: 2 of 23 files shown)


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FragmentConvertor.java
```diff
@@ -90,4 +91,15 @@ default byte[] convertFinalAggFragment(RelNode fragment) {
...
     }
+
+    /**
+     * Returns the Arrow schema the data node's prepared physical plan will emit
+     * for the given partial-aggregate plan bytes, or {@code null} if the backend
+     * has no opinion (caller falls back to a row-type-based derivation).
+     *
+     * @param partialAggBytes bytes from a prior {@link #attachPartialAggOnTop}
+     */
+    default Schema partialAggOutputSchema(byte[] partialAggBytes) {
```
mch2: This isn't really Fragment conversion - lets move this underneath DF's sink?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
```diff
@@ -80,6 +80,7 @@ public static void convertAll(QueryDAG dag, CapabilityRegistry registry) {
         if (root.getExchangeSinkProvider() != null && !root.getPlanAlternatives().isEmpty()) {
             AnalyticsSearchBackendPlugin backend = registry.getBackend(root.getPlanAlternatives().getFirst().backendId());
             root.setInstructionHandlerFactory(backend.getInstructionHandlerFactory());
+            root.setFragmentConvertor(backend.getFragmentConvertor());
```
mch2: we should keep this out of Stage, fragment conversion has happend already.  This should imo be a direct native call from DF - we can update ExchangeSinkContext.ChildInput if needed send the fragment instead of schema and derive the schema underneath.

sandeshkr419: Thanks, makes sense. Addressed.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ExchangeSinkProvider.java
```diff
@@ -33,4 +35,15 @@ public interface ExchangeSinkProvider {
...
     ExchangeSink createSink(ExchangeSinkContext context, BackendExecutionContext backendContext);
+
+    /**
+     * Returns the Arrow schema the data-node prepared physical plan will emit
+     * for the given partial-aggregate plan bytes, or {@code null} if the backend
+     * has no opinion (caller falls back to a row-type-based derivation).
+     *
+     * @param partialAggBytes bytes from a prior {@link FragmentConvertor#attachPartialAggOnTop}
+     */
+    default Schema partialAggOutputSchema(byte[] partialAggBytes) {
```
expani, line 46: This is a Coordinator Execution API that's only used in the Planner to infer whether a backend has the capability to create a sink. 

Can we think around reusing AggregateFunction since we already added IntermediateField ( IF ) that has an Arrow type ?

sandeshkr419:
Closing in lieu of https://github.com/opensearch-project/OpenSearch/pull/21690

PR #21677, opened 2026-05-15
[Analytics Engine] Wire broadcast join end-to-end via DataFusion backend

LantaoJin:
### Description
Lands the production path for broadcast-join (milestone 1) execution on the DataFusion backend on top
of PR #21639 (distributed join planning). The `JoinStrategyAdvisor` already picked `BROADCAST` when one join side fit inside the broadcast gates, but `DefaultPlanExecutor` fell back to coordinator-centric.
This PR replaces that fallback with a two-pass dispatcher that:

1. **Pass 1: build only.** Runs the `BROADCAST_BUILD` child stage against a new `BroadcastCaptureSink` that buffers its Arrow batches into a single Arrow-IPC byte buffer.
2. **Pass 2: probe + root.** Appends a `BroadcastInjectionInstructionNode` carrying the IPC bytes to every probe-stage plan alternative, then dispatches through the normal `QueryScheduler` path. Each probe data node decodes the payload, registers it as a `MemTable` under `"broadcast-<buildStageId>"` on its **shard-scan `SessionContextHandle`**, and the native engine executes `Join(ShardScan, NamedScan("broadcast-<buildStageId>"))`.

bc-join

**Scope:** The join runs on each probe data node in parallel, against its local shards plus the injected memtable. The coordinator only gathers joined rows -- there is no coord-side join in this M1 path.

**End-to-end verified.** `BroadcastJoinIT` runs INNER and LEFT OUTER broadcast joins on a 2-node cluster against parquet-backed indices and asserts row-multiset parity with the coord-centric baseline plus a strategy-counter delta proving BROADCAST actually fired.

PR#21639 landed a Volcano split-rule architecture that changed planner-shape contracts our
M0 work was built on. This PR rebases on top of it and reconciles:

**M0/M1 <--> PR#21639 reconciliation**
- **`OpenSearchHashJoinRule` removed.** M0's HASH-shuffle Volcano rule was incompatible with PR#21639's split-rule design — registering it caused a memo explosion (verified empirically: `PlanShapeTests.testJoinThenSort_2shard` hit a 16-minute suite timeout before this fix). M2 hash-shuffle support needs to be redesigned as a sibling split rule. Tracked as a follow-up.
- **`ExchangeInfo` reverted to PR#21639's 2-field shape.** M0 added `partitionCount` for the M2 hash-shuffle work; auto-generated record `toString` mismatch broke ~11 PR#21639-introduced `*PlanShapeTests`. The field comes back when M2 hash-shuffle lands.
- **`OpenSearchJoin` SEMI/ANTI fix preserved.** M0 patched `getOutputFieldStorage()` to return left-only storage for `SEMI`/`ANTI`; PR#21639's reset of the file dropped that. Re-applied (`if (getJoinType().projectsRight())`).
- **`CapabilityRegistry` join indexing removed.** M0's per-format Equi/Theta indexing was built on the M0 `JoinCapability` API; PR#21639 replaced that API with a simpler `record JoinCapability(Set<JoinKind>, Set<String>)`. The indexing was dead code under the new API.
- **Stale fixtures updated.** Six tests asserted M0's coord-centric shape on`shardCount=1`; under PR#21639's split rule, single-shard same-table joins go SHARD-local (no ERs, no separate child stages). Switched fixtures to `shardCount=2` or `3` so they exercise the COORDINATOR-localized path they were designed for.
- **Theta-join tests now assert PR#21639's failure mode.** PR#21639 rejects non-equi joins at the rule level (`!info.isEqui()` → unmarked `LogicalJoin` survives, Volcano's trait converter crashes). The M0/M1 tests that asserted "theta routes coord-centric" now use `expectThrows(RuntimeException.class, ...)` to pin the failure mode. M2 follow-up: re-enable theta joins through a coord-centric fallback consistent with the new split-rule architecture.

### Related Issues

### Check List


LantaoJin:
> https://github.com/opensearch-project/OpenSearch/pull/21677#issuecomment-4457377042 

Real bug: fixed
Finding 5 (Low)
Finding 1 (Medium)

False alarm: no action
Finding 2 (Medium)
Finding 3 (Medium)
Finding 4 (Low)

LantaoJin:
BTW, here is the new configurations added. Any thoughts? @mch2 

1. **analytics.mpp.enabled** — boolean, default true. Master kill switch. When false, every join routes through coordinator-centric regardless of the advisor's decision.
2. **analytics.mpp.broadcast_max_rows** — long, default 1,000,000. Pre-flight row gate; the eligible build side's IndicesStats.primaries.docs.count must be > 0 AND ≤ this value for BROADCAST to fire.
3. **analytics.mpp.broadcast_max_bytes** — ByteSize, default 32MB. Runtime byte cap on the accumulated build-side payload at BroadcastCaptureSink. When exceeded, close() fails the future with BroadcastSizeExceededException.

LantaoJin:
The CI failures not related, it introduced by #21703 which removed the `try { root.close() } catch (IllegalStateException) { logger.warn(...) }` block from `DefaultArrowAllocatorService`

LantaoJin:
> ✅ Gradle check result for [568a099](https://github.com/opensearch-project/OpenSearch/commit/568a0992a773a8a548a6e3dd2608af95d20812a0): [SUCCESS](https://build.ci.opensearch.org/job/gradle-check/77485/)

the previous CI passed, please take a review @mch2  @sandeshkr419

LantaoJin:
@mch2 as discussed offline. I will continue developing the M2 and further works in my branch https://github.com/LantaoJin/OpenSearch/tree/feature/scaffold_mpp_stype_join. Close this PR and reopen later.

PR #21689, opened 2026-05-16
Feature/handle foyer disk full

vishwasgarg18:
---

**Description**

This PR fixes several correctness issues in `FoyerBlockCache`'s key index management and stats accounting, adds a background sweeper to prevent unbounded key index memory growth, and exposes the sweep interval as an operator-configurable setting.

### Problem

The `FoyerCache` implementation used Foyer's `EventListener.on_leave` callbacks to manage its key index and update stats counters. This was fundamentally broken for a disk-only (`memory(1)`) configuration:

- **`Event::Evict`** fires when an entry is promoted from the 1-byte DRAM tier to disk — the entry is still alive, but the handler was incorrectly removing it from the key index and counting it as an eviction.
- **`Event::Remove`** and **`Event::Replace`** fire only when an entry is still in the 1-byte DRAM tier, which is near-zero frequency — so `removed_count`/`removed_bytes`/`used_bytes` were almost never updated.
- Foyer's **disk reclaimer** silently evicts disk blocks with no callback, causing the key index to grow unboundedly with stale entries.

### Solution

1. **Removed `KeyIndexListener` entirely** — all `on_leave` callbacks were no-ops or incorrect for `memory(1)`. Stats are now updated directly in the hot-path operations that call them.

2. **Fixed stats accounting**:
   - `put()` → `used_bytes += size` (unchanged, correct)
   - `evict_prefix()` → `removed_count`, `removed_bytes`, `used_bytes -= size` (moved from unreliable `Event::Remove`)
   - `clear()` → `removed_count`, `removed_bytes` from key index scan before wipe; `used_bytes = 0`

3. **Added background key index sweeper** — a Tokio task that periodically calls `inner.contains(key)` (O(1) in-RAM indexer lookup, no disk I/O) on each key index entry and prunes stale ones, updating `eviction_count`, `eviction_bytes`, and `used_bytes`.

4. **`key_byte_size()` utility** — parses `end - start` from `"path\x1Fstart-end"` key strings to derive exact byte sizes without storing sizes in the index. This enables accurate `eviction_bytes`, `removed_bytes`, and `used_bytes` accounting.

5. **Configurable sweep interval** — `block_cache.foyer.key_index_sweep_interval_seconds` (default: 0 = 30s). Wired end-to-end from `opensearch.yml` → Java → Rust.

6. **Fixed `inner.contains()` documentation** — correctly describes it as an in-RAM sharded hashmap lookup with possible false positives on 64-bit hash collision (harmless), not disk I/O.

7. **Fixed stale setting references** — `format_cache.*` → `block_cache.foyer.*` throughout.

### Testing

- 82 Rust unit tests (all passing), including new tests for:
  - `key_byte_size()`: 8 unit tests covering all edge cases
  - Sweeper: idempotency, stale entry removal, byte accounting, malformed keys
  - `evict_prefix()`: byte and used_bytes correctness
  - `clear()`: removed_count/removed_bytes update
- Java tests updated: `testGetSettingsRegistersAllFourSettings` → `testGetSettingsRegistersAllSettings` (5 settings, added `KEY_INDEX_SWEEP_INTERVAL_SETTING` assertion)
- `FoyerBlockCacheTests`: added `sweepIntervalSecs` validation guard test

### Key design decisions

- **No type change to `key_index`** — stays `DashMap<String, Vec<String>>`. Sizes derived from keys at use time.
- **`used_bytes` over-counts between sweeps** — acceptable known limitation; corrected at each sweep cycle.
- **Sweep interval tuning guidance** — set to your `_nodes/stats` polling interval for optimal freshness.

---

**Check List**
- [x] Functionality includes testing (82 Rust tests + Java tests updated)
- [ ] API changes companion pull request created, if applicable
- [ ] Public documentation issue/PR created, if applicable


sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18, line 324: Ok(Some(e)) => { ... Some(Bytes::...) }
    _ => {                   // ← **Err AND Ok(None) both land here**
        self.stats.miss_count.fetch_add(1, ...);
        None
    }

vishwasgarg18:
Duplicate - https://github.com/opensearch-project/OpenSearch/pull/21704

PR #21697, opened 2026-05-17
Update pull-based ingestion multi-threaded writer flow batchStartPointer computation

varunbharadwaj:
### Description
1. Update the pull-based ingestion multi-threaded write flow to handle the rare race scenario where a flush is called before the very first message is read by a processor thread, but another processor thread has read a message. In this case, use the first queued message pointer for determining batchStartPointer. This can only happen during uneven hash distribution on docID partitioning which is rare.
2. Update Kafka consumer to be synchronized to avoid concurrent access when reset API is called.

### DIFF
--- a/plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
+++ b/plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
@@ -148,7 +148,7 @@ public class KafkaPartitionConsumer implements IngestionShardConsumer<KafkaOffse
      * @throws TimeoutException
      */
     @Override
-    public List<ReadResult<KafkaOffset, KafkaMessage>> readNext(
+    public synchronized List<ReadResult<KafkaOffset, KafkaMessage>> readNext(
         KafkaOffset offset,
         boolean includeStart,
         long maxMessages,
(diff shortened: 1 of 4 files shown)


plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
```diff
@@ -263,14 +263,12 @@ public int getShardId() {
 
     /**
      * Compute Kafka offset based lag as the difference between latest available offset and last consumed offset.
```
yupeng9, line 265: with the annotation here removed, does it mean it's intended to have other caller than the poller thread to access this method?

server/src/main/java/org/opensearch/indices/pollingingest/PartitionedBlockingQueueContainer.java
```diff
@@ -169,10 +185,17 @@ public void updateErrorStrategy(IngestionErrorStrategy errorStrategy) {
     }
 
     /**
-     * Returns the current shard pointers from each message processor thread.
+     * Returns the list of lowest/first shard pointer per partition that is in-flight or already processed.
+     * If the processor thread is yet to process the very first message, the first queued pointer for the partition is considered.
```
yupeng9, line 189: why it's ok the consider the first queued pointer? and is it guaranteed that the pointer exists?

varunbharadwaj: The issue we are trying to solve here is if we have 2 blocking queues/partitions, and we write a message to both of them, and one processor thread has not yet consumed its first message (it is still in the queue). Here, we have a window where we might have written offsets 1 and 2 to the queues, but the batchStartPointer can be read as 2. 

This is not an issue after all processor threads have read the very first message. So to avoid this, we fallback to the first queued message if a processor thread has not yet consumed anything. This is okay because the messages are written to the queue in the order they are consumed.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
varunbharadwaj: Only the poller holds an instance of the consumer and will be the only one calling. But with this, we synchronize the access to the consumer implementations so removing the comment.

server/src/main/java/org/opensearch/indices/pollingingest/PartitionedBlockingQueueContainer.java
varunbharadwaj: Updated the comments explaining the fallback logic. The fallback pointer can possibly be null, which is okay, and gets filtered out.

yupeng9 (approved):

msfroh (approved):

PR #21698, opened 2026-05-17
[analytics-engine] Skip unsupported field types when building Calcite schema

mengweieric:
## Problem

Queries against any index that contains a field type the analytics-engine cannot scan (e.g. `geo_point`, `geo_shape`, `alias`) fail at planning time with `IllegalArgumentException: Unsupported OpenSearch field type: <type>` — even when the query never references the unsupported column.

## Solution

Drop the unsupported column from the Calcite row type during schema build instead of throwing.

- Queries that reference the dropped column → clean `column not found` validator error.
- Queries that don't reference it → plan and execute normally.

## Verification

**Unit tests** (16/16 green, 9 new): cover Calcite end-to-end validator behavior on dropped columns, mixed mappings, nested objects with unsupported leaves, all-unsupported indices, and the full type catalogue.

**Same-cluster A/B IT regression sweep** on a force-routed analytics-engine cluster, identical 4872-test run with fix OFF vs fix ON:

| | fix-OFF (baseline) | fix-ON | Δ |
|---|---|---|---|
| Pass | 1401 | 1405 | **+4** |
| Fail | 3266 | 3262 | **−4** |
| Skip | 205 | 205 | 0 |
| Unique tests hitting `Unsupported field type: geo_point` | **214** (across 5 classes) | **0** | **−214** |
| Cluster-log `Unsupported OpenSearch field type` throws (raw, JVM-level) | **443** | **0** | **−443** |
| Regressions on previously-passing tests | — | **0 classes** | — |

The 214 unblocked tests span `CalciteDataTypeIT`, `CalciteDateTimeComparisonIT`, `CalciteGeoPointFormatsIT`, `CalciteMultiValueStatsIT`, `CalcitePPLCastFunctionIT`. The pass-delta is small (+4) because most unblocked tests immediately re-fail on the *next* engine gap (`TIMESTAMP` / `IS_NULL` / `IS_NOT_NULL` scalar functions, etc.) — this fix is the necessary prerequisite for those tests to ever reach those gaps in the first place. The load-bearing metric is the 443 → 0 / 214 → 0 elimination, not the pass-rate %.

Per-class pass wins: `CalciteGeoPointFormatsIT` 0 → 2, `CalcitePPLStringBuiltinFunctionIT` 44 → 46.

## Test plan

- [x] `:sandbox:libs:analytics-api:check`
- [x] `:sandbox:plugins:analytics-engine:check`
- [x] `spotlessCheck` on both modules
- [x] Unit tests for all field-type scenarios (supported, unsupported, nested, mixed, malformed)
- [x] Calcite validator behavior tested end-to-end via `Frameworks.getPlanner`
- [x] Same-cluster A/B IT regression sweep — error fully eliminated, 0 regressions

### DIFF
--- a/sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/OpenSearchSchemaBuilder.java
+++ b/sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/OpenSearchSchemaBuilder.java
@@ -64,32 +64,31 @@ public class OpenSearchSchemaBuilder {
     }
 
     /**
-     * Maps an OpenSearch field type string to a Calcite SqlTypeName.
+     * Maps an OpenSearch field type string to a Calcite SqlTypeName, or {@code null} when the type
+     * has no scalar Calcite representation here (geo_point, geo_shape, nested, completion, …) or
+     * is unrecognized. Callers omit the column from the schema so a query referencing it surfaces
+     * a Calcite "column not found" via the validator rather than a planner-time crash.
      *
      * <p>Type mapping:
      * <ul>
      *   <li>keyword/text/match_only_text -> VARCHAR</li>
-     *   <li>long -> BIGINT</li>
-     *   <li>unsigned_long -> BIGINT</li>
-     *   <li>integer -> INTEGER</li>
-     *   <li>short -> SMALLINT</li>
-     *   <li>byte -> TINYINT</li>
-     *   <li>double -> DOUBLE</li>
-     *   <li>float -> REAL</li>
-     *   <li>half_float -> REAL</li>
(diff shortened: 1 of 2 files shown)


mengweieric:
The sandbox-check failure is in CoordinatorResilienceIT.testShardHostNodeKillAndRestart and
  testCoordinatorSurvivesIntermittentNetwork — distributed-quorum resilience tests, with no overlap with
  OpenSearchSchemaBuilder.java. Same branch passed sandbox-check at 01:42 UTC (run 25978261158); concurrent unrelated
  PRs (#21691, #21681, #21680) also show sandbox-check fail — known flaky pattern.

mch2 (approved):

PR #21699, opened 2026-05-17
analytics-engine: Refactor Scheduler

mch2:
### Description

This PR introduces a series of changes to our scheduler to better organize
  responsibility, position us for shuffle/broadcast flows, explain/metrics, and
  shard-level retries. Also adds a hook for child stages to hand metadata to parent
stages on completion. (see StageListenerWiring). 

  The analytics-engine coordinator had three concerns tangled inside `QueryScheduler` + `PlanWalker` + `ExecutionGraph`: per-query lifecycle, graph topology, and task
  dispatch. Each `Stage*Scheduler` did two unrelated jobs: build the `StageExecution`
  AND dispatch its tasks. That left "when and how to run a task" scattered across one
  class per stage type, and nothing in the system had a single view of dispatch.

  This refactor draws the line cleanly:

  - **Stages decide *what* to dispatch** — they produce `StageTask`s and expose a
    runner to initiate execution.
  - **The scheduler decides *when* and *how*** — `QueryScheduler::scheduleStage` is
    the one site that transitions tasks to RUNNING and hands them to the runner.

  ## What changed

  - **`QueryExecution` (new)** — owns the per-query lifecycle: tracks state, fires
    the result listener once, cleans up at terminal. Pulled out of `QueryScheduler`.
  - **`ExecutionGraph`** — now just topology. The graph-build methods from `PlanWalker`
    moved here; the scheduler decides what to run.
  - **`Stage*Scheduler` → `Stage*ExecutionFactory` (rename)** — these classes build
    stage executions. They never actually scheduled anything; the old name was
    misleading.
  - **`TaskRunner` (new) + `LocalTaskRunner`, `ShardTaskRunner`** — per-stage task
    runners. Local runs on a virtual-thread executor; Shard goes over Arrow Flight.
  - **`StageTask` + `StageTaskId` + `StageTaskState` (new)** — tasks are now
    first-class with their own identity and state, instead of bare partition indices.
  - **`PlanWalker` (deleted)** — its job is split between `ExecutionGraph.build()`
    (build the graph) and `QueryExecution` (run the lifecycle).

  ```mermaid
 flowchart LR
      QS[QueryScheduler]
      QE[QueryExecution]
      EG[ExecutionGraph]
      SE[StageExecution]
      TR[TaskRunner]
      ST[StageTask]

      QS -- creates --> QE
      QS -- builds --> EG
      EG -- contains --> SE
      SE -- owns --> TR
      SE -- produces --> ST
      QS -- dispatches --> ST
      TR -- runs --> ST
  ```

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/AbstractStageExecution.java
```diff
@@ -79,54 +85,83 @@ public final void addStateListener(StageStateListener listener) {
...
         if (previous == State.CREATED) {
             metrics.recordStart();
         }
-        if (isTerminal(target)) {
+        if (target.isTerminal()) {
             metrics.recordEnd();
         }
 
-        // Transition succeeded. Fire state listeners.
         for (StageStateListener l : stateListeners) {
```
Bukhtawar, line 214: Should it use `CopyOnWriteArrayList` to ensure publication safety

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/StageExecution.java
```diff
@@ -33,67 +28,74 @@ public interface StageExecution {
...
+
+    /** Idempotent. Transitions to CANCELLED and tears down stage-owned resources. */
+    void cancel(String reason);
+
+    // ── Scheduler-driven dispatch hooks ───────────────────────────────────
+
+    /** Empty until {@link #start()} populates from resolved targets. */
+    default List<StageTask> tasks() {
+        return List.of();
+    }
```
Bukhtawar, line 57: This can be mutated by anyone who has the reference. Ideally it should only be allowed for `QueryScheduler`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/StageListenerWiring.java
```diff
@@ -0,0 +1,94 @@
...
+ * <p>Parent: on FAILED / CANCELLED, sweep still-running children — they shouldn't
+ * keep producing into a sink whose owner has terminated.
+ *
+ * <p>Thread-safe under the documented stage contracts: pending counter is atomic;
+ * child state reads + cancel are idempotent; {@code wire} runs during graph build
+ * (before any transitions fire), so listener registration never races with firing.
+ *
+ * @opensearch.internal
+ */
+final class StageListenerWiring {
```
Bukhtawar: This looks a bit weird, maybe better placed at `AbstractStageExecution`

```
void attachChildren(List<StageExecution> children, Consumer<StageExecution> scheduler) {
      this.children = children;
      ....
      for (StageExecution child : children) {
          child.addStateListener(this::onChildStateChange);
      }
  }
```

mch2: good call. moved to AbstractStageExecution.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/StageExecution.java
mch2: I've got a commit to trim StageExecution and make all the core scheduling methods pkg private, this will be only what we expose about a stage outside.  Would like to follow up with that one though as it touches a lot of files.

mch2:
cut this down a bit in latest revision - have a stacked pr https://github.com/opensearch-project/OpenSearch/pull/21712 to add cancellation e2e and convert the reduce cycle in DataFusionReduce sink to be a stage task.

sandeshkr419 (approved):

PR #21700, opened 2026-05-17
Add ingestion-hive plugin for pull-based ingestion from Hive tables

### LINKED ISSUE #21451: [RFC] Hive Table Source for Pull-Based Ingestion
Pull-Based Ingestion (PBI) enables OpenSearch to ingest data from streaming sources (Kafka, Kinesis) directly into Lucene segments without a traditional indexing pipeline. However, a large portion of analytical data lives in Hive tables on storage systems such as HDFS or S3. To ingest this data via PBI today, users must first publish it to Kafka, then consume from Kafka into OpenSearch. This intermediate step adds operational complexity, infrastructure cost, and ingestion latency.

Alternative approaches exist (e.g., opensearch-hadoop Spark connector for batch loading), but they do not benefit from PBI's incremental ingestion model.

A new `ingestion-hive` plugin that implements `IngestionConsumerPlugin`, allowing OpenSearch to read from Hive tables directly via PBI. The pipeline simplifies to: Hive table → OpenSearch.

Each shard independently...

1. Connects to Hive Metastore via Thrift
2. Discovers new partitions since its last watermark (incremental fetch)
3. Determines which partitions it owns via consistent hashing
4. Reads Parquet files from those partitions via Hadoop FileSystem API
5. Emits rows as JSON messages into PBI's standard processing pipeline

No coordinator process is involved. Each shard operates in complete isolation, the same way Kafka shards independently consume their assigned partitions.

#### Example configuration

```json
PUT /events-index
{
  "settings": {
    "number_of_shards": 3,
    "number_of_replicas": 0,
    "replication.type": "SEGMENT",
    "ingestion_source": {
      "type": "HIVE",
      "pointer.init.reset": "earliest",
      "param.metastore_uri": "thrift://metastore:9083",
      "param.database": "analytics",
      "param.table": "events",
      "param.monitor_interval": "300s",
      "param.partition_order": "partition-name",
      "param.consume_start_offset": "dt=2026-04-01"
    }
  }
}
```

#### Architecture

```
┌─────────────────────────────────────────────────────────────────┐
│                        Hive Metastore                            │
│                       (Thrift RPC)                               │
└──────────┬──────────────────┬──────────────────┬────────────────┘
           │                  │                  │
    getPartitionsByFilter  (same)             (same)
           │                  │                  │
           ▼                  ▼                  ▼
┌──────────────────┐ ┌──────────────────┐ ┌──────────────────┐
│   Shard 0        │ │   Shard 1        │ │   Shard 2        │
│                  │ │                  │ │                  │
│ 1. Query MS      │ │ 1. Query MS      │ │ 1. Query MS      │
│ 2. Filter by     │ │ 2. Filter by     │ │ 2. Filter by     │
│    hash % 3 == 0 │ │    hash % 3 == 1 │ │    hash % 3 == 2 │
│ 3. List files    │ │ 3. List files    │ │ 3. List files    │
│ 4. Read Parquet  │ │ 4. Read Parquet  │ │ 4. Read Parquet  │
│ 5. Write Lucene  │ │ 5. Write Lucene  │ │ 5. Write Lucene  │
└──────────────────┘ └──────────────────┘ └──────────────────┘
           │                  │                  │
           ▼                  ▼                  ▼
┌─────────────────────────────────────────────────────────────────┐
│         Storage (local / HDFS / S3 via Hadoop FileSystem)       │
└─────────────────────────────────────────────────────────────────┘
```

#### Dependencies

| Component | Library | Purpose |
|-----------|---------|---------|
| Metastore client | Thrift-generated code from Hive's `hive_metastore.thrift` IDL | Partition listing, table metadata. No dependency on Hive or Hadoop server libraries. Same approach as [Trino's hive-thrift](https://github.com/trinodb/hive-thrift). |
| Thrift transport | `libthrift` | Wire protocol for Metastore communication |
| File access (local/HDFS) | `hadoop-client-api` + `hadoop-client-runtime` (version aligned with `repository-hdfs` via `versions.hadoop3`) | FileSystem abstraction for local and HDFS. Same pattern as OpenSearch's `repository-hdfs` plugin. |
| File access (S3) | `S3HadoopFileSystem` (plugin-internal) + AWS SDK v2 individual modules | Minimal read-only Hadoop `FileSystem` for S3, registered for s3/s3a/s3n schemes. Avoids `hadoop-aws` which requires the ~300MB `software.amazon.awssdk:bundle`. |
| Parquet reading | `parquet-hadoop` + `parquet-column` | Parquet file reading and column decoding |
| Kerberos | `hadoop-minikdc` (test only) + Hadoop's built-in Kerberos support | SASL/GSSAPI authentication for secured Metastore |

#### Processing Flow
**Initialization**: On the first read, each shard connects to Hive Metastore via Thrift and initializes the Hadoop FileSystem.

**Partition Discovery (Incremental Fetch)**: Every `monitor_interval`, each shard queries Metastore for partitions added since its watermark. Only new partitions are returned, so the response size is proportional to the number of newly added partitions, not the total partition count.

**File Reading**: For each assigned partition, the shard lists data files via Hadoop FileSystem API, reads Parquet rows, converts them to JSON, and feeds them into PBI's processing pipeline.

**Continuous Monitoring**: After all current partitions are processed, the shard waits for `monitor_interval` to elapse, then re-queries Metastore for partitions newer than its watermark.

#### Example Flow
Table: `analytics.events`, 3 shards, `monitor_interval=5m`, `consume_start_offset` not set.

```
Time 0:00 - Index created
  Metastore has partitions: [dt=2026-04-13, dt=2026-04-14, dt=2026-04-15]
  Shard 0 gets: [dt=2026-04-13]         (2 files, 150 rows)
  Shard 1 gets: [dt=2026-04-14]         (2 files, 150 rows)
  Shard 2 gets: [dt=2026-04-15]         (2 files, 150 rows)

Time 0:03 - All shards finish
  Total documents in index: 450

Time 0:05 - monitor_interval fires, all shards re-query Metastore
  Metastore now has: [..., dt=2026-04-16]  (new partition added by upstream ETL)
  dt=2026-04-16 hashes to shard 1
  Shard 1 starts processing dt=2026-04-16

Time 0:06 - Shard 1 finishes dt=2026-04-16
  Total documents in index: 600
```

#### Recovery
Each shard tracks its position within the table using a pointer that encodes the current partition, file, row index, and a monotonic sequence number.

| Component | Type | Example |
|-----------|------|---------|
| partitionName | String | `dt=2026-04-15` |
| filePath | String | `file:///warehouse/events/dt=2026-04-15/part-00001.parquet` |
| rowIndex | long | `4200` |
| sequenceNumber | long | `12345` |

This pointer is periodically committed to Lucene. When a shard restarts, it reads the last committed pointer, seeks to the partition/file/row encoded in it, and resumes reading. Rows between the last commit and the crash are re-read (at-least-once). The plugin sets `_id` to the serialized pointer, so re-read rows produce the same `_id` and result in idempotent upserts, achieving effectively exactly-once semantics.

#### Configuration
| Parameter | Description | Default |
|-----------|-------------|---------|
| `param.metastore_uri` | Hive Metastore Thrift endpoint | (required) |
| `param.database` | Hive database name | (required) |
| `param.table` | Hive table name | (required) |
| `param.monitor_interval` | How often to check for new partitions (TimeValue format: `60s`, `5m`) | `300s` |
| `param.partition_order` | How to determine partition ordering: `partition-name`, `partition-time`, `create-time` | `partition-name` |
| `param.partition_time_pattern` | Pattern for extracting time from partition values (e.g., `$year-$month-$day $hour:00:00`). Required when `partition_order=partition-time` | (none) |
| `param.consume_start_offset` | Start from this partition onward (exclusive) | (none, reads all) |
| `param.transport_mode` | Thrift transport: `unframed` (Hive 4) or `framed` (Hive 3) | `unframed` |
| `param.max_retries` | Maximum connection retry attempts | `3` |
| `param.retry_interval` | Delay between retries (TimeValue format) | `5s` |
| `param.authentication` | Authentication mode: `none` or `kerberos` | `none` |
| `param.kerberos_principal` | Client Kerberos principal (required if authentication=kerberos) | (none) |
| `param.kerberos_keytab` | Path to keytab file (required if authentication=kerberos) | (none) |
| `param.metastore_service_principal` | Metastore service principal. `_HOST` is replaced with the metastore hostname | (none, defaults to `hive/<host>`) |
| `param.connect_timeout` | Metastore connection timeout (milliseconds) | `10000` |
| `param.hadoop_config.*` | Pass-through Hadoop configuration properties (e.g., `param.hadoop_config.fs.s3a.endpoint` for custom S3 endpoint) | (none) |

**Partition Ordering**: `partition_order` determines how the plugin decides which partitions are "newer" than the watermark.

`partition-name` (default): Lexicographic comparison. Server-side Metastore filter (`partition_column > 'watermark'`). Works when partition values sort lexicographically in chronological order (e.g., `dt=2026-04-01`). Breaks with non-zero-padded numerics.

`partition-time`: Extracts timestamps from partition values using a configurable pattern, then compares them. Requires full partition list retrieval with client-side filtering. Same approach as [Flink's `partition-time` ordering](https://nightlies.apache.org/flink/flink-docs-release-1.15/zh/docs/connectors/table/hive/hive_read_write/).

`create-time`: Uses the partition's `createTime` from Metastore. Requires full partition list retrieval with client-side filtering. Works regardless of naming conventions. Breaks with backfilled partitions.

**Backfill handling**: When backfilling historical partitions (adding old dates after newer dates have been processed), neither ordering mode guarantees correct incremental detection. The recommended workflow for backfill is to delete and recreate the ingestion index with `consume_start_offset` set to cover the backfilled range, or to use `pointer.init.reset: earliest` to re-scan the full table.

#### Constraints

- Partitions must be immutable once registered in Metastore. Files appended to an existing partition after it has been processed will not be detected.
- Partition addition must be atomic from Metastore's perspective. A partition should be fully written before being registered.
- No row-level UPDATE or DELETE. Plain Hive tables have no change tracking. The plugin supports INSERT/append semantics only.
- Segment replication required. PBI does not support document replication.

These are the same constraints documented by [Flink's Hive streaming source](https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/connectors/table/hive/hive_read_write/#streaming).

#### Shard-to-Partition Relationship
Partitions are assigned to shards via `hash(partitionName) % numShards`. Unlike Kafka (where partition count is fixed at topic creation), Hive partition count changes over time as upstream jobs add new partitions.

| Scenario | Behavior |
|----------|----------|
| partitions > shards | Each shard handles multiple partitions. Distribution is roughly even due to hashing. |
| partitions = shards | Each shard handles one partition. |
| partitions < shards | Some shards have no work and sit idle. Not an error, because new partitions may appear later. |

Data skew is possible: if partition sizes vary, the shard assigned the large partition takes longer. No dynamic rebalancing is performed. With incremental partition fetch, per-poll Metastore load is proportional to the number of new partitions, not total partition count, so scaling to many shards does not create a Metastore bottleneck.

#### Design Consideration: Coordinator
Distributed engines (Spark, Flink, Trino) use a coordinator to centralize partition discovery and task assignment. This proposal does not include one, as PBI's framework assumes independent, shared-nothing shards. The coordinator-less design works for Hive tables because incremental fetch keeps Metastore load constant and consistent hashing distributes work without coordination. A coordinator can be introduced later without changing the consumer interface.

#### Future Work

- **Iceberg source**: Iceberg's snapshot metadata contains file lists directly, which removes the need for directory listing. Incremental ingestion can track snapshot IDs as the watermark.
- **Hudi source**: Hudi's CDC feature (RFC-51, Hudi 0.13.0+) records row-level changes within file groups. Since each file group is self-contained, CDC can be computed per-shard without cross-shard shuffling.
- **ORC/Avro format support**: The file reading logic is abstracted behind an interface. Adding ORC or Avro requires a new reader implementation.
- **AWS Glue Data Catalog**: The catalog access is abstracted behind an interface. AWS Glue support can be added without changing the consumer logic.
- **Coordinator-based design**: For very large tables or dynamic load balancing, a coordinator could centralize Metastore queries and distribute assignments via shared storage.
- **Zero-copy payload passing**: The PBI framework currently requires serialized JSON payloads, which forces a Map → JSON → Map round-trip for sources like Hive that already produce structured data. Allowing structured payloads directly would eliminate this overhead.

### Related component

Indexing

**Use a coordinator (Spark Driver / Flink SplitEnumerator pattern)**: See "Design Consideration: Coordinator" above. Not adopted because PBI's framework assumes shared-nothing shards.

**Use Data Prepper or Spark + opensearch-hadoop**: Keep the existing pipeline (Hive → external engine → OpenSearch). This works today but requires intermediate infrastructure and does not benefit from PBI's direct Lucene writes and translog-free recovery.

### Additional context

This RFC was motivated by a discussion with @msfroh about eliminating the Kafka intermediate layer for Hive-to-OpenSearch ingestion.

lawofcycles:
### Description

Implements the `ingestion-hive` plugin proposed in #21451. OpenSearch can now ingest directly from Hive tables via the Pull-Based Ingestion framework, eliminating the need for an intermediate streaming layer like Kafka.

Hive tables store data as files (typically Parquet) in partitioned directories on storage systems like HDFS or S3. A separate service called Hive Metastore tracks which partitions and files belong to each table, and exposes this metadata via a Thrift API.

Each shard independently connects to Hive Metastore via Thrift, discovers partitions assigned to it via consistent hashing, reads Parquet files through Hadoop's `FileSystem` API, and emits rows into PBI's processing pipeline. Partition discovery is incremental and only fetches partitions added after the shard's watermark.

The plugin consists of four main components.
- **`HiveShardConsumer`**: Implements `IngestionShardConsumer`. Orchestrates partition discovery, file listing, and row reading.
- **`ThriftMetastoreCatalog`**: Thrift client for Hive Metastore. Handles partition listing with filtering. Supports both unauthenticated and Kerberos (SASL/GSSAPI) authentication.
- **`ParquetHiveFileReader`**: Reads Parquet files via `parquet-hadoop`, with schema projection from Metastore table definition.
- **`S3HadoopFileSystem`**: Minimal read-only Hadoop `FileSystem` for S3, registered for s3/s3a/s3n schemes.

#### Example usage

```json
PUT /events-index
{
  "settings": {
    "number_of_shards": 3,
    "number_of_replicas": 0,
    "replication.type": "SEGMENT",
    "ingestion_source": {
      "type": "HIVE",
      "pointer.init.reset": "earliest",
      "param.metastore_uri": "thrift://metastore:9083",
      "param.database": "analytics",
      "param.table": "events",
      "param.monitor_interval": "300s",
      "param.partition_order": "partition-name"
    }
  }
}
```

See #21451 for the full parameter list. Key parameters include `partition_order` (`partition-name`, `partition-time`, `create-time`), `authentication` (`none`, `kerberos`), and `hadoop_config.*` for pass-through Hadoop configuration (e.g., custom S3 endpoint).

#### Key points
- **Coordinator-less**: Each shard self-assigns partitions via `hash(partitionName) % numShards`, following PBI's shared-nothing model.
- **S3 via `S3HadoopFileSystem`**: `hadoop-aws` requires `software.amazon.awssdk:bundle` (~300MB). Instead, a minimal `FileSystem` backed by AWS SDK v2 individual modules is registered so `FileSystem.get()` dispatches storage transparently without per-format branching.
- **Thrift-generated Metastore client**: Hive Metastore exposes a Thrift API for table and partition metadata. The plugin uses a client generated from Hive's [`hive_metastore.thrift`](https://github.com/apache/hive/blob/master/standalone-metastore/metastore-common/src/main/thrift/hive_metastore.thrift) IDL, keeping the dependency footprint minimal. Kerberos (SASL/GSSAPI) authentication is included since most production Metastore deployments require it.
- **Pointer-based `_id`**: Each row's `_id` is set to its serialized pointer (partition + file + row index), making crash recovery idempotent without requiring user-side `_id` configuration.
- **Framework change**: Added `IngestionConsumerFactory.initialize(source, numberOfShards)` so plugins can receive shard count for partition assignment without relying on params injection.

### Related Issues
Resolves #21451

### Check List


plugins/ingestion-hive/bin/generate-thrift.sh
```diff
@@ -0,0 +1,57 @@
...
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+'
+
+echo "Generating Thrift code from ${THRIFT_IDL}..."
+
+rm -rf "${OUTPUT_DIR}"
```
varunbharadwaj, line 37: I noticed above we cd into the required directory and hardcode the OUTPUT_DIR, but is this always safe for forced deletion?

plugins/ingestion-hive/README.md
```diff
@@ -0,0 +1,117 @@
...
+        "init": {
+          "reset": "earliest"
+        }
+      },
+      "param": {
+        "metastore_uri": "thrift://hive-metastore:9083",
+        "database": "my_database",
+        "table": "my_table"
+      },
+      "mapper_type": "field_mapping"
```
varunbharadwaj: Should this be left to use the default mapper? It looks like the json we generate is `{_id: xyz, _source: ...}` format which is the default mapper format. Using field mapper might disregard the _id fields set from the Hive consumer flow. Let us double check.

plugins/ingestion-hive/README.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+## Configuration Parameters
+
+| Parameter | Required | Default | Description |
+|-----------|----------|---------|-------------|
+| `metastore_uri` | Yes | - | Thrift URI of the Hive Metastore (e.g., `thrift://host:9083`) |
+| `database` | Yes | - | Hive database name |
+| `table` | Yes | - | Hive table name |
+| `monitor_interval` | No | `60s` | How often to check for new partitions |
```
varunbharadwaj: Is the default 300s in the code? Can we double check the defaults for partition_order and others as well?

plugins/ingestion-hive/README.md
```diff
@@ -0,0 +1,117 @@
...
+| `partition_order` | No | `create-time` | Partition ordering (`create-time` or `name`) |
```
varunbharadwaj: I think the values are create-time, partition-time etc? Please check all the entries in the table in case we updated anything later on.

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -95,6 +95,8 @@ private void initializeStreamPoller(
         assert indexMetadata != null;
         IngestionSource ingestionSource = Objects.requireNonNull(indexMetadata.getIngestionSource());
 
+        ingestionSource.params().put("_number_of_shards", indexMetadata.getNumberOfShards());
```
varunbharadwaj: ingestionSource.params is meant for user provided consumer configs. Can we maybe add a new map to store additional metadata in IngestionSource where we can record the number of shards? This can be extended to capture other information in future when needed. We can highlight in comments that metadata map is only updated during engine initialization.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,680 @@
...
+    @Override
+    public List<ReadResult<HivePointer, HiveMessage>> readNext(
+        HivePointer pointer,
+        boolean includeStart,
+        long maxMessages,
+        int timeoutMillis
+    ) {
+        return executeWithRetry(() -> {
+            ensureInitialized();
+            if (!resumed && pointer != null && !pointer.getPartitionName().isEmpty()) {
```
varunbharadwaj: This method can be called the very first time a shard is initialized with the initial start pointer, or later if any message processing fails mid-batch, so we retry from the failed message. I'm guessing the retry on failed message flow will skip the seek as resumed = true?  Should we always read from provided pointer (includeStart decides inclusive/exclusive) ?

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,680 @@
...
+
+    @Override
+    public List<ReadResult<HivePointer, HiveMessage>> readNext(long maxMessages, int timeoutMillis) {
+        return executeWithRetry(() -> doReadNext(maxMessages));
+    }
+
+    private List<ReadResult<HivePointer, HiveMessage>> executeWithRetry(ReadAction action) {
+        try {
+            return action.execute();
+        } catch (TTransportException e) {
```
varunbharadwaj: Should IOException also be considered here?

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,680 @@
...
+            logger.warn("Metastore connection lost for shard {}, attempting reconnect: {}", shardId, e.getMessage());
+            try {
+                reconnectMetastore();
```
varunbharadwaj: if reconnect attempts are failing, and we are returning empty list as the result, will it result in too many reconnection attempts? Do we need a sleep here maybe if reconnect attempt fails?

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,680 @@
...
+                break;
+            default:
+                partitions = discoverByPartitionName();
+                break;
+        }
+
+        logger.info("Shard {} discovered {} candidate partitions from Metastore", shardId, partitions.size());
+
+        // Sort partitions according to ordering strategy
+        switch (config.getPartitionOrder()) {
```
varunbharadwaj: Looks like both the switch statements are on partition order? Do we want to keep it like this or combine them?

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,680 @@
...
+    @Override
+    public IngestionShardPointer earliestPointer() {
+        return new HivePointer("", "", 0, 0);
+    }
+
+    @Override
+    public IngestionShardPointer latestPointer() {
+        // Signals that existing partitions should be skipped. seekToPointer sets all watermarks
+        // to their respective maximums so that subsequent discovery returns no existing partitions.
+        return new HivePointer("__LATEST__", "", 0, Long.MAX_VALUE - 1);
```
varunbharadwaj: Should `__LATEST__` be an enum or a static final string constant defined for the class?

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,680 @@
...
+
+        // Assign partitions to this shard using consistent hashing on partition name.
+        // Each shard deterministically owns a subset of partitions (no coordination needed).
+        for (MetastoreCatalog.PartitionInfo partition : partitions) {
+            String partName = partitionToName(partition);
+            if (Math.floorMod(partName.hashCode(), numShards) == shardId) {
+                List<String> files = listDataFiles(partition.getLocation());
+                if (!files.isEmpty()) {
+                    String partTime = extractPartitionTime(partition);
+                    pendingWork.add(new PartitionWork(partName, partTime, files, partition.getCreateTime()));
```
varunbharadwaj, line 411: Where do we remove items from `pendingWork` once complete? 🤔 
Wondering if currently this list grows forever?

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/ThriftMetastoreCatalog.java
```diff
@@ -0,0 +1,236 @@
...
+                    t = new TFramedTransport(socket);
+                } else {
+                    t = socket;
+                }
+
+                if (config.getAuthMode() == HiveSourceConfig.AuthMode.KERBEROS) {
+                    t = createSaslTransport(t, host);
+                    final TTransport saslTransport = t;
+                    try {
+                        UserGroupInformation.getLoginUser().doAs((PrivilegedExceptionAction<Void>) () -> {
```
varunbharadwaj: Just wondering - does this work if we have multiple hive ingestion based indices created?

plugins/ingestion-hive/src/internalClusterTest/java/org/opensearch/plugin/hive/IngestFromHiveIT.java
```diff
@@ -0,0 +1,620 @@
...
+        });
+
+        // Pause ingestion
+        client().admin().indices().pauseIngestion(Requests.pauseIngestionRequest(indexName)).get();
+
+        // Add a new partition while paused
+        addPartition("2026-04-16", 100);
+
+        // Resume ingestion
+        client().admin().indices().resumeIngestion(Requests.resumeIngestionRequest(indexName)).get();
```
varunbharadwaj, line 185: Before we resume should we be validating if the pause has taken effect, or if the new partition is not visible?

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,680 @@
...
+        String result = pattern;
+        for (int i = 0; i < partitionKeys.size(); i++) {
+            result = result.replace("$" + partitionKeys.get(i), values.get(i));
+        }
+        return result;
+    }
+
+    byte[] rowToJson(Map<String, Object> row, HivePointer pointer) {
+        StringBuilder sb = new StringBuilder("{\"_id\":\"");
+        sb.append(escapeJson(pointer.asString()));
```
msfroh: I imagine this is probably fine for the typical Hive use-case which is probably going to be append-only.

Does it make sense to support the presence of an ID column in the Hive table? Are we ever going to need to support updates to existing rows? I'm guessing not, but I don't know enough about Hive to know for sure.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,680 @@
...
+                sb.append("\"").append(escapeJson(value.toString())).append("\"");
+            }
+        }
+        sb.append("}}");
+        return sb.toString().getBytes(StandardCharsets.UTF_8);
+    }
+
+    private static String escapeJson(String value) {
+        return value.replace("\\", "\\\\").replace("\"", "\\\"").replace("\n", "\\n").replace("\r", "\\r").replace("\t", "\\t");
+    }
```
msfroh: Instead of escaping JSON, I'm wondering if it might be easier (or at least more robust) to use the existing `XContentBuilder` to construct the JSON content.

So, your `rowToJson` could just do something like:

```
XContentBuilder builder = JsonXContent.contentBuilder();
builder.startObject();
builder.field("_id", pointer.asString());
builder.startObject("_source");
builder.mapContents(row);
builder.endObject(); // _source
builder.endObject(); // root object

return BytesReference.toBytes(BytesReference.bytes(builder));
```

plugins/ingestion-hive/README.md
lawofcycles: Thanks, the README was stale after several refactoring iterations. Will update the defaults, parameter list, and usage examples to match the current code.

plugins/ingestion-hive/README.md
lawofcycles: Confirmed. The default in code is `300s`, not `60s`. Fixed in the README update.

plugins/ingestion-hive/README.md
lawofcycles: Confirmed. The valid values are `partition-name` (default), `create-time`, and `partition-time`. Fixed in the README update.

plugins/ingestion-hive/README.md
lawofcycles: You're right. The plugin produces `{"_id": "...", "_source": {...}}` format, which is what the default mapper expects. Removed `mapper_type` from the examples.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Extracted to `private static final String LATEST_PARTITION_SENTINEL`.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Fixed. `discoverNewPartitions()` now resets `pendingWork` and `currentWorkIndex` at the start of each discovery cycle. This method is only called after `shouldRefreshPartitions()` confirms all prior work is consumed, so no in-progress entries are dropped. Added a unit test to verify the reset behavior.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Combined into a single switch. Each case now handles both fetch and sort together.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Reworked. ThriftMetastoreCatalog wraps all TExceptions in IOException, so the previous TTransportException catch was unreachable. Now catches IOException and recovers both the file reader and Metastore connection before retrying. This handles both Metastore disconnects and transient file read errors uniformly.

plugins/ingestion-hive/bin/generate-thrift.sh
lawofcycles: Added a guard that verifies OUTPUT_DIR is within src/main/java/ before the rm -rf. The script also has set -eu so a failed cd would abort early, but this makes the intent explicit.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Added exponential backoff on repeated failures, resets on success.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Good call. Replaced with XContentBuilder. The manual escapeJson was only covering 5 characters, so this is strictly better.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Good catch. Confirmed that DefaultStreamPoller sets forcedShardPointer from processRecords on batch failure, which calls readNext(pointer, ...) again. The `resumed` flag was preventing seek on that second call, causing data loss. Removed it. The guard was unnecessary because the pointer-less readNext(maxMessages, timeout) is a separate method that never hits the seek path.

plugins/ingestion-hive/src/internalClusterTest/java/org/opensearch/plugin/hive/IngestFromHiveIT.java
lawofcycles: Added. The test now asserts poller is actually in PAUSED state via getIngestionState API before adding the new partition, then verifies document count remains at 450.

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
lawofcycles: Agree, `params()` should be reserved for user-provided consumer configs. Added a `metadata()` map to `IngestionSource` for framework-injected runtime information, as you suggested. `params()` stays for user configs, `metadata()` is populated by IngestionEngine during poller initialization. This also fixed an `UnsupportedOperationException` that was occurring because `params()` became unmodifiable upstream.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/ThriftMetastoreCatalog.java
lawofcycles: Good point. The previous global `loginUserFromKeytab` would break with multiple indices since the second login overwrites the first. Followed the `repository-hdfs` pattern and switched to `loginUserFromKeytabAndReturnUGI` so each `ThriftMetastoreCatalog` holds its own UGI instance. Added `testMultiplePrincipalsCoexist` IT that creates two indices with `client1` and `client2` principals simultaneously to verify they work independently.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Hive tables are designed around partition-level immutability. The most common ingestion pattern is append-only (new partitions are added over time), but partition-level overwrites (INSERT OVERWRITE) do happen when Spark or similar engines rewrite data for corrections or compaction. Row-level updates and deletes require rewriting entire partitions or tables, which is prohibitively expensive in both compute resources and processing time. This is one of the core motivations behind formats like Iceberg and Hudi.

On the current implementation side, the watermark-based discovery assumes partitions are immutable once processed, and the consistent hashing assigns partitions to shards with shard-local writes. This means partition overwrites or partition column changes would not be handled correctly today.

Given these constraints, I think keeping the Hive source as append-only for now makes sense, and revisiting update/delete support when we add Iceberg or Hudi sources (where CDC is a first-class concept) would be a more natural fit. That said, it depends on your organization's and the community's actual needs, so it would be good to validate whether append-only covers the intended use cases.

Regarding an ID column in Hive tables, it's not impossible but uncommon in practice. Hive's ecosystem doesn't generally have mechanisms that operate on a row identity column, and partitions are typically treated as immutable units.

lawofcycles:
@varunbharadwaj @msfroh Thanks for the thorough review! Addressed all comments and merged latest main. The feedback really helped improve the robustness of the plugin, especially around the Kerberos isolation and error handling. Ready for re-review when you get a chance.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,673 @@
...
+                return results;
+            } catch (Exception retryEx) {
+                logger.error("Recovery failed for shard {}: {}", shardId, retryEx.getMessage());
+                backoffOnFailure();
+                return Collections.emptyList();
+            }
+        } catch (Throwable e) {
+            logger.error("Error reading from Hive table for shard {}: {}", shardId, e.getMessage());
+            backoffOnFailure();
+            return Collections.emptyList();
```
varunbharadwaj: Looks like we return empty list on read errors. This should be fine, but i think we will skip incrementing the consumer error count metric (done in the DefaultStreamPoller). If we have a bad schema/mappings or continuously failing for any other reason, what should be the expectation? Should we throw an exception at some point or do you feel this is okay? 🤔

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,673 @@
...
+     * Opens the next data file from the pending work queue. When all files in a partition
+     * are consumed, updates the watermark and advances to the next partition.
+     */
+    private boolean openNextFile() throws IOException {
+        while (currentWorkIndex < pendingWork.size()) {
+            PartitionWork work = pendingWork.get(currentWorkIndex);
+            if (work.currentFileIndex < work.files.size()) {
+                // More files in current partition: open the next one
+                currentFile = work.files.get(work.currentFileIndex);
+                work.currentFileIndex++;
```
varunbharadwaj: Should currentFileIndex be incremented after the file is opened successfully? Also, if there is an error later on while reading, do we end up moving to the next file skipping the current one?

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -95,6 +95,8 @@ private void initializeStreamPoller(
         assert indexMetadata != null;
         IngestionSource ingestionSource = Objects.requireNonNull(indexMetadata.getIngestionSource());
 
+        ingestionSource.metadata().put("_number_of_shards", indexMetadata.getNumberOfShards());
```
varunbharadwaj: There was a recent change (https://github.com/opensearch-project/OpenSearch/pull/21841) that gives the IngestionConsumerFactory access to IndexMetadata via method `createShardConsumer(String clientId, int shardId, IndexMetadata indexMetadata)`. This can help us access the number of shards in the hive plugin without making any changes to the IngestionEngine and IngestionSource. Can you please have a look.

lawofcycles: Done. Rebased on latest main and switched to createShardConsumer(clientId, shardId, IndexMetadata).

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Fixed. On IOException, the consumer retries once after reconnecting to Metastore (which itself retries up to `max_retries` times internally). If recovery fails, the exception propagates to DefaultStreamPoller, which increments the error count and pauses ingestion. Non IO errors propagate immediately without retry. This aligns with how the Kafka and Kinesis consumers handle errors.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Fixed. Moved the increment to after createFileReader() succeeds, so a failed open does not advance past the file. Added a unit test that verifies currentFileIndex stays unchanged when file open throws.

plugins/ingestion-hive/src/internalClusterTest/java/org/opensearch/plugin/hive/IngestFromHiveIT.java
```diff
@@ -0,0 +1,639 @@
...
+        client().admin().indices().pauseIngestion(Requests.pauseIngestionRequest(indexName)).get();
+
+        // Verify pause has taken effect
+        assertBusy(() -> {
+            GetIngestionStateResponse state = client().admin()
+                .indices()
+                .getIngestionState(Requests.getIngestionStateRequest(indexName))
+                .get();
+            for (ShardIngestionState shardState : state.getShardStates()) {
+                assertTrue("Poller should be paused", shardState.isPollerPaused());
```
varunbharadwaj: We can refer to the kafka tests [here](https://github.com/opensearch-project/OpenSearch/blob/main/plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/KafkaIngestionBaseIT.java#L245) for reference. We need to wait for the shardState.isPollerPaused() and also poller state to be `paused`, to ensure there is no race condition that can cause flakiness in the test.

plugins/ingestion-hive/licenses/third-party-jackson-core-NOTICE.txt
```diff
@@ -0,0 +1,14 @@
+OpenSearch (https://opensearch.org/)
```
varunbharadwaj: Is the contents of this file correct?

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,654 @@
...
+    @FunctionalInterface
+    private interface ReadAction {
+        List<ReadResult<HivePointer, HiveMessage>> execute() throws Exception;
+    }
+
+    /**
+     * Core read loop. Discovers new partitions if needed, then reads rows from the current
+     * file and converts them to JSON messages. Returns up to maxMessages results per call.
+     */
+    private List<ReadResult<HivePointer, HiveMessage>> doReadNext(long maxMessages) throws Exception {
```
varunbharadwaj, line 303: If there is a transient read failure, have we checked if we will resume from the same file when openNextFile() is called or if we are skipping rest of the file? We could add a integration test / unit test for this.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
```diff
@@ -0,0 +1,654 @@
...
+    }
+
+    @Override
+    public int getShardId() {
+        return shardId;
+    }
+
+    @Override
+    public long getPointerBasedLag(IngestionShardPointer expectedStartPointer) {
+        if (pendingWork == null) return 0;
```
varunbharadwaj: I think we can return -1 here when pendingWork is null / not initialized. 0 indicates that we have caught up and there is no lag, which can then skip warmup feature, if it is configured.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Good catch, this was a bug. A mid-file failure discarded the rows already read in that batch and then skipped the rest of the file on retry, because openNextFile had already advanced past it.

The consumer now returns the rows read so far as a partial batch, re-queues the file, and skips the delivered rows when reopening, so nothing is lost or duplicated. Added testTransientReadFailureResumesSameFileWithoutLossOrDuplicates covering exactly this scenario.

plugins/ingestion-hive/src/main/java/org/opensearch/plugin/hive/HiveShardConsumer.java
lawofcycles: Fixed. It now returns -1 both for the null guard and before the first partition discovery has run, since the lag is simply unknown at that point.

plugins/ingestion-hive/src/internalClusterTest/java/org/opensearch/plugin/hive/IngestFromHiveIT.java
lawofcycles: Fixed. The test now follows the Kafka test pattern and waits for both isPollerPaused() and the poller state to become PAUSED (with no failed shards) before asserting that the document count stays unchanged.

lawofcycles:
@varunbharadwaj  Ready for another review. Since the last pass I addressed the remaining comments, did a final self review that fixed a few issues, and merged the latest main to resolve the buildSrc conflict. Could you take another look?

plugins/ingestion-hive/licenses/libthrift-LICENSE.txt
```diff
@@ -0,0 +1,6 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+   (full text at http://www.apache.org/licenses/LICENSE-2.0)
```
varunbharadwaj: Looks like this got truncated? We can copy the full content here. The NOTICE file also seems to be empty. We can copy the contents for it as well.

varunbharadwaj:
Thanks for the changes @lawofcycles. It looks good to me overall from the pull-based ingestion integration perspective, and we can iterate further in subsequent PRs. I noticed some of the license files and notice files (will share the list offline) are not complete. Could you please review them once again and update where ever needed?

@msfroh could you also take a look at the PR again?

msfroh (approved):
It looks good to me, overall. As @varunbharadwaj said, we can iterate some more as we try it out.

IMO, this is really exciting!

lawofcycles:
@msfroh @varunbharadwaj Thank you for the thorough review! I fixed the issue with the license file and merged the latest main into my branch.

I plan to work on the following tasks next.

* Blog post: Writing an OpenSearch blog post to introduce this new feature, including some performance benchmarks.
* Documentation: Adding a documentation page for this feature.
* Enhancements: Working on several improvements such as enhanced shard mapping.
* Open Table Format support: Considering how to integrate with Iceberg.

Beyond this PR, I'm also keeping a close eye on the Iceberg integration in the Pluggable Component Bundling discussion (https://github.com/opensearch-project/OpenSearch/issues/20644#issuecomment-4885896658). It would enable direct querying from OpenSearch to Iceberg via DataFusion. I think the choice between PBI and direct query depends on the use case. For example, if users need lexical search backed by Lucene, PBI is the right fit, whereas if they need a seamless query experience, direct query would be better. If there is room for contribution, I would like to help on the direct query side as well.

FYI, I'm also trying to write up an RFC for Lance table format integration, which would be another form of direct query from OpenSearch to Lance.

varunbharadwaj (approved):

PR #21701, opened 2026-05-17
Performance Delegation to Lucene from Datafusion for Filter Predicates

expani:
## Overview

  Predicates that both DataFusion and Lucene can evaluate (e.g. tag = 'hello' on indexed keyword fields) now run faster. The engine consults Lucene per row group when
  DataFusion's own page-level pruning isn't selective enough, narrowing the candidate set before parquet decode.

## Followups

  - Selectivity gate is currently a 5% threshold; cluster setting plumbing is on the roadmap. Also, include considering `Weight.count` as a gate.
  - OR/NOT-shaped queries with dual-viable predicates fall back to native DataFusion evaluation cleanly. Will be a follow up for NOT condition.  
  - Combining multiple predicates belonging to the same backend into one `AND` condition is a follow up so that BooleanQuery is generated in Lucene. Currently, one eligible predicate is used for perf delegation.
  - Adding delegation based on pureFilterColumns and known slow cases like Wildcard on high cardinality will be follow ups.
  - Dynamic Mapping for Filter Delegation requires more work to return a signal that says query doesn't match anything V/S IT can't match because field is absent.
  - Validating that if a backend is marked for delegation it also has a serializer for that predicate. Currently, it silently skips as the UT surface area was huge.
  - Asserting in Integs whether Lucene was actually called (vs DataFusion handling everything natively) is only visible in cluster logs today. Code-level assertion needs a mock test plugin and is deferred.

##  Tests

  - New IT exercising dual-viable EQUALS with aggregation and proving the disjunctive shape doesn't crash the data node.
  - Existing MATCH delegation IT still passes.
  - Rust unit tests + Java unit tests green.

## Additional Fixes

- Caught an existing `UTF8View == UTF8` bug where we were not converting physical expressions properly.
Queries like below will no longer fail with Filter Delegation
```
where match(message, 'hello') and tag = 'hello' | stats sum(value) as total
```
-  Fixed using `index.composite.secondary_data_formats` for determining Lucene as a backend.


sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_table/eval/single_collector.rs
```diff
@@ -115,10 +147,39 @@ impl SingleCollectorEvaluator {
...
+) -> bool {
+    let surviving_rows = match page_ranges {
+        None => rg.num_rows as i64,
+        Some(ranges) => ranges.iter().map(|(lo, hi)| (hi - lo) as i64).sum::<i64>(),
+    };
+    if rg.num_rows == 0 {
+        return false;
+    }
+    let surviving_fraction = surviving_rows as f64 / rg.num_rows as f64;
+    surviving_fraction > threshold
```
bharath-techie, line 223: Should this also consider weight.count equivalent?

What if surviving fraction is still less than weight.count equivalent? Atleast we must do this for o(1) queries

expani: Yes we should consider it and other factors like presence of pureFilter columns and some predicate functions like Wildcard that are known to be slower. 

This will be a significant change to include so I am planning to order them up in the following way : 

- Combine AND predicates to ensure BooleanQuery is formed in Lucene ( Significant as it touches Fragment Conversion a lot ) 
- Introduce segment pruning at analytics-engine instead of DF backend. This would introduce using PruningPredicate at the same level as Weight.count. Will be very involved with FragmentConversion.
- Enhance the FilterDelegationHandle to perform smart performance based delegation ( This should resolve your comment ) 
- Introduce a mechanism to filter out Wildcard queries and other cases that are known to be slower at a predicate filter level via Cluster/Index settings and then execution hints.

bharath-techie (approved):
LGTM

aasom143:
Hi @expani , After picking up your changes, following query seems to be stuck, can you please verify it this regression introduced with your change?
```
curl -XPOST 'http://localhost:9200/_plugins/_ppl' -H 'Content-Type: application/json' -d '{"query": "source=clickbench | where match(Title, '\''google'\'') AND BrowserCountry = '\''B8'\'' | stats count()"}'
```
Also following query giving the wrong result if this query also supposed to be covered with your PR, raising this point because before your PR, it was failing with `Sort/dc streaming fragment failure`, so can you please check once. Currently with your change, it's giving `17630976`, but the correct result is `2523`. 
```
curl -XPOST 'http://localhost:9200/_plugins/_ppl' -H 'Content-Type: application/json' -d '{"query": "source=clickbench | where match(Title, '\''google'\'') AND RegionID = 229 AND Age > 0 | stats dc(UserID)"}'
```

alchemist51:
```
  So the flow is:
  1. Java sees AND with performance-delegated peers → tree_shape=CONJUNCTIVE (=1)
  2. Rust gets (tree_shape=1, delegated_predicate_count=3) → (1, _) matches → FilterClass::Tree
  3. Tree path → BitmapTreeEvaluator runs → panics on DelegationPossible node
  ```

alchemist51:
As part of the above commits, the following changes are covered:

1. Lucene field-type whitelisting need to only allow what is being ingested in the pluggable format. Have added the same. Have also taken a todo to link it with LuceneFieldRegistry. It was making the numeric/data/boolean fields lucene viable resulting in wrong results or crashes.
2. NOT was being classifield as CONJUNCTIVE(SingleCollector) while it actually needs a tree shape to solve.
3. In Conjunctive shape, if we had multiple delegation, it was going into the Tree shape, now multiple delegations also goes via the SingleCollector path.
4. In case of dc queries, the planner was adding a NOT NULL as a top level filter. Currently findNode function only checks for the first filterNode and tells the shape, it should ideally recursively check the tree. Have added a rule for Merging the filters while taken a todo to recursively check if we find more such cases.

@expani  @mch2  please let me know if you folks see concerns.

sandeshkr419 (approved):

PR #21703, opened 2026-05-17
Unified native allocator framework to track and monitor Arrow allocations

Bukhtawar:
Introduces a new plugin (`native-allocator-arrow`) that owns a single Arrow RootAllocator for the node, with named pool-level children for each subsystem (flight, query, datafusion, ingest). A background rebalancer redistributes unused capacity across pools every 5 seconds so active pools can burst beyond their guarantee when others are idle.

The SPI (`libs/arrow-spi`) defines the Arrow-agnostic interface, pool config constants, and Writeable stats shape. The plugin implements it with Arrow's BufferAllocator and registers configurable limits as dynamic cluster settings.

This is a framework-only PR — no existing code is modified. Follow-up PRs will wire consumers (arrow-flight-rpc, analytics-engine, analytics-backend-datafusion, parquet-data-format) to use the unified allocator and expose stats via _nodes/stats.

```
libs/arrow-spi/  (depends on libs:core only — NO Arrow)
  ├── NativeMemoryAllocator.java      ← interface: getOrCreatePool, setPoolLimit, stats()
  ├── NativeMemoryPoolConfig.java     ← pool names, setting keys, static defaults
  └── NativeMemoryPoolStats.java      ← Writeable + ToXContent stats blob

  plugins/native-allocator-arrow/  (depends on libs:arrow-spi + arrow-memory-core)
  ├── ArrowNativeMemoryAllocator.java ← implements NativeMemoryAllocator using RootAllocator
  └── NativeAllocatorArrowPlugin.java ← registers settings from yaml, creates pools on startup
                                         exposes allocator as a component

  plugins/arrow-flight-rpc/        ← extendedPlugins = ['native-allocator-arrow']
  plugins/analytics-engine/        ← extendedPlugins = ['native-allocator-arrow']
  plugins/analytics-backend-datafusion/ ← extendedPlugins = ['native-allocator-arrow']
  plugins/parquet-data-format/     ← extendedPlugins = ['native-allocator-arrow']
```

  The indexing path now flows through, other flows need to wired similarly

```
  NativeAllocatorArrowPlugin (root)
    └── "ingest" pool (limit from yaml)
          └── ArrowBufferPool.createChildAllocator("vsr-N")
                └── ManagedVSR (per-batch Arrow buffers)

```

Needs to be integrated with https://github.com/opensearch-project/OpenSearch/pull/21465

### Description

### Related Issues

### Check List


plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/ArrowNativeAllocator.java
```diff
@@ -0,0 +1,291 @@
...
+     *   <li>Identify active pools (currently allocated > 50% of guarantee)</li>
+     *   <li>Distribute headroom proportionally among active pools</li>
+     *   <li>Set each pool's effective limit = guarantee + its share of headroom</li>
+     *   <li>Inactive pools are capped at their guarantee</li>
+     * </ol>
+     *
+     * <p>A pool's limit never drops below its current allocation (to avoid
+     * failing in-progress operations) and never below its guaranteed limit.
+     */
+    void rebalance() {
```
alchemist51, line 199: Do we want to abstract it inside a strategy? To easily switch between greedy/fair distribution type of strategy later on?

plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/NativeAllocatorArrowPlugin.java
```diff
@@ -0,0 +1,203 @@
...
+        Long.MAX_VALUE,
+        0L,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    /** Interval in seconds between pool rebalance cycles. */
+    public static final Setting<Long> REBALANCE_INTERVAL_SETTING = Setting.longSetting(
+        "native.allocator.rebalance.interval_seconds",
+        0L,
```
alchemist51, line 101: let's make it something like 1m?

libs/arrow-spi/src/main/java/org/opensearch/arrow/spi/NativeAllocatorPoolConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Returns the setting key for the given pool's limit.
+     *
+     * @param poolName the pool name
+     */
+    public static String settingKey(String poolName) {
+        return switch (poolName) {
+            case POOL_FLIGHT -> SETTING_FLIGHT_LIMIT;
+            case POOL_QUERY -> SETTING_QUERY_LIMIT;
```
bharath-techie: Is there a diff between this and DF pool?

alchemist51: Is this per query limit?

plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/NativeAllocatorArrowPlugin.java
bharath-techie: 0 is disabled

bharath-techie: Maybe that should be added as doc here

plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/NativeAllocatorArrowPlugin.java
```diff
@@ -0,0 +1,203 @@
...
+            ROOT_LIMIT_SETTING,
+            FLIGHT_LIMIT_SETTING,
+            QUERY_LIMIT_SETTING,
+            DATAFUSION_LIMIT_SETTING,
+            INGEST_LIMIT_SETTING,
+            REBALANCE_INTERVAL_SETTING
+        );
+    }
+
+    private static void validatePoolLimits(long rootLimit, long... poolLimits) {
```
bharath-techie: Do we need to check the current usage.? What if current usage is above the limit ?

Also do we need to check if sum exactly matches root allocator?

libs/arrow-spi/src/main/java/org/opensearch/arrow/spi/NativeAllocatorPoolConfig.java
Bukhtawar: coordinator allocates the pool for handling responses from data nodes

plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/NativeAllocatorArrowPlugin.java
Bukhtawar: This is handler through ` effectiveLimit = Math.max(effectiveLimit, currentAllocation)`

plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/ArrowNativeAllocator.java
```diff
@@ -0,0 +1,335 @@
...
+
+            // Cap at pool's max
+            effectiveLimit = Math.min(effectiveLimit, max);
+            // Never drop below current allocation or min
+            effectiveLimit = Math.max(effectiveLimit, currentAllocation);
+            effectiveLimit = Math.max(effectiveLimit, min);
+            // Never exceed root
+            effectiveLimit = Math.min(effectiveLimit, rootLimit);
+
+            alloc.setLimit(effectiveLimit);
```
mch2, line 241: is this safe without re-creating the child allocator?  how do we handle this with concurrent bursts on in-flight allocations/requests? Meaning how does arrow handle if we call setLimit and we're already over the new max, those allocs/reqs fail? or rebalance fails?

libs/arrow-spi/src/main/java/org/opensearch/arrow/spi/NativeAllocatorPoolConfig.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /** Setting key for the root allocator limit. */
+    public static final String SETTING_ROOT_LIMIT = "native.allocator.root.limit";
+
+    /** Setting key for the Flight pool minimum. */
+    public static final String SETTING_FLIGHT_MIN = "native.allocator.pool.flight.min";
+    /** Setting key for the Flight pool maximum. */
+    public static final String SETTING_FLIGHT_MAX = "native.allocator.pool.flight.max";
+    /** Setting key for the ingest pool minimum. */
+    public static final String SETTING_INGEST_MIN = "native.allocator.pool.ingest.min";
```
mch2, line 42: not for a first version here, but thinking the plugins that use/provide these functionalities should register / define these settings.

plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/ArrowNativeAllocator.java
Bukhtawar: I think Arrow's BufferAllocator.setLimit() is non-disruptive, i.e. it only affects future allocations

Bukhtawar:
```
REPRODUCE WITH: ./gradlew ':server:internalClusterTest' --tests 'org.opensearch.recovery.RecoveryWhileUnderLoadIT' -Dtests.method='testRecoverWhileUnderLoadWithReducedAllowedNodes {p0={"cluster.indices.replication.strategy":"SEGMENT"}}' -Dtests.seed=59F5E5CA00DB5BC2 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=ar-SS -Dtests.timezone=America/Guayaquil -Druntime.java=25

RecoveryWhileUnderLoadIT > testRecoverWhileUnderLoadWithReducedAllowedNodes > testRecoverWhileUnderLoadWithReducedAllowedNodes {p0={"cluster.indices.replication.strategy":"SEGMENT"}} FAILED
    java.lang.AssertionError: replica shards haven't caught up with primary expected:<29> but was:<26>
        at __randomizedtesting.SeedInfo.seed([59F5E5CA00DB5BC2:3D6CEB1AC8888F91]:0)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at org.opensearch.test.OpenSearchIntegTestCase.lambda$waitForReplication$1(OpenSearchIntegTestCase.java:2570)
        at org.opensearch.test.OpenSearchTestCase.assertBusy(OpenSearchTestCase.java:1193)
        at org.opensearch.test.OpenSearchIntegTestCase.waitForReplication(OpenSearchIntegTestCase.java:2557)
        at org.opensearch.recovery.RecoveryWhileUnderLoadIT.assertAfterRefreshAndWaitForReplication(RecoveryWhileUnderLoadIT.java:504)
        at org.opensearch.recovery.RecoveryWhileUnderLoadIT.testRecoverWhileUnderLoadWithReducedAllowedNodes(RecoveryWhileUnderLoadIT.java:350)
```

plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/ArrowNativeAllocator.java
Bukhtawar: Let me write a test for this case

plugins/native-allocator-arrow/src/main/java/org/opensearch/arrow/allocator/ArrowNativeAllocator.java
```diff
@@ -0,0 +1,335 @@
...
+     * <ol>
+     *   <li>Every pool is guaranteed at least its configured min</li>
+     *   <li>Compute headroom = rootLimit - sum(all pool current allocations)</li>
+     *   <li>Identify active pools (allocated > 0)</li>
+     *   <li>Distribute headroom equally among active pools, capped at each pool's max</li>
+     *   <li>Inactive pools are set to their min</li>
+     *   <li>No pool's limit ever drops below its current allocation or its min</li>
+     * </ol>
+     */
+    void rebalance() {
```
bowenlan-amzn, line 199: I'm not sure this min/max rebalance logic is useful. Maybe start with one limit per pool (dynamic cluster setting) — I feel that would suffice.

The use case I'm thinking about is when ingest and query compete for Arrow memory at the same time. Say total we have 8GB, and both at peak use 8GB, so we set their limit to 8GB. When both burst, the rebalancer can't help — headroom is zero, Math.max(effectiveLimit, currentAllocation) prevents shrinking, so there's nothing to redistribute.

What would actually help under contention:
- Query path: when allocation hits the limit, surface a circuit breaker exception so the client knows to retry
- Ingest path: integrate native memory into IndexingPressure so bulk requests get rejected with 429 before Arrow OOMs

I hope we can hook into existing circuit breaker framework. Like a "native_memory" child breaker backed by rootAllocator.getAllocatedMemory() — that makes Arrow bytes visible to the parent breaker, gives you _nodes/stats for free, and lets the parent trip when heap + native exceeds the node budget.

alchemist51: +1 to keep this simple for the first iteration and use the existing resiliency framework

Bukhtawar: This is just the framework PR folks, I'd like to keep it simple to allow other PRs to consume and improvise. The only useful circuit breaker is the real memory circuit breaker that looks into the point in time heap. It also triggers a GC if it goes beyond. The other circuit breakers aren't effective. Since each of the limits across indexing/search/flight is tunable, I'd want to expose individual child stats for visibility and better control(nothing comes for free :)) Overall aligned on rejecting requests and handling the failures gracefully.

plugins/arrow-base/src/main/java/org/opensearch/arrow/allocator/ArrowNativeAllocator.java
```diff
@@ -0,0 +1,335 @@
...
+ * Each pool has a <em>guaranteed</em> limit (configured via settings). When other
+ * pools are idle, an active pool can temporarily grow beyond its guarantee up to
+ * the root limit. When contention rises, pools shrink back toward their guarantee.
+ * This prevents idle capacity from being wasted while maintaining isolation under load.
+ *
+ * <p>The singleton instance is accessible via {@link #instance()} after the
+ * plugin creates it. This allows child plugins (in the same classloader) to
+ * locate the allocator without explicit injection.
+ */
+public class ArrowNativeAllocator implements NativeAllocator {
```
bowenlan-amzn, line 44: This can substitute https://github.com/opensearch-project/OpenSearch/blob/main/plugins/arrow-base/src/main/java/org/opensearch/arrow/memory/DefaultArrowAllocatorService.java

arrow-base pluging can contain this, so no need for a new native-allocator-arrow plugin.

Bukhtawar: I merged both the plugins now, earlier I assumed both could co-exist and subsequent PRs wire the newer SPIs and deprecate the arrow-base. Looks like we ran into conflicts. So had to merge the SPI plugins.

bowenlan-amzn: Thanks! Could you please keep the arrow-base? current code change seems remove arrow-base and keep native-allocator-arrow.

Bukhtawar: We can't have both(we will run into conflicts). I have merged it into the native-allocator-arrow spi

bowenlan-amzn (approved):

mch2 (approved):

PR #21704, opened 2026-05-17
[Block Cache] Cache registry disk-full metadata cleanup, add activeInBytes stat

vishwasgarg18:
### Description

This PR bundles three improvements to the Foyer-backed warm node block cache:

#### 1. Fix key_index metadata cleanup on disk-full

When the Foyer disk cache reaches capacity and eviction occurs, the in-memory
`key_index` could retain stale byte-accounting entries, causing incorrect
`used`/`active` byte counts. This change:

- Adds a `key_byte_size(raw_key)` helper in `range_cache.rs` that parses the
  byte size directly from the Foyer cache key string (`"path\x1Fstart-end"`)
  without requiring the value to be in memory — enabling correct cleanup of
  evicted/removed entries even under disk pressure.
- Refactors `foyer_cache.rs` to use `key_byte_size` in eviction, removal, and
  sweep callbacks so the key_index stays accurate when the disk is full.
- Introduces a configurable background sweep via a new
  `block_cache.foyer.key_index_sweep_interval_seconds` setting (default 30 s),
  passed through the `foyer_create_cache` FFM call. The sweep periodically
  reconciles the key_index against the live cache state.

#### 2. Remove `data_to_cache_ratio` plugin setting and add `activeInBytes` stat

- **Removes** `block_cache.foyer.data_to_cache_ratio`. The warm-to-SSD ratio
  is now read exclusively from the cluster-level `cluster.filecache.remote_data_ratio`
  setting, which is the single source of truth for SSD budget allocation.
- **Adds `activeInBytes`** to the block cache stats pipeline, tracking bytes
  currently actively referenced in the cache. The counter flows end-to-end:
  Rust atomic (`FoyerStatsCounter`) → FFM buffer (`FoyerAggregatedStats`,
  buffer 18→20 longs) → Java SPI (`BlockCacheStats.activeInBytes`) →
  `NodeCacheService.mergeStats()` folded into `AggregateFileCacheStats.active`.

#### 3. Rename `NodeCacheOrchestrator` → `NodeCacheService`

`NodeCacheService` better reflects the component's role as the single authority
for all warm-node SSD caches — owning budget allocation, lifecycle, stats
aggregation, and `BlockCacheProvider` registration. All references updated in
`Node.java`, `NodeService.java`, `IndicesService.java`, `IndexModule.java`,
`IndexService.java`, `WarmFsService.java`, `FsServiceProvider.java`, and all
associated test files. `NodeCacheOrchestratorCleaner` is renamed to
`NodeCacheServiceCleaner` accordingly.

### Related Issues
Resolves #[Issue number]

### Check List


sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -295,7 +202,99 @@ impl FoyerCache {
...
+            if sweep_interval_secs == 0 { "default" } else { "operator config" }
+        );
+        {
+            let sweep_inner = inner.clone();
+            let sweep_key_index = Arc::clone(&key_index);
+            let sweep_stats = Arc::clone(&stats);
+            let sweep_shutdown = Arc::clone(&shutdown);
+            let sweep_interval = Duration::from_secs(effective_sweep_secs);
+
+            rt.spawn(async move {
```
nishchay21, line 313: This spawned task has no shutdown mechanism. When FoyerCache is dropped, the task continues running indefinitely because the Arc<Runtime> keeps the runtime alive. Add a CancellationToken and cancel it in a Drop impl.

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -295,7 +202,99 @@ impl FoyerCache {
...
+                loop {
+                    tokio::time::sleep(sweep_interval).await;
+                    if sweep_shutdown.load(Ordering::Relaxed) {
+                        native_bridge_common::log_debug!("[block-cache] sweeper: shutdown signal received, exiting");
+                        break;
+                    }
+
+                    let mut stale_removed = 0usize;
+                    let mut freed_bytes = 0i64;
+                    for mut entry in sweep_key_index.iter_mut() {
```
nishchay21: If the sweep holds DashMap write locks during iteration (e.g., via retain()), it can block concurrent put() calls on the same shard. Consider a two-phase approach: read-only scan first, then targeted per-entry removal with brief write locks.

server/src/main/java/org/opensearch/index/store/remote/filecache/NodeCacheService.java
```diff
@@ -341,13 +339,14 @@ private BlockCacheStats combineBlockCacheStats() {
...
-            totalCapacity
+            totalCapacity,
+            activeInBytes
         );
     }
 
     private static AggregateFileCacheStats mergeStats(AggregateFileCacheStats fc, BlockCacheStats bc) {
         FileCacheStats mergedOverall = new FileCacheStats(
-            fc.getActive().getBytes(),
+            fc.getActive().getBytes() + bc.activeInBytes(),
```
nishchay21, line 349: No validation that s.activeInBytes() is non-negative. If a BlockCache impl has a bug causing negative values, the accumulated stat will be wrong. Consider Math.max(0L, s.activeInBytes()) or at least an assertion. we should have thsi on all

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -295,7 +202,99 @@ impl FoyerCache {
...
+
+        Self { inner, key_index, _runtime: Arc::new(rt), stats, shutdown }
+    }
+
+    /// Run one sweep of the key_index, removing entries whose underlying disk data is gone.
+    ///
+    /// Used in tests to trigger the sweep without waiting 30s.
+    /// In production the background task calls this logic automatically.
+    #[cfg(test)]
+    pub(crate) fn sweep_once(&self) -> usize {
```
nishchay21: For large caches with millions of entries, iterating all keys every 30s could spike CPU. Consider a MAX_KEYS_PER_SWEEP budget (e.g., 1000) to spread work across multiple cycles.

server/src/main/java/org/opensearch/index/store/remote/filecache/NodeCacheService.java
vishwasgarg18: Good point.

but `activeInBytes` is the only value whose value fluctuates. Added the max func
. All other 9 fields are  monotonic cumulative counters, adding a max to them would silently hide real accounting bugs.

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: Already have the shutdown mechanism, updated CancellationToken now though
```
impl Drop for FoyerCache {
    fn drop(&mut self) {
        self.shutdown.store(true, Ordering::Relaxed);
    }
}
```

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: Two-phase approach won't work

```
  Phase 1: read-only scan → collect stale keys → release all read locks
                                               ↑ GAP HERE
  Phase 2: write lock → remove stale keys

Race in the gap:
  put() inserts key X as a new live entry
  Phase 2 removes key X (it was in the stale list from Phase 1)
  Result: key X is gone from key_index but still in Foyer disk
          evict_prefix() can never see it — memory leak + index corruption

Current retain() approach:
  Write lock → retain() filter → release write lock (all in one atomic operation per shard)
  No gap. No race. Correct.
```


Now the problem with retain is performance
```
The proposal is to go ahead with DashMap level shards granulaity

DashMap has exactly 16 shards (configurable at construction but we use the default)
Each shard is processed completely in one retain() pass
The cursor records which shard to start from next cycle
Over ≤ 16 cycles, every shard is swept exactly once
```

Perform for search queries will only be impacted for queries which are not cached in disk and is not in the same dashmap shard.

Performance Analysis

Avg. Entry size | Cache size | Total entries | Entries/shard | retain() per shard | Cycles to full sweep | Max put() block
-- | -- | -- | -- | -- | -- | --
64 MB | 25 GB | 400 | 25 | ~250ns | 1 | ~250ns
1 MB | 25 GB | 25,600 | 1,600 | ~16μs | 1 | ~16μs
100 KB | 25 GB | 256,000 | 16,000 | ~160μs | 1 | ~160μs
4 KB | 25 GB | 6,500,000 | 406,000 | ~4ms | 16 (8 min) | ~4ms




Last thing - I'm adding % based sweeper. Only run after 30% 9 (setting). Tune it later.

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: Please refer to the above comment.

nishchay21 (approved):

sandbox/plugins/block-cache-foyer/src/main/rust/src/tests.rs
```diff
@@ -305,20 +533,279 @@ fn lru_eviction_removes_stale_keys_from_key_index() {
     }
     std::thread::sleep(std::time::Duration::from_millis(500));
```
mgodwan: This can be flaky
Can we do a busy assert pattern here?

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
```diff
@@ -337,12 +499,20 @@ impl BlockCache for FoyerCache {
...
         // index_key() normalizes by stripping leading '/' — evict_prefix uses the same normalization.
         let idx = Self::index_key(raw).to_string();
-        self.key_index.entry(idx).or_default().push(k);
+        // HashSet::insert returns true only when the key is new.
+        // If it returns false the key was already present (concurrent put / re-put after eviction)
+        // and we must NOT increment used_bytes again — the caller already paid for it on the first put.
+        let is_new = self.key_index.entry(idx).or_default().insert(k);
+        if is_new {
+            self.stats.used_bytes.fetch_add(size, Ordering::Relaxed);
+        }
```
mgodwan, line 510: same key cannot be updated, right?

mgodwan (approved):

mgodwan:
Aside from the statsflow, Will the foyer cache be able to recover on disk data on restart?

vishwasgarg18:
> Aside from the statsflow, Will the foyer cache be able to recover on disk data on restart?

No, since the foyer doesn't provider iterator we won't be able to update the registry (key-index), hence no recovery for on restart

sandbox/plugins/block-cache-foyer/src/main/rust/src/foyer/foyer_cache.rs
vishwasgarg18: Nope same key can't be updated
this will prevent from duplicate insertion - `let is_new = self.key_index.entry(idx).or_default().insert(k);`

sandbox/plugins/block-cache-foyer/src/main/rust/src/tests.rs
vishwasgarg18: done

PR #21170, opened 2026-04-08
Move bundled JDK to JDK-26

reta:
### Description
Move bundled JDK to JDK-26

### Related Issues
Most notable improvements:
 - [JEP 522: G1 GC: Improve Throughput by Reducing Synchronization](https://openjdk.org/jeps/522)
 - [JEP 517: HTTP/3 for the HTTP Client API](https://openjdk.org/jeps/517)

And, not yet relevant for us but still - [JDK-8369238: Allow virtual thread preemption on some common class initialization paths](https://bugs.openjdk.org/browse/JDK-8369238)

### Check List
- [x] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.
- [ ] Benchmarks run

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21718, opened 2026-05-18
Support for HTTP/3 (client side)

### LINKED ISSUE #20634: [FEATURE] Support for HTTP/3 (client side)

In https://github.com/opensearch-project/OpenSearch/issues/4451, we delivered experimental HTTP/3 support on the server side. However, none of the standard OpenSearch clients (rest client, opensearch-java, opensearch-rs, ...) do support HTTP/3 yet.

Add support of the HTTP/3 protocol on the client side (starting from rest client).

### Related component

Clients

N/A

### Additional context

As of today, the Apache HttpClient 5 that we rely on for rest client transport does not support HTTP/3 (and unluckily will in nearest future), see please https://issues.apache.org/jira/browse/HTTPCORE-793.

reta:
### Description
Support for HTTP/3 (client side) by introducing alternative, zero dependencies HTTP client built on top of JDK's `HttpClient`. 

#### Motivation
Recently we have introduced an experimental HTTP/3 support on the served side. The client side turned out to be more complicated since the Apache HttpClient 5 that we rely on for rest client transport does not support HTTP/3 (and unluckily will in nearest future), see please https://issues.apache.org/jira/browse/HTTPCORE-793. This limitation leads to uneasy options:
 - no support of the HTTP/3 on client side
 - provide alternative client implementation to support HTTP/3 (and HTTP/1.1, HTTP/2, ... ) 

#### Alternative base clients
There are basically two production ready clients that fit into OpenSearch technical stack:
 - Netty Client (and Reactor Netty that is built on top of it)
 - OpenJDK `HttpClient` that supports HTTP/3 as of JDK-26 [1], [2]

OpenJDK looks particularly appealing since it requires no dependencies vs Netty that provides the foundation but requires more implementation work. Although our baseline is JDK-21, the applications could use JDK-26 (or later) and get HTTP/3 support for free, no changes required on OpenSearch side.  Contrary, Netty's HTTP/3 implementation is built on top of native library (https://github.com/cloudflare/quiche) and is not available on many platforms.

#### Do we need new HTTP client?
The Rest Client is considered deprecated for a long time (we recommend to use `opensearch-java`), so we could consider it to be an internal change. Also, we stuck on Java 8 baseline for `opensearch-java` and `HttpClient` is only available in JDK-11, adding a support for it is possible as multi-release jar and will be added accordingly.

[1] https://github.com/opensearch-project/OpenSearch/pull/21170
[2]  https://openjdk.org/jeps/517

### Implementation Notes
The new client borrows a large chunk of implementation from the existing `RestClient` and preserves the attribution (license headers). The new client though has no dependencies besides JDK itself.

### Related Issues
Closes https://github.com/opensearch-project/OpenSearch/issues/20634

### Check List


reta:
@andrross @cwperks looking for early feedback folks. The pull request is not in ready state yet, mostly looking into "does it make sense or not" perspective, since we reach some limits of the client HTTP libraries we use. Thanks a lot, would really appreciate thoughts.

reta:
@andrross @cwperks I believe the pull request is in the good shape, wondering what do you think of it folks, thank you.

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RestHttpClient.java
```diff
@@ -0,0 +1,1095 @@
...
+                requestHeader.getValue().forEach(v -> httpRequest.header(requestHeader.getKey(), v));
+                requestNames.add(requestHeader.getKey());
+            }
+            for (Map.Entry<String, List<String>> defaultHeader : defaultHeaders.entrySet()) {
+                if (requestNames.contains(defaultHeader.getKey()) == false) {
+                    defaultHeader.getValue().forEach(v -> httpRequest.header(defaultHeader.getKey(), v));
+                }
+            }
+            if (compressionEnabled) {
+                httpRequest.header("Content-Encoding", "gzip");
```
andrross, line 870: Is this being added unconditionally even when there is no body (i.e. GET or HEAD)?

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/Request.java
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /**
+     * Query string parameters. The returned map is an unmodifiable view of the
+     * map in the request so calls to {@link #addParameter(String, String)}
+     * will change it.
+     */
+    public Map<String, String> getParameters() {
+        if (options.getParameters().isEmpty()) {
+            return unmodifiableMap(parameters);
```
andrross, line 40: Nitpick:


```suggestion
            return Map.of();
```

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RestHttpClient.java
```diff
@@ -0,0 +1,1095 @@
...
+                } else {
+                    fullPath = pathPrefix + "/" + path;
+                }
+            } else {
+                fullPath = path;
+            }
+
+            final String additionalQuery = params.entrySet()
+                .stream()
+                .map(e -> e.getKey() + "=" + URLEncoder.encode(e.getValue(), StandardCharsets.UTF_8))
```
andrross: I think `e.getValue()` comes from Request.getParameters(), which I think allows for a null parameter value, which would cause an NPE here. Is that right?

client/rest-http-client/build.gradle
```diff
@@ -0,0 +1,145 @@
...
+}
+
+tasks.named('forbiddenApisTest').configure {
+  //we are using jdk-internal instead of jdk-non-portable to allow for com.sun.net.httpserver.* usage
+  bundledSignatures -= 'jdk-non-portable'
+  bundledSignatures += 'jdk-internal'
+}
+
+// JarHell is part of es server, which we don't want to pull in
+// TODO: Not anymore. Now in :libs:opensearch-core
```
andrross: Is this still accurate? At least change the `es` text :)

client/rest-http-client/build.gradle
```diff
@@ -0,0 +1,145 @@
...
+      'reactor.blockhound.integration.BlockHoundIntegration'
+   )
+   ignoreViolations(
+      'reactor.core.publisher.CallSiteSupplierFactory$SharedSecretsCallSiteSupplierFactory',
+      'reactor.core.publisher.CallSiteSupplierFactory$SharedSecretsCallSiteSupplierFactory$TracingException'
+   )
+}
+
+tasks.withType(JavaCompile) {
+  // Suppressing '[options] target value 8 is obsolete and will be removed in a future release'
```
andrross: Does this need to stay?

andrross (commented):
> The Rest Client is considered deprecated for a long time (we recommend to use opensearch-java), so we could consider it to be an internal change.

What's the long term path not having to support both the existing low-level client and this one? The reality is that the high-level client has a lot of usage still and I've heard from users that there is a lot of concern about it going away (they don't want to re-write their applications), so I think we'll have to continue supporting the HLRC for the foreseeable future.

reta:
Thanks @andrross 

> What's the long term path not having to support both the existing low-level client and this one? The reality is that the high-level client has a lot of usage still and I've heard from users that there is a lot of concern about it going away (they don't want to re-write their applications), so I think we'll have to continue supporting the HLRC for the foreseeable future.

I think we could keep HLRC but depreciate `RestClient` in favour of this new `RestHttpClient`, it should remove large chunk of dependencies from us (plus, many things we could get from free just moving to latest JDKs officially (LTS) or unofficially), thanks!

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/Request.java
reta: We actually need readonly view over `parameters`, not an empty map instance

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RestHttpClient.java
reta: That is right, if body is not set or is empty, it is noop for server

client/rest-http-client/build.gradle
reta: Nope, removing, thank you

client/rest-http-client/build.gradle
reta: Nope, removing, thank you

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RestHttpClient.java
reta: Correct, should not happen, but definitely allowed by `HashMap`, addressed, thank you

andrross: @reta Request.addParameter seems to allow null?

> If {@code null} then the parameter is sent as {@code name} rather than {@code name=value}

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RestHttpClient.java
andrross: Isn't a content-encoding header on a request with no content technically incorrect? Seems like there might be some risk of an intermediate layer like a proxy or load balancer mishandling this.

client/rest-http-client/build.gradle
```diff
@@ -0,0 +1,134 @@
...
+apply from: "$rootDir/gradle/fips.gradle"
+
+java {
+  targetCompatibility = JavaVersion.VERSION_21
+  sourceCompatibility = JavaVersion.VERSION_21
+}
+
+base {
+  group = 'org.opensearch.client'
+  archivesName = 'opensearch-rest-http-client'
```
andrross, line 25: Is this going to publish a `opensearch-rest-http-client` artifact on the next release? At least for now I think this is intended as an experimental, internal-only low-level client. Do we have a way to communicate that more clearly, and also differentiate this from the existing `opensearch-rest-client` artifact?

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RestHttpClient.java
reta: Yes it does, I added filtering in the latest commit [1] to address that:

```
.filter(e -> e.getValue() != null)
```

Thank you!

[1] https://github.com/opensearch-project/OpenSearch/pull/21718/changes/2748cb6572cae444d143861694a07e011af54a80

client/rest-http-client/build.gradle
reta: > Is this going to publish a opensearch-rest-http-client artifact on the next release

Correct, for 3.8.0

> At least for now I think this is intended as an experimental, internal-only low-level client. Do we have a way to communicate that more clearly, and also differentiate this from the existing opensearch-rest-client artifact?

Yes, aligned with https://github.com/opensearch-project/OpenSearch/pull/21718#issuecomment-4623682456, the suggestion would be to finally deprecate `RestClient`. Also, which I haven't thought about, we could host the new client under `org.opensearch.internal.httpclient` or `org.opensearch.internal.rest` to emphasize the fact this is indeed internal-only low-level client. The `@ExprimentalApi` annotation would be somewhat useful here, but this new client has literally zero dependencies (not even `opensearch-common`) so I added experimental part to every single javadoc for public classes. Wdyt? Thanks!

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RestHttpClient.java
reta: There is a trade off here, the JDK's `HttpClient` operates over `BodyPublisher`, to understand if there is data to be sent, we have to consume it. I don't really think it could be mishandled since empty body would not be subject to decompression, but if you could point me out to some examples, I would be happy to look what kind of issues it may cause. Thanks again!

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RequestLogger.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Helper class that exposes static methods to unify the way requests are logged.
+ * Includes trace logging to log complete requests and responses in curl format.
+ * Useful for debugging, manually sending logged requests via curl and checking their responses.
+ * Trace logging is a feature that all the language clients provide.
+ */
+final class RequestLogger {
```
andrross, line 55: Is it worth using record classes where possible? I think there are a number of places to reduce boilerplate and avoid the manual equals/hashcode/toString implementations.

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/Request.java
```diff
@@ -0,0 +1,207 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import static java.util.Collections.unmodifiableMap;
+
+/**
+ * HTTP Request to OpenSearch.
+ * Note: This is an experimental API.
+ */
+public final class Request {
```
andrross: I think you mostly ported this structure from the existing HTTP client, but is it worth refactoring some of these to use a builder pattern with immutable instances (and use record classes)?

client/rest-http-client/build.gradle
andrross: I like the idea of putting `internal` in the package name. Either option is good with me, maybe a slight preference toward `org.opensearch.internal.httpclient`.

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RequestLogger.java
reta: Yeah, the [HttpHost](https://github.com/opensearch-project/OpenSearch/pull/21718/changes#diff-091d7882ea12ec09095649743cf002eddb424110cbd7dc6576fa42243eb72f16) is a good example, this one (RequestLogger) is not a data class - it is purely log helper methods

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/Request.java
reta: Yeah, I think it would make sense to modernize it a bit, I will work on it shortly

client/rest-http-client/build.gradle
reta: Moved under `org.opensearch.internal.httpclient`, thank you

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/Request.java
reta: @andrross I think I migrated most of the classes to records, the `StreamRequest` / `StreamResponse` pair was not converted due to the fact records cannot hide/keep state., that would expose the internals (like publishers) to the outside, not a desirable outcome. Hope it makes sense, thank you!

client/rest-http-client/src/main/java/org/opensearch/internal/httpclient/RestHttpClient.java
andrross: If the value is null don't you still need to include the key? For example for a request like `http://localhost?pretty`

reta: This is good point, thanks @andrross , fixed that and also added test cases to both clients to make sure the behaviour is consistent and identical.

andrross (approved):

PR #21719, opened 2026-05-18
[sandbox] Fix Utf8View vs Utf8 type mismatch in Substrait filter evaluation

aasom143:
### Description
Fixes Utf8View vs Utf8 type mismatch errors when evaluating filter predicates on parquet-backed indices.

The Substrait plan carries Utf8 literals but the parquet schema stores string columns as Utf8View, causing Arrow to reject mixed-type comparisons and like predicates. This applies DataFusion's `TypeCoercionRewriter` to expressions before creating physical expressions, inserting the necessary casts (e.g. Utf8 → Utf8View) using the same logic as DataFusion's analyzer pass.

**Issues fixed:**
1. LIKE predicate assertion failure (`Utf8View AND Utf8 of like physical should be same`) on text fields.
2. Equality comparison error (`Invalid comparison operation: Utf8View == Utf8`) on keyword fields.

**Verified on ClickBench 100M document setup:**
```
curl -XPOST 'http://localhost:9200/_plugins/_ppl' -H 'Content-Type: application/json' -d '{"query": "source=clickbench | where match(Title, '\''google'\'') AND BrowserCountry = '\''B8'\'' | stats count()"}'
{
  "schema": [
    {
      "name": "count()",
      "type": "bigint"
    }
  ],
  "datarows": [
    [
      7
    ]
  ],
  "total": 1,
  "size": 1
}
```

### Related Issues

### Check List

### DIFF
--- a/distribution/build.gradle
+++ b/distribution/build.gradle
@@ -311,7 +311,7 @@ configure(subprojects.findAll { ['archives', 'packages'].contains(it.name) }) {
    *             Properties to expand when copying packaging files             *
    *****************************************************************************/
   configurations {
-    ['libs', 'libsPluginCli', 'libsKeystoreCli', 'libsFipsInstallerCli', 'bcFips'].each {
+    ['libs', 'libsPluginCli', 'libsKeystoreCli', 'libsFipsInstallerCli', 'libsHeapProfCli', 'bcFips'].each {
       create(it) {
         canBeConsumed = false
         canBeResolved = true
@@ -340,6 +340,7 @@ configure(subprojects.findAll { ['archives', 'packages'].contains(it.name) }) {
     libsPluginCli project(':distribution:tools:plugin-cli')
     libsKeystoreCli project(path: ':distribution:tools:keystore-cli')
     libsFipsInstallerCli project(path: ':distribution:tools:fips-demo-installer-cli')
+    libsHeapProfCli project(path: ':distribution:tools:heap-prof-cli')
 
     bcFips libs.bundles.bouncycastle
 
@@ -367,6 +368,9 @@ configure(subprojects.findAll { ['archives', 'packages'].contains(it.name) }) {
         // Add FIPS installer CLI
         into('tools/fips-demo-installer-cli') {
           from(configurations.libsFipsInstallerCli)
+        }
+        into('tools/heap-prof-cli') {
+          from(configurations.libsHeapProfCli)
         }
       }
     }
--- a/distribution/src/bin/opensearch-heap-prof
+++ b/distribution/src/bin/opensearch-heap-prof
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e -o pipefail
+
+OPENSEARCH_MAIN_CLASS=org.opensearch.tools.cli.heapprof.HeapProfCli \
+  OPENSEARCH_ADDITIONAL_CLASSPATH_DIRECTORIES=lib/tools/heap-prof-cli \
+  "`dirname "$0"`"/opensearch-cli \
+  "$@"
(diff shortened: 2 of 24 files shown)


alchemist51:
Can we please add ITs?

alchemist51:
let's add in both: indexed path fuzzy testing and also end to end. Thanks!

mch2:
this should no longer be required after https://github.com/opensearch-project/OpenSearch/commit/35ce14790c2d3daabb6084eea7689e53a4ab8d93 @sandeshkr419 can u take a stab here?

sandeshkr419:
Yeah, this shouldn't be required after my recent changes.

aasom143:
> this should no longer be required after [35ce147](https://github.com/opensearch-project/OpenSearch/commit/35ce14790c2d3daabb6084eea7689e53a4ab8d93) @sandeshkr419 can u take a stab here?
> Yeah, this shouldn't be required after my recent changes.

@mch2 @sandeshkr419 thanks for checking the PR.
This fix aims to fix single collector scenarios where residual expression contains the filtering on keyword type field.
I have tried following query in main branch which still seems failing with same issue. Please check the following query and let me know.
```
curl -XPOST 'http://localhost:9200/_plugins/_ppl' -H 'Content-Type: application/json' -d '{"query": "source=clickbench | where match(Title, '\''google'\'') AND BrowserCountry = '\''B8'\'' | stats count()"}'
```

sandeshkr419 (changes requested):
@aasom143, confirmed my fix doesn't cover this path., it fixes the producer→consumer StreamingTable schema flow on the coordinator; this site (expr_to_bool_tree on the data node) bypasses DataFusion's analyzer entirely and goes straight to create_physical_expr, so TypeCoercion never runs. Real bug, your fix addresses the right concern, but not at the right place.

Your mechanism: TypeCoercionRewriter produces `Cast(literal AS Utf8View)`. We're skipping the optimizer on this path too, so the Cast survives into the physical expression and into the parquet pushdown predicate handed to ParquetSource::with_predicate — every BoolNode::Predicate leaf evaluates a Cast per batch.

What we could do instead is walk the Expr once before convert_expr and retag ScalarValue::Utf8(x) / ScalarValue::LargeUtf8(x) → ScalarValue::Utf8View(x) when the sibling resolves to Utf8View. Same end state (Utf8View literal vs Utf8View column), no Cast node, no per-batch coercion. Covers =/<>/</<=/>/>= and the IS DISTINCT / regex-match family, plus Like, Between, InList. AND/OR/NOT recurse through.

I sketched the draft proposed changes on a commit: https://github.com/sandeshkr419/OpenSearch/commit/2b29a65dac8b0a160e8d34367032c4d24333b724, 16 unit tests covering each shape, swapped operands, LargeUtf8, NULL widening, all comparison ops, recursion, and end-to-end through expr_to_bool_tree. Lib suite (816 tests) green. Feel free to take (cherry-pick) it as-is, and please add yout IT coverage on top.

alchemist51:
@sandeshkr419  instead of doing that, I have another simpler fix for handling more cases actually:
```
// Pull the underlying array out of the Scalar wrapper
      let scalar_inner = scalar.into_inner();   // Arc<dyn Array>

      // If literal type ≠ column type, cast the literal to match
      let scalar_inner = if scalar_inner.data_type() != col_arr.data_type() {
          datafusion::arrow::compute::cast(scalar_inner.as_ref(), col_arr.data_type())
              .map_err(|e| e.to_string())?
      } else {
          scalar_inner
      };

      // Re-wrap as Scalar so the kernel treats it as a single-row scalar
      let scalar = datafusion::arrow::array::Scalar::new(scalar_inner);

      let kernel_result = match *bin.op() {
          Operator::Eq => eq(col_arr, &scalar),    // ← now matches
          ...
      };
```

PR #21721, opened 2026-05-18
feat(wlm): add search.max_buckets to workload group settings

### LINKED ISSUE #20556: [Feature Request] Onboard search.max_buckets to WLM group search settings

OpenSearch’s max buckets setting limits the maximum number of aggregation buckets that a single search response is allowed to produce. If a query’s aggregations would generate more buckets than the limit, OpenSearch fails the request instead of trying to build an excessively large response.
​
This setting is important because bucket aggregations can explode in size (for example, high-cardinality terms aggregations or many nested bucket aggregations), and the work to build/reduce those buckets consumes memory and CPU. By capping the bucket count, search.max_buckets acts as a guardrail that helps protect cluster stability and prevents a single expensive aggregation query from overwhelming nodes.

Onboard the max buckets setting as a per-WLM group setting as part of https://github.com/opensearch-project/OpenSearch/issues/20555. This will provide OpenSearch administrators finer control when setting limits and protecting memory usage.

### Related component

Search:Resiliency

_No response_

### Additional context

_No response_

dzane17:
### Description
Onboards `search.max_buckets` as a Workload Management search setting. Today, `search.max_buckets` is a single cluster-wide setting, leaving operators with no way to apply per-tenant bucket limits. This feature gives users the flexibility to assign more restrictive bucket limits to protect the system, or more permissive limits when their use case calls for it.

Unlike the four settings added in #21523, `search.max_buckets` does **not** exist as a per-request parameter — it only exists as a cluster setting. The value is enforced by the `MultiBucketConsumerService` class at two sites: (1) during the query phase on each data node, and (2) during the final reduce on the coordinator. This PR includes the necessary piping so that `MultiBucketConsumerService` can resolve the WLM `max_buckets` value from the workload group ID passed in thread context (already propagated to shards).

Because there is no per-request value to override, `override_request_values` is not relevant for this setting. The WLM value, when defined, always wins over the cluster setting — workload groups are admin-configured and more granular than the cluster-wide setting.

### Related Issues
Resolves #20556
Part of #20555

### Check List

### DIFF
--- a/plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
+++ b/plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
@@ -158,7 +158,8 @@ public class WorkloadManagementRestIT extends OpenSearchRestTestCase {
                     "search.default_search_timeout": "30s",
                     "search.cancel_after_time_interval": "1m",
                     "search.max_concurrent_shard_requests": "5",
-                    "search.batched_reduce_size": "512"
+                    "search.batched_reduce_size": "512",
+                    "search.max_buckets": "1000"
                 }
             }""";
         Response response = performOperation("PUT", "_wlm/workload_group", createJson);
@@ -173,6 +174,7 @@ public class WorkloadManagementRestIT extends OpenSearchRestTestCase {
         assertTrue(responseBody.contains("\"search.cancel_after_time_interval\":\"1m\""));
         assertTrue(responseBody.contains("\"search.max_concurrent_shard_requests\":\"5\""));
         assertTrue(responseBody.contains("\"search.batched_reduce_size\":\"512\""));
+        assertTrue(responseBody.contains("\"search.max_buckets\":\"1000\""));
 
         // Update search settings
         String updateJson = """
@@ -181,7 +183,8 @@ public class WorkloadManagementRestIT extends OpenSearchRestTestCase {
                     "search.default_search_timeout": "1m",
                     "search.cancel_after_time_interval": "5m",
                     "search.max_concurrent_shard_requests": "10",
-                    "search.batched_reduce_size": "256"
+                    "search.batched_reduce_size": "256",
+                    "search.max_buckets": "500"
                 }
             }""";
         Response updateResponse = performOperation("PUT", "_wlm/workload_group/search_test", updateJson);
@@ -194,6 +197,7 @@ public class WorkloadManagementRestIT extends OpenSearchRestTestCase {
         assertTrue(responseBody2.contains("\"search.cancel_after_time_interval\":\"5m\""));
         assertTrue(responseBody2.contains("\"search.max_concurrent_shard_requests\":\"10\""));
         assertTrue(responseBody2.contains("\"search.batched_reduce_size\":\"256\""));
+        assertTrue(responseBody2.contains("\"search.max_buckets\":\"500\""));
 
         performOperation("DELETE", "_wlm/workload_group/search_test", null);
     }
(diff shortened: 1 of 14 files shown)


server/src/main/java/org/opensearch/search/aggregations/MultiBucketConsumerService.java
```diff
@@ -79,6 +99,43 @@ private void setMaxBucket(int maxBucket) {
...
+     * The WLM-set value, when present, always wins — {@code override_request_values} is not
+     * relevant because {@code search.max_buckets} is not a per-request parameter.
+     */
+    int resolveMaxBuckets() {
+        try {
+            if (threadPool == null || workloadGroupService == null) {
+                return maxBucket;
+            }
+            ThreadContext threadContext = threadPool.getThreadContext();
+            if (threadContext == null) {
```
cwperks: I know this is defensive coding, but are threadPool and threadContext really nullable here?

server/src/main/java/org/opensearch/search/aggregations/MultiBucketConsumerService.java
```diff
@@ -79,6 +99,43 @@ private void setMaxBucket(int maxBucket) {
...
+                return maxBucket;
+            }
+            String workloadGroupId = threadContext.getHeader(WorkloadGroupTask.WORKLOAD_GROUP_ID_HEADER);
+            if (workloadGroupId == null) {
+                return maxBucket;
+            }
+            WorkloadGroup workloadGroup = workloadGroupService.getWorkloadGroupById(workloadGroupId);
```
cwperks: Is there not an existing place to extract workloadGroup from the threadContext already? This code looks like it may be shared in multiple places.

plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
```diff
@@ -315,6 +319,67 @@ public void testSearchSettingsInvalidSettingsRejected() throws Exception {
...
+        ResponseException invalidMaxBucketsException = expectThrows(
+            ResponseException.class,
+            () -> performOperation("PUT", "_wlm/workload_group", invalidMaxBucketsJson)
+        );
+        String invalidMaxBucketsBody = EntityUtils.toString(invalidMaxBucketsException.getResponse().getEntity());
+        assertTrue(invalidMaxBucketsBody.contains("search.max_buckets"));
+        assertTrue(invalidMaxBucketsBody.contains("Invalid value"));
+    }
+
+    public void testSearchMaxBucketsCreateAndUpdate() throws Exception {
```
cwperks, line 344: What about the case where a request exceeds a WLM Group's search.max_buckets size and below the cluster-level search.max_buckets size? I think we should have a test to ensure the workload group's setting is prioritized if a request matches the group.

server/src/main/java/org/opensearch/search/aggregations/MultiBucketConsumerService.java
dzane17: Good catch, `threadPool` is always non-null at the only production call site (`SearchService`), and `ThreadPool.getThreadContext()` returns a `final` field initialized in the constructor. I've removed those two checks.

server/src/main/java/org/opensearch/search/aggregations/MultiBucketConsumerService.java
dzane17: The same pattern existed in `applyWorkloadGroupSearchSettings`. I created a new helper method `WorkloadGroupService.resolveFromThreadContext(ThreadContext)` and updated both call sites to use it. Also added unit tests in `WorkloadGroupServiceTests`.

plugins/workload-management/src/javaRestTest/java/org/opensearch/rest/WorkloadManagementRestIT.java
dzane17: Added an integ test `testSearchMaxBucketsEnforcedAtRequestPath` which creates a WLM group with `search.max_buckets=1`, indexes two docs into a single-shard index, then runs a `terms` aggregation request twice. First try without the WLM group tag succeeds since the cluster default setting permits it. Second try with the `workloadGroupId` header set is rejected with `too_many_buckets` exception because the group setting is `max_buckets=1`. That demonstrates the WLM value taking precedence and limiting the request.

server/src/main/java/org/opensearch/wlm/WorkloadGroupService.java
```diff
@@ -334,6 +335,18 @@ public WorkloadGroup getWorkloadGroupById(String workloadGroupId) {
         return clusterService.state().metadata().workloadGroups().get(workloadGroupId);
     }
 
+    /**
+     * Resolves the workload group attached to the calling thread context, or null if there is
+     * no workload group ID header set or the referenced group does not exist.
+     */
+    public WorkloadGroup resolveFromThreadContext(ThreadContext threadContext) {
```
cwperks: Isn't threadPool already at class level? Can we use that?

dzane17: Fixed

server/src/main/java/org/opensearch/search/aggregations/MultiBucketConsumerService.java
```diff
@@ -79,6 +93,35 @@ private void setMaxBucket(int maxBucket) {
...
+     * <p>
+     * The WLM-set value, when present, always wins — {@code override_request_values} is not
+     * relevant because {@code search.max_buckets} is not a per-request parameter.
+     */
+    int resolveMaxBuckets() {
+        try {
+            if (workloadGroupService == null) {
+                return maxBucket;
+            }
+            WorkloadGroup workloadGroup = workloadGroupService.resolveFromThreadContext();
```
cwperks: Can we rename to `getCurrentWorkloadGroup` which would be nullable if the current request does not map to a workload group?

dzane17: Done

cwperks (approved):

PR #21723, opened 2026-05-18
analytics-engine: drive reduce through stage tasks

mch2:
### Description
  Splits reduce-sink work into two responsibilities and collapses three execution variants into one.

  **Before:** `AbstractDatafusionReduceSink.close()` was overloaded — it both *executed* the reduce plan and *released* native resources. The coordinator-reduce side had three near-duplicate `StageExecution` subclasses (one per
  scheduling mode) because they each wired the `close()` overload differently.

  **After:**
  - New `ReducingExchangeSink` SPI separates `reduce(ActionListener<Void>)` (does the work, fires listener on completion) from `close()` (idempotent cleanup, runs on every terminal regardless of whether reduce ran).
  - One `ReduceStageExecution` covers both streaming and buffered modes — the sink declares which it wants via `supportsEagerScheduling()`, and the scheduler does the rest. Streaming sink anchors native cleanup on the `reduce()` caller
  via a `SinkState` machine so cancel-before-reduce paths release resources without double-firing or leaking.

  Also relocates the per-stage classes into `coordinator/` and `shard/` subpackages (`ReduceStageExecution`, `LocalStageTask`/`LocalTaskRunner`, `PassThroughStageExecution`, `ShardFragmentStageExecution`, etc.) to make the topology
  obvious and avoid a flat `stage/` package with 20+ files.

  ### Reviewer guide

  38 files but only ~6 are load-bearing. Most of the rest are file moves into the new subpackages and corresponding test relocations.

  **Read these carefully:**
  - `sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ReducingExchangeSink.java` — the new SPI contract
  - `sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/AbstractDatafusionReduceSink.java` — base lifecycle (now smaller)
  - `sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionReduceSink.java` — streaming variant with `SinkState`
  - `sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionMemtableReduceSink.java` — buffered variant
  - `sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/coordinator/ReduceStageExecution.java` — single execution unifying the three old variants
  - `sandbox/plugins/analytics-backend-datafusion/src/test/java/org/opensearch/be/datafusion/DatafusionReduceSinkTests.java` — before-/after-first-batch cancellation test coverage

  Everything else is mechanical:
  - Stage subclass moves into `coordinator/` and `shard/` subpackages
  - `LocalStageExecutionTests` deletion (covered by new `ReduceStageExecutionTests`)
  - Per-test signature updates for the new `reduce(listener)` API
### Related Issues

### Check List


mch2:
pr reviewer - 
2nd and 3rd are non issues - 2 - The memtable variant doesn't have a pending input channel — reduce() runs registerMemtable against the already-buffered Arrow arrays, 3 - outStream is closed exactly once via closeUnderLock() → outStream.close().

Fixing 1.

sandeshkr419 (commented):
Potential miss in the cancel-during-REDUCING path - can you validate if this is intended or a miss?

-----

Trace through external close() arriving while drain is parked:

```  
external close() ─── AbstractDatafusionReduceSink.close()
                         sync(feedLock):
                           closed=false → set closed=TRUE       ← (A) base flag flipped
                         closeImpl()
                           CAE(READY,DONE) returns REDUCING     ← (B) state unchanged
                           REDUCING branch:
                             fireCancelQuery()
                             return null                        ← (C) NO teardown here


  drain unwinds (cancel-Err) ── reduce()'s finally:
                         state.set(DONE)
                         super.close()
                           sync(feedLock):
                             closed=TRUE → return early         ← (D) closeImpl NEVER runs
```

  The docstring on closeImpl says: "REDUCING: drain is parked. Fire cancel_query so it unwinds — the in-flight reduce()'s finally will re-enter via super.close() with state == DONE and do the actual teardown then." 

But the base class's closed short-circuit at (D) prevents that re-entry from doing anything. Net effect: outStream.close() and session.close() never run on the cancel-while-reducing path. (sender.close() is still reachable via ChildSink.close() from the orchestrator, but that's not guaranteed to fire on every cancel path.)

  The before-/after-first-batch tests only assert that cancel_query is invoked and that the listener fires, they don't assert that outStream/session actually got torn down. Worth adding a leak assertion on DatafusionReduceSinkTests.java, or restructure so closeImpl runs exactly once on the drain's thread. Fix should be small (e.g. closeImpl runs unconditionally in the REDUCING-detected branch from reduce's finally, not via super.close()).

sandeshkr419 (approved):

PR #21726, opened 2026-05-19
Add configurable coordinator buffer limit for per-query Arrow allocator

bowenlan-amzn:
## Summary

- Replace hardcoded 256MB per-query Arrow allocator limit with dynamic cluster setting `analytics.coordinator.buffer_limit`
- Add debug/warn logging at allocator close to detect memory leaks (WARN if bytes > 0 at close)

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
@@ -18,6 +18,7 @@ import org.opensearch.action.ActionRequest;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.action.support.TimeoutTaskCancellationUtility;
+import org.opensearch.analytics.AnalyticsPlugin;
 import org.opensearch.analytics.EngineContext;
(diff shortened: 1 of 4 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/QueryContext.java
mch2, line 97: can we treat a limit of -1 as don't spawn a child allocator?

bowenlan-amzn: done. now have a coordinator level arrow allocator. and create per query child allocator if the setting is not 0.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -96,7 +101,10 @@ public DefaultPlanExecutor(
         this.taskManager = transportService.getTaskManager();
         this.client = client;
         this.scheduler = scheduler;
-        this.allocatorService = allocatorService;
+        this.coordinatorAllocator = allocatorService.newChildAllocator("coordinator", Long.MAX_VALUE);
```
mch2, line 104: when it lands, lets set a bound on pls from https://github.com/opensearch-project/OpenSearch/pull/21703

mch2 (approved):

PR #21727, opened 2026-05-19
Add DELETE task API for stored task results

### LINKED ISSUE #21724: [Feature Request] Automatic cleanup of completed task results in .tasks index
## Is your feature request related to a problem? Please describe

When async operations (update-by-query, reindex, delete-by-query, force-merge, etc.) are submitted with `wait_for_completion=false`, their results are persisted to the `.tasks` system index via `TaskResultsService.storeResult()`. However, **there is no mechanism to remove these results after they have been consumed or after a configurable retention period**.

In production clusters that heavily use async operations (e.g., automated UBQ pipelines, scheduled reindex jobs), the `.tasks` index grows unbounded over time. This leads to:

1. **Wasted storage** — completed task results that will never be read again accumulate indefinitely
2. **Cluster state overhead** — the `.tasks` index metadata contributes to cluster state size
3. **Operational burden** — users must manually run delete-by-query against `.tasks` or set up external ISM policies as a workaround

## Describe the solution you'd like

A two-layer cleanup strategy:

### Layer 1: Delete on retrieval (primary mechanism)

When a client retrieves a completed task result via `GET /_tasks/<taskId>`, the document is deleted from the `.tasks` index after the response is returned. This is the natural lifecycle — the result exists only so the client can poll for it; once read, it's served its purpose.

- **Default behavior**: delete completed results on retrieval (`cleanup=true` by default)
- **Opt-out**: `GET /_tasks/<taskId>?cleanup=false` to retain the result for re-reading
- **Safe**: only deletes *completed* results; in-progress tasks are unaffected
- **Non-blocking**: deletion happens asynchronously after the response is sent; failure to delete doesn't affect the client response

### Layer 2: TTL-based fallback (safety net)

A background service catches orphaned results that are never retrieved (client crashed, fire-and-forget patterns, etc.):

| Setting | Default | Description |
|---------|---------|-------------|
| `task.result.ttl` | `-1` (disabled) | Time-to-live for completed task results. Results older than this are eligible for deletion. Example: `12h`, `7d` |
| `task.result.cleanup.interval` | `1h` | How often the cleanup service runs |
| `task.result.cleanup.batch_size` | `1000` | Maximum documents deleted per cleanup cycle |

### Combined behavior

```
Client submits async UBQ (wait_for_completion=false)
  → Task runs, completes, result stored in .tasks
  → Client polls GET /_tasks/<id>
  → Response returned, .tasks document deleted  ← Layer 1

If client never polls:
  → After TTL expires, background service deletes it  ← Layer 2
```

### Layer 1: Delete on retrieval

The change is in `TransportGetTaskAction.onGetFinishedTaskFromIndex()`. After successfully parsing and returning the `TaskResult`, issue an async delete:

```java
// After returning the response to the client:
if (cleanup && result.isCompleted()) {
    client.prepareDelete(TaskResultsService.TASK_INDEX, response.getId())
        .execute(ActionListener.wrap(
            r -> logger.trace("Cleaned up completed task result [{}]", response.getId()),
            e -> logger.debug("Failed to clean up task result [{}]: {}", response.getId(), e.getMessage())
        ));
}
```

Changes needed:
- `GetTaskRequest`: add `boolean cleanup` field (version-gated for wire compat)
- `RestGetTaskAction`: parse `?cleanup=true|false` parameter (default: `true`)
- Cluster-level override: `task.result.cleanup_on_get` setting (default: `true`)

### Layer 2: TTL-based cleanup service (`TaskResultsCleanupService`)

- Runs on elected cluster manager only (implements `ClusterStateListener`)
- Periodically runs delete-by-query against `.tasks` for completed results older than TTL
- Disabled by default (`task.result.ttl=-1`)
- Batched deletions to avoid resource spikes
- Requires adding a `completion_time` field to the task result mapping (version bump)

### Settings

```
task.result.cleanup_on_get = true          # Layer 1 default
task.result.ttl = -1                       # Layer 2 disabled by default
task.result.cleanup.interval = 1h          # Layer 2 frequency
task.result.cleanup.batch_size = 1000      # Layer 2 batch limit
```

## Backward Compatibility

- **Layer 1** defaults to `true` but is overridable per-request (`?cleanup=false`) and per-cluster
- **Layer 2** is disabled by default
- Wire format: `cleanup` field in `GetTaskRequest` is version-gated; older nodes ignore it
- Existing `.tasks` documents work fine — Layer 1 doesn't need any mapping change, Layer 2 can fall back to `start_time_in_millis` for old documents

## Affected operations

Any operation that supports `wait_for_completion=false`:
- Update by query
- Delete by query  
- Reindex
- Force merge
- Open index
- Resize (split/shrink/clone)

## Related component

Cluster Manager

## Related issues

- #9183 — Tasks not completed or canceled (orphaned in-memory tasks, different problem)

cwperks:
### Description

Adds Layer 1 of #21724 with a dedicated `DELETE /_tasks/{task_id}` API for deleting stored completed task results.

The new delete API removes a completed task result from the `.tasks` index and returns an acknowledged response. Running tasks are not cancelled or deleted by this API; callers should continue to use the cancel tasks API for running tasks.

The change wires the new action through transport, REST routing, REST API spec, the Java transport client, and the high-level REST client. It also marks the built-in `.tasks*` system index descriptor as `UnrestrictedSystemIndexDescriptor`, which allows security-aware plugins to distinguish task result reads from normal restricted system-index access while continuing to block direct writes through their existing system-index protections.

### API Behavior

Delete a stored completed task result:

```bash
curl -XDELETE "http://localhost:9200/_tasks/node-1:12345"
```

Successful response:

```json
{
  "acknowledged": true
}
```

After the stored result is deleted, a follow-up `GET /_tasks/node-1:12345` returns `404` because the task is no longer running and no stored task result remains.

This API intentionally does not cancel or remove running tasks. If the task is still running on the node that owns it, OpenSearch returns a conflict response:

```json
{
  "error": {
    "type": "opensearch_status_exception",
    "reason": "task [node-1:12345] is still running and cannot be deleted; use the cancel tasks API to cancel running tasks"
  },
  "status": 409
}
```

If the task id does not correspond to a running task or a stored result in `.tasks`, OpenSearch returns `404`:

```json
{
  "error": {
    "type": "resource_not_found_exception",
    "reason": "task [node-1:12345] isn't running and hasn't stored its results"
  },
  "status": 404
}
```

If a task id belongs to a node that is no longer part of the cluster, the delete API still attempts to remove a stored result from `.tasks`. If no stored result exists, OpenSearch returns `404` with context that the owning node is no longer in the cluster.

If a stored task result is found but is not marked completed, OpenSearch returns `409` and leaves the stored document in place.

### Related Issues

Resolves #21724

### Check List

- [x] All tests pass.
- [x] New functionality has been documented.

### Testing

- `./gradlew :server:test --tests org.opensearch.action.admin.cluster.node.tasks.delete.DeleteTaskRequestTests :client:rest-high-level:test --tests org.opensearch.client.TasksRequestConvertersTests`
- `./gradlew :server:internalClusterTest --tests org.opensearch.action.admin.cluster.node.tasks.TasksIT.testTaskStoringSuccessfulResult --tests org.opensearch.action.admin.cluster.node.tasks.TasksIT.testTaskStoringFailureResult --tests org.opensearch.action.admin.cluster.node.tasks.TasksIT.testNodeNotFoundButTaskFound`
- `./gradlew :server:precommit :client:rest-high-level:precommit :rest-api-spec:validateRestSpec :rest-api-spec:validateNoKeywords`
- `./gradlew spotlessApply`
- `./gradlew :server:test --tests org.opensearch.indices.SystemIndicesTests`
- `./gradlew publishToMavenLocal -Pcrypto.standard=FIPS-140-3 -Dbuild.snapshot=true -i`

Attempted a security consumer test after publishing Core locally:

```bash
./gradlew integrationTest --tests org.opensearch.security.systemindex.SystemIndexTests.regularUserShouldBeAbleToSearchUnrestrictedSystemIndex
```

That security build currently stops before the test runs due to the existing integration-test classpath conflict:

```text
Could not resolve com.fasterxml.jackson.core:jackson-annotations:2.22.
Conflict found for module 'com.fasterxml.jackson.core:jackson-annotations': between versions 2.22 and 2.21
```

<details>
<summary>Manual REST verification against a local 3.7.0-SNAPSHOT node</summary>

With OpenSearch running at `http://localhost:9200`:

Start from a clean slate so the test indices do not inherit state from a previous run.

```bash
curl -s -XDELETE 'http://localhost:9200/movies_source,movies_archive?ignore_unavailable=true'
```

Create a small source index with no replicas so the test is deterministic on a single local node.

```bash
curl -s -XPUT 'http://localhost:9200/movies_source' \
  -d '{"settings":{"number_of_shards":1,"number_of_replicas":0}}'
```

Index a couple of documents and refresh immediately so the async reindex task has known source content.

```bash
curl -s -XPOST 'http://localhost:9200/movies_source/_bulk?refresh=true' \
  -H 'Content-Type: application/x-ndjson' \
  -d '{"index":{"_id":"1"}}
{"title":"The Matrix","year":1999}
{"index":{"_id":"2"}}
{"title":"Spirited Away","year":2001}
'
```

Start `_reindex` asynchronously. This returns a task id and causes OpenSearch to persist the completed task result in `.tasks`.

```bash
TASK_ID=$(curl -s -XPOST 'http://localhost:9200/_reindex?wait_for_completion=false' \
  -d '{"source":{"index":"movies_source"},"dest":{"index":"movies_archive"}}' \
  | jq -r '.task')
```

Fetch the task with `wait_for_completion=true`. This waits for completion and confirms the completed stored task result can be read.

```bash
curl -s -i "http://localhost:9200/_tasks/${TASK_ID}?wait_for_completion=true"
```

Fetch the same task again to confirm the stored result is still available before deletion.

```bash
curl -s -i "http://localhost:9200/_tasks/${TASK_ID}"
```

Delete the stored completed task result using the dedicated delete API.

```bash
curl -s -i -XDELETE "http://localhost:9200/_tasks/${TASK_ID}"
```

Fetch the task again. The expected `404` confirms the stored completed task result was deleted.

```bash
curl -s -i "http://localhost:9200/_tasks/${TASK_ID}"
```

Remove the temporary indices created by the manual test.

```bash
curl -s -XDELETE 'http://localhost:9200/movies_source,movies_archive?ignore_unavailable=true'
```

Expected behavior:

- The first fetch returns `200` for the completed task result.
- The second fetch also returns `200`, confirming the stored result is available before deletion.
- `DELETE /_tasks/${TASK_ID}` returns `200` with `"acknowledged":true`.
- A subsequent fetch returns `404`, confirming the stored completed task result was deleted.

</details>


rishabhmaurya:
Thanks @cwperks for working on it. I'm not sure if passing a param to GET api is the right way to cleanup stored tasks here.

cwperks:
> Thanks @cwperks for working on it. I'm not sure if passing a param to GET api is the right way to cleanup stored tasks here.

@rishabhmaurya what do you suggest for cleanup? FYI if you look at the linked issue this PR is one of 2 proposals. The other is a TTL on .tasks entries so they can periodically be cleaned up on expiry (if the task is completed). wdyt?

I suggested this because I think it would make sense for a caller to explicitly request cleanup when they are done with the task.

rishabhmaurya:
@cwperks can we support POST on this API?
TTL based approach might be a breaking change for users as polling for async task beyond this TTL would result in empty response. Ideally, we can start with a high TTL, say 15 or 30 days as default and also have this POST api with cleanup param. wdyt?

sgup432:
I am not sure if adding a cleanup parameter is the right way to do it.
I believe we are trying to provide a way for users to easily manage their growing `.tasks` index file. If yes, then introducing a cleanup parameter is as good as explicitly  calling DELETE API on a task_id, as it requires extra manual step to do it. In fact existing DELETE api is a better way to do it, with the reasons @rishabhmaurya mentioned.

If we really want to solve this, a better way might be to introduce a dynamic cluster type setting where user can define their own time interval, and retain the task results until a certain point after it got completed or fetched. By default it would be disabled i.e. the current behavior.

cwperks:
> @cwperks can we support POST on this API? TTL based approach might be a breaking change for users as polling for async task beyond this TTL would result in empty response. Ideally, we can start with a high TTL, say 15 or 30 days as default and also have this POST api with cleanup param. wdyt?

We can make sure the cleanup would only clean completed tasks and make sure they don't linger in perpetuity.

I like @sgup432 suggestion of a separate DELETE _task API as well. This actually doesn't exist.

See https://docs.opensearch.org/latest/api-reference/tasks/tasks/ for the list of Tasks APIs.

sgup432:
>I like @sgup432 suggestion of a separate DELETE _task API as well. This actually doesn't exist.

Oh wow! I am surprised that we don't have any way to delete a task_id!? We certainly need this then.

Actually considering `.tasks` is a system index, can we not use the document API to delete a task_id using `DELETE /.tasks/_doc/<task_id>`? @cwperks

cwperks:
> Actually considering .tasks is a system index, can we not use the document API to delete a task_id using DELETE /.tasks/_doc/<task_id>? @cwperks

Not in a cluster with FGAC. Any mutating operation to a system index is blocked when using FGAC as writing to a system index can cause corruption and any writing should be done through dedicated APIs where the system is the one doing system index access.

I pivoted this PR to reveal a DELETE _task/{task_id} API to delete from this index and it properly guards to ensure that in-flight tasks cannot be deleted from the index.

cwperks:
@rishabhmaurya @reta would I be able to get reviews on this PR?

reta:
> @rishabhmaurya @reta would I be able to get reviews on this PR?

Sorry @cwperks , didn't get to it over the weekend, will try to look this week

server/src/main/java/org/opensearch/action/admin/cluster/node/tasks/delete/TransportDeleteTaskAction.java
```diff
@@ -0,0 +1,181 @@
...
+                return;
+            }
+            deleteTaskResult(response.getId(), thisTask, listener);
+        }
+    }
+
+    private void deleteTaskResult(String taskResultId, Task thisTask, ActionListener<AcknowledgedResponse> listener) {
+        DeleteRequest delete = new DeleteRequest(TaskResultsService.TASK_INDEX, taskResultId);
+        delete.setParentTask(clusterService.localNode().getId(), thisTask.getId());
+        client.delete(delete, ActionListener.wrap(response -> onDeleteTaskResult(taskResultId, response, listener), listener::onFailure));
```
reta, line 232: I think we discussed that briefly on other pill request but tasks have parent / child relationships (Task::parentTask). If we delete parent task but keep its dependent tasks, that would cause problem down the line, we should probably validate that before, wdyt?

reta:
@cwperks the change is looking good overall, thank you, the only concern I have is dangling parent tasks which I believe we should think about, thanks!

server/src/main/java/org/opensearch/action/admin/cluster/node/tasks/delete/TransportDeleteTaskAction.java
cwperks: Addressed in the latest commit. I think that illustrates exactly why putting this behind an API makes sense as in prior versions of OpenSearch you could delete from `.tasks` index directly.

reta (approved):
LGTM, thanks @cwperks !

cwperks:
Added 2 lines in test to refresh post deletion to fix the flaky test. Preferring this over setting RefreshPolicy.IMMEDIATE on the delete operation as this can wait til normal refresh.

PR #21731, opened 2026-05-19
[analytics-engine] wire PPL TAKE / FIRST / LAST / LIST / VALUES end-to-end

sandeshkr419:
## Summary

  Wires five PPL state-expanding aggregates through the analytics engine with correct multi-shard behavior:

  - `take(field, n)` — bounded array of first N values
  - `first(field)` / `last(field)` — first/last value per group
  - `list(field)` / `values(field)` — array of values (values = distinct)

  All work both single-shard and across shards.

  ## Commits

  1. **SPI** (`analytics-framework`): five new `AggregateFunction` enum entries plus `IntermediateTypeResolver` so the planner can derive the FINAL exchange-column type from the actual call's arg type rather than a constant.

  2. **Backend** (`analytics-backend-datafusion`):
     - Rust UDAF for `take(field, n)` with bounded `List<elem>` state and coordinator-side merge in `update_batch` (DataFusion's substrait consumer ignores `AggregationPhase`, so FINAL is also a single-pass aggregate that needs to un-nest the per-shard state).
     - Substrait extensions (`take`, `first_value`, `last_value`, `array_agg`). `take`'s 2-arg shape uses `any1`/`any2` so isthmus's wildcard binding accepts mismatched arg types.
     - `LOCAL_*_OP` stubs + a pre-emit RelShuttle that rebinds PPL aggregations onto the stubs so isthmus resolves them through our extension catalog.
     - `SubstraitPlanRewriter.visit(Aggregate)` inlines literal Project columns as Substrait `Expression.Literal` on aggregate-call args (Calcite materializes constant aggregate-args as `$f1` columns).

  3. **Planner** (`analytics-engine`):
     - Move `AggregateCallAnnotation` out of `AggregateCall.rexList` onto an `OpenSearchAggregate` side-map. Calcite's `AggCallBinding.preOperands` would otherwise corrupt return-type inference for PPL's `ARG0_ARRAY`.
     - `DistributedAggregateRewriter` drives intermediate-field types through `IntermediateTypeResolver` and pins the FINAL aggCall return type to the StageInputScan column for state-expanding aggregates.
     - Cross-shard literal-arg forwarding for `take`'s N: `OpenSearchAggregateSplitRule` captures literal aggregate-args from the original Project; the rewriter re-creates them as a constant Project below FINAL so the Rust accumulator preserves N.

  4. **Cross-shard LIST / VALUES** (`analytics-engine`):
     - PPL `STRING_ARRAY` return type forces `ARRAY<VARCHAR>` regardless of input element. `OpenSearchAggregateSplitRule` repairs the PARTIAL aggCall return type to `ARRAY<actual-arg0>` (PARTIAL only — FINAL keeps the original to satisfy Volcano's parent row-type check).
     - DataFusion's substrait consumer ignores `AggregationPhase`, so a FINAL `array_agg(state)` re-wraps each shard's list rather than concatenating. Custom Rust UDAFs `list_merge` / `list_merge_distinct` un-nest the per-shard state. The convertor routes LIST/VALUES at FINAL (arg0 is already a list type) to the merge UDAFs; PARTIAL keeps `array_agg`.

  ## Test plan

  - [x] Rust unit tests for `take` and `list_merge` accumulators (PARTIAL, FINAL
        un-nest, distinct dedup, null handling).
  - [x] `CoordinatorReduceIT` — single-shard + cross-shard for each of the five
        aggregates. **18/18 pass, 0 deferred.**
  - [x] Full `:sandbox:qa:analytics-engine-rest:integTest` green.
  - [x] Existing planner unit tests updated for shape changes from the
        annotation side-map move.

  ## Notes

  The IT cluster runs with `-da:org.apache.calcite...` to silence Calcite's
  `Aggregate.typeMatchesInferred` debug assertion. The wire schema is
  derived in Rust from the producer's substrait plan (post #21690), not from
  Calcite types, so this assertion is architecturally redundant for our setup
  — enforced at runtime by Rust's `derive_schema_from_partial_plan` + Arrow's
  `ensure_field_compatibility`. Production runs without `-ea`.

### Check List


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
```diff
@@ -72,6 +79,62 @@ public Optional<Rel> visit(io.substrait.relation.Filter filter, EmptyVisitationC
                     .build()
             );
         }
+
```
expani, line 75: Why does this Class exists ? Might have missed it from previous additions

We should be doing Plan Rewrites in Analytics-Engine and BackendPlanAdapter was meant for this.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DistributedAggregateRewriter.java
```diff
@@ -13,55 +13,34 @@
...
- * no recursive tree walks.
+ * Post-Volcano distributed-aggregate decomposition for FINAL aggregate calls. Classifies
+ * each call via {@link AggregateFunction#intermediateFields} into pass-through,
+ * function-swap (e.g. COUNT→SUM), or engine-native merge (reducer == self), and rebuilds
+ * its argList, function, and return type accordingly. Sole owner of
+ * {@link AggregateCall#create} on the FINAL side.
  *
  * @opensearch.internal
  */
 final class DistributedAggregateRewriter {
```
expani, line 40: We should defintely Revisit the need for this class. 

Either we don't use AggregateFunctionDecomposition of Calcite OR we don't rewrite the aggregates here. 

I think this also ties back to moving DAGBuilder down the pipeline after everything since we are rewriting the StageScanInput here again

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
```diff
@@ -377,6 +482,111 @@ private byte[] convertToSubstrait(RelNode fragment) {
...
+     * The {@code isDistinct} flag is preserved <i>except</i> for VALUES, which forces
+     * {@code isDistinct=true} so isthmus emits {@code INVOCATION_DISTINCT} and
+     * DataFusion routes to {@code array_agg(DISTINCT)} server-side.
+     *
+     * <p>Mirrors the pattern of {@link UntypedNullPreprocessor#rewrite(RelNode)} and
+     * {@link DatetimeOutputCastRewriter#rewrite(RelNode)}; called from
+     * {@link #convertToSubstrait(RelNode)} and {@link #convertStandalone(RelNode)}
+     * just before the isthmus visitor runs.
+     */
+    private static RelNode rewritePplAggregateCalls(RelNode root) {
```
expani: This is purely RelNode based and should be happening via BackendPlanAdapter.

Also, let's use switch statements and Set<SqlAggFunction> for comparisons.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DistributedAggregateRewriter.java
sandeshkr419: Ack, will keep the restructure as a follow-up.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
sandeshkr419: Removed the substrait rewriter changes for now. Will handle removing this entirely in a follow-up.

sandeshkr419:
@expani Thanks for review, addressed your comments.

mch2 (approved):

PR #21732, opened 2026-05-19
Native allocator: dynamic settings, query/datafusion pools, plugin nodeStats

gaurav-amz:
## Description

Builds on #21703 to add the query, datafusion with allocator, dynamic-tuning, stats, and pool-wiring pieces.

## Architecture: three native-memory trackers, three knobs

Each tracker owns the bytes it can actually see, with a process-level cap above all of them.

```
            ┌──────────────────────────────────────────────────────────┐
            │  Process-level cap (admission control)                   │
            │                                                          │
            │  node.native_memory.limit                                │
            │    Defaults to 79% × (RAM − JVM heap), cgroup-aware.     │
            │                                                          │
            │  ┌────────────────────────┐  ┌────────────────────────┐  │
            │  │ Arrow tracker          │  │ DataFusion tracker     │  │
            │  │ (ArrowNativeAllocator. │  │ (Rust MemoryPool)      │  │
            │  │  root)                 │  │                        │  │
            │  │                        │  │ Tracks:                │  │
            │  │ Tracks:                │  │   HashAggregate        │  │
            │  │   POOL_FLIGHT (5% NM)  │  │   Sort buffers         │  │
            │  │   POOL_INGEST (8% NM)  │  │   TopK heap            │  │
            │  │   POOL_QUERY  (5% NM)  │  │   Spill staging        │  │
            │  │                        │  │                        │  │
            │  │ Bounded by:            │  │ Bounded by:            │  │
            │  │  native.allocator      │  │  datafusion.memory_    │  │
            │  │  .root.limit           │  │  pool_limit_bytes      │  │
            │  │  (20% × NM by default) │  │  (75% × NM by default) │  │
            │  └────────────────────────┘  └────────────────────────┘  │
            │                                                          │
            └──────────────────────────────────────────────────────────┘
```

Three layers, three responsibilities. Each is necessary; none can be replaced by another.

- **Arrow tracker** (`ArrowNativeAllocator`) accounts for every Arrow `BufferAllocator` allocation in the JVM. Partitioned into FLIGHT / INGEST / QUERY pools so one plugin can't starve another. All Arrow buffers descend from the same `RootAllocator`, preserving Arrow's same-root invariant for cross-plugin zero-copy handoff.
- **DataFusion tracker** (Rust `MemoryPool`) accounts for DataFusion's own working memory and triggers spill / fail-fast when a query exceeds budget. Lives entirely on the Rust side; updates flow in via FFM through `NativeBridge.setMemoryPoolLimit`.
- **Admission control** caps the OS process. `node.native_memory.limit` is the operator-declared off-heap budget AC throttles against; framework-derived defaults (root, pools, DataFusion pool) all scale from this single number.

`POOL_DATAFUSION` is intentionally not present. DataFusion's working memory is allocated by Rust operators directly and reported only to DataFusion's own `MemoryPool` — it never flows through Arrow's `BufferAllocator` API. Adding a Java-side pool that pretended to track it would have required either a per-allocation FFM round-trip (performance disaster) or a config-only mirror (a setting that returns HTTP 200 and silently does nothing). The Rust-side `datafusion.memory_pool_limit_bytes` is the honest knob for that layer.

## Worked example: 64 GB / 16 GB-heap node

With operator-declared `-Xmx16g` on a 64 GB host (bare metal or cgroup-limited container), defaults compose to:

```
64 GB RAM
  │ ├── JVM Heap (16 GB / 25%)
  │ │   ├── Indexing pressure, Writer buffer
  │ │   ├── Transport Netty
  │ │   └── etc.
  │ │
  │ └── Off-Heap (48 GB / 75%)
  │     │
  │     ├── node.native_memory.limit (37.92 GB / 79% of off-heap)
  │     │   │
  │     │   ├── Java Arrow Root (7.58 GB / 20% of NM)
  │     │   │   ├── POOL_FLIGHT  (1.90 GB / 5% of NM)
  │     │   │   │   └── server, client — flight transport batches
  │     │   │   ├── POOL_INGEST  (3.03 GB / 8% of NM)
  │     │   │   │   └── ArrowBufferPool / parquet VSR allocators
  │     │   │   └── POOL_QUERY   (1.90 GB / 5% of NM)
  │     │   │       └── analytics-search-service → per-query children
  │     │   │
  │     │   └── DataFusion Rust pool (28.44 GB / 75% of NM)
  │     │       └── datafusion.memory_pool_limit_bytes
  │     │           ├── hash tables, sort buffers, aggs, intermediate batches
  │     │           └── triggers spill on exhaustion
  │     │
  │     └── Unmanaged (~10.08 GB / 21% of off-heap)
  │         ├── Lucene mmap
  │         ├── OS Page Cache
  │         └── Sidecars / non-Arrow native consumers
  │
  │ Independent (disk):
  │   datafusion.spill_memory_limit_bytes (32 GB / 50% of physical RAM)
```

As a flat table for quick reference:

```
RAM                                              64 GB
JVM heap (operator-configured -Xmx)              16 GB
Off-heap (RAM − heap)                            48 GB
node.native_memory.limit  (79% of off-heap)      37.92 GB    ← AC throttle threshold
  native.allocator.root.limit  (20% of NM)        7.58 GB    ← Arrow framework cap
    pool.flight.max  (5% of NM)                   1.90 GB    ← Flight transport pool
    pool.ingest.max  (8% of NM)                   3.03 GB    ← Parquet VSR pool
    pool.query.max   (5% of NM)                   1.90 GB    ← analytics-engine pool
  datafusion.memory_pool_limit  (75% of NM)      28.44 GB    ← Rust runtime pool (sibling)
Unmanaged (Lucene mmap, OS page cache, etc.)     10.08 GB
```

Independent budget (disk staging, not memory):
```
datafusion.spill_memory_limit_bytes (50% of physical RAM)    32 GB
```

## How a query flows through these layers

Take a concrete example: a user issues a PPL query that goes through analytics-engine and dispatches to DataFusion.

```
  1. Coordinator receives the request.
     → Admission control checks node.native_memory.limit budget.
     → If OK, proceeds; otherwise rejects with 429.

  2. AnalyticsSearchService creates a per-fragment Arrow allocator:
        allocator = getPoolAllocator(POOL_QUERY).newChildAllocator("frag-N", 0, Long.MAX_VALUE)
     → Future BufferAllocator.buffer(...) calls on this allocator
       increment POOL_QUERY's counter and the root counter via Arrow's
       parent-cap chain at allocateBytes.
     → Bounded by native.allocator.pool.query.max.

  3. AnalyticsSearchService dispatches to DataFusion via NativeBridge.
     → NativeBridge.executeQueryAsync(...) marshals plan bytes via FFM.

  4. DataFusion runs the query in Rust:
     → HashAggregate builds a hash table.
     → reservation.try_grow(50MB) → DataFusion MemoryPool counter += 50MB.
     → Bounded by datafusion.memory_pool_limit_bytes.
     → If exceeded, HashAggregate spills to disk
       (which itself is bounded by datafusion.spill_memory_limit_bytes).

  5. DataFusion produces a result batch in Rust.
     → Java imports it via Arrow C Data Interface.
     → The import allocates Java ArrowBufs under the per-fragment allocator
       from step 2. POOL_QUERY counter += result_size.

  6. Result returns to coordinator.
     → Per-fragment allocator closes; POOL_QUERY counter decrements.
     → DataFusion query completes; MemoryPool counter decrements.
```

Each layer accounts for what it owns. No double-counting (the result bytes are counted in DataFusion's pool only while they exist as Rust buffers; after Java imports them, ownership transfers and the Java side counts them; when Java closes the import, the bytes are freed). Each operator-tunable knob bounds a real, observable thing.

## Operator surface

After this PR, an operator inspecting a node running the query above sees:

```json
{
  "native_allocator": {
    "root": {"allocated": "150MB", "limit": "7.58GB"},
    "pools": {
      "flight":  {"allocated": "20MB",  "limit": "1.90GB"},
      "ingest":  {"allocated": "30MB",  "limit": "3.03GB"},
      "query":   {"allocated": "100MB", "limit": "1.90GB"}
    }
  },
  "datafusion": {
    "memory_pool": {"usage": "2.4GB", "limit": "28.44GB"},
    "spill":       {"usage": "0",     "limit": "32GB"}
  }
}
```

Three numbers, three sources, three knobs. The operator looks at each in isolation when tuning that layer:

- "DataFusion queries are OOMing" → bump `datafusion.memory_pool_limit_bytes`.
- "Flight RPC is starving Parquet ingest" → bump `parquet.native.pool.flight.max` or lower `parquet.native.pool.ingest.max`.
- "Whole node is using too much memory" → lower `node.native_memory.limit`.

All pool min/max settings are `Setting.Property.Dynamic`. Pool limit changes propagate to consumer-side child allocators automatically via Arrow's parent-cap check at `allocateBytes` — child allocators are created with `Long.MAX_VALUE` and inherit the live parent cap on every allocation, so dynamic resizes reach in-flight workloads without restart and without an explicit notification SPI. The grouped validator rejects cross-setting violations (sum of pool mins > root, per-pool min > max) at PUT time with HTTP 400 rather than at the next allocation.

## Behavior change to call out

**Admission control is now active by default.** `node.native_memory.limit` defaults to 79% × (RAM − JVM heap) instead of `0` (unconfigured). On upgrade, AC will start tracking native memory utilization and the framework's pool caps will derive sensible values from the operator's declared off-heap budget without any explicit configuration.

Operators who want pre-existing opt-out behavior (AC unconfigured, all framework caps unbounded) can set:

```
node.native_memory.limit: 0b
```

This restores the prior "explicit-opt-in" semantics — useful for nodes where Lucene mmap or non-Arrow native consumers dominate and the operator does not want admission control throttling search/index traffic.

## Changes after initial review

- **Drop `NativeAllocatorListener` SPI.** The SPI was emulating Arrow's native parent-cap check at `allocateBytes`. With consumer-side child allocators created at `Long.MAX_VALUE`, dynamic pool resizes propagate automatically through Arrow's `Accountant.allocate` parent walk — no listener machinery needed. Resolves @bowenlan-amzn (https://github.com/opensearch-project/OpenSearch/pull/21732#discussion_r3269655712, https://github.com/opensearch-project/OpenSearch/pull/21732#discussion_r3269660170) and @Bukhtawar (https://github.com/opensearch-project/OpenSearch/pull/21732#discussion_r3267186318).
- **Fix Guice duplicate-binding regression.** Removed the redundant `ArrowBasePlugin#createGuiceModules` override; `Node.java` already auto-binds every component returned from `createComponents`. This was the root cause of the `gradle-check` Jenkins CI failures across multiple SHAs.
- **`@ExperimentalApi` on `Plugin#nodeStats()`** to match the annotation already on `PluginNodeStats`. Resolves @bowenlan-amzn (https://github.com/opensearch-project/OpenSearch/pull/21732#discussion_r3269563833).
- **FQN → import in `ParquetIndexingEngine`** (3 constructor parameters). Resolves @bowenlan-amzn (https://github.com/opensearch-project/OpenSearch/pull/21732#discussion_r3269589758).
- **Default tuning**: `node.native_memory.limit` derives from RAM−heap (79%); `root.limit` is 20% of NM (Arrow gets a small fraction; DataFusion gets the larger 75% as a sibling); pool maxes anchor to NM at 5%/8%/5%; `datafusion.memory_pool_limit_bytes` is 75% of NM (replaces the prior JVM-heap-derived default flagged by @bharath-techie at https://github.com/opensearch-project/OpenSearch/pull/21732#discussion_r3271093086); `datafusion.spill_memory_limit_bytes` is 50% of physical RAM (independent disk-staging budget).
- **3 boundary integration tests** in `NativeAllocatorBoundaryIT` (`plugins/arrow-flight-rpc/src/internalClusterTest`) boot a real cluster with tight memory settings and verify cap enforcement end-to-end through actual Arrow allocations: per-pool max rejection, root-level rejection across pools, and dynamic pool-resize propagation to in-flight `Long.MAX_VALUE` children.


libs/arrow-spi/src/main/java/org/opensearch/arrow/spi/NativeAllocatorListener.java
```diff
@@ -0,0 +1,39 @@
...
+public interface NativeAllocatorListener {
+
+    /**
+     * Invoked after a pool's limit has been updated.
+     *
+     * @param poolName logical pool name
+     * @param newLimit the new limit in bytes
+     */
+    void onPoolLimitChanged(String poolName, long newLimit);
+}
```
Bukhtawar: Can we not use ClusterSetting's update consumer instead.

Bukhtawar (commented):
We might need to reject the request, throw OpenSearchRejectedException rather than oom-ing. Also lets try to see how we can wire up circuit-breaker for this. Maybe circuit-breaker stats is something we leverage for tracking memory used

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransport.java
```diff
@@ -147,9 +151,27 @@ public FlightTransport(
     protected void doStart() {
         boolean success = false;
         try {
-            flightAllocator = allocatorService.newChildAllocator("flight", Integer.MAX_VALUE);
+            // Use the unified native allocator's flight pool so flight allocations are tracked
+            // and capped by the framework alongside ingest, query, and datafusion. Hard-fail if
+            // the framework plugin is missing — silently falling back to a separate root would
+            // break the same-root invariant for cross-plugin Arrow handoff.
+            ArrowNativeAllocator nativeAllocator = ArrowNativeAllocator.instance();
```
bowenlan-amzn: Please change the way of providing ArrowNativeAllocator.

See ArrowAllocatorService as the example.

Initialize: https://github.com/opensearch-project/OpenSearch/blob/10f3617e1e2161d90dd4aea6ac983de871876fd2/plugins/arrow-base/src/main/java/org/opensearch/arrow/allocator/ArrowBasePlugin.java#L207-L208
And return in createComponents, so will be part of GUICE

How plugin get it: 
From plugin's createComponents
https://github.com/opensearch-project/OpenSearch/blob/f121d1e8ca90817b95f5e4969b0f1bc006629049/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightStreamPlugin.java#L106-L107

Or through GUICE
https://github.com/opensearch-project/OpenSearch/blob/efe121068efddacbe219f8e892d6b55db58bfc41/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java#L93

And we can remove the ArrowAllocatorService using this PR, as it's for providing the root allocator to other extension plugin. And ArrowNativeAllocator should be the one doing that now, through this `getPoolAllocator`.

bowenlan-amzn:
This PR seems to be the good place to get some alignment on the memory assignment across different components in the system.
Let's take 64G RAM as example.

```
64 GB RAM
├── JVM Heap (16 GB / 25%)
│   ├── Indexing pressure, Writer buffer
│   ├── Transport Netty
│   └── etc. ...
|
└── Off-Heap (48 GB / 75%)
    │
    ├── node.native_memory.limit (40 G)
    │   │
    │   ├── Java Arrow Root (12G / 30%)
    │   │   ├── POOL_FLIGHT  (2G / 5%)
    │   │   │   └── flight/server, flight/client — batch ser/deser
    │   │   ├── POOL_INGEST  (4G / 10%)
    │   │   │   └── ArrowBufferPool children (per-VSR, pool_limit/divisor)
    │   │   └── POOL_QUERY   (2G / 5%)
    │   │       └── analytics-search-service → per-query children (256MB each)
    │   │
    │   └── Rust Memory Pool (28G / 70%)
    │       └── DataFusion Rust pool (datafusion.memory_pool_limit_bytes)
    │           ├── hash tables, sort buffers, aggs, intermediate batches
    │           └── spill staging (datafusion.spill_memory_limit_bytes)
    │
    └── Rest (no knobs, at least 8G)
        ├── Lucene mmap (~2 GB)
        └── Page Cache (whatever's left)
```

I think the Java Arrow memory are mostly used for intermediate data transfer, while Rust memory are used for query execution. That's one reason to provide more to Rust side.

@Bukhtawar @alchemist51 @mch2 @expani

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransport.java
gaurav-amz: Thanks for point this out, will do this change.

libs/arrow-spi/src/main/java/org/opensearch/arrow/spi/NativeAllocatorListener.java
gaurav-amz: After the FlightTransport / AnalyticsSearchService / DefaultPlanExecutor migration, the SPI had zero production callers.

server/src/main/java/org/opensearch/plugins/Plugin.java
```diff
@@ -249,6 +249,20 @@ public List<Setting<?>> getSettings() {
...
+     * Returns plugin-contributed node statistics that surface under {@code _nodes/stats}.
+     * Each entry renders at top-level under {@code nodes.<id>.<getWriteableName()>}.
+     *
+     * <p>Plugins that override this method must also register the concrete
+     * {@link PluginNodeStats} subclass via {@link #getNamedWriteables()} so the
+     * coordinator can deserialize per-node payloads received over transport.
+     *
+     * <p>Default: empty.
+     */
+    public List<PluginNodeStats> nodeStats() {
```
bowenlan-amzn, line 266: I hope we can have a ExperimentalApi annotation here.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
```diff
@@ -102,7 +103,8 @@ public ParquetIndexingEngine(
         Supplier<Schema> schemaSupplier,
         Supplier<Long> mappingVersionSupplier,
         IndexSettings indexSettings,
-        ThreadPool threadPool
+        ThreadPool threadPool,
+        org.opensearch.arrow.allocator.ArrowNativeAllocator nativeAllocator
```
bowenlan-amzn: nit: use import. And all the other places

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransport.java
```diff
@@ -147,9 +150,26 @@ public FlightTransport(
     protected void doStart() {
         boolean success = false;
         try {
-            flightAllocator = allocatorService.newChildAllocator("flight", Integer.MAX_VALUE);
+            // Use the unified native allocator's flight pool so flight allocations are tracked
+            // and capped by the framework alongside ingest, query, and datafusion. Hard-fail if
+            // the framework plugin is missing — silently falling back to a separate root would
+            // break the same-root invariant for cross-plugin Arrow handoff.
+            BufferAllocator flightPool = nativeAllocator.getPoolAllocator(NativeAllocatorPoolConfig.POOL_FLIGHT);
+            flightAllocator = flightPool.newChildAllocator("flight", 0, flightPool.getLimit());
```
bowenlan-amzn: It seems not necessary to create flightAllocator here, the flightPool should be the flightAllocator.

For the 2 child allocator, server and client, can we set the limit to be `Long.MAX_VALUE`. I think they will still be bound by their parent which is the POOL_FLIGHT here.

This way seems we also don't need the listener setup here.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -61,33 +64,57 @@ public class AnalyticsSearchService implements AutoCloseable {
...
         this.backends = backends;
         this.listener = new AnalyticsOperationListener.CompositeListener(listeners);
-        this.allocator = allocatorService.newChildAllocator("analytics-search-service", Long.MAX_VALUE);
+        this.nativeAllocator = nativeAllocator;
+        // Source the service-level allocator from the unified framework's query pool so all
+        // analytics-engine allocations are tracked and capped by the framework. Hard-fail if
+        // the framework is missing — silently falling back to a separate root would break
+        // Arrow's same-root invariant for cross-plugin handoff.
+        BufferAllocator queryPool = nativeAllocator.getPoolAllocator(NativeAllocatorPoolConfig.POOL_QUERY);
+        this.allocator = queryPool.newChildAllocator("analytics-search-service", 0, queryPool.getLimit());
```
bowenlan-amzn: Same as this comment in FlightTransport https://github.com/opensearch-project/OpenSearch/pull/21732/changes#r3269631251

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -101,10 +109,28 @@ public DefaultPlanExecutor(
...
         this.scheduler = scheduler;
-        this.coordinatorAllocator = allocatorService.newChildAllocator("coordinator", Long.MAX_VALUE);
+        this.nativeAllocator = nativeAllocator;
+        // Source the coordinator allocator from the framework's QUERY pool so coordinator-side
+        // allocations (held shard responses, intermediate batches during reduce) count against
+        // parquet.native.pool.query.max alongside the data-node-side per-fragment allocators in
+        // AnalyticsSearchService. Before this, coordinator allocations were root-siblings outside
+        // any pool and invisible to _nodes/stats.native_allocator.pools.query.allocated.
+        BufferAllocator queryPool = nativeAllocator.getPoolAllocator(NativeAllocatorPoolConfig.POOL_QUERY);
+        this.coordinatorAllocator = queryPool.newChildAllocator("coordinator", 0, queryPool.getLimit());
```
bowenlan-amzn: Same as this comment in FlightTransport https://github.com/opensearch-project/OpenSearch/pull/21732/changes#r3269631251

bowenlan-amzn (approved):
Overral looks good! Thanks for refactoring the allocator related changes in arrow-base.
Left some comments that should be easy to accommodate. Approving now.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -73,12 +73,21 @@ public class DataFusionPlugin extends Plugin implements SearchBackEndPlugin<Data
...
+     * <p>
+     * Dynamic only when the loaded native library exports {@code df_set_spill_limit}
+     * (see {@link org.opensearch.be.datafusion.nativelib.NativeBridge#isSpillLimitDynamic()}).
+     * When the symbol is absent the setting can still be updated at the cluster level,
+     * but the new value only takes effect after a node restart — the live update consumer
+     * logs a warning in that case.
+     */
     public static final Setting<Long> DATAFUSION_SPILL_MEMORY_LIMIT = Setting.longSetting(
         "datafusion.spill_memory_limit_bytes",
         Runtime.getRuntime().maxMemory() / 8,
```
bharath-techie: @mch2 found that this is JVM heap, can we do total ram / 8 as default ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
bharath-techie: Same here as https://github.com/opensearch-project/OpenSearch/pull/21732/changes#r3271093086

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransport.java
gaurav-amz: Done

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransport.java
gaurav-amz: Done. The intermediate flightAllocator is gone, flightPool is the parent directly, and serverAllocator / clientAllocator are children at Long.MAX_VALUE.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
gaurav-amz: Same shape applied.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
gaurav-amz: Done

server/src/main/java/org/opensearch/plugins/Plugin.java
gaurav-amz: Done

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/engine/ParquetIndexingEngine.java
gaurav-amz: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
gaurav-amz: datafusion.memory_pool_limit_bytes now defaults to 75% × node.native_memory.limit
(the off-heap budget, cgroup-aware via OsProbe).

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
gaurav-amz: Fixed, Now the spill will be 50% of RAM.

gaurav-amz:
> We might need to reject the request, throw OpenSearchRejectedException rather than oom-ing. Also lets try to see how we can wire up circuit-breaker for this. Maybe circuit-breaker stats is something we leverage for tracking memory used

OpenSearchRejectedException handling has been addressed in this PR. Circuit breaker-related changes will be included in a follow-up PR. Stats-related refactoring will be needed once Ajay’s PR is merged.

Bukhtawar (approved):

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetSettings.java
```diff
@@ -105,6 +105,25 @@ private ParquetSettings() {}
...
+     * to 1/10 of the pool"; a divisor of 2 means "each VSR may use up to 1/2".
+     */
+    public static final Setting<Integer> MAX_PER_VSR_ALLOCATION_DIVISOR = Setting.intSetting(
+        "parquet.max_per_vsr_allocation_divisor",
+        10,
+        1,
+        100,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
```
Bukhtawar, line 125: Should this be function of cores?

server/src/main/java/org/opensearch/action/admin/cluster/node/stats/NodeStats.java
```diff
@@ -252,6 +261,51 @@ public NodeStats(StreamInput in) throws IOException {
...
+                // entry. Drop the whole map. This branch is defensive; production paths
+                // always set a registry when they expect named writeables.
+                continue;
+            }
+            try (
+                StreamInput rawIn = payload.streamInput();
+                NamedWriteableAwareStreamInput payloadIn = new NamedWriteableAwareStreamInput(rawIn, registry)
+            ) {
+                payloadIn.setVersion(in.getVersion());
+                PluginNodeStats stats = payloadIn.readNamedWriteable(PluginNodeStats.class);
```
Bukhtawar, line 325: Will `_node/stats?metrics=<metric_name>` work?

gaurav-amz: Yes, The stats will be under this path `_nodes/stats/native_allocator`

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetSettings.java
gaurav-amz: Will discuss and change this accordingly in follow up PR.

gaurav-amz:
Closing and reopening to trigger the build again.

gaurav-amz:
Follow up PR changes.
1. Native allocator stats under native memory.
2. Refactoring of native allocator stats.

PR #21743, opened 2026-05-19
analytics-engine: More Math function support

mch2:
### Description
Brings PPL scalar function coverage on the analytics-engine route up to parity.

  Arithmetic
  - x % y and mod(x, y) on floating-point operands (fp32, fp64, and mixed fp/int)
  - mod(x, 0) returns NULL instead of raising
  - x / 0 returns NULL instead of raising

  Trigonometric
  - sin, cos, tan, asin, acos, atan, atan2, cot on integer-typed columns (was failing with "Unable to convert call SIN(i32?)" etc.)

  Base conversion
  - conv(n, fromBase, toBase) — convert integer between bases 2–36, matches MySQL/Java semantics (lowercase digit output, NumberFormatException on out-of-range radix)
  - Accepts both string and integer first argument
  - Round-trips negative numbers

  Predicates
  - where rand() > 0 and other non-deterministic filter predicates (was rejected by the planner)

  No new PPL syntax — these are existing PPL functions that were silently failing or throwing on the analytics-engine route. Users gain consistent behavior whether their query hits the analytics path.

### Related Issues

### Check List


vinaykpud (approved):

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
```diff
@@ -548,11 +538,22 @@ public Map<ScalarFunction, ScalarFunctionAdapter> scalarFunctionAdapters() {
...
                     Map.entry(ScalarFunction.MVZIP, new MvzipAdapter()),
                     Map.entry(ScalarFunction.MVAPPEND, new MvappendAdapter()),
+                    // Trigonometric functions: substrait `sin/cos/tan/...` declare fp32/fp64 impls only.
+                    // PPL emits e.g. SIN(integer_field) which Calcite types as SIN(i32?) — isthmus then
+                    // fails signature lookup. NumericToDoubleAdapter widens any numeric operand to fp64
+                    // before substrait conversion (no-op when the operand is already fp32/fp64).
+                    Map.entry(ScalarFunction.ACOS, new NumericToDoubleAdapter(SqlStdOperatorTable.ACOS)),
+                    Map.entry(ScalarFunction.ASIN, new NumericToDoubleAdapter(SqlStdOperatorTable.ASIN)),
+                    Map.entry(ScalarFunction.ATAN, new NumericToDoubleAdapter(SqlStdOperatorTable.ATAN)),
+                    Map.entry(ScalarFunction.ATAN2, new NumericToDoubleAdapter(SqlStdOperatorTable.ATAN2)),
```
vinaykpud, line 583: We will need similar adapters for `RADIANS and DEGREES`
I think we need similar fix for them as well. We need to convert the arguments in those functions to Double as well.

please add
```
Map.entry(ScalarFunction.RADIANS, new NumericToDoubleAdapter(SqlStdOperatorTable.RADIANS)),
Map.entry(ScalarFunction.DEGREES, new NumericToDoubleAdapter(SqlStdOperatorTable.DEGREES)),
```

vinaykpud (approved):

sandeshkr419 (approved):

PR #21744, opened 2026-05-19
Replace oneshot channel with JoinHandle in IndexReader prefetch

himshikhagupta:
### Description
IndexReader used a oneshot channel between spawn_blocking and the poll loop for row group prefetch. When the blocking task panicked (e.g. DelegationPossible reaching subtree_cost in the bitmap tree evaluator), the sender was dropped without sending. The Poll::Ready(Err(_)) arm retried unconditionally, creating an infinite loop that left the Java search thread stuck on future.join() forever.

Replace the oneshot channel with JoinHandle from spawn_blocking, which provides explicit panic detection via JoinError::is_panic():
- Panic: propagate immediately as DataFusionError with the panic message
- Cancelled (runtime shutdown): retry (transient, preserves existing behavior)

This also fixes waker registration: JoinHandle properly wakes the parent task on completion, unlike the oneshot pattern which required the receiver to be polled before it could register a waker.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_table/stream.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_table/stream.rs
@@ -47,7 +47,7 @@ use datafusion::physical_plan::{
 };
 use datafusion_common::DataFusionError;
 use futures::{Future, Stream};
-use tokio::sync::oneshot;
+use tokio::task::JoinHandle;
 
 use super::eval::{PrefetchedRg, RowGroupBitsetSource};
 use super::metrics::StreamMetrics;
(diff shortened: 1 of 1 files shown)


Bukhtawar (commented):
Can we add a test to demonstrate the fix

himshikhagupta:
> Can we add a test to demonstrate the fix

@Bukhtawar  Tests are already added to ensure panics are handled. Are you looking for some specific test?

expani (approved):
Thanks @himshikhagupta 

LGTM JoinHandle is the right way for getting a handle to errors in one-shot channels.

The test timeout of 2 secs seems a bit high. Maybe there is a way to make it less flaky and reduce timeout.

alchemist51 (approved):

Bukhtawar (approved):
LGTM

PR #21747, opened 2026-05-19
[Analytics Backend] Timechart route end-to-end closure (1/30 → 30/30)

ahkcs:
## Summary

End-to-end coverage closure for PPL `timechart` on the analytics-engine route. Five coupled commits in `analytics-backend-datafusion` take `CalciteTimechart{Command,PerFunction}IT` from **1/30 → 30/30 ✅** on `upstream/main`. Each commit corresponds to a distinct Substrait-binding gap surfaced by progressive IT runs against a `:run` cluster with `-Dtests.analytics.force_routing=true -Dtests.analytics.parquet_indices=true`.

### 1. `[Analytics] Wire TIMESTAMPDIFF/TIMESTAMPADD + skip non-prefix-groupSet split`

Two coupled changes:

- Register `TIMESTAMPDIFF` and `TIMESTAMPADD` as scalar functions (enum constants in `ScalarFunction`, added to `STANDARD_PROJECT_OPS`). Clears the capability-layer rejection at `OpenSearchProjectRule.annotateExpr`; isthmus-side substrait conversion is still gated by per-call adapters (commit 2).
- `OpenSearchAggregateSplitRule.onMatch()` skips the PARTIAL/FINAL alternative when it would emit a row type that fails Volcano's `typeMatchesInferred`. Two unsafe shapes today:
    - `percentile_approx` (2-arg aggregate whose FINAL phase needs `(tdigest_state, percent_literal)`)
    - **Cross-family non-prefix groupSet** — when a non-prefix group-key index lands on PARTIAL's agg-output slot whose `SqlTypeFamily` differs from the ORIGINAL input column's family (e.g. PPL timechart's no-`by` form: `groupSet={2}` on a `[@timestamp:TIMESTAMP, cpu_usage:DOUBLE, SPAN(@timestamp):TIMESTAMP]` input where the SUM aggregate's DOUBLE return at position 2 clashes with TIMESTAMP at the same input position).
  Same-family non-prefix cases (e.g. `group={1}` over two INTEGER columns + a NUMERIC agg) still pass through the split unchanged — `PlanShapeTests.testJoinWithDifferentGroupKeys_multiShard` verifies this. The skip lives in `onMatch` (not `matches`) so the SINGLE+SINGLETON alternative is still registered, giving the planner a coordinator-gather path for the skipped shapes.

### 2. `[Analytics Backend] Add TimestampDiffAdapter peephole for timechart per_*`

Peephole that constant-folds `TIMESTAMPDIFF(out_unit, t, TIMESTAMPADD(in_unit, n, t))` to a BIGINT literal when both units are fixed-length (MICROSECOND…WEEK). Eliminates both PPL UDF references in one step — neither has a Substrait extension binding, so the unfolded calls would surface as `Unable to convert call TIMESTAMPADD(...)`. `OperatorAnnotation` wrappers introduced by `OpenSearchProjectRule` are peeled at each operand before structural comparison.

### 3. `[Analytics Backend] UInt64 → Int64 bridge at partition-stream boundary`

Analytics-engine reduce-stage fragments crashed with `Substrait error: Field '<name>' in Substrait schema has a different type (Int64) than the corresponding field in the table schema (UInt64)` when the upstream PARTIAL fragment emitted an Arrow type Substrait/Calcite has no native equivalent for. Canonical case: DataFusion's `row_number()` physical op (output: UInt64) feeding a consumer plan that types `ROW_NUMBER OVER` as BIGINT/Int64.

Extends the existing `schema_coerce::coerce_inferred_schema` bridge — currently applied only at the parquet-scan boundary — to the partition-stream registration boundary used by every inter-stage exchange:

- `api::register_partition_stream` runs `coerce_inferred_schema` on the producer-derived schema before `session.register_partition` and the IPC return path.
- `api::sender_send` adds `coerce_batch_to_sender_schema` — casts incoming `RecordBatch` columns to match the sender's coerced schema via `arrow::compute::cast`. Without this, DataFusion's HashJoin / RepartitionExec strict-downcasts to `Int64Type` and panics with `"primitive array"`.
- `DatafusionReduceSink.typesMatch` relaxed to treat same-width `Int` (any signedness) as compatible, mirroring the existing `Timestamp` precision tolerance.

### 4. `[Analytics Backend] SpanAdapter — multi-unit fixed-length time bucket`

Pre-existing `SpanAdapter` handled time spans only when `interval==1`. Multi-unit spans (`span=2m`, `span=12h`, `span=5d`, ...) fell through unchanged. Rewrite for fixed-length units (s, m, h, d, w):

```
SPAN(t, N, '<unit>')
  → CAST(from_unixtime(CAST((to_unixtime(t) / B) * B AS DOUBLE)) AS <result>)
  where B = N * unit_seconds
```

Three component choices were each forced by a binding-time substrait constraint:

1. **`LOCAL_TO_UNIXTIME_OP` / `LOCAL_FROM_UNIXTIME_OP`** (package-local) instead of Calcite stdlib `UNIX_SECONDS` / `TIMESTAMP_SECONDS`. The latter bind to BigQuery-named substrait functions DF's consumer has no entries for.
2. **No `FLOOR` between divide and multiply.** Integer ÷ integer already truncates (correct for non-negative epoch seconds); `FLOOR(i64)` doesn't bind in substrait (DF's `floor` is floating-point only).
3. **`CAST(... AS DOUBLE)` before `from_unixtime`.** yaml signature is `(fp64) → precision_timestamp<6>`.

Variable-length units (M, q, y) intentionally fall through — bucketing by calendar month isn't `N × fixed_seconds`. Commit 5 handles those.

### 5. `[Analytics Backend] TimestampDiffAdapter — variable-month runtime rewrite`

Extends commit 2's peephole to handle variable-length inner units (MONTH/QUARTER/YEAR) when the out-unit is fixed-length. Folding isn't safe (Feb 2025 = 28 days, Oct 2025 = 31 days — value depends on calendar month). Atomic runtime rewrite:

```
TIMESTAMPDIFF(out_unit, t, TIMESTAMPADD(MONTH|QUARTER|YEAR, n, t))
  →  (to_unixtime(DATETIME_PLUS(t, INTERVAL n*m MONTH)) - to_unixtime(t)) * out_factor
```

where `m` converts QUARTER→3 / YEAR→12, `INTERVAL_YEAR_MONTH` is built via `SqlIntervalQualifier(MONTH)` (same idiom as `EarliestLatestAdapter#makeIntervalAdd`, which has prior wiring proof that `DATETIME_PLUS(t, INTERVAL_MONTH)` binds end-to-end via Substrait's standard `add(timestamp, interval_year_month)` into DataFusion's native interval add).

PPL's `Chart#transformPerFunction` always emits MILLISECOND out-unit, so the multiplier path (×1000) is the hot case; the divisor path (MINUTE→÷60 etc.) covers consumer code shapes that compute coarser durations.

## Pass-rate impact

Verified locally against `upstream/main` + this PR's full stack:

| Test class | Baseline | + c1 (TIMESTAMPDIFF wire) | + c2 (peephole) | + c3 (UInt64 bridge) | + c4 (SpanAdapter multi-unit) | + c5 (variable-month rewrite) |
|---|---|---|---|---|---|---|
| `CalciteTimechartCommandIT` | 1/18 | 1/18 | 5/18 | 18/18 ✅ | 18/18 ✅ | **18/18 ✅** |
| `CalciteTimechartPerFunctionIT` | 0/12 | 0/12 | 1/12 | 1/12 | 10/12 | **12/12 ✅** |
| **Combined** | **1/30** | **1/30** | **6/30** | **19/30** | **28/30** | **30/30 ✅** |

`PlanShapeTests.testJoinWithDifferentGroupKeys_multiShard` (the multi-shard join-with-non-prefix-groupSet plan-shape regression test) and the 12 other `PlanShapeTests` cases also pass — no regression to the existing split rule's coverage.

## Sandbox check status

`./gradlew check -p sandbox -Dsandbox.enabled=true` results on this branch:

- ✅ All analytics-backend-datafusion / analytics-engine / analytics-framework / parquet-data-format / composite-engine unit tests pass (`-Werror` compile, spotless, checkstyle, forbiddenApis, licenseHeaders, dependencyLicenses, thirdPartyAudit all clean).
- ⚠️ Pre-existing failures in `:sandbox:qa:analytics-engine-coordinator:internalClusterTest` and `:sandbox:plugins:composite-engine:internalClusterTest` (9 + 2 tests) all surface the same `Allocator[ROOT] closed with outstanding child allocators` teardown leak. Identical failures reproduce on plain `upstream/main` without any of this PR's commits — confirmed in a separate run. Unrelated to the code paths this PR touches (`SpanAdapter`, `TimestampDiffAdapter`, `OpenSearchAggregateSplitRule`, `register_partition_stream`, `sender_send`, `typesMatch`).

## Test plan

- [x] `CalciteTimechartCommandIT` — 18/18 on analytics-engine route
- [x] `CalciteTimechartPerFunctionIT` — 12/12 on analytics-engine route
- [x] `PlanShapeTests` (analytics-engine) — 13/13, no regression to non-prefix-groupSet handling
- [x] `./gradlew check -p sandbox -Dsandbox.enabled=true` — unit-test surface clean; pre-existing flakes in unrelated `internalClusterTest` modules

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
@@ -559,11 +559,21 @@ pub unsafe fn sql_to_substrait(
 }
 
 /// Lowers a partial-aggregate Substrait plan against a throwaway session and
-/// returns its physical output schema. NamedTable references are resolved
-/// against empty MemTables built from the substrait base_schema — the plan
-/// itself is the source of truth for the producer side, so no view-type or
-/// timestamp-precision rewrites are applied here. The plan is dropped at
+/// returns its physical output schema **narrowed via
+/// [`schema_coerce::coerce_inferred_schema`]** to types Substrait can bind
+/// against. NamedTable references are resolved against empty MemTables built
+/// from the substrait base_schema — the plan itself is the source of truth for
+/// the producer side, so no view-type or timestamp-precision rewrites are
+/// applied here beyond the post-physical coercer. The plan is dropped at
 /// function exit; only the schema is returned.
+///
+/// <p>The coercer flips Arrow-only types (notably `UInt64` from DataFusion's
+/// `row_number()` physical op) back to their Substrait-compatible counterparts
+/// (`Int64` matching isthmus's `ROW_NUMBER OVER` declaration). The producer
+/// side runs the same coercer + wraps its physical plan with
+/// [`crate::relabel_exec::RelabelExec`] (zero-copy bit-tag flip per mismatched
+/// column), so runtime batches arrive with the same type tags the consumer
+/// registers here — no per-batch cast needed at the partition-stream feed.
 fn derive_schema_from_partial_plan(
     substrait_bytes: &[u8],
 ) -> Result<arrow::datatypes::SchemaRef, DataFusionError> {
@@ -641,7 +651,7 @@ fn derive_schema_from_partial_plan(
 
     let logical_plan = futures::executor::block_on(from_substrait_plan(&session_state, &plan))?;
     let physical_plan = futures::executor::block_on(session_state.create_physical_plan(&logical_plan))?;
-    Ok(physical_plan.schema())
+    Ok(crate::schema_coerce::coerce_inferred_schema(physical_plan.schema()))
 }
 
 /// Encodes a Schema as Arrow IPC stream-format bytes (a schema-only message
@@ -794,6 +804,12 @@ pub unsafe fn register_partition_stream(
     partial_plan_bytes: &[u8],
 ) -> Result<(i64, Vec<u8>), DataFusionError> {
     let session = &mut *(session_ptr as *mut LocalSession);
+    // derive_schema_from_partial_plan applies `schema_coerce::coerce_inferred_schema`
+    // to the physical plan's output schema, matching what isthmus declared on the wire
+    // for the producer (e.g. `Int64` for `ROW_NUMBER OVER`). The producer side runs the
+    // same coercer + wraps its physical plan with `RelabelExec`, so the batches arriving
+    // here through the partition channel are already typed to match this schema — no
+    // per-batch cast at feed time.
     let schema = derive_schema_from_partial_plan(partial_plan_bytes)?;
     let schema_ipc = schema_to_ipc_bytes(schema.as_ref())?;
     let sender = session.register_partition(input_id, schema)?;
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScalarFunction.java
@@ -228,6 +228,24 @@ public enum ScalarFunction {
     DATE_FORMAT(Category.SCALAR, SqlKind.OTHER_FUNCTION),
     TIME_FORMAT(Category.SCALAR, SqlKind.OTHER_FUNCTION),
     STR_TO_DATE(Category.SCALAR, SqlKind.OTHER_FUNCTION),
+    /**
+     * PPL {@code TIMESTAMPDIFF(unit, start, end)} — emits the integer number of {@code unit}s
+     * between two timestamps. Resolves through the SQL plugin's {@code TimestampDiffFunction}
+     * UDF named {@code "TIMESTAMPDIFF"}. Lowering target for PPL `timechart`'s {@code per_*}
+     * aggregations, which expand to {@code DIVIDE(agg * scale, TIMESTAMPDIFF('MILLISECOND',
+     * @timestamp, TIMESTAMPADD(unit, n, @timestamp)))}. The analytics-backend-datafusion adapter
+     * rewrites the call into a DataFusion-native expression because the PPL UDF itself is
+     * unknown to isthmus's default extension catalog.
+     */
+    TIMESTAMPDIFF(Category.SCALAR, SqlKind.OTHER_FUNCTION),
+    /**
+     * PPL {@code TIMESTAMPADD(unit, value, timestamp)} — emits {@code timestamp + value units}.
+     * Resolves through the SQL plugin's {@code TimestampAddFunction} UDF named
+     * {@code "TIMESTAMPADD"}. Lowering target for PPL `timechart`'s {@code per_*}
+     * aggregations (see {@link #TIMESTAMPDIFF}). The analytics-backend-datafusion adapter
+     * rewrites the call into a DataFusion-native interval-add expression.
+     */
+    TIMESTAMPADD(Category.SCALAR, SqlKind.OTHER_FUNCTION),
 
     // ── JSON ────────────────────────────────────────────────────────
     JSON_APPEND(Category.SCALAR, SqlKind.OTHER_FUNCTION),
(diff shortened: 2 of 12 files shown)


sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -934,9 +942,52 @@ pub unsafe fn sender_send(
...
+
     sender.send_blocking(Ok(batch), io_handle)
 }
 
+/// Cast batch columns to match the sender's declared schema where they differ. No-op when
+/// schemas already match (the common case). Returns an error if a cast is not representable
+/// via Arrow's `cast` kernel — callers should not register a sender whose schema is not
+/// reachable from the batch's via cast (the registration-time `schema_coerce` only narrows
+/// types the cast kernel handles).
+fn coerce_batch_to_sender_schema(
```
sandeshkr419: @ahkcs the row_number() UInt64↔Int64 bridge is real, but `coerce_batch_to_sender_schema` does unnecessary per-row work. UInt64 and Int64 share an identical Arrow buffer layout — the `x as i64` reinterpret produces the same bytes the buffer already holds — so `arrow::compute::cast` does an  O(rows) memcpy + buffer allocation per batch that's observationally a no-op.

Suggesting two coupled changes that drop the per-batch cast entirely:

   **1.** In `derive_schema_from_partial_plan`, return the LogicalPlan schema instead of the physical plan schema:
   ```
   // before:
   let physical = state.create_physical_plan(&logical).await?;
   let schema = physical.schema();
   // after:
   let schema: SchemaRef = Arc::new(logical.schema().as_arrow().clone());
```

   The leaf mirrors keep working (they apply to base_schema before from_substrait_plan and propagate through the LogicalPlan). For row_number(), this returns Int64 directly — drops the coerce_inferred_schema call at register_partition_stream.

   2. Add a thin RelabelExec operator on the producer side. Wraps an ExecutionPlan, exposes a different SchemaRef, retags each batch's columns via zero-copy ArrayData rebuild:

``` 
  let data = src.to_data().into_builder()
       .data_type(dst.clone())
       .build()?;
   make_array(data)
```

   Insert after create_physical_plan on the producer plan-build path when physical schema diverges from logical schema on bit-compatible pairs (Int8↔UInt8 .. Int64↔UInt64). Per-batch cost: 1× Arc clone + 1× ArrayData rebuild per mismatched column, no buffer alloc, no per-row work. Errors on non-bit-compatible mismatches (those should be Cast at the LogicalPlan level instead).

   With both, coerce_batch_to_sender_schema and the typesMatch Int-signedness arm both go away and batches arrive Int64 by construction. Timestamp tolerance in typesMatch stays (pre-existing, handles leaf-level Timestamp(s)→Timestamp(ms), unrelated) - I think they are still okay but **I want to get rid of this schema co-ercion**.

ahkcs: Done in [`cc0b9ab`](https://github.com/opensearch-project/OpenSearch/pull/21747/commits/cc0b9abcc0d).

- `coerce_batch_to_sender_schema` removed.
- New `RelabelExec` does the zero-copy retag via `ArrayData::into_builder().data_type(...).build()` + `make_array(data)`.
- Wrapped at every producer `create_physical_plan` site (`query_executor`, `indexed_executor`, `session_context`, `local_executor`).
- `typesMatch` Int-signedness arm dropped. Timestamp tolerance kept (unrelated, pre-existing).

One adjustment to your point 1: in DF 53, `from_substrait_plan` already maps Substrait `i64` ROW_NUMBER to `UInt64` at the LogicalPlan layer, so `logical.schema()` is `UInt64` and the wrap would be a no-op. I used `schema_coerce::coerce_inferred_schema(physical.schema())` instead — same narrowing as the parquet-scan boundary, no new code, both producer wrap and consumer registration use it so they agree by construction.

sandeshkr419 (approved):
Thanks for making the changes.

PR #21749, opened 2026-05-20
Improve/native memory admission control's precision addresses multiple bug fixes

Bukhtawar:
## Summary

  - Auto-derive effective native memory budget as `totalPhysicalMemory - jvmMaxHeap` when
    `node.native_memory.limit` is not explicitly configured, eliminating the hard requirement
    to set this setting for admission control to function
  - Subtract JVM non-heap committed memory (metaspace + code cache + compressed class space)
    from the native usage calculation so the metric reflects actual native allocations
    (Arrow buffers, jemalloc/DataFusion, direct byte buffers) rather than fixed JVM overhead
  - Downgrade log from WARN to DEBUG when the limit is unconfigured to prevent log spam on
    every poll cycle (default 2s from 500ms)

### Related Issues

```
  #  │                           Commit                            │                                                                                       Changes                                                                                       │
  ├─────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 1   │ d240228a11c Improve native memory admission control         │ Auto-derive effectiveNativeMemory as totalRAM - jvmMaxHeap when node.native_memory.limit is unset. Subtract JVM non-heap (metaspace + code cache) from nativeUsed so metric         │
  │     │ accuracy                                                    │ reflects actual native allocations. Downgrade log spam from WARN to DEBUG when limit is unconfigured.                                                                               │
  ├─────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 2   │ ac164789e31 Add integration tests for native memory         │ 5 ITs: auto-derived limit, explicit limit tracking, monitor mode pass-through, below-threshold success, dynamic limit update.                                                       │
  │     │ admission control                                           │                                                                                                                                                                                     │
  ├─────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 3   │ 6063e625186 Fix integration tests                           │ Use stats-injection assertions instead of transport-rejection (which requires full shard routing). Tests verify utilization values are correctly tracked and observable.            │
  ├─────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 4   │ 2859e30fffd Change native memory polling interval default   │ Reduce polling from 500ms to 2s (/proc/self/status read is unnecessary at sub-second frequency). Keep RssAnon unavailable log at WARN. Fix tests that used windows shorter than 2s. │
  │     │ to 2s                                                       │                                                                                                                                                                                     │
  ├─────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 5   │ 8eeb786ebf8 Close outstanding child allocators on           │ Force-close any remaining child allocators (e.g., coordinator) before closing the root allocator. Fixes IllegalStateException: closed with outstanding child allocators on node     │
  │     │ ArrowNativeAllocator shutdown                               │ shutdown. Also fixed a second 1s window in ClusterInfoServiceIT.                                                                                                                    │
  ├─────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 6   │ 53f8dc8f561 Fork fragment execution handler to search       │ Change ThreadPool.Names.SAME → ThreadPool.Names.SEARCH for indices:data/read/analytics/fragment. Prevents blocking the transport I/O thread during query execution (observed 115s   │
  │     │ thread pool                                                 │ blocking in production). Added logger to AnalyticsSearchService. Test verifying correct executor registration.                                                                      │
  ├─────┼─────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 7   │ b60f1b525a6 Skip flaky TieringStatusIT.testTieringStatus    │ @AwaitsFix on a flaky unrelated test.
```

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
```diff
@@ -81,7 +81,7 @@ private static void registerStreamingFragmentHandler(
     ) {
         transportService.registerRequestHandler(
             FragmentExecutionAction.NAME,
-            ThreadPool.Names.SAME,
+            ThreadPool.Names.SEARCH,
```
bharath-techie: Should we receive it in SAME and then fork it to search threadpool similar to `SearchTransportService`? 
```
 IndexShard shard = indicesService.indexServiceSafe(request.getShardId().getIndex())
                        .getShard(request.getShardId().id());
```
For example : code like this happens in transport threadpool.

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
```diff
@@ -43,6 +44,7 @@ public class AverageNativeMemoryUsageTracker extends AbstractAverageUsageTracker
 
     private final LongSupplier rssAnonSupplier;
     private final LongSupplier heapCommittedSupplier;
+    private final LongSupplier nonHeapCommittedSupplier;
```
pradeep-L, line 47: Non heap memory accounts for metadata , code caches so we are removing it from RSSAnon, Then do you think we need to account for thread stack and direct buffers as well because they are also part of RSSAnon.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
Bukhtawar: We can but is there is no material difference right?

server/src/main/java/org/opensearch/node/resource/tracker/AverageNativeMemoryUsageTracker.java
Bukhtawar: Is there an easy way to do it?

bharath-techie (approved):
LGTM

PR #21750, opened 2026-05-20
[Sandbox] Add local index authorization check to analytics engine before planning phase

finnegancarroll:
### Description

The analytics engine's query execution path (PPL, SQL) bypasses the security plugin's index-level privilege evaluation. The `FragmentExecutionRequest` on the data node is executed with `dispatchFragmentStreaming` and never evaluated for index permissions by the security plugin `SecurityFilter` (wraps the `ActionFilter`).

### SQL / PPL queries with analytics engine

In a typical SQL / PPL query authorization is handled at the node-to-node transport layer on the coordinator as the request is being dispatched to shards. The security plugin [provides an ActionFilter](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L964-L971) which every TransportAction child holds a reference to, and [executes on new requests](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/TransportAction.java#L213-L228).

The overall flow for a regular DSL search query is then:
```
NodeClient.execute(SearchAction.INSTANCE, searchRequest)
    → NodeClient.executeLocally(action, request, listener)
      → transportAction(action).execute(request, listener)
        → TransportAction.execute(task, request, listener)
          → RequestFilterChain.proceed()
```

The path for an analytics engine SQL / PPL query avoids the `RequestFilterChain` entirely. The `ShardTaskRunner` is  responsible for dispatching a `FragmentExecutionRequest` on the transport layer.

```
    @Override
    public void run(ShardStageTask task, ActionListener<Void> listener) {
        ShardExecutionTarget target = (ShardExecutionTarget) task.target();
        FragmentExecutionRequest request = requestBuilder.apply(target);
        PendingExecutions pending = pendingFor(target);
        transport.dispatchFragmentStreaming(request, target.node(), stage.responseListenerFor(listener), config.parentTask(), pending);
    }
```

The `FragmentExecutionRequest` fetches a new connection from the transport service, and [directly executes ](https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java#L193-L204)the request itself. Skipping the `TransportAction` framework which typically holds and applies the `ActionFilter`.

```
        pending.tryRun(() -> {
            try {
                Transport.Connection connection = getConnection(null, targetNode.getId());
                transportService.sendChildRequest(connection, FragmentExecutionAction.NAME, request, parentTask, options, handler);
            } catch (Exception e) {
                try {
                    listener.onFailure(e);
                } finally {
                    pending.finishAndRunNext();
                }
            }
        });
```

### DSL queries with analytics engine

DSL permissions are actually evaluated correctly already. This comes from how DSL is routed into the analytics plugin. DSL queries construct a regular `SearchRequest` which is handled by the `TransportSearchAction` and are executed on the transport layer with the local node client as usual.

Handling of DSL queries is done by a new `SearchActionFilter` which is also injected into the `RequestFilterChain` and [redirects these `SearchRequest`s to the `DslExecuteAction.INSTANCE` handler](https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java) instead of executing them as normal search requests. Since this `SearchActionFilter` executes AFTER the security plugin ActionFilter (which has highest priority), privileges are evaluated correctly.

| Path | Unauthorized Index | Status |
|------|-------------------|--------|
| DSL (`POST /{index}/_search`) | ✅ Denied | Secure |
| PPL (`POST /_plugins/_ppl`) | ❌ Allowed | **Gap** |
| SQL (`POST /_plugins/_sql`) | ❌ Allowed | **Gap** |

### Proposed fix

These changes somewhat mimic the functionality of DSL queries for SQL and PPL by adding a pre-planning authorization check which passes a no-op `AnalyticsAuthRequest` through the `ActionFilter` to determine if permissions are valid for the equivalent set of  `FragmentExecutionRequest`s.

No actual request is executed on the transport layer, the `AnalyticsAuthRequest` is passed through the full `ActionFilter` chain to invoke security plugin privilege evaluation, and proceeds with the analytics engine planning + query only if this probing `AnalyticsAuthRequest` is authorized for the same indices.

### Testing

Testing this feature requires an analytics engine enabled OpenSearch node, with SQL plugin installed, and security plugin installed. Currently SQL plugin supports this environment with distribution level security enabled tests. This seems like the most ideal place to house ITs validating index level security over the analytics plugin.

Test implementation is pending. Currently changes are validated with local single node testing.

### Related Issues
N/A

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
@@ -21,6 +21,8 @@ import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.action.support.TimeoutTaskCancellationUtility;
 import org.opensearch.analytics.AnalyticsPlugin;
 import org.opensearch.analytics.EngineContext;
+import org.opensearch.analytics.exec.action.AnalyticsAuthAction;
+import org.opensearch.analytics.exec.action.AnalyticsAuthRequest;
 import org.opensearch.analytics.exec.action.AnalyticsQueryAction;
 import org.opensearch.analytics.exec.task.AnalyticsQueryTask;
 import org.opensearch.analytics.exec.task.AnalyticsQueryTaskRequest;
@@ -109,10 +111,24 @@ public class DefaultPlanExecutor extends HandledTransportAction<ActionRequest, A
 
     @Override
     public void execute(RelNode logicalFragment, Object context, ActionListener<Iterable<Object[]>> listener) {
-        // Fork the entire query lifecycle (planning, scheduling, cleanup) onto the SEARCH
-        // executor so the calling thread — which may be a transport thread — is freed
-        // immediately. The scheduler then drives execution asynchronously and fires
-        // {@code listener} once the query terminates; nothing on this path blocks.
+        // Extract target indices from the logical plan and check authorization before
+        // doing any planning work. Dispatches AnalyticsAuthRequest (which extends SearchRequest)
+        // through the action filter chain so the SecurityFilter evaluates index-level permissions.
+        // Using SearchRequest as the base ensures the legacy security evaluator's IndexResolverReplacer
+        // correctly resolves the target indices.
+        String[] indices = extractIndices(logicalFragment);
+        if (indices.length > 0) {
+            client.execute(
+                AnalyticsAuthAction.INSTANCE,
+                new AnalyticsAuthRequest(indices),
+                ActionListener.wrap(ok -> doExecuteOnSearchPool(logicalFragment, listener), listener::onFailure)
+            );
+        } else {
+            doExecuteOnSearchPool(logicalFragment, listener);
+        }
+    }
+
+    private void doExecuteOnSearchPool(RelNode logicalFragment, ActionListener<Iterable<Object[]>> listener) {
         searchExecutor.execute(() -> {
             try {
                 executeInternal(logicalFragment, listener);
@@ -128,6 +144,23 @@ public class DefaultPlanExecutor extends HandledTransportAction<ActionRequest, A
                 listener.onFailure(new IllegalStateException("Analytics-engine executor rejected the plan: " + e.getMessage(), e));
             }
         });
+    }
+
+    /** Extracts index names from LogicalTableScan nodes in the plan. */
+    private static String[] extractIndices(RelNode plan) {
+        java.util.Set<String> indices = new java.util.LinkedHashSet<>();
+        collectIndices(plan, indices);
+        return indices.toArray(String[]::new);
+    }
+
+    private static void collectIndices(RelNode node, java.util.Set<String> indices) {
+        if (node instanceof org.apache.calcite.rel.core.TableScan scan) {
+            List<String> names = scan.getTable().getQualifiedName();
+            indices.add(names.get(names.size() - 1));
+        }
+        for (RelNode input : node.getInputs()) {
+            collectIndices(input, indices);
+        }
     }
 
     /**
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
@@ -20,7 +20,9 @@ import org.opensearch.analytics.exec.DefaultPlanExecutor;
 import org.opensearch.analytics.exec.QueryPlanExecutor;
 import org.opensearch.analytics.exec.QueryScheduler;
 import org.opensearch.analytics.exec.Scheduler;
+import org.opensearch.analytics.exec.action.AnalyticsAuthAction;
 import org.opensearch.analytics.exec.action.AnalyticsQueryAction;
+import org.opensearch.analytics.exec.action.TransportAnalyticsAuthAction;
 import org.opensearch.analytics.planner.CapabilityRegistry;
 import org.opensearch.analytics.planner.FieldStorageResolver;
 import org.opensearch.analytics.schema.OpenSearchSchemaBuilder;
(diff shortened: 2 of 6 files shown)


Bukhtawar (commented):
I was wondering if we add a transport interceptor that can also validate if the request was indeed authorised.
```
public class AnalyticsAuthInterceptor implements TransportInterceptor {
      @Override
      public <T extends TransportRequest> TransportRequestHandler<T> interceptHandler(
          String action, ..., TransportRequestHandler<T> actualHandler, ...) {

          if (action.equals(FragmentExecutionAction.NAME)) {
              return (request, channel, task) -> {
                  String token = threadContext.getHeader("auth_token");
                  if (token == null || !verify(token)) {
                      throw new SecurityException("Analytics auth not performed");
                  }
                  actualHandler.messageReceived(request, channel, task);
              };
          }
          return actualHandler;
      }
  }
```

finnegancarroll:
> I was wondering if we add a transport interceptor that can also validate if the request was indeed authorised.

Thanks for taking a look @Bukhtawar, I spent a small bit of time trying to implement this and it seems like we would need to work this authorization token down to the thread context from the `QueryPlanExecutor` with some additional plumbing.

I can look at implementing this but i'm wondering if we are guarding against anything additional here since the `QueryPlanExecutor` is the single entry point for analytics engine.

cwperks:
@finnegancarroll another problem to think about with regards to security is how to extract indices from a request pertinent to indices.

There currently are 2 ways:

1. Inside the [IndexResolverReplacer](https://github.com/opensearch-project/security/blob/324a038c845deb30ad4d259ee9459d92d9880f65/src/main/java/org/opensearch/security/privileges/actionlevel/legacy/IndexResolverReplacer.java#L679-L876) of the security plugin
2. The modern way is to implement [TransportIndicesResolvingAction](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/search/TransportSearchAction.java#L143-L144) so that security puts that onus on the transport action to be able to resolve the indices.

cwperks:
> The FragmentExecutionRequest on the data node is executed with dispatchFragmentStreaming and never evaluated for index permissions by the security plugin SecurityFilter (wraps the ActionFilter).

Agreed on putting the call to `dispatchFragmentStreaming` within a transport action. @finnegancarroll should we get the security plugin in a state where we can test with the sandbox from core?

cwperks:
Quickly came up with https://github.com/cwperks/security/pull/97 which adds test that would be expected to fail until a fix like this lands.

cwperks:
@finnegancarroll I'm also seeing some action names that I think we should rename: https://github.com/cwperks/OpenSearch/pull/362

^ I haven't been tracking sandbox too closely so lmk what you think and if there are other actions similar.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -130,6 +146,23 @@ public void execute(RelNode logicalFragment, Object context, ActionListener<Iter
...
     }
 
+    /** Extracts index names from LogicalTableScan nodes in the plan. */
+    private static String[] extractIndices(RelNode plan) {
+        java.util.Set<String> indices = new java.util.LinkedHashSet<>();
+        collectIndices(plan, indices);
+        return indices.toArray(String[]::new);
+    }
+
+    private static void collectIndices(RelNode node, java.util.Set<String> indices) {
```
expani:
```
RelNodeUtils#findNode(RelNode node, Class<T> type)
```

We need to write a variation of this method that also takes a parameter of depth as a guard and returns empty if Scan is not found. 

The depth parameter can be HardCoded to 10/15 as a constant.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -109,10 +111,24 @@ public DefaultPlanExecutor(
...
-        // Fork the entire query lifecycle (planning, scheduling, cleanup) onto the SEARCH
-        // executor so the calling thread — which may be a transport thread — is freed
-        // immediately. The scheduler then drives execution asynchronously and fires
-        // {@code listener} once the query terminates; nothing on this path blocks.
+        // Extract target indices from the logical plan and check authorization before
+        // doing any planning work. Dispatches AnalyticsAuthRequest (which extends SearchRequest)
+        // through the action filter chain so the SecurityFilter evaluates index-level permissions.
+        // Using SearchRequest as the base ensures the legacy security evaluator's IndexResolverReplacer
+        // correctly resolves the target indices.
+        String[] indices = extractIndices(logicalFragment);
```
expani, line 119: If we are extracting indices here, it means Index patterns will also be resolved here. 

So, to avoid duplication the `OpenSearchTableScanRule` needs to be able to read the indices extracted here. Maybe via the PlannerContext or some other common place.

expani (commented):
Thanks for getting this started @finnegancarroll

DarshitChanpura:
For my understanding, why are we implementing security check outside Security plugin? we are trying to move away from that model, by making security aware of the related actions.

Adding a transport action should allow using the authz layer of security plugin. (I didn't read through the conversations, maybe this is already discuseed)

finnegancarroll:
>For my understanding, why are we implementing security check outside Security plugin? we are trying to move away from that model, by making security aware of the related actions.

> Adding a transport action should allow using the authz layer of security plugin. (I didn't read through the conversations, maybe this is already discuseed)

+1

Initially I thought there was some design limitation here but reading through the flow for `FragmentExecutionRequest` it's not clear what stops us from executing these normally as transport actions. Giving this refactor a try.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
finnegancarroll: I think PrivilegesEvaluator in security plugin holds onto a `IndexNameExpressionResolver` and should handle this for us. So we should only need to provide the literal index names here.

finnegancarroll:
>Agreed on putting the call to dispatchFragmentStreaming within a transport action. @finnegancarroll should we get the security plugin in a state where we can test with the sandbox from core?

Copying offline discussion here. One issue testing security for this feature is we need all three of SQL plugin, Security plugin, OpenSearch core. Just the security plugin and opensearch core alone do not provide us a front end to test analytics engine with, and using the test-ppl-frontend is insufficient because it does not integrate with security plugin  authentication handlers.

finnegancarroll:
Hi @DarshitChanpura @finnegancarroll @expani @Bukhtawar .
Already discussed with a couple of you offline but migrating to a cleaner approach where `DefaultPlanExecutor` executes an analytics engine query locally through a transport action.

Please let me know your thoughts. Adding integration tests to SQL plugin.

https://github.com/opensearch-project/OpenSearch/pull/21789

PR #5513, opened 2022-12-09
[Feature/Identity] Identity use cases

peternied:
### Description

Adding details uses cases for identity, this is going to get long!

### Check List
- [ ] ~New functionality includes testing.~
  - [ ] ~All tests pass~
- [ ] ~New functionality has been documented.~
  - [ ] ~New functionality has javadoc added~
- [X] Commits are signed per the DCO using --signoff
- [ ] ~Commit changes are listed out in CHANGELOG.md file (See: [Changelog](../blob/main/CONTRIBUTING.md#changelog))~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21753, opened 2026-05-20
Fix StringView buffer bloat in stream_next FFI export

bowenlan-amzn:
### Description

`StringViewArray::slice()` shares ALL backing buffers via Arc. When DataFusion's hash aggregate emits via `EmitTo::All` and slices the result into per-batch chunks, each batch carries the full backing buffer pool — not just the strings it references. This caused up to 435x data amplification on multi-shard aggregate queries with many distinct string values.

**Root cause:** Arrow IPC and C Data Interface intentionally do not compact StringView backing buffers during serialization ([arrow-rs #5513](https://github.com/apache/arrow-rs/issues/5513)). Compaction is the caller's responsibility at the serialization boundary.

**Fix:** Add `compact_string_view_columns()` in `stream_next()` before FFI export. Calls `gc()` on StringView/BinaryView columns only when backing buffers are significantly over-allocated. Three-level precondition avoids unnecessary work:

```
stream_next() per batch:
  1. has_views flag?         → No: skip (O(1), set once at stream creation)
  2. view_needs_gc()?        → No: buffers compact (O(n) scan, 2x + 10KB threshold)
  3. gc()                    → compact the bloated columns (O(n) copy)
```

### Performance

| Query type | Cost per batch |
|---|---|
| No StringView columns (numeric agg, scan) | Zero — O(1) bool check |
| StringView but not sliced (scan, filter) | O(n) view scan, no allocation |
| Sliced StringView (hash aggregate emit) | O(n) scan + copy into compact buffer |

Validated on 4-shard ClickBench q19 (14M groups, SearchPhrase):
- Before: 9748 batches × 174MB = 1.7TB transferred, 12 minutes
- After: 9748 batches × 0.4MB = 4GB transferred, 11 seconds

### Testing

```bash
# Unit tests (function correctness + optimization)
cargo test -p opensearch-datafusion --lib -- api::tests

# Integration test (regression guard for stream_next call site)
cargo test -p opensearch-datafusion --test stringview_gc_test
```

The integration test feeds a sliced 10K→100 StringView batch through `df_stream_next` and asserts output buffers are compact (<10KB). **Fails immediately if the gc() call is removed** (gets 310KB instead).

### Check List
- [x] Commits are signed off

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
@@ -35,6 +35,8 @@ use std::num::NonZeroUsize;
 use std::path::PathBuf;
 use std::sync::Arc;
 
+use arrow_schema::DataType;
+
 use arrow_array::ffi::FFI_ArrowArray;
 use arrow_array::RecordBatch;
 use arrow_array::{Array, StructArray};
@@ -454,6 +456,7 @@ pub async unsafe fn stream_next(stream_ptr: i64) -> Result<i64, DataFusionError>
 
     match result {
         Some(batch) => {
+            let batch = compact_string_view_columns(batch);
             let struct_array: StructArray = batch.into();
             let array_data = struct_array.into_data();
             let ffi_array = FFI_ArrowArray::new(&array_data);
@@ -461,6 +464,36 @@ pub async unsafe fn stream_next(stream_ptr: i64) -> Result<i64, DataFusionError>
         }
         None => Ok(0),
     }
+}
+
+/// Compacts StringView/BinaryView columns in a RecordBatch by calling `gc()`.
+/// This eliminates shared backing buffers that inflate batch size when sliced
+/// from a larger array (e.g., hash aggregate EmitTo::All + slice()).
+fn compact_string_view_columns(batch: RecordBatch) -> RecordBatch {
+    let schema = batch.schema();
+    let has_views = schema.fields().iter().any(|f| {
+        matches!(f.data_type(), DataType::Utf8View | DataType::BinaryView)
+    });
+    if !has_views {
+        return batch;
+    }
+    let columns: Vec<Arc<dyn Array>> = batch
+        .columns()
+        .iter()
+        .zip(schema.fields().iter())
+        .map(|(col, field)| match field.data_type() {
+            DataType::Utf8View => {
+                let view: &arrow_array::StringViewArray = col.as_any().downcast_ref().unwrap();
+                Arc::new(view.gc()) as Arc<dyn Array>
+            }
+            DataType::BinaryView => {
+                let view: &arrow_array::BinaryViewArray = col.as_any().downcast_ref().unwrap();
+                Arc::new(view.gc()) as Arc<dyn Array>
+            }
+            _ => Arc::clone(col),
+        })
+        .collect();
+    RecordBatch::try_new(schema, columns).unwrap_or(batch)
 }
 
 /// Closes a result stream. Safe to call with 0 (no-op).
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
@@ -1171,22 +1171,20 @@ mod tests {
         let strings: Vec<String> = (0..1000)
             .map(|i| format!("long_string_value_{:06}_padding", i))
             .collect();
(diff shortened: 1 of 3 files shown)


bharath-techie:
Similar fix in DF - https://github.com/apache/datafusion/pull/20381

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -463,6 +466,60 @@ pub async unsafe fn stream_next(stream_ptr: i64) -> Result<i64, DataFusionError>
...
+        return batch;
+    }
+    let needs_compaction = batch
+        .columns()
+        .iter()
+        .zip(schema.fields().iter())
+        .any(|(col, field)| match field.data_type() {
+            DataType::Utf8View => {
+                let view: &arrow_array::StringViewArray = col.as_any().downcast_ref().unwrap();
+                view_needs_gc(view.data_buffers(), view.total_buffer_bytes_used())
```
bowenlan-amzn, line 495: Pre-condition

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -72,17 +74,22 @@ pub struct QueryStreamHandle {
...
     _session_ctx: Option<datafusion::prelude::SessionContext>,
+    has_views: bool,
 }
 
 impl QueryStreamHandle {
     pub fn new(
         stream: RecordBatchStreamAdapter<CrossRtStream>,
         query_context: QueryTrackingContext,
     ) -> Self {
+        let has_views = stream.schema().fields().iter().any(|f| {
```
expani: nit: can be an Utility method.

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -463,6 +479,60 @@ pub async unsafe fn stream_next(stream_ptr: i64) -> Result<i64, DataFusionError>
...
+        .collect();
+    RecordBatch::try_new(schema, columns).expect("gc'd columns must match schema")
+}
+
+/// Returns true if the backing buffers are over-allocated relative to
+/// the data actually referenced by views.
+#[inline]
+fn view_needs_gc(buffers: &[arrow::buffer::Buffer], bytes_used: usize) -> bool {
+    let bytes_allocated: usize = buffers.iter().map(|b| b.len()).sum();
+    bytes_allocated > 2 * bytes_used && (bytes_allocated - bytes_used) > 10_240
```
expani: Can you add a comment on how `10_240` is arrived ? 

Should we think of making is a Index setting in future ? 

nit : The condition is a bit hard to read maybe we can break it down.

expani (commented):
Great finding and deep dive @bowenlan-amzn 👏

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -463,6 +479,60 @@ pub async unsafe fn stream_next(stream_ptr: i64) -> Result<i64, DataFusionError>
...
+/// Prevents sliced StringView batches from carrying full backing buffers across FFI.
+fn compact_string_view_columns(batch: RecordBatch) -> RecordBatch {
+    let schema = batch.schema();
+    let has_views = schema.fields().iter().any(|f| {
+        matches!(f.data_type(), DataType::Utf8View | DataType::BinaryView)
+    });
+    if !has_views {
+        return batch;
+    }
+    let needs_compaction = batch
```
expani, line 487: I think we can inspect the Substrait schema itself. This way we avoid checking if each batch has Utf8View or BinaryView ?

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
bowenlan-amzn: done

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
bowenlan-amzn: done

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
bowenlan-amzn: At this place it's not checking view, the check is done per stream creation once.

mch2 (approved):

PR #21758, opened 2026-05-20
[BUG] Fix for intermediate field name handling

nssuresh2007:
#### Description

Fixes a plan conversion failure when a PPL query uses EVAL (project) followed by WHERE (filter) on the derived column — e.g.:
```
SOURCE=index | EVAL answerId = AcceptedAnswerId + 1 | WHERE answerId > 200
```

#### Problem

Substrait does not carry intermediate field names — all references are positional ordinals. When DataFusion's Substrait consumer deserializes a ProjectRel, it invents
names for expression-derived output fields using the expression string (e.g. "table.col + Int32(1)"). If a FilterRel above the ProjectRel references such a derived
field, DataFusion's optimizer (FilterProjectTransposeRule) pushes the filter below the projection. The filter then references the expression-derived column name against
the raw table schema, causing:

`Schema error: No field named "table.col + Int32(1)"`

#### Fix

Added eliminateIntermediateRelNodes() which rewrites Filter(Project(Scan)) into Project(Filter(Scan)) before Substrait conversion. The filter condition is inlined with
the project's expressions so it references raw table columns directly:

Before (problematic):

```
Filter(condition: >(answerId, 200))
  Project(..., answerId=[+($0, 1)])
    TableScan(index)
```
After (safe):
```
Project(..., answerId=[+($0, 1)])
  Filter(condition: >(+($0, 1), 200))
    TableScan(index)
```
This ensures the Substrait plan never contains a filter referencing synthetic intermediate field names.

#### Testing

Added 3 unit tests in DataFusionFragmentConvertorTests:

- testFilterOverProjectOverScan_RewrittenToProjectOverFilterOverScan — verifies the rewrite produces ProjectRel(FilterRel(ReadRel))
- testProjectOverScan_NoFilter_PassesThroughUnchanged — no false triggering
- testFilterOverScan_NoProject_PassesThroughUnchanged — no false triggering

Manually verified with PPL query on a local cluster:
```
SOURCE=parquet_test | EVAL newScore = score + 1 | WHERE newScore > 5
```

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
@@ -22,11 +22,15 @@ import org.apache.calcite.rel.RelReferentialConstraint;
 import org.apache.calcite.rel.RelRoot;
 import org.apache.calcite.rel.core.AggregateCall;
 import org.apache.calcite.rel.core.TableScan;
+import org.apache.calcite.rel.logical.LogicalFilter;
+import org.apache.calcite.rel.logical.LogicalProject;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rel.type.RelDataTypeFactory;
 import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexShuttle;
 import org.apache.calcite.schema.ColumnStrategy;
 import org.apache.calcite.sql.SqlAggFunction;
 import org.apache.calcite.sql.SqlFunctionCategory;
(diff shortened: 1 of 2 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
```diff
@@ -600,6 +627,49 @@ private static Rel withAggregationPhase(Rel rel, Expression.AggregationPhase pha
...
+     * Eliminates intermediate project nodes that introduce derived fields referenced by
+     * a filter above them. Inlines the project's expressions into the filter condition
+     * and moves the project above the filter.
+     *
+     * <p>This works around a Substrait/DataFusion limitation: Substrait does not carry
+     * intermediate field names, so DataFusion's Substrait consumer derives them from
+     * expression strings. DataFusion's optimizer then pushes filters past projections
+     * using those synthetic names, which fail schema resolution against the raw table.
+     */
+    private static RelNode eliminateIntermediateRelNodes(RelNode node) {
```
expani, line 640: Maybe we can explore how to do this before FragmentConversionDriver calls into DatafusionFragmentConvertor, one place is `BackendPlanAdapter`

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
```diff
@@ -458,6 +462,29 @@ private byte[] convertToSubstrait(RelNode fragment) {
         // LOCAL_*_OP stubs so isthmus's AggregateFunctionConverter binds them by
         // operator identity through ADDITIONAL_AGGREGATE_SIGS.
         preprocessed = PplAggregateCallRewriter.rewrite(preprocessed);
+
+        // Substrait does not name intermediate fields in a plan — all references are
```
expani, line 466: Let's reduce the verbosity of comments maybe 2-3 lines

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
mch2: those adapters are per function though?

expani: @nssuresh2007 What are we trying to achieve at a high level ? 

Is there a standard Calcite out of the box Rule that can be used ?

PR #21763, opened 2026-05-20
Add projectCapability for CAST in analytics backend

Swiddis:
### Description
Fixes ~14 integ tests failures from PPL in the new backend for aggregations involving casts ("No backend supports scalar function [CAST]"). The capability checking here is based on the return type of the function so we need to add rules that say CAST returns appropriate types. This comes up with operations on mixed-type arrays, coercing arrays to function inputs like `mvzip`, and certain constant folding ops.

### Related Issues
N/A

### Check List


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
```diff
@@ -505,6 +497,15 @@ public Set<ProjectCapability> projectCapabilities() {
                 for (ScalarFunction op : MAP_RETURNING_PROJECT_OPS) {
                     caps.add(new ProjectCapability.Scalar(op, Set.of(FieldType.MAP), formats, true));
                 }
+                // CAST and SAFE_CAST are type-conversion functions that must support casting TO
```
vinaykpud: The loop at above already iterates those sets and registers each entry against FieldType.ARRAY / FieldType.MAP ie exactly what these new lines do for CAST / SAFE_CAST. Reusing the constants keeps `projectCapabilities()` to one pattern. CAST was already in STANDARD_PROJECT_OPS but `STANDARD_PROJECT_OPS` is registered against `SUPPORTED_FIELD_TYPES`, which looks like intentionally excludes ARRAY and MAP. 

Two ways to reuse the existing patterns:
Option 1:  add to existing sets `ARRAY_RETURNING_PROJECT_OPS` 

Option 2:
new dedicated constant 

```
private static final Set<ScalarFunction> POLYMORPHIC_RETURN_PROJECT_OPS =
      Set.of(ScalarFunction.CAST, ScalarFunction.SAFE_CAST);
```
```
for (ScalarFunction op : POLYMORPHIC_RETURN_PROJECT_OPS) {
    caps.add(new ProjectCapability.Scalar(op, Set.of(FieldType.ARRAY), formats, true));
    caps.add(new ProjectCapability.Scalar(op, Set.of(FieldType.MAP), formats, true));
}
```

This keeps meaning of  `ARRAY_RETURNING_PROJECT_OPS` as "always ARRAY" without polluting with cast.

Swiddis: went with the polymorphic constant approach and decided to just register CAST and SAFE_CAST as every type since they definitionally can return anything

vinaykpud (approved):

mch2 (approved):

PR #21765, opened 2026-05-20
[Analytics Backend / DataFusion] Wire sub-second multi-unit span() via date_bin

ahkcs:
## What this fixes

Dashboard logs-tab histogram queries that bucket timestamps into
**sub-second** widths (e.g. `span(@timestamp, 40ms)`,
`span(@timestamp, 250us)`) fail on the analytics-engine route against
parquet-backed indices with:

```
Unable to convert call SPAN(precision_timestamp<0>?, i32, char<1>)
```

## Why it happens

`SpanAdapter` already had two lowering paths for time-mode SPAN:

1. `interval == 1`, any unit → `date_trunc(<unit>, field)` (predates this PR)
2. fixed-second unit (`s`, `m`, `h`, `d`, `w`) with `interval > 1` →
   integer-seconds arithmetic via `to_unixtime / from_unixtime` (added by
   #21747)

Sub-second units (`us`, `ms`) with `interval > 1` are in the gap between
those two paths. The arithmetic approach can't be reused because
`to_unixtime` returns BIGINT seconds and would truncate sub-second
precision. So `SPAN(t, 40, 'ms')` falls through unchanged and isthmus
then fails to bind the PPL SPAN UDF.

## What this PR changes

Adds a third lowering path checked after the existing two:

- `interval == 1`, any unit       → `date_trunc(<unit>, field)`             *(unchanged)*
- fixed-second + `interval > 1`   → integer-seconds arithmetic              *(#21747, unchanged)*
- **sub-second + `interval > 1`   → `date_bin("<N> <unit>", field)`**         *(new)*

DataFusion's `date_bin(stride, source)` accepts a string-form interval
stride like `"40 milliseconds"` or `"250 microseconds"` natively. Emitting
the stride as a plain string literal avoids substrait interval-type
plumbing that would otherwise be needed for sub-second precision, and
matches how DataFusion's native parser accepts strides at the SQL
surface.

Multi-unit month/quarter/year still fall through unchanged — calendar-
dependent bucket length needs `date_bin` with an interval-month stride;
tracked separately.

### Wiring

1. **`opensearch_scalar_functions.yaml`** — new `date_bin` entry with
   signature `(stride: string, source: precision_timestamp<P>) -> precision_timestamp<P>`.
2. **`SpanAdapter.java`** — `SUB_SECOND_UNIT_TO_DATE_BIN_STRIDE` map plus a
   `LOCAL_DATE_BIN_OP` Calcite `SqlFunction`, used by the new branch in
   `adapt()`.
3. **`DataFusionFragmentConvertor.java`** —
   `FunctionMappings.s(SpanAdapter.LOCAL_DATE_BIN_OP, "date_bin")`.

### Unit (`SpanAdapterSubSecondTests`, 5/5 pass)

- `testMillisecondMultiUnitRewritesToDateBin` — `40ms` lowers to
  `date_bin("40 milliseconds", field)`.
- `testMicrosecondMultiUnitRewritesToDateBin` — `250us` lowers to
  `date_bin("250 microseconds", field)`.
- `testSubSecondUnitOneStaysOnDateTruncPath` — `N == 1` for `us`/`ms`
  keeps using the cheaper `date_trunc` path (regression guard).
- `testFractionalIntervalForSubSecondFallsThrough` — non-integer
  intervals are not rewritten.
- `testAdaptedDateBinCallPreservesOriginalReturnType` — Project rowType
  invariant.

Full `:sandbox:plugins:analytics-backend-datafusion:test` suite passes
(no regression of the timechart fixed-second arithmetic path).

### QA integration (`SpanTimeCommandIT`, 5/5 pass)

Self-contained REST test in `sandbox/qa/analytics-engine-rest` against
a parquet-backed cluster, exercising the full PPL → CalciteRelNodeVisitor
→ SpanAdapter → Substrait → DataFusion pipeline. Uses a focused 5-row
`span_time` dataset with explicit sub-second timestamps (the existing
`calcs` dataset is whole-second only).

| Test | Path | Result |
|---|---|---|
| `testSpanUnitOneHourLowersToDateTrunc` | date_trunc | 5 rows in one 1h bucket |
| `testSpanMultiUnitHoursLowersToArithmetic` | arithmetic (#21747) | 5 rows in one 2h bucket |
| `testSpanFortyMillisecondsBucketsEachRowDistinctly` | **date_bin (new)** | 5 distinct 40ms buckets |
| `testSpanMicrosecondsBucketsEachRowDistinctly` | **date_bin (new)** | 5 distinct 250us buckets |
| `testSpanOneMillisecondStaysOnDateTruncPath` | date_trunc | regression guard: `N == 1 ms` stays on cheaper path |

Run via:

```bash
./gradlew :sandbox:qa:analytics-engine-rest:integTest \
  -Dsandbox.enabled=true --tests "*SpanTimeCommandIT"
```

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SpanAdapter.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SpanAdapter.java
@@ -14,8 +14,14 @@ import org.apache.calcite.rex.RexBuilder;
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.SqlFunction;
+import org.apache.calcite.sql.SqlFunctionCategory;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlOperator;
 import org.apache.calcite.sql.fun.SqlLibraryOperators;
 import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.OperandTypes;
+import org.apache.calcite.sql.type.ReturnTypes;
 import org.apache.calcite.sql.type.SqlTypeName;
 import org.opensearch.analytics.spi.FieldStorageInfo;
 import org.opensearch.analytics.spi.ScalarFunctionAdapter;
@@ -35,11 +41,25 @@ import java.util.Map;
  *       {@code (field / interval) * interval} for integer types (where Calcite's
  *       integer division already truncates).</li>
  *   <li><b>Time span</b> ({@code unit} is a single-letter unit string like
- *       {@code "y"}, {@code "M"}, {@code "d"}, etc.): rewritten to
- *       {@code DATE_TRUNC(<unit>, field)} when {@code interval == 1}. Multi-unit
- *       intervals like {@code 12h} aren't expressible as {@code date_trunc} and
- *       fall through to the original UDF, which surfaces as a normal substrait
- *       binding error rather than a silent wrong-result.</li>
+ *       {@code "y"}, {@code "M"}, {@code "d"}, etc.) lowers through three paths,
+ *       checked in order:
+ *       <ul>
+ *         <li><b>{@code interval == 1}, any unit:</b> rewritten to
+ *             {@code DATE_TRUNC(<unit>, field)} — the cheapest DataFusion path.</li>
+ *         <li><b>Fixed-length unit (s/m/h/d/w) with {@code interval > 1}:</b> bucketed
+ *             via integer-seconds arithmetic ({@code TIMESTAMP_SECONDS(FLOOR(UNIX_SECONDS(t) / B) * B)}).
+ *             Exact because the epoch is a fixed reference for these units.</li>
+ *         <li><b>Sub-second unit (us/ms) with {@code interval > 1}:</b> rewritten to
+ *             DataFusion's native {@code date_bin("<N> <unit>", field)}. The arithmetic
+ *             path can't be used here because {@code to_unixtime} returns BIGINT seconds
+ *             and loses sub-second precision. Emitting the stride as a plain string
+ *             literal (e.g. {@code "40 milliseconds"}) avoids substrait interval-type
+ *             plumbing and matches how DataFusion's native parser accepts strides.</li>
+ *       </ul>
+ *       Multi-unit month / quarter / year intervals (e.g. {@code 12M}) still fall through
+ *       to the original UDF — their bucket length is calendar-dependent, requiring a
+ *       {@code date_bin} with an interval-month stride; tracked separately.
+ *   </li>
  * </ul>
  *
  * <p>The unit-letter mapping mirrors PPL's {@code SpanUnit} enum (defined in the SQL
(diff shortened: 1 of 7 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SpanAdapter.java
```diff
@@ -119,13 +169,60 @@ public RexNode adapt(RexCall original, List<FieldStorageInfo> fieldStorage, RelO
...
+            if (bd.scale() > 0 && bd.stripTrailingZeros().scale() > 0) {
+                return null;
+            }
+            try {
+                n = bd.longValueExact();
+            } catch (ArithmeticException e) {
+                return null;
+            }
+        } else if (value instanceof Number num) {
+            double d = num.doubleValue();
```
expani: nit: We can create a generic utility somewhere to check for Rounding of a Double value

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SpanAdapter.java
```diff
@@ -119,13 +169,60 @@ public RexNode adapt(RexCall original, List<FieldStorageInfo> fieldStorage, RelO
...
+     * sub-second date_bin path bakes the unit into the stride string instead.
+     */
+    private static Long extractPositiveInteger(RexNode interval) {
+        if (!(interval instanceof RexLiteral lit)) {
+            return null;
+        }
+        Object value = lit.getValue();
+        long n;
+        if (value instanceof BigDecimal bd) {
+            if (bd.scale() > 0 && bd.stripTrailingZeros().scale() > 0) {
```
expani: nit: Should just `bd.stripTrailingZeros().scale() > 0` be enough ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SpanAdapter.java
expani, line 144: Unrelated to this PR's scope. 

Is this the structure of SQL/PPL plugin leaking into the Datafusion Backend ? 

Can we add a TODO to move such transformations out of backends and Analytics-Engine to send a common structure that is easy to adapt ?

ahkcs: Added TODO above `adapt()` in `57508e77f11` — calls out the SPAN-name match as PPL-coupling and names the long-term direction (Analytics-Engine sends a backend-neutral bucketing primitive).

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SpanAdapter.java
ahkcs: Good catch — dropped the redundant `bd.scale() > 0` conjunct in `57508e77f11`.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SpanAdapter.java
ahkcs: Extracted `isIntegralDouble(double)` in `57508e77f11` and folded `bucketSecondsIfPositiveInteger` to delegate to `extractPositiveInteger` so the literal-parsing lives in one place.

mch2 (approved):

PR #21766, opened 2026-05-20
Add Liquid Cache integration to analytics-backend-datafusion

cocosz:
# Liquid Cache Integration

## Overview

Liquid Cache is a byte-level columnar cache that sits between DataFusion's query engine and Parquet scans. It caches frequently-accessed numeric predicate columns in a compressed columnar format (Liquid encoding), eliminating repeated Parquet IO and decode overhead on hot query patterns.

This PR vendors a minimal, **in-memory-only** subset of liquid-cache directly into the sandbox native workspace (`opensearch-liquid-cache-{core,datafusion}`), so OpenSearch takes no external Cargo dependency on the upstream liquid-cache package.

## Problem

When the same Parquet columns are queried repeatedly (common in analytics dashboards, monitoring, and ad-hoc exploration), each query re-reads and re-decodes the same column data. This is wasteful for columns that participate in `WHERE` predicates — they are accessed on every query that filters on them.

## Solution

Liquid cache intercepts Parquet reads at the DataFusion physical plan level, caching numeric predicate columns in memory using a space-efficient columnar encoding (Liquid format). Subsequent queries on the same columns serve from the cache, avoiding Parquet IO and decode. Under memory pressure, cached Arrow batches are transcoded to the compressed Liquid form; when the budget is still exceeded, entries are evicted.

## Vendored Subset

The vendored copy is intentionally minimal. Relative to upstream liquid-cache it **drops**:

- the disk tier (t4 / io-uring)
- client/server mode (arrow-flight / tonic)
- FSST string caching
- variant shredding
- the lineage optimizer

Because io-uring is gone, the cache is **memory-only and platform-independent** (no `#[cfg(target_os = "linux")]` gating).

## Architecture

```
Query -> DataFusion Planning -> Physical Optimizer
                                     |
                         LocalModeOptimizer rewrites
                         ParquetSource -> LiquidParquetSource
                                     |
                                     v
                         LiquidParquetSource.read()
                         +--------+--------+
                         |                 |
                     Cache HIT         Cache MISS
                  (serve from       (read from Parquet,
                   memory)           insert into cache)
```

### Optimizer

- **LocalModeOptimizer** (physical) — Walks the physical plan tree, finds `DataSourceExec` nodes containing `ParquetSource`, and rewrites them to `LiquidParquetSource`, routing reads through the cache. The upstream LineageOptimizer (logical) is intentionally excluded — it adds planning overhead that regresses string-heavy queries.

### Engagement Guard

The cache engages on two independent scan paths, each with its own column-count guard:

1. **Indexed-query path** (`indexed_table/parquet_bridge.rs`) — bounded by `datafusion.liquid_cache.indexed_query.max_columns` (default 10).
2. **Listing-table path** (`LocalModeOptimizer`) — bounded by `datafusion.liquid_cache.listing_table.max_columns` (default 4, lower because this path decodes full columns so per-column cache overhead outweighs savings sooner).

For either path, all of the following must hold before the cache engages:

1. The cache is enabled globally (`is_enabled_globally()` — feature flag on and runtime toggle on).
2. All projected columns are numeric/date/timestamp/boolean (string columns bypass the cache).
3. The number of projected columns does not exceed the path's `max_columns`.

### Cache Lifecycle

1. **Initialization** — Once per process as a singleton. The cache survives across queries and sessions.
2. **Cache miss** — Column batch is read from Parquet, then inserted into the cache in Arrow format.
3. **Memory pressure** — Entries are transcoded Arrow -> Liquid columnar format (2–5x compression); when the memory budget is still exceeded, entries are evicted according to the eviction policy.
4. **Cache hit** — Data is served directly from the in-memory representation, with optional predicate evaluation on the compressed form (avoiding full decompression).

## Guardrails

LC is gated by multiple independent layers — all must be true for the cache to engage:

| Layer | Mechanism | Default |
|-------|-----------|---------|
| Startup feature flag | `opensearch.experimental.feature.liquid_cache.enabled` | `false` |
| Runtime toggle | `datafusion.liquid_cache.enabled` cluster setting | `true` (when flag is on) |
| Per-query column type check | Only numeric/date/timestamp/boolean columns | Bypassed for strings |
| Per-query column count check | `indexed_query.max_columns` / `listing_table.max_columns` | 10 / 4 |

### Feature Flag (required to enable)

```yaml
# opensearch.yml or JVM property
opensearch.experimental.feature.liquid_cache.enabled: true
```

This is a `NodeScope` static setting — must be set at node startup. When `false`, the cache is never initialized and no cluster settings listeners are registered.

### Cluster Settings

| Setting | Type | Default | Dynamic | Description |
|---------|------|---------|---------|-------------|
| `datafusion.liquid_cache.enabled` | boolean | `true` | yes | Runtime on/off toggle. Pauses cache engagement for new queries without restart. Only effective when the feature flag is on. |
| `datafusion.liquid_cache.size_bytes` | long | `1073741824` (1 GB) | yes | Maximum memory budget for cached entries. When exceeded, the eviction policy selects victims for transcode/eviction. |
| `datafusion.liquid_cache.eviction_policy` | string | `lru` | no (`NodeScope + Final`) | Eviction policy. `lru` = standard LRU across all entry types. `liquid` = independent FIFO queues per batch type. Validated case-insensitively; an invalid value fails fast at startup. |
| `datafusion.liquid_cache.indexed_query.max_columns` | int | `10` | yes | Max projected columns for LC engagement on the indexed-query path. Queries projecting more columns bypass the cache. |
| `datafusion.liquid_cache.listing_table.max_columns` | int | `4` | yes | Max projected columns for LC engagement on the listing-table path. |

### Example Configuration

```yaml
# opensearch.yml
opensearch.experimental.feature.liquid_cache.enabled: true
datafusion.liquid_cache.size_bytes: 2147483648                  # 2 GB memory cache
datafusion.liquid_cache.eviction_policy: liquid                 # per-type FIFO queues
datafusion.liquid_cache.indexed_query.max_columns: 12
datafusion.liquid_cache.listing_table.max_columns: 6
```

### Rust (native layer)

| File | Purpose |
|------|---------|
| `liquid_cache.rs` | `LiquidOnlyRuntime` singleton — builds cache, extracts optimizer, exposes stats/reset/dynamic-update APIs |
| `api.rs` | `DataFusionRuntime` extended with `liquid_cache_optimizer` field; `create_global_runtime` accepts cache params |
| `ffm.rs` | FFM exports: `df_create_global_runtime` (updated), `df_clear_liquid_cache`, `df_set_liquid_cache_enabled`, `df_set_liquid_cache_memory_limit`, `df_set_liquid_cache_indexed_max_columns`, `df_set_liquid_cache_listing_max_columns` |
| `session_context.rs` | Injects the optimizer rule into per-query DataFusion sessions (listing-table path) |
| `query_executor.rs` | Optimizer injection for the standalone query execution path |
| `indexed_table/parquet_bridge.rs` | Per-scan engagement guard: checks `is_enabled_globally()`, column types, and column count before wrapping `ParquetSource` with `LiquidParquetSource` |

### Java (settings + bridge)

| File | Purpose |
|------|---------|
| `FeatureFlags.java` | `opensearch.experimental.feature.liquid_cache.enabled` feature flag definition and registration |
| `DatafusionSettings.java` | LC cluster settings (`enabled`, `size_bytes`, `eviction_policy`, `indexed_query.max_columns`, `listing_table.max_columns`); registered in `ALL_SETTINGS` |
| `DataFusionPlugin.java` | Reads feature flag and settings at startup; passes to the service builder; wires dynamic cluster-settings listeners and applies initial values to the native runtime |
| `DataFusionService.java` | Builder fields for LC params; passes to native `createGlobalRuntime` |
| `NativeBridge.java` | FFM bridge — updated `createGlobalRuntime`; added `clearLiquidCache`, `setLiquidCacheEnabled`, `setLiquidCacheMemoryLimit`, `setLiquidCacheIndexedMaxColumns`, `setLiquidCacheListingMaxColumns` |
| `action/LiquidCacheClearAction.java` | REST handler for `POST _plugins/analytics_backend_datafusion/liquid_cache/clear`; only registered when the feature flag is on |
| `action/stats/RestDataFusionStatsAction.java`, `TransportDataFusionStatsAction.java` | Expose Liquid Cache stats through the DataFusion stats endpoint under the `liquid_cache` stat section |

### Stats endpoint

Liquid Cache stats are exposed via the DataFusion cluster stats API under the `liquid_cache` section:

```
GET /_plugins/_analytics_backend_datafusion/stats/liquid_cache
GET /_plugins/_analytics_backend_datafusion/{nodeId}/stats/liquid_cache
```

Stat reads are non-destructive (they do not reset counters).

### Debug logging

Cache statistics are also logged at DEBUG level after each query stream closes:

```
[LiquidCache] entries=1, mem=120/1073741824, arrow=1(120 B), liquid=0
[LiquidCache] hits=1, misses=0, predicate_evals=1
```

To enable:

```yaml
logger.org.opensearch.nativebridge.spi.RustLoggerBridge: DEBUG
```

## Running Locally

```bash
# Build the native library
cd sandbox/libs/dataformat-native/rust
cargo build --release

# Run OpenSearch with liquid cache enabled
./gradlew run -Dsandbox.enabled=true \
  -PinstalledPlugins="['arrow-base','arrow-flight-rpc','analytics-engine','parquet-data-format','analytics-backend-datafusion','analytics-backend-lucene','dsl-query-executor','composite-engine','org.opensearch.plugin:opensearch-job-scheduler:3.7.0.0-SNAPSHOT','org.opensearch.plugin:opensearch-sql-plugin:3.7.0.0-SNAPSHOT']" \
  -Dtests.jvm.argline="-Djava.library.path=sandbox/libs/dataformat-native/rust/target/release \
    -Dopensearch.experimental.feature.pluggable.dataformat.enabled=true \
    -Dopensearch.pluggable.dataformat.merge.enabled=true \
    -Dopensearch.experimental.feature.transport.stream.enabled=true \
    -Dopensearch.experimental.feature.liquid_cache.enabled=true \
    -Xms4g -Xmx4g" \
  -x javadoc -x test -x missingJavadoc
```

## Getting Started

```bash
# Create a composite/parquet index
curl -X PUT "localhost:9200/test_idx" -H 'Content-Type: application/json' -d '{
  "settings": {
    "index.number_of_shards": 1,
    "index.number_of_replicas": 0
  }
}'

# Ingest data
curl -X POST "localhost:9200/_bulk" -H 'Content-Type: application/json' -d '
{"index":{"_index":"test_idx"}}
{"name":"Alice","age":30,"salary":75000}
{"index":{"_index":"test_idx"}}
{"name":"Bob","age":25,"salary":60000}
{"index":{"_index":"test_idx"}}
{"name":"Charlie","age":35,"salary":90000}
'

# Force merge to parquet segments
curl -X POST "localhost:9200/test_idx/_forcemerge?max_num_segments=1&flush=true"
sleep 5

# Query with numeric predicate (triggers cache)
curl -X POST 'localhost:9200/_plugins/_ppl' -H 'Content-Type: application/json' \
  -d '{"query": "source=test_idx | where age > 25 | stats sum(salary) as total"}'

# Run again -- should hit cache (check DEBUG logs, or the stats endpoint, for hits=1)
curl -X POST 'localhost:9200/_plugins/_ppl' -H 'Content-Type: application/json' \
  -d '{"query": "source=test_idx | where age > 25 | stats sum(salary) as total"}'

# Inspect cache stats
curl "localhost:9200/_plugins/_analytics_backend_datafusion/stats/liquid_cache"

# Clear the cache if needed
curl -X POST 'localhost:9200/_plugins/analytics_backend_datafusion/liquid_cache/clear'
```

## Known Limitations

- **In-memory only** — The disk tier (t4/io-uring) is dropped in the vendored subset. Only the memory cache is available; there is no disk spill for evicted entries.
- **Numeric predicate columns only** — String columns bypass the cache; only numeric/date/timestamp/boolean projections engage it.
- **Vendored subset** — `opensearch-liquid-cache-{core,datafusion}` is a trimmed in-tree copy. It will be re-synced or replaced with an upstream release once the library is published under the `opensearch-project` org.
- **Experimental** — Gated behind `opensearch.experimental.feature.liquid_cache.enabled`. API and settings are subject to change.

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
@@ -313,6 +313,8 @@ pub struct DataFusionRuntime {
     pub runtime_env: datafusion::execution::runtime_env::RuntimeEnv,
     pub custom_cache_manager: Option<CustomCacheManager>,
     pub dynamic_limit_handle: DynamicLimitHandle,
+    pub liquid_cache_optimizer:
+        Option<Arc<dyn datafusion::physical_optimizer::PhysicalOptimizerRule + Send + Sync>>,
 }
 
 /// Per-file metadata passed from Java at shard view creation time.
@@ -417,7 +419,22 @@ impl DataFusionRuntime {
             runtime_env,
             custom_cache_manager: None,
             dynamic_limit_handle: handle,
+            liquid_cache_optimizer: None,
         }
+    }
+
+    pub fn apply_liquid_cache_optimizers(
+        &self,
+        mut builder: SessionStateBuilder,
+    ) -> SessionStateBuilder {
+        if let Some(ref optimizer) = self.liquid_cache_optimizer {
+            builder = builder.with_physical_optimizer_rule(optimizer.clone());
+        }
+        builder
+    }
+
+    pub fn has_liquid_cache(&self) -> bool {
+        self.liquid_cache_optimizer.is_some()
     }
 }
 
@@ -511,6 +528,9 @@ pub fn create_global_runtime(
     cache_manager_ptr: i64,
     spill_dir: &str,
     spill_limit: i64,
+    liquid_cache_enabled: bool,
+    liquid_cache_size: i64,
+    liquid_cache_eviction_policy: &str,
 ) -> Result<i64, DataFusionError> {
     if memory_pool_limit < 0 {
         return Err(DataFusionError::Configuration(format!(
@@ -712,10 +732,21 @@ pub fn create_global_runtime(
         .with_cache_manager(cache_manager_config)
         .build()?;
 
+    let liquid_cache_optimizer = if liquid_cache_enabled {
+        let liquid_runtime = crate::liquid_cache::LiquidOnlyRuntime::init(
+            liquid_cache_size as u64,
+            liquid_cache_eviction_policy,
+        )?;
+        Some(liquid_runtime.optimizer())
+    } else {
+        None
+    };
+
     let runtime = DataFusionRuntime {
         runtime_env,
         custom_cache_manager,
         dynamic_limit_handle,
+        liquid_cache_optimizer,
     };
     Ok(Box::into_raw(Box::new(runtime)) as i64)
 }
@@ -1511,6 +1542,23 @@ pub fn cancel_query(context_id: i64) {
     query_tracker::cancel_query(context_id);
 }
 
+/// Clears all caching layers: Liquid Cache (in-memory) and DataFusion
+/// metadata caches (parquet footers + column statistics).
+///
+/// # Safety
+/// `runtime_ptr` must be 0 or a valid pointer from `create_global_runtime`.
+pub unsafe fn clear_liquid_cache(runtime_ptr: i64) {
+    crate::liquid_cache::LiquidOnlyRuntime::reset_cache_if_initialized();
+
+    if runtime_ptr == 0 {
+        return;
+    }
+    let runtime = &*(runtime_ptr as *const DataFusionRuntime);
+    if let Some(ref cache_manager) = runtime.custom_cache_manager {
+        cache_manager.clear_all();
+    }
+}
+
 /// Converts SQL to Substrait plan bytes (test only).
 ///
 /// # Safety
@@ -2174,7 +2222,8 @@ mod tests {
         // memory_guard SPILL_ENABLED flag off so per_query_spill_budget returns
         // Disabled (not Critical) — preventing the 1-partition clamp.
         let _guard = SPILL_GLOBALS_LOCK.lock().unwrap_or_else(|e| e.into_inner());
-        let ptr = create_global_runtime(64 * 1024 * 1024, 0, "", 0).expect("runtime build");
+        let ptr = create_global_runtime(64 * 1024 * 1024, 0, "", 0, false, 0, "lru")
+            .expect("runtime build");
         assert!(ptr > 0);
         let runtime = unsafe { &*(ptr as *const DataFusionRuntime) };
         assert!(
@@ -2211,8 +2260,16 @@ mod tests {
             "sentinel must exist before runtime build"
         );
 
-        let ptr = create_global_runtime(64 * 1024 * 1024, 0, spill_path, 1024 * 1024 * 1024)
-            .expect("runtime build");
+        let ptr = create_global_runtime(
+            64 * 1024 * 1024,
+            0,
+            spill_path,
+            1024 * 1024 * 1024,
+            false,
+            0,
+            "lru",
+        )
+        .expect("runtime build");
         assert!(ptr > 0);
 
         // Phase 1 renames the sentinel file to leaked_from_prior_run.tmp.stale
@@ -2275,8 +2332,16 @@ mod tests {
         assert!(top_file.exists());
         assert!(nested_file.exists());
 
-        let ptr = create_global_runtime(64 * 1024 * 1024, 0, spill_path, 1024 * 1024 * 1024)
-            .expect("runtime build");
+        let ptr = create_global_runtime(
+            64 * 1024 * 1024,
+            0,
+            spill_path,
+            1024 * 1024 * 1024,
+            false,
+            0,
+            "lru",
+        )
+        .expect("runtime build");
         assert!(ptr > 0);
 
         // Phase 1: original names gone (renamed to *.stale).
@@ -2320,7 +2385,8 @@ mod tests {
         // accidental fs::remove_dir_all("") would error and break boot. This test
         // guards against future refactors that hoist the cleanup out of the else-branch.
         let _guard = SPILL_GLOBALS_LOCK.lock().unwrap_or_else(|e| e.into_inner());
-        let ptr = create_global_runtime(64 * 1024 * 1024, 0, "", 0).expect("runtime build");
+        let ptr = create_global_runtime(64 * 1024 * 1024, 0, "", 0, false, 0, "lru")
+            .expect("runtime build");
         assert!(ptr > 0);
         unsafe { close_global_runtime(ptr) };
     }
@@ -2338,7 +2404,7 @@ mod tests {
         fs::write(&bad_path, b"not a directory").expect("seed regular file");
         let bad_path_str = bad_path.to_str().expect("utf-8 path");
 
-        let err = create_global_runtime(64 * 1024 * 1024, 0, bad_path_str, 0)
+        let err = create_global_runtime(64 * 1024 * 1024, 0, bad_path_str, 0, false, 0, "lru")
             .expect_err("create_global_runtime must fail when cleanup fails");
 
         // Operator-facing message must include the offending path + io kind so the
@@ -2412,7 +2478,7 @@ mod tests {
         fs::set_permissions(parent.path(), locked).expect("chmod parent 555");
 
         let spill_str = spill_path.to_str().expect("utf-8 path");
-        let result = create_global_runtime(64 * 1024 * 1024, 0, spill_str, 0);
+        let result = create_global_runtime(64 * 1024 * 1024, 0, spill_str, 0, false, 0, "lru");
 
         // Restore parent perms via RAII so tempdir cleanup runs even on assertion failure.
         struct RestorePerms<'a> {
@@ -2491,7 +2557,8 @@ mod tests {
         assert!(link.is_symlink(), "precondition: link is a symlink");
 
         let spill_str = spill_path.to_str().expect("utf-8 path");
-        let ptr = create_global_runtime(64 * 1024 * 1024, 0, spill_str, 0).expect("runtime build");
+        let ptr = create_global_runtime(64 * 1024 * 1024, 0, spill_str, 0, false, 0, "lru")
+            .expect("runtime build");
         assert!(ptr > 0);
 
         // Phase 1: symlink is renamed inline (fs::rename does not follow symlinks).
@@ -2541,7 +2608,8 @@ mod tests {
         fs::create_dir(&leaked_b).expect("create leaked b");
         fs::write(leaked_b.join("tmp_002.arrow"), b"data b").expect("write b");
 
-        let ptr = create_global_runtime(64 * 1024 * 1024, 0, spill_path, 0).expect("runtime build");
+        let ptr = create_global_runtime(64 * 1024 * 1024, 0, spill_path, 0, false, 0, "lru")
+            .expect("runtime build");
         assert!(ptr > 0);
 
         // Originals were renamed inline — gone immediately by the original name.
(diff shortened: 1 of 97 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSettings.java
```diff
@@ -164,6 +164,55 @@ public final class DatafusionSettings {
...
+     * Entries evicted from memory are written to disk up to this limit.
+     * Default is unlimited (Long.MAX_VALUE).
+     */
+    public static final Setting<Long> LIQUID_CACHE_MAX_DISK_BYTES = Setting.longSetting(
+        "datafusion.liquid_cache.max_disk_bytes",
+        Long.MAX_VALUE,
+        0L,
+        Setting.Property.NodeScope,
+        Setting.Property.Final
+    );
```
Bukhtawar, line 232: Can we make it dynamic? Als lets understand what happens if we increase/decrease the size

cocosz: I have made the settings dynamic in the next commit. You can check also I have redesigned liquid cache to be dynamic as well.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSettings.java
```diff
@@ -184,6 +184,67 @@ public final class DatafusionSettings {
...
+     *   <li>{@code liquid} (default) — Independent FIFO queues per batch type (LiquidPolicy)</li>
+     *   <li>{@code lru} — Standard LRU eviction across all entry types</li>
+     * </ul>
+     */
+    public static final Setting<String> LIQUID_CACHE_EVICTION_POLICY = Setting.simpleString(
+        "datafusion.liquid_cache.eviction_policy",
+        "liquid",
+        Setting.Property.NodeScope,
+        Setting.Property.Final
+    );
```
Bukhtawar: Do we want it to be LRU or S3FIFO?

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -235,7 +244,36 @@ pub fn create_global_runtime(
...
+                )
+            }
+            Err(e) => {
+                return Err(e);
+            }
+        }
+    } else {
+        (runtime_env, None, None)
+    };
+
```
Bukhtawar: Lets change the liquid cache branch to only extract the optimizer rules while keeping the original runtime_env
```
let (liquid_optimizer, liquid_lineage_optimizer) = if liquid_cache_enabled {
      match crate::liquid_cache::LiquidOnlyRuntime::init(...) {
          Ok(liquid_runtime) => (Some(liquid_runtime.optimizer()), Some(liquid_runtime.lineage_optimizer())),
          Err(e) => return Err(e),
      }
  } else {
      (None, None)
  };

  let runtime = DataFusionRuntime {
      runtime_env,  // always the user-configured one
      custom_cache_manager,
      dynamic_limit_handle,
      liquid_cache_optimizer: liquid_optimizer,
      liquid_cache_lineage_optimizer: liquid_lineage_optimizer,
  };
```

sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
```diff
@@ -104,12 +104,27 @@ pub async fn execute_query(
     config.options_mut().execution.parquet.pushdown_filters = query_config.parquet_pushdown_filters;
     config.options_mut().execution.target_partitions = query_config.target_partitions;
     config.options_mut().execution.batch_size = query_config.batch_size;
+    config.options_mut().execution.parquet.schema_force_view_types = false;
+    config.options_mut().execution.parquet.skip_arrow_metadata = false;
+    config.options_mut().execution.parquet.skip_metadata = false;
```
Bukhtawar: Is this change only needed for liquid cache?

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -235,7 +244,36 @@ pub fn create_global_runtime(
...
         .build()?;
 
-    let runtime = DataFusionRuntime { runtime_env, custom_cache_manager, dynamic_limit_handle };
+    let (final_runtime_env, liquid_optimizer, liquid_lineage_optimizer) = if liquid_cache_enabled {
+        match crate::liquid_cache::LiquidOnlyRuntime::init(
+            liquid_cache_size as u64,
+            liquid_cache_max_disk_bytes as u64,
+            liquid_cache_dir,
+            liquid_cache_eviction_policy,
+        ) {
```
Bukhtawar: Why cant we reuse the same DF tokio runtime given this is just an optimiser

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSettings.java
cocosz: We can make it LRU

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
cocosz: Done. The RuntimeEnv (memory pool, disk spill, cache manager) is now always preserved. LC only provides optimizer rules — LiquidParquetSource internally manages its own IO through the t4 store and doesn't depend on the session's RuntimeEnv.

sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
cocosz: Yes. These 3 parquet settings are now gated behind if runtime.liquid_cache_optimizer.is_some() — they only apply when LC is initialized. When LC is disabled, DataFusion's defaults are preserved unchanged.

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
cocosz: Fixed. Now passing the RuntimeManager's IO runtime handle into LiquidOnlyRuntime::init() instead of creating a throwaway tokio::Runtime.

Bukhtawar (commented):
How are we restricting numeric/timestamp fields?

Bukhtawar (commented):
How did we verify if this works for both the IndexedTableProvider and ListingTable paths?

cocosz:
> How are we restricting numeric/timestamp fields?

The restriction happens inside the liquid-cache library itself, not in our OpenSearch integration code. The LineageOptimizer (logical optimizer rule) inspects filter predicates in the query plan and identifies which columns are used in predicate expressions. When those columns are cached, the cache's CachePolicy only admits columns with numeric Arrow data types (Int8/16/32/64, UInt8/16/32/64, Float16/32/64, Decimal128/256, Date32/64, Time32/64, Timestamp, Duration). String, Binary, and other complex types are excluded at the cache admission layer.

cocosz:
>How did we verify if this works for both the IndexedTableProvider and ListingTable paths?

Liquid Cache hooks in at the optimizer level — it's a physical optimizer rule that rewrites how Parquet I/O happens (cache column chunks), not what the scan's filter logic does. The IndexedTableProvider's evaluator (single-collector or bitmap-tree) operates on the same decoded Arrow arrays regardless of whether they came from cache or raw Parquet decode.However I will change the integ test to add cases to cover both paths.

PR #21767, opened 2026-05-20
[analytics-engine] DatePartAdapters: coerce VARCHAR operand to TIMESTAMP and validate string literals

mengweieric:
### Problem

PPL allows bare string-literal datetime arguments like `DAY('2020-09-16')`. Calcite types those operands as `VARCHAR`, so `DatePartAdapters` (and the siblings `DayOfWeekAdapter` / `SecondAdapter`) produced `DATE_PART('<unit>', <varchar>)`. `opensearch_scalar_functions.yaml` only declares `date_part` for `(string, precision_timestamp<P>)` and `(string, date)`, so isthmus' substrait matcher rejected the call with:

```
IllegalArgumentException: Unable to convert call DATE_PART(string, string)
```

This blocked **31 PPL date-part tests** across `CalcitePPLBuiltinDatetimeFunctionInvalidIT` (18), `CalciteDateTimeFunctionIT` (11), and `CalcitePPLBuiltinFunctionsNullIT` (2).

Two follow-on issues uncovered while iterating:
- `DayOfWeekAdapter` / `SecondAdapter` build their `date_part` calls directly, so they didn't pick up the original `DatePartAdapters` fix.
- Invalid literals (e.g. `'2025-13-02'`) failed inside DataFusion's CAST kernel; the Arrow error text is dropped by Flight RPC on the worker→coordinator hop, so the legacy `"unsupported format"` wording the IT assertions expect never reached the user.

### Solution

Engine-side only — all changes under `sandbox/plugins/analytics-backend-datafusion/`.

1. **VARCHAR→TIMESTAMP coercion** in `DatePartAdapters.adapt`; `TIMESTAMP` / `DATE` operands pass through.
2. **Shared helper** `coerceCharacterOperandToTimestamp` so `DayOfWeekAdapter` / `SecondAdapter` (which build `date_part` directly) get the same wrapping.
3. **Plan-time literal validation** mirroring legacy `DateTimeParser.parse` — flat try-chain (`LocalDateTime` with space→T, then `LocalDate`, then `LocalTime`) that throws `IllegalArgumentException("timestamp:<v> in unsupported format, ...")` without attaching the cause (SQL plugin's `unwrapCause` would otherwise surface the inner JDK message instead of the legacy wording).

#### Unit tests
`:sandbox:plugins:analytics-backend-datafusion:check` — **BUILD SUCCESSFUL** locally under JDK 25. Covers TIMESTAMP / DATE pass-through, VARCHAR + CHAR coercion, nullability preservation, unit-literal propagation, NULL-literal pass-through, column-ref-not-eagerly-validated, and 4 invalid-literal rejection cases (month-13, second-61, bare-time-second-61, garbage) across all three adapter test classes.

#### End-to-end ITs (force-routed AE cluster, `:integ-test:integTestRemote -Dtests.analytics.parquet=true`)
- **6 originally-tracked failing tests now pass**: `testDAYOFWEEKInvalid`, `testDAY_OF_WEEKInvalid`, `testSECONDInvalid`, `testSECOND_OF_MINUTEInvalid` in `CalcitePPLBuiltinDatetimeFunctionInvalidIT`; `testDayOfWeek`, `testDay_of_week` in `CalciteDateTimeFunctionIT`.
- **Full 31-test set verified** by running the 3 affected IT classes end-to-end and grepping the test-result XML directory for any lingering `DATE_PART(string, string)` mention:

before

```
31 test methods across 3 classes fail with Unable to convert call DATE_PART(string, string).

  ┌───────┬────────────────────────────────────────────┬────────────────────────────────────────────────────────────────┐
  │ Tests │                   Class                    │                        Outer exception                         │
  ├───────┼────────────────────────────────────────────┼────────────────────────────────────────────────────────────────┤
  │    18 │ CalcitePPLBuiltinDatetimeFunctionInvalidIT │ AssertionError (test-side — they assert on the error response) │
  ├───────┼────────────────────────────────────────────┼────────────────────────────────────────────────────────────────┤
  │    11 │ CalciteDateTimeFunctionIT                  │ ResponseException (server-side — query returns HTTP 500)       │
  ├───────┼────────────────────────────────────────────┼────────────────────────────────────────────────────────────────┤
  │     2 │ CalcitePPLBuiltinFunctionsNullIT           │ AssertionError (test-side)
```

after

| Class | Total tests | `DATE_PART(string, string)` matches |
|---|---|---|
| CalcitePPLBuiltinDatetimeFunctionInvalidIT | 45 | **0** |
| CalciteDateTimeFunctionIT | 65 | **0** |
| CalcitePPLBuiltinFunctionsNullIT | 71 | **0** |

The error mode is fully eliminated across the 3 affected classes. Residual failures in those classes are unrelated error types (other engine gaps), not the DATE_PART signature mismatch this PR targets.

### Check List
- [ ] API changes companion pull request created, if applicable. *(N/A — internal adapter behavior, engine-side only.)*
- [ ] Public documentation issue/PR created, if applicable. *(N/A — bug fix, no PPL-visible behavior change.)*

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatePartAdapters.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatePartAdapters.java
@@ -8,21 +8,84 @@
 
 package org.opensearch.be.datafusion;
 
+import org.apache.calcite.plan.RelOptCluster;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.sql.fun.SqlLibraryOperators;
+import org.apache.calcite.sql.type.SqlTypeFamily;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.opensearch.analytics.spi.AbstractNameMappingAdapter;
+import org.opensearch.analytics.spi.FieldStorageInfo;
 
+import java.util.ArrayList;
 import java.util.List;
 
 /**
  * Date-part extractor adapters — rewrite {@code FN(ts)} to {@code date_part('<unit>', ts)}.
  * Alias pairs (e.g. MONTH_OF_YEAR → MONTH) share an adapter instance at registration.
  *
+ * <p>Operand coercion: PPL allows bare string-literal datetime arguments
+ * (e.g. {@code DAY('2020-09-16')}). Calcite types those operands as
+ * {@code VARCHAR}, but {@code opensearch_scalar_functions.yaml} declares
+ * {@code date_part} only for {@code (string, precision_timestamp<P>)} and
+ * {@code (string, date)}. Without coercion, isthmus' signature matcher fails
+ * with {@code "Unable to convert call DATE_PART(string, string)"}. This adapter
+ * wraps any character-family operand in {@code CAST(operand AS TIMESTAMP)} so
+ * the call resolves to the precision_timestamp impl. DataFusion's CAST kernel
+ * parses ISO-8601 datetime strings; malformed inputs surface as a runtime
+ * Arrow cast error from the engine.
+ *
  * @opensearch.internal
  */
 final class DatePartAdapters extends AbstractNameMappingAdapter {
 
+    private final String unit;
+
     DatePartAdapters(String unit) {
         super(SqlLibraryOperators.DATE_PART, List.of(unit), List.of());
+        this.unit = unit;
+    }
+
+    @Override
+    public RexNode adapt(RexCall original, List<FieldStorageInfo> fieldStorage, RelOptCluster cluster) {
+        if (original.getOperands().stream().noneMatch(DatePartAdapters::isCharacterOperand)) {
+            return super.adapt(original, fieldStorage, cluster);
+        }
+        RexBuilder rexBuilder = cluster.getRexBuilder();
+        List<RexNode> coerced = new ArrayList<>(original.getOperands().size());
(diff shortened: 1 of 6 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatePartAdapters.java
```diff
@@ -8,21 +8,149 @@
...
+     * <p>Non-literal operands (column refs, expressions) and NULL literals pass through — column
+     * value validation is a separate concern (Arrow CAST per-row error handling, tracked
+     * separately).
+     *
+     * <p>Accept-set mirrors legacy {@code DateTimeParser.parse}: try {@link LocalDateTime} (date+time
+     * with optional nanos), {@link LocalDate} (bare date), {@link LocalTime} (bare time), throw on
+     * all-failed. Same try/catch-fall-through shape used by
+     * {@link TimestampFunctionAdapter#parseTimestamp}.
+     */
+    static void validateDatetimeLiteral(RexNode operand) {
```
vinaykpud, line 125: this function duplicates parse logic from `TimestampFunctionAdapter.parseTimestamp`. Same try-LocalDateTime / LocalDate ladder, just with different format coverage and error message. Worth giving a look to see if we can extract the parsing to common method

mengweieric: I looked at extracting, but the overlap seems thinner than it looks. The two methods have different signatures (String → TimestampString vs RexNode → void), inverted accept sets (parseTimestamp needs OffsetDateTime/Instant, this one needs bare LocalTime for HOUR('10:30:00')), and different error contracts (this one must throw causeless to survive unwrapCause). A shared helper ends up parameterized enough that I think it reads worse than the two inline ladders.

vinaykpud (approved):

mch2 (approved):

PR #21775, opened 2026-05-21
[analytics-engine] Per-shard bucket oversampling for TopK aggregation queries

sandeshkr419:
Description:

  Adds a post-CBO rewriter that inserts a per-partition Sort+Limit between PARTIAL aggregate and ExchangeReducer, so each shard only ships top-N×factor groups instead of all groups.

  - Index setting index.analytics.shard_bucket_oversampling_factor (default 0, opt-in)
  - OpenSearchTopKRewriter finds bottommost Sort with collation above grouped FINAL→ER→PARTIAL and inserts per-shard Sort
  - reduce_eval scalar UDF for evaluating opaque aggregate state (approx_distinct HLL) to a sortable scalar — generic pattern for future approximate aggregates
  - OpenSearchSort.perPartition flag skips the singleton cost gate

  Tests: TopKRewriterPlanShapeTests (2), ShardBucketOversamplingIT (7 on ClickBench multi-shard), reduce_eval, Rust unit test.
### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
```diff
@@ -140,7 +171,7 @@ public Collection<Module> createGuiceModules() {
 
     @Override
     public List<Setting<?>> getSettings() {
-        return List.of(COORDINATOR_BUFFER_LIMIT);
+        return List.of(COORDINATOR_BUFFER_LIMIT, INDEX_ANALYTICS_SHARD_BUCKET_OVERSAMPLING_FACTOR);
```
expani: Can we categorize the settings into different classes ? 

Concurrency/Memory/Perf; etc.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/AggregateShardBucketTests.java
```diff
@@ -0,0 +1,156 @@
...
+        RelNode result = runPlanner(plan, singleShardContext());
+        assertPlanShape("""
+            OpenSearchSort(sort0=[$1], dir0=[ASC], fetch=[100], viableBackends=[[mock-parquet]])
+              OpenSearchAggregate(group=[{0}], cnt=[COUNT()], mode=[SINGLE], viableBackends=[[mock-parquet]])
+                OpenSearchTableScan(table=[[test_index]], viableBackends=[[mock-parquet]])
+            """, result);
+    }
+
+    /** No LIMIT in the query: rule skips (no coord-required size to derive shardSize from). */
+    public void testNoLimit_noRewrite() {
```
mch2: today we have a default fetch size of i think 10, should we honor this setting as default?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public OpenSearchAggregateShardBucketRule(PlannerContext context) {
+        super(operand(OpenSearchSort.class, any()), "OpenSearchAggregateShardBucketRule");
+        this.context = context;
+    }
+
+    @Override
+    public boolean matches(RelOptRuleCall call) {
+        OpenSearchSort sort = call.rel(0);
+        if (!(sort.fetch instanceof RexLiteral)) return false;
```
expani: I think we should always be performing the Approximation when settings are enabled Irrespective of whether there is a limit in the query OR Sort Columns are empty.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,161 @@
...
+ * {@code index.analytics.shard_bucket_oversampling_factor} across all tables
+ * scanned under the PARTIAL aggregate. Any index with {@code factor == 0}
+ * disables the rewrite for the whole query.
+ *
+ * <p>Approximate: a group missing from every shard's top-{@code shardSize}
+ * cannot reach the coordinator.
+ *
+ * @opensearch.internal
+ */
+public class OpenSearchAggregateShardBucketRule extends RelOptRule {
```
expani: This needs to be a generic PlannerPhase that handles any rewrites related to ExecutionHints provided by the User. Defaults saved in Indexing or Cluster Setting.

Performing a TopK Approximation is just one of the many rewrites that will be performed in the future based on User's desire.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,161 @@
...
+        );
+
+        RelNode newEr = shape.er.copy(shape.er.getTraitSet(), java.util.List.of(shardSort));
+        RelNode newFinal = shape.finalAgg.copy(shape.finalAgg.getTraitSet(), java.util.List.of(newEr));
+        RelNode newOuter = outerSort.copy(outerSort.getTraitSet(), newFinal, outerSort.getCollation(), outerSort.offset, outerSort.fetch);
+        call.transformTo(newOuter);
+    }
+
+    /** Pattern-match the FINAL ← ER ← PARTIAL chain under a Sort. */
+    private static Shape findShape(RelNode input) {
```
expani: We need to use the Marking phase of the Planner to attach information about the depth of the relational operator in the plan. This will be similar to how we MARK `viableBackends` in the HepRules today. 

Using the depth, we can then figure out whether the Sort we are changing is the one that's at the bottommost location of the plan or not.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -103,6 +104,8 @@ public static RelNode runAllOptimizations(RelNode rawRelNode, PlannerContext con
         // AnnotatedPredicates under OR/NOT (Lucene call buys nothing in those positions).
         modifiedRelNode = cbo(modifiedRelNode, rawRelNode, context, listener);
         LOGGER.info("After CBO:\n{}", RelOptUtil.toString(modifiedRelNode));
+        modifiedRelNode = shardBucketOversample(modifiedRelNode, context, listener);
```
expani: I think this should happen before CBO so that we don't need to worry about the exchange operators.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: Done for the LIMIT case, the rule now fires when no LIMIT is present and falls back to DEFAULT_COORD_LIMIT = 10 (matching PPL head's default). The empty-sort-collation check stays, without an ordering there's no meaningful top-K to compute and we'd be dropping arbitrary groups. Pure shard-side LIMIT without ORDER BY is a different optimization (raw-doc top-K) and is out of scope here.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: Done, execution-hints is now a generic pre-CBO HEP phase. Future rewrites driven by user/index/cluster execution-shape hints (raw-doc top-K shipping factor, partition pruning, memory-budget directives) plug in alongside OpenSearchAggregateShardBucketRule without architectural changes.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: Deferred. HEP TOP_DOWN match order combined with the per-aggregate idempotency check (getShardBucketHint() == null) handles the outermost-Sort case naturally. Calcite's redundant-sort-elimination already collapses chained-sort patterns before this rule fires. Explicit depth tracking via the marking phase is a future enhancement; we can revisit in your QTF enablement?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
sandeshkr419: Done, moved entirely to pre-CBO via the execution-hints HEP phase. The rule attaches a ShardBucketHint to the SINGLE-mode aggregate before any PARTIAL/FINAL/ER exists.
OpenSearchAggregateSplitRule consumes the hint at split time when it has the full split context. No exchange-operator awareness in the rule itself.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
sandeshkr419: Done, created AnalyticsApproximationSettings as the accuracy/memory-tradeoff category.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/AggregateShardBucketTests.java
sandeshkr419: Done, DEFAULT_COORD_LIMIT = 10L (matching PPL head's default count) is used both as the floor for the shardSize calculation and as the fallback when Sort.fetch is null. Test renamed testNoLimit_defaultFallback; asserts the rewrite fires with shardSize = ceil(max(10, 10) * 1.5) + 10 = 25 over a no-LIMIT input.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/hll_estimate.rs
```diff
@@ -0,0 +1,311 @@
...
+const HLL_Q: usize = 64 - HLL_P;
+const NUM_REGISTERS: usize = 1 << HLL_P;
+
+pub fn register_all(ctx: &SessionContext) {
+    ctx.register_udf(ScalarUDF::from(HllEstimateUdf::new()));
+}
+
+/// `hll_estimate(binary)` → `i64`.
+#[derive(Debug)]
+pub struct HllEstimateUdf {
```
expani: Would we end up requiring one UDF per aggregate like Approx Distinct Count ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchSort.java
```diff
@@ -43,9 +47,52 @@ public OpenSearchSort(
...
+                    + "] must match collation field count ["
+                    + collation.getFieldCollations().size()
+                    + "]"
+            );
+        }
+        this.sortExprs = sortExprs == null ? null : List.copyOf(sortExprs);
+    }
+
+    /** True when this Sort is a shard-local top-K — see {@link #localTopK}. */
+    public boolean isLocalTopK() {
```
expani: A rel node sub-shape is topK and not an individual relational operator like Sort.

So, not sure why a sort is being marked as topK ?

expani (commented):
Thanks for the changes @sandeshkr419 

Dropping some early comments, still going through the PR.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/hll_estimate.rs
sandeshkr419: The shard-bucket rule needs to sort per-shard groups by their "cardinality" while shipping the Binary HLL sketch (so coord can merge correctly). DataFusion's `approx_distinct` exposes cardinality only inside an aggregate, not as a scalar. So we need a scalar `hll_estimate(Binary state) → BIGINT`.


It's **one UDF per engine-native-merge aggregate family**. Pass-through (SUM/MIN/MAX/COUNT) and primitive-decomposed (AVG/STDDEV/VAR) aggregates need zero UDFs as their sort keys are already first-class column refs or standard arithmetic expressions.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchSort.java
sandeshkr419: Fair point, the flag isn't "this Sort is a topK"; it's "evaluate this Sort per-partition (don't insert a gather below it)". Renaming to perPartition (or partitionLocal/shardLocal) reflects the actual semantic in a follow-up commit.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/hll_estimate.rs
sandeshkr419: I'm looking into another way to get this removed in a fast follow-up. Need to clean up other pieces of architecture as well.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -197,6 +209,74 @@ public void onMatch(RelOptRuleCall call) {
...
+     *
+     * <p>Mode is FINAL for additive aggregates (SUM/MIN/MAX/COUNT/AVG/...): shard emits per-group
+     * scalars, coord FINAL re-aggregates. Mode is SHARD_MERGE when any aggCall is engine-native
+     * merge (e.g. APPROX_COUNT_DISTINCT): shard emits intermediate state (Binary HLL sketch),
+     * coord runs state-merge via {@link org.opensearch.analytics.planner.dag.DistributedAggregateRewriter};
+     * the Sort sortExprs use {@link AggregateFunction#finalizeOperator} (e.g. {@code hll_estimate(state)}).
+     *
+     * <p>Returns {@code partial} unchanged when no hint is present.
+     */
+    private static RelNode maybeInsertShardSort(
```
expani: We could have done this in the rewriter itself and leave this rule to do just the splitting

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,271 @@
...
+
+    public OpenSearchAggregateShardBucketRule(PlannerContext context) {
+        super(operand(OpenSearchSort.class, any()), "OpenSearchAggregateShardBucketRule");
+        this.context = context;
+    }
+
+    @Override
+    public boolean matches(RelOptRuleCall call) {
+        OpenSearchSort sort = call.rel(0);
+        if (sort.getCollation().getFieldCollations().isEmpty()) return false; // top-K needs a sort key
```
expani: We can sort on group by key ? Default bucket order

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,271 @@
...
+            anyTable[0] = true;
+            return;
+        }
+        for (RelNode input : current.getInputs()) {
+            walkSingleShard(input, anyMultiShard, anyTable);
+        }
+    }
+
+    /** {@link Project} immediately below {@code sort}, or {@code null}. */
+    private static Project projectBelow(OpenSearchSort sort) {
```
expani: These checks seem very fragile. What are we trying to do here ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,271 @@
...
+        if (current instanceof OpenSearchTableScan scan) {
+            String tableName = scan.getTable().getQualifiedName().getLast();
+            IndexMetadata indexMetadata = context.getClusterState().metadata().index(tableName);
+            if (indexMetadata != null && indexMetadata.getNumberOfShards() > 1) {
+                anyMultiShard[0] = true;
+            }
+            anyTable[0] = true;
+            return;
+        }
+        for (RelNode input : current.getInputs()) {
```
expani: You can directly find the RelNode using an utility method in RelNodeUtils

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,271 @@
...
+ *
+ * <p>The hint's {@code sortExprs} carry per-collation-field RexNodes evaluated at the shard.
+ * Three sources: pass-through column ref (SUM/MIN/MAX/COUNT), the recompose Project's
+ * RexNode (AVG/STDDEV/VAR), or {@link AggregateFunction#finalizeOperator} (engine-native
+ * merge — APPROX_COUNT_DISTINCT). Approximate: a group missing from every shard's
+ * top-{@code shardSize} cannot reach the coordinator.
+ *
+ * @opensearch.internal
+ */
+public class OpenSearchAggregateShardBucketRule extends RelOptRule {
```
expani: Looks like we are rewriting even in the case of Joins/Unions. 

Should we explicitly reject such cases ? OR add more tests for them ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,271 @@
...
+
+    /** Min {@code shard_bucket_oversampling_factor} across involved tables; {@code 0.0} disables for the whole query. */
+    private double resolveFactor(RelNode subtree) {
+        double[] state = { Double.MAX_VALUE };
+        boolean[] any = { false };
+        collectFactor(subtree, state, any);
+        return any[0] ? state[0] : 0.0;
+    }
+
+    private void collectFactor(RelNode node, double[] state, boolean[] any) {
```
expani: Can't we store the factor in PlannerContext ? So, that it's easier to read instead of extracting from RelNode again ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: Addressed. aggregateBelow now uses RelNodeUtils.findNode. projectBelow is a single-line instanceof check that's unavoidable (we need to detect the optional recompose Project between Sort and Aggregate).

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: Addressed. Replaced collectFactor/readFactor/walkSingleShard with visitScans helper that reads factor directly from OpenSearchTableScan.getShardBucketFactor() (set at marking time by OpenSearchTableScanRule). Rule no longer accesses ClusterState/PlannerContext.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: Addressed. Added hasMultiInput guard in matches().

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: Addressed, stored it on OpenSearchTableScan itself (set during OpenSearchTableScanRule from IndexMetadata.getSettings()). The rule reads it directly via scan.getShardBucketFactor() without any tree traversal or ClusterState lookup. PlannerContext dependency removed entirely from the rule.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
```diff
@@ -0,0 +1,267 @@
...
+        RelNode current = RelNodeUtils.unwrapHep(node);
+        if (current.getInputs().size() > 1) return true;
+        for (RelNode input : current.getInputs()) {
+            if (hasMultiInput(input)) return true;
+        }
+        return false;
+    }
+
+    @Override
+    public void onMatch(RelOptRuleCall call) {
```
expani: Can you check for query shapes that can contain multiple Sorts from PPL plugin ? 

The test-ppl frontend won't show the same error. 

Here is one such query and it's RelNode
```
curl -s -X POST "http://localhost:9200/_plugins/_ppl" -H "Content-Type: application/json" -d '{"query": "source = clickbench | where GoodEvent=1 | sort - ResolutionWidth | head 3 | fields URL, ResolutionWidth, EventTime"}'
```

RelNode from PPL plugin 
```
LogicalSystemLimit(sort0=[$1], dir0=[DESC-nulls-last], fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(URL=[$0], ResolutionWidth=[$1], EventTime=[CAST($2):VARCHAR])
    LogicalSort(sort0=[$1], dir0=[DESC-nulls-last])
      LogicalProject(URL=[$12], ResolutionWidth=[$9], EventTime=[$3])
        LogicalSort(sort0=[$9], dir0=[DESC-nulls-last], fetch=[3])
          LogicalFilter(condition=[=($4, 1)])
            LogicalTableScan(table=[[opensearch, clickbench]])
```

Let's say we add a group by to this, then I believe the detection logic so go for the bottom most sort which this matching won't handle.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: Deferred as part of this change. Currently the rule only fires when there's an explicit user-facing Sort with collation. Will review this in a follow-up.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
expani: Uh I meant store in the planner context :) but nvm it's fine maybe add a TODO for later.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
sandeshkr419: Done. Moved shard-sort insertion from OpenSearchAggregateSplitRule to DistributedAggregateRewriter.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateShardBucketRule.java
sandeshkr419: The rule fires on every OpenSearchSort in the tree independently (not just the topmost). The inner Sort(fetch=3) → Project → Aggregate matches naturally via aggregateBelow(), while the outer SystemLimit is a no-op since it doesn't see an Aggregate below it.
 
 Can iterate this approach to something better next?
 
 Validated with sql plugin as well.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/reduce_eval.rs
```diff
@@ -0,0 +1,137 @@
...
+    let expr: Arc<dyn datafusion::physical_plan::PhysicalExpr> = Arc::new(Column::new("x", 0));
+    let ret_field: Arc<Field> = Arc::new(Field::new("r", DataType::UInt64, true));
+
+    let mut results = Vec::with_capacity(binary.len());
+    for i in 0..binary.len() {
+        if binary.is_null(i) {
+            results.push(0u64);
+            continue;
+        }
+        let mut acc = approx_distinct_udaf().accumulator(AccumulatorArgs {
```
expani, line 77: Can we try to make this generic for any UDAF that has a accumulator in Datafusion and try to see if there is a simple way to add support for the remaining one's ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchSort.java
```diff
@@ -98,6 +118,9 @@ public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
         if (getCollation().getFieldCollations().isEmpty()) {
             return planner.getCostFactory().makeTinyCost();
         }
+        if (perPartition) {
```
expani: We don't need to affect Calcite's Trait based propagation unless it's for Sort Removal in case the underlying table is sorted. 

TopK is purely a Rewriter concern and any RelNode modifications required should be done within it.

expani (commented):
Thanks @sandeshkr419 it was fun iterating on this offline.

Really appreciate you sticking to the architecture and design intentions 🫡 

Have a few comments around a few oddities that should be quick.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/ShardBucketOversamplingIT.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    /** Q13 shape: count by keyword, sort desc, head 10. */
+    public void testCountByGroup_sortDesc_head10() throws Exception {
+        ensureProvisioned();
+        Map<String, Object> result = executePPL(
+            "source = " + INDEX + " | stats count() as c by RegionID | sort - c | head 10"
+        );
+        assertRowCount(result, 10);
```
expani, line 42: Can we assert on the results WITH AND WITHOUT the use of TopK approximation across single or multiple shards ? 

This should give us a clear view on the expected difference in the results.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+
+        double factor = resolveOversamplingFactor(partial, context);
+        if (factor <= 0.0) return Optional.empty();
+
+        long coordLimit = (sort.fetch instanceof RexLiteral lit) ? RexLiteral.intValue(lit) : 10_000L;
+        long shardSize = (long) Math.ceil(coordLimit * factor) + coordLimit;
+        if (shardSize > Integer.MAX_VALUE) return Optional.empty();
+
+        RexBuilder rb = sort.getCluster().getRexBuilder();
+        RexNode fetchLiteral = rb.makeLiteral((int) shardSize, sort.getCluster().getTypeFactory().createSqlType(SqlTypeName.INTEGER), true);
```
expani: nit: shardSizeLiteral. fetch would confuse future readers with QTF/LateMaterialization

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+        OpenSearchTableScan scan = findScan(partial);
+        if (scan == null) return 0.0;
+        String indexName = scan.getTable().getQualifiedName().get(scan.getTable().getQualifiedName().size() - 1);
+        IndexMetadata meta = context.getClusterState().metadata().index(indexName);
+        if (meta == null) return 0.0;
+        if (meta.getSettings() == null) return 0.0;
+        return AnalyticsApproximationSettings.INDEX_ANALYTICS_SHARD_BUCKET_OVERSAMPLING_FACTOR.get(meta.getSettings());
+    }
+
+    private static OpenSearchTableScan findScan(RelNode node) {
```
expani: Let's use `RelNodeUtils#findNode`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+ * for TopK aggregation queries. Runs before DAGBuilder splits stages.
+ *
+ * <p>Pattern: Sort(collation) → [Project] → Aggregate(FINAL) → ER → Aggregate(PARTIAL) → Scan
+ */
+public final class OpenSearchTopKRewriter {
+
+    private OpenSearchTopKRewriter() {}
+
+    public static Optional<RelNode> rewrite(RelNode root, PlannerContext context) {
+        SortAboveFinal match = findSortAboveFinal(root);
```
expani, line 38: Shouldn't we perform TopK even in the case of a single shard ? 

Let's add a TODO to reconsider in a fast-followup

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+            if (finalAgg != null) return new SortAboveFinal(sort, finalAgg);
+        }
+        for (RelNode child : node.getInputs()) {
+            SortAboveFinal found = findSortAboveFinal(child);
+            if (found != null) return found;
+        }
+        return null;
+    }
+
+    private static OpenSearchAggregate findFinalBelow(RelNode node) {
```
expani: nit: findFinalAgg is a more suitable name.

Also, can be a utility in `RelNodeUtils`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+            if (newChildren[i] != children.get(i)) changed = true;
+        }
+        if (!changed) return root;
+        return root.copy(root.getTraitSet(), List.of(newChildren));
+    }
+
+    private static double resolveOversamplingFactor(OpenSearchAggregate partial, PlannerContext context) {
+        OpenSearchTableScan scan = findScan(partial);
+        if (scan == null) return 0.0;
+        String indexName = scan.getTable().getQualifiedName().get(scan.getTable().getQualifiedName().size() - 1);
```
expani: Can you see how this ties with IndexPattern and Alias ? 

I am fine with adding a TODO for now. But, we might need to resolve settings to a common value in such cases.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private static OpenSearchAggregate findFinalBelow(RelNode node) {
+        if (node instanceof OpenSearchAggregate agg && agg.getMode() == AggregateMode.FINAL) return agg;
+        if (node.getInputs().size() == 1) return findFinalBelow(node.getInputs().get(0));
+        return null;
+    }
+
+    /** Replaces oldNode with newNode in the tree (single occurrence). */
+    private static RelNode replaceInTree(RelNode root, RelNode oldNode, RelNode newNode) {
```
expani, line 106: nit: Looks like another common utility which might already exist.

PS Name can be better.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/TopKRewriterPlanShapeTests.java
```diff
@@ -0,0 +1,91 @@
...
+import org.opensearch.common.settings.Settings;
+
+import java.util.List;
+
+import static org.mockito.Mockito.when;
+
+/**
+ * Plan shape tests for the TopK rewriter.
+ */
+public class TopKRewriterPlanShapeTests extends PlanShapeTestBase {
```
expani, line 32: Let's categorize the Plan related tests by 2 top level variations : 
- Detection
Ensure TopK doesn't fire when Index setting is off, Plan is not supported ( No Aggregate ); etc.  
- Rewrite
These can go more deeper like Rewrite is correct when 2 sorts are present and we modify only the bottom-most sort and a few with Opaque aggregates like DC and Standard Aggregates like Avg

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
mch2: This will break with patterns, the name on the scan node remains the pattern/alias we receive given we don't split it per concrete index.

mch2: i think INDEX_ANALYTICS_SHARD_BUCKET_OVERSAMPLING_FACTOR should be cluster lvl not index?

expani: Agreed that having a cluster setting is the saner starting point.

We can eventually move to using IndexSetting and in the long term future an Execution Hint.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/reduce_eval.rs
sandeshkr419: Added TODO in reduce_eval.rs. Current dispatch is by name ("approx_distinct" => ...). To generalize: look up UDAF by name from the session registry at runtime. Blocked because ScalarFunctionArgs doesn't carry a FunctionRegistry. Need to dig separately.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
sandeshkr419: Removed custom findScan entirely.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
sandeshkr419: Changed from index.analytics.shard_bucket_oversampling_factor (IndexScope) to analytics.shard_bucket_oversampling_factor (NodeScope, Dynamic). Avoids index-pattern/alias resolution issue entirely.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
sandeshkr419: Added a TODO.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/ShardBucketOversamplingIT.java
sandeshkr419: Currently the test PPL frontend (/_analytics/ppl) doesn't support sort + head after stats. The ITs verify row counts only. Added TODO. Will assert exact results once sort+head is supported on the test frontend (works   _plugins/_ppl with sql-plugin).

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchSort.java
sandeshkr419: Done. Removed the if (perPartition) return makeTinyCost() block from computeSelfCost. TopK is purely a rewriter concern.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
sandeshkr419: Done.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
sandeshkr419: Added todo, Can move to RelNodeUtils in follow-up.

mch2 (approved):

PR #21780, opened 2026-05-21
Add STATS bundle aggregate decomposition

varun-st:
### Description
  
  Adds support for the SQL `STATS(field)` bundle aggregate function in the analytics engine sandbox. `STATS` returns a struct of `{count, min, max, avg, sum}` — same shape and field order as legacy OpenSearch's `InternalStats` aggregation.
  
  #### Approach: HEP-time decomposition into primitives
  
  A new HEP rule, `OpenSearchStatsReduceRule`, runs in `PlannerImpl.decomposeAggregates` (alongside the existing `OpenSearchAggregateReduceRule` that handles AVG/STDDEV/VAR). When the rule sees a `LogicalAggregate` containing a `STATS` aggregate call, it rewrites the plan into:
  
  Project [stats = STRUCT(count, min, max, avg, sum)]
    └── LogicalAggregate [SUM(x), MIN(x), MAX(x), COUNT(x)]
         └── (input)
  
  The Project assembles the result struct using Calcite's `ROW` operator. `avg` is computed inline as `CAST(sum AS DOUBLE) / CAST(count AS DOUBLE)`. Subsequent planner phases (marking, Volcano CBO, partial/final split, fragment conversion, DataFusion execution) see only standard primitive aggregates,
  which already have full backend support.
  
  #### Why HEP-time decomposition
  
  `STATS` has no executor implementation in this codebase and isn't a Calcite standard `SqlKind` that the stock `AggregateReduceFunctionsRule` knows how to reduce. Rewriting at HEP keeps `STATS` invisible to downstream phases while reusing battle-tested primitives.
  
  #### SPI additions
  
  - `OpenSearchAggregateOperators.STATS` — custom `SqlAggFunction` singleton with the struct return type. Public so frontends can emit calls to it.
  - `AggregateFunction.STATS` — registers the function category and `SqlKind.OTHER`.
  - `ScalarFunction.ROW` — registers Calcite's struct constructor used by the assembly Project.
  
  #### Notable design choices
  
  - **Emission order `[SUM, MIN, MAX, COUNT]`** (not `[COUNT, MIN, MAX, SUM]`). After Volcano split, FINAL's aggregate calls reuse the original `argList` against PARTIAL's output. Putting `SUM` first means `PARTIAL.col[1]` has the same Calcite type as the source field, so FINAL's `typeMatchesInferred`
  assertion passes for MIN/MAX/SUM. `DistributedAggregateRewriter` corrects argList semantics post-Volcano.
  - **Struct field order `count, min, max, avg, sum`** — matches `InternalStats.doXContentBody` for backward compatibility with any client tooling that consumes results positionally.
  
  #### Tests
  
  - `OpenSearchStatsReduceRuleTests` — 9 unit tests covering single STATS, multiple STATS, STATS mixed with other aggregates, STATS with GROUP BY, FILTER and DISTINCT propagation, struct field names/types, and the no-op case.
  - `StatsAggregatePlanShapeTests` — 4 integration tests through the full `PlannerImpl.runAllOptimizations` covering 1-shard, 2-shard, GROUP BY, and STATS-mixed-with-COUNT scenarios.

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchStatsReduceRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchStatsReduceRule.java
@@ -1,0 +1,289 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.analytics.planner.rules;
+
+import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.RelOptRuleCall;
+import org.apache.calcite.rel.RelCollations;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.core.AggregateCall;
+import org.apache.calcite.rel.logical.LogicalAggregate;
+import org.apache.calcite.rel.logical.LogicalProject;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.sql.SqlAggFunction;
+import org.apache.calcite.sql.fun.SqlStdOperatorTable;
+import org.apache.calcite.sql.type.SqlTypeName;
+import org.apache.calcite.sql.type.SqlTypeUtil;
+import org.opensearch.analytics.spi.OpenSearchAggregateOperators;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * HEP rule that decomposes the {@link OpenSearchAggregateOperators#STATS STATS} bundle aggregate
+ * into the four primitive aggregates ({@code COUNT}, {@code MIN}, {@code MAX}, {@code SUM})
+ * plus a {@link LogicalProject} that re-assembles them — together with a derived {@code AVG}
+ * — into the struct shape the original {@code STATS} call declared.
+ *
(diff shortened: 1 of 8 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchStatsReduceRule.java
```diff
@@ -0,0 +1,289 @@
...
+        int newMaxPos = groupCount + mapping[2];
+        int newSumPos = groupCount + mapping[3];
+
+        RexNode countRef = ensureType(rexBuilder, newAggFields.get(newCountPos).getType(), newCountPos, statsFields.get(0).getType());
+        RexNode minRef = ensureType(rexBuilder, newAggFields.get(newMinPos).getType(), newMinPos, statsFields.get(1).getType());
+        RexNode maxRef = ensureType(rexBuilder, newAggFields.get(newMaxPos).getType(), newMaxPos, statsFields.get(2).getType());
+        RexNode sumRef = ensureType(rexBuilder, newAggFields.get(newSumPos).getType(), newSumPos, statsFields.get(4).getType());
+        RexNode avgRef = buildAvgExpression(newSumPos, newCountPos, statsFields.get(3).getType(), newAggFields, rexBuilder, typeFactory);
+
+        return rexBuilder.makeCall(statsType, SqlStdOperatorTable.ROW, List.of(countRef, minRef, maxRef, avgRef, sumRef));
```
nssuresh2007, line 217: Doesn't this mean that the RexNode returned would have ROW as the operator in the generated RelNode right? Just trying to understand, how would it get resolved when it would need to be converted to Substrait?

nssuresh2007 (commented):
Just a high level question - Would this logic easier to be handled in SQL plugin where we can convert stats function into the equivalent RelNode representation directly?

Eg:
```
LogicalProject [count_x, min_x, max_x, sum_x / count_x AS avg_x, sum_x]
     LogicalAggregate [COUNT(x), MIN(x), MAX(x), SUM(x)]
```

PR #21785, opened 2026-05-21
Feature/analytics can match

Bukhtawar:
### Description
 Add can-match pre-filter phase for the analytics engine. Before dispatching
  fragment execution to data-node shards, the coordinator checks Parquet row-group
  metadata statistics against the query's range predicates. Shards where no row
  group can match are eliminated, avoiding expensive full-scan execution.

  Primary use cases:
  - Time-range queries: `WHERE timestamp > '2026-01-01'` skips historical shards
  - Wildcard index patterns: `index-2026*` prunes indices outside the query range

  ### Architecture

  Coordinator: extract range predicates from plan → dispatch can-match per shard
  Data Node:   read Parquet footer stats → check column min/max vs filter range
  Response:    canMatch=true/false → coordinator skips false shards

  ### Changes

  **analytics-framework (SPI):**
  - `AnalyticsSearchBackendPlugin.canMatch(ctx, filterBytes)` — default true

  **analytics-engine (coordinator + transport):**
  - `CanMatchFilterExtractor` — walks Calcite RelNode, extracts GT/GTE/LT/LTE predicates
  - `CanMatchFilter` — serializable range predicate (column, min, max)
  - `CanMatchPreFilterPhase` — coordinator-side parallel dispatch skeleton
  - `ShardFragmentStageExecution.applyCanMatchFilter()` — integration point
  - Transport handler registered for `indices:data/read/analytics/can_match`
  - `AnalyticsSearchService.canMatch()` — delegates to backend SPI

  **analytics-backend-datafusion (Rust + Java):**
  - `can_match.rs` — reads Parquet footer, checks row-group column stats (INT32/INT64)
  - `df_can_match` FFM entry point (returns 1/0/-1)
  - `NativeBridge.canMatch()` Java downcall
  - `DataFusionAnalyticsBackendPlugin.canMatch()` override (TODO: wire file paths)

  ### Remaining TODOs
  - [ ] Pass shard file paths through CommonExecutionContext to backend
  - [ ] Coordinator transport dispatch in CanMatchPreFilterPhase
  - [ ] Timestamp literal support (Calendar/TimestampString → epoch millis)
  - [ ] INT96 timestamp support in Rust stats evaluation
  - [ ] Merge overlapping range filters on same column

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/can_match.rs
```diff
@@ -0,0 +1,110 @@
...
+pub fn can_match_range(
+    file_path: &str,
+    column_name: &str,
+    filter_min: i64,
+    filter_max: i64,
+) -> CanMatchResult {
+    // Read Parquet footer metadata directly from the file.
+    // This is cheap (~1-10KB read from end of file) and will be cached by the OS page cache.
+    // TODO: wire into the cached metadata path once CachedParquetMetaData is accessible.
+    let file = match File::open(file_path) {
```
nishchay21, line 42: This will not work on warm nodes can we rely on store where we are trying to open file on warm store will handle

nishchay21: We should do something like:
```
use object_store::{ObjectStore, path::Path};

/// Evaluate can-match using an ObjectStore (works for local AND remote).
pub async fn can_match_range(
    store: &dyn ObjectStore,
    file_path: &str,
    column_name: &str,
    filter_min: i64,
    filter_max: i64,
) -> CanMatchResult {
    let path = Path::from(file_path);

    // 1. Get file size (TieredObjectStore serves from registry — no I/O)
    let meta = match store.head(&path).await {
        Ok(m) => m,
        Err(_) => return CanMatchResult::Unknown,
    };
    if meta.size < 12 {
        return CanMatchResult::Unknown;
    }

    // 2. Read Parquet magic + footer length (last 8 bytes)
    let tail = match store.get_range(&path, (meta.size - 8)..meta.size).await {
        Ok(b) => b,
        Err(_) => return CanMatchResult::Unknown,
    };
    let footer_len = i32::from_le_bytes(tail[0..4].try_into().unwrap()) as u64;

    // 3. Read footer metadata
    let footer_start = meta.size - 8 - footer_len;
    let footer_bytes = match store.get_range(&path, footer_start..(meta.size - 8)).await {
        Ok(b) => b,
        Err(_) => return CanMatchResult::Unknown,
    };

    // 4. Parse and evaluate
    let metadata = match parquet::file::footer::decode_footer(&footer_bytes) {
        Ok(m) => m,
        Err(_) => return CanMatchResult::Unknown,
    };
    evaluate_row_group_stats(&metadata, column_name, filter_min, filter_max)
}
```

someting like https://github.com/nishchay21/OpenSearch/blob/main/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs#L168

PR #21787, opened 2026-05-21
Fix safe-commit info and replication checksum for DFA shards

ask-kamal-nayan:
### Description
Two bug fixes in the DataFormat-Aware (DFA) replication path, surfaced by DataFormatAwareReplicationIT.
  
  1. Move getSafeCommitInfo to CatalogSnapshotManager / deletion policy
  
  DataFormatAwareEngine.getSafeCommitInfo() delegated to committer.getSafeCommitInfo(), where the only implementation threw UnsupportedOperationException("TODO:: with index deleter").
  Any caller of IndexShard.getSafeCommitInfo() therefore crashed — most visibly the periodic retention-lease sync, which calls it via
  ReplicationTracker.getMinimumReasonableRetainedSeqNo() whenever a replica is missing. Stale peer-recovery leases consequently never expired.
  
  Fix: move getSafeCommitInfo() from the Committer SPI to CatalogSnapshotDeletionPolicy / CatalogSnapshotManager, where CombinedCatalogSnapshotDeletionPolicy already maintains a cached
  volatile SafeCommitInfo updated in synchronized onCommit. New chain mirrors InternalEngine → CombinedDeletionPolicy.getSafeCommitInfo() exactly:
  
  IndexShard → DataFormatAwareEngine → CatalogSnapshotManager → CatalogSnapshotDeletionPolicy
  
  Replica path (DataFormatAwareNRTReplicationEngine.getSafeCommitInfo) is unchanged. getSafeCommitInfo is dropped from the Committer interface (only one caller existed) and from the
  corresponding stubs.
  
  2. Fix validateLocalChecksum to use format-aware checksum for DFA shards
  
  AbstractSegmentReplicationTarget.validateLocalChecksum unconditionally called CodecUtil.retrieveChecksum on all files, which fails on non-Lucene formats (e.g. parquet) with
  CorruptIndexException: codec footer mismatch. This caused segment replication on DFA shards to unnecessarily delete and re-download valid local files after a crash recovery.
  
  The fix mirrors the primary-side logic in Store.checksumFromFile: unwrap the directory, and if it's a DataFormatAwareStoreDirectory, delegate to calculateUploadChecksum (which
  dispatches per format). Otherwise fall back to the existing Lucene CodecUtil path.
  
  Also removes the unused Store.getFileMetadata(Collection<String>) method and its orphaned java.util.Collection import (dead code from a prior iteration).
  

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/engine/DataFormatAwareNRTReplicationEngine.java
+++ b/server/src/main/java/org/opensearch/index/engine/DataFormatAwareNRTReplicationEngine.java
@@ -529,7 +529,7 @@ public class DataFormatAwareNRTReplicationEngine implements Indexer {
     @Override
     public GatedCloseable<CatalogSnapshot> acquireSafeCatalogSnapshot() throws EngineException {
         ensureOpen();
-        return catalogSnapshotManager.acquireSnapshot();
+        return acquireLastCommittedSnapshot(false);
     }
 
     @Override
@@ -544,7 +544,7 @@ public class DataFormatAwareNRTReplicationEngine implements Indexer {
     @Override
     public SafeCommitInfo getSafeCommitInfo() {
         ensureOpen();
-        try (GatedCloseable<CatalogSnapshot> snapshot = catalogSnapshotManager.acquireSnapshot()) {
+        try (GatedCloseable<CatalogSnapshot> snapshot = catalogSnapshotManager.acquireCommittedSnapshot(true)) {
             return new SafeCommitInfo(localCheckpointTracker.getProcessedCheckpoint(), (int) snapshot.get().getNumDocs());
         } catch (IOException e) {
             throw new UncheckedIOException(e);
@@ -991,6 +991,13 @@ public class DataFormatAwareNRTReplicationEngine implements Indexer {
             final CatalogSnapshot snapshot = lastSnapshot;
             acquiredSnapshots.add(snapshot);
             return new GatedCloseable<>(snapshot, () -> { acquiredSnapshots.remove(snapshot); });
+        }
+
+        @Override
+        public SafeCommitInfo getSafeCommitInfo() {
+            // Replicas track safe commit info on the engine itself (see DataFormatAwareNRTReplicationEngine#getSafeCommitInfo);
+            // this policy is not used to compute it.
+            return SafeCommitInfo.EMPTY;
         }
 
         /**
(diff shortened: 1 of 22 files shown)


server/src/main/java/org/opensearch/index/engine/exec/CatalogSnapshotDeletionPolicy.java
```diff
@@ -81,4 +82,12 @@ default GatedCloseable<CatalogSnapshot> acquireCommittedSnapshot(boolean acquiri
     default CatalogSnapshot findSafeCommit(List<CatalogSnapshot> commits) throws IOException {
         throw new UnsupportedOperationException("findSafeCommit not supported by this policy");
     }
+
+    /**
+     * Returns information about the safe commit, for making decisions about recoveries.
+     * Defaults to {@link SafeCommitInfo#EMPTY} for policies that do not track a safe commit.
+     */
+    default SafeCommitInfo getSafeCommitInfo() {
```
mgodwan: DAFNRE has no impl for it?
Can we not make this default?

server/src/main/java/org/opensearch/indices/replication/AbstractSegmentReplicationTarget.java
```diff
@@ -271,8 +273,19 @@ protected List<StoreFileMetadata> getFiles(CheckpointInfoResponse checkpointInfo
...
+            final String checksum;
+            if (dfasd != null) {
+                // Format-aware path: matches Store.checksumFromFile on the primary so checksums
+                // computed for non-Lucene files (e.g. parquet) can be compared directly.
+                checksum = Store.digestToString(Long.parseLong(dfasd.calculateUploadChecksum(file.name())));
+            } else {
+                try (IndexInput indexInput = directory.openInput(file.name(), IOContext.READONCE)) {
+                    checksum = Store.digestToString(CodecUtil.retrieveChecksum(indexInput));
+                }
+            }
```
mgodwan: Can this be delegated into a store method?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareNRTReplicationEngine.java
mgodwan, line 556: This should get from CatalogSnapshotManager?

mgodwan (approved):

mgodwan (approved):

PR #21788, opened 2026-05-21
Memory leak upon installing transport-reactor-netty4 plugin

### LINKED ISSUE #21717: [BUG] Memory leak upon installing `transport-reactor-netty4` plugin
### Describe the bug

In context of built-in MCP server testing, as per [blog post](https://opensearch.org/blog/introducing-mcp-in-opensearch/), I tried installing `transport-reactor-netty4` in couple of fairly inactive clusters, and set `http.type: reactor-netty4-secure` for the clusters to actually work (security plugin is enabled). Main activity in the cluster is imposed by the elasticsearch_exporter, running every 10 seconds to collect metrics.

At this stage, it doesn't matter if `"plugins.ml_commons.mcp_server_enabled": "true"` is set or not, as the leak has been observed either way, typically about 48 hours after the instalment, it becomes more evident and it never drops, unless restarted (I had to engage `kill -9` to succeed that). After the restart, the leak starts building again.

cluster 1
Image

cluster 2
Image

For what is worth, here's my full `opensearch.yml` and it's pretty identical on both clusters that I tested.

```
cluster.name: cms-compops-archi-test1
cluster.publish.timeout: 120s
cluster.routing.allocation.awareness.attributes: landb_service_name_IP
cluster.routing.allocation.node_concurrent_recoveries: 20
cluster.routing.allocation.node_initial_primaries_recoveries: 30
cluster.routing.allocation.same_shard.host: true
discovery.request_peers_timeout: 10s
discovery.seed_hosts:
- redacted:9315
- redacted:9315
- redacted:9315
http.compression: true
http.compression_level: 3
http.max_header_size: 16kb
http.port: 9215
http.type: reactor-netty4-secure
indices.recovery.max_bytes_per_sec: 250mb
network.bind_host: 0.0.0.0
network.publish_host: _enp33s0f0_
node.attr.landb_service_name_IP: IP412
node.attr.node_attr: default
node.name: osaqa101-cms-compops-archi-test1_master
node.processors: 4
node.roles:
- ingest
- remote_cluster_client
- cluster_manager
- data
path.data: "/var/lib/opensearch/cms-compops-archi-test1_master"
path.logs: "/var/log/opensearch/cms-compops-archi-test1_master"
plugins.alerting.alert_history_max_age: 30d
plugins.alerting.alert_history_max_docs: 2500000
plugins.alerting.alert_history_retention_period: 60d
plugins.alerting.alert_history_rollover_period: 1d
plugins.index_state_management.history.max_age: 30d
plugins.index_state_management.history.max_docs: 2500000
plugins.index_state_management.history.rollover_check_period: 1d
plugins.index_state_management.history.rollover_retention_period: 7d
plugins.index_state_management.job_interval: 60
plugins.security.allow_default_init_securityindex: true
plugins.security.allow_unsafe_democertificates: false
plugins.security.audit.config.log4j.level: INFO
plugins.security.audit.config.log4j.logger_name: audit
plugins.security.audit.type: log4j
plugins.security.authcz.admin_dn:
- 'CN=Robot: redacted'
plugins.security.authcz.rest_impersonation_user.superadmin:
- "*"
plugins.security.check_snapshot_restore_write_privileges: true
plugins.security.enable_snapshot_restore_privilege: true
plugins.security.kerberos.acceptor_keytab_filepath: krb5.keytab
plugins.security.kerberos.acceptor_principal: HTTP/os-redacted
plugins.security.kerberos.krb5_filepath: krb5.conf
plugins.security.nodes_dn:
- CN=*.redacted,OU=computers,DC=redacted
plugins.security.restapi.roles_enabled:
- all_access
- security_rest_api_access
plugins.security.ssl.http.enabled: true
plugins.security.ssl.http.pemcert_filepath: node-public.pem
plugins.security.ssl.http.pemkey_filepath: node-private.pem
plugins.security.ssl.http.pemtrustedcas_filepath: company-root.pem
plugins.security.ssl.transport.enabled: true
plugins.security.ssl.transport.pemcert_filepath: node-public.pem
plugins.security.ssl.transport.pemkey_filepath: node-private.pem
plugins.security.ssl.transport.pemtrustedcas_filepath: company-root.pem
plugins.security.ssl_cert_reload_enabled: true
reindex.remote.allowlist:
- os-redacted:443
- os-redacted2:443
reindex.ssl.verification_mode: none
s3.client.default.endpoint: redacted.redacted.redacted
s3.client.default.max_retries: 3
s3.client.default.path_style_access: false
s3.client.default.protocol: https
s3.client.default.read_timeout: 50s
s3.client.default.region: us-east-2
s3.client.default.use_throttle_retries: true
transport.port: 9315
transport.ssl.enforce_hostname_verification: false
transport.ssl.resolve_hostname: false
```

### Related component

Plugins

### To Reproduce

1. Bootstrap an OpenSearch cluster with security plugin enabled
2. Install `transport-reactor-netty4`
3. Add `http.type: reactor-netty4-secure` on opensearch.yml
4. Run elasticsearch exporter against the cluster to keep it busy with some requests every 10 seconds
5. Await for couple of days and observe the metrics of heap usage
### Expected behavior

no memory leak upon `transport-reactor-netty4` plugin installation.

### Additional Details

**Plugins**
All default + repository-s3 + transport-reactor-netty4

**Screenshots**

Cluster 1, barely busy, the plugin was installed 15/05 11:40 and after 24 hours you can already see a trend, but becomes more evident on the following hours.

Image

Cluster 2, again barely having any activity. I had to manually restart the nodes (with `kill -9` as it needed some force) and then the leak started building up again:

Image

 - OS: AlmaLinux 9.7
 - OpenSearch v3.4.0

**Additional context**
I am trying to use the plugin in context of setting up the built-in MCP server. It actually works as expected, and the cluster (incl. the MCP server functionality) remains operational, but the memory leak gets the cluster unresponsive at some point.

reta:
### Description
Memory leak upon installing `transport-reactor-netty4` plugin

Image

The Project Reactor creates an ephemeral connection over Netty channel all the time and runs request / response conversation over it. It does not work well with persistent connections (this is exactly what `elastisearch_exporter` is doing) as it leads to explosion of `HttpChannel` instances that are tracked by `AbstractHttpServerTransport` and eventually could cause out of memory.

Rested overnight, no issues for 24h run.

image

### Related Issues
Closes https://github.com/opensearch-project/OpenSearch/issues/21717

### Check List


reta:
@andrross @cwperks would love your feedback folks, would be great to have it in time for 3.7, thank you

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpServerTransport.java
```diff
@@ -601,4 +623,112 @@ private static void disposeQuietly(final DisposableChannel disposable) {
...
+        request.withConnection(c -> connection[0] = c);
+
+        if (connection[0] == null) {
+            throw new IllegalStateException("Failed to obtain connection from HttpServerRequest");
+        }
+
+        final ChannelGroup channels = channelGroups.computeIfAbsent(
+            connection[0].channel(),
+            key -> new ChannelGroup(key, ActionListener.wrap(() -> channelGroups.remove(key)))
+        );
```
andrross: I think there is a _tiny_ race condition here since in theory it is possible for the channel to be closed and the listener you're registering here be fired while you're still inside the execution of `computeIfAbsent()`. I think that ends up being recursive update to the same key inside computeIfAbsent, which is undefined or not allowed.

Can you separate the creation from registering the callback? something like:

```
final Channel nettyChannel = connection[0].channel();
final ChannelGroup channels = channelGroups.computeIfAbsent(nettyChannel, ChannelGroup::new);
channels.attachRemoval(() -> channelGroups.remove(nettyChannel));
```

plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpChannelsReleaseIntegTests.java
```diff
@@ -67,17 +66,9 @@ public void testAcceptedChannelsGetCleanedUpOnTheNodeShutdown() throws Interrupt
...
-        // - if additional plugins are used (like telemetry), the same HTTP channel may
-        // be wrapped around, inflating the number of HTTP channels being tracked
-        assertThat(
-            "All channels remain open",
-            RestCancellableNodeClient.getNumChannels(),
-            greaterThanOrEqualTo(initialHttpChannels + 10 /* default connections per route */)
-        );
+        // The Project Reactor creates an ephemeral connection over Netty channel all the time and run request / response conversation over
+        // it. The underlying HttpChannel / RestChannel instances are always closed once the response is delivered.
+        assertThat("All channels are closed", RestCancellableNodeClient.getNumChannels(), equalTo(initialHttpChannels));
```
andrross, line 71: Is there a potential race here? The latch only waits for the client to get the response. I'm guessing the server-side closing of channels is async?

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpServerTransport.java
```diff
@@ -601,4 +623,112 @@ private static void disposeQuietly(final DisposableChannel disposable) {
...
+            httpChannels.remove(httpChannel);
+        }
+
+        /**
+         * Closes this channel group
+         */
+        @Override
+        public void close() {
+            if (closeContext.isDone() == false) {
+                Netty4Utils.addListener(channel.close(), closeContext);
```
andrross, line 671: Wasn't this already done on line 639?

plugins/transport-reactor-netty4/src/internalClusterTest/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpChannelsReleaseIntegTests.java
reta: No, this is "sync" operation in a sense when response is sent - channel is closed

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpServerTransport.java
reta: No, this is `channel.close()` vs `channel.closeFuture()`, basically if for some reason the future did not kick in, we close the channel, thanks @andrross

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpServerTransport.java
reta: Better be safe for sure, addressed, thanks @andrross !

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpServerTransport.java
```diff
@@ -601,4 +623,104 @@ private static void disposeQuietly(final DisposableChannel disposable) {
             // Do nothing
         }
     }
+
+    /**
+     * The {@link ChannelGroup} abstraction binds {@link HttpChannel} instance to a single Netty channel and
+     * separate the lifecycle of those: {@link HttpChannel} is closed upon every response delivered whereas {@link ChannelGroup}
+     * is closed only when the Netty channel it is attached to is closed.
+     */
+    static final class ChannelGroup implements CloseableChannel {
```
cwperks: @reta the PR looks ok to me but ChannelGroup sounds semantically incorrect for what this is. With this change, if my understanding is correct we are essentially ignoring persistent connections? i.e. each request and response has a single channel that's closed on response regardless if its a persistent (`keepalive: true`?) connection.

reta: > @reta the PR looks ok to me but ChannelGroup sounds semantically incorrect for what this is.

Thanks @cwperks , open for suggestions here! What about `ParentChannel` ?

> With this change, if my understanding is correct we are essentially ignoring persistent connections? i.e. each request and response has a single channel that's closed on response regardless if its a persistent (keepalive: true?) connection. 

No no, the issue is a bit subtle. So in case of Netty transport, once the channel is initialized, it is bound to `HttpChannel` instance and the handler takes care of read / write operation. With respect to Reactor Netty, the channel abstraction is hidden behind request / response abstraction (read / write on the channel) - we created an `HttpChannel` every time for each request / response pair. We didn't ignore persistent connection, we just created too many `HttpChannel` instances.

cwperks: > We didn't ignore persistent connection, we just created too many HttpChannel instances.

Makes sense. I see the issue now.

cwperks (approved):
Thank you @reta! Approving this PR. Had one question though on the tests and whether there's a way to catch this leak through the test.

plugins/transport-reactor-netty4/src/main/java/org/opensearch/http/reactor/netty4/ReactorNetty4HttpServerTransport.java
cwperks: Maybe WrappedChannel?

reta: It is not really a wrapper to be fair: basically one Netty channel hosts zero or more `HttpChannel`, may be `HostChannel`?

reta:
> Thank you @reta! Approving this PR. Had one question though on the tests and whether there's a way to catch this leak through the test.

Thanks @cwperks ! This test now captures it https://github.com/opensearch-project/OpenSearch/pull/21788/changes#diff-9d1dcef1f14b5701aa396a29ad51838c26257d3f724985b8af25858bb7ada63dR71, but I will look if I could add something more targeted, thank you!

reta:
@andrross how does it look for you? thanks!

reta:
Merging it to make it before the code freeze, will address any feedback afterwards, thanks @cwperks @andrross !

PR #21789, opened 2026-05-21
[analytics engine] Execute analytics engine plan as a local transport action to provide ActionFilter authorization

finnegancarroll:
### Description

The analytics engine's query execution path (PPL, SQL) bypasses the security plugin's index-level privilege evaluation. The `FragmentExecutionRequest` on the data node is executed with `dispatchFragmentStreaming` and never evaluated for index permissions by the security plugin `SecurityFilter` (wraps the `ActionFilter`).

### SQL / PPL queries with analytics engine

In a typical SQL / PPL query authorization is handled at the node-to-node transport layer on the coordinator as the request is being dispatched to shards. The security plugin [provides an ActionFilter](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/OpenSearchSecurityPlugin.java#L964-L971) which every TransportAction child holds a reference to, and [executes on new requests](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/TransportAction.java#L213-L228).

The overall flow for a regular DSL search query is then:
```
NodeClient.execute(SearchAction.INSTANCE, searchRequest)
    → NodeClient.executeLocally(action, request, listener)
      → transportAction(action).execute(request, listener)
        → TransportAction.execute(task, request, listener)
          → RequestFilterChain.proceed()
```

The path for an analytics engine SQL / PPL query avoids the `RequestFilterChain` entirely. The `ShardTaskRunner` is  responsible for dispatching a `FragmentExecutionRequest` on the transport layer.

```
    @Override
    public void run(ShardStageTask task, ActionListener<Void> listener) {
        ShardExecutionTarget target = (ShardExecutionTarget) task.target();
        FragmentExecutionRequest request = requestBuilder.apply(target);
        PendingExecutions pending = pendingFor(target);
        transport.dispatchFragmentStreaming(request, target.node(), stage.responseListenerFor(listener), config.parentTask(), pending);
    }
```

The `FragmentExecutionRequest` fetches a new connection from the transport service, and [directly executes ](https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java#L193-L204)the request itself. Skipping the `TransportAction` framework which typically holds and applies the `ActionFilter`.

```
        pending.tryRun(() -> {
            try {
                Transport.Connection connection = getConnection(null, targetNode.getId());
                transportService.sendChildRequest(connection, FragmentExecutionAction.NAME, request, parentTask, options, handler);
            } catch (Exception e) {
                try {
                    listener.onFailure(e);
                } finally {
                    pending.finishAndRunNext();
                }
            }
        });
```

### DSL queries with analytics engine

DSL permissions are actually evaluated correctly already. This comes from how DSL is routed into the analytics plugin. DSL queries construct a regular `SearchRequest` which is handled by the `TransportSearchAction` and are executed on the transport layer with the local node client as usual.

Handling of DSL queries is done by a new `SearchActionFilter` which is also injected into the `RequestFilterChain` and [redirects these `SearchRequest`s to the `DslExecuteAction.INSTANCE` handler](https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/dsl-query-executor/src/main/java/org/opensearch/dsl/action/SearchActionFilter.java) instead of executing them as normal search requests. Since this `SearchActionFilter` executes AFTER the security plugin ActionFilter (which has highest priority), privileges are evaluated correctly.

| Path | Unauthorized Index | Status |
|------|-------------------|--------|
| DSL (`POST /{index}/_search`) | ✅ Denied | Secure |
| PPL (`POST /_plugins/_ppl`) | ❌ Allowed | **Gap** |
| SQL (`POST /_plugins/_sql`) | ❌ Allowed | **Gap** |

### Proposed fix

These changes somewhat mimic the functionality of DSL queries for SQL and PPL by adding a pre-planning authorization check which passes a no-op `AnalyticsAuthRequest` through the `ActionFilter` to determine if permissions are valid for the equivalent set of  `FragmentExecutionRequest`s.

No actual request is executed on the transport layer, the `AnalyticsAuthRequest` is passed through the full `ActionFilter` chain to invoke security plugin privilege evaluation, and proceeds with the analytics engine planning + query only if this probing `AnalyticsAuthRequest` is authorized for the same indices.

### Testing

These changes route all analytics engine query execution through the existing `AnalyticsQueryAction` and `TransportAction` via `NodeClient.execute()` during `DefaultPlanExecutor.execute()`. This triggers the `ActionFilter` chain (including SecurityFilter) before even planning begins.

Test implementation is pending. Currently changes are validated with local single node testing.

### Related Issues
N/A

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
@@ -15,13 +15,14 @@ import org.apache.calcite.rel.metadata.JaninoRelMetadataProvider;
 import org.apache.calcite.rel.metadata.RelMetadataQueryBase;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.opensearch.action.ActionRequest;
 import org.opensearch.action.support.ActionFilters;
 import org.opensearch.action.support.HandledTransportAction;
 import org.opensearch.action.support.TimeoutTaskCancellationUtility;
 import org.opensearch.analytics.AnalyticsPlugin;
 import org.opensearch.analytics.EngineContext;
 import org.opensearch.analytics.exec.action.AnalyticsQueryAction;
+import org.opensearch.analytics.exec.action.AnalyticsQueryRequest;
+import org.opensearch.analytics.exec.action.AnalyticsQueryResponse;
 import org.opensearch.analytics.exec.profile.ProfiledResult;
 import org.opensearch.analytics.exec.profile.QueryProfile;
 import org.opensearch.analytics.exec.profile.QueryProfileBuilder;
@@ -30,6 +31,7 @@ import org.opensearch.analytics.exec.task.AnalyticsQueryTaskRequest;
 import org.opensearch.analytics.planner.CapabilityRegistry;
 import org.opensearch.analytics.planner.PlannerContext;
 import org.opensearch.analytics.planner.PlannerImpl;
+import org.opensearch.analytics.planner.RelNodeUtils;
 import org.opensearch.analytics.planner.dag.BackendPlanAdapter;
 import org.opensearch.analytics.planner.dag.DAGBuilder;
 import org.opensearch.analytics.planner.dag.FragmentConversionDriver;
@@ -40,7 +42,6 @@ import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.unit.TimeValue;
 import org.opensearch.core.action.ActionListener;
-import org.opensearch.core.action.ActionResponse;
 import org.opensearch.search.SearchService;
 import org.opensearch.tasks.Task;
 import org.opensearch.tasks.TaskManager;
@@ -67,7 +68,7 @@ import static org.opensearch.action.search.TransportSearchAction.SEARCH_CANCEL_A
  *
  * @opensearch.internal
  */
-public class DefaultPlanExecutor extends HandledTransportAction<ActionRequest, ActionResponse>
+public class DefaultPlanExecutor extends HandledTransportAction<AnalyticsQueryRequest, AnalyticsQueryResponse>
     implements
         QueryPlanExecutor<RelNode, Iterable<Object[]>> {
 
@@ -98,9 +99,7 @@ public class DefaultPlanExecutor extends HandledTransportAction<ActionRequest, A
         Scheduler scheduler,
         ArrowAllocatorService allocatorService
     ) {
-        super(AnalyticsQueryAction.NAME, transportService, actionFilters, in -> {
-            throw new UnsupportedOperationException("Transport path not implemented yet");
-        });
+        super(AnalyticsQueryAction.NAME, transportService, actionFilters, AnalyticsQueryRequest::new);
         this.capabilityRegistry = capabilityRegistry;
         this.clusterService = clusterService;
         this.searchExecutor = threadPool.executor(ThreadPool.Names.SEARCH);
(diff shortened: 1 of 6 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryRequest.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public void writeTo(StreamOutput out) throws IOException {
+        throw new UnsupportedOperationException("AnalyticsQueryRequest is local-dispatch only");
+    }
+
+    public RelNode getPlan() {
+        return plan;
+    }
+
+    public Object getContext() {
```
expani, line 56: nit: I don't see this getting called anywhere

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -111,21 +114,45 @@ public DefaultPlanExecutor(
             .addSettingsUpdateConsumer(AnalyticsPlugin.COORDINATOR_BUFFER_LIMIT, v -> perQueryBufferLimit = v);
     }
 
+    private static final int MAX_COLLECT_INDICES_DEPTH = 15;
```
cwperks: What is this number for?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -245,4 +265,24 @@ static Iterable<Object[]> batchesToRows(Iterable<VectorSchemaRoot> batches) {
...
+    private static void collectIndices(RelNode node, java.util.Set<String> indices, int depth) {
+        if (depth >= MAX_COLLECT_INDICES_DEPTH) {
+            return;
+        }
+        if (node instanceof org.apache.calcite.rel.core.TableScan scan) {
+            List<String> names = scan.getTable().getQualifiedName();
+            indices.add(names.get(names.size() - 1));
+        }
+        for (RelNode input : node.getInputs()) {
+            collectIndices(input, indices, depth + 1);
```
cwperks: For my own understanding, what is the purpose of recursion here? Is there no better way to extract the indices?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
finnegancarroll: The `RelNode` which describes a query is a tree structure with leaves describing discrete query components (filter/sort/join/ect). To get all of the indices which this `RelNode` operates on we end up recursing through this tree structure.

on the stale PR @expani pointed out that recursing with no max depth could be problematic and we should guard max depth here:
https://github.com/opensearch-project/OpenSearch/pull/21750#discussion_r3279956161

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -245,4 +262,24 @@ static Iterable<Object[]> batchesToRows(Iterable<VectorSchemaRoot> batches) {
         }
         return rows;
     }
+
+    /** Extracts index names from TableScan nodes in the plan. */
+    private static String[] extractIndices(RelNode plan) {
+        java.util.Set<String> indices = new java.util.LinkedHashSet<>();
```
expani: nit: Import statements :)

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -245,4 +262,24 @@ static Iterable<Object[]> batchesToRows(Iterable<VectorSchemaRoot> batches) {
...
+        collectIndices(plan, indices, 0);
+        return indices.toArray(String[]::new);
+    }
+
+    private static void collectIndices(RelNode node, java.util.Set<String> indices, int depth) {
```
expani: We can move this to `RelNodeUtils` as this will be a useful common helper for others.

Also, add some tests using complex operators like Aggregates/Joins/Unions to ensure the table names extracted are appropriate.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
expani: RelNode are logical plans in relational algebra that are recursive by nature. For instance, a query like 
```
select url, request_time from http_logs where verb = 'GET'  order by size limit 100;
```

Will roughly generate a plan like 
```
Project(url, request_time)
 Sort(size, fetch=100)
   Filter(verb='GET')
     Scan(http_logs)
```

The recursion is to reach the Scan and extract the table name out.

cwperks: Thanks for the explanation @expani !

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryRequest.java
finnegancarroll: Yea initially i didn't wrap this at all but am told it will be used in the future. I can remove the getter for now until it is actually needed.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
cwperks: +1 let's add tests for this with different expected scenarios

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
finnegancarroll: Ended up making the depth check strict and throwing an exception to avoid any case where we stop recursing and miss an index in the `RelNode`.

finnegancarroll:
Associated initial IT suite in SQL plugin:
https://github.com/opensearch-project/sql/pull/5462

Currently still running these locally and working on having them run correctly in CI with all analytics plugin + security dependencies.

cwperks (approved):
This PR LGTM! Looks like there's now a merge conflict that needs to be resolved. TY for fixing this @finnegancarroll !

PR #21794, opened 2026-05-21
Convert varchar to str in analytics engine Project operations

Swiddis:
### Description
String literals in some contexts of PPL queries produce VarChars instead of Strings. In the ITs, most prominently as part of `addtotals ... label='Sum'`. VarChar isn't a recognized type by Datafusion (needs to be a Utf8View per https://datafusion.apache.org/user-guide/sql/data_types.html), causing an error resembling `Unsupported literal_type: Some(VarChar(VarChar { value: $V, length: $N }))`. So we need a conversion layer for Project operations that converts these to strings.

### Related Issues
Integ tests for SQL against the AE backend

### Check List


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
```diff
@@ -73,6 +73,33 @@ public Optional<Rel> visit(io.substrait.relation.Filter filter, EmptyVisitationC
             );
         }
 
+        // Rewrite expressions inside project expressions (where literals often appear)
+        @Override
+        public Optional<Rel> visit(io.substrait.relation.Project project, EmptyVisitationContext ctx) {
+            Optional<Rel> newInput = project.getInput().accept(this, ctx);
+            java.util.List<Expression> oldExpressions = project.getExpressions();
```
vinaykpud: nit: drop java.util prefixes

expani:
Thanks for getting this started @Swiddis 

I am assuming this is applicable for the literals across all the ScalarFunctions emitted by PPL. 

My 2 cents is that the rewrite can be performed at the Calcite layer itself but the problem is that all Rewrites are currently tied to a ScalarFunction via `ScalarFunctionAdapter` 

Given that you want this to be applied in a generic manner independent of the ScalarFunction, we would need to provide a LiteralAdapter for expressions in Project at https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java#L184-L254

If you wish to handle it in a different PR, we can start with adding a TODO so it's not missed.

Swiddis:
image
interesting failure

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanRewriter.java
```diff
@@ -73,6 +74,36 @@ public Optional<Rel> visit(io.substrait.relation.Filter filter, EmptyVisitationC
             );
         }
 
+        // Rewrite expressions inside project expressions (where literals often appear)
+        @Override
+        public Optional<Rel> visit(Project project, EmptyVisitationContext ctx) {
```
expani, line 79: Do we need similar handling for literals in Filter and other applicable operators ? 

If you wish to take it up separately, please add a TODO as well. However, I feel given the existing code should be straight forward.

expani (approved):

mch2 (approved):

PR #21660, opened 2026-05-14
[analytics engine] Add /_analytics/ppl/_explain endpoint with stage profiling

finnegancarroll:
### Description

Adds an explain/profile API to the analytics engine, accessible via `POST /_analytics/ppl/_explain`. The endpoint executes the query and returns per-stage timing from the coordinator's perspective alongside normal results.

Follow up work which is not included in this PR:
- Integrate with SQL plugin's `RestUnifiedQueryAction.explain()` to use `executeWithProfile`
- Data-node-level profiling for fragments executed on shards

### Response format
Full response format with SQL plugin ([SQL PR](https://github.com/opensearch-project/sql/pull/5442)) with some test data.
```json
{
    "columns": [
        "avg(score)",
        "name"
    ],
    "rows": [
        [
            92.7,
            "eve"
        ],
        [
            91.0,
            "carol"
        ],
        [
            95.5,
            "alice"
        ],
        [
            88.1,
            "dave"
        ],
        [
            87.3,
            "bob"
        ]
    ],
    "profile": {
        "query_id": "a3009ddf-629b-45bc-ac61-5ab607d38cf5",
        "full_plan": [
            "OpenSearchProject(avg(score)=[$1], name=[$0], viableBackends=[[datafusion]])",
            "  OpenSearchProject(name=[$0], avg(score)=[ANNOTATED_PROJECT_EXPR(id=2, backends=[datafusion], /($1, $2))], viableBackends=[[datafusion]])",
            "    OpenSearchAggregate(group=[{0}], agg#0=[SUM(AGG_CALL_ANNOTATION(id=0, viableBackends=[datafusion]), $1)], agg#1=[COUNT(AGG_CALL_ANNOTATION(id=1, viableBackends=[datafusion]), $1)], mode=[FINAL], viableBackends=[[datafusion]])",
            "      OpenSearchExchangeReducer(viableBackends=[[datafusion]], exchange=[ExchangeInfo[distributionType=SINGLETON, partitionKeyIndices=[]]])",
            "        OpenSearchAggregate(group=[{0}], agg#0=[SUM(AGG_CALL_ANNOTATION(id=0, viableBackends=[datafusion]), $1)], agg#1=[COUNT(AGG_CALL_ANNOTATION(id=1, viableBackends=[datafusion]), $1)], mode=[PARTIAL], viableBackends=[[datafusion]])",
            "          OpenSearchProject(name=[$1], score=[$2], viableBackends=[[datafusion]])",
[...]

PR #21796, opened 2026-05-21
[analytics engine] Add `/_plugins/_analytics/stats` endpoint

OVI3D0:
### Description
Builds ontop of #21660 and adds per stage timing but rolled up so users won't need to hit `_explain` for every request

Example:
```
$ curl -s 'http://localhost:9200/_plugins/_analytics/stats' | jq
{
  "analytics": {
    "queries": {
      "elapsed_ms":  { "count": 46, "sum_ms": 615 },
      "planning_ms": { "count": 46, "sum_ms": 504 }
    },
    "stages_by_type": {
      "SHARD_FRAGMENT": {
        "started": 46, "succeeded": 46, "failed": 0, "cancelled": 0,
        "rows_processed_total": 334,
        "elapsed_ms": { "count": 46, "sum_ms": 615 }
      }
    },
    "fragments": {
      "total": 46, "succeeded": 46, "failed": 0,
      "elapsed_ms": { "count": 46, "sum_ms": 613 }
    }
  }
}
```

Some follow ups that can be added:
- `TransportNodesAction` to broadcast/merge results, since right now this API is per-node only
- Planning-time failures don't increment `queries.failed` — query lifecycle events fire from `QueryScheduler.execute()`, but parse / table-resolution / planning failures fail upstream of that

Also separately adds a `SearchStatsContributor` so analytics-engine queries surface in `_nodes/stats` `search.query_total` / `query_time_in_millis`. This means existing tooling can see Mustang traffic without changes.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
@@ -27,6 +27,7 @@ import org.opensearch.analytics.planner.CapabilityRegistry;
 import org.opensearch.analytics.planner.FieldStorageResolver;
 import org.opensearch.analytics.schema.OpenSearchSchemaBuilder;
 import org.opensearch.analytics.spi.AnalyticsSearchBackendPlugin;
+import org.opensearch.analytics.stats.AnalyticsStats;
 import org.opensearch.analytics.stats.AnalyticsStatsCollector;
 import org.opensearch.analytics.stats.RestAnalyticsStatsAction;
 import org.opensearch.arrow.allocator.ArrowNativeAllocator;
@@ -47,10 +48,12 @@ import org.opensearch.core.common.io.stream.NamedWriteableRegistry;
 import org.opensearch.core.xcontent.NamedXContentRegistry;
 import org.opensearch.env.Environment;
 import org.opensearch.env.NodeEnvironment;
+import org.opensearch.index.search.stats.SearchStats;
 import org.opensearch.plugins.ActionPlugin;
 import org.opensearch.plugins.ExtensiblePlugin;
 import org.opensearch.plugins.Plugin;
 import org.opensearch.plugins.PluginComponentRegistry;
+import org.opensearch.plugins.SearchStatsContributor;
 import org.opensearch.repositories.RepositoriesService;
 import org.opensearch.rest.RestController;
 import org.opensearch.rest.RestHandler;
@@ -74,7 +77,7 @@ import java.util.function.Supplier;
  *
  * @opensearch.internal
  */
-public class AnalyticsPlugin extends Plugin implements ExtensiblePlugin, ActionPlugin {
+public class AnalyticsPlugin extends Plugin implements ExtensiblePlugin, ActionPlugin, SearchStatsContributor {
 
     private static final Logger logger = LogManager.getLogger(AnalyticsPlugin.class);
 
@@ -211,6 +214,16 @@ public class AnalyticsPlugin extends Plugin implements ExtensiblePlugin, ActionP
 
     static int schedulerPoolSize() {
         return Math.max(2, Runtime.getRuntime().availableProcessors() / 2);
+    }
+
+    @Override
+    public SearchStats contributeSearchStats() {
+        AnalyticsStats.LatencyStats elapsed = statsCollector.snapshot().queries().elapsedMs();
+        if (elapsed.count() == 0) {
+            return null;
+        }
+        SearchStats.Stats stats = new SearchStats.Stats.Builder().queryCount(elapsed.count()).queryTimeInMillis(elapsed.sumMs()).build();
+        return new SearchStats(stats, 0, null);
     }
 
     @Override
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
@@ -27,15 +27,21 @@ import org.opensearch.analytics.planner.CapabilityRegistry;
 import org.opensearch.analytics.planner.FieldStorageResolver;
 import org.opensearch.analytics.schema.OpenSearchSchemaBuilder;
 import org.opensearch.analytics.spi.AnalyticsSearchBackendPlugin;
+import org.opensearch.analytics.stats.AnalyticsStatsCollector;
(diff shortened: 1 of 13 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -140,12 +145,23 @@ public void executeFragmentStreamingAsync(
     ) {
         try {
             executor.execute(() -> {
+                final long startNanos = System.nanoTime();
+                long rows = 0L;
                 try (FragmentResources ctx = executeFragmentStreaming(request, shard, task)) {
                     Iterator<EngineResultBatch> it = ctx.stream().iterator();
                     while (it.hasNext()) {
-                        responseHandler.onBatch(it.next());
+                        EngineResultBatch batch = it.next();
```
finnegancarroll: Do we need this? `ProfileResult` should already include the total rows processed here right? Is this a redundant count?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -157,14 +161,22 @@ private void executeInternal(RelNode logicalFragment, boolean profile, ActionLis
...
+        final long planStartNanos = System.nanoTime();
         RelNode plan = PlannerImpl.createPlan(logicalFragment, new PlannerContext(capabilityRegistry, clusterService.state()));
         final String fullPlan = profile ? org.apache.calcite.plan.RelOptUtil.toString(plan) : null;
         QueryDAG dag = DAGBuilder.build(plan, capabilityRegistry, clusterService);
         PlanForker.forkAll(dag, capabilityRegistry);
         BackendPlanAdapter.adaptAll(dag, capabilityRegistry);
         FragmentConversionDriver.convertAll(dag, capabilityRegistry);
-        final long planningTimeMs = profile ? java.util.concurrent.TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - planStartNanos) : 0;
+        final long planningTimeNanos = System.nanoTime() - planStartNanos;
+        final long planningTimeMs = java.util.concurrent.TimeUnit.NANOSECONDS.toMillis(planningTimeNanos);
```
finnegancarroll: I think planning time should be available in the `ProfileResult` as well and we shouldn't need to capture it separately in these `AnalyticsOperationListener`s.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/stats/AnalyticsStatsCollector.java
```diff
@@ -0,0 +1,167 @@
...
+/**
+ * Node-local {@link AnalyticsOperationListener} that aggregates query / stage / fragment
+ * lifecycle events into counters surfaced via the {@code _stats} REST endpoint.
+ *
+ * <p>Recording is wait-free: every counter is a {@link LongAdder}, every latency is a
+ * {@link LatencyHistogram} (HdrHistogram-backed Recorder), and the per-stage-type map is
+ * a {@link ConcurrentHashMap}. {@link #snapshot()} reads each counter once and folds the
+ * latest histogram interval into the cumulative accumulator.
+ */
+public final class AnalyticsStatsCollector implements AnalyticsOperationListener {
```
finnegancarroll: Is this redundant with the `QueryProfileBuilder`/`QueryProfile`? Each `ExecutionGraph` independently tracks execution timings for each stage already and makes them available in these profiles at the end of query execution.

Instead of attaching this second listener can we fetch everything we need from the `QueryProfile` at the end of execution within the `DefaultPlanExecutor`? We should always build this `QueryProfile` regardless of whether we are executing through the explain api or regular path.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
finnegancarroll: >TransportNodesAction to broadcast/merge results, since right now this API is per-node only

Instead of creating a new transport action for this I'm wondering if implementing the extension point for the existing opensearch stats API can help here.

If the stats API extension points are added:
https://github.com/opensearch-project/OpenSearch/pull/21809/changes#diff-92aa74bd95173ec7b0662a3e215f99d6d9679d1446a78dc1f3f836b6ff4ef9b3R192-R201

Then `/_plugins/_analytics/stats` can piggyback off of this existing API and just filter only the analytics engine related stats using.

```
NodesStatsRequest request = new NodesStatsRequest();
client.admin().cluster().nodesStats(request, listener);

# Filter response for only analytics engine stats
```

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
OVI3D0: removed

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
OVI3D0: yep i see, removed

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
OVI3D0: discussed offline & building a new transport action seems like the cleanest solution, will add this in a follow up PR

finnegancarroll (approved):

sandeshkr419 (changes requested):
`QueryProfileBuilder.snapshot()` now runs on every query (not just _explain) to feed the stats collector, is that correct? If so, could we instead accumulate just the raw counters (elapsed, planning_ms, stage counts) in the stats collector without building the full QueryProfile object on the hot path?

Basically I want to avoid any unnecessary overhead.

OVI3D0:
> `QueryProfileBuilder.snapshot()` now runs on every query (not just _explain) to feed the stats collector, is that correct? If so, could we instead accumulate just the raw counters (elapsed, planning_ms, stage counts) in the stats collector without building the full QueryProfile object on the hot path?
> 
> Basically I want to avoid any unnecessary overhead.

yeah, i can add a method to walk the execution graph and only build the `QueryProfile` on the _explain path

sandeshkr419 (approved):
Thanks for working on this and addressing the concerns.

PR #21797, opened 2026-05-21
PPL patterns command (BRAIN + SIMPLE) on the analytics-engine route

ahkcs:
## What

Wires PPL `patterns` (BRAIN + SIMPLE methods, label + aggregation modes) through the analytics-engine route, including the OpenSearch Dashboards BRAIN-label panel query:

```ppl
patterns message method=brain mode=label
| stats count() as pattern_count, take(message, 1) as sample_logs by patterns_field
| sort - pattern_count
| fields patterns_field, pattern_count, sample_logs
```

Companion to **opensearch-project/sql#5467** (`UnifiedQueryContext` PATTERN_* defaults + dashboard IT + planner unit test). Both PRs are required.

## Changes

**Rust / DataFusion** (`sandbox/plugins/analytics-backend-datafusion`)
- BRAIN classifier ported to Rust (scaffold → full classifier → scalar eval functions → `PATTERN_PARSER` UDF wiring).
- `internal_pattern` window UDF (BRAIN label mode) + aggregate UDF (BRAIN aggregation mode).
- Per-field scalar UDFs `pattern_parser_get_pattern` / `pattern_parser_get_tokens` — workaround for DataFusion's substrait consumer's "Direct reference StructField with child not supported".
- `SAFE_CAST` / `ITEM` overloads for ARRAY + MAP returns; `FORCE_NULLABLE` on `LOCAL_TAKE_OP` / `LOCAL_ARRAY_AGG_OP`.

**Analytics-engine planner** (`sandbox/plugins/analytics-engine`)
- `PplWindowCallRewriter` — bottom-up RelNode shuttle that retypes `RexInputRef`s before `copy()` so the BRAIN window-UDF result type propagates through enclosing Projects.
- `OpenSearchAggregateSplitRule` — skip PARTIAL/FINAL split when STATE_EXPANDING aggregates (TAKE/FIRST/LAST/LIST/VALUES) are present (avoids argList-shift crash on `take(field, 1)` in stats).
- `ArrowValues.toJavaValue` — recursively unwrap Arrow `Text` so nested `Map<String, List<String>>` returns hand the SQL layer pure Java types.

**DataFusion `regexp_replace` adapter** (`RegexpReplaceAdapter`)
- Append `"g"` flag to every 3-arg `REGEXP_REPLACE` going to DataFusion. DataFusion's `regexp_replace` defaults to first-match-only; Calcite's 3-arg form is already replace-all. The append preserves PPL contract on both backends — no SQL-core knowledge of DataFusion semantics required (companion change in opensearch-project/sql#5467 keeps SQL core on plain 3-arg).

**Capabilities + substrait wiring**
- `WindowFunction.PATTERN`, `AggregateFunction.PATTERN`.
- `opensearch_window_functions.yaml` registers `internal_pattern` window-UDF variants.
- `DataFusionFragmentConvertor` operator declarations + `ADDITIONAL_*_SIGS` for isthmus binding.

## Results

| Suite | Before | After |
|---|---|---|
| `CalcitePPLDashboardPatternsIT` (new) | — | **1 / 1** |
| `CalcitePPLPatternsIT` (analytics-engine route) | 3 / 15 | **10 / 15** |

Remaining 5 / 15: 3 × `BrainAggregationMode_*` (planner needs `LogicalCorrelate` post-aggregate support) + 2 × `Brain*Mode_ShowNumberedToken` (substrait type wiring for `_ShowNumberedToken`). Tracked separately, none affect the dashboard query.


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchProjectRule.java
```diff
@@ -301,6 +301,10 @@ private static Set<WindowFunction> collectWindowFunctions(List<? extends RexNode
                 @Override
                 public RexNode visitOver(RexOver over) {
                     WindowFunction fn = WindowFunction.fromSqlKind(over.getAggOperator().getKind());
+                    if (fn == null) {
+                        // OTHER-kind operators (PPL's INTERNAL_PATTERN, etc.) — fall back to name lookup.
+                        fn = WindowFunction.fromName(over.getAggOperator().getName());
```
mch2: nitpick - we should put this in WindowFunction.resolveFunction or similar.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -95,6 +95,24 @@ private static boolean shouldSkipPartialFinalSplit(OpenSearchAggregate aggregate
...
+            // input is the PARTIAL output where those positions hold STATE columns
+            // of different types (count_state BIGINT, take_state ARRAY<VARCHAR>).
+            // The operator's return-type inference applied to the shifted args no
+            // longer agrees with the declared output type — the rule throws before
+            // DistributedAggregateRewriter (which DOES remap correctly) can run.
+            // Skip the split for these; the SINGLE-on-SINGLETON alternative still
+            // executes correctly and is the only viable choice on a single shard
+            // anyway. Distributed parallelism for state-expanding aggs is a
+            // follow-up (requires per-phase operator overloads or constructing the
+            // FINAL with remapped argList + explicit type here).
```
mch2: is this comment correct? the ITs look like they exercise multi shard?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchProjectRule.java
ahkcs: Done — extracted to `WindowFunction.resolveFunction(SqlOperator)` (SqlKind lookup + name fallback for OTHER kinds). `OpenSearchProjectRule` now calls it directly. Added unit tests for all three branches.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
ahkcs: Good catch — comment was stale. Updated

mch2 (approved):

PR #21799, opened 2026-05-22
coordinator reduce: adaptive memory budget via phantom reservation

mch2:
Apply the same adaptive budget system to coordinator-reduce LocalSessions that shard scans already use. The reduce path previously created sessions with no phantom reservation — all concurrent reduces shared the 128MB DataFusion pool without admission control, allowing unbounded sessions that exhausted the pool and prevented spill.

Changes:
- create_local_session now calls acquire_budget_inner_for_reduce() which tries target_partitions=4 and adaptively reduces (4→2→1) under memory pressure, same as the shard scan path
- LocalSession.new_with_budget() accepts target_partitions + batch_size from the budget instead of hardcoding defaults
- Add acquire_budget_inner_for_reduce() to query_budget.rs using fixed estimates for reduce input (200 bytes/row, 6 columns)
- Phantom reservation held for session lifetime, released on close

This ensures the pool tracks concurrent reduce memory upfront and triggers backpressure before all sessions pile up and OOM.

### Description

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/query_budget.rs
```diff
@@ -278,6 +278,21 @@ fn acquire_budget_inner(
...
+/// and applies the same halving logic as the shard path.
+pub fn acquire_budget_inner_for_reduce(
+    pool: &Arc<dyn MemoryPool>,
+    configured_target_partitions: usize,
+    configured_batch_size: usize,
+) -> Result<QueryMemoryBudget, DataFusionError> {
+    // Reduce input is partially-aggregated Arrow batches. Estimate ~200 bytes/row
+    // (group keys + accumulators) with ~6 columns equivalent.
+    let avg_row_bytes: usize = 200;
+    let num_columns: usize = 6;
```
Bukhtawar: do we want this to be configurable

Bukhtawar (approved):
LGTM, minor suggestion

sandbox/plugins/analytics-backend-datafusion/rust/src/query_budget.rs
mch2: Yeah will add this. We re-size these almost immediately though based on the actual partial schema when we register partitions, and sets it as the largest partition's size. Java side orchestrates this creation in phases. These are used as defaults when we initially create the session. This could be registered based on schema up front but would rather set to the largest partition given for join cases one partition could return more cols. there's an issue here that we could be over reserving if no partition exceeds the defaults never under.

sandbox/plugins/analytics-backend-datafusion/rust/src/query_budget.rs
```diff
@@ -278,6 +278,21 @@ fn acquire_budget_inner(
 }
 
 
+/// Adaptive budget for coordinator-reduce sessions. Uses a fixed row-byte
+/// estimate (reduce consumes pre-aggregated Arrow batches, not raw parquet)
+/// and applies the same halving logic as the shard path.
+pub fn acquire_budget_inner_for_reduce(
```
aravindsagar: The commit message says it's called in `create_local_session` but I don't see this being called at all

sandbox/plugins/analytics-backend-datafusion/rust/src/query_budget.rs
mch2: updating to configurable.

sandbox/plugins/analytics-backend-datafusion/rust/src/query_budget.rs
mch2: doh, dead code thanks.

mch2:
[Race Condition](https://github.com/opensearch-project/OpenSearch/pull/21799/files#diff-f596150be09d3e677637152e19120e7e833843b0947c3cc7f602ec772345d6f7R99-R106) - not valid, we always invoke registerPartitionStream sequentially from java side before query execution starts.

mch2:
@Bukhtawar @aravindsagar I've cleaned this up a bit.  At session registration we don't actually create any phantom reservation, we'll create and add one when partitions are registered from the java side.  I've wired this up to also reuse the existing methods in query_budget, and set the partitions to be adaptive to mem pressure as with shard scans.  It will start with 4 and reduce.

aravindsagar (approved):

PR #21469, opened 2026-05-04
Add live version map in DataFormatAwareEngine.

thorkous:
### Description

Integrates `LiveVersionMap` into `DataFormatAwareEngine`.

**What this PR does:**

- Adds a `LiveVersionMap` to `DataFormatAwareEngine` and wires it into the `IndexingStrategyPlanner` (replacing the no-op `(a, b) -> null` version resolver)
- Introduces `DataFormatVersionValue`, extending `IndexVersionValue` with `rowId` (0-based row position in the parquet file) and a `Writer` reference, enabling the delete path to locate documents in format-specific storage. 
- Stores version entries in the map after successful indexing, keyed by document `_id`
- Adds version map lifecycle hooks in the refresh path: `beforeRefresh()` / `afterRefresh()` / `pruneTombstones()`
- Implements `resolveDocVersion()` for version lookups. 
- Implements `getVersionFromMap()` . 
- Adds `rowId` field to `WriteResult.Success` so writers can report the row position back to the engine

### Related Issues
Resolves #[Issue number]

### Check List
- [ ] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21803, opened 2026-05-22
Add non-Lucene get-by-id via DocumentLookupProvider SPI and integrated version-map. 

thorkous:
## Description

Adds version resolution and get-by-id document lookup to `DataFormatAwareEngine`, enabling the engine to detect whether a document already exists before indexing. This is needed for update semantics (delete-then-reindex) in the pluggable dataformat path.

### What this does

- Wires `resolveDocVersion()` in `DataFormatAwareEngine` with 
  -  Check the in-memory `LiveVersionMap` (fast path, same as `InternalEngine`)
  - Fall through to `DocumentLookupProvider.getById()` when the version map has no entry
- Implements `DocumentLookupProvider` in `DataFusionPlugin`, backed by `GetService` which:
  - Looks up the `_id` in the Lucene secondary index to find the segment generation and row ID
  - Fetches the document row from the parquet file via the native DataFusion runtime
  - Returns version, seqNo, and primaryTerm for the planner to make update/delete decisions

## Related Issues

Part of the pluggable dataformat update/delete support epic.

## Check List

---


server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1679,6 +1725,58 @@ private void closeReaders() throws IOException {
         }
     }
 
+    protected VersionValue resolveDocVersion(final Engine.Operation op, boolean loadSeqNo) throws IOException {
+        assert incrementVersionLookup();
+        VersionValue versionValue = getVersionFromMap(op.uid().bytes());
+        if (versionValue == null) {
+            if (documentLookupProvider == null) {
+                return null;
```
RS146BIJAY, line 2190: So if documentLookupProvider is null, will we keep on appending the document in the shard?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1679,6 +1725,58 @@ private void closeReaders() throws IOException {
...
+            }
+            assert incrementIndexVersionLookup();
+            DocumentLookupResult lookupResult = getById(new Engine.Get(false, false, op.id(), op.uid()));
```
RS146BIJAY: Are we skipping the entire CAS integration here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/InternalEngine.java#L800 as done in InternalEngine?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -376,13 +399,13 @@ public DataFormatAwareEngine(EngineConfig engineConfig) {
             this.indexingStrategyPlanner = new IndexingStrategyPlanner(
                 engineConfig.getIndexSettings(),
                 engineConfig.getShardId(),
-                new LiveVersionMap(),
+                this.versionMap,
```
RS146BIJAY, line 449: Are we not rotating the version map here on refresh?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -376,13 +399,13 @@ public DataFormatAwareEngine(EngineConfig engineConfig) {
             this.indexingStrategyPlanner = new IndexingStrategyPlanner(
```
RS146BIJAY, line 446: What will happen to the version map incase reboot node happens? Are we not repopulating version map: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/InternalEngine.java#L343?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareReadOnlyEngine.java
```diff
@@ -305,6 +306,11 @@ public Engine.NoOpResult noOp(Engine.NoOp noOp) throws IOException {
         throw new UnsupportedOperationException("DataFormatAwareReadOnlyEngine does not support no-ops");
     }
 
+    @Override
+    public DocumentLookupResult getById(Engine.Get get) throws IOException {
```
RS146BIJAY: Why getById is not supported for read only engine? ReadOnlyEngine should only prevent writes. Why we are blocking reads as well?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Yes, because lucene is not present for that index and then it becomes a append only case.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareReadOnlyEngine.java
RS146BIJAY: I guess this should be similar to ReadOnlyEngine which allows getById

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: I had raised these changes in my prev PR https://github.com/opensearch-project/OpenSearch/pull/21469/changes but missed here. Let me add them again.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareReadOnlyEngine.java
thorkous: DataFormatAwareReadOnlyEngine was raised and merged after I raised my PR, I haven't integrated my changes in DataFormatAwareReadOnlyEngine till yet.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Do we need CAS integration in dataformatAwareEngine? 

If we want CAS integration then we will require changes in writer because we need to validate ImmutableField not Updated. 

Do let me know your thoughts on this? We can also have CAS integration in phase 2 because we have no support in dataFormatAwareEngine as of now.

RS146BIJAY: Yeah. I think it may not be a mandatory thing. We can skip it for now.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Added a function called restoreVersionMapAndCheckpointTracker.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Done in new changes.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareReadOnlyEngine.java
thorkous: Integrated.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1684,6 +1751,101 @@ private void closeReaders() throws IOException {
...
 
+    protected VersionValue resolveDocVersion(final Engine.Operation op, boolean loadSeqNo) throws IOException {
+        assert incrementVersionLookup();
+        VersionValue versionValue = getVersionFromMap(op.uid().bytes());
+        if (versionValue == null) {
+            if (documentLookupProvider == null) {
+                return null;
+            }
+            assert incrementIndexVersionLookup();
+            DocumentLookupResult lookupResult = getById(new Engine.Get(false, false, op.id(), op.uid()));
```
RS146BIJAY: I don't think we should call getById here. This should be dependent on reader and documentLookupProvider directly

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1523,6 +1569,26 @@ public byte[] serializeSnapshotToRemoteMetadata(CatalogSnapshot catalogSnapshot)
...
 
+    /**
+     * Delegates get-by-id to the installed {@link DocumentLookupProvider}, acquiring a
+     * per-format reader on the current snapshot and passing it to the plugin.
+     * Throws {@link UnsupportedOperationException} if no plugin is wired.
+     */
+    @Override
+    public DocumentLookupResult getById(Engine.Get get) throws IOException {
+        ensureOpen();
+        if (documentLookupProvider == null) {
```
RS146BIJAY: This does not have translog integration as well. Also this seem to missing multiple version conflict checks: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/InternalEngine.java#L649

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1684,6 +1751,101 @@ private void closeReaders() throws IOException {
...
+                + versionValue.term;
+            return OpVsEngineDocStatus.OP_STALE_OR_EQUAL;
+        } else {
+            return OpVsEngineDocStatus.OP_STALE_OR_EQUAL;
+        }
+    }
+
+    private void restoreVersionMapAndCheckpointTracker() {
+        try {
+            List<DocMetadata> docs = indexingExecutionEngine.getDocsAboveCheckpoint(localCheckpointTracker.getPersistedCheckpoint());
```
RS146BIJAY: I think this should be using document look up provider.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: What's the difference in calling getById and directly documentLookUpProvider? 

We will require checks like `documentLookupProvider == null` and `documentLookupProvider.getById(get, readerRef.get(), shardId.getIndexName())` which is same logic in getById?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: restoreVersionMapAndCheckpointTracker needs to scan all docs above the persisted checkpoint — it's a range query over `_seq_no`, not a point lookup by `_id.` DocumentLookupProvider.getById() is a point-lookup API that goes through the native DataFusion runtime to fetch from parquet. Using it here would require already knowing which doc IDs to restore (chicken-and-egg problem) and would make unnecessary calls to the parquet layer when all the metadata we need (_seq_no, _version, _primary_term, _id) lives in the Lucene secondary index. The current approach queries Lucene directly via IndexingExecutionEngine.getDocsAboveCheckpoint() which seems correct and efficient for this use case.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Done

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1696,6 +1741,81 @@ public byte[] serializeSnapshotToRemoteMetadata(CatalogSnapshot catalogSnapshot)
...
     }
 
+    /**
+     * Delegates get-by-id to the installed {@link DocumentLookupProvider}, acquiring a
+     * per-format reader on the current snapshot and passing it to the plugin.
+     * Throws {@link UnsupportedOperationException} if no plugin is wired.
+     */
+    @Override
+    public DocumentLookupResult getById(Engine.Get get) throws IOException {
+        ensureOpen();
```
RS146BIJAY: Move ensure open in try block

thorkous: Done

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Done

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -686,6 +717,10 @@ private Engine.IndexResult indexIntoEngine(Engine.Index index, IndexingStrategy
             final Translog.Location location;
             if (indexResult.getResultType() == Engine.Result.Type.SUCCESS) {
                 location = translogManager.add(new Translog.Index(index, indexResult));
+                versionMap.maybePutIndexUnderLock(
+                    index.uid().bytes(),
+                    new IndexVersionValue(location, indexResult.getVersion(), index.seqNo(), index.primaryTerm())
```
RS146BIJAY, line 766: Why we are enforcing saving location in the version map: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/InternalEngine.java#L975

thorkous: Aren't we doing the same thing in internalEngine? 

What's the issue in saving the location?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -247,6 +268,8 @@ public DataFormatAwareEngine(EngineConfig engineConfig) {
         this.shardId = engineConfig.getShardId();
         this.store = engineConfig.getStore();
         this.throttle = new IndexingThrottler();
+        this.documentLookupProvider = documentLookupProvider;
+        this.versionMap = new LiveVersionMap();
```
RS146BIJAY, line 287: Can you elaborate on how we are rotating liveVersionMap here?

RS146BIJAY: Got it. It seems to be in beforeRefresh

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1322,7 +1363,11 @@ public long getMaxSeqNoOfUpdatesOrDeletes() {
 
     @Override
     public void advanceMaxSeqNoOfUpdatesOrDeletes(long maxSeqNoOfUpdatesOnPrimary) {
-        throw new UnsupportedOperationException("updates/deletes not supported");
+        if (maxSeqNoOfUpdatesOnPrimary == SequenceNumbers.UNASSIGNED_SEQ_NO) {
```
RS146BIJAY: I think we can skip this for now and add it as a part of deletes integration.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -2003,6 +2247,21 @@ private void markSeqNoAsSeen(long seqNo) {
         localCheckpointTracker.advanceMaxSeqNo(seqNo);
     }
 
+    private final Object refreshIfNeededMutex = new Object();
+
+    /**
+     * Refresh this engine **internally** iff the requesting seq_no is greater than the last refreshed checkpoint.
+     */
+    protected final void refreshIfNeeded(String source, long requestingSeqNo) {
```
RS146BIJAY: Why is this required?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareReadOnlyEngine.java
```diff
@@ -305,6 +315,17 @@ public Engine.NoOpResult noOp(Engine.NoOp noOp) throws IOException {
...
```
RS146BIJAY: You need to implement VersionConflict changes even here as well: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/Engine.java#L705

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: removed.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,637 @@
...
+ * DataFusion native runtime. The single returned Arrow record batch is flattened into a
+ * JSON source with reserved columns stripped.
+ */
+@ExperimentalApi
+public class GetService {
+
+    private static final Logger logger = LogManager.getLogger(GetService.class);
+
+    /** Reserved columns that must not leak into the JSON _source payload. */
+    private static final Set<String> RESERVED_FIELDS = Set.of(
```
expani: This is not a Datafusion Backend construct. 

It should be coming from the Core layer ( `analytics-engine` thru `analytics-api` )

expani (changes requested):
Thanks for getting this started @thorkous 

I think the overall approach is tightly coupling Lucene Backend with Datafusion.

We need a GetServiceProvider in Core ( `analytics-api` ) and default mock implementations ( `analytics-engine` ) 

I would say that we should instead model it around accessing doc values either in large Batches or randomly a few.

The `RESERVED_FIELDS` should be a planner construct so that we can perform related validations at the Marking phase of whether a GetServiceProvider/DocValueProvider is present in the backend while validating it's capabilities.

All our interactions at the Core layer ( `analytics-api/analytics-engine` ) should be using APIs based on Calcite's RexNode/RexCall.

.idea/runConfigurations/Debug_OpenSearch.xml
```diff
@@ -1,11 +1,15 @@
...
-    <configuration default="false" name="Debug OpenSearch" type="Remote">
-        <option name="USE_SOCKET_TRANSPORT" value="true" />
-        <option name="SERVER_MODE" value="true" />
-        <option name="SHMEM_ADDRESS" />
-        <option name="HOST" value="localhost" />
-        <option name="PORT" value="5005" />
-        <option name="AUTO_RESTART" value="true" />
-        <method v="2" />
-    </configuration>
+  <configuration default="false" name="Debug OpenSearch" type="Remote">
```
RS146BIJAY: Why are adding debug configuration here?

thorkous: Incorrectly added, removed it.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareReadOnlyEngine.java
thorkous: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: Will it fine if the resolution logic lives in `LuceneRowLocator` in the server module and gets injected via a `RowLocator` interface. DataFusion plugin will have zero Lucene imports and it just receives the row location and does the parquet read.

expani: @thorkous We Cannot have any `org.apache.lucene` imports present in Datafusion Backend plugin. 

That breaks the whole architecture of keeping the backends independent of each other.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
RS146BIJAY: No. I am not sure this is a valid scenario

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1864,6 +1980,129 @@ private void closeReaders() throws IOException {
...
+                versionValue = new IndexVersionValue(null, lookupResult.version(), lookupResult.seqNo(), lookupResult.primaryTerm());
+            }
+        } else {
+            if (engineConfig.isEnableGcDeletes()
+                && versionValue.isDelete()
+                && (engineConfig.getThreadPool().relativeTimeInMillis()
+                    - ((DeleteVersionValue) versionValue).time) > getGcDeletesInMillis()) {
+                versionValue = null;
+            }
+        }
```
RS146BIJAY, line 2211: Lets ensure this is integrated with CAS as well. Non of the existing feature should break if a new feature is introduced.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Can you explain then what should be scenario when documentLookupProvider is null?

RS146BIJAY: I think add an assertion here that documentLookupProvider cannot be null.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: I am not sure why CAS integration is required here when the index writer are not supporting CAS.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: I have made the changes which eliminates org.apache.lucene from Datafusion Backend plugin.

Bukhtawar (commented):
Can we not have seq no(we already do)/primary_term/version all three in Lucene, else update/delete paths would be super slow?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneRowLocator.java
```diff
@@ -0,0 +1,101 @@
...
+    private final DataFormat luceneFormat;
+
+    public LuceneRowLocator(DataFormat luceneFormat) {
+        this.luceneFormat = luceneFormat;
+    }
+
+    @Override
+    public DocumentRowLocation resolveRowLocation(IndexReaderProvider.Reader reader, String id) throws IOException {
+        DirectoryReader luceneReader = directoryReader(reader);
+        if (luceneReader == null || luceneReader.numDocs() == 0) {
```
expani: A null reader should cause an exception here OR we validate the case and throw IllegalArgException even before this is called.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneRowLocator.java
```diff
@@ -0,0 +1,101 @@
...
+            return null;
+        }
+        BytesRef idBytes = Uid.encodeId(id);
```
expani: Why do we need encoding as BytesRef ? Is TermQuery using a string as a parameter not working ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,343 @@
...
+            long runtimePtr = dfPlugin.getDataFusionService().getNativeRuntime().get();
+            // ReaderHandle registers the native pointer with NativeHandle so downstream
+            // validatePointer() calls in executeQueryAsync() find it in the live set.
+            MonoFileWriterSet segment = MonoFileWriterSet.of(parquetDir, writerGeneration, parquetFile, 0L);
+            try (ReaderHandle readerHandle = new ReaderHandle(parquetDir, List.of(segment), null)) {
+                long readerPtr = readerHandle.getPointer();
+                if (rowId < 0) {
+                    throw new IllegalArgumentException("rowId must be non-negative, got: " + rowId);
+                }
+                String sql = "SELECT * FROM \"" + GET_BY_ID_TABLE_ALIAS + "\" LIMIT 1 OFFSET " + Long.toUnsignedString(rowId);
```
expani: If the query just has rowId that changes then we can cache the Substrait after creation to avoid creating one for every call. 

If you wish to take this as a follow up, add a TODO

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneRowLocator.java
```diff
@@ -0,0 +1,101 @@
...
+            return null;
+        }
+        int docId = topDocs.scoreDocs[0].doc;
+        int leafOrd = ReaderUtil.subIndex(docId, luceneReader.leaves());
+        LeafReaderContext leafCtx = luceneReader.leaves().get(leafOrd);
+        int localDocId = docId - leafCtx.docBase;
+
+        SortedNumericDocValues rowIdDv = leafCtx.reader().getSortedNumericDocValues(DocumentInput.ROW_ID_FIELD);
+        if (rowIdDv == null || !rowIdDv.advanceExact(localDocId)) {
+            throw new IllegalStateException("Leaf segment missing " + DocumentInput.ROW_ID_FIELD + " doc values");
```
expani: This is error is true for `rowIdDV == null` not for `!rowIdDv.advanceExact(localDocId)` in the latter case it means that docId is missing which means either user is trying to look for a invalid DocId or some bug exists in upstream code

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,343 @@
...
+                case Binary:
+                case LargeBinary:
+                case FixedSizeBinary:
+                case BinaryView:
+                    return null;
+                default:
+                    break;
+            }
+            Object raw = vec.getObject(idx);
+            switch (id) {
```
expani: Check if there is an existing utility for this. If not move this as a utility for others to reuse in future.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,343 @@
...
+                        return ISO_FORMATTER.format(toInstant(((Number) raw).longValue(), ts.getUnit()));
+                    }
+                    return raw == null ? null : raw.toString();
+                default:
+                    // TODO type coverage (list, struct, decimal)
+                    return null;
+            }
+        }
+
+        private static Instant toInstant(long v, TimeUnit unit) {
```
expani: Similar comment about utility here

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DocumentRowReader.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * Fetch all rows with {@code _seq_no > fromSeqNoExclusive} across the snapshot's storage.
+     *
+     * @param reader a point-in-time reader snapshot (used to resolve backend files)
+     * @param indexName the logical table name
+     * @param fromSeqNoExclusive the exclusive lower bound on _seq_no
+     * @return list of rows as field-name → value maps
+     */
+    List<Map<String, Object>> executeRangeQuery(IndexReaderProvider.Reader reader, String indexName, long fromSeqNoExclusive)
```
expani: This wants to execute a range query but assumes that the filter clause will be ` > fromSeqNoExclusive` which doesn't seem correct for the API Contract.

expani: We can call it get all fields greater than a value or something similar.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,343 @@
...
+    }
+
+    /**
+     * Executes a {@code SELECT * FROM tableName LIMIT 1 OFFSET rowId} query over a single parquet
+     * file via the DataFusion native bridge and returns the first row as a map. Uses
+     * {@link NativeBridge#sqlToSubstrait} to let the native planner project the parquet file's
+     * true schema — passing a hand-rolled Substrait plan with an empty {@code base_schema} yields
+     * zero-column projection.
+     */
+    interface NativeExecutor extends DocumentRowReader {}
```
expani: Don't think this marker interface is useful ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        @Override
+        public void close() {
+            sharedAllocator.close();
+        }
+
+        WriterFileSet findParquetSet(CatalogSnapshot snapshot, long writerGeneration) {
+            for (Segment seg : snapshot.getSegments()) {
+                WriterFileSet candidate = seg.dfGroupedSearchableFiles().get(PARQUET_FORMAT);
```
expani: Given there is nothing specific to a backend here, can't this happen at a Core layer and we directly pass in the candidate segments ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,343 @@
...
+                String parquetDir = parquetSet.directory();
+                String parquetFile = parquetSet.files().iterator().next();
+                MonoFileWriterSet writerSet = MonoFileWriterSet.of(parquetDir, parquetSet.writerGeneration(), parquetFile, 0L);
+                try (ReaderHandle readerHandle = new ReaderHandle(parquetDir, List.of(writerSet), null)) {
+                    long readerPtr = readerHandle.getPointer();
+                    String sql = "SELECT \"_id\", \"_seq_no\", \"_primary_term\", \"_version\" FROM \""
+                        + GET_BY_ID_TABLE_ALIAS
+                        + "\" WHERE \"_seq_no\" > "
+                        + seqNoFloor;
+                    byte[] substraitPlan = NativeBridge.sqlToSubstrait(readerPtr, GET_BY_ID_TABLE_ALIAS, sql, runtimePtr);
```
expani: Similar comment on creating the substrait once and only changing the value of `seqNoFloor` when required.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,343 @@
...
+                StreamHandle streamHandle = new StreamHandle(streamPtr, dfPlugin.getDataFusionService().getNativeRuntime());
+                DatafusionResultStream stream = new DatafusionResultStream(streamHandle, sharedAllocator)
+            ) {
+                var iter = stream.iterator();
+                while (iter.hasNext()) {
+                    var batch = iter.next();
+                    try (VectorSchemaRoot root = batch.getArrowRoot()) {
+                        FieldVector idVec = root.getVector(IdFieldMapper.NAME);
+                        for (int i = 0; i < root.getRowCount(); i++) {
+                            Map<String, Object> row = rowToMap(root, i);
```
expani: readSingleRow, readAllRows and rowToMap also look like common utilities that should be present in Core as they only deal with Arrow and StreamHandle both of which are available.

expani (changes requested):

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneRowLocator.java
thorkous: you can't drop Uid.encodeId() and use a plain string — the indexed bytes carry a type-prefix that a raw string TermQuery won't match.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: readSingleRow and rowToMap operate on Arrow VectorSchemaRoot/FieldVector and the native StreamHandle, which are DataFusion's in-memory representation rather than generic types. The Core abstraction already sits one level up: DocumentRowReader.executeSingleRow(...) returns a plain Map<String, Object>, and Core (DocumentLookupService.buildResultFromRow) only ever sees that map. Pulling these into Core would bring Arrow + StreamHandle into the core layer and tie it to the Parquet/DataFusion format — a future backend on a different format would have to produce Arrow, or we'd end up reimplementing the conversion per format anyway. Keeping the Arrow→Map step inside the DataFusion backend is what keeps Core format-agnostic, so I'd prefer to leave them here.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneRowLocator.java
thorkous: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: Done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneRowLocator.java
thorkous: Done

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DocumentRowReader.java
thorkous: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: Done.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DocumentLookupService.java
```diff
@@ -0,0 +1,150 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Core orchestrator for document lookup. Coordinates row location resolution,
+ * backend-specific execution, and result building. Lives in the analytics core
```
expani: This currently lives in the `analytics-framework` but as the comment said should be in core.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
expani: I think we should look at using `SearchExecEngine` which is already the Abstraction used by Core to interact with Backends instead of tightly coupling the generic GetService to Datafusion Backend. 

We can look at extending it with a DocumentRowReader API.

It is already the hook point for getting CatalogSnapshot from the IndexShard and Engine abstractions

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ReservedFields.java
```diff
@@ -0,0 +1,36 @@
...
+        "_id",
+        "_seq_no",
+        "_primary_term",
+        "_version",
+        "_doc_count",
+        "_size",
+        "_routing",
+        "_ignored",
+        "__row_id__"
+    );
```
mgodwan: Can we use PutMappingRequest#RESERVED_FIELDS ?

sandbox/libs/analytics-api/build.gradle
```diff
@@ -7,10 +7,11 @@
  */
 
 /*
- * Analytics Engine API surface consumable from JDK 21 code paths.
+ * Analytics Engine API surface.
+ * Targets JDK 25 (required for Panama FFM APIs used by the native DataFusion bridge).
  */
 
-java { sourceCompatibility = JavaVersion.toVersion(21); targetCompatibility = JavaVersion.toVersion(21) }
+java { sourceCompatibility = JavaVersion.toVersion(25); targetCompatibility = JavaVersion.toVersion(25) }
```
mgodwan: Is this change needed as there is no other change in this package?

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshot.java
```diff
@@ -243,6 +243,24 @@ public Map<String, Collection<String>> getFilesByFormat() {
...
+     */
+    public WriterFileSet findFileSet(String format, long writerGeneration) {
+        for (Segment seg : getSegments()) {
+            WriterFileSet candidate = seg.dfGroupedSearchableFiles().get(format);
+            if (candidate == null || candidate.files().isEmpty()) continue;
+            if (candidate.writerGeneration() != writerGeneration) continue;
+            return candidate;
+        }
+        return null;
+    }
```
mgodwan, line 262: Please add a UT for this

server/src/main/java/org/opensearch/indices/IndicesService.java
```diff
@@ -1254,12 +1256,41 @@ private IngestionConsumerFactory getIngestionConsumerFactory(final IndexSettings
...
     private IndexerFactory getIndexerFactory(final IndexSettings idxSettings) {
         if (idxSettings.isPluggableDataFormatEnabled()) {
-            return new DataFormatAwareIndexerFactory();
+            DataFormatAwareIndexerFactory factory = new DataFormatAwareIndexerFactory();
+            DocumentMetadataResolver documentResolver = resolveDocumentResolver();
+            DocumentLookupProvider provider = resolveGetByIdPlugin();
+            if (provider != null) {
+                provider.setDocumentResolver(documentResolver);
+            }
+            factory.setGetByIdPlugin(provider);
```
mgodwan: Setter doesn't seem to be the correct pattern here. Lets move DocumentLookupProvider to EngineConfig

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDocumentResolver.java
```diff
@@ -0,0 +1,125 @@
...
+ * <p>An empty reader or an id that is not present reports "not found".
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public final class LuceneDocumentResolver implements DocumentMetadataResolver {
+
+    private final DataFormat luceneFormat;
+
+    public LuceneDocumentResolver(DataFormat luceneFormat) {
```
mgodwan: This should be hard wired to LuceneDataFormat

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDocumentResolver.java
```diff
@@ -0,0 +1,125 @@
...
+            throw new IllegalStateException("Cannot resolve document row location: no Lucene reader is available");
+        }
+        if (readerObj instanceof DirectoryReader dr) {
+            return dr;
+        }
+        if (readerObj instanceof LuceneReader lr) {
+            return lr.directoryReader();
+        }
+        throw new IllegalStateException("Reader for format [lucene] has unexpected type: " + readerObj.getClass().getName());
+    }
```
mgodwan, line 86:
```
LuceneReader luceneReader = reader.getReader(plugin.getDataFormat(), LuceneReader.class);
return luceneReader.directoryReader();
```

server/src/main/java/org/opensearch/plugins/DocumentLookupProvider.java
```diff
@@ -0,0 +1,76 @@
...
+     * @throws IOException if resolution fails
+     */
+    DocumentLookupResult getById(Engine.Get get, IndexReaderProvider.Reader reader, String indexName) throws IOException;
+
+    /**
+     * Resolves only version metadata ({@code _version}/{@code _seq_no}/{@code _primary_term}) for an id,
+     * skipping {@code _source} reconstruction. Used by the engine's version-resolution hot path.
+     * Defaults to not-found; backends override.
+     */
+    default DocumentLookupResult getVersionMetadata(String id, IndexReaderProvider.Reader reader, String indexName) throws IOException {
```
mgodwan:
```
getVersionMetadata(String id, IndexReaderProvider.Reader reader, Index index)
```

mgodwan: Applicable across all methods in the interface.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -1532,7 +1533,19 @@ public Engine.GetResult get(Engine.Get get) {
...
+        }
+        try {
+            DocumentLookupResult lookup = indexer.getById(get);
+            if (lookup == null || lookup.exists() == false) {
+                return GetResult.NOT_EXISTS;
+            }
+            return lookup.toGetResult();
+        } catch (IOException e) {
+            throw new OpenSearchException("get-by-id via pluggable path failed for id [" + get.id() + "]", e);
+        }
```
mgodwan: Can we unify the logic by calling acquireSearcher method within EngineBackedIndexer while wrapping `engine.get` output in DocumentLookupResult? That should prevent the cast/check in shard

server/src/main/java/org/opensearch/index/engine/exec/Indexer.java
```diff
@@ -198,4 +201,14 @@ Translog.Snapshot newChangesSnapshot(String source, long fromSeqNo, long toSeqNo
...
      */
     GatedCloseable<CatalogSnapshot> acquireLastCommittedSnapshot(boolean flushFirst) throws EngineException, IOException;
+
+    /**
+     * Resolves a document by id via a pluggable lookup path (see
+     * {@link DocumentLookupProvider}). Not supported by default;
+     * implementations that have a non-Lucene row store override this.
+     */
+    default DocumentLookupResult getById(Engine.Get get) throws IOException {
+        throw new UnsupportedOperationException("getById not supported for " + getClass().getSimpleName());
```
mgodwan: This needs to be implemented in DataFormatAwareNRTReplicationEngine as well. Can we put the common code at a single place which all three DataFormatAware*Engine can use?

mgodwan: Also, lets avoid a default impl here. We can have EngineBackedIndexer implement this method

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -466,6 +477,18 @@ DataFormatRegistry getDataFormatRegistry() {
         return dataFormatRegistry;
     }
 
+    @Override
+    public void setDocumentResolver(DocumentMetadataResolver documentResolver) {
```
mgodwan: Lets see how to avoid the setter pattern here. Will check on plugin wiring once for suggestion, but want to avoid any races via this

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: Done

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DocumentLookupService.java
thorkous: Corrected the doc.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ReservedFields.java
thorkous: PutMappingRequest has different reserve fields and these are different so we can't use them.

sandbox/libs/analytics-api/build.gradle
thorkous: Removed.

server/src/main/java/org/opensearch/index/engine/exec/coord/CatalogSnapshot.java
thorkous: Tests are added in DataformatAwareCatalogSnapshotTests

server/src/main/java/org/opensearch/indices/IndicesService.java
thorkous: Done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDocumentResolver.java
thorkous: Done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDocumentResolver.java
thorkous: done

server/src/main/java/org/opensearch/plugins/DocumentLookupProvider.java
thorkous: Done

server/src/main/java/org/opensearch/index/shard/IndexShard.java
thorkous: Done

server/src/main/java/org/opensearch/index/engine/exec/Indexer.java
thorkous: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
thorkous: Done

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ReservedFields.java
mgodwan: This is quite error prone. We should be relying on mappers rather than a list like this

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDocumentResolver.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private DirectoryReader directoryReader(IndexReaderProvider.Reader reader) {
+        LuceneReader luceneReader = reader.getReader(LucenePlugin.DATA_FORMAT, LuceneReader.class);
+        return luceneReader.directoryReader();
+    }
+
+    private LeafDoc locate(DirectoryReader luceneReader, String id) throws IOException {
+        BytesRef idBytes = Uid.encodeId(id);
+        TopDocs topDocs = new IndexSearcher(luceneReader).search(new TermQuery(new Term(IdFieldMapper.NAME, idBytes)), 1);
```
mgodwan: Can we ensure that this lookup is in reverse order of segment creation?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,245 @@
...
+                        }
+
+                        @Override
+                        public void onFailure(Exception e) {
+                            future.completeExceptionally(e);
+                        }
+                    }
+                );
+                try {
+                    return future.join();
```
mgodwan: This should've a timeout

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -890,6 +935,7 @@ public void refresh(String source) throws EngineException {
             refreshLock.lock();
             try (GatedCloseable<CatalogSnapshot> catalogSnapshot = catalogSnapshotManager.acquireSnapshot()) {
                 if (store.tryIncRef()) {
+                    versionMap.beforeRefresh();
```
mgodwan: Can we not have this wired as a RefreshListener?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1034,10 +1080,15 @@ public void refresh(String source) throws EngineException {
                     }
                     if (refreshed) {
                         lastRefreshedCheckpointListener.updateRefreshedCheckpoint(localCheckpointBeforeRefresh);
+                        versionMap.pruneTombstones(
```
mgodwan: Is this the only place to prune? We need to do it on onSettingsChanged/flush as well?
Also, override from `maybePruneDeletes`?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,245 @@
...
+            }
+            String parquetDir = parquetSet.directory();
+            String parquetFile = parquetSet.files().iterator().next();
+            long runtimePtr = dfPlugin.getDataFusionService().getNativeRuntime().get();
+            // ReaderHandle registers the native pointer with NativeHandle so downstream
+            // validatePointer() calls in executeQueryAsync() find it in the live set.
+            MonoFileWriterSet segment = MonoFileWriterSet.of(parquetDir, parquetSet.writerGeneration(), parquetFile, 0L);
+            try (ReaderHandle readerHandle = new ReaderHandle(parquetDir, List.of(segment), null, List.of(), List.of())) {
+                long readerPtr = readerHandle.getPointer();
+                // TODO: only the OFFSET (rowId) varies between calls — cache the Substrait plan
```
bharath-techie: select * from table where rowId = 'some_row_id' 

why are we not using this query for example ? this will prune the pages based on the filter and decode only the page that is required and get the document.

I'm not sure about the offset query - did we measure performance pitting the two queries against each other ?

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/DataFormatAwareReadonlyEngineBaseIT.java
```diff
@@ -136,7 +136,7 @@ protected Collection<Class<? extends Plugin>> nodePlugins() {
                 .filter(p -> p != MockFsRepositoryPlugin.class),
             Stream.<Class<? extends Plugin>>of(
                 ArrowBasePlugin.class,
-                ParquetOnlyDataFormatPlugin.class,
+                ParquetDataFormatPlugin.class,
```
mgodwan, line 139: Please add end to end tests for DFAE flows with active indexing and refreshes while fetching from version map and rows directly

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
```diff
@@ -0,0 +1,245 @@
...
+                try (ReaderHandle readerHandle = new ReaderHandle(parquetDir, List.of(writerSet), null, List.of(), List.of())) {
+                    long readerPtr = readerHandle.getPointer();
+                    // TODO: only the _seq_no floor varies between calls -- cache the Substrait plan
+                    // per reader and rebind the bound to avoid re-planning on every restore scan.
+                    String sql = "SELECT \"_id\", \"_seq_no\", \"_primary_term\", \"_version\" FROM \""
+                        + GET_BY_ID_TABLE_ALIAS
+                        + "\" WHERE \"_seq_no\" > "
+                        + seqNoFloor;
+                    byte[] substraitPlan = NativeBridge.sqlToSubstrait(readerPtr, GET_BY_ID_TABLE_ALIAS, sql, runtimePtr);
+                    WireConfigSnapshot configSnapshot = WireConfigSnapshot.builder(dfPlugin.getDatafusionSettings().getSnapshot())
```
bharath-techie: maybe for this particular query we need to execute in single partition as well

server/src/main/java/org/opensearch/indices/IndicesService.java
```diff
@@ -1260,6 +1262,28 @@ private IndexerFactory getIndexerFactory(final IndexSettings idxSettings) {
         }
     }
 
+    private DocumentLookupProvider resolveGetByIdPlugin() {
+        List<DocumentLookupProvider> plugins = pluginsService.filterPlugins(DocumentLookupProvider.class);
```
mgodwan: Can we get this via DataFormatRegistry? Lets not expose the plugin here?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ReservedFields.java
thorkous: I have removed RESERVED and used `IndicesModule.getBuiltInMetadataFields()`

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDocumentResolver.java
thorkous: Done

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GetService.java
thorkous: Added.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Done

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Done. Added maybePruneDeletes to onSettingsChanged/flush

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/DataFormatAwareReadonlyEngineBaseIT.java
thorkous: Added ITs for DFAE, NRT and ReadOnlyEngine.

server/src/main/java/org/opensearch/indices/IndicesService.java
thorkous: Done.

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/DataFormatAwareGetByIdIT.java
```diff
@@ -0,0 +1,124 @@
...
+    public void testReindexSameIdAcrossRefresh() {
+        createManualRefreshIndex();
+        assertEquals(RestStatus.CREATED, indexDoc("1", "doc_1", 1).status());
+        refreshIndex(INDEX);
+
+        // Re-index (overwrite) the same id with a new source; version-resolution bumps the version to 2.
+        // This is a plain index op, not the _update API. Do NOT refresh yet.
+        IndexResponse update = indexDoc("1", "doc_1_updated", 100);
+        assertEquals(2L, update.getVersion());
+
```
mgodwan: This should not be allowed without updates/deletes in place

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDocumentResolver.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+        return dv.nextValue();
+    }
+
+    private DirectoryReader directoryReader(IndexReaderProvider.Reader reader) {
+        LuceneReader luceneReader = reader.getReader(LucenePlugin.DATA_FORMAT, LuceneReader.class);
+        return luceneReader.directoryReader();
+    }
+
+    private LeafDoc locate(DirectoryReader luceneReader, String id) throws IOException {
```
mgodwan, line 88: How useful would VersionsAndSeqNoResolver be here?

mgodwan: loadDocIdAndVersion method does similar work only (the doc values part is not reusable there, but getting doc id is) Its ok to keep it separate, but wondering the optimizations present there around perthread lookup cache may be lost

mgodwan (approved):

PR #21804, opened 2026-05-22
[Analytics-Engine] Lower LITERAL_AGG and accept all join condition shapes

songkant-aws:
### Description

Two complementary marking-phase fixes that build on top of #21795. Both keep Calcite as the source of truth and unblock end-to-end correlated-subquery execution on the analytics-engine route.

1. **LITERAL_AGG lowering** — after `RelDecorrelator` runs in `PlannerImpl.removeSubQueries`, a new Hep pass invokes `ExtractLiteralAggRule` (ported from Apache Impala with attribution) to rewrite `LITERAL_AGG(literal)` aggregate calls into an `Aggregate` + `Project` shape. `SubQueryRemoveRule.rewriteIn` emits `LITERAL_AGG` for `NOT IN`'s null-aware semantics; Calcite ships no rule to remove it, and each downstream engine handles it differently (Druid registers a native impl, Ignite wires an accumulator, Impala lowers, Flink forks `SubQueryRemoveRule`). The Impala approach is a clean Calcite-rule-style rewrite — port it.

2. **`OpenSearchJoinRule.matches()` relaxed** — accept all join condition shapes, not just equi or mixed equi+non-equi. `RelDecorrelator`'s output for correlated subqueries with non-equi correlation predicates (e.g. `WHERE id > uid`) is a pure non-equi `LogicalJoin`. DataFusion handles non-equi via NestedLoopJoin downstream, so there is no capability-level reason to reject these in marking.

`PlannerImpl` is also refactored to a fluent `HepPhase` builder so each Hep pass attaches the existing `RuleProfilingListener` consistently and the phase boundary is recorded in the planner profile.

### Dependency on #21795

This PR is **draft** until #21795 merges. The branch contains all of #21795's commits plus the three commits in this layer; once #21795 lands, this PR will rebase onto `main` and only show the layered diff.

### Test results (on this branch = #21795 + these fixes)

| Suite | Result |
| --- | --- |
| `ExistsSubqueryCommandIT` | 11/11 |
| `ScalarSubqueryCommandIT` | 13/13 |
| `InSubqueryCommandIT` | 11/13 |
| `StreamstatsCommandIT.testStreamstatsReset` | pass |
| `JoinCommandIT` | 12/12 (no regression from join relaxation) |
| **Total** | **36/38 ≈ 95%** |

The two remaining `InSubqueryCommandIT` failures are unrelated to subquery decorrelation:
- `testTwoExpressionsInSubquery` — row-ordering assertion vs. hash-join bucketing. Test issue.
- `testTwoExpressionsNotInSubquery` — `OpenSearchAggregateSplitRule` rejects nullable filter expressions ("filter must be BOOLEAN NOT NULL") on a multi-column `NOT IN`'s `COUNT FILTER`. Independent split-rule bug.

### Related Issues

Builds on #21795.

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -89,6 +84,8 @@ public static RelNode runAllOptimizations(RelNode rawRelNode, PlannerContext con
 
         RelNode modifiedRelNode = rawRelNode;
         modifiedRelNode = removeSubQueries(modifiedRelNode, listener);
+        LOGGER.info("After removeSubQueries:\n{}", RelOptUtil.toString(modifiedRelNode));
```
songkant-aws: Remove this line of log

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -126,59 +123,67 @@ public static RelNode runAllOptimizations(RelNode rawRelNode, PlannerContext con
...
+     * rewrites are the only Calcite source of {@code LITERAL_AGG} that we know
+     * of, but the rule is kept as its own pass — independent of the subquery
+     * removal phase — for two reasons: (1) phase listing reflects each
+     * concern separately so profiling output stays meaningful, (2) the rule
+     * still fires correctly if a future frontend / pushdown rule constructs
+     * {@code LITERAL_AGG} directly via {@code RelBuilder.literalAgg(...)}.
+     */
+    private static RelNode extractLiteralAgg(RelNode input, RuleProfilingListener listener) {
+        return HepPhase.named("literal-agg-extract")
+            .addRules(new org.opensearch.analytics.planner.rules.ExtractLiteralAggRule())
```
songkant-aws: Use import instead of the full class path

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/ExtractLiteralAggRule.java
```diff
@@ -0,0 +1,173 @@
...
+                    "LITERAL_AGG must carry exactly one RexLiteral preOperand"
+                );
+                literalAtSlot.add(ac.rexList.get(0));
+            } else {
+                literalAtSlot.add(null);
+                keptCalls.add(ac);
+            }
+        }
+        // Nothing to do if no LITERAL_AGG present.
+        if (literalAtSlot.stream().noneMatch(java.util.Objects::nonNull)) {
```
songkant-aws: Use import instead of full function path

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -187,26 +189,18 @@ private static RelNode reduceExpressions(RelNode input, RuleProfilingListener li
...
     private static RelNode pushdownRules(RelNode input, RuleProfilingListener listener) {
-        HepProgramBuilder builder = new HepProgramBuilder();
-        builder.addMatchOrder(HepMatchOrder.BOTTOM_UP);
-        // Push Filters below Project/Aggregate/Join.
-        builder.addRuleCollection(
-            List.of(CoreRules.FILTER_PROJECT_TRANSPOSE, CoreRules.FILTER_AGGREGATE_TRANSPOSE, CoreRules.FILTER_INTO_JOIN)
-        );
-        // Merge adjacent Filters into one — must run after transposes so any
-        // auto-injected NOT NULL collapses with the user's WHERE.
-        builder.addRuleInstance(CoreRules.FILTER_MERGE);
```
expani, line 222: We would need the rule instance and rule collection in the refactored class.

Currently, it's based on size but this was added for proper ordering

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/HepPhase.java
```diff
@@ -0,0 +1,132 @@
...
+    /** Build the Hep planner, run it on {@code input}, and apply the post-processing
+     *  hook (if any). The listener (when non-null) sees a matched begin/end phase pair
+     *  around both stages. */
+    public RelNode run(RelNode input, RuleProfilingListener listener) {
+        HepProgramBuilder builder = new HepProgramBuilder();
+        if (bottomUp) {
+            builder.addMatchOrder(HepMatchOrder.BOTTOM_UP);
+        }
+        if (rules.size() == 1) {
+            builder.addRuleInstance(rules.get(0));
```
expani: This has to be deterministic based on the Hep Phase

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
songkant-aws: Yeah, I just confirmed that Calcite run HepInstruction(either collection or single instance here) until optimization fixed point is reached. So they have different semantics and boundary. Changed to support addRuleCollection and addRuleInstance differently.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/HepPhase.java
songkant-aws: Ditto

mch2 (approved):

PR #21807, opened 2026-05-22
Added UDT for IP and Binary support

vinaykpud:
### Description

Adds Calcite UDTs for OpenSearch `ip` and `binary` columns to the analytics-engine path, plus two backend adapters and two Rust UDFs that make `CIDRMATCH` and `cast(<ip|binary> as STRING)` work end-to-end against DataFusion. Closes the rendering, schema-label, `CIDRMATCH`-literal and `[B`-writer-crash bugs.

#### Background

Today, `OpenSearchSchemaBuilder.addLeafFields` collapses both OpenSearch `ip` and `binary` field types to `SqlTypeName.VARBINARY`. Three downstream consequences:

1. The Calcite plan's row type carries plain `VARBINARY` for the column. The DataFusion backend ships back a 16-byte ipv4-mapped-ipv6 buffer and `byte[]` reaches the SQL plugin's row converter, which has no `byte[]` case — `unsupported object class [B`.
2. The response schema reports `"type": "binary"` for an IP column.
3. `CIDRMATCH('1.2.3.4', '1.2.3.0/24')` (literal/literal) reaches DataFusion as a UDF call no backend implements.
4. PPL `cast(host as STRING)` lands as `SAFE_CAST(<ip-col> AS VARCHAR)`. DataFusion's `cast(binary, utf8)` kernel decodes the raw 16-byte buffer as UTF-8 — garbage strings or NULL.

#### Approach

Three coordinated pieces:

**1. UDT split (`analytics-api`).** New `IpType` and `BinaryType` extending `AbstractSqlType` with `VARBINARY` underneath. They keep `getSqlTypeName() == VARBINARY` so Substrait, Arrow, DataFusion, and operator dispatch (cidrmatch byte-range rewrite, `=`/`IN`/`BETWEEN` coercion) all keep working unchanged. The UDT distinction lives only in the Calcite RelNode and is read at exactly two sites in the SQL plugin (response-schema build + row converter). Wired through `OpenSearchSchemaBuilder.buildLeafType`.

**2. CIDRMATCH backend adapter (`analytics-backend-datafusion`).** New `CidrMatchFunctionAdapter` implementing `ScalarFunctionAdapter`.

**3. Cast rewrite + Rust UDFs (`analytics-backend-datafusion`).** New `IpBinaryCastFunctionAdapter` (also `ScalarFunctionAdapter`) registered for both `ScalarFunction.CAST` and `ScalarFunction.SAFE_CAST` (PPL's `cast(... AS STRING)` reliably emits `SAFE_CAST` via `rexBuilder.makeCast(..., true, true)`, but Calcite's planner-internal coercion can also emit plain `CAST`). For VARCHAR-target casts whose source is `IpType` the adapter emits `ip_to_string(<col>)`; for `BinaryType`, `binary_to_base64(<col>)`. Otherwise returns the original unchanged.

The two emitted UDFs bind through the Substrait extension catalog (`opensearch_scalar_functions.yaml`) and are implemented in two new Rust UDFs:
- `ip_to_string` — detects ipv4-mapped form (10 zero bytes + `0xff 0xff` + 4 IPv4 bytes), emits dotted-quad; otherwise renders the 16-byte buffer via `Ipv6Addr::to_string()` (RFC 5952). Output matches `InetAddresses.toAddrString` semantics.
- `binary_to_base64` — base64-encodes per the OpenSearch `binary` field wire contract.

The cast rewrite has to happen **before** Substrait conversion because by the time the row reaches Java the cell is already a (mangled) `String`, not `byte[]`.

### Companion change in the SQL plugin

The SQL plugin recognizes the new UDT at the response-schema boundary (instanceof dispatch in `OpenSearchTypeFactory`, byte[] formatting in `AnalyticsExecutionEngine.convertRows`) and removes the legacy in-plugin `CidrMatchAdapter` now that dispatch lives on the backend. Companion PR: [opensearch-project/sql#5463](https://github.com/opensearch-project/sql/pull/5463) — must merge together.

### Testing

**Unit tests added:**
- `OpenSearchSchemaBuilderTests` — projected ip/binary columns are `IpType`/`BinaryType` but `getSqlTypeName() == VARBINARY`; digest-based equality.
- `CidrMatchFunctionAdapterTests` — both-literal in-range/out-of-range, IPv6 literal fold, varbinary column + cidr literal byte-range AND, dynamic cidr falls through, unparseable IP falls through.
- `IpBinaryCastFunctionAdapterTests` — IpType cast, BinaryType cast, SAFE_CAST IpType, non-VARCHAR target unchanged, plain VARBINARY (no UDT) unchanged, INTEGER unchanged.
- Rust per-module tests — `ip_to_string`: ipv4-mapped, pure IPv6 loopback, arbitrary IPv6, null pass-through, wrong-length null. `binary_to_base64`: round-trip a known payload, null pass-through.

**End-to-end:** ran the IP/binary PPL test set against a single-node `gradle run` cluster on x86_64 + Corretto JDK 25:
- `cidrmatch(host, '1.2.3.0/24')` (column form) — passes, fires tier 2.
- `CIDRMATCH('1.2.3.4', '1.2.3.0/24')` (literal/literal) — passes, fires tier 1.
- `cast(host as STRING)` — returns `"1.2.3.4"` / `"::1"` (was a 16-byte garbage buffer).
- `cast(blob as STRING)` — base64-encoded buffer matches `| fields blob`.
- `where host = '1.2.3.4'` — coerces cleanly with no `Unable to convert call IP(string)` regression.
- `FieldTypeCoverageIT#testIpFilters` — unchanged, still passing.


sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/BinaryType.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Calcite type marker for an OpenSearch {@code binary} column. Underlying SQL type is
+ * {@link SqlTypeName#VARBINARY}, matching the on-disk encoding;
+ * {@link AbstractSqlType} handles {@code getSqlTypeName()}, {@code isNullable()}, and
+ * {@code getFamily()} for free. The dedicated subclass exists so the SQL plugin's
+ * response-schema build can {@code instanceof}-dispatch to render the column type as
+ * {@code "binary"}, and so {@code AnalyticsExecutionEngine.convertRows} can base64-encode
+ * the {@code byte[]} cell to match the OpenSearch {@code binary} field wire contract.
+ */
+public final class BinaryType extends AbstractSqlType {
```
penghuo, line 22: Calcite support SqlTypeName.BINARY/SqlTypeName.VARBINARY. No need to create BinaryType?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/IpBinaryOutputCastRewriter.java
```diff
@@ -0,0 +1,178 @@
...
+            // Match both CAST and SAFE_CAST. PPL's `cast(... as STRING)` lowers via
+            // `rexBuilder.makeCast(targetType, expr, true /* matchNullability */, true /* safe */)`
+            // — the `safe` flag produces a SAFE_CAST RexKind, not a CAST.
+            SqlKind kind = visited.getKind();
+            if (kind != SqlKind.CAST && kind != SqlKind.SAFE_CAST) {
+                return visited;
+            }
+            if (visited.getType().getSqlTypeName() != SqlTypeName.VARCHAR) {
+                return visited;
+            }
```
dai-chen: I think we're removing the rule that adds Project(CAST date/time/ts as string) due to datetime type name issue. @mengweieric If so, we need to check each project item here.

vinaykpud: I changed the `IpBinaryOutputCastRewriter.java` to `IpBinaryCastFunctionAdapter` small change in the implementation which updates the the RelNode before substrait conversion step.

For your comment,
I had to add this to support the queries like
```
source=my_index | eval s = cast(host as STRING) | fields s
```
I dint understand your comment fully, can you take look into my current implementation in `IpBinaryCastFunctionAdapter`

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/IpBinaryCastFunctionAdapter.java
```diff
@@ -0,0 +1,113 @@
...
+        "binary_to_base64",
+        SqlKind.OTHER_FUNCTION,
+        ReturnTypes.VARCHAR_2000_NULLABLE,
+        null,
+        OperandTypes.ANY,
+        SqlFunctionCategory.SYSTEM
+    );
+
+    @Override
+    public RexNode adapt(RexCall original, List<FieldStorageInfo> fieldStorage, RelOptCluster cluster) {
```
vinaykpud, line 64:
```
  ┌──────────────────────┬────────────────────────────────────────┬────────────────────────────┬────────────────────┐
  │         PPL          │            Calcite RexCall             │       After adapter        │ Substrait function │
  ├──────────────────────┼────────────────────────────────────────┼────────────────────────────┼────────────────────┤
  │ cast(host as STRING) │ SAFE_CAST(<IpType col> AS VARCHAR)     │ IP_TO_STRING_OP(<col>)     │ ip_to_string       │
  ├──────────────────────┼────────────────────────────────────────┼────────────────────────────┼────────────────────┤
  │ cast(blob as STRING) │ SAFE_CAST(<BinaryType col> AS VARCHAR) │ BINARY_TO_BASE64_OP(<col>) │ binary_to_base64   │
  └──────────────────────┴────────────────────────────────────────┴────────────────────────────┴────────────────────┘
```

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/IpBinaryOutputCastRewriter.java
dai-chen: FYI https://github.com/opensearch-project/sql/pull/5454

vinaykpud: But this is for IP/Binary not for Date/Timestamp, am i missing anything?

sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/BinaryType.java
vinaykpud: Yes, but at storage layer there is not distinction. our setup we collapse everything to VARBINARY at the storage layer. 
But having this Logical Types defined gives us control when we have any future types like GeoIP ect

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/IpBinaryOutputCastRewriter.java
mengweieric: @dai-chen good point. did a quick scan, seems like ` DatetimeOutputCastRule` only wraps `DATE/TIME/TIMESTAMP` fields (isDatetimeType per UdtMapping), it doesn't touch VARBINARY-backed IP/Binary at all, so `the CAST(ip/binary AS VARCHAR)` this adapter rewrites was never coming from that rule. User-issued cast(host as STRING) lowers through CalciteRexNodeVisitor.visitCast → rexBuilder.makeCast at AST→Rel time, which is separate from #5454/#21748. So checking each project item: datetime ones no-op via the instanceof guard, and IP/Binary ones should hit the rewrite branch the same way they do today.

mch2 (approved):

PR #21815, opened 2026-05-23
fix(search): make bitmap scorer repeatable

### LINKED ISSUE #21781: The combination of bitmap terms lookup + additional filter + text match produces wrong results, and force merge fixes it.
### Describe the bug

Bitmap terms lookup (`value_type: "bitmap"`) produces incorrect results (0 hits) when the target index has multiple segments and the query combines the bitmap filter with additional filter clauses and a text `match` query. After force-merging the target index to a single segment, the same query returns correct results.

### Related component

Search:Performance

### To Reproduce

1. Create two indices: `index_main` (with multiple segments) and `index_user` (containing bitmap data).
2. `index_user` has a `binary` field with `"store": true` containing a roaring bitmap.
3. `index_main` has an integer field that the bitmap is matched against, a boolean field, and a text field.
4. Run the following query:

```json
{
  "query": {
    "bool": {
      "minimum_should_match": 1,
      "should": [
        { "match": { "text_field": "some_term" } }
      ],
      "filter": [
        { "term": { "boolean_field": true } },
        {
          "terms": {
            "integer_field": {
              "index": "index_user",
              "id": "some_doc_id",
              "path": "bitmap_field",
              "store": true
            },
            "value_type": "bitmap"
          }
        }
      ]
    }
  }
}
```

### Expected behavior

The query should return documents that match all three conditions (text match + boolean filter + bitmap filter). Each pair of conditions works correctly in isolation:

| Query Combination | Result |
|-------------------|--------|
| Bitmap filter only | Correct hits |
| Bitmap filter + text match (no boolean filter) | Correct hits |
| Bitmap filter + boolean filter (no text match) | Correct hits |
| Bitmap filter + boolean filter + text match | **0 hits (incorrect)** |

### Actual Behavior

When all three clauses are present and `index_main` has multiple segments, the query returns 0 hits. An aggregation on the same data confirms the intersection is non-empty.

### Workaround

Force-merging `index_main` to a single segment (`_forcemerge?max_num_segments=1`) resolves the issue. After the merge, the identical query returns the expected results.

### Additional Observations

- The issue does not depend on query clause ordering (`must` vs `should` vs `filter` placement).
- `post_filter` with the text match works correctly even with multiple segments.
- `wildcard` and `prefix` queries work correctly in place of `match`, suggesting the issue is specific to analyzed text queries interacting with bitmap evaluation across segments.
- The problem does not occur when using explicit `terms` arrays (e.g., `"terms": {"integer_field": [123, 456]}`) instead of bitmap lookup.

### Environment

- Single-node cluster
- Index with 4 segments (approximately 20K documents, ~15 MB)
- Bitmap stored as base64-encoded roaring bitmap in a `binary` field with `store: true`

gingeekrishna:
### Description

`BitmapIndexQuery` and `Bitmap64IndexQuery` had their `DocIdSetBuilder` and `MergePointVisitor` declared as fields of the `ScorerSupplier` anonymous class, causing all calls to `ScorerSupplier.get()` to share mutable state. This violates the Lucene contract that each `get()` call must produce an independent scorer, and caused incorrect empty results in complex bool queries using bitmap terms lookup on multi-segment indices.

**Root causes fixed:**

- **Shared mutable state in `get()`** — `DocIdSetBuilder` and `MergePointVisitor` were constructed once and reused across multiple `get()` calls. Moved both into the body of `get()` so each invocation builds a fresh, independent `DocIdSet`.
- **`BytesRef` aliasing in `bitmapEncodedIterator.next()`** — the returned `BytesRef` shared the same backing array as the internal buffer, so subsequent iteration silently overwrote previously returned values. Fixed by returning `BytesRef.deepCopyOf(encoded)`.
- **Stale `hasBuffered` flag in `Bitmap64IndexQuery.bitmapEncodedIterator.advance()`** — when the iterator was exhausted without finding a value `>= target`, `hasBuffered` was left in whatever state it had before, allowing a stale buffered value to surface on the next `next()` call. Fixed by explicitly setting `hasBuffered = false` when the loop exits without a match.

### Testing

Added regression tests to both `BitmapIndexQueryTests` and `Bitmap64IndexQueryTests`:

- `testScorerSupplierGetIsRepeatable` — obtains two independent `ScorerSupplier` instances per leaf, calls `get()` on each, and asserts both return identical match sets.
- `testScoreNoMatchingDocs` — verifies correct empty-result handling when the bitmap contains no values present in the index.
- `testRewrite` — verifies that an empty bitmap rewrites to `MatchNoDocsQuery`.

### Related Issues

Resolves #21781

### Check List

- [x] API changes companion pull request created, if applicable. **Not applicable** — no public API contract change.
- [x] Public documentation issue/PR created, if applicable. **Not applicable** — internal query execution correctness fix.


gingeekrishna:
### Description
This change makes bitmap scorer suppliers repeatable by rebuilding mutable intersection state on every ScorerSupplier#get(...) call instead of reusing state across calls.

What this fixes:
- Prevents stale iterator/intersection state from leaking between scorer instances.
- Fixes incorrect empty results seen in complex bool queries using bitmap terms lookup on multi-segment indices.
- Ensures each scorer retrieval is independent and deterministic.

Implementation details:
- Rebuilds DocIdSetBuilder and merge/intersection visitor on each ScorerSupplier#get(...) invocation.
- Avoids reusing mutable intersection artifacts between repeated get() calls.
- Keeps behavior consistent for both 32-bit and 64-bit bitmap index query paths.

Regression coverage:
- Added repeatability tests that call ScorerSupplier#get(...) twice and assert identical matches on both passes for:
  - BitmapIndexQueryTests (32-bit)
  - Bitmap64IndexQueryTests (64-bit)

### Related Issues
Resolves #21781

### Check List
- [x] API changes companion pull request created, if applicable. Not applicable (no API contract change).
- [x] Public documentation issue/PR created, if applicable. Not applicable (internal query execution correctness fix).

msfroh:
> BytesRef aliasing in bitmapEncodedIterator.next()

@gingeekrishna -- Can you please clarify the impact of this? This is a pretty standard pattern in Lucene -- you allocate a buffer and reuse it as you iterate. As long as the buffer is only held by a single thread, it's okay. The alternative (using the `deepCopyOf`) allocates a new array of 4 or 8 bytes on every iteration. Since these allocations are tracked by the memory circuit breaker and there could be millions of iterations, it's possible that the memory circuit breaker will trip. I'm not convinced that reusing the buffer is a problem (unless the iterator was accessed from multiple threads, which should be fixed by allocating a separate iterator per thread).

The other two root causes are good catches that I'm glad you fixed!

Also, @bowenlan-amzn, can you please take a look at this, as the original author of `BitmapIndexQuery`?

gingeekrishna:
Hi @msfroh, thank you for the detailed review!

You're right — I over-corrected on the `BytesRef` aliasing. Looking at `MergePointVisitor.matches()` and `compare()`, `nextQueryPoint` is always reassigned immediately after calling `iterator.next()`, so there's no scenario where a stale reference would cause incorrect comparisons. Returning the same reused buffer is safe here. I'll revert the `deepCopyOf` changes on both `BitmapIndexQuery` and `Bitmap64IndexQuery` to avoid the per-iteration allocation overhead.

The primary root cause was the `DocIdSetBuilder` and `MergePointVisitor` being initialized as `ScorerSupplier` fields — once `result.build()` is called, the builder is consumed and a second call to `get()` returns an empty result. The `hasBuffered = false` fix in `advance()` is an independent correctness bug I'll keep.

I'll push an updated version shortly with only those two fixes.

msfroh (approved):

gingeekrishna:
Hi @msfroh 

The `Sandbox Check` failure on this PR is unrelated to the bitmap scorer changes.

msfroh:
> Hi @msfroh
> 
> The `Sandbox Check` failure on this PR is unrelated to the bitmap scorer changes.

Agreed. Annoying that it failed, but I'm not going to consider it a blocker.

PR #21816, opened 2026-05-24
Fixes for query cache integration with analytics engine

bharath-techie:
### Description
We need shardId to be resolved via OpenSearchDirectoryReader for indices cache to be properly integrated with analytics engine. Earlier e2e tests didn't catch this as they were skipped with `@awaitsFix` annotation. So removed it, ran it 100 times for checking flakiness and moved the query cache tests to a new file as well.

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -199,17 +199,22 @@ private FragmentResources startFragment(FragmentExecutionRequest request, Resolv
 
                 if (task != null && taskResourceTrackingService != null) {
                     long taskId = task.getId();
+                    long fragmentThreadId = Thread.currentThread().threadId();
                     TaskResourceTrackingService service = taskResourceTrackingService;
                     backend.setDelegationThreadTracker(new DelegationThreadTracker() {
                         @Override
                         public long trackStart() {
                             long threadId = Thread.currentThread().threadId();
+                            if (threadId == fragmentThreadId) {
```
bharath-techie: Without this workaround, IT assertions fail that same threadId is getting tracked again. This is not the right fix - we need to check whether collector, provider calls need to be invoked on search threadpool and that should get tracked here.

Bukhtawar (approved):
LGTM

PR #21822, opened 2026-05-25
Analytics engine - add support for index patterns, aliases, multi-index search

mch2:
## Summary

  Adds support for alias, index pattern, and multi-index queries to the analytics engine. A query like `source=my_alias` now fans out across all backing indices, with schema widening to null-fill columns that individual shards don't have.

  **Planning:** When the planner encounters a table name that resolves to multiple indices, it validates that any field shared across indices has a compatible type (e.g. `text` and `keyword` both map to VARCHAR —
  fine; `long` vs `keyword` — rejected with a clear error). It then builds a single scan node whose row type is the union of all fields across all backing indices, and routes the query to shards on every backing index.

  **Execution:** Each shard registers its local parquet table with the schema inferred from its files. But the Substrait plan references the full union schema. To bridge the gap, during session creation extract the plan's declared schema (`base_schema`) and appends any columns this shard doesn't have as nullable. DataFusion's built-in adapter then produces null values for those columns at read time.

For single-index queries this widening is a no-op: the plan's schema matches the shard's schema, so the field-name comparison exits immediately with no work done.

 **Table name binding:** The plan references the logical name (e.g. `"my_alias"`), but each shard knows itself by its concrete index name (e.g. `"idx_a"`). Session creation extracts the logical name from the plan's `NamedTable` and registers the table under it, so the Substrait consumer can bind. For single-index queries, the logical name equals the concrete name — same code path, no branching.

  **Wildcard support (test frontend):** The PPL frontend's schema lookup was flattening tables into a static map, losing the lazy resolution that handles wildcards. Fixed by delegating `get()` to the underlying
  schema which resolves expressions on demand.

  **Filter delegation (indexed path):** The indexed execution path (filter delegation to Lucene) now also receives plan bytes, enabling multi-index support for queries with `MATCH` predicates across aliases.

### Related Issues

### Check List


sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/OpenSearchSchemaBuilder.java
```diff
@@ -11,56 +11,135 @@
...
+     */
+    @SuppressWarnings("unchecked")
+    private static Table resolveTable(ClusterState clusterState, IndexNameExpressionResolver resolver, String expression) {
+        String[] concrete;
+        try {
+            // Comma-split first: concreteIndexNames treats each vararg as one expression, and
+            // splitting lets the resolver honor exclusions across tokens (e.g. "test*,-test1").
+            concrete = resolver.concreteIndexNames(
+                clusterState,
+                IndicesOptions.lenientExpandOpen(),
```
sandeshkr419: Do we need param supports equivalent to `ignore_unavailable`, `allow_no_indices`, and `expand_wildcards` instead of this hard-coded `IndicesOptions.lenientExpandOpen()`? 

Or are you handling such cases elsewhere?

sandeshkr419 (commented):
Thanks for the changes @mch2 

Let's add up test cases for these as well if:

- case-insensitive table resolution
- test for `findTableName` on a multi-input join/union shape
- rust tests for `widen_schema_from_plan`
- `first_named_table_name` / `base_schema_for_table` with multi-scan plans
- nested field schema mismatch
- IndexResolutionTests tests for the closed-index case on the concrete-index path.
- `IndexResolutionTests.testMissingNameThrows` doesn't pass a resolver, so it can't actually exercise the wildcard fallback, check once.
- `OpenSearchSchemaBuilderTests.testCommaSeparatedSourcesResolveToUnionedTable` doesn't test field-conflict behavior.

(I may have missed in the code if some of the above tests actually exist and I overlooked through them - just validate that they do please)

sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/OpenSearchSchemaBuilder.java
mch2: i don't know that we need to support these yet, as we don't get these from front-end.  This defaults to open indices only which is the default in _search endpoint.

sandeshkr419 (approved):

sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/OpenSearchSchemaBuilder.java
```diff
@@ -11,56 +11,135 @@
...
+                {
+                    // Eagerly register concrete indices and aliases so getTableNames() lists them —
+                    // case-insensitive resolution (Calcite uppercases unquoted identifiers) matches
+                    // against that set. Wildcard/comma/exclusion expressions aren't enumerable and
+                    // resolve lazily in get().
+                    for (String name : clusterState.metadata().getIndicesLookup().keySet()) {
+                        Table table = resolveTable(clusterState, resolver, name);
+                        if (table != null) {
+                            super.put(name, table);
+                        }
```
Bukhtawar: This might be pretty expensive right?

mch2: doh, yes fixing this to be lazy at get time like sql does today.

PR #21823, opened 2026-05-25
[analytics-engine] Wire dc/earliest/latest/nth_value to DataFusion built-ins

songkant-aws:
### Description

Wires PPL `dc`/`distinct_count`/`distinct_count_approx`, `earliest`/`latest`, and `nth_value` to DataFusion built-ins on the analytics-engine route.

Backend rewrite via a new per-function `WindowFunctionAdapter` SPI (parallel to `ScalarFunctionAdapter`). DataFusion plugin registers three rewrites that `BackendPlanAdapter` applies before substrait emission:

- `ARG_MIN(value, ts)` → `FIRST_VALUE(value) ORDER BY ts ASC` (PPL `earliest()`)
- `ARG_MAX(value, ts)` → `LAST_VALUE(value) ORDER BY ts ASC` (PPL `latest()`)
- `DISTINCT_COUNT_APPROX(x)` → `APPROX_COUNT_DISTINCT(x)` (PPL `dc()` / `distinct_count()` / `distinct_count_approx()`)

A small Rust wrapper UDAF aliases substrait's `approx_count_distinct` name to DataFusion's built-in `approx_distinct` so end-to-end execution works.

Unit-level coverage in `WindowFunctionAdaptersTests` pins the rewrite contracts. End-to-end IT in `EventstatsCommandIT` and `StreamstatsCommandIT` (8 dc methods, plus single-shard and multi-shard variants for dc / dc by) assert exact rows.

### Why this PR depends on opensearch-project/sql#5466

`PPLFuncImpTable.INSTANCE` is a JVM-process-level singleton with a mutable `aggExternalFunctionRegistry` (ConcurrentHashMap). On current sql-plugin main, `DISTINCT_COUNT_APPROX` is registered into that map only as a side effect of constructing `OpenSearchExecutionEngine` — which is Guice-bound lazily and only ever instantiated when the V3 PPL path is exercised. Because the singleton registry is shared across paths once populated, the registration *does* become visible to the unified-query path eventually — but only if the running process has serviced at least one V3 request first.

Our IT environment only sends requests through `/_analytics/ppl` (`RestUnifiedQueryAction`), never through V3, so `OpenSearchExecutionEngine` is never constructed and `DISTINCT_COUNT_APPROX` is never registered. PPL parser then throws "Cannot resolve function: DISTINCT_COUNT_APPROX" before our backend rewrite gets a chance to run. opensearch-project/sql#5466 moves the registration into `PPLFuncImpTable.populate()` (class-init self-registration) using a logical marker UDAF, eliminating the V3-path precondition entirely.

### Related Issues

- Depends on opensearch-project/sql#5466 for the dc-IT path (positive `assertRowsEqual` instead of negative `assertErrorContains`).

### Check List


sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/EventstatsCommandIT.java
```diff
@@ -934,44 +934,93 @@ public void testEventstatsVarianceWithNullBy() throws IOException {
...
+    // calcs.str3 is "e" on every non-null row → dc(str3)=1. calcs.str0 has three values
+    // {FURNITURE, OFFICE SUPPLIES, TECHNOLOGY} on every row → dc(str0)=3. eventstats
+    // broadcasts the global aggregate to every row.
 
-    /** sql IT: testEventstatsDistinctCount. {@code dc()} resolves to DISTINCT_COUNT_APPROX in
-     *  the PPL parser; that aggregate isn't registered in analytics-engine, so the request
-     *  fails before reaching the window-function gate with
-     *  {@code "Cannot resolve function: DISTINCT_COUNT_APPROX"}. */
+    /** sql IT: testEventstatsDistinctCount. */
     public void testEventstatsDistinctCount() throws IOException {
```
mch2, line 945: can we test these with multi shard case? calcs/mapping.json - i have tried setting for clickbench tests here https://github.com/opensearch-project/OpenSearch/pull/21848 to add randomization 1-2 shard, we can do somethign similar here.

We'll likely have variance with dc() being approximate, but to ensure the queries succeed.

songkant-aws: Added multi-shard variants in 31702eac7bb:

- `testEventstatsDistinctCount_3shard` and `testEventstatsDistinctCountByCountry_3shard` reuse the existing `DATASET_MULTI` (3 shards) — eventstats dc is broadcast pattern so the per-row expectations are identical to single-shard, and HLL is exact at calcs's 17-row scale.
- `testStreamstatsDistinctCountByCountry_3shard` collapses to per-partition finals via `| stats max(dc_str3) by str0 | sort str0` since streamstats running values aren't deterministic when shard parallelism reorders input rows — same shape as `testStreamstatsBy_3shard`.

Didn't go with `DatasetProvisioner.provision`-level random shard count for this PR since #21848's randomization helper isn't merged yet; sticking with the existing per-test 3-shard pattern this file already uses.

songkant-aws: Sanity-checked datafusion 53.1's HLL with a standalone Rust harness running the real algorithm over 3 shards × 100 iterations — `dc(str3)` always returns 1 and `dc(str0)` always returns 3, with zero variance. The merge step is just register-wise max with a fixed seed, so multi-shard and single-shard paths produce bit-identical state, and for these two specific values the count output sits well clear of any rounding boundary.

sandeshkr419 (changes requested):
I see you are doing 2 things:

Part 1: Wire dc/earliest/latest/nth_value: no comments for now, good to go!
Part 2: Unify reduce-stage subtree conversion: This part should simplify after my changes: https://github.com/opensearch-project/OpenSearch/pull/21775 (aiming to get this in today). I have introduced state-shipping which unifies all agg state as Binary IPC, so the schema never shifts between SINGLE and FINAL - this should simplify your code a lot.

songkant-aws:
@sandeshkr419 thanks for the review. Rebased on main — both #21775's `reduce_eval` Binary-IPC path and #21875's `overrideExchangeType` transformer are on this branch now. Those address PARTIAL/FINAL aggregate row-type drift at the rel layer for STATE_EXPANDING / opaque-state aggregates so the schema doesn't shift between SINGLE and FINAL.

Could you take another look?

sandbox/plugins/analytics-backend-datafusion/rust/src/udaf/approx_count_distinct.rs
sandeshkr419, line 1: We already have `FunctionMappings.s(APPROX_COUNT_DISTINCT, "approx_distinct")` mapping it correctly for aggregates. Check if the window-function substrait emission bypasses this. If so, either fix the emission to use the same mapping, or just add a second YAML entry name: `approx_count_distinct` pointing to the same function. No Rust wrapper should be needed.

sandeshkr419 (changes requested):
1 major comment on support of `dc`, rest looks fine. 

While you are addressing this, can you also please concise up the comments in your changes to avoid overflowing intermittent design decisions.

sandbox/plugins/analytics-backend-datafusion/rust/src/udaf/approx_count_distinct.rs
songkant-aws: You're right. Missed the parallel agg mapping. Wrapper dropped, comments cleaned up.

sandeshkr419 (approved):

PR #21824, opened 2026-05-25
Hybrid approach

Ashu-tosh27:
### Description
This PR introduces a new POST /{index}/_star_tree/upgrade API that retroactively adds star tree indexes to existing indices without requiring re-indexing.

Today, star tree indexes can only be configured at index creation time. For large indices with terabytes of historical data, re-indexing is prohibitively expensive. This feature enables in-place upgrade by building star tree data directly from existing doc values on a per-segment basis.

### How it works
The upgrade is a 3-phase operation per shard:

1. Mapping update — Adds the star tree field to the index mapping via cluster state, bypassing the index.composite_index Final setting restriction using MergeReason.STAR_TREE_UPGRADE. Also force-sets index.composite_index=true and index.append_only.enabled=true.

2. Per-segment star tree build — Swaps to a ReadOnlyEngine (reads remain available), then builds star tree files (.cid, .cim, .cidvd, .cidvm) for each segment in parallel. Deleted documents (both hard and soft deletes) are filtered out using a FixedBitSet + LiveDocsFilteredDocValuesProducer wrapper.
 
3. SegmentInfos rewrite — For clean segments (docValuesGen == -1), switches codec to Composite912Codec. For segments with soft deletes (docValuesGen != -1), keeps the original codec and serves star tree data via a StarTreeDirectReader cache that bypasses the codec pipeline.

After upgrade, new segments are written natively with Composite104Codec and background merges gradually converge all segments to native composite format.


server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
```diff
@@ -0,0 +1,658 @@
...
+                    commitInfo.info.name);
+            }
+        }
+
+        return liveBits;
+    }
+
+    public static void rewriteSegmentInfos(Directory directory, Set<String> upgradedSegmentNames) throws IOException {
+        Lock writeLock = directory.obtainLock(IndexWriter.WRITE_LOCK_NAME);
+        try {
```
Shailesh-Kumar-Singh: we can do `try (Lock writeLock = directory.obtainLock(IndexWriter.WRITE_LOCK_NAME))` instead?

server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
```diff
@@ -0,0 +1,658 @@
...
+        int softDeleteCount = commitInfo.getSoftDelCount();
+
+        logger.debug("buildLiveDocsBitset: segment={} hardDel={} softDel={} maxDoc={}",
+            commitInfo.info.name, hardDeleteCount, softDeleteCount, maxDoc);
+
+        if (hardDeleteCount == 0 && softDeleteCount == 0) {
+            return null;
+        }
+
+        org.apache.lucene.util.FixedBitSet liveBits = new org.apache.lucene.util.FixedBitSet(maxDoc);
```
Shailesh-Kumar-Singh: `org.apache.lucene.util.FixedBitSet` - import and use, change in all places.

server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
```diff
@@ -0,0 +1,658 @@
...
+        if (hardLiveDocs != null) {
+            for (int i = 0; i < maxDoc; i++) {
+                if (hardLiveDocs.get(i) == false) {
+                    liveBits.clear(i);
+                }
+            }
+        }
+
+        // Apply soft deletes via segmentReader.getNumericDocValues() which routes to the update file.
+        String softDeleteField = org.opensearch.common.lucene.Lucene.SOFT_DELETES_FIELD;
```
Shailesh-Kumar-Singh: `org.opensearch.common.lucene.Lucene.SOFT_DELETES_FIELD`
import and use

server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
```diff
@@ -0,0 +1,658 @@
...
+                        commitInfo.info.getCodec().getName()
+                    );
+                    // Add star tree files to file set so IndexWriter doesn't GC them.
+                    SegmentInfo oldInfo = commitInfo.info;
+                    Set<String> files = new HashSet<>(oldInfo.files());
+                    String segName = oldInfo.name;
+                    files.add(IndexFileNames.segmentFileName(segName, "", Composite912DocValuesFormat.DATA_EXTENSION));
+                    files.add(IndexFileNames.segmentFileName(segName, "", Composite912DocValuesFormat.META_EXTENSION));
+                    files.add(IndexFileNames.segmentFileName(segName, "", Composite912DocValuesFormat.DATA_DOC_VALUES_EXTENSION));
+                    files.add(IndexFileNames.segmentFileName(segName, "", Composite912DocValuesFormat.META_DOC_VALUES_EXTENSION));
```
Shailesh-Kumar-Singh: duplicate code snippet below as well [ L598 to L601], can be extracted to a helper function

server/src/main/java/org/opensearch/index/compositeindex/CompositeIndexValidator.java
```diff
@@ -33,7 +34,23 @@ public static void validate(
...
+        boolean isCompositeFieldPresent,
+        MergeReason mergeReason
+    ) {
+        if (mergeReason == MergeReason.STAR_TREE_UPGRADE) {
+            // Skip the "no new composite fields during update" check for star tree upgrade.
+            // Still validate dims/metrics against existing fields.
+            StarTreeValidator.validate(mapperService, compositeIndexSettings, indexSettings);
+            return;
+        }
+        if (isCompositeFieldPresent == false && mapperService.isCompositeIndexPresent()) {
```
Shailesh-Kumar-Singh: we can just do this?
```
      if (mergeReason != MergeReason.STAR_TREE_UPGRADE
          && isCompositeFieldPresent == false
          && mapperService.isCompositeIndexPresent()) {
          throw new IllegalArgumentException(
              "Composite fields must be specified during index creation, "
                  + "addition of new composite fields during update is not supported"
          );
      }
      StarTreeValidator.validate(mapperService, compositeIndexSettings, indexSettings);
```

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -2515,10 +2520,225 @@ public void resetToWriteableEngine() throws IOException, InterruptedException, T
...
+        logger.info("{} starting per-segment star tree upgrade", shardId);
+
+        final AtomicInteger upgradedCount = new AtomicInteger(0);
+        try {
+            // First flush: reduce work needed after blocking
+            flush(new FlushRequest().force(true));
+
+            indexShardOperationPermits.blockOperations(30, TimeUnit.MINUTES, () -> {
+                // Second flush: ensure all data committed before engine close.
+                flush(new FlushRequest().waitIfOngoing(true));
```
Shailesh-Kumar-Singh, line 2549: given that we are already flushing at L2546 [ just before the block ] ,is this flush for the in-flight writes?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -2515,10 +2520,225 @@ public void resetToWriteableEngine() throws IOException, InterruptedException, T
...
+                }
+                // Check if this segment has star tree files in its file set
+                String cimFile = IndexFileNames.segmentFileName(segName, "", Composite912DocValuesFormat.META_EXTENSION);
+                boolean hasStarTreeFiles = false;
+                try {
+                    for (String file : commitInfo.files()) {
+                        if (file.equals(cimFile)) {
+                            hasStarTreeFiles = true;
+                            break;
+                        }
```
Shailesh-Kumar-Singh: just do `hasStarTreeFiles = commitInfo.files().contains(cimFile);`

server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
```diff
@@ -0,0 +1,658 @@
...
+                4096, /* Lucene90DocValuesFormat#DEFAULT_SKIP_INDEX_INTERVAL_SIZE */
+                Composite912DocValuesFormat.DATA_DOC_VALUES_CODEC,
+                Composite912DocValuesFormat.DATA_DOC_VALUES_EXTENSION,
+                Composite912DocValuesFormat.META_DOC_VALUES_CODEC,
+                Composite912DocValuesFormat.META_DOC_VALUES_EXTENSION
+            );
+
+            // Build star tree data using StarTreesBuilder
+            try (StarTreesBuilder starTreesBuilder = new StarTreesBuilder(state, mapperService, new AtomicInteger())) {
+                starTreesBuilder.build(metaOut, dataOut, fieldProducerMap, compositeDocValuesConsumer);
```
Shailesh-Kumar-Singh, line 433: If this throws exception, the outputs have partial data written. In finally block we are closing them, this would persist corrupt/partial files on disk?

server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
```diff
@@ -0,0 +1,658 @@
...
+        directory.syncMetaData();
+
+        logger.info(
+            "SegmentInfos rewrite complete — committed new segment infos for {} upgraded segments. "
+                + "Generation: {}, files: {}, directory listing: {}",
+            upgradedSegmentNames.size(),
+            newSegmentInfos.getGeneration(),
+            newSegmentInfos.files(true),
+            java.util.Arrays.toString(directory.listAll())
+        );
```
Shailesh-Kumar-Singh: `java.util.Arrays.toString(directory.listAll())`
This would list all directories, is expensive for info logs
Use debug log or remove.

server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
```diff
@@ -0,0 +1,658 @@
...
+            // Open a DirectoryReader and find the matching SegmentReader by segment name
+            directoryReader = DirectoryReader.open(directory);
+            SegmentReader segmentReader = null;
+            for (LeafReaderContext leafContext : directoryReader.leaves()) {
+                SegmentReader candidate = Lucene.segmentReader(leafContext.reader());
+                if (candidate.getSegmentName().equals(segmentName)) {
+                    segmentReader = candidate;
+                    break;
+                }
+            }
```
Shailesh-Kumar-Singh, line 311: instead of opening every segment just to find one, can we just directly open the target segment?

server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
Ashu-tosh27: Done

server/src/main/java/org/opensearch/index/compositeindex/datacube/startree/StarTreeUpgradeService.java
Ashu-tosh27: Done

server/src/main/java/org/opensearch/index/shard/IndexShard.java
Ashu-tosh27: Yes, the second flush is for in-flight writes that completed between the first flush and the operation block being acquired.

server/src/main/java/org/opensearch/action/admin/indices/startree/RestStarTreeUpgradeAction.java
```diff
@@ -0,0 +1,53 @@
...
+ * and delegates to {@link TransportStarTreeUpgradeAction} via
+ * {@link StarTreeUpgradeAction}.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class RestStarTreeUpgradeAction extends BaseRestHandler {
+
+    @Override
+    public List<Route> routes() {
```
bharath-techie, line 36: Can we reuse existing index upgrade action , we can add optional params for star tree which routes to star tree based upgrade ?

server/src/main/java/org/opensearch/action/admin/indices/startree/StarTreeUpgradeRequest.java
```diff
@@ -0,0 +1,345 @@
...
+ * A request to upgrade an existing index by adding star tree configuration.
+ * The star tree field configuration is parsed from the request body and
+ * serialized as XContent bytes for transport across nodes.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public class StarTreeUpgradeRequest extends BroadcastRequest<StarTreeUpgradeRequest> {
+
+    private static final String STAR_TREE = "star_tree";
```
bharath-techie, line 35: How to reuse this from existing star tree field mapper ? this has a challenge that if star tree mapper adds it , we might miss adding functionality here.

Ashu-tosh27: The request now stores the raw config JSON as opaque BytesReference and only extracts the name field . All actual parsing, validation, and feature handling happens exclusively through StarTreeMapper.Builder when TransportStarTreeUpgradeAction calls mapperService.merge() with MergeReason.STAR_TREE_UPGRADE.

So if the mapper adds new functionality in the future, the upgrade path will pick it up automatically .

server/src/main/java/org/opensearch/action/admin/indices/startree/RestStarTreeUpgradeAction.java
Ashu-tosh27: We can try adding a param to the existing _upgrade endpoint and route to our transport action from there . But the two operations are a little different . the existing one doesnt take a request body and is a simple per shard segment format upgrade but our upgrade is a two phase upgrade and requires a body so doing that will need some significant changes . Open to going either way though

PR #21825, opened 2026-05-25
Fix SSE blob store reload during repository metadata update

ask-kamal-nayan:
#### Summary
  
1. When repository metadata was updated on an SSE-enabled S3 repository (e.g., rotating credentials, changing endpoint), S3Repository.reload() only invoked reload() on the result of getBlobStore(), which returned only the non-SSE blob store.
  The serverSideEncryptedBlobStore was silently skipped, leaving SSE-enabled repositories using stale config until node restart.

2. Added "lucene/" in `EXCLUDED_SUBDIRECTORIES` so that the files in this directory doesn't come in the file list of `SubdirectoryAwareDirectory`.
  
  #### Bug
  
  When repository metadata was updated on an SSE-enabled S3 repository (e.g., rotating credentials, changing endpoint), S3Repository.reload() only invoked reload() on the result of getBlobStore(), which returned only the non-SSE blob store.
  The serverSideEncryptedBlobStore was silently skipped, leaving SSE-enabled repositories using stale config until node restart.
  
  #### Changes
  
  - BlobStoreProvider.reloadBlobStore(RepositoryMetadata) — new method that reloads both the non-SSE and SSE blob stores when initialized.
  - S3Repository.reload() — now delegates to blobStoreProvider.reloadBlobStore() instead of reloading only one store.
  - BlobStoreRepository.blobStoreProvider — visibility relaxed from private to protected so subclasses can invoke the provider.
  

### Related Issues

### Check List

### DIFF
--- a/plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/S3Repository.java
+++ b/plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/S3Repository.java
@@ -596,8 +596,9 @@ class S3Repository extends MeteredBlobStoreRepository {
         s3AsyncService.releaseCachedClients();
 
         // Reload configs for S3BlobStore
-        BlobStore blobStore = getBlobStore();
-        blobStore.reload(metadata);
+        if (blobStoreProvider.get() != null) {
+            blobStoreProvider.get().reloadBlobStore(newRepositoryMetadata);
+        }
     }
 
     /**
--- a/server/src/main/java/org/opensearch/index/store/SubdirectoryAwareDirectory.java
+++ b/server/src/main/java/org/opensearch/index/store/SubdirectoryAwareDirectory.java
@@ -39,7 +39,7 @@ import java.util.stream.Collectors;
  */
 public class SubdirectoryAwareDirectory extends FilterDirectory {
     private static final Logger logger = LogManager.getLogger(SubdirectoryAwareDirectory.class);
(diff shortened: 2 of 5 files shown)


gbbafna (approved):

plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/S3Repository.java
```diff
@@ -596,8 +596,9 @@ public void reload(RepositoryMetadata newRepositoryMetadata) {
         s3AsyncService.releaseCachedClients();
 
         // Reload configs for S3BlobStore
-        BlobStore blobStore = getBlobStore();
-        blobStore.reload(metadata);
+        if (blobStoreProvider.get() != null) {
```
Bukhtawar, line 599: When is it null?

Bukhtawar (approved):

plugins/repository-s3/src/main/java/org/opensearch/repositories/s3/S3Repository.java
ask-kamal-nayan: `blobStoreProvider` is a SetOnce that's lazy-initialized inside `BlobStoreRepository.blobStore(boolean)`. So .get() is null until the first blob-store operation on this node materializes it.

mgodwan (approved):

PR #21828, opened 2026-05-25
[DFA Tiering] Enforce write blocks for DFA H2W tiering, survive DiskThresholdMonitor removal, and set replicas to 1`

vishwasgarg18:
### Description

This PR addresses write-block correctness and engine improvements for DFA (pluggable dataformat) tiering.

__1. Write block enforcement during H2W tiering (`blocks.write` only)__

For DFA indices, `TransportHotToWarmTierAction.addWriteBlockAndPrepare()` adds `INDEX_WRITE_BLOCK` (`blocks.write=true`) before pre-tiering sync (flush + refresh + remote store upload). Unlike `read_only_allow_delete`, `blocks.write` cannot be auto-removed by `DiskThresholdMonitor` under disk pressure, ensuring writes remain blocked through the entire tiering operation.

- Both `IndexMetadata` settings (`blocks.write=true`) and `ClusterBlocks` (`INDEX_WRITE_BLOCK`) are updated atomically: settings survive CM failover/restart, ClusterBlocks enforce the block immediately at transport layer.
- `removeReadOnlyBlock()` (renamed `removeWriteBlock()` internally) clears `blocks.write` on prepare failure to avoid leaving the index stuck.

__2. Template method pattern for ClusterBlocks management in TieringService__

Added two abstract template methods, symmetric with the existing settings pattern:

- `getTieringStartClusterBlocksToAdd(blocksBuilder, indexName)` — `HotToWarmTieringService`: no-op (blocks set by `TransportHotToWarmTierAction` before `tier()`); `WarmToHotTieringService`: removes `INDEX_WRITE_BLOCK` (index going to HOT).
- `getIndexTierClusterBlocksToRestoreAfterCancellation(blocksBuilder, indexName)` — `HotToWarmTieringService`: removes `INDEX_WRITE_BLOCK` (back to HOT); `WarmToHotTieringService`: restores `INDEX_WRITE_BLOCK` (back to WARM).

Both template methods in `TieringService.tier()` and `cancelTiering()` are guarded by `TieringUtils.isDfaIndex()` — non-DFA indices never have write blocks applied.

__3. Write-block settings guarded by DFA check in subclasses__

`getTieringStartSettingsToAdd(IndexMetadata)` and `getIndexTierSettingsToRestoreAfterCancellation(IndexMetadata)` now receive `IndexMetadata` and conditionally include `INDEX_BLOCKS_WRITE_SETTING` only for DFA indices. This prevents the bug where cancelling W2H tiering on a non-DFA index would incorrectly set `blocks.write=true`, permanently blocking writes on an index that was never blocked.

__4. `number_of_replicas` normalization during tiering start__

`updateIndexMetadataForTieringStart()` now atomically sets `number_of_replicas=1` in both `IndexMetadata` and the routing table when `currentReplicas != 1`, ensuring `ReplicationTracker` consistency on warm nodes during shard relocation.

__5. `CatalogSnapshotStatsCache` in `DataFormatAwareReadOnlyEngine`__

Stats methods (`docStats()`, `segmentsStats()`, `segments()`) previously acquired a new catalog snapshot on every call. For the read-only engine the snapshot is permanent, so stats are computed once at engine construction via `statsCache.forceRefresh()` and served from cached fields on subsequent calls — O(1) instead of O(segments).

__6. `remotePtr != 0L` guard in `ParquetDataFormatStoreHandler`__

Added an early `IllegalStateException` when `remotePtr == 0L` (null or dead `NativeStoreRepository`) to fail fast before any Rust FFM call, preventing undefined behavior.

### Related Issues

Resolves #[Issue number]

### Check List

- Functionality includes testing.
- API changes companion pull request created, if applicable.
- Public documentation issue/PR created, if applicable.


server/src/main/java/org/opensearch/storage/tiering/TieringService.java
```diff
@@ -527,11 +563,11 @@ void updateIndexMetadataForTieringStart(
...
     ) {
         try {
-            // 1. Build settings
+            // 1. Build settings.
             Settings.Builder indexSettingsBuilder = Settings.builder().put(indexMetadata.getSettings()).put(getTieringStartSettingsToAdd());
 
-            // 2. Handle replica updates using auto_expand_replicas
-            indexSettingsBuilder.put(IndexMetadata.SETTING_AUTO_EXPAND_REPLICAS, "0-" + 1);
+            // 2. Always set number_of_replicas to 1 to ensure routing table consistency.
+            indexSettingsBuilder.put(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, 1);
```
nishchay21: lets get current replicas and if current replica != as previous then only make this change

server/src/main/java/org/opensearch/storage/tiering/TieringService.java
```diff
@@ -544,6 +580,15 @@ void updateIndexMetadataForTieringStart(
...
 
             metadataBuilder.put(indexMetadataBuilder);
+
+            // 5. Update routing table to ensure ReplicationTracker consistency.
+            // This must happen in the same cluster state update to keep the routing table
+            // consistent with the metadata. Without this, the ReplicationTracker on the warm
+            // node may see stale allocation IDs during shard relocation, causing an assertion
+            // failure in renewPeerRecoveryRetentionLeases.
+            final String[] indices = new String[] { index.getName() };
+            routingTableBuilder.updateNumberOfReplicas(1, indices);
```
nishchay21: same here add current replica check as before

server/src/main/java/org/opensearch/storage/tiering/HotToWarmTieringService.java
```diff
@@ -130,9 +131,24 @@ protected Settings getIndexTierSettingsToRestoreAfterCancellation() {
...
+    protected ClusterBlocks.Builder getTieringStartClusterBlocksToAdd(ClusterBlocks.Builder blocksBuilder, String indexName) {
+        return blocksBuilder; // no-op: write blocks set by TransportHotToWarmTierAction before tier()
+    }
+
+    @Override
+    protected ClusterBlocks.Builder getIndexTierClusterBlocksToRestoreAfterCancellation(
+        ClusterBlocks.Builder blocksBuilder,
+        String indexName
+    ) {
+        return blocksBuilder.removeIndexBlock(indexName, IndexMetadata.INDEX_READ_ONLY_ALLOW_DELETE_BLOCK)
```
nishchay21: we should only keep write block readonly block should be removed

vishwasgarg18: Ack, removed

server/src/main/java/org/opensearch/storage/tiering/TieringService.java
vishwasgarg18: Ack, reverted to how it was before our change

server/src/main/java/org/opensearch/storage/tiering/TieringService.java
vishwasgarg18: Done

server/src/main/java/org/opensearch/storage/tiering/WarmToHotTieringService.java
```diff
@@ -107,10 +108,23 @@ protected Settings getIndexTierSettingsToRestoreAfterCancellation() {
             .put(IS_WARM_INDEX_SETTING.getKey(), true)
             .put(INDEX_TIERING_STATE.getKey(), WARM)
             .put(INDEX_COMPOSITE_STORE_TYPE_SETTING.getKey(), TIERED_COMPOSITE_INDEX_TYPE)
-            .put(IndexMetadata.INDEX_BLOCKS_READ_ONLY_ALLOW_DELETE_SETTING.getKey(), true)
+            .put(IndexMetadata.INDEX_BLOCKS_WRITE_SETTING.getKey(), true)
```
nishchay21: this will cause write block on lucene lets verify and fix this

server/src/main/java/org/opensearch/storage/common/tiering/TieringUtils.java
```diff
@@ -372,4 +374,22 @@ public String value() {
         }
     }
 
+    /**
+     * Checks if the index has pluggable data format enabled (is a DFA index).
+     */
+    public static boolean isDfaIndex(String indexName, ClusterState state) {
```
mgodwan: [nit] the 2 methods can be combined

server/src/main/java/org/opensearch/storage/tiering/HotToWarmTieringService.java
```diff
@@ -124,13 +126,28 @@ protected Settings getTieringStartSettingsToAdd() {
...
+            .put(INDEX_COMPOSITE_STORE_TYPE_SETTING.getKey(), "default");
+        if (TieringUtils.isDfaIndex(indexMetadata)) {
+            builder.put(IndexMetadata.INDEX_BLOCKS_WRITE_SETTING.getKey(), false);
+        }
+        return builder.build();
+    }
+
+    @Override
+    protected ClusterBlocks.Builder getTieringStartClusterBlocksToAdd(ClusterBlocks.Builder blocksBuilder, String indexName) {
+        return blocksBuilder; // no-op: write blocks set by TransportHotToWarmTierAction before tier()
```
mgodwan: We should apply the block directly here

vishwasgarg18: Done

server/src/main/java/org/opensearch/storage/common/tiering/TieringUtils.java
vishwasgarg18: Updated

mgodwan (approved):

PR #21830, opened 2026-05-26
Revert NonClosingReaderWrapper cache (#20921)

bugmakerrrrrr:
### Description
This reverts the NonClosingReaderWrapper cache introduced in #20921.

NonClosingReaderWrapper only prevents close() from propagating to the underlying DirectoryReader. The wrapper itself is still a normal IndexReader with its own closed state. Once a wrapped searcher closes the cached NonClosingReaderWrapper, that wrapper becomes closed even though the underlying reader remains open.

If the cache later returns the same NonClosingReaderWrapper for another wrapped searcher, the caller may operate on an already-closed wrapper and hit AlreadyClosedException, for example when accessing leaves(). Therefore the wrapper should remain scoped to a single wrapped searcher instead of being shared from a cache.

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/opensearch/index/shard/IndexShard.java
@@ -252,7 +252,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletionService;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorCompletionService;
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -4595,9 +4547,7 @@ public void beforeRefresh() {}
                 public void afterRefresh(boolean didRefresh) {
                     if (!didRefresh) return;
                     // Use the engine directly (not IndexShard.acquireSearcher) so that we do NOT
-                    // go through IndexShard.wrapSearcher / nonClosingReaderWrapperSupplier.
```
kkewwei, line 4598: unrelated changes

kkewwei (approved):

PR #21831, opened 2026-05-26
Inline merge at refresh for composite engine: produce one segment per…

Bukhtawar:
… refresh cycle

When multiple writers flush at refresh time, consolidate them into a single segment per format using the existing composite merger. This eliminates small-file proliferation (previously N files per refresh, now 1) without changing total disk I/O — the merge that would happen later in the background is simply moved inline.

Key changes:
- RefreshInput: add consolidationGeneration field to signal consolidation
- DataFormatAwareEngine: allocates a consolidation generation when multiple writers flush
- CompositeIndexingExecutionEngine.refresh(): when shouldConsolidate() is true, calls CompositeMerger.merge() on the per-writer segments to produce one consolidated segment with correct sort + RowIdMapping alignment
- ParquetIndexingEngine: add consolidateAndWrite() for future in-memory path
- LuceneIndexingExecutionEngine: unchanged (addIndexes(Directory...) as before, composite merger handles Lucene merge during consolidation)

Result: catalog sees 1 Parquet file + 1 Lucene segment per refresh cycle instead of N+N. Fewer files for readers, remote store, and query path. Same I/O budget — merge cost moved from background to refresh inline.

### Description

### Related Issues

### Check List


server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -971,7 +971,8 @@ public void refresh(String source) throws EngineException {
                         notifyRefreshListenersBefore();
                         if (refreshed) {
                             final long engineRefreshStartNanos = System.nanoTime();
-                            RefreshInput refreshInput = new RefreshInput(existingSegments, newSegments);
+                            long nextGen = newSegments.size() > 1 ? writerGenerationCounter.incrementAndGet() : RefreshInput.NO_GENERATION;
+                            RefreshInput refreshInput = new RefreshInput(existingSegments, newSegments, nextGen, refreshLock::unlock);
```
mgodwan: Instead of this, should we change warm hook logic to:

```
this.committer = engineConfig.getCommitterFactory().getCommitter(new CommitterConfig(engineConfig, () -> {
                if (!refreshLock.isHeldByCurrentThread()) {
                    refreshLock.lock();
                }
            }));
```

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -971,7 +971,8 @@ public void refresh(String source) throws EngineException {
                         notifyRefreshListenersBefore();
                         if (refreshed) {
                             final long engineRefreshStartNanos = System.nanoTime();
-                            RefreshInput refreshInput = new RefreshInput(existingSegments, newSegments);
+                            long nextGen = newSegments.size() > 1 ? writerGenerationCounter.incrementAndGet() : RefreshInput.NO_GENERATION;
```
mgodwan, line 910: Can this be a supplier instead of calculating here? That way generation is used only if composite decides to merge on refresh

mgodwan (approved):

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/CompositeIndexingExecutionEngine.java
```diff
@@ -204,24 +208,97 @@ public Merger getMerger() {
...
+            Set<Long> existingGens = refreshInput.existingSegments().stream().map(Segment::generation).collect(Collectors.toSet());
+            List<Segment> onlyNew = newSegments.stream().filter(s -> existingGens.contains(s.generation()) == false).toList();
+
+            if (onlyNew.size() > 1) {
+                Merger primaryMerger = primaryEngine.getMerger();
+                MergeInput primaryMergeInput = MergeInput.builder()
+                    .segments(onlyNew)
+                    .newWriterGeneration(refreshInput.nextAvailableGeneration())
+                    .build();
+                MergeResult primaryResult = primaryMerger.merge(primaryMergeInput);
```
rayshrey: In case of any errors/exceptions thrown from the merges looks like we will fail the refresh as well ?
Shall we fallback to returning the normal refresh results in such cases ?

Can we also add tests for the same as well ?

(For parquet since the merge will happen via the merge path, it will not be aware that this is happening on the write path, and might treat this as a background merge request - in case we have separate pools for merges/writes and lot of background merges are happening, merges might be rejected in such cases)

Bukhtawar: makes sense

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
Bukhtawar: Simplified this

PR #21834, opened 2026-05-26
Add derived-source nested mapping support 

### LINKED ISSUE #9568: Optimizing Data Storage and Retrieval for Time Series data.
In OpenSearch, documents are stored using three primary formats: indexed, stored, and docValues. In case of time series data, a document usually consists of dimensions, time point and quantitative measurements that are used monitor various aspects of a system, process, or phenomenon. In cases like these, the numeric, keyword and similar datatypes are stored as the "stored" field as well as docValues that serves specific purposes related to search performance and retrieval. DocValues are a columnar storage format used by Lucene to store indexed data in a way that facilitates efficient aggregation, sorting etc and stored fields, on the other hand, are used to store the actual values of fields as they were inserted into the index. 
For example, lets look at a document consists of performance related data points of an ec2 instance. 
```
{
	"hostName": "xyz",
	"hostIp": "x.x.x.x",
	"zone" : "us-east-1a"
	"@timestamp": 1693192062,
	"cpu": 80
	"jvm": 85
}
```
Here, hostName, hostIp and zone uses keyword as field type and timestamp/cpu/jvm uses numeric fields. Values for dimensions and mesurements are stored in docValue as well in stored fields as _source. The most common search query for such data set is aggregations like min/max/avg/sum. As we can see, data is stored twice here. 

Currently the _source field stores the original documents as stored field. We can possibly skip storing the _source field in such cases and retrieve the field values from docValue instead. This will help in reducing the storage cost significantly. Based on the nature of the query we can skip or fetch some or all of the fields from docValues to serve the search queries.

bugmakerrrrrr:
### Description
The [original derived source design](https://github.com/opensearch-project/OpenSearch/issues/9568#issuecomment-2525797878) did not include support for `nested` mappings. This PR extends derived source reconstruction to support nested fields.

### Related Issues
Resolves #9568

### Check List


bugmakerrrrrr:
@mgodwan Would you mind taking a look?

server/src/main/java/org/opensearch/index/mapper/ObjectMapper.java
```diff
@@ -1101,4 +1101,38 @@ public void deriveSource(XContentBuilder builder, LeafReader leafReader, int doc
         }
         builder.endObject();
     }
+
+    @Override
+    void deriveSource(XContentBuilder builder, DerivedSourceContext context) throws IOException {
+        if (this.nested.isNested()) {
+            List<Integer> nestedDocIds = context.nestedDocs(this);
+            if (nestedDocIds.isEmpty()) {
```
kkewwei, line 1109: nested field: comments

```
PUT source_derived_nested/_doc/1
{
   "comments": []
}
```
we will return nothing by the condition, I'm not sure whether is appropriate.

server/src/main/java/org/opensearch/index/mapper/ObjectMapper.java
```diff
@@ -1101,4 +1101,38 @@ public void deriveSource(XContentBuilder builder, LeafReader leafReader, int doc
...
+                return;
+            }
+            builder.startArray(simpleName());
```
kkewwei, line 1112: why not is `startObject`?

server/src/main/java/org/opensearch/index/mapper/DerivedSourceContext.java
```diff
@@ -0,0 +1,111 @@
...
+    private BitSet bitSetForRoot() throws IOException {
+        return bitSetForQuery(Queries.newNonNestedFilter());
+    }
+
+    private BitSet bitSetForNested(ObjectMapper nestedMapper) throws IOException {
+        return bitSetForQuery(nestedMapper.nestedTypeFilter());
+    }
+
+    private BitSet bitSetForQuery(Query query) throws IOException {
+        BitSetProducer producer = bitSetProducer.apply(query);
```
kkewwei, line 108: We can use `IndicesBitsetFilterCache.bitsetFromQuery(query, leafReaderContext)` directly, instead of passing it in from the outside.

VijayanB:
Hi @bugmakerrrrrr, thanks for this PR! Quick question — did you do any performance benchmarking comparing stored _source vs. derived source with nested documents? It would be helpful to see numbers (latency, throughput) to understand the cost of reconstructing nested objects, especially for documents with high-cardinality nested arrays.

Issue #21653, opened 2026-05-14
[Analytics Engine] Data node row ID emission and fetch-by-row-ids for Query-Then-Fetch late materialization


alchemist51:
## Description

  The analytics engine currently materializes all projected columns during the query phase scan. For queries with ORDER BY ... LIMIT N, this reads and transfers far more data than necessary — the sort only needs the sort key and a row identifier, not the full row payload.

  Query-Then-Fetch (QTF) late materialization splits execution into two phases:
  1. Query phase: Each shard returns only __row_id__ + sort/filter columns (lightweight)
  2. Fetch phase: Coordinator identifies survivors after sort+limit, sends their row IDs back to data nodes which read only those rows' full columns

  This issue covers the data node side — emitting shard-global row IDs during the query phase, and efficiently fetching specific rows by ID during the fetch phase.

## Proposed Solution

Query Phase: Row ID Emission

  Two execution paths emit shard-global __row_id__ values:

  ListingTable + Optimizer path — used when the query has parquet-only predicates with sort, or no filter with sort. Reads __row_id__ from parquet and computes the global ID via __row_id__ + row_base. This path exists because the indexed path doesn't support sort pushdown; when the data node must sort and there's no collector, it goes through ListingTable where DataFusion can plan sort naturally.

  Indexed path — used for everything else (any collector present, or no sort required). Computes row IDs from position: global_base + rg.first_row + position_in_rg. Once a collector is involved, the indexed path is required regardless of sort.

  Both paths produce the same output contract: Int64, non-negative, unique within a shard, globally accounts for multi-segment files.

  Both paths produce the same output contract: Int64, non-negative, unique within a shard, globally accounts for multi-segment files.

  Fetch Phase: Row Retrieval

  Given global row IDs from the coordinator, the data node:
  - Resolves IDs to per-file positions using segment boundaries
  - Builds targeted parquet access plans (row-group skipping + row-level selection)
  - Reads only the requested columns for the identified rows
  - Returns results with __row_id__ attached for coordinator-side positional assembly

  Reader Context

  The fetch phase must read from the same shard snapshot as the query phase. A reader context store holds acquired readers open between phases with configurable keep-alive and a reaper for cleanup.

  Out of scope (coordinator side — separate issue)

  - Planner changes to inject __row_id__ as a virtual column in the schema
  - New stage execution type for the two-phase QTF flow in the DAG
  - Fetch transport action, shard_id injection, position map, and result assembly

### Related component

[...]

PR #21836, opened 2026-05-26
Late Materialization (Query-Then-Fetch) for pure Sort queries

expani:
# Late Materialization (Query-Then-Fetch) for top-N queries

Implements QTF for `Sort + Limit` queries: shards return only the sort columns plus a row-id helper, the coordinator picks the top-K rows globally, then fetches the remaining columns from the originating shards by row id and stitches them back into sort order.

  ## Why

  For a query like "select a few columns from a large index, ordered by one column, limit K", the existing path materializes every projected column on every matching row cross every shard before the global sort. With QTF, shards only ship the sort columns and row ids, the coordinator picks K winners, and only those K rows drive a second per-shard fetch for the remaining columns. This drops shard-to-coordinator network I/O and avoids per-row materialization on rows that were never going to win.

  ## DAG shape

  Four stages (three if the cluster has a single shard, where the rewriter short-circuits):

  Stage 0: shard scan      (returns sort columns + row id, per-shard ordinal stamped at the boundary)
  Stage 1: coord reduce    (Sort + Limit picks K winners)
  Stage 2: late materialization (drain → group rows by shard → scatter fetch → stitch back into order)
  Stage 3: coord reduce    (post-LM derived projection / filter / aggregate)

  ---

  ## Coordinator

  ### Planner

  - **OpenSearchLateMaterializationRewriter** — new rule that detects a Sort+Limit over a viable scan, inserts the LM marker, and short-circuits in single-shard case.
  - **OpenSearchLateMaterialization** — new marker RelNode that records which columns are needed in the second-pass fetch.
  - **OpenSearchTableScan** — keeps its overridden row type through Calcite normalization so the helper columns survive.
  - **PlannerImpl** — adds Calcite's standard sort-project-transpose and project-merge rules so the planner output reaches the canonical shape QTF detection expects.
  - **RelNodeUtils** — small helpers used by the rewriter when threading helper columns through the row type.
  - **ArrowCalciteTypes** — new bidirectional Arrow ↔ Calcite type converter used when the LM stage builds its output schema.
  - **DAGBuilder** — adds the cut around the LM stage and adjusts the Stage 1 input-scan placeholder so the schema includes the helper columns.
  - **FragmentConversionDriver** — emits a schema-only Substrait Read for the LM stage (it has no source table to lower) and records the post-decoration schema for any stage where the wire schema differs from the producer's natural one.
  - **StagePlan** — carries the post-decoration schema so the consuming reduce sink registers against the schema that actually crosses the wire.
  - **InputSinkDecorator** — new SPI that lets a reduce stage wrap its input sink. Today the only implementation stamps the per-shard ordinal column.
  - **Stage**, **StageExecutionType** — new LM stage type with the wiring needed to provide an exchange sink and an instruction handler factory.

  ### Scheduler / Core

  - **LateMaterializationStageExecution** — new stage orchestrator that runs the four QTF phases. It drains the upstream reduce output to recover row ids and per-shard ordinals, partitions rows by shard, fans out a fetch transport call per shard, and lets the stitcher write each response back into the K-row output in sort order.
  - **LateMaterializationStageExecutionFactory** — builds the LM stage execution and resolves the descendant shard-stage that supplies the per-shard target lookup.
  - **Stitcher** — new per-stage helper that owns the K-row output and copies cells from each shard's response into the correct row position. Emits one batch when all shards complete.
  - **OrdinalAppendingSink** — new decorator that stamps a per-shard ordinal column onto each batch before it lands in the reduce sink. Used to identify which shard a winning row came from.
  - **VectorUtils.appendConstantInt** — new zero-copy helper for adding the constant ordinal column. Works around an Arrow 18.x change that rejects the obvious "append at end" path.
  - **QueryContext** — new side-table mapping a stage's resolved per-shard targets so a downstream stage can map an ordinal back to a node and shard.
  - **ReduceStageExecution** — consults the optional input-sink decorator; multi-input dispatch is now keyed by the logical child-stage count rather than a marker
  interface.
  - **ReduceStageExecutionFactory** — when registering child inputs, prefers the post-decoration schema over the producer's natural one.
  - **FetchByRowIdsAction**, **FetchByRowIdsRequest** — new transport action and request type for the second-pass fetch.

  ### DataFusion backend (coordinator-relevant pieces)

  - **DataFusionFragmentConvertor** — new method that builds a schema-only Substrait Plan directly (bypassing isthmus, since the LM stage has no source table to lower).
  - **DataFusionInstructionHandlerFactory** — factory methods now take a flag that requests row-id emission on the shard.
  - **DatafusionReduceSink** — adds a schema-mismatch check on incoming batches and tolerates Timestamp precision/timezone divergence so the schema declared on the coordinator round-trips through the data-node parquet reader.

  ---

  ## Data node

  ### Core

  - **AnalyticsSearchService** — new handler that, given a sorted list of row ids and a column projection, produces an Arrow stream of those exact rows in ascending row-id order.
  - **AnalyticsSearchTransportService** — registers the new fetch action with the streaming transport and exposes the coordinator-side dispatch helper.
  - **ReaderContext** / **ReaderContextStore** — keep-alive plumbing so the reader handle survives across the query → fetch round trip.

  ### DataFusion backend (Java)

  - **ShardScanInstructionHandler** — when row-id emission is requested, switches to the indexed execution path that can produce row ids.
  - **ShardScanWithDelegationHandler** — same gating for the delegation path.
  - **NativeBridge** — Java FFM signature for the indexed-session-creation call now carries the row-id flag.

  ### DataFusion backend (Rust)

https://github.com/opensearch-project/OpenSearch/issues/21653

  **Row-id emission infrastructure:**

  - **project_row_id_analyzer.rs**, **project_row_id_optimizer.rs**, **shard_table_provider.rs** — new modules that wire row-id emission into the vanilla (non-indexed) execution path. The analyzer detects when a query references the row-id helper column and the optimizer rewrites the physical plan so the table provider produces row ids alongside the data columns.
  - **indexed_table/row_id_injection.rs** — sibling injection point for the indexed execution path; row ids come from the position within the row group rather than from the table provider.
  - **api.rs** — plan inspection now reports both whether the query is indexed and whether it requests row ids, so dispatch routes to the right executor.
  - **datafusion_query_config.rs** — new enum that captures the strategy decision for the non-indexed-but-row-id-needed case.
  - **indexed_executor.rs** — reads the row-id-emission flag from session config.
  - New end-to-end and benchmark tests covering both emission paths and the strategy enum's branches.

  **Core-side opt-in:**

  - **ffm.rs** — the indexed-session FFM signature now carries a byte for the row-id-emission flag set on the Java side.
  - **session_context.rs** — stores the flag on the indexed-execution config so the executor can read it without re-parsing the plan.

  ---

  ## Cross-cutting (Framework SPI)

  - **ShardScanInstructionNode**, **ShardScanWithDelegationInstructionNode** — new field that carries the row-id-emission request, with serde.
  - **FragmentInstructionHandlerFactory** — factory methods accept the new flag.
  - **FragmentConvertor** — new default-throwing schema-only-read method that backends opt into.
  - **ExchangeSink** — new default method that lets sinks distinguish a per-source ordinal when one is available.
  - **FieldStorageInfo** — exposes per-column storage hints so the late-materialization stage can build its fetch projection (i.e., decide which columns to ask the data
  node for in the second-pass fetch).

  ---

  ## Tests

  - New end-to-end IT that asserts QTF fires and that the post-LM stage runs derived expressions (e.g., `UPPER(url)`) over the stitched output, with per-row value
  assertions.
  - New IT that verifies Substrait conversion at each stage.
  - New planner-shape goldens for the rewrite.
  - DAG-shape and fragment-conversion unit tests updated to cover the LM stage type and the post-decoration schema plumbing.

  ## Known limitations / follow-ups

  - **Single-batch emission from the stitcher** — the full stitched output is emitted in one feed when all shards complete, so the post-LM stage uses a buffered sink.
  Streaming the post-LM stage (per-shard sub-batch emission, dropping the position-sortedness invariant) is deferred until a real workload demands it.
  - **Type coverage in the Arrow ↔ Calcite converter** — Timestamp currently hardcodes millisecond precision; Date, Time, smallint, tinyint, decimal, and timestamp-with-local-timezone are unmapped. Broaden before exposing QTF beyond keyword and date columns.


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -300,4 +313,143 @@ private ShardScanExecutionContext buildContext(
...
+     *
+     * @param task transport-layer task for cancellation propagation; may be null in tests
+     */
+    public FragmentResources executeFetchByRowIds(FetchByRowIdsRequest request, IndexShard shard, AnalyticsShardTask task) {
+        if (task != null && task.isCancelled()) {
+            throw new TaskCancelledException("Fetch task cancelled before execution: " + task.getReasonCancelled());
+        }
+        long[] rowIds = request.getRowIds();
+        String[] columns = request.getColumns();
+        if (rowIds == null || rowIds.length == 0 || columns == null || columns.length == 0) {
```
mch2: are we short circuiting at coordinator here as well?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -300,4 +313,143 @@ private ShardScanExecutionContext buildContext(
...
+                    + shard.shardId()
+                    + " requires non-empty rowIds and columns; got rowIds="
+                    + (rowIds == null ? "null" : rowIds.length)
+                    + ", columns="
+                    + (columns == null ? "null" : columns.length)
+            );
+        }
+        boolean hasRowIdField = false;
+        for (String c : columns) {
+            if (DocumentInput.ROW_ID_FIELD.equals(c)) {
```
mch2: are we not making this guarantee already? is this not a precondition?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
```diff
@@ -118,6 +122,47 @@ public void onFailure(Exception e) {
...
+            false,
+            true,
+            AdmissionControlActionType.SEARCH,
+            FetchByRowIdsRequest::new,
+            (request, channel, task) -> {
+                IndexShard shard = indicesService.indexServiceSafe(request.getShardId().getIndex()).getShard(request.getShardId().id());
+                try (FragmentResources ctx = searchService.executeFetchByRowIds(request, shard, (AnalyticsShardTask) task)) {
+                    Iterator<EngineResultBatch> it = ctx.stream().iterator();
+                    // FIXME do we need to drain the entire stream here, or can we hand back
+                    // pressure to the engine when the channel/coordinator is slow?
```
mch2: can we not reuse what `FragmentExecutionAction` does here? it already implements this backpressure - unless we are limiting aggressively here and you expect this as a single batch..

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
```diff
@@ -151,6 +196,8 @@ public void handleStreamResponse(StreamTransportResponse<FragmentExecutionArrowR
                 try {
                     FragmentExecutionArrowResponse current;
                     FragmentExecutionArrowResponse last = null;
+                    // FIXME do we need to drain the entire stream here, or should we hand back
+                    // pressure when the listener / downstream consumer is slow?
```
mch2: the listener here should be blocking `listener.onStreamResponse(last, false);` per batch, so nextResponse would not be getting invoked until reduce per batch has processed.

`responseListenerFor > StreamingResponseListener > outputSink.feed(vsr);` is synchronous

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSettings.java
```diff
@@ -182,6 +182,47 @@ public final class DatafusionSettings {
...
+     *       Reads {@code __row_id__} from parquet and adds the file's cumulative row offset via
+     *       ProjectRowIdOptimizer ({@code __row_id__ + row_base = global_id}). Works with standard
+     *       DataFusion ListingTable scan path.</li>
+     *   <li>{@code indexed} — Uses the indexed pipeline (segment partitioning, PositionMap).
+     *       Computes row IDs from position ({@code global_base + rg.first_row + position_in_rg}).
+     *       Zero I/O for the row ID column. Fastest path when the indexed executor is available.</li>
+     * </ul>
+     * Default: {@code indexed}.
+     */
+    public static final Setting<String> INDEXED_QUERY_STRATEGY = Setting.simpleString(
```
mch2, line 207: Is this solely for testing?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/VectorUtils.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param input     source batch; not closed by this method
+     * @param name      name of the new column
+     * @param value     constant value written to every row of the new column
+     * @param allocator allocator for the new {@link IntVector}
+     * @return new VSR with the appended column; row count and existing column data unchanged
+     */
+    public static VectorSchemaRoot appendConstantInt(VectorSchemaRoot input, String name, int value, BufferAllocator allocator) {
+        int rowCount = input.getRowCount();
+        // FIXME [RemoveBeforeMainMerge] diagnostics for QTF VSR.addVector failure
+        LOGGER.info(
```
mch2: this is likely very noisy / per query, lets set to debug scope.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/Stitcher.java
```diff
@@ -0,0 +1,227 @@
...
+    /**
+     * Copies one shard's response batch into the output VSR. Reads positions starting at
+     * {@code rowsCopiedSoFar} (the listener's per-shard "rows copied so far" counter) and
+     * places each response row at {@code positions[rowsCopiedSoFar + srcRow]}.
+     *
+     * <p>The response batch's column layout is {@code [___row_id, fetch-col-0, fetch-col-1, ...]} —
+     * fetch cols ordered as the request's {@code columns[]} (excluding the helper). The
+     * helper column is identified by name and skipped during copy.
+     */
+    public synchronized void acceptBatch(VectorSchemaRoot batch, int[] positions, int rowsCopiedSoFar) {
```
expani: Will make these be not `synchronized` if not required, across this class.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/VectorUtils.java
```diff
@@ -0,0 +1,93 @@
...
+            "FIXME [RemoveBeforeMainMerge] appendConstantInt: name={} value={} rowCount={} fieldVectorsSize={} schema={}",
+            name,
+            value,
+            rowCount,
+            input.getFieldVectors().size(),
+            input.getSchema()
+        );
+
+        // TODO/FIXME: reserve `rowCount * 4` bytes against the query CircuitBreaker before allocateNew.
```
mch2: would this not hit our allocator's limit?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/VectorUtils.java
```diff
@@ -0,0 +1,93 @@
...
+        for (int i = 0; i < rowCount; i++) {
+            constantVector.set(i, value);
+        }
+        constantVector.setValueCount(rowCount);
+
+        // Zero-copy: existing FieldVectors are reused by reference; only the new column allocates.
+        List<FieldVector> combined = new ArrayList<>(input.getFieldVectors().size() + 1);
+        combined.addAll(input.getFieldVectors());
+        combined.add(constantVector);
+        return new VectorSchemaRoot(combined);
```
mch2, line 76: do we need to transfer/drop the input vsr?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/QueryContext.java
```diff
@@ -104,6 +123,29 @@ public List<AnalyticsOperationListener> operationListeners() {
         return operationListeners;
     }
 
+    /**
+     * Records the {@link ShardExecutionTarget}s resolved for a stage. Called once by the
+     * stage execution after {@code TargetResolver.resolve(...)} runs. See the field-level
+     * Javadoc on {@code resolvedTargetsByStage} for context on why this lives on
+     * {@code QueryContext}.
+     */
+    public void recordResolvedTargets(int stageId, List<ShardExecutionTarget> targets) {
```
mch2, line 132: Are we restricting QTF optimization on join / multi child stage queries? if so we need a concurrent collection on resolvedTargetsByStage

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DAGBuilder.java
```diff
@@ -101,6 +120,115 @@ private static RelNode sever(
...
+        // Schema is the wrapper's output rowType (= aboveAnchorPhysicalFields).
+        //
+        // TODO Stage 3 sink mode. The COORDINATOR_REDUCE that wraps the post-LM ops currently
+        // inherits whatever sink the cluster setting selects (streaming vs memtable). For QTF
+        // it should be memtable: LM emits a single VSR after full stitch (see Stitcher TODO),
+        // so streaming buys nothing and the eager-scheduling deadlock-avoidance the streaming
+        // sink is designed for doesn't apply. Once Stitcher supports incremental emission, we
+        // can pick streaming for Camp-A post-LM ops (Filter/Project/hash Aggregate) and keep
+        // memtable for Camp-B (Sort/TopN/global Aggregate). Detect post-LM op shape here and
+        // hand a per-stage hint to the sink-provider selection.
```
mch2, line 243: do we need to do this now ? is there downside to streaming here?

mch2 (commented):
thanks @expani still going through this but pasting initial batch

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/StagePlan.java
```diff
@@ -21,28 +21,36 @@
...
- * @param backendId             the primary backend for this plan
- * @param convertedBytes        backend-specific serialized plan bytes (null before conversion)
- * @param delegatedExpressions  serialized delegated expressions (empty if no delegation)
- * @param instructions          ordered instruction nodes for data-node execution (empty before resolution)
+ * @param resolvedFragment           fragment with all viableBackends narrowed to single choices
+ * @param backendId                  the primary backend for this plan
+ * @param convertedBytes             backend-specific serialized plan bytes (null before conversion)
+ * @param delegatedExpressions       serialized delegated expressions (empty if no delegation)
+ * @param instructions               ordered instruction nodes for data-node execution (empty before resolution)
+ * @param postDecorationSchemaBytes  schema-only Read taking precedence over {@code convertedBytes} when
```
mch2, line 29: lets be explicit this is only for QTF's query phase?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
expani: Yes if something from a Shard doesn't survive Coordinator won't even send it a Fetch task as it will never come up.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
expani: This is a guarantee. Might have stayed due to independent development.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSettings.java
expani: No it was added as a tuning for testing some performance variations. 
Currently, uses default all the time

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/VectorUtils.java
expani: Yeah it would, removed it

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
expani: Made it similar to FragmentStreaming

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/QueryContext.java
expani: Yes we are restricting QTF on Join/Union/SubQueries; etc. so this is thread safe

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/VectorUtils.java
expani: `DatafusionReduceSink.feedToSender` should close the outer one which frees up the inner as well.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DAGBuilder.java
expani: This is a TODO, we are not doing it now.

We will wait to stitch back the entire results after FETCH phase and then feed to parent stage.

This comment was done to capture that we can incrementally feed that which I will think can be a followup

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/StagePlan.java
expani: I didn't add QTF because something later that alters the plan can reuse it. 

Idea is you can alter the batches that is fed to a sink. The first use-case to do so is QTF.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/VectorUtils.java
expani: Yeah will remove once I am done adding more tests. 

Keeping this comment unresolved till then so I don't forget

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
expani: Done

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
expani: Removed it

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java
```diff
@@ -143,22 +142,33 @@ private static void registerFetchByRowIdsHandler(
...
+
+                        @Override
+                        public void onComplete() {
+                            channel.completeStream();
+                        }
+
+                        @Override
+                        public void onFailure(Exception e) {
+                            if (e instanceof StreamException se && se.getErrorCode() == StreamErrorCode.CANCELLED) {
+                                return;
```
alchemist51: Why are we returning here?

expani: I have just refactored this into a common method. The behavior is same as mainline https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchTransportService.java#L108


Having said that, I have refactored this to throw now. Good catch

mch2 (approved):
this is a huge feature thanks @expani and @alchemist51 for taking this on

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/VectorUtils.java
expani: Removed this

PR #16483, opened 2024-10-25
rollover alias supports restored searchable snapshot index

kkewwei:
### Description
Using ISM to rollover alias and restore remote index will use _aliases and _rollover api, we should exclude the searchable snapshot index from `checkBlock`.

### Related Issues
Resolves #16419

### Check List
- [x] Functionality includes testing.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21838, opened 2026-05-26
Rollover: scope checkBlock to write index only, skipping non-write alias members

### LINKED ISSUE #21629: [BUG] Rollover precheck overscans alias members; should check only the write index
### Describe the bug

Disclaimer: parts of this issue were generated by AI.

## Problem

`TransportRolloverAction.checkBlock` resolves the rollover alias to **all** of its concrete member indices, then aborts the rollover if **any** of them carries a `METADATA_WRITE` cluster block.

This looks like an overscan. Rollover only ever modifies the alias's write index. A `METADATA_WRITE` block on any *other* alias member has no bearing on whether the rollover can complete, but it still kills the request before the write-index code path runs.

### Concrete trigger

Cross-cluster-replication applies an `INDEX_REPLICATION_BLOCK` to follower indices at level `WRITE | METADATA_WRITE`. When a follower lands on a shared rollover alias (e.g., active-active topologies where the alias replicates with the data), the next rollover fails with:

```
index [<follower>] blocked by: [FORBIDDEN/1000/index read-only(cross-cluster-replication)];
```

…even though the follower is not the rollover target and the real write index is unblocked. The error names the irrelevant blocked index, not the would-be rollover target, making it especially confusing.

CCR is just the visible trigger; any plugin or operator that places a `METADATA_WRITE` block on a non-write alias member breaks rollover the same way.

### Related component

Indexing

### To Reproduce

Real-world topology that triggered this in production: bidirectional CCR between two datacenters (call them `dc-a` and `dc-b`), each ingesting locally, with a single shared write alias so the ingest agent's config is DC-agnostic.

1. Set up bidirectional CCR between cluster `dc-a` and cluster `dc-b`. Configure auto-follow on each side so indices created on the peer are replicated locally as followers.
2. On `dc-a`, create a write index `idx-dc-a-2026.04.03-000001` with the shared write alias `idx-active`. Symmetrically on `dc-b`, create `idx-dc-b-2026.05.07-000001` with the same alias.
3. Configure ISM on both clusters with a rollover policy targeting the `idx-active` alias (e.g., rollover at 50 GB or 1 day).
4. Start ingesting locally on each cluster via `idx-active` — writes route to the local write index because alias resolution is cluster-local.
5. Wait for CCR to replicate. The `dc-b`-origin index now exists on `dc-a` as a CCR follower carrying the `idx-active` alias (CCR replicates aliases blindly, only stripping `is_write_index`). Symmetrically on `dc-b`.
6. ISM on `dc-a` reaches the rollover threshold and calls `_rollover` against `idx-active`. The precheck scans every alias member, finds the `dc-b`-origin follower carries the CCR `INDEX_REPLICATION_BLOCK`, and aborts with the error above.

### Expected behavior

Rollover should have happened

## Prior art

This overscan has already been hit and patched narrowly once:

- [opensearch-project/OpenSearch#16419](https://github.com/opensearch-project/OpenSearch/issues/16419) — same precheck overscan, different trigger. Reporter set `index.blocks.read_only=true` on a non-write alias member that became a remote/searchable snapshot; next rollover failed with the same shape of error.
- [opensearch-project/OpenSearch#16483](https://github.com/opensearch-project/OpenSearch/pull/16483) (merged Nov 2024) — fix for #16419. Added a per-setting carve-out (`INDEX_DATA_READ_ONLY_BLOCK_SETTINGS` in `ClusterBlocks`) that excludes searchable-snapshot indices from the precheck. PR description: *"we should exclude the searchable snapshot index from `checkBlock`."*

The CCR-follower case in this issue is the same bug class with a different trigger. Adding another per-plugin carve-out would push the overscan problem to whichever plugin lands a `METADATA_WRITE` block next; scoping the precheck to the write index resolves the class.

mishail:
Disclaimer: parts of this PR were generated by LLM

### Description
The rollover precheck in `TransportRolloverAction.checkBlock` resolved the rollover target to every alias member and aborted on any `METADATA_WRITE` block on any of them. That means a non-write alias member carrying such a block — for example a cross-cluster-replication follower's index-replication block — kills otherwise-valid rollovers against a sibling write index. The same shape applies to data-stream rollovers when an older backing carries a metadata-write block.

This PR scopes the precheck to the single write index via `IndexAbstraction.getWriteIndex()` — the same path `MetadataRolloverService` uses to perform the rollover — so the precheck and the actual rollover agree on which index matters. The same code path handles alias and data-stream targets and works regardless of which plugin placed the block.

The block check itself still goes through `ClusterBlocks.indicesWithRemoteSnapshotBlockedException`(added by https://github.com/opensearch-project/OpenSearch/pull/16483) — only the input changes from "all alias members" to "the single write index". The helper's existing carve-out for `remote_snapshot` indices (skip the abort unless the user has set `index.blocks.read_only`, `index.blocks.metadata`, or `index.blocks.read_only_allow_delete`) therefore continues to apply, so searchable-snapshot indices keep working.

`ClusterBlocks.indicesWithRemoteSnapshotBlockedException` itself is unchanged; `TransportIndicesAliasesAction` still uses it for alias-modification blocking and is unaffected.

### Related Issues
Resolves #21629

### Check List


mishail:
@andrross it's been a month already. Are there anything wrong with these changes?

andrross (approved):

PR #21840, opened 2026-05-26
Add search support for handling deleted documents

RS146BIJAY:
### Description
For a Composite format where one of the format is Lucene (or lucene only format), deleted documents are tracked by Lucene's `liveDocs` bitmap. When a query's filter contains at least one predicate naturally served by Lucene, that predicate's `Weight.scorer(...).iterator()` already excludes deletes — so deleted docs are filtered out.

When **every** predicate is Parquet-native (DataFusion drives the scan, no Lucene delegation occurs), there is no Lucene-produced bitset to AND with the row-group survivor mask. Parquet has no concept of soft-delete, so deleted documents leak into query results.

**The fix**: inject a synthetic Lucene-bound `MATCHALL()` annotation under a top-level AND so at least one Lucene `Weight.scorer` runs per row group, producing a live-docs bitset that gets ANDed into the survivor mask.

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
@@ -16,7 +16,11 @@ import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.rex.RexUtil;
+import org.apache.calcite.sql.SqlFunction;
+import org.apache.calcite.sql.SqlFunctionCategory;
 import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.type.OperandTypes;
+import org.apache.calcite.sql.type.ReturnTypes;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.opensearch.analytics.planner.CapabilityRegistry;
@@ -35,6 +39,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import static org.apache.calcite.sql.type.SqlTypeName.BOOLEAN;
+
 /**
  * Converts {@link Filter} → {@link OpenSearchFilter}.
  *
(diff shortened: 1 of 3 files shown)


mch2:
do we have tests here?

RS146BIJAY:
> do we have tests here?

Added the test cases covering all scenarios.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -107,6 +146,66 @@ public void onMatch(RelOptRuleCall call) {
         );
     }
 
+    // ---- Predicate annotation ----
+    private RexNode injectLuceneLiveDocsClauseIfNeeded(
```
expani, line 150: This is meant only for Performance and Correctness based Filter delegation if Lucene is present as a backend. 

Given deletes are also meant to be supported using pure parquet with datafusion WITHOUT lucene as a backend, it seems like a temporary workaround. 

Why are we storing the deleted docs in LiveDocs in Lucene during indexing ?

RS146BIJAY: For the initial cut, we will only support updates/deletes when of the format is Lucene. For parquet only format indices will be [append only](https://docs.opensearch.org/latest/im-plugin/append-only-index/). This is still a temporary solution as having a separate bitsets/delete vector will require implementing more complex structure like DeleteQueue, doc id upto etc.

expani: Can we look at moving this to the Backend especially restricted to the Shard/DataNode layer of Analytics Core ? 

Deletion awareness needs to bought into the Planner if we require it for making a cost based decision at the datanode/Shard.

Even in such cases, it should be hidden behind the Backends.

>as having a separate bitsets/delete vector will require implementing more complex structure like DeleteQueue. doc id upto; etc

Then this is a relationship b/w 2 backends is arising for delegating LiveDocs which is independent of Filter Delegation for either performance or Correctness. 

I don't see a need to introduce this to the Coordinator side of things yet ( Planner/Execution/Scheduler )

PR #21841, opened 2026-05-26
[pull-based-ingestion] Pass IndexMetadata instead of only IngestionSource when initiating the shardConsumer

zhaih:
### Description
IngestionSource only exposes a very limited set of index settings and sometimes the plugin might want to see more, for example:
1. We only pass in the current shardId but don't know the total number of shards, this requires plugin must have some pre-existing knowledge about sharding (like, kafka is sharded the same way as OS index)
2. The plugin might want to have access to the index mapping to apply correct transformation or filtering, when the input schema and index schema is not exactly aligned.

### Check List

### DIFF
--- a/plugins/ingestion-kafka/src/test/java/org/opensearch/plugin/kafka/KafkaConsumerFactoryTests.java
+++ b/plugins/ingestion-kafka/src/test/java/org/opensearch/plugin/kafka/KafkaConsumerFactoryTests.java
@@ -8,13 +8,14 @@
 
 package org.opensearch.plugin.kafka;
 
+import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.test.OpenSearchTestCase;
 import org.junit.Assert;
 
 public class KafkaConsumerFactoryTests extends OpenSearchTestCase {
     public void testCreateShardConsumerWithNullSource() {
         KafkaConsumerFactory factory = new KafkaConsumerFactory();
-        expectThrows(NullPointerException.class, () -> factory.createShardConsumer("test-client", 0, null));
+        expectThrows(NullPointerException.class, () -> factory.createShardConsumer("test-client", 0, (IndexMetadata) null));
     }
 
     public void testParsePointerFromString() {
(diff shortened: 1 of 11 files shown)


plugins/ingestion-kafka/src/test/java/org/opensearch/plugin/kafka/KafkaConsumerFactoryTests.java
```diff
@@ -8,13 +8,14 @@
 
 package org.opensearch.plugin.kafka;
 
+import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.test.OpenSearchTestCase;
 import org.junit.Assert;
 
 public class KafkaConsumerFactoryTests extends OpenSearchTestCase {
     public void testCreateShardConsumerWithNullSource() {
```
varunbharadwaj: Should we rename the test from "nullSource" to "nullIndexMetadata"?

plugins/ingestion-kinesis/src/test/java/org/opensearch/plugin/kinesis/KinesisConsumerFactoryTests.java
```diff
@@ -19,7 +20,7 @@ public void testConstructor() {
 
     public void testCreateShardConsumerWithNullSource() {
```
varunbharadwaj: Same as above - test needs rename?

varunbharadwaj (approved):

server/src/main/java/org/opensearch/index/IngestionConsumerFactory.java
```diff
@@ -67,9 +68,28 @@ default T createShardConsumer(String clientId, int shardId) {
...
+     *
+     * @param clientId      the client id assigned to the consumer
+     * @param shardId       the id of the shard
+     * @param indexMetadata the index metadata containing the ingestion source configuration
+     * @return the created consumer
+     */
+    @SuppressWarnings("deprecation")
+    default T createShardConsumer(String clientId, int shardId, IndexMetadata indexMetadata) {
+        initialize(indexMetadata.getIngestionSource());
+        return createShardConsumer(clientId, shardId);
```
msfroh: I think it's safest to make this a one-line `return createShardConsumer(clientId, shardId, indexMetadata.getIngestionSource());`

That way, in case someone followed @varunbharadwaj's earlier deprecation notice and implemented `createShardConsumer(String clientId, int shardId, IngestionSource ingestionSource)`, their logic would still get called.

msfroh (approved):

PR #21844, opened 2026-05-27
[Analytics-Engine] MPP framework: supporting multi-stage distributed join and aggregation

LantaoJin:
### Description

HOW-JOIN-WORKS

---
HOW-AGG-WORKS

### Configurations

*Core:*

| Setting | Purpose | Default |
|---|---|---|
| `analytics.mpp.enabled` | master switch + incident kill switch | `false` |
| `analytics.mpp.distribute.min_rows` | size floor — a join/agg whose larger scan subtree is below this stays coordinator-centric | `1000000` |
| `analytics.mpp.shuffle.aggregate.enabled` | sub-toggle for distributed aggregation (distributed joins unaffected when off) | `true` |

*Strategy / cost tuning:*

| Setting | Purpose | Default |
|---|---|---|
| `analytics.mpp.broadcast.max_bytes` | cap on broadcast build size; a larger build is planned as (or falls back to) hash-shuffle | `64mb` |
| `analytics.mpp.broadcast.probe_estimate` | probe-node count the broadcast cost estimate uses; `-1` = cluster data-node count at planning time | `-1` |
| `analytics.mpp.shuffle.partitions` | fixed hash-shuffle partition count; `-1` = per-query default (probe-side data-node count) | `-1` |

*Memory / spill:*

| Setting | Purpose | Default |
|---|---|---|
| `analytics.mpp.shuffle.node_budget_percent` | per-node on-heap shuffle budget, as a percent of `-Xmx` (`0` disables the bound) | `80` |
| `analytics.mpp.shuffle.spill.enabled` | spill shuffle intermediates to disk instead of failing fast | `false` |
| `analytics.mpp.shuffle.spill.directory` | spill root (one subdir per query); `""` → `<path.data>/shuffle_spill` | `""` |
| `analytics.mpp.shuffle.spill.max_bytes` | disk ceiling for spill, per node | `50gb` |
| `analytics.coordinator.buffer_limit` | per-query coordinator allocator cap in bytes for the `FINAL` gather; `0` = no per-query cap (share the coordinator allocator) | `0` |

*Reliability:*

| Setting | Purpose | Default |
|---|---|---|
| `analytics.mpp.shuffle.recv_timeout` | per-partition receive timeout — backstop against a stuck shuffle producer | `60s` |

*Shuffle payload:*

| Setting | Purpose | Default |
|---|---|---|
| `analytics.mpp.shuffle.prune_columns` | drop columns no operator references before the shuffle, so only the join keys + downstream-referenced columns cross the wire (shrinks the shuffle 5–25× on wide fact tables) | `true` |
| `analytics.mpp.shuffle.compress` | compress shuffle IPC chunks (standard Arrow IPC compression) to shrink the on-heap buffered bytes — trades CPU for heap headroom | `false` |
| `analytics.mpp.compression.codec` | shuffle compression codec when enabled — `zstd` or `lz4` | `zstd` |
| `analytics.mpp.compression.zstd.level` | ZSTD level when the codec is `zstd` (matches Spark's shuffle default); range `1`–`22` | `1` |

### Benchmark
**SF=1:**
11-sf1-three-way

**SF=10:**
sf10-base-vs-feature

### Related Issues

### Check List


Bukhtawar (commented):
Should we evaluate datafusion distrib before committing on the existing distributed framework which I am not confident it can scale to high number of shards/concurrent requests espl the Rust -> Java(blocking stream_next) -> Rust switch across layers making pipelining inherently harder. @mch2 thoughts?

LantaoJin:
> Should we evaluate datafusion distrib before committing on the existing distributed framework which I am not confident it can scale to high number of shards/concurrent requests espl the Rust -> Java(blocking stream_next) -> Rust switch across layers making pipelining inherently harder. @mch2 thoughts?

Thanks @Bukhtawar. The "datafusion distrib", do you mean Ballista or datafusion-distributed? 
Ballista is a standalone engine, its own scheduler + executor processes, deployed separately (Docker/K8s), disk-based shuffle, and it assumes it owns data partitioning. It has no notion of a Lucene shard pinned to a node by OpenSearch's allocation/routing, and no integration with cluster state, transport, circuit breakers, cancellation, or security. Adopting it means running a second cluster and shipping shard data to it -- losing shard-locality/pushdown and doubling operational surface. It also doesn't remove the Rust↔Java boundary: our coordinator/planner/transport are the JVM, so an embedded engine still meets the JVM at the coordinator -- Ballista just relocates that boundary into a cross-cluster hop.

datafusion-distributed's data path is all Rust↔Rust (Flight/gRPC), so the JVM hop disappears only if coordination moves to Rust, which means giving up the Calcite planner, PPL/REST, cluster state, transport, security, breakers, and the MPP scheduling already built, and running a second gRPC mesh outside OpenSearch transport/security. Keep the JVM coordinating and the same boundary just reappears at the coordinator over FFM. However, I am open to the idea. We could design a concrete plan addressing the details mentioned, conduct ongoing evaluations, and run a PoC. I am not sure if @mch2 has evaluated the `df-distributed` library before.

Bukhtawar:
We just need to avoid Java hop in the response path data node to coordinator path if we even want to start small. Yes I mean Datafusion distributed not Ballista. I am hoping Datafusion distributed takes in a Substrait plan, so no API/Planner changes. Some transport actions and security might have to be re-wired as a result of Rust based distributed executor but I feel it would be totally worth the simplification/scalability/performance benefits it would offer

expani:
>Rust -> Java(blocking stream_next) -> Rust switch across layers making pipelining inherently harder.

I agree that we should keep this natively in Rust which means that either `arrow-flight-rpc` supports a Pure Rust mode OR we go around it to make pipelining in Datafusion. 

If we can achieve that, curious to see what other issues exist that would force us to move to datafusion-distributed.

LantaoJin:
Thanks @expani @Bukhtawar , could we consider moving more of the distributed execution logic to the Rust side (df-distributed solution) as a follow-up task and making it opt-in? The current PR covers too much ground; this specific improvement doesn't require a complete rewrite, and I've already had an improvement plan generated by AI which is based on `df-distributed`.

LantaoJin:
Can this PR be merged first? @mch2. I've started a PoC on **a Rust-native Flight exchange under Java control** which adopt df-distributed's `NetworkShuffleExec`/`worker_connection_pool` streaming model as a Rust component inside our `.so`: the JVM still plans (Calcite) and assigns worker URLs + shard targets, but hands each worker a serialized DataFusion sub-plan + peer URLs, and the shuffle edge becomes Flight-over-tonic **Rust↔Rust** with the x MiB backpressure budget. The JVM boundary shrinks to plan handoff + result gather at the coordinator, exactly the "start small, avoid the Java hop in the data path".

Bukhtawar:
> Thanks @expani @Bukhtawar , could we consider moving more of the distributed execution logic to the Rust side (df-distributed solution) as a follow-up task and making it opt-in? The current PR covers too much ground; this specific improvement doesn't require a complete rewrite, and I've already had an improvement plan generated by AI which is based on `df-distributed`.

Yeah we shouldn't block this PR, will take a look

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DataTransferCapability.java
```diff
@@ -0,0 +1,37 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.analytics.spi;
+
+/**
+ * Declares a backend's ability to produce or consume a shuffle-partitioned data transfer.
+ *
+ * <p>For hash-shuffle join the planner must intersect producer-side capabilities with consumer-side
```
expani, line 14: This can be a sub-field under existing `JoinCapability` as algoType capturing Hash Shuffle, Broadcast and other types of Join

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/join/GeneralShuffleDAGRewriter.java
```diff
@@ -0,0 +1,390 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.analytics.exec.join;
```
expani: I think these belong in the `org.analytics.planner` as custom Rewrite passes OR we can introduce as part of custom trait derivation extension for Joins.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchBroadcastScan.java
```diff
@@ -0,0 +1,156 @@
...
+ * satisfy this scan at execution time.
+ *
+ * <p>The build stage id determines the name used on the wire and in the Substrait plan —
+ * {@code "broadcast-&lt;buildStageId&gt;"}. Both the coordinator (when constructing the
+ * {@link org.opensearch.analytics.spi.BroadcastInjectionInstructionNode}) and the data node
+ * (when registering the memtable) agree on the string.
+ *
+ * @opensearch.internal
+ */
+public class OpenSearchBroadcastScan extends AbstractRelNode implements OpenSearchRelNode {
```
expani, line 42: This can be an optional field in the OpenSearchJoin operator with sub-classes for individual algorithms related metadata like build-stage scan.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchBroadcastExchange.java
```diff
@@ -0,0 +1,116 @@
...
+ * roughly {@code inputRowCount × probeNodeEstimate} — this is what discriminates BROADCAST from
+ * HASH_SHUFFLE during CBO: a small build side broadcasts cheaply (small × N), but a large build
+ * gets expensive (rows × N grows fast), so HASH wins when both sides are large.
+ *
+ * <p>{@code probeNodeEstimate} is resolved at split-rule time from the cluster setting
+ * {@code analytics.mpp.broadcast.probe_estimate}, defaulting to the cluster's data-node count.
+ *
+ * @opensearch.internal
+ */
+public class OpenSearchBroadcastExchange extends SingleRel implements OpenSearchRelNode {
```
expani, line 48: With trait propagation, we wouldn't need a new Relational Operator to store the cost given we can define it as a custom Trait of the OpenSearchJoin operator.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/join/MppStrategy.java
```diff
@@ -0,0 +1,104 @@
...
+    HASH_SHUFFLE,
+
+    /** Hash-shuffle aggregate: a distributed aggregate split into PARTIAL (per-partition, on the worker)
+     *  + FINAL (coordinator gather). */
+    HASH_SHUFFLE_AGG;
+
+    // ── Read-only DAG-shape predicates (used by DefaultPlanExecutor to pick the dispatch counter) ──
+
+    /** True iff any stage fragment in {@code dag} contains an {@link OpenSearchJoin}. */
+    public static boolean containsJoin(QueryDAG dag) {
```
expani, line 49: We can use the `RelNodeUtils#findNode` or enhance it if something is amiss.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/shuffle/ShuffleCompression.java
```diff
@@ -0,0 +1,122 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.be.datafusion;
+
+import org.apache.arrow.compression.CommonsCompressionFactory;
```
expani, line 11: Based on the imports, this can be a part of execution layer in Analytics Core.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/CalciteToArrowSchema.java
```diff
@@ -0,0 +1,109 @@
...
+import org.apache.arrow.vector.types.pojo.Schema;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Minimal Calcite {@link RelDataType} → Arrow {@link Schema} converter. Used by the M1
```
expani, line 26: I feel this should be in Analytics Core because Calcite and Arrow are first class citizens.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/shuffle/FastLz4CompressionCodec.java
```diff
@@ -0,0 +1,113 @@
...
+ * it. We only swap the IMPLEMENTATION, not the format.
+ *
+ * <p>Extends {@link AbstractCompressionCodec} so the shared length-prefix framing and the
+ * "incompressible buffer is stored uncompressed" fallback are inherited; this class implements only
+ * the raw byte transform, mirroring Arrow's {@code Lz4CompressionCodec.doCompress/doDecompress}
+ * shape ({@code getBytes → stream copy → setBytes}).
+ *
+ * @opensearch.internal
+ */
+final class FastLz4CompressionCodec extends AbstractCompressionCodec {
```
expani, line 48: These can be migrated up either to `analytics-engine` OR `arrow-base` if it's more generic

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeCall.java
```diff
@@ -219,6 +219,21 @@ public MemorySegment longs(long[] data) {
...
     }
 
+    /**
+     * Allocate a segment from an int array. Returns an empty (zero-byte) segment if the array
+     * is empty so callers can pass it as a non-null pointer with count zero.
+     */
+    public MemorySegment ints(int[] data) {
+        ensureOpen();
+        if (data == null) {
+            throw new NullPointerException("Cannot marshal null int array to native");
```
expani, line 229: nit : We can throw IllegalState instead

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchJoin.java
expani, line 124: I am trying to replace all derivations out of computeSelfCost with Calcite's PhysicalNode used by Volcano. I think deriving join algorithm (Row count based) and order should naturally get implemented as custom Trait implementations of a RelationalOperator

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchAggregate.java
```diff
@@ -307,9 +324,96 @@ public RelOptCost computeSelfCost(RelOptPlanner planner, RelMetadataQuery mq) {
                 return planner.getCostFactory().makeInfiniteCost();
             }
         }
+        // FINAL placement + parallel-merge cost (our hash-shuffle feature): FINAL is legal only over
```
expani, line 335: All the implementations of `computeSelfCost` will be replaced with `PhysicalNode.deriveTraits` and `PhysicalNode.passThroughTraits` to properly utilize Volcano CBO for doing Trait Propagation.

expani (commented):
Thanks for starting this @LantaoJin 

I think we can do this concisely once TraitPropagation is implemented using PhysicalNode for Calcite Volcano Planner.

Agreed that using Rust for streaming is a separate thing on its own.

mch2:
> Can this PR be merged first? @mch2. I've started a PoC on **a Rust-native Flight exchange under Java control** which adopt df-distributed's `NetworkShuffleExec`/`worker_connection_pool` streaming model as a Rust component inside our `.so`: the JVM still plans (Calcite) and assigns worker URLs + shard targets, but hands each worker a serialized DataFusion sub-plan + peer URLs, and the shuffle edge becomes Flight-over-tonic **Rust↔Rust** with the x MiB backpressure budget. The JVM boundary shrinks to plan handoff + result gather at the coordinator, exactly the "start small, avoid the Java hop in the data path".

---
Thanks @LantaoJin for getting this going.  This was a significant effort and I definitely don't want to hold this back from sandbox as is, so I think we should move forward with this.  Though, if possible, can we cut this into smaller logical chunks for review?  I know @expani had some planned cleanup to exchange insertion that will overlap here.

Though I agree with others here on moving more rust side. It gets us to a more natural pull-based pipelined execution model.  I've been prototyping a similar implementation ([distributed-engine-mpp](https://github.com/opensearch-project/OpenSearch/compare/main...mch2:OpenSearch:distributed-engine-mpp?expand=1)) that uses datafusion-distributed directly, so it does the stage cutting and task scheduling.  The coordinator still does a initial planning for marking / delegation and hands off the whole query as Substrait with worker URLs representing the DiscoveryNodes. Workers run as long-lived gRPC servers on each data node and at execution upcalls into Java only for reader acquisition and delegation setup. A custom TaskEstimator/WorkerResolver exists so each leaf task lands on the node hosting that shard.

Either way I don't want to block your PR, let's get it merged behind the feature flag and iterate from there?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/DataTransferCapability.java
LantaoJin: `DataTransferCapability(Kind{PRODUCER,CONSUMER}, format)` is about shuffle transport direction/format, orthogonal to `JoinCapability(kinds, formats)` which is SQL join-type support. They're used by different resolvers (`CapabilityResolutionUtils` / `ShuffleEnrichment` vs `OpenSearchJoinRule`). Merging conflates transport with relational semantics.

But, your point — "expose join algorithm as a capability" is somewhat reasonable and currently the algorithm is implied by which exchange CBO picks, not declared, could be as a follow-up?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/join/GeneralShuffleDAGRewriter.java
LantaoJin: Okey, trying to move it to `org.opensearch.analytics.planner.dag`, but keep `WorkerLevel` stays in `exec.join.ShuffleEnrichment`.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchBroadcastScan.java
LantaoJin: `OpenSearchBroadcastScan` is a standalone `AbstractRelNode` with `getBuildStageId()`. Consolidating is feasible but load-bearing -- `UnifiedDispatch` locates the broadcast build via `getBuildStageId()`.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/join/MppStrategy.java
LantaoJin: `MppStrategy#containsJoinRecursive` now calls `RelNodeUtils#findNode`.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/shuffle/ShuffleCompression.java
LantaoJin: Okey, trying to move `ShuffleCompression` + `FastLz4CompressionCodec` to analytics-engine and add arrow-compression + commons-compress + lz4-java deps (plus Unsafe forbidden-API whitelist + jar bundling) to analytics-engine/build.gradle.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/CalciteToArrowSchema.java
LantaoJin: sure

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/shuffle/FastLz4CompressionCodec.java
LantaoJin: sure

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeCall.java
LantaoJin: :229 one of 5 identical null-arg guards (:83, :113, :119, :202, :214, :229). For a null argument, `NullPointerException` (or Objects.requireNonNull) is arguably more correct than `IllegalStateException`(which is for bad object state, already used at :55/:149/:152 for the genuinely stateful cases). If changing, `IllegalArgumentException` fits better than IllegalStateException.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchJoin.java
LantaoJin: Agreed on the direction -- and part of it is already in this PR, just not wired through `PhysicalNode`. Two things are bundled in `computeSelfCost` and we've separated them:

1. **Exchange placement is no longer cost-driven**. It's a per-operator distribution algebra: see `DistributionAware` (this class implements it — `OpenSearchJoin ... implements ... DistributionAware`). Each operator declares `requiredInputDistribution` ("what partitioning must input i deliver") and `deriveOutputDistribution` ("given my inputs' distributions, what do I output"). A post-CBO pass places an exchange on an input only when the input's actual distribution doesn't `satisfy()` the requirement — so a co-partitioned child needs no exchange, and multi-tier cascades emerge for any shape. That's exactly `passThroughTraits`/`deriveTraits`, expressed as plain methods instead of `PhysicalNode` hooks. So the content you're asking for exists and is unit-tested per operator (`DistributionAwareTests`).

2. **Join-algorithm selection (broadcast/shuffle/coord) is still cost-seeded** via the split rules + `computeSelfCost`, because Volcano runs bottom-up here and gathers each join to `COORDINATOR+SINGLETON`.

We deliberately kept the algebra as plain methods rather than flipping Volcano to `setTopDownOpt` + `PhysicalNode.passThrough` — that flip is a large refactor (it also rewrites the marker/split-rule machinery, see below) and isn't needed for correctness, since the placement algebra already generalizes. I'd treat "join algorithm + order as a derived trait via PhysicalNode" as a tracked follow-up, not a blocker on this PR. I can drop a `// TODO` at these `computeSelfCost` sites pointing at `DistributionAware` so the migration path is discoverable.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchAggregate.java
LantaoJin: Same as the join comment — agreed as the target, partly here already. For aggregation, the `PARTIAL/FINAL` split placement is already trait-driven, not cost-driven: this class implements `DistributionAware`, declaring "FINAL requires `SINGLETON` input, `PARTIAL` derives worker-partitioned output," and the post-CBO pass inserts the reducer between them. `computeSelfCost` only ranks the alternatives the split rules emit.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchBroadcastExchange.java
LantaoJin: Partly agree, with one structural caveat. You're right that the cost of a broadcast wouldn't need its own `RelNode` under trait propagation — it'd fall out of `deriveOutputDistribution` on `OpenSearchJoin`. But this exchange node isn't only a cost holder: `DAGBuilder` cuts a child Stage at each exchange, so the node is the physical stage boundary that becomes a `BROADCAST_BUILD` fragment with its own transport (Arrow Flight), distinct from a shuffle edge. A trait on the join alone doesn't give the executor a stage to schedule — we'd still need a materialized boundary for the DAG builder to cut on.

 So the end state I'd expect: keep a lightweight exchange boundary node, but source its cost/selection from the join's derived distribution trait rather than a bespoke `computeSelfCost`. I don't think we can delete the `RelNode` outright while `DAGBuilder` cuts stages on it.

LantaoJin:
Thanks @mch2, really appreciate the thorough look and the support to move this forward behind the flag.

On cutting it into smaller chunks, I want to be straight about the trade-off rather than just say yes or no:

Why I'd prefer not to retroactively split this PR:
  - The layers are tightly interdependent — the SPI (`JoinCapability`/`DataTransferCapability`/instruction nodes), the `OpenSearch*` RelNodes + Volcano split rules, the DAG builder + post-CBO `DistributionEnforcementPass` + scheduler, the DataFusion backend handlers, and the native Rust all interlock. I can't cut a chunk that compiles green and passes the ITs without stubbing the layers around it, so a split series would have intermediate PRs that don't build — harder to review, not easier.
  - The blast-radius argument that usually motivates splitting is mostly absent here: it's all sandbox, all behind a default-off flag.
  - Of the ~955 files, the actual reviewable feature surface is much smaller — ~324 non-test sandbox files, and a big share of the rest is upstream-sync churn (jackson/netty license + .sha1 bumps ≈ 255 files, plus ~442 non-sandbox files from the periodic **main** re-sync). So the diff reads larger than the logic being reviewed.

FYI, the reviewer's map:

  - SPI / capabilities (libs/analytics-framework, ~22 files)
  - Planner RelNodes + Volcano rules (planner/rel, planner/rules, ~17)
  - DAG + post-CBO distribution/scheduler (planner/dag, exec/**, ~52)
  - DataFusion backend handlers (analytics-backend-datafusion Java, ~19)
  - Native Rust shuffle/Flight (rust/**, the largest chunk)
  - QA multi-node ITs (qa/analytics-engine-rest)

LantaoJin:
Here are the findings of switching the Rust→Java→Rust to pure Rust→Rust based on this PR + **datafusion-distributed**:
1. TPC-H Q11, Q21 regression which caused by df-distributed does not support **materialized shuffle** (map writes files, reduce fetches later).
2. No significant performance gains in the PoC on TPC-H SF=10
3. To support materialized shuffle in df-distributed could be a follow-up (giving the native Flight buffer real spill)
4. Switch to datafusion-distributed based on this PR has less code changes.
cc @mch2 @Bukhtawar

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/OpenSearchJoin.java
expani: Yes I am working on Trait Propagation as it is essential to avoid doing cost hacks for exchange placements ( happy to collaborate )
My concern is that introducing Join related CBO will make the flip more larger and refactoring will be a lot.

Dropped a general comment on enabling MPP for High Cardinality aggregation queries and then following up with Joins

expani:
We can add the support for Shuffles in this PR and apply them to high-cardinality GroupBy queries reducing the amount of changes done to introduce Joins.

Once the trait propagation is in, we can introduce Joins on top of it. This way we don't have to throwaway a lot of code required to make Calcite's Volcano behave as expected. Let me know your thoughts on the same.

LantaoJin:
> We can add the support for Shuffles in this PR and apply them to high-cardinality GroupBy queries reducing the amount of changes done to introduce Joins.

@expani , supporting shuffle in high-cardinality GroupBy queries has already supported, just enable it by `analytics.mpp.shuffle.aggregate.enabled=true`, or do you mean support **materialized** shuffle for datafusion-distributed in this PR?

LantaoJin:
~~@expani This PR was opened since 2 months ago, personally, I don't want to make it larger. The work (datafusion-distributed + support materialized shuffle) includes not only the code refactor, but also entire re-benchmark and documentations. If you insist, I will continue with this PR.~~

Ah, I misread, you mean scope this PR to agg-shuffle and defer joins until trait propagation lands, not materialized shuffle. Thanks for clarifying, and yes to collaborating on the flip.
On the concern: the join CBO surface that would actually need rework is narrower than it looks. Exchange **placement** is already out of `computeSelfCost`, it's the `DistributionAware` algebra (`requiredInputDistribution`/`deriveOutputDistribution`) consumed by a post-CBO pass, which is the same `deriveTraits`/`passThroughTraits` logic as plain methods. That should port to `PhysicalNode` largely as-is rather than being thrown away. What genuinely is bottom-up scaffolding is `OpenSearchDistributionDeriveRule` + the split rules, and those are the same machinery agg-shuffle needs (`OpenSearchAggregateSplitRule` etc.), so deferring joins wouldn't avoid that rewrite.

Given @mch2's "let's get it merged behind the feature flag and iterate from there", both agg-shuffle and joins are already default-off, could we land this as the flagged scaffold, then do the `PhysicalNode` flip as a joint follow-up? Happy to pair on it: I can port `DistributionAware` to the `PhysicalNode` hooks and delete the derive-rule scaffolding as part of your flip.

mch2 (approved):
Thanks @LantaoJin for this massive contribution.

I am giving my approval here, I think we should go ahead with this and iterate but I don't want to slow @expani down with his work on trait propagation by merging just yet.  @expani ?

Also there are some interfaces here that are not used leaving this largely independent and risk-free. We can refactor to tighten that up particularly in scheduling.

LantaoJin:
Thanks @mch2 , let me sync with main to resolve the latest conflicts (imports) and trigger the CI again.

LantaoJin:
Identified the current failures are not related. @mch2

LantaoJin:
failures not related

Bukhtawar (commented):
Curious if we have some early benchmarks to review

LantaoJin:
> Curious if we have some early benchmarks to review

Thanks @Bukhtawar , first, the current PR serves primarily as a scaffold, and there is known room for optimization; many of these optimizations can be developed in parallel based on this PR, and Marc also advised me to not make this PR larger. Second, MPP is primarily designed to accelerate queries involving shuffle operators, specifically JOIN operations. Since our earlier benchmarks (such as Big5 and ClickBench) did not include JOIN queries, I focused on TPC-H which involves numerous complex JOINs (note that TPC-H is not a new benchmark; it has long existed in the SQL repository).

LantaoJin:
@Bukhtawar I ran FULL ClickBench — 99,997,497 docs, 6 shards, 4-node cluster , here is the report and at least no regression.
```
  ┌───────────────────┬───────────┬────────────────────┐
  │                   │  mpp=off  │       mpp=on       │
  ├───────────────────┼───────────┼────────────────────┤
  │ Total (41 ops)    │ 55,204 ms │ 55,186 ms          │
  ├───────────────────┼───────────┼────────────────────┤
  │ Ratio             │ —         │ 1.00x              │
  ├───────────────────┼───────────┼────────────────────┤
  │ Errors            │ 0         │ 0                  │
  └───────────────────┴───────────┴────────────────────┘
```
TPC-H is the workload that exercises the join paths — ClickBench structurally can't.

PR #21848, opened 2026-05-27
analytics: Update ClickBench PPL test harness + planner fix

mch2:
Adds more expected output assertions for clickbench queries.  The remaining skipped are flaky between 1 and 2 shards due to topk sort (tie breaker is not deterministic) and dc() is approximate count distinct.

Tests
-----
- PplClickBenchIT runs all 43 ClickBench PPL queries randomizing shards (1 or 2) and replicas (0 or 1) per class run so we exercise both single-shard and multi-shard planner paths.
- DatasetProvisioner gains shards/replicas overrides and stable-_id bulk ingest so doc-to-shard routing is reproducible.
- ResponseValidator supports a per-query "tolerance" field for approximate aggregates (dc(), percentile_approx).
- 30 of 43 expected-response baselines pinned (the other 13 hit a known per-shard top-K truncation issue in "sort | head N" — they still exercise parse/plan/execute end-to-end via PASS_ON_MISSING).
- Data: 1002 docs in bulk.json with stable _ids; mapping unchanged.
- 13 .ppl files lightly edited so the smaller sample dataset returns ≥1 row (lowered thresholds / dropped offsets / added tiebreaker sorts).

Fix
---
- OpenSearchAggregateSplitRule: align FINAL aggCall types with the operator-inferred return types and cast-wrap to the original row type. Avoids a "BIGINT vs BIGINT NOT NULL" Volcano transformTo rejection that surfaced only on the real opensearch-sql PPL path (the test-ppl shim used stricter aggCall types and didn't hit it).

### Description

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -180,21 +181,75 @@ public void onMatch(RelOptRuleCall call) {
...
+            result = RelOptUtil.createCastRel(finalAggregate, aggregate.getRowType(), /*rename=*/true);
+        }
+
         call.getPlanner().ensureRegistered(singleOnSingleton, aggregate);
-        call.transformTo(finalAggregate);
+        call.transformTo(result);
+    }
+
+    /**
+     * Re-derive each aggCall's declared return type via the operator's
```
expani, line 254: Then we should align the types before the split itself ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -180,21 +181,75 @@ public void onMatch(RelOptRuleCall call) {
...
+                call.isApproximate(),
+                call.ignoreNulls(),
+                call.rexList,
+                call.getArgList(),
+                call.filterArg,
+                call.distinctKeys,
+                call.collation,
+                hasEmptyGroup,
+                input,
+                null,  // type=null → Calcite derives via op.inferReturnType(callBinding)
```
vinaykpud, line 283: if we pass `aggregate.getAggCallList()` directly, the FINAL OpenSearchAggregate's constructor errors out with a stored-type-vs-inferred-type mismatch.

So basically fix here is: 
We reconstruct each AggregateCall with type=null so that Calcite re-derives the return type against `gathered`, giving the new call with stored type that agrees with what the FINAL constructor's validation will infer.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
vinaykpud: how can we align before the split? we need to know the type which `gathered` produced for this right?

I am curious why this issue happens only for the queries with `count() sum() with groupby`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
vinaykpud: Fix works but for me seems like we are doing some work around  to make Calcite's internal consistency check pass

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
mch2: yeah we need the gathered (partial) relNode to align the types.

expani: The gathered RelNode is using RelOptUtil.convert which is just aligning the Traits. 

I am not able to understand how is that becoming a problem for the Return Type of the aggregate ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
vinaykpud: I checked why this happens only for certain queries like bellow:
```
source = indexname | stats count(), sum(AdvEngineID) by RegionID
```
ie when it splits into PARTIAL and FINAL

Original RelNode:
```  
OpenSearchAggregate(group={0}, count()=COUNT(), sum(AdvEngineID)=SUM($1), mode=SINGLE)
     └─ OpenSearchProject(RegionID=$65, AdvEngineID=$0)       
          └─ OpenSearchTableScan(clickbench)
````

So `OpenSearchProject(RegionID=$65, AdvEngineID=$0)`  is the original input for the aggregation before the split.

inputs are:
```
slot 0:  RegionID       (from underlying TableScan column $65)
slot 1:  AdvEngineID    (from underlying TableScan column $0)
```

So the original input's row type that the SINGLE aggregate sits over is exactly [RegionID, AdvEngineID] — 2 columns, indexes 0 and 1.

  The aggregate calls

```
- count() — argList=[]. Doesn't reference any input column.
- sum(AdvEngineID)=SUM($1) — argList=[1]. 

group={0} means group by input column 0 = RegionID.
 ```


When we have split,

```
  PARTIAL = OpenSearchAggregate(
      input  = OpenSearchProject(RegionID=$65, AdvEngineID=$0)   ← ORIGINAL input
      group  = {0}
      calls  = [COUNT(),  SUM($1)]                                ← argList=[1] points at AdvEngineID
      mode   = PARTIAL
  )
```

Partials output:

```
  slot 0:  RegionID                                  (group key, NOT NULL)
  slot 1:  count_partial          : BIGINT NOT NULL  (COUNT's PARTIAL output)
  slot 2:  sum(AdvEngineID)_partial : BIGINT (nullable)  (SUM's PARTIAL output)
```

then we get  `gathered = convert(partial, finalTraits)`

When we construct the final:

```
FINAL = OpenSearchAggregate(
      input  = gathered                                          ← DIFFERENT input ie partials output
      group  = {0}
      calls  = [COUNT(), SUM($1)]                       ← argList=[1](this is the issue), should be SUM($2)
      mode   = FINAL
  )
```

But now SUM's argList=[1] indexes into gathered's row type ie `BIGINT NOT NULL` instead of original `BIGINT`, basically they shifted since we changed the input.

vinaykpud: Can you try adding tests for single shard and multi shard for the query like `Count Sum By Key` kind of shape? I was able to reproduce the issue in both scenario

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
vinaykpud: I added the issue steps here https://github.com/opensearch-project/OpenSearch/pull/21848/changes#r3313371145 can you check

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -180,21 +181,75 @@ public void onMatch(RelOptRuleCall call) {
...
+                call.getName()
+            );
+            if (derived.getType().equals(call.getType())) continue;
```
vinaykpud, line 248: Phase 1 (logical) 
```
LogicalSystemLimit(fetch=[10000], type=[QUERY_SIZE_LIMIT])
  LogicalProject(count()=[$1], sum(AdvEngineID)=[$2], RegionID=[$0])
    LogicalAggregate(group=[{0}], count()=[COUNT()], sum(AdvEngineID)=[SUM($1)])
      LogicalProject(RegionID=[$65], AdvEngineID=[$0])
        LogicalTableScan(table=[[opensearch, clickbench]])
```
Phase 2 (after marking) :
```
OpenSearchProject(count()=[$1], sum(AdvEngineID)=[$2], RegionID=[$0], viableBackends=[[datafusion]])
  OpenSearchAggregate(group=[{0}], count()=[COUNT()], sum(AdvEngineID)=[SUM($1)], mode=[SINGLE], viableBackends=[[datafusion]])
    OpenSearchProject(RegionID=[$65], AdvEngineID=[$0], viableBackends=[[datafusion]])
      OpenSearchTableScan(table=[[clickbench]], viableBackends=[[datafusion]])
```

```
[2026-05-27T20:59:57,316][ERROR][o.o.s.p.r.RestPPLQueryAction] [runTask-0] Error happened during query handling (status INTERNAL_SERVER_ERROR)
java.lang.IllegalStateException: Analytics-engine executor rejected the plan: type mismatch:
aggCall type:
BIGINT
inferred type:
BIGINT NOT NULL
        at org.opensearch.analytics.exec.DefaultPlanExecutor.lambda$doExecute$1(DefaultPlanExecutor.java:277)
        at org.opensearch.threadpool.TaskAwareRunnable.doRun(TaskAwareRunnable.java:78)
        at org.opensearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:52)
        at org.opensearch.common.util.concurrent.TimedRunnable.doRun(TimedRunnable.java:59)
        at org.opensearch.common.util.concurrent.ThreadContext$ContextPreservingAbstractRunnable.doRun(ThreadContext.java:1011)
        at org.opensearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:52)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
        at java.base/java.lang.Thread.run(Thread.java:1474)
Caused by: java.lang.AssertionError: type mismatch:
aggCall type:
BIGINT
inferred type:
BIGINT NOT NULL
        at org.apache.calcite.util.Litmus.lambda$static$0(Litmus.java:31)
        at org.apache.calcite.plan.RelOptUtil.eq(RelOptUtil.java:2223)
        at org.apache.calcite.rel.core.Aggregate.typeMatchesInferred(Aggregate.java:454)
        at org.apache.calcite.rel.core.Aggregate.<init>(Aggregate.java:177)
        at org.opensearch.analytics.planner.rel.OpenSearchAggregate.<init>(OpenSearchAggregate.java:89)
        at org.opensearch.analytics.planner.rules.OpenSearchAggregateSplitRule.onMatch(OpenSearchAggregateSplitRule.java:192)
        at org.apache.calcite.plan.volcano.VolcanoRuleCall.onMatch(VolcanoRuleCall.java:223)
        at org.apache.calcite.plan.volcano.IterativeRuleDriver.drive(IterativeRuleDriver.java:61)
        at org.apache.calcite.plan.volcano.VolcanoPlanner.findBestExp(VolcanoPlanner.java:525)
        at org.opensearch.analytics.planner.PlannerImpl.cbo(PlannerImpl.java:267)
        at org.opensearch.analytics.planner.PlannerImpl.runAllOptimizations(PlannerImpl.java:104)
        at org.opensearch.analytics.planner.PlannerImpl.createPlan(PlannerImpl.java:77)
        at org.opensearch.analytics.exec.DefaultPlanExecutor.executeInternal(DefaultPlanExecutor.java:161)
        at org.opensearch.analytics.exec.DefaultPlanExecutor.lambda$doExecute$1(DefaultPlanExecutor.java:265)
```



-----
After the fix:

Phase 1 (logical) 
```
LogicalSystemLimit(fetch=10000)
  LogicalProject(count()=$1, sum(AdvEngineID)=$2, RegionID=$0)
    LogicalAggregate(group={0}, count()=COUNT(), sum(AdvEngineID)=SUM($1))
      LogicalProject(RegionID=$65, AdvEngineID=$0)
        LogicalTableScan(clickbench)
```

Phase 2 (after marking) 

```
OpenSearchProject(count()=$1, sum(AdvEngineID)=$2, RegionID=$0)
  OpenSearchAggregate(group={0}, count()=COUNT(), sum(AdvEngineID)=SUM($1), mode=SINGLE)
    OpenSearchProject(RegionID=$65, AdvEngineID=$0)
      OpenSearchTableScan(clickbench)
```

Phase 3 (after CBO) 
```
OpenSearchProject(count()=$1, sum(AdvEngineID)=$2, RegionID=$0)
  OpenSearchAggregate(group={0}, count()=COUNT(), sum(AdvEngineID)=SUM($1), mode=SINGLE)
    OpenSearchProject(RegionID=$65, AdvEngineID=$0)
      OpenSearchExchangeReducer(distributionType=SINGLETON)      
        OpenSearchTableScan(clickbench)
```

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
vinaykpud: With this approach I found some difference in way Plan get picked for similar shaped query:

### Query A: [count, sum] order 

```
source = clickbench | stats count(), sum(AdvEngineID) by RegionID
```


```
  OpenSearchProject(count()=$1, sum(AdvEngineID)=$2, RegionID=$0)
    OpenSearchAggregate(group={0}, count()=COUNT(), sum(AdvEngineID)=SUM($1), mode=SINGLE)
      OpenSearchProject(RegionID=$65, AdvEngineID=$0)
        OpenSearchExchangeReducer(SINGLETON)        
          OpenSearchTableScan(clickbench)
```

No PARTIAL/FINAL split pickedup  by volcano planner


### Query B: [sum, count] order 

```
source = clickbench | stats sum(AdvEngineID), count() by RegionID
```

```
  OpenSearchProject(sum(AdvEngineID)=$1, count()=$2, RegionID=$0)
    OpenSearchAggregate(group={0}, sum(AdvEngineID)=SUM($1), count()=COUNT(), mode=FINAL)
      OpenSearchExchangeReducer(SINGLETON)
        OpenSearchAggregate(group={0}, sum(AdvEngineID)=SUM($1), count()=COUNT(), mode=PARTIAL)
          OpenSearchProject(RegionID=$65, AdvEngineID=$0)
            OpenSearchTableScan(clickbench)
```

PARTIAL, then FINAL Aggregation split pickedup by volcano planner



Lets say we have 


[cnt, sum] as order, FINAL aggCalls looks like [COUNT(), SUM($1)]
[sum, cnt] as order, FINAL aggCalls looks [SUM($1), COUNT()]

Important thing to note is argList [1] for SUM unchanged

Now run `typeMatchesInferred` on each:

[cnt, sum] order : slot 1 of FINAL aggCalls is SUM($1):
  - argList=[1] → reads gathered[1] = count_partial : BIGINT NOT NULL
  - AGG_SUM calls deriveSumType(BIGINT NOT NULL) → returns BIGINT NOT NULL
  - Stored stamped to BIGINT NOT NULL ← but the user wrote SUM(AdvEngineID), which should be SUM-of-INTEGER, not SUM-of-count_partial
  - Row type tightens, cast Project gets added, Volcano picks SINGLE


[sum, cnt] order : slot 0 of FINAL aggCalls is SUM($1):
  - argList=[1] → reads gathered[1] = sum_partial : INTEGER NOT NULL
  - AGG_SUM calls deriveSumType(INTEGER NOT NULL) → returns INTEGER NOT NULL
  - Stored matches user's stamped INTEGER NOT NULL
  - No row-type drift, no cast, Volcano picks PARTIAL+FINAL


We need a different approach something similar to https://github.com/opensearch-project/OpenSearch/compare/main...vinaykpud:OpenSearch:feature/fix-5-queries?expand=1 
where we need to adjust the FINAL aggregate aggFunctions to point to right inputRef pointer.

Going to continue this PR in separate branch

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -180,21 +181,113 @@ public void onMatch(RelOptRuleCall call) {
...
+        for (int i = 0; i < aggCalls.size(); i++) {
+            AggregateCall call = aggCalls.get(i);
+            AggregateCall derived = AggregateCall.create(
+                call.getAggregation(),
+                call.isDistinct(),
+                call.isApproximate(),
+                call.ignoreNulls(),
+                call.rexList,
+                call.getArgList(),
+                call.filterArg,
```
songkant-aws, line 278: Thanks for fixing the grouping sets mapping. One of my command test case is also impacted by this bug. There is one more edge case needs to be addressed here. 

Sometimes AGG query can carry filters on raw data rows like `COUNT() FILTER ($predicates)`. We can drop filterArg on FINAL aggCall: Filter is consumed by PARTIAL on raw data rows. FINAL's input is PARTIAL output where the predicate column no longer exists, so retaining filterArg trips Aggregate.\<init\>'s isPredicate validation, which will throw validation error if the filterArg index is out of row type boundary.

vinaykpud: Thanks @songkant-aws for taking look and pointing to edge case. Could you please provide an example PPL query. I am currently working the fix for this

songkant-aws: @vinaykpud A simple example query could be like: `source=accounts | head 5 | fields firstname, age, balance | transpose`. Transpose command will create such agg shape with filter conditions

vinaykpud:
we have merged fix with https://github.com/opensearch-project/OpenSearch/pull/21875, there are other changes in `qa/analytics-engine-rest/src/test` package in this PR which may need to be merged

PR #21849, opened 2026-05-27
[DFA Tiering / Block Cache] Defer H2W cancel write-block removal, add Foyer key_index persistence, and fix retention lease renewal

vishwasgarg18:
---

### Description

This PR bundles three related areas of work for the DFA (pluggable-dataformat) tiered storage path:

**1. Deferred write-block removal after H2W cancel**

Cancelling a hot-to-warm migration for a DFA index was immediately removing the write block during the cancel cluster state task. This created a window where write requests could reach warm-node shards that were still running the read-only Parquet engine (before physical relocation completed), causing ingestion errors.

- `HotToWarmTieringService.getIndexTierSettingsToRestoreAfterCancellation()` no longer sets `INDEX_BLOCKS_WRITE=false`; the base class `getIndexTierClusterBlocksToRestoreAfterCancellation()` no longer removes `INDEX_WRITE_BLOCK`
- Added `TieringService.removeWriteBlockForCancelledDfaIndices()` — a cluster-event listener that removes the block only when: DFA index, `INDEX_TIERING_STATE=HOT`, `INDEX_BLOCKS_WRITE=true`, not actively tiering, and **all shards are `started` on hot nodes**
- W2H cancel path unchanged: re-adds the write block immediately (correct — reverting to warm means read-only engine)
- Unit tests: `TieringServiceTests`, `HotToWarmTieringServiceTests`, `WarmToHotTieringServiceTests`

**2. Foyer key_index persistence (file-based shard registry)**

Adds a disk-backed key_index to the Foyer block cache to survive warm node restarts without cold-read penalties:

- `key_index_store.rs`: atomic JSON serialization/deserialization with version validation; graceful fallback on corrupt or wrong-version files
- Independent periodic persist task (configurable interval, default 60s) decoupled from the sweep task
- Sweep threshold guard: skips DashMap scan when cache is less than N% full (default 70%)
- `FoyerBlockCacheSettings`: new settings for `key_index_sweep_interval_seconds`, `key_index_sweep_threshold`, `key_index_persist_interval_seconds`; default cache size updated to 50%
- Live FFM downcall handles: `foyer_update_sweep_threshold`, `foyer_update_sweep_interval`, `foyer_update_persist_interval` via `FoyerBridge`
- Rust integration tests: 10k-entry scale tests for key_index serialization, periodic persist, bulk evict_prefix, prune
- IT infrastructure: `BlockCacheKeyIndexRecoveryIT`, `BlockCacheKeyIndexScaleIT`, `BlockCacheStatsIT`, `PruneBlockCacheIT` — all with `nodeSettings()` override to enable Foyer via `FeatureFlags.PLUGGABLE_DATAFORMAT_

### DIFF
--- a/server/src/main/java/org/opensearch/index/shard/IndexShard.java
+++ b/server/src/main/java/org/opensearch/index/shard/IndexShard.java
@@ -132,6 +132,7 @@ import org.opensearch.index.cache.request.ShardRequestCache;
 import org.opensearch.index.codec.CodecService;
 import org.opensearch.index.engine.CommitStats;
 import org.opensearch.index.engine.DataFormatAwareEngine;
+import org.opensearch.index.engine.DataFormatAwareReadOnlyEngine;
 import org.opensearch.index.engine.Engine;
 import org.opensearch.index.engine.Engine.GetResult;
 import org.opensearch.index.engine.EngineBackedIndexer;
@@ -823,7 +824,16 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
                     if (currentRouting.initializing() && currentRouting.isRelocationTarget() == false && newRouting.active()) {
                         // the cluster-manager started a recovering primary, activate primary mode.
                         replicationTracker.activatePrimaryMode(getLocalCheckpoint());
-                        postActivatePrimaryMode();
+                        // DFA warm primaries: skip postActivatePrimaryMode (no remote translog upload
+                        // needed) but still ensure peer recovery retention leases exist for replicas.
+                        // Reset hasAllPeerRecoveryRetentionLeases to false first to prevent assertion
+                        // failures in renewPeerRecoveryRetentionLeases() during the async window.
+                        if (getIndexer() instanceof DataFormatAwareReadOnlyEngine == false) {
+                            postActivatePrimaryMode();
+                        } else {
+                            replicationTracker.resetHasAllPeerRecoveryRetentionLeases();
+                            ensurePeerRecoveryRetentionLeasesExist();
+                        }
                     }
                 } else {
                     assert currentRouting.primary() == false : "term is only increased as part of primary promotion";
@@ -896,14 +906,22 @@ public class IndexShard extends AbstractIndexShardComponent implements IndicesCl
                                 // Force update the checkpoint post engine reset.
                                 updateReplicationCheckpoint();
                             }
-
                             replicationTracker.activatePrimaryMode(getLocalCheckpoint());
                             if (indexSettings.isSegRepEnabledOrRemoteNode()) {
                                 // force publish a checkpoint once in primary mode so that replicas not caught up to previous primary
                                 // are brought up to date.
                                 checkpointPublisher.publish(this, getLatestReplicationCheckpoint());
                             }
-                            postActivatePrimaryMode();
+                            // DFA warm primaries: activate primary mode (needed for initiateTracking
+                            // during replica recovery) but skip postActivatePrimaryMode (no remote
+                            // translog upload). Reset hasAllPeerRecoveryRetentionLeases and ensure
+                            // retention leases exist for replicas.
+                            if (getIndexer() instanceof DataFormatAwareReadOnlyEngine == false) {
+                                postActivatePrimaryMode();
+                            } else {
+                                replicationTracker.resetHasAllPeerRecoveryRetentionLeases();
+                                ensurePeerRecoveryRetentionLeasesExist();
+                            }
                             /*
                              * If this shard was serving as a replica shard when another shard was promoted to primary then
                              * its Lucene index was reset during the primary term transition. In particular, the Lucene index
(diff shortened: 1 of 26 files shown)


nishchay21 (approved):

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -823,7 +824,16 @@ public void updateShardState(
                     if (currentRouting.initializing() && currentRouting.isRelocationTarget() == false && newRouting.active()) {
                         // the cluster-manager started a recovering primary, activate primary mode.
                         replicationTracker.activatePrimaryMode(getLocalCheckpoint());
-                        postActivatePrimaryMode();
+                        // DFA warm primaries: skip postActivatePrimaryMode (no remote translog upload
+                        // needed) but still ensure peer recovery retention leases exist for replicas.
+                        // Reset hasAllPeerRecoveryRetentionLeases to false first to prevent assertion
+                        // failures in renewPeerRecoveryRetentionLeases() during the async window.
+                        if (getIndexer() instanceof DataFormatAwareReadOnlyEngine == false) {
+                            postActivatePrimaryMode();
```
mgodwan, line 832: Can we avoid cast here, and check for warm+pluggable_format enabled?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -4135,16 +4153,30 @@ public void activateWithPrimaryContext(final ReplicationTracker.PrimaryContext p
             // So, we can use a stricter check where local checkpoint of new primary is checked against that of old primary.
             allocationId = primaryContext.getRoutingTable().primaryShard().allocationId().getId();
         }
+
+        // DFA warm primaries: relax checkpoint assertion. During hot-to-warm relocation after cancel+bulk+retry,
```
mgodwan, line 4157: Are we not flushing on prev primary before recovery is initiated on warm?

vishwasgarg18: Yes, we are flushing.

But the below race condition exists

T0: Hot Primary -> Retention leases (LocalCP - 0)
T1: Document indexed -> local CP -> 2 but the 
     - Retention leases (LocalCP - 0) (if AsyncRetentionLeaseSyncTask hasn't fired yet)
T2: Recovery starts
   - Create the clone retention lease for Warm Primary (LocalCP - 0)
T3: AsyncRetentionLeaseSyncTask runs on Hot primary 
  - Updates the Hot Primary Retention Lease (LocalCP -2) 
  - Warm Replica shard started and marks the allShardsStarted = true
  - Expires the stale WP retention lease (LocalCP - 0) if allShardsStarted = true

T4: Recovery completes (doesn't have lease for WP - already expired)
  -  handoffPrimaryContext 
      - tries to renew WP lease but skips since (hasAllPeerRecoveryRetentionLeases = true and was never updated)
     - Hence the assertion fails
     
```
assert (checkpoints.get(shardRouting.allocationId().getId()).tracked
                        && checkpoints.get(shardRouting.allocationId().getId()).replicated == false)
                        || checkpoints.get(shardRouting.allocationId().getId()).tracked == false
                        || hasAllPeerRecoveryRetentionLeases == false;


As there is not retention lease for Warm replica
```

--------------
If before the AsyncRetentionLeaseSyncTask could expire the recovery completes and renew the cloned retention lease we don't see this issue

server/src/main/java/org/opensearch/index/shard/IndexShard.java
vishwasgarg18: I think it's better to keep the cast as DataFormatAwareReadOnlyEngine doesn't really care about the leases as there are no translog replays required and tomorrow for writable warm we don't have to correct these changes again.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -4135,16 +4153,30 @@ public void activateWithPrimaryContext(final ReplicationTracker.PrimaryContext p
...
         synchronized (mutex) {
             replicationTracker.activateWithPrimaryContext(primaryContext); // make changes to primaryMode flag only under mutex
         }
-        postActivatePrimaryMode();
+        // DFA warm primaries: skip postActivatePrimaryMode (no remote translog upload needed).
+        // Reset hasAllPeerRecoveryRetentionLeases and ensure retention leases exist for replicas.
+        if (getIndexer() instanceof DataFormatAwareReadOnlyEngine == false) {
+            postActivatePrimaryMode();
+        } else {
+            replicationTracker.resetHasAllPeerRecoveryRetentionLeases();
```
shourya035, line 4177: I get why we are doing this and RLs would probably be never used in warm as the source of truth would always be Remote Store in this situation. 

I am okay with resetting all RLs here, ensure that all the existing ITs are passing in this case.

shourya035 (approved):
Approving in terms of resetting the RLs during H2W recovery

server/src/main/java/org/opensearch/index/shard/IndexShard.java
vishwasgarg18: ITs have passed.

mgodwan (approved):

PR #21865, opened 2026-05-28
Prune search shard groups using index-level field domains before can_match

### LINKED ISSUE #21566: [RFC/Proposal]: Index-level search pruning before can_match using field bounds metadata
Author @lamlamine

_Note: Portions of this document were reformatted and edited with the assistance of generative AI tools for clarity and readability. The design, technical content, and proposal originate from the author._ 
 
# TL;DR

Introduce a lightweight coordinator-side optimization that prunes irrelevant indices before `can_match` using trusted index-level field bounds metadata.

The optimization targets workloads where wildcard index patterns or explicitly provided index lists expand to large numbers of indices while queries only match a small subset of them. This is especially beneficial for TSDB-style deployments with narrow time-range filters, but the approach is generic and can improve performance in any scenario involving many indices and selective query constraints.

The feature is generic and can improve performance in any scenario involving many indices and selective mandatory query constraints. Time-range filtering is one example, but the same architecture can also support other types of selective predicates such as low-cardinality term or boolean constraints in future evolutions.
 
Current `can_match` pruning occurs after shard fanout and still performs transport/query rewrite work against many irrelevant shards. This proposal introduces an earlier pruning stage to reduce shard fanout itself.

The initial implementation is intentionally conservative:

- one configured range field
- finalized/trusted bounds only
- local cluster only
- mandatory query constraints only
- metadata-driven pruning only

Bounds metadata will initially be published by ISM during rollover once indices become stable/finalized.

---

### Expected Improvement

Expected gains are primarily coordinator CPU reduction and lower `can_match` fanout costs for TSDB-style wildcard searches.

For workloads with:

- hundreds/thousands of historical indices
- wildcard index patterns
- narrow timestamp filters

the optimization can significantly reduce:

- shard iterator fanout
- `can_match` transport requests
- query rewrite overhead
- Lucene BKD/PointValues range checks
- coordinator CPU utilization

Example scenario:

```text
wildcard expands to 500 historical indices
query timestamp overlaps only latest index
````

Current behavior:

```text
500 indices
→ can_match fanout across many shards
→ query rewrite/BKD checks on all candidates
```

Expected behavior after pruning:

```text
500 indices
→ early index-level pruning
→ only relevant index/shards reach can_match
```

Expected impact:

* substantial reduction in unnecessary `can_match` work
* lower coordinator CPU usage
* lower transport fanout
* better scalability for large TSDB clusters
* improved behavior for high-segment-count historical clusters

while preserving correctness via conservative pruning semantics.

---

# Problem Statement

 

We investigated a production OpenSearch workload experiencing very high coordinator CPU usage caused by wildcard TSDB-style searches.

The workload stores operational metrics and observability data in time-series-like indices such as:

```text
metrics-service-*
````

Client queries often target broad wildcard index patterns such as:

```text
metrics-*
```

while also filtering on a very narrow timestamp range (for example last 1–2 minutes).

Example simplified query:

```json
{ 
  "query": {
    "bool": {
      "filter": [
        {
          "range": {
            "@timestamp": {
              "gte": "now-2m",
              "lte": "now"
            }
          }
        }
      ]
    }
  }
}
```

In practice, the wildcard expansion matches hundreds of historical indices, even though only the latest/current index actually overlaps the requested timestamp range.

Currently OpenSearch still:

1. expands the wildcard to all matching indices
2. builds shard iterators for all matching indices
3. executes `can_match` fanout across many irrelevant historical shards

Production hot threads and flamegraphs showed significant CPU in:

* `SearchService.canMatch`
* `RangeQueryBuilder.doRewrite`
* `DateFieldMapper.isFieldWithinQuery`
* Lucene `PointValues` / BKD metadata checks

The issue is not query execution itself, but the coordinator fanout and `can_match` work performed against large numbers of irrelevant historical indices.

 
  
---

# Current Limitation

Today `can_match` operates at shard-level and occurs after shard iterator construction and shard fanout.

This means OpenSearch already performs transport fanout and query rewrite work across many shards before discovering that most shards cannot possibly match.

For workloads with:

* many historical indices
* wildcard index patterns
* narrow timestamp filters

this can create substantial unnecessary CPU overhead.

---

# Proposed Solution

Introduce a new optional coordinator-side optimization layer:

```text
Index-level predicate pruning before can_match
```

using trusted index-level field bounds metadata.

High-level flow:

```text
search request
  → wildcard expansion
  → shard iterator creation
  → index-level pruning (new)
  → can_match
  → query phase
```

The optimization would prune shard iterators before `can_match` when OpenSearch can safely prove that an index cannot contain matching documents.

---

# Example

Assume:

```text
query range:
  @timestamp between May 4 18:25 and 18:27
```

and index metadata:

```text
index-a:
  @timestamp min/max = Apr 1 → Apr 2

index-b:
  @timestamp min/max = May 4 18:20 → May 4 18:30
```

Then:

```text
index-a can be pruned safely
index-b must be searched
```

This prevents unnecessary `can_match` requests to irrelevant historical indices.

---

# V1 Scope

The initial implementation intentionally keeps scope conservative.

V1 supports:

* local cluster only (no CCS pruning)
* one configured range field
* mandatory query constraints only
* finalized/trusted bounds only
* metadata-driven pruning only

V1 intentionally does NOT support:

* arbitrary query analysis
* should/must_not pruning semantics
* multi-field pruning
* remote cluster pruning
* automatic bounds computation
* aggressive pruning heuristics

---

# Proposed Architecture

## Query constraints

Extract only mandatory constraints from:

* `bool.filter`
* `bool.must`

Ignore:

* `should`
* `must_not`

Example V1 constraint:

```text
RangeQueryConstraint(@timestamp, from, to)
```

---

## Field bounds abstraction

Introduce generic abstraction:

```java
interface FieldBounds {
    String field();
    boolean finalized();
    boolean canMatch(QueryConstraint constraint);
}
```

V1 implementation:

```text
RangeFieldBounds
```

This keeps the design extensible for future:

* boolean fields
* low-cardinality term fields
* multiple constraints
* additional predicate types

---

## Metadata source

Bounds are provided via trusted index metadata.

For the current proposed implementation, the metadata is published by ISM during rollover, when the previous write index becomes stable/finalized.

Example:

```json
{
  "index_field_bounds": {
    "@timestamp.type": "range",
    "@timestamp.min": "1777852800000",
    "@timestamp.max": "1777939199999",
    "@timestamp.finalized": "true",
    "@timestamp.source": "ism_rollover"
  }
}
```

The important design distinction is that:

* ISM acts only as a metadata publisher
* the search layer remains generic and metadata-driven
* search pruning itself does not depend on ISM internals

This avoids coupling the search engine directly to TSDB/ISM semantics while still leveraging rollover lifecycle events to publish stable trusted bounds.

Future publishers could include:

* data streams
* manual APIs
* plugins
* future automatic collectors
* other lifecycle management systems

without changing the coordinator-side pruning logic.

---
 
  

# Planned Future Evolution

The current implementation is intentionally conservative and limited to a single configured range field in V1.

The architecture was designed to evolve incrementally without changing the core pruning flow.

Planned V2 work includes support for:

* multiple mandatory constraints
* multiple configured fields
* combining range pruning with low-cardinality term/boolean pruning

Examples:

* `@timestamp`
* `replication_role`
* `environment`
* boolean flags

Planned V3 work includes:

* additional `FieldBounds` implementations beyond numeric/date ranges
* low-cardinality `TermsFieldBounds`
* richer query constraint extraction
* data stream integration
* automatic bounds publishers
* optional CCS pruning support
* planner/cost-based pruning
* shard-level metadata

The abstraction intentionally separates:

* query constraint extraction
* bounds storage/publication
* pruning execution

so future optimizations can be added without rewriting the coordinator search flow or changing existing correctness guarantees.

 

_No response_

### Related component

None

_No response_

### Additional context

_No response_

see above

### Related component

Search

_No response_

### Additional context

_No response_

laminelam:
Resolves [#21566](https://github.com/opensearch-project/OpenSearch/issues/21566)

Implements the search-core consumer side of index-level search pruning described in the [RFC](https://github.com/opensearch-project/OpenSearch/issues/21566).

This adds a conservative coordinator-side pruning step before `can_match` that uses finalized index-level field-domain metadata from `IndexMetadata` custom data. If metadata is missing, malformed, unfinalized, unsupported, or unsafe to evaluate, the shard group is kept.

The implementation introduces generic field-domain abstractions with initial date-range support, mandatory query-constraint extraction, dynamic pruning settings, PIT safety guards, and test coverage.  

### Benchmark
 

- 90 daily indices
- 1M docs 

Note: Benchmark was done on local (**no network involved**).  Although the improvement is already significant locally it should be even more pronounced in a distributed cluster, where pruning also avoids remote can_match requests.

.

| Metric | Pruning disabled | Pruning enabled | Reduction | Improvement |
|---|---:|---:|---:|---:|
| Avg wall time | 60.084 ms | 11.845 ms | 80.29% | 5.07x faster |
| P50 wall time | 66.713 ms | 11.610 ms | 82.60% | 5.75x faster |
| P90 wall time | 85.724 ms | 13.432 ms | 84.33% | 6.38x faster |
| Avg response took | 59.200 ms | 11.100 ms | 81.25% | 5.33x faster |
| Avg can_match requests | 27 | 0.6 | 97.78% | 45x fewer | 
 
 
 
. 
  ### Core Classes

| Class | Role |
|---|---|
| `SearchIndexPruningService` | Main coordinator-side service. Computes pruned shard groups and fails open on pruning errors. |
| `SearchIndexPruningSettings` | Dynamic cluster settings: enabled flag, min shard threshold, configured pruning fields. |
| `SearchIndexPruningResult` | Immutable result describing which original shard groups may be skipped. |
| `QueryConstraint` | Generic mandatory query constraint abstraction. |
| `RangeQueryConstraint` | Query-side range constraint. It is intentionally not date-specific. |
| `QueryConstraintExtractor` | Extracts mandatory constraints from a search request. |
| `MandatoryQueryConstraintExtractor` | Walks mandatory query clauses and emits constraints for configured pruning fields. |
| `FieldDomain` | Index-level description of values that may exist for one field in one concrete index. |
| `DateRangeFieldDomain` | Initial field-domain implementation for date |
| `FieldDomainProvider` | Resolves field domains for an index and field. |
| `IndexFieldDomainMetadata` | Codec for `index_field_domains` custom metadata. |
| `FieldDomainParser` | Parser/encoder for one field-domain metadata type. |
| `FieldDomainParserRegistry` | Maps metadata `type` values to parser implementations. |
| `FieldDomainEvaluator` | Proves whether a field domain can or cannot match a query constraint. |
| `DateRangeFieldDomainEvaluator` | Evaluates date range domains against range query constraints using OpenSearch date parsing. |
| `FieldDomainEvaluationContext` | Request-scoped evaluation context, currently search start time for date math `now`. |
| `ClusterStateFieldDomainProvider` | Reads field-domain metadata from `IndexMetadata.getCustomData(...)`. |
  

.

image

  
### Check List


laminelam:
> ## PR Reviewer Guide 🔍
> Here are some key observations to aid the review process:
> 
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ✅ **No TODO sections**
> 🔀 **No multiple PR themes**
> ⚡ **Recommended focus areas for review**
> 
> [**Possible Issue**](https://github.com/opensearch-project/OpenSearch/pull/21865/files#diff-3918f4a2920f8c97bca1dc2736095eca4bca6c71a7da6a60a2426e62e2ec9030R224-R226)
> When all shard groups are skipped by pruning, the code attempts to set the first active shard index. However, if `activeShardIndexToOriginalShardIndex` is empty, `firstActiveShardIndex` defaults to 0, which may not be a valid shard group index if all shards were skipped before can_match. This could cause an out-of-bounds access or incorrect shard selection when `possibleMatches.set(results.getFirstActiveShardIndex())` is called on line 225.
> 
> ```java
> if (results.hasActiveShards()) {
>     possibleMatches.set(results.getFirstActiveShardIndex());
> }
> ```
> 
> [**Possible Issue**](https://github.com/opensearch-project/OpenSearch/pull/21865/files#diff-53d3a491629c8e6e8c7650918a720ada8140a304bd9b091f54a44c5fe558a1d6R181-R188)
> When all shard groups would be pruned, the code falls back to the original iterators without clearing any skip flags that may have been set. If any iterator was already marked as skipped before pruning (e.g., by a previous phase), this fallback preserves that skip state, potentially causing zero shards to be queried when at least one was expected. The fallback should ensure at least one shard group is queryable.
> 
> ```java
> if (pruned == originalSize) {
>     /*
>      * Pruning is an optimization. If every shard group would be skipped, fall back to
>      * the original iterators until zero-shard response semantics are covered explicitly
>      * for this pre-can-match stage.
>      */
>     return SearchIndexPruningResult.notPruned(shardIterators);
> }
> ```
> 
> [**Possible Issue**](https://github.com/opensearch-project/OpenSearch/pull/21865/files#diff-03d20bd807401468393765eb8ed312ddfe608b56bbc1fa83575c9ec7e7719c0bR97-R102)
> When converting an exclusive lower bound to inclusive, the code increments `lowerInclusive` without checking for `Long.MAX_VALUE` overflow. If `lowerInclusive` is already `Long.MAX_VALUE`, incrementing it wraps to `Long.MIN_VALUE`, producing an incorrect range that may prune matching shards. This scenario is narrow but possible with date_nanos fields or extreme date math expressions.
> 
> ```java
> if (constraint.includeLower() == false) {
>     if (lowerInclusive == Long.MAX_VALUE) {
>         return Optional.of(NormalizedRange.EMPTY);
>     }
>     lowerInclusive++;
> }
> ```


1- When all shard groups are skipped before can_match, _hasActiveShards()_ is false, so possibleMatches.set(…) is not called.


2- Pruning should not clear existing skip flags because they may come from another search step.

3- This won’t happen because we are already checking  if (lowerInclusive == Long.MAX_VALUE)

laminelam:
@msfroh 

 If you have time can you please take a look?
 
 To facilitate the review I have  generated a visual and some classes description.

In short, the way it works is:
- Another service (for ex ISM) publishes the field domain (think of it as min max for timestamp for ex) to the index metadata.
- The field configuration is loaded from the index metadata and gets parsed and registered (hence the Registry). Only "finalized" fields are taken in account (we ignore "temporary" data)
- At query time, the query is parsed to extract the mandatory clauses (query constraints) for the configured field.
- Evaluators evaluate the field domain (index info) against query constraint (query info) to decide whether an index can_match 
- Non matching indices are skipped.

**The implementation is completely generic**, for now only _DateRangeFieldDomain_  is implemented but it is easy to add other implementations serving other types of data. 
The feature is disabled by default (opt-in).

server/src/main/java/org/opensearch/action/search/pruning/SearchIndexPruningSettings.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Query fields eligible for pruning.
+     */
+    public static final Setting<List<String>> FIELDS = Setting.listSetting(
+        "search.index_pruning.fields",
+        List.of(),
+        s -> s,
+        Dynamic,
+        NodeScope
```
msfroh, line 44: Nitpick: The common pattern (which IMO could be an overload for `listSetting` -- maybe even a dedicated `stringListSetting`, given how common it is) is to use `Collections.emptyList()` and `Function.identity()`. See e.g. https://github.com/opensearch-project/OpenSearch/blob/353b8901b7dd9b1788001e1e49d1c6a736c69ee4/server/src/main/java/org/opensearch/index/IndexModule.java#L210-L216

server/src/main/java/org/opensearch/action/search/pruning/MandatoryQueryConstraintExtractor.java
```diff
@@ -0,0 +1,89 @@
...
+            return;
+        }
+
+        if (query instanceof RangeQueryBuilder) {
+            RangeQueryConstraint constraint = buildRangeConstraint((RangeQueryBuilder) query, pruningFields);
+            if (constraint != null) {
+                constraints.add(constraint);
+            }
+        }
+    }
```
msfroh: This implements a sort of visitor pattern over `QueryBuilder`, but doesn't use `QueryBuilderVisitor`.

I think you can just use an implementation of `QueryBuilderVisitor` like:

```
    private static class MandatoryQueryConstraintVisitor implements QueryBuilderVisitor {
        private final Set<String> pruningFields;
        private final List<QueryConstraint> constraints = new ArrayList<>();

        private MandatoryQueryConstraintVisitor(Set<String> pruningFields) {
            this.pruningFields = pruningFields;
        }

        @Override
        public void accept(QueryBuilder qb) {
            if (query instanceof RangeQueryBuilder) {
                RangeQueryConstraint constraint = buildRangeConstraint((RangeQueryBuilder) query, pruningFields);
                if (constraint != null) {
                    constraints.add(constraint);
                }
            }
        }

        @Override
        public QueryBuilderVisitor getChildVisitor(BooleanClause.Occur occur) {
            switch (occur) {
                case MUST, FILTER:
                    return this;
                default:
                    return QueryBuilderVisitor.NO_OP_VISITOR;
            }
        }
    }
```

That has the advantage that it handles traversing through all `QueryBuilder` types, provided that their `visit` implementation is correct.

server/src/main/java/org/opensearch/action/search/pruning/FieldDomainParser.java
```diff
@@ -0,0 +1,49 @@
...
+    Optional<FieldDomain> parse(String field, Map<String, String> customData, String prefix);
+
+    /**
+     * Encodes bounds into a target metadata map.
+     *
+     * @param domain domain to encode
+     * @param target target metadata map
+     * @param prefix field-specific metadata prefix
+     */
+    void encode(FieldDomain domain, Map<String, String> target, String prefix);
```
msfroh: If the assumption is that a `FieldDomainParser` will be created for each `FieldDomain` type, I think we can add a type parameter to specify the corresponding `FieldDomain`. That is, make this `public interface FieldDomainParser<T extends FieldDomain>`, then `parse` can return `Optional<T>` and `encode` can take `T` as the first argument.

What do you think?

server/src/main/java/org/opensearch/action/search/SearchIndexPruningService.java
```diff
@@ -0,0 +1,258 @@
...
+
+        if (pruned == activeShardGroups) {
+            /*
+             * Pruning is an optimization. If every currently active shard group would be skipped, fall back to
+             * the original iterators until zero-shard response semantics are covered explicitly for this
+             * pre-can-match stage. Existing skip flags are preserved because they may have been set by another
+             * search execution step.
+             */
+            return SearchIndexPruningResult.notPruned(shardIterators);
+        }
```
msfroh, line 194: I'm curious about this part. If we do have a query that prunes away all active shard groups, shouldn't we return an empty response? Or is this just here for now to "fail open" if we accidentally prune everything?

msfroh (commented):
Thanks @laminelam!

Overall, it looks pretty good. Thank you for taking the time to think about how to evolve this over time, while keeping your logic largely scoped to the new `pruning` package.

I have a few comments/suggestions, but I like the overall shape of the solution.

server/src/main/java/org/opensearch/action/search/SearchIndexPruningService.java
```diff
@@ -0,0 +1,258 @@
...
+                prunedShardGroupIndexes.set(shardGroupIndex);
+                pruned++;
+            }
+        }
+
+        if (pruned == 0) {
+            return SearchIndexPruningResult.notPruned(shardIterators);
+        }
+
+        if (pruned == activeShardGroups) {
```
owaiskazi19, line 186: Can add a warn log here. If all active shard groups would be pruned could be because of stale metadata or something

server/src/main/java/org/opensearch/action/search/SearchIndexPruningService.java
```diff
@@ -0,0 +1,258 @@
...
+    ) {
+        if (config.enabled == false) {
+            return true;
+        }
+
+        if (request == null || request.source() == null) {
+            return true;
+        }
+
+        if (request.pointInTimeBuilder() != null) {
```
owaiskazi19, line 208: Can add javadoc here why we are skipping for PIT. Something along the line
```
// PIT searches must query all shards from the PIT creation snapshot. 
// Pruning would break the PIT contract by skipping shards
```

server/src/main/java/org/opensearch/action/search/SearchIndexPruningService.java
```diff
@@ -0,0 +1,258 @@
...
+            return true;
+        }
+
+        if (shardIterators.size() < config.minShards) {
```
owaiskazi19: Instead of comparing with all shards including the skipped ones should we compare it with only active shards here?

server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
```diff
@@ -1273,6 +1282,65 @@ AbstractSearchAsyncAction<? extends SearchPhaseResult> searchAsyncAction(
...
+        SearchRequest searchRequest,
+        GroupShardsIterator<SearchShardIterator> shardIterators,
+        ClusterState clusterState,
+        SearchTimeProvider timeProvider
+    ) {
+        final SearchIndexPruningResult pruningResult = searchIndexPruningService.prune(
+            searchRequest,
+            shardIterators,
+            clusterState,
+            new FieldDomainEvaluationContext(timeProvider::getAbsoluteStartMillis)
```
owaiskazi19, line 1319: nit: Can add a debug log to emit no. of shards pruned?

server/src/main/java/org/opensearch/action/search/pruning/DateRangeFieldDomainEvaluator.java
```diff
@@ -0,0 +1,207 @@
...
+                DateFieldMapper.DateFieldType.parseToLong(value, roundUp, zone, parser, context.nowInMillisSupplier(), resolution)
+            );
+        } catch (RuntimeException e) {
+            return Optional.empty();
+        }
+    }
+
+    private static DateFieldMapper.Resolution resolveResolution(DateRangeFieldDomain domain) {
+        if (domain.resolution() == null || domain.resolution().isEmpty()) {
+            return DateFieldMapper.Resolution.MILLISECONDS;
```
owaiskazi19: We should document/comment this default resolution

server/src/main/java/org/opensearch/action/search/pruning/MandatoryQueryConstraintExtractor.java
```diff
@@ -0,0 +1,98 @@
...
+            if (queryBuilder instanceof RangeQueryBuilder) {
+                RangeQueryConstraint constraint = buildRangeConstraint((RangeQueryBuilder) queryBuilder, pruningFields);
+                if (constraint != null) {
+                    constraints.add(constraint);
+                }
+            }
+        }
+
+        @Override
+        public QueryBuilderVisitor getChildVisitor(BooleanClause.Occur occur) {
```
owaiskazi19, line 62: Does `dis_max` emit children as `SHOULD`? Worth adding a test for dis max

server/src/main/java/org/opensearch/action/search/pruning/MandatoryQueryConstraintExtractor.java
```diff
@@ -0,0 +1,98 @@
...
+                    return QueryBuilderVisitor.NO_OP_VISITOR;
+            }
+        }
+
+        private List<QueryConstraint> constraints() {
+            return List.copyOf(constraints);
+        }
+    }
+
+    private static RangeQueryConstraint buildRangeConstraint(RangeQueryBuilder range, Set<String> pruningFields) {
```
owaiskazi19, line 77: How field alias would be handled if the index has one?  For example `search.index_pruning.fields: ["@timestamp"]` if configured

server/src/main/java/org/opensearch/action/search/pruning/SearchIndexPruningSettings.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Enables coordinator-side index pruning before can-match/query execution.
+     */
+    public static final Setting<Boolean> ENABLED = Setting.boolSetting("search.index_pruning.enabled", false, Dynamic, NodeScope);
+
+    /**
+     * Minimum number of shard groups required before pruning is attempted.
+     */
+    public static final Setting<Integer> MIN_SHARDS = Setting.intSetting("search.index_pruning.min_shards", 128, 1, Dynamic, NodeScope);
```
owaiskazi19, line 34: Wouldn't 128 be a little high? Can scope it down little or we can make it auto adapt based on cluster topology for future

server/src/main/java/org/opensearch/action/search/pruning/SearchIndexPruningResult.java
```diff
@@ -0,0 +1,109 @@
...
+     * Creates a result containing pruned shard group indexes.
+     *
+     * At least one shard group must remain unpruned.
+     */
+    public static SearchIndexPruningResult pruned(
+        GroupShardsIterator<SearchShardIterator> shardIterators,
+        BitSet prunedShardGroupIndexes,
+        int prunedShardGroups
+    ) {
+        if (prunedShardGroups < 1 || prunedShardGroups >= shardIterators.size()) {
```
owaiskazi19, line 47: Should we check for activeShard here?

server/src/internalClusterTest/java/org/opensearch/action/search/SearchIndexPruningIT.java
```diff
@@ -0,0 +1,278 @@
...
+                failure.set(e);
+                latch.countDown();
+            }
+        });
+
+        assertTrue(latch.await(30, TimeUnit.SECONDS));
+        if (failure.get() != null) {
+            throw failure.get();
+        }
+    }
```
owaiskazi19, line 317: Can we have one test for multiple indices spanning the query range and  verify that partial overlap indices are kept?

owaiskazi19 (commented):
Promising improvement. Thanks @laminelam! Overall LG with few comments

server/src/main/java/org/opensearch/action/search/SearchIndexPruningService.java
laminelam: > I'm curious about this part. If we do have a query that prunes away all active shard groups, shouldn't we return an empty response? Or is this just here for now to "fail open" if we accidentally prune everything?

Hi @msfroh 

Yes, this is intentional fail-open behavior. 
In the existing _can_match_ implementation, if no shard can match, OpenSearch [still keeps one active shard ](https://github.com/opensearch-project/OpenSearch/blob/08ac04b1d95e94aae56d69be257b9be4be14467f/server/src/main/java/org/opensearch/action/search/CanMatchPreFilterSearchPhase.java#L155)so the query phase can produce a valid response, especially for aggregations.

This PR follows the same safe behavior and falls back when pruning would skip every active shard.

server/src/main/java/org/opensearch/action/search/pruning/FieldDomainParser.java
laminelam: @msfroh 
Forgot to quote you but I agree with those changes and I have implemented them. I went further in refactoring (notably moved the field-domain metadata classes under the _index.fielddomain_ package) so the future producer (the code that produces/publishes the metadata) implementation can reuse the same typed metadata model without adding search-core coupling.

laminelam:
> Promising improvement. Thanks @laminelam! Overall LG with few comments

Thank you @owaiskazi19  for the review.
I have implemented the suggestions, except for the debug logging which I think it adds unnecessary noisy logs for a feature that is disabled for most of the time.

owaiskazi19 (approved):
LGTM. Thanks @laminelam! Will wait for another review by @msfroh before merging this

laminelam:
@msfroh

msfroh (approved):
I'm looking forward to trying this out. I think it should help speed up a decent number of observability queries (assuming people populate the field domain metadata).

As a follow-up, I'm wondering if we should populate the field domain metadata as part of the "transition to read-only" logic or add a dedicated ISM action to populate domain metadata.

laminelam:
> I'm looking forward to trying this out. I think it should help speed up a decent number of observability queries (assuming people populate the field domain metadata).
> 
> As a follow-up, I'm wondering if we should populate the field domain metadata as part of the "transition to read-only" logic or add a dedicated ISM action to populate domain metadata.

I am thinking we need a dedicated action, but before that I need to push the "producer" PR (built on the top of this one). The idea is to have a separation of concern by separating the producer part (domain metadata writer) from the consumer (search). ISM would just call the producer API to write the domain metadata.
Thanks for the approval and merge.
@msfroh

PR #21867, opened 2026-05-28
Lucene as a driving backend for shard-local count fragments

alchemist51:
## What changed

Lucene is now a first-class driving backend for ungrouped count fragments expressible against the inverted index. The shard runs `IndexSearcher.count(boolQuery)`; DataFusion still drives the coordinator reduce that sums per-shard counts.

### Capability surface
- `ScanCapability.InvertedIndex` — metadata-only scan, no row values.
- `OpenSearchTableScanRule` — two-phase viability: strict for value-producing backends (must cover every requested field), permissive for the metadata-only driver (viable if it covers some indexed field, with type-aware `indexScanBackendsForField`).

### Coordinator-side alternative selection
- `PlanAlternativeSelector` runs after fork+adapter and before convert. With `analytics.planner.prefer_metadata_driver` (default `true`) and `canMetadataDriverExecute(fragment)` — fragment top is `OpenSearchAggregate` AND `groupSet.isEmpty()` — the Lucene alternative wins; otherwise non-drivable Lucene alternatives are dropped so the data-node fallback can't pick them either.
- Stage now ships exactly one `PlanAlternative`; convertor runs once per stage.

### Lucene driver
- `LuceneFragmentConvertor` lowers to `[outputColumnNames][hasFilter?][BoolQueryBuilder]` (NamedWriteable). `attachPartialAggOnTop` re-emits the wire bytes with the partial-Aggregate's column names spliced in front of the preserved tail (without this the multi-shard path stalled with a 0-column batch).
- `producesSubstraitFragments() = false` and `convertSchemaOnlyRead` returns a Substrait `Plan{Read{named_table; base_schema}}` stub so the coordinator's reduce sink can decode the partition schema. Stub columns are nullable Int64 to match the runtime batch (Calcite types COUNT as `BIGINT NOT NULL`, but the runtime emits nullable — runtime wins).
- `LuceneSearchExecEngine` runs `IndexSearcher.count(boolQuery)`; the count is exported through Arrow C-Data so Flight's `transferRoot` round-trips without zeroing the buffer.

### Delegation combiner
- New cluster setting `analytics.delegation.fuse_dual_viable` (default `false`). On OR/NOT, default behavior throws performance-delegated leaves back to native (driver evaluates, peer consults opportunistically); with the flag on, the entire boolean ships to the peer as one delegated expression.
- `unwrapAnnotations(RexNode)` strips nested `AnnotatedPredicate` markers before building the `delegation_possible` placeholder.

## Translation example

`where userID = 'arpit' AND event_type = 'click' AND match(message, 'alpha') | stats count()`

```
mark     every node viable = {lucene, datafusion}

fork     leaf stage alternatives = [lucene, datafusion]
         coord stage alternatives = [datafusion]   (Lucene has no ExchangeSink)

select   prefer_metadata_driver=true  →  leaf collapses to [lucene]

convert  leaf (Lucene):
           convertedBytes = [cnt][hasFilter=true]
                            [BoolQueryBuilder must=[Term(userID,'arpit'),
                                                     Term(event_type,'click'),
                                                     Match(message,'alpha')]]
           postDecorationSchemaBytes = Plan{Read{named_table:"input-1";
                                                  schema:[cnt: I64 nullable]}}
         coord (DataFusion):
           Substrait FinalAggregate(SUM(cnt)) over input-1

execute  shard:  searcher.count(boolQuery) → Arrow batch [cnt = N_i]
         coord:  registers partition with Lucene stub schema, sums N_i over shards
                 → result row [cnt = sum(N_i)]
```


sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/FilterDelegationIT.java
```diff
@@ -292,6 +292,95 @@ public void testOrOfTwoAndArms() throws Exception {
...
+
+        // Default (fuse=false) — carve-out path
+        setFuseDualViable(false);
+        try {
+            Map<String, Object> result = executePplViaShim(ppl);
+            @SuppressWarnings("unchecked")
+            List<List<Object>> rows = (List<List<Object>>) result.get("rows");
+            assertEquals("fuse=false: count for OR(MATCH, EQUALS)", 10L, ((Number) rows.get(0).get(0)).longValue());
+
+            // fuse=true — entire boolean ships to peer
```
bharath-techie: Is there no way to validate this via stats or anything? The test merely runs with two different settings but doesn't validate if the entire query is fused and delegated as one single query.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/CountFastPathIT.java
```diff
@@ -0,0 +1,501 @@
...
+     * (DataFusion-only). The numeric leaf forces DataFusion as the operator backend on the
+     * Filter, so the keyword leaf either correctness-delegates to Lucene (when DataFusion
+     * doesn't have it natively) or stays as a {@code delegation_possible} performance hint.
+     * Either way, the count must match the oracle.
+     */
+    public void testMixedAndFilter_keywordPlusNumeric() throws Exception {
+        createIndex();
+        ingestThreeSegments();
+
+        assertCount(
```
bharath-techie, line 150: Can we assert the plan, will the user id be delegated to Lucene or solved via DF in this case? Since this is a perf delegation.

Can we have a mixed query case with match as well (correctness delegation)

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/CountFastPathIT.java
```diff
@@ -0,0 +1,501 @@
...
+
+    /**
+     * Pure Lucene-side AND/OR over keyword fields. Both leaves are correctness-delegated to
+     * Lucene; the combiner emits a single fused {@code BoolQueryBuilder} which Lucene's
+     * {@code IndexSearcher.count} resolves via the term dictionary.
+     */
+    public void testKeywordBooleanFilter_luceneFused() throws Exception {
+        createIndex();
+        ingestThreeSegments();
+
```
bharath-techie, line 171: Let's add = + match as well ?

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/FragmentConversionDriverTests.java
```diff
@@ -1349,6 +1380,150 @@ public void testInterleavedAndOrNot() {
...
+     * <p>Setup: {@code status} is index=true (indexed integer) and {@code message} is keyword
+     * indexed — EQUALS on both is dual-viable (performance-delegation candidate). FUZZY on
+     * message is correctness-delegated to Lucene. Two distinct fields are used so Calcite's
+     * SEARCH/Sarg rewrite doesn't fold the two EQUALS into a single sargable predicate.
+     */
+    public void testOrPerformanceLeavesAndCorrectness_fuseDualViable_explicitFalse() {
+        RecordingConvertor dfConvertor = new RecordingConvertor();
+        RecordingSerializer serializer = new RecordingSerializer();
+        // Two distinct performance-delegation candidates:
+        // - EQUALS on status (integer, indexed) — dual-viable
```
bharath-techie: How is this dual viable today since only DF/parquet holds number fields?

bharath-techie: Do we mean index=true, BKD case here? Is it supported in prod use case as well? 

If not can we have keyword fields instead.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
```diff
@@ -92,6 +92,48 @@ public class AnalyticsPlugin extends Plugin implements ExtensiblePlugin, ActionP
...
+     * {@code delegation_possible} pattern (driver evaluates natively, peer-consults
+     * opportunistically) is incorrect: the driver can't tell whether a peer miss means "this
+     * leaf didn't match" or "no leaf matched". Forcing fusion trades that semantic guarantee
+     * for fewer round-trips. Flip to {@code true} only when measurements show the peer's
+     * filter (e.g. Lucene's term dictionary) is decisively faster than the driver's column
+     * scan AND the workload's OR shape benefits from the consolidation.
+     */
+    public static final Setting<Boolean> DELEGATION_FUSE_DUAL_VIABLE = Setting.boolSetting(
+        "analytics.delegation.fuse_dual_viable",
+        false,
```
bharath-techie: Should the defaults be true? Fusion should always be favorable since random access of Lucene can be utilised efficiently compared to managing bitsets in DF?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/CapabilityRegistry.java
```diff
@@ -259,6 +259,21 @@ public List<String> scanBackendsForField(FieldStorageInfo field) {
         return result;
     }
 
+    /**
+     * All backends that can scan this field's inverted index across all its formats. A backend
```
bharath-techie, line 263: Let's keep it generic as index? We don't need to mention inverted index specifically except in the examples.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanAlternativeSelector.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Lucene-as-driver only emits a single count value per shard — no grouping, no value
+     * materialization. Drivable fragments today are aggregates with an EMPTY group-set: a
+     * Calcite Aggregate's {@code groupSet.isEmpty()} means "count(*) over all rows", which
+     * Lucene's {@code IndexSearcher.count} answers directly. Anything else
+     * ({@code count() BY RegionID}, a bare {@code TableScan}, a Project, a Filter without
+     * a parent agg) needs row values or per-group values Lucene can't produce.
+     *
+     * <p>TODO: this is a coordinator-side coupling — the selector decides what Lucene can
```
bharath-techie: I wonder how hard is it to do this as part of the PR itself, if the scope of changes are minimal, can we do the todo and arrive at right abstractions?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanAlternativeSelector.java
```diff
@@ -0,0 +1,114 @@
...
+     * physically execute by inspecting RelNode shapes that really belong to the backend's
+     * own contract. When a second metadata-only backend lands, this should move behind a
+     * backend-owned predicate (e.g. {@code FragmentConvertor.canDriveFragment(RelNode)})
+     * so each backend declares its own drivability rule and the selector stays generic.
+     */
+    private static boolean canMetadataDriverExecute(RelNode fragment) {
+        return fragment instanceof OpenSearchAggregate agg && agg.getGroupSet().isEmpty();
```
bharath-techie: Can this aggregate be anything other than count like sum etc? How do we ensure this evaluates as true only for count(*)

Can we add tests as well for the same?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
```diff
@@ -162,16 +188,22 @@ private static void populatePostDecorationSchemas(Stage stage, CapabilityRegistr
...
-            // Cheap int compare first, then digest string compare via equals.
-            if (produced.getFieldCount() == expected.getFieldCount() && produced.equals(expected)) continue;
+            boolean schemaMismatch = produced.getFieldCount() != expected.getFieldCount() || produced.equals(expected) == false;
 
             List<StagePlan> updated = new ArrayList<>(child.getPlanAlternatives().size());
+            boolean changed = false;
             for (StagePlan plan : child.getPlanAlternatives()) {
                 FragmentConvertor convertor = registry.getBackend(plan.backendId()).getFragmentConvertor();
+                boolean nonSubstraitProducer = convertor.producesSubstraitFragments() == false;
+                if (schemaMismatch == false && nonSubstraitProducer == false) {
```
bharath-techie: Nit : Do we really need double negative wording with non substrait producer? Can we simply use the processSubstraitFragments directly?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
```diff
@@ -76,12 +76,23 @@ public class FragmentConversionDriver {
 
     private FragmentConversionDriver() {}
 
+    /**
+     * Backwards-compatible overload — equivalent to {@code convertAll(dag, registry, false)}.
```
bharath-techie: Why do we need to maintain bwc? We are within same release cycle?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneFragmentConvertor.java
```diff
@@ -0,0 +1,373 @@
...
+     * builds {@code FieldType(true, Int(64,true), null)} — the leading {@code true} is
+     * nullable). The Substrait stub MUST reflect the producer's actual runtime schema, not
+     * the Calcite logical type, otherwise the Rust-side partition stream registers as
+     * NOT-NULL and silently stalls when nullable batches arrive. Force nullable for now;
+     * when the driver supports more shapes, this will need a per-column source-of-truth.
+     *
+     * <p>TODO: when Lucene-driver shapes beyond COUNT land (group-by-count keys), wire in a
+     * proper Calcite→Substrait converter so the stub describes real producer schemas.
+     */
+    private static io.substrait.proto.Type toSubstraitType(RelDataType type) {
```
bharath-techie: Nit: can we import packages and just use the class names instead of full names everywhere?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/ScanCapability.java
```diff
@@ -28,4 +28,13 @@ record DocValues(Set<String> formats, Set<FieldType> supportedFieldTypes) implem
...
     record StoredFields(Set<String> formats, Set<FieldType> supportedFieldTypes) implements ScanCapability {
     }
+
+    /**
+     * Inverted index — drives metadata-only ops (count today; group-by-count, top-K terms
+     * later) via the term dictionary. Cannot deliver row values; consumers needing values
+     * must check value-producing caps. Today only Lucene-secondary, for keyword/text/
+     * match_only_text.
+     */
+    record InvertedIndex(Set<String> formats, Set<FieldType> supportedFieldTypes) implements ScanCapability {
```
bharath-techie: Can we keep this generic as index?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/AnalyticsPlugin.java
alchemist51: I was able run benchmarks and true makes a lot of sense:
```
Setting prefer_metadata_driver=false …

=== broad_text  + broad_kw  (Title:google ∪ BrowserCountry=h1 ~28%) ===
    source = clickbench | where match(Title, 'google') or BrowserCountry = 'h1' | stats count() as cnt
  cnt=28313573      fuse=false: 0.256s  fuse=true: 0.156s  delta -39.0%

=== broad_text  + narrow_kw (Title:google ∪ BrowserCountry=UA ~0.09%) ===
    source = clickbench | where match(Title, 'google') or BrowserCountry = 'UA' | stats count() as cnt
  cnt=126747        fuse=false: 0.238s  fuse=true: 0.026s  delta -89.1%

=== narrow_text + broad_kw  (URL:kinopoisk ∪ BrowserCountry=h1 ~28%) ===
    source = clickbench | where match(URL, 'kinopoisk') or BrowserCountry = 'h1' | stats count() as cnt
  cnt=28349909      fuse=false: 0.256s  fuse=true: 0.149s  delta -41.7%

=== narrow_text + narrow_kw (URL:kinopoisk ∪ BrowserCountry=UA ~0.09%) ===
    source = clickbench | where match(URL, 'kinopoisk') or BrowserCountry = 'UA' | stats count() as cnt
  cnt=185195        fuse=false: 0.232s  fuse=true: 0.025s  delta -89.2%

=== broad_text  + med_kw    (Title:google ∪ MobilePhoneModel=iPad ~5%) ===
    source = clickbench | where match(Title, 'google') or MobilePhoneModel = 'iPad' | stats count() as cnt
  cnt=5141331       fuse=false: 0.192s  fuse=true: 0.051s  delta -73.6%

=== very_broad_text + verybroad_kw (Title:google ∪ BrowserLanguage=S0 ~92%) ===
    source = clickbench | where match(Title, 'google') or BrowserLanguage = 'S0' | stats count() as cnt
  cnt=92363100      fuse=false: 0.233s  fuse=true: 0.356s  delta +52.7%

=== 3-way OR all delegatable (Title:google ∪ BrowserCountry=h1 ∪ MobilePhoneModel=iPad) ===
    source = clickbench | where match(Title, 'google') or BrowserCountry = 'h1' or MobilePhoneModel = 'iPad' | stats count() as cnt
  cnt=30451323      fuse=false: 0.335s  fuse=true: 0.187s  delta -44.2%
```

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/FilterDelegationIT.java
alchemist51: Updated the IT, it now capture the backend type and also the shape of the tree

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/CountFastPathIT.java
alchemist51: Updated

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/CountFastPathIT.java
alchemist51: Done

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/FragmentConversionDriverTests.java
alchemist51: Fixed, I have added a regression test here: `testCountByNumericField_drivenByDataFusion` which exactly validates it in the CountFastPathIT

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanAlternativeSelector.java
alchemist51: Covered it in here: PlanAlternativeSelectorTests.java

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanAlternativeSelector.java
alchemist51: Updated

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DelegatedPredicateCombiner.java
```diff
@@ -133,7 +147,24 @@ private Classified combine(RexCall call, List<Classified> kids, Function<Operato
...
+        // boolean into independently-evaluable arms — `delegation_possible` only composes under
+        // AND (driver evaluates the perf arm natively, peer is opportunistic). With a correctness
+        // sibling under OR, the driver can't evaluate at all, so the entire OR/NOT must ship to
+        // the peer as one delegated_predicate. Bubble up `call` itself; the parent's eventual
+        // `convertor.convertSubtree(call, ...)` walks the original AnnotatedPredicate operands
+        // and emits one BoolQuery (LuceneSubtreeConvertor handles AND/OR/NOT generically).
+        // Annotation id is "any one of the bubbled leaves" — same convention as the homogeneous
+        // bubble-up branches above; the (N-1) other ids become orphans (no placeholder and no
+        // DelegatedExpression references them since the OR collapses to one leaf at the top).
+        if (isOrNot && ordered.size() == correctnessChildren.size() + performanceChildren.size()) {
```
aasom143: Here it should be And condition with `fuseDualViable`, why isOrNot?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DelegatedPredicateCombiner.java
```diff
@@ -229,15 +260,44 @@ RexNode finalizeDelegated(Delegated d) {
...
+     * <p>{@link Delegated#subtree()} is either a single {@link AnnotatedPredicate} (leaf
+     * case) or a raw {@code RexCall} that bubbled up an AND/OR/NOT of same-backend children
+     * (the bubble branches in {@link #classify}). The driving backend evaluates the unwrapped
+     * expression natively, so every {@code AnnotatedPredicate} marker — leaf or under a
+     * bubbled AND/OR — has to come off before the placeholder is built.
      */
     RexNode makePlaceholder(Delegated d) {
         if (d.performanceDelegation()) {
-            RexNode original = ((AnnotatedPredicate) d.subtree()).unwrap();
+            RexNode original = unwrapAnnotations(d.subtree());
```
aasom143: Why we have changed this, this flow won't come in picture for your use case..? We have created Delegated node with performanceDelegation false.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DelegatedPredicateCombiner.java
aasom143: Instead of all changes in DelegatedPredicateCombiner class, we can make the performance delegate false, then it will treat that predicate as correctness and everything will be handled smoothly..

```
return new Delegated(peerBackend, node, ap.getAnnotationId(), !this.fuseDualViable);
```

alchemist51: Have done this

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DelegatedPredicateCombiner.java
alchemist51: it was done to handle a different condition have removed it

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DelegatedPredicateCombiner.java
alchemist51: I was trying to handle the OrNot Case when we have performance and correctness both, now I have merged them so this is not needed.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DelegatedPredicateCombiner.java
```diff
@@ -122,18 +139,18 @@ private Classified combine(RexCall call, List<Classified> kids, Function<Operato
         }
 
         if (ordered.size() == correctnessChildren.size() && performanceChildren.isEmpty()) {
-            // All children are correctness-delegated to the same backend — bubble up
+            // All children are correctness-delegated to the same backend — bubble up.
             int firstId = correctnessChildren.getFirst().firstAnnotationId();
             return new Delegated(commonBackend, call, firstId, false);
         }
 
-        if (ordered.size() == performanceChildren.size() && correctnessChildren.isEmpty()) {
```
aasom143, line 130: This change is required, performance delegate will not be merged incase of OR condition, else if there is AND i will resolve to a single delegatePossible. Please revert this back.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
```diff
@@ -162,16 +181,25 @@ private static void populatePostDecorationSchemas(Stage stage, CapabilityRegistr
...
             RelDataType expected = inputScan.getRowType();
-            // Cheap int compare first, then digest string compare via equals.
-            if (produced.getFieldCount() == expected.getFieldCount() && produced.equals(expected)) continue;
+            boolean schemaMismatch = produced.getFieldCount() != expected.getFieldCount() || produced.equals(expected) == false;
 
             List<StagePlan> updated = new ArrayList<>(child.getPlanAlternatives().size());
+            boolean changed = false;
             for (StagePlan plan : child.getPlanAlternatives()) {
                 FragmentConvertor convertor = registry.getBackend(plan.backendId()).getFragmentConvertor();
+                boolean producesSubstrait = convertor.producesSubstraitFragments();
```
expani: Substrait is not a Analytics Core concept but is specific to a backend

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FragmentConvertor.java
```diff
@@ -97,4 +97,25 @@ default byte[] attachFragmentOnTop(RelNode fragment, byte[] innerBytes) {
...
     }
+
+    /**
+     * Whether {@link #convertFragment} emits Substrait bytes the reducer can decode for
+     * partition-schema derivation. Default {@code true}; backends with a custom wire format
+     * (e.g. Lucene's count fast path) return {@code false} and MUST also override
+     * {@link #convertSchemaOnlyRead} so the orchestrator emits a Substrait stub at the
+     * partition boundary.
+     */
+    default boolean producesSubstraitFragments() {
```
expani: What is this representing about a backend ? 

Substrait is not a Core concept. It's backend specific

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
```diff
@@ -179,13 +187,16 @@ private void executeInternal(
...
+            new PlannerContext(capabilityRegistry, planningState, indexNameExpressionResolver, false, preferMetadataDriver)
         );
         final String fullPlan = profile ? org.apache.calcite.plan.RelOptUtil.toString(plan) : null;
         QueryDAG dag = DAGBuilder.build(plan, capabilityRegistry, clusterService, indexNameExpressionResolver);
         PlanForker.forkAll(dag, capabilityRegistry);
         BackendPlanAdapter.adaptAll(dag, capabilityRegistry);
-        FragmentConversionDriver.convertAll(dag, capabilityRegistry);
+        // Collapse multi-backend stages to a single chosen alternative before conversion
+        // so the convertor runs once per stage and the wire request carries one PlanAlternative.
+        PlanAlternativeSelector.selectAll(dag, capabilityRegistry, preferMetadataDriver);
```
expani, line 214: This cannot be determined at the Coordinator if the ShardStats are used to make a decision on whether the fast path needs to be selected. 

This should just be set as a Function that runs at the Shard based on the CatalogSnapshot of each backends through some API via `analytics-api`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanAlternativeSelector.java
```diff
@@ -0,0 +1,76 @@
...
+
+        // Keep value-producing alternatives unconditionally; keep metadata-only ones only
+        // when drivable. If a drivable metadata-only survives, collapse to it. If we dropped
+        // any, persist the filtered list so the data-node fallback can't pick a non-drivable.
+        StagePlan drivableMetadataOnly = null;
+        List<StagePlan> survivors = new ArrayList<>(alternatives.size());
+        for (StagePlan plan : alternatives) {
+            AnalyticsSearchBackendPlugin backend = registry.getBackend(plan.backendId());
+            boolean metadataOnly = backend.getCapabilityProvider().isMetadataOnlyDriver();
+            if (metadataOnly && backend.getFragmentConvertor().canDriveFragment(plan.resolvedFragment()) == false) continue;
```
expani: `metadataOnly`
These are also executing the RelNodes and providing results just that internally the use Indices.
So, we don't need to advertise it explicitly ?

It should instead be a ShardLocalDecider that picks backends given certain conditions, for instance deleted docs always prefers Lucene so that requires a higher preference over other one's

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
alchemist51: Updated

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FragmentConvertor.java
alchemist51: Updated

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/DefaultPlanExecutor.java
alchemist51: Tried to make it simple, please check now.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanAlternativeSelector.java
alchemist51: Done

Bukhtawar (approved):
Thanks overall lgtm, maybe we should try extend this to group_by keyword fields/count distinct w and w/o text filter would help for low selectivity/low cardinality use cases(terms.size) can give us a sense.

bharath-techie (approved):
LGTM

PR #21868, opened 2026-05-28
IT: assert lastCommitFileName invariant under concurrent flush/refresh

ask-kamal-nayan:
### What
  
  This PR adds two changes:
  
  1. A regression IT (`CompositeConcurrentIndexingIT#testConcurrentFlushAndRefreshHoldsRefreshLock`) for the `refreshLock` fix in `DataFormatAwareEngine.flush()`.
  2. A bug fix + regression IT for an `indexSort` mismatch in `LuceneCommitter.discoverAndTrimUnsafeCommits` that prevents engine reinit when the safe commit has many Lucene segments.
  
  ---
  
  ### Change 1 — IT for `refreshLock` fix in `DataFormatAwareEngine.flush()`
  
 #### Why
  
  Without holding `refreshLock` across the full flush body, a refresh that fires between `committer.commit()` and `updateLastCommitInfo()` advances `latestCatalogSnapshot` so the
  just-committed snapshot keeps a stale `lastCommitFileName`. The stale name eventually points to a `segments_<N>` the deletion sweep has already removed, and surfaces as peer-recovery
  `NoSuchFileException`.
  
  The existing IT covered concurrent index/refresh/flush correctness but did not assert the on-disk invariant that the fix protects.
  
  #### What the test does
  
  - 4 indexer + 3 flusher + 3 refresher threads, started in lockstep via a `CyclicBarrier`
  - After every flush, the flusher reads back the safe catalog snapshot and asserts
      - `getLastCommitFileName()` is non-null and starts with `segments_`
      - the referenced file is still listed in `store().directory().listAll()`
  - Final `verifyIndex` confirms doc counts and segment-generation invariants
  
  #### Validation
  
  | Configuration | Result |
  |---|---|
  | With `refreshLock` fix | PASS in ~15s |
  | `refreshLock` removed | FAIL: `Committed snapshot's lastCommitFileName must always match an existing segments_<N> on disk` (also surfaces `afterRefresh called but not beforeRefresh`)
  |
  
  Test fails deterministically without the fix and passes with it.
  
  ---
  
  ### Change 2 — Fix `indexSort` mismatch in `LuceneCommitter.discoverAndTrimUnsafeCommits`
  
  #### Why
  
  `LuceneCommitter.discoverAndTrimUnsafeCommits` opens a temp `IndexWriter` in `APPEND` mode at the safe-commit point with no `setMergePolicy` (defaults to `TieredMergePolicy`) and no
  `setIndexSort`. When the safe commit references ≥ ~10 Lucene segments, the temp writer fires a merge during open/close and produces a segment with `source=merge` and `indexSort=null`.
  The new (sorted) `MergeIndexWriter` then fails to open at the polluted commit:
  
  IllegalArgumentException: cannot change previous indexSort=null
      (from segment=_X(...source=merge,...))
      to new indexSort=<sortednumeric: "__row_id__"> ...
      at IndexWriter.validateIndexSort:1225
      at LuceneCommitter.<init>:126
  
  The shard cannot start; the index goes RED.
  
  #### Fix
  
  Pin `NoMergePolicy.INSTANCE` on the temp writer — matches the existing guard vanilla `Store.trimUnsafeCommits` uses (`Store.java:2255`). The temp writer's job is to re-anchor the
  commit point; it must never merge.
  
  #### Test
  
  `DataFormatAwareReplicaResilienceIT#testEngineReinitDoesNotFailWithIndexSortMismatch` fails deterministically without the fix (RED) and passes with it (~60s GREEN).
  ````

### Related Issues
<!-- List any other related issues here -->

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
@@ -379,10 +379,15 @@ public class LuceneCommitter extends SafeBootstrapCommitter {
         }
         // Open a temp IndexWriter at the target commit and re-commit. The default deletion policy
         // (KeepOnlyLastCommitDeletionPolicy) discards all other segments_N files, cleaning up
-        // both unsafe commits and orphan non-CatalogSnapshot commits as well, if any
+        // both unsafe commits and orphan non-CatalogSnapshot commits as well, if any.
+        // Pin the merge policy to NoMergePolicy: this writer's only job is to re-anchor the
+        // commit point. The default TieredMergePolicy would otherwise merge segments without
+        // honoring the engine's index sort, producing an unsorted merged segment that the
+        // subsequent (sorted) MergeIndexWriter cannot open.
         IndexWriterConfig iwc = new IndexWriterConfig().setOpenMode(IndexWriterConfig.OpenMode.APPEND)
             .setCommitOnClose(false)
-            .setIndexCommit(targetCommit);
+            .setIndexCommit(targetCommit)
+            .setMergePolicy(NoMergePolicy.INSTANCE);
         try (IndexWriter tempWriter = new IndexWriter(store.directory(), iwc)) {
             tempWriter.setLiveCommitData(targetCommit.getUserData().entrySet());
             tempWriter.commit();
--- a/sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeConcurrentIndexingIT.java
+++ b/sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeConcurrentIndexingIT.java
@@ -17,13 +17,17 @@ import org.opensearch.arrow.allocator.ArrowBasePlugin;
 import org.opensearch.be.datafusion.DataFusionPlugin;
 import org.opensearch.be.lucene.LucenePlugin;
 import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.common.concurrent.GatedCloseable;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.FeatureFlags;
+import org.opensearch.core.index.shard.ShardId;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.index.engine.CommitStats;
 import org.opensearch.index.engine.exec.Segment;
 import org.opensearch.index.engine.exec.WriterFileSet;
+import org.opensearch.index.engine.exec.coord.CatalogSnapshot;
 import org.opensearch.index.engine.exec.coord.DataformatAwareCatalogSnapshot;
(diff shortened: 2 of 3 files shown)


sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
```diff
@@ -379,10 +379,15 @@ protected void discoverAndTrimUnsafeCommits(Store store, Path translogPath) thro
...
+        // both unsafe commits and orphan non-CatalogSnapshot commits as well, if any.
+        // Pin the merge policy to NoMergePolicy: this writer's only job is to re-anchor the
+        // commit point. The default TieredMergePolicy would otherwise merge segments without
+        // honoring the engine's index sort, producing an unsorted merged segment that the
+        // subsequent (sorted) MergeIndexWriter cannot open.
         IndexWriterConfig iwc = new IndexWriterConfig().setOpenMode(IndexWriterConfig.OpenMode.APPEND)
             .setCommitOnClose(false)
-            .setIndexCommit(targetCommit);
+            .setIndexCommit(targetCommit)
+            .setMergePolicy(NoMergePolicy.INSTANCE);
```
RS146BIJAY, line 390: This will set no merge policy for parent writer. Right? Are we disabling segment merges for parent writer altogether?

ask-kamal-nayan: It's not parent writer , its just a temporary writer that we create during bootstrap?

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeConcurrentIndexingIT.java
```diff
@@ -263,6 +267,147 @@ public void testInterleavedIndexRefreshFlush() throws Exception {
...
+     * {@code lastCommitFileName} that may point to a {@code segments_<N>} the deletion sweep
+     * has already removed — surfacing later as peer-recovery {@code NoSuchFileException}.
+     *
+     * <p>The test pushes high contention with multiple parallel indexer + flusher + refresher
+     * threads. After every flush, each flusher reads back the just-committed snapshot's
+     * {@code lastCommitFileName} and asserts it (a) is non-null and (b) references a Lucene
+     * commit file that actually exists on disk — the direct invariant the {@code refreshLock}
+     * fix protects.
+     */
+    public void testConcurrentFlushAndRefreshHoldsRefreshLock() throws Exception {
```
RS146BIJAY, line 284: Did we validated that without refresh lock this will fail?

RS146BIJAY (approved):

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeConcurrentIndexingIT.java
ask-kamal-nayan: Yes it's validated.

PR #21871, opened 2026-05-28
[analytics-engine] Fix signed/unsigned long overflow on sum() using Neumaier summation

aasom143:
Replace DataFusion's built-in SUM with a custom UDAF that casts all numeric inputs to f64 and uses Neumaier's improved Kahan compensated summation algorithm. This prevents i64 wrapping overflow when summing large values.

Key changes:
- New custom sum UDAF (udaf/sum.rs) returning Float64
- Neumaier compensated summation for cross-batch precision
- Vectorized GroupsAccumulator for GROUP BY queries (zero perf regression)
- Arrow SIMD-optimized intra-batch sum + Neumaier for cross-batch merge

Performed the performance benchmarking and perf numbers are comparable without this change.

### Description

### Related Issues

### Check List


Bukhtawar (commented):
How does coordinator handle this across partial shard results

alchemist51:
Do we not need changes in the Coordinator since the result from the data node is of different type? Can we write a multi node multi shard IT?

sandeshkr419 (commented):
Let's check whether relevant test cases exist in CoordinatorReduceIT,java so that we don't end up breaking this agg.

Thanks!

aasom143:
> How does coordinator handle this across partial shard results

Both executors call udaf register_all(&ctx), so the coordinator session has the same SumUdaf registered. The Final AggregateExec therefore uses same accumulator. When shard partial results arrive, DataFusion calls 
NeumaierSumAccumulator's merge_batch. Adding a IT for the same.

sandeshkr419 (changes requested):
The `AggregateFunction.SUM` enum in `AggregateFunction.java` should declare intermediate fields with Float64 type resolution, similar to how `COUNT` already does, something like:

 ```
 // Current:
  SUM(Type.SIMPLE, SqlKind.SUM),

  // Should be:
  SUM(Type.SIMPLE, SqlKind.SUM, fields(IF("sum", new ArrowType.FloatingPoint(DOUBLE), SUM))),
```

  This would make `DistributedAggregateRewriter.overrideExchangeType()` correctly retype the StageInputScan at plan time, and the Substrait emitted for the reduce plan would declare Float64 in the base_schema from the start. The existing infrastructure already handles this (it does it for `APPROX_COUNT_DISTINCT → VARBINARY`, `COUNT → BIGINT`).

The below rust components should vanish then:
- patch_input_read_schemas
- arrow_schema_to_named_struct

The UDAF `sum.rs` is well-written and the core approach is sound but the `patch_input_read_schemas` hack is the wrong layer to fix this because it compensates at runtime for what should be a compile/plan-time type declaration.

PR #21875, opened 2026-05-28
Refactor distributed-aggregate rewriter

vinaykpud:
### Description

Refactors `DistributedAggregateRewriter` into a transformer pipeline and fixes aggCall slot adjustment in `OpenSearchAggregateSplitRule` when the input schema changes during the aggregate split.

`DistributedAggregateRewriter` was originally introduced in #21645 and #21731 to handle post-Volcano FINAL-aggregate rewrites. As it grew, three loosely-related concerns ended up under a single `rewrite()` method, with construction-time aggCall building reaching across packages.

#21848 took an initial pass at the empty-group failures by skipping the PARTIAL/FINAL split when the aggregate has no `by` clause. As discussed in that PR's comments, the skip was a workaround rather than a fix — distributed parallelism was traded for correctness on every empty-group query, and the underlying row-type-drift issue between SINGLE and FINAL went unaddressed.

This PR separates the two timings (construction-time vs. post-Volcano), introduces a single source of truth for per-call classification, and fixes the row-type drift directly so empty-group queries can use the parallel PARTIAL/FINAL alternative without the skip.

### Changes

**Refactor**
- Consolidate the post-Volcano transformer pipeline (interface + two implementations) and `FinalAggCallBuilder` into `DistributedAggregateRewriter` as nested types and private static methods.
- Move construction-time aggCall building (slot rebase, COUNT→SUM swap, type pinning) into `FinalAggCallBuilder`.
-  Introduce per-call `IntermediateField` classification stored on `OpenSearchAggregate`. Classifying once on the ORIGINAL aggCalls avoids picking the wrong reducer when post-swap calls are re-classified later.
- Add `OpenSearchAggregate.finalAfterRewrite` static factory so the rewriter's return path doesn't repeat the constructor argument list.

**Fix**
- `OpenSearchAggregateSplitRule` now builds FINAL via `FinalAggCallBuilder` with the stashed classification, fixing aggCall slot adjustment when the input schema changes during the aggregate split.
- For empty-group `count()`, FINAL's COUNT→SUM swap returns nullable `BIGINT` while SINGLE keeps `BIGINT NOT NULL` — Volcano rejects the mismatch. Wrap FINAL in a CAST-projection back to SINGLE's row type. Restores PARTIAL/FINAL parallelism for no `group by` queries.

### Tests

- `AggregatePlanShapeTests` — 10 plan-shape assertions across count / sum / avg / sum+count with `by`, plus count empty-group, at 1-shard and 2-shard.
- `DistributedAggregateRewriterTests` 
- End-to-end verified against ClickBench at 1n/1sh, 1n/4sh, and 4n/4sh: 43/43 PPL queries pass; the 5 originally-failing queries (Q10, Q29, Q31, Q32, Q33) green; new empty-group `count()`, `count(), sum(x)`, `count(), avg(x)`, `count(), min(string)` shapes all pass.
- The original failing reproducer and its swap-control both return identical row counts:
  ```
  source = clickbench | stats count(), sum(AdvEngineID) by RegionID   (was failing pre-fix)
  source = clickbench | stats sum(AdvEngineID), count() by RegionID   (was passing; control)
  ```

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/RelNodeUtils.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/RelNodeUtils.java
@@ -82,7 +82,8 @@ public class RelNodeUtils {
                 aggregate.getMode(),
                 aggregate.getViableBackends(),
                 aggregate.getCallAnnotations(),
-                aggregate.getFinalExtraLiteralArgs()
+                aggregate.getFinalExtraLiteralArgs(),
+                aggregate.getIntermediateFields()
             );
         } else if (node instanceof OpenSearchSort sort) {
             return new OpenSearchSort(
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DistributedAggregateRewriter.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DistributedAggregateRewriter.java
@@ -31,21 +31,14 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * Post-Volcano orchestrator: adapts a FINAL {@link OpenSearchAggregate}'s input chain (re-types
- * the StageInputScan when a typeResolver dictates a different shape, inserts a literal-Project
- * for STATE_EXPANDING aggregates) and rebinds its aggCalls via the nested
- * {@link FinalAggCallBuilder} when anything changed.
+ * Adapts a FINAL {@link OpenSearchAggregate}'s input chain after the DAG is fragmented:
+ * re-types the StageInputScan, inserts a literal-Project for STATE_EXPANDING aggregates, and
+ * rebinds aggCalls via {@link FinalAggCallBuilder} when anything changed. Called from
+ * {@code BackendPlanAdapter.adaptNode}.
  *
- * <p>Runs after Volcano CBO and DAG fragmentation, called from
- * {@code BackendPlanAdapter.adaptNode} once per FINAL aggregate per backend. At that point
- * FINAL's input chain ends at an {@link OpenSearchStageInputScan} placeholder — PARTIAL lives
- * in a sibling stage.
- *
- * <p>TODO: {@link FinalAggCallBuilder} also runs at split-rule time (called from
- * {@code OpenSearchAggregateSplitRule} in {@code planner.rules}), so its current home here
- * means {@code planner.rules} reaches into {@code planner.dag} — backwards from the usual
- * layered direction. Move it to a top-level class in {@code planner.rel} once a third
- * construction-time consumer appears.
+ * <p>TODO: {@link FinalAggCallBuilder} is also used by the split rule, so {@code planner.rules}
+ * reaches into {@code planner.dag} — backwards from the usual layering. Move it to
+ * {@code planner.rel} once a third construction-time consumer appears.
  *
  * @opensearch.internal
  */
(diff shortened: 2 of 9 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FinalInputTransformer.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * One step in the post-Volcano transformation of a FINAL {@link OpenSearchAggregate}'s input
+ * chain. Transformers are composed by {@code DistributedAggregateRewriter} — each must return
+ * {@code current} unchanged when it has nothing to do, so the orchestrator can skip the
+ * aggCall rebuild on a no-op pass.
+ *
+ * @opensearch.internal
+ */
+interface FinalInputTransformer {
```
expani: Can we keep all these as static inner classes in the Rewriter ? 

I don't think handling partial aggs should require all such transformations if we had moved DAG Builder to the end of the planner pipeline.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/FinalAggCallBuilder.java
```diff
@@ -0,0 +1,151 @@
...
+ * {@code groupCount + i}, applies the COUNT→SUM-style reducer swap, and pins return types
+ * for engine-native merges. Pure helper — no rel-tree walking.
+ *
+ * <p>Called by the split rule (so Volcano's {@code typeMatchesInferred} passes) and again by
+ * {@code DistributedAggregateRewriter} when post-Volcano input adapters change the chain.
+ * Idempotent across both calls as long as the same {@link #classify} result is reused.
+ *
+ * @opensearch.internal
+ */
+public final class FinalAggCallBuilder {
```
expani: So, the complexity moved out of rewriter into this class and the 2 transformers ? 

IMO Let's still keep it contained in the same class even if we are distinguishing as separate classes/methods

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/CallDecomposition.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Per-call decomposition decision produced by {@link FinalAggCallBuilder#classify} and read
+ * by post-Volcano transformers. Either carries an {@link IntermediateField} describing how
+ * PARTIAL emits state and how FINAL merges it, or is the {@link #PASS_THROUGH} marker for
+ * aggregates with no decomposition (FINAL re-uses the original function).
+ *
+ * @opensearch.internal
+ */
+public record CallDecomposition(IntermediateField field) {
```
expani: I had kept AggregateDecomposition in the past AND then it was removed later in a change by @sandeshkr419 

We are introducing this again so at least we should capture the Reduce Expression of the aggregate with it.

vinaykpud: Added it so that handling null was confusing in the if conditions, so `PASS_THROUGH` gives a self documentation. 
But i see your concern on `DELAY`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rel/FinalAggCallBuilder.java
vinaykpud: sure

vinaykpud: Reason why this logic was moved out because it was used in OpenSearchAggregateSplitRule.java, now i am moving this class inside the `DistributedAggregateRewriter.java`

mch2 (approved):

PR #21878, opened 2026-05-28
[Analytics Engine] Make PPL chart work on the analytics-engine route (partition-boundary coercion + non-prefix agg-split type fix)

ahkcs:
### Description

Two type-handling fixes on the analytics-engine route, both surfaced by PPL `chart` (force-routed through the analytics-engine path). No SQL-plugin change — chart's lowering is already merged and unchanged.

#### 1. Coerce streaming-partition schema & batches at the reduce-stage boundary (DataFusion backend)

A producer fragment emits batches in DataFusion's raw physical Arrow types — `UInt64` (e.g. `row_number()`), `BinaryView` (view-typed columns), `Float16` (`half_float`) — but the consumer fragment's Substrait plan declares the coerced forms (`Int64` / `Binary` / `Float32`) that `coerce_inferred_schema` already applies to scan-inferred schemas. `LocalSession::register_partition` registered the reduce-stage `StreamingTable` under the **raw** producer schema, so when a column of one of those types crosses the reduce exchange the consumer plan can fail to bind:

```
Substrait error: Field 'X' in Substrait schema has a different type (Int64) than the corresponding field in the table schema (UInt64).
```

Register the `StreamingTable` under the coerced schema and coerce each incoming batch (new `coerce_record_batch` + `CoercingReceiverPartition`). The channel/sender keep the un-coerced schema so the Java-side tripwire still validates producer batches; `coerce_inferred_schema` returns the same `Arc` when nothing needs rewriting, so plans without these types keep the cheaper passthrough partition (no behavior change).

#### 2. Skip partial/final split on non-prefix groupSet type mismatch — exact `SqlTypeName` (analytics-engine planner)

`OpenSearchAggregateSplitRule.shouldSkipPartialFinalSplit()` skips the split for non-prefix group sets, where a group-key ordinal `k >= groupCount` lands on a PARTIAL agg-output slot — the structural FINAL reuses original ordinals over the keys-first PARTIAL output, so the group key takes the agg-output type, which must equal the original input-field type or Volcano's `typeMatchesInferred` / row-type-equivalence check throws. The collision check compared `SqlTypeFamily`, which is too coarse: `BIGINT` and `INTEGER` share the `NUMERIC` family, so a span group key (`INTEGER`) over a measure declared `BIGINT` could slip through and planning crash with:

```
AssertionError: type mismatch: aggCall type: BIGINT inferred type: INTEGER
IllegalArgumentException: Type mismatch: rel rowtype ... $f2: BIGINT -> INTEGER
```

Compare the exact `SqlTypeName`, so such cases fall back to a SINGLE coordinator aggregate (always correct). Prefix group sets (the common case) are unaffected — they early-return before this check.

Both are surfaced by `chart ... by <field> span=...` over an `INTEGER` measure and by chart's ranked `ROW_NUMBER() OVER (...)` subplan crossing a join exchange; the equivalent `stats ... by <field> span=...` and any `unsigned_long` / `binary` / `float16` column crossing a reduce stage hit the same paths.

### Tests

New Rust unit tests: `coerce_record_batch_rewrites_uint64_column`, `coerce_record_batch_passthrough_when_already_matching`.

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/local_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/local_executor.rs
@@ -43,7 +43,9 @@ use datafusion_substrait::logical_plan::consumer::from_substrait_plan;
 use prost::Message;
 use substrait::proto::Plan;
 
-use crate::partition_stream::{channel, PartitionStreamSender, SingleReceiverPartition};
+use crate::partition_stream::{
+    channel, CoercingReceiverPartition, PartitionStreamSender, SingleReceiverPartition,
+};
 
 /// Coordinator-reduce DataFusion session.
 ///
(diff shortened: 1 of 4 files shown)


mch2:
This coercion should not be required once https://github.com/opensearch-project/OpenSearch/pull/21775 lands cc @sandeshkr419

ahkcs:
Closing as no longer needed. The type-handling issues these two fixes targeted on the analytics-engine route (the reduce-stage `row_number()` Int64/UInt64 mismatch and the non-prefix agg-split BIGINT/INTEGER crash) are resolved on current `main` — `CalciteChartCommandIT` force-routed through the analytics-engine path is at its in-scope ceiling (the only remaining failures are the datetime wire-format and the `attributes.client.ip` Binary/BinaryView scan mismatch, both tracked separately). Verified against current `main` with and without this PR (identical result), so it adds no measurable change.

PR #21882, opened 2026-05-29
Fix for text functions with non-text filter condition

nssuresh2007:
### Description
Reject text-relevance functions (`query_string`, `simple_query_string`, `multi_match`) when applied to non-text/non-keyword fields, with a clear, actionable error at planning time — covering both explicitly-named fields and fields referenced *inside* the query string.

#### Problem
Text-relevance functions only have their Lucene capability declared for `text` and `keyword` types, but the planner was not validating field types before routing. For example, `query_string(['severityNumber'], 'severityNumber:>15')` on a `long` field was routed to Lucene and produced a generic "no backend can evaluate" error downstream (or a
silently wrong result). The problem applied to several shapes:
- explicit fields: `query_string(['severityNumber'], '...')`
- mixed explicit fields: `query_string(['body','severityNumber'], '...')`
- fields named inside the query body: `query_string(['body'], 'severityNumber:15')`
- the field-less `search` form, which lowers to a query-only `query_string`:
  `search source=idx severityNumber=15` → `query_string(MAP('query','severityNumber:15'))`

#### Fix
- Added `TextRelevanceFieldValidator`, wired into `OpenSearchFilterRule`: for any text-relevance predicate, every resolved **literal** field must be `text`/`keyword`, otherwise planning fails with `IllegalArgumentException` naming the function, field, and mapping type, plus a hint to use a typed comparison.
```
Text-relevance function [query_string] cannot be applied to field [severityNumber] of type [long]. Only text and keyword fields are supported. Use a typed comparison (e.g. [severityNumber > value]) for non-text fields.
```

- Introduced a `FieldReferenceExtractor` SPI in `analytics-framework`, implemented by the Lucene backend (`LuceneFieldReferenceExtractor`) and resolved via a capability registry. For `query_string` it parses the query body with Lucene's classic grammar (no-op analyzer, no `QueryShardContext` required) to surface inline field references, including `_exists_:field`. Extraction reuses the same operand path as execution so plan-time and
runtime classification cannot drift.
- Literal vs. pattern classification uses `Regex.isSimpleMatchPattern` (the same predicate the field resolver uses). Only **literal** fields are validated; **pattern** tokens (e.g. an escaped `severity\*:15`, or a wildcard in the `fields` list like `body*`) pass through and are expanded at execution.
- `lenient=true` opts out of the eager rejection; the failure mode then matches the existing backend-selection behavior rather than the friendly message.
- In-string parsing is best-effort: on any parse failure the extractor skips inline extraction and falls back to the explicit `fields` operand (the authoritative parse still happens at execution). This also fixes a regression where a field-less `query_string('category:A')` with empty literal field names was wrongly rejected — it now falls back to TEXT instead of throwing.

#### Tests
- Unit: `LuceneFieldReferenceExtractorTests`, `FilterRuleTests` (literal/pattern/inline extraction, lenient, mixed fields).
- Integration (`TextRelevanceValidationIT`, real PPL over `otel_logs`): explicit numeric reject (`source=` and `search` forms), text-field success, mixed text+numeric reject, `simple_query_string`/`multi_match` numeric reject, in-string literal reject, in-string regex-*value* reject (`severityNumber:/1[0-9]/`), escaped in-string wildcard passthrough (text-only `serviceName\*` and numeric-spanning `severity\*`), field-less literal numeric reject via the `search` command, and the `lenient=true` routing-failure case.

#### Out of scope
- Splitting mixed-type predicates across Lucene + DataFusion. For example, `query_string(['title','severity'], 'login AND severity:>15')` could in principle be split into a Lucene clause and a DataFusion clause, but the rewrite is non-trivial (OR/NOT, score preservation, default-field semantics), so this PR rejects the predicate instead.
- Unescaped wildcard-in-field-name inside the query body (`severity*:15`) is invalid Lucene `query_string` syntax — the classic grammar accepts only a plain `TERM` or a bare `*` before `:`, so it fails to parse at execution. The escaped form `severity\*:15` is the supported syntax.

### Check List


mch2:
@nssuresh2007 can you check/please add if not an IT for this case ls in /sandbox/qa

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -163,9 +164,42 @@ private List<String> resolveViableBackends(
...
             // Multi-field full-text functions (multi_match, query_string, simple_query_string)
             // encode field names as string literals in nested MAPs rather than RexInputRef.
-            // Resolve viability against any backend that supports the function on text fields.
+            // Extract the literal field names and resolve viability per-field using the actual
+            // FieldStorageInfo lookup — same code path as RexInputRef-based fields. This ensures
+            // that, e.g., query_string(['severityNumber'], ...) on an INTEGER field doesn't get
+            // routed to a backend that only declared (QUERY_STRING, TEXT) capability.
             if (function.getCategory() == ScalarFunction.Category.FULL_TEXT) {
-                return new ArrayList<>(registry.filterBackendsAnyFormat(function, FieldType.TEXT));
+                List<String> literalFieldNames = extractLiteralFieldNames(predicate);
```
expani: We can extract this entire logic in a separate file itself and call it from here.

This class seems to be growing bigger.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -234,6 +268,113 @@ private void collectFieldIndices(RexNode node, Set<Integer> result) {
...
+     *     MAP('fields', MAP('field1':VARCHAR, boost1:DOUBLE, 'field2':VARCHAR, boost2:DOUBLE, ...)),
+     *     MAP('query', '...':VARCHAR)
+     *   )
+     * </pre>
+     * This method walks the call tree to find all string literals that represent field names —
+     * specifically, RexLiteral children at even indices of a nested MAP whose outer key is "fields"
+     * or "field". Returns an empty list if no such names are found (e.g. zero-field functions
+     * like MATCHALL or QUERY without an explicit field list).
+     */
+    private List<String> extractLiteralFieldNames(RexCall predicate) {
```
expani: Won't this be dependent on the type of function ? 

Should we look at tying it to ScalarFunction ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -234,6 +268,113 @@ private void collectFieldIndices(RexNode node, Set<Integer> result) {
...
+     * <p>Fields whose storage info cannot be resolved (unknown columns, dynamic
+     * mappings) are skipped — they fall through to the existing capability-based
+     * matching where they get the conservative TEXT-type assumption.
+     */
+    private void rejectNonTextFieldsForTextFunction(
+        String functionName,
+        List<String> fieldNames,
+        List<FieldStorageInfo> fieldStorageInfos
+    ) {
+        Set<FieldType> allowed = new HashSet<>();
```
expani: Can be a constant

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
nssuresh2007: Done

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
nssuresh2007: Done

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
nssuresh2007: Yes, have added check for the specific text functions for which we should use this flow.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/TextRelevanceValidationIT.java
```diff
@@ -0,0 +1,93 @@
...
+            dataProvisioned = true;
+        }
+    }
+
+    /**
+     * {@code query_string(['severityNumber'], ...)} on a {@code long} field is rejected at
+     * planning. The error body names the offending field and the supported types.
+     */
+    public void testQueryStringOnNumericFieldIsRejected() {
+        String ppl = "source=" + DATASET.indexName + " | where query_string(['severityNumber'], 'severityNumber:>15')";
```
expani, line 52: Can we add some cases using the `Search` command of PPL ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -165,8 +165,51 @@ private List<String> resolveViableBackends(
...
-            // Resolve viability against any backend that supports the function on text fields.
+            // Extract the literal field names and resolve viability per-field using the actual
+            // FieldStorageInfo lookup — same code path as RexInputRef-based fields. This ensures
+            // that, e.g., query_string(['severityNumber'], ...) on an INTEGER field doesn't get
+            // routed to a backend that only declared (QUERY_STRING, TEXT) capability.
             if (function.getCategory() == ScalarFunction.Category.FULL_TEXT) {
+                if (TextRelevanceFieldValidator.usesLiteralFieldEncoding(function)) {
+                    List<String> literalFieldNames = TextRelevanceFieldValidator.extractLiteralFieldNames(predicate);
+                    if (literalFieldNames.isEmpty()) {
+                        // No field references at all — fall back to TEXT type assumption.
```
expani: Which case do you see for this ? 
We should rather throw exceptions

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/TextRelevanceFieldValidator.java
```diff
@@ -0,0 +1,189 @@
...
+     *     MAP('fields', MAP('field1':VARCHAR, boost1:DOUBLE, 'field2':VARCHAR, boost2:DOUBLE, ...)),
+     *     MAP('query', '...':VARCHAR)
+     *   )
+     * </pre>
+     * This method walks the call tree to find all string literals that represent field names —
+     * specifically, RexLiteral children at even indices of a nested MAP whose outer key is "fields"
+     * or "field". Returns an empty list if no such names are found (e.g. zero-field functions
+     * like MATCHALL or QUERY without an explicit field list).
+     */
+    public static List<String> extractLiteralFieldNames(RexCall predicate) {
```
expani: Let's add a TODO to push this extraction to the FullText function.

Similar to how you handled extraction in QuerySerializers

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/TextRelevanceValidationIT.java
nssuresh2007: Done

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
nssuresh2007: Done

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/TextRelevanceFieldValidator.java
nssuresh2007: Added TODO

expani (approved):

mch2 (approved):

mch2 (approved):

sandeshkr419:
@nssuresh2007 Resolve conflicts please. Thanks for addressing comments!

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
nssuresh2007: Revisiting this comment, since we identified scenario where queries like below would not have fields present but still it would be valid query:

`search source=test_data category=A | fields category, value | head 5`

Hence, I would like to revert this to the older format, where we just fallback to Lucene backend for such scenarios. One downside is that, if a query uses such syntax but specifies a field other than text/keyword then we will not be able to identify it still but assuming that would be a rare case.
Please let me know if you see any concern.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -165,8 +165,57 @@ private List<String> resolveViableBackends(
...
+                    for (String fieldName : literalFieldNames) {
+                        FieldStorageInfo storageInfo = null;
+                        for (FieldStorageInfo info : fieldStorageInfos) {
+                            if (fieldName.equals(info.getFieldName())) {
+                                storageInfo = info;
+                                break;
+                            }
+                        }
+                        if (storageInfo == null) {
+                            // Unknown field — fall back to TEXT type assumption for this field.
```
expani: Why do we need this assumption ? What happens if we fail with an explicit error ?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldReferenceExtractor.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Returns the fields a relevance predicate references.
+     *
+     * @param call         the relevance {@link RexCall} (e.g. {@code QUERY_STRING(MAP('fields', ...), MAP('query', ...))})
+     * @param fieldStorage per-column storage metadata; {@link org.apache.calcite.rex.RexInputRef}
+     *                     indices in {@code call} index into this list
+     * @return the referenced fields split into validated literals vs. passed-through patterns,
+     *         plus fan-out and lenient metadata
+     */
+    FieldReferences referencedFields(RexCall call, List<FieldStorageInfo> fieldStorage);
```
expani: Can't this be a method inside the Serializer itself ?

nssuresh2007: No. Because this is the SPI through which planner interacts with the backend. Planner needs to know the list of fields referenced in the query, based on which it decides if possible execute query or throws exception.
Please let me know if that makes sense.

expani: Isn't `DelegatedPredicateSerializer` already an existing SPI for the same ? 

I am thinking if a backend is serializing a predicate, it can also tell the fields that were involved in the predicate if it's opaque to the Planner. ( In this case query string and match are those cases ) 

Other serializers can default to not implemented or something equivalent ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/CapabilityRegistry.java
```diff
@@ -73,6 +74,10 @@ public class CapabilityRegistry {
     private final Map<DelegationType, List<String>> delegationAcceptors = new HashMap<>();
     private final Map<FullTextParamKey, Set<String>> fullTextParamIndex = new HashMap<>();
 
+    // Per-function field-reference extractors, aggregated across backends (first declaration wins).
+    // Full-text is Lucene-only today, so a single function -> extractor map suffices.
+    private final Map<ScalarFunction, FieldReferenceExtractor> fieldReferenceExtractors = new HashMap<>();
```
expani: This will also not be required if it's part of the Serializer as a new method

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldReferenceExtractor.java
nssuresh2007: Updated to use new method in existing SPI itself as suggested

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/CapabilityRegistry.java
nssuresh2007: Yes this is removed now

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/serializers/RelevanceFieldExtractor.java
```diff
@@ -0,0 +1,174 @@
...
+     * field and are not emitted as field references.
+     */
+    private static void collectInStringFields(String functionName, String queryString, Set<String> tokens) {
+        QueryParser parser = new QueryParser(DEFAULT_FIELD_SENTINEL, Lucene.KEYWORD_ANALYZER);
+        parser.setAllowLeadingWildcard(true);
+        Query query;
+        try {
+            query = parser.parse(queryString);
+        } catch (ParseException | RuntimeException e) {
+            LOGGER.debug(
```
expani: Why is this debug ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
expani: Can we remove this assumption and see what fails ?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldReferences.java
```diff
@@ -0,0 +1,50 @@
...
+ * fan-out for unqualified terms) is resolved at execution by the data node, which matches
+ * OpenSearch's best-effort, never-erroring treatment of those forms.
+ *
+ * @param literalFields explicitly-named concrete field names (from the {@code fields}/{@code field}
+ *                      operand and, for {@code query_string}, literal {@code field:} tokens in the
+ *                      query string). Validated by the planner. First-appearance ordered.
+ * @param patternTokens explicitly-named wildcard/regex tokens (e.g. {@code cat*}, {@code *}),
+ *                      classified via {@code Regex.isSimpleMatchPattern}. Passed through — not
+ *                      expanded, not rejected.
+ * @param lenient       effective lenient flag: the explicit {@code lenient} param when set,
```
expani, line 34: Who sets it ? The end user as an execution hint via the SQL/PPL plugin ?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/serializers/RelevanceFieldExtractor.java
expani: Also, if it fails here shouldn't we fail the query instead of letting it happen downstream ? 

Can you explain the cases where this is required taking a Raw PPL query as an example ?

expani (commented):
Thanks for the followups @nssuresh2007 

Dropping some comments around clarification

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldReferences.java
nssuresh2007: Lenient is a parameter present in `simple_query_string`, `query_string, `match` and `multi_match` queries passed by the customer.

https://docs.opensearch.org/latest/query-dsl/full-text/simple-query-string/
https://docs.opensearch.org/latest/query-dsl/full-text/match/

This is also supported by PPL: https://github.com/opensearch-project/sql/blob/main/docs/user/ppl/functions/relevance.md

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/serializers/RelevanceFieldExtractor.java
nssuresh2007: > Why is this debug ?

Thanks for pointing out, it was a miss. Updated it to warn.

> Can you explain the cases where this is required taking a Raw PPL query as an example ?

We are using Lucene's QueryParser which still does not have some extensions that OpenSearch's query parser does at execution time. Due to this reason we swallow the exception since there is a possibility that the parser failed for a valid query still. Following is an example:

Query parser successfully extracts the field severityNumber and validates against it:
```
curl -s -w "\nHTTP %{http_code}\n" -X POST "http://localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d '{"query": "source=otel_logs | where query_string([\"body\"], \"severityNumber:15 OR GET~2\")"}'
{
  "error": {
    "reason": "Invalid Query",
    "details": "Text-relevance function [query_string] cannot be applied to field [severityNumber] of type [long]. Only text and keyword fields are supported. Use a typed comparison (e.g. 
[severityNumber > value]) for non-text fields.",                                                                                                                                                "type": "IllegalArgumentException"
  },
  "status": 400
}
HTTP 400
```

Query parser fails recognizing `~2.5` even though it is a valid query. Since we swallow the exception, it skips this validation and the actual query executes to return 0 results.

```
curl -s -w "\nHTTP %{http_code}\n" -X POST "http://localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d '{"query": "source=otel_logs | where query_string([\"body\"], \"severityNumber:15 OR GET~2.5\") | stats count() as c"}'
{
  "schema": [
    {
      "name": "c",
      "type": "bigint"
    }
  ],
  "datarows": [
    [
      0
    ]
  ],
  "total": 1,
  "size": 1
}
HTTP 200
```


**Another example:**

Query parser successfully parses and throws error at planning phase itself:

```
curl -s -w "\nHTTP %{http_code}\n" -X POST "http://localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d '{"query": "source=otel_logs | where query_string([\"body\"], \"severityNumber:15~2\")"}' 2>&1 | head -c 1000

{
  "error": {
    "reason": "Invalid Query",
    "details": "Text-relevance function [query_string] cannot be applied to field [severityNumber] of type [long]. Only text and keyword fields are supported. Use a typed comparison (e.g. 
[severityNumber > value]) for non-text fields.",                                                                                                                                                "type": "IllegalArgumentException"
  },
  "status": 400
}
HTTP 400
```

Query parser fails to parse, swallows exception and allows runtime execution to handle it and we get error from runtime:

```
curl -s -w "\nHTTP %{http_code}\n" -X POST "http://localhost:9200/_plugins/_ppl" -H 'Content-Type: application/json' -d '{"query": "source=otel_logs | where query_string([\"body\"], \"severityNumber:15~2.5\") | stats count() as c"}' 2>&1 | head -c 800

{
  "error": {
    "reason": "Error occurred in OpenSearch engine: Can only use fuzzy queries on keyword and text fields - not on [severityNumber] which is of type [long]",
    "details": "StreamException[errorCode=INTERNAL, message=Can only use fuzzy queries on keyword and text fields - not on [severityNumber] which is of type [long], metadata={opensearch.co
rrelation-id=[1868629307266433050]}]\nFor more details, please send request for Json format to see the raw response from OpenSearch engine.",  "type": "StreamException"
  },
  "status": 500
}
HTTP 500
```

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
nssuresh2007: Agreed. This was a more safer check added for wildcard in field names, but that will already get handled at  Line#194 (Empty literalFieldNames condition) and hence throwing exception when this happens. Updated now

nssuresh2007: > Can we remove this assumption and see what fails ?

As of now, no UT or integ test is failing when this condition throws Exception

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/serializers/RelevanceFieldExtractor.java
expani: >We are using Lucene's QueryParser which still does not have some extensions that OpenSearch's query parser does at execution time

Then we should look at using OpenSearch's query parser. If it's simple enough let's do in this PR or add a TODO to follow-up in another PR

expani (approved):

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/serializers/RelevanceFieldExtractor.java
nssuresh2007: I evaluated that, but that was not possible since `QueryStringQueryParser` requires `QueryShardContext` which is available only at execution time at shard level. Due to this reason it was not possible to use it in the planning layer.

expani: Got it, although not ideal we can mock a QueryShardContext since it's only looking for the MapperService which should be available at the Coordinator. 

See https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneAnalyticsBackendPlugin.java#L265-L284 where we mock in Lucene Backend Plugin to access Mappers.

mch2 (approved):

PR #21888, opened 2026-05-29
[Analytics Engine] Skip partial/final split on non-prefix groupSet type mismatch (exact SqlTypeName)

ahkcs:
### Description

`OpenSearchAggregateSplitRule.shouldSkipPartialFinalSplit()` skips the PARTIAL/FINAL split for **non-prefix** group sets — those where a group-key ordinal `k >= groupCount` lands on a PARTIAL agg-output slot. The structural FINAL reuses the original ordinals over the keys-first PARTIAL output, so such a group key would take the **agg-output column's type**, which must equal the original input-field type or Volcano's `Aggregate.typeMatchesInferred` / row-type-equivalence check throws.

The collision check compared `SqlTypeFamily`, which is too coarse: `BIGINT` and `INTEGER` both belong to the `NUMERIC` family, so a span group key (`INTEGER`) paired with a measure declared `BIGINT` slipped through, the split was attempted, and planning crashed with either:

```
AssertionError: type mismatch: aggCall type: BIGINT inferred type: INTEGER
```
or
```
IllegalArgumentException: Type mismatch: rel rowtype ... $f2: BIGINT -> INTEGER
```

### Fix

Compare the exact `SqlTypeName` instead of the type family, so these cases fall back to a SINGLE coordinator aggregate (always correct) rather than an unsound partial/final split. One-line semantics change; no behavior change for prefix group sets (handled by the earlier early-return) or for non-prefix sets whose collision-slot types already match.

### How it surfaced / verification

Surfaced by PPL `chart ... by <field> span=...` over an `INTEGER` measure on the analytics-engine route (e.g. `chart max(balance) by age span=10`, `chart usenull=... avg(balance) over gender by age span=10`). The equivalent `stats ... by <field> span=...` shape hits the same path.

Verified against `CalciteChartCommandIT` (opensearch-project/sql) force-routed through the analytics-engine path. The two span tests that crashed on planning go red → green:

| Test (`CalciteChartCommandIT`) | Before | After |
|---|---|---|
| `testChartMaxBalanceByAgeSpan` | ❌ `aggCall type: BIGINT inferred type: INTEGER` | ✅ |
| `testChartUseNullTrueWithNullStr` | ❌ `Type mismatch … $f2: BIGINT -> INTEGER` | ✅ |

Combined with the partition-boundary coercion fix (#21878), chart on the analytics-engine route goes from 8/15 → 10/15; the remaining 5 are out of scope here (datetime wire-format sql#5420 ×2, and the `attributes.client.ip` Binary/BinaryView scan mismatch ×3).

### Follow-up

A self-contained QA IT under `sandbox/qa/analytics-engine-rest` exercising `stats max(<int>) by <field> span=N` would regression-cover this in OpenSearch CI (no existing Span/Stats QA IT covers the `stats max(int-measure) by … span` non-prefix shape). Not included here.

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
@@ -20,7 +20,7 @@ import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.sql.SqlAggFunction;
-import org.apache.calcite.sql.type.SqlTypeFamily;
+import org.apache.calcite.sql.type.SqlTypeName;
 import org.apache.calcite.util.ImmutableBitSet;
 import org.opensearch.analytics.planner.PlannerContext;
 import org.opensearch.analytics.planner.RelNodeUtils;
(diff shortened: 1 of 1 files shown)


ahkcs:
Folded into #21878 — combining the partition-boundary coercion and the non-prefix agg-split type fix into a single PR, since together they're what makes PPL chart work end-to-end on the analytics-engine route.

PR #21889, opened 2026-05-29
Derived source. Do not initialise parent object during _source reconstruction when document does not have field with some value

### LINKED ISSUE #21886: [BUG] [Derived source] ObjectMapper adds empty objects {} into reconstructed _source
### Describe the bug

**Description**
During __source_ reconstruction, _org.opensearch.index.mapper.ObjectMapper#deriveSource_ iterates over the defined index mappings to reconstruct the original document on-the-fly. However, if an index defines nested or deep object mappings (e.g., _metadata.user.id_) but a specific document does not contain any values for those fields, the reconstruction logic still initializes and appends the parent JSON nodes.

**Expected Behavior**
If a document does not contain a field with value, the reconstructed __source_ should completely omit them, accurately reflecting the original ingested document.

**Actual Behavior**
The reconstructed __source_ includes empty JSON object for every missing object field defined in the schema.

**Impact**
This behavior causes severe payload bloat and performance regressions for users with large, sparse schemas (hundreds of fields where only a few are populated per document). It also alters the semantic structure of the document, which can break downstream application logic that checks for the existence of parent keys.

### Related component

Search

### To Reproduce

```bash
curl -X DELETE localhost:9200/derived-source-test-index

curl -X PUT localhost:9200/derived-source-test-index -H 'Content-type: application/json' -d '{"settings":{"index":{"derived_source":{"enabled":true}}}}'

curl -X PUT localhost:9200/derived-source-test-index/_doc/1 -H 'Content-type: application/json' -d '{"timestamp":"2026-05-27T22:12:01.000Z"}'
curl -X PUT localhost:9200/derived-source-test-index/_doc/2 -H 'Content-type: application/json' -d '{"timestamp":"2026-05-27T22:12:02.000Z", "message": "some-text"}'
curl -X PUT localhost:9200/derived-source-test-index/_doc/3 -H 'Content-type: application/json' -d '{"timestamp":"2026-05-27T22:12:03.000Z", "some":{"very":{"long":{"path":{"field_1":1,"field_2":true,"field_3":"keyword"}}}}}'
curl -X PUT localhost:9200/derived-source-test-index/_doc/4 -H 'Content-type: application/json' -d '{"timestamp":"2026-05-27T22:12:04.000Z", "another":{"very":{"long":{"path":{"field_1":1,"field_2":true,"field_3":"keyword"}}}}}'

curl -X POST localhost:9200/derived-source-test-index/_refresh
clear

curl localhost:9200/derived-source-test-index/_search?pretty
```

### Expected behavior

```bash
curl localhost:9200/derived-source-test-index/_doc/1?pretty
```
the `_source` should have only `timestamp` field, without empty `some.very.long.path = {}` and `another.very.long.path = {}`

### Additional Details

This issue was introduced by design, in `org.opensearch.index.mapper.ObjectMapper#deriveSource` method.

Tested on versions 3.4 and 3.6

---

I have already developed a fix for this issue locally.
However, since this is my first time submitting an issue to OpenSearch, I wanted to open this tracking issue first and verify with the maintainers that this is indeed the agreed-upon intended behavior before opening the Pull Request. 

Please let me know if you would like me to submit the PR for review. Sorry for any inconvenience, and looking forward to contributing!

OVyshnevskyi:
### Description
Lazy parent object initialisation in `org.opensearch.index.mapper.ObjectMapper`, that prevents empty objects during _source reconstruction.

### Related Issues
Resolves #21886

### Check List

### DIFF
--- a/libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java
+++ b/libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java
@@ -64,7 +64,7 @@ import java.util.function.Function;
  * A utility to build XContent (ie json).
  */
 @PublicApi(since = "1.0.0")
-public final class XContentBuilder implements Closeable, Flushable {
+public class XContentBuilder implements Closeable, Flushable {
 
     /**
      * Create a new {@link XContentBuilder} using the given {@link XContent} content.
@@ -263,7 +263,7 @@ public final class XContentBuilder implements Closeable, Flushable {
      * @param os       the output stream
      * @param includes the inclusive filters: only fields and objects that match the inclusive filters will be written to the output.
      * @param excludes the exclusive filters: only fields and objects that don't match the exclusive filters will be written to the output.
-     * @param parent references the parent this instance was copied from
+     * @param parent if non-null, this builder shares the parent's underlying generator and write context
      * @param prettyPrint use pretty printer
      */
     private XContentBuilder(
@@ -279,6 +279,13 @@ public final class XContentBuilder implements Closeable, Flushable {
         this.includes = includes;
         this.excludes = excludes;
         this.prettyPrint = prettyPrint;
+        if (parent != null) {
+            // Share the parent's underlying generator so that all write operations act on the same
+            // generator and write context. Used by LazyXContentBuilder so that nested lazy builders
+            // do not each spin up their own generator on the shared stream (which would start in the
+            // root context and fail with "Current context not Object but root" when closing objects).
+            this.generator = parent.generatorInstance();
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


libs/core/src/main/java/org/opensearch/core/xcontent/XContentBuilder.java
```diff
@@ -1173,4 +1173,31 @@ private static void ensureNoSelfReferences(final Iterable<?> value, Object origi
         }
     }
 
+    public static final class LazyXContentBuilder extends XContentBuilder {
```
OVyshnevskyi, line 1187: This is the simplest possible and minimalistic solution i was able to produce, but a bit dirty.
Likely better option here would be to introduce an interface for `XContentBuilder` and avoid direct inheritance ( too big change to pull it out from start in first PR without discussion :) )

server/src/test/java/org/opensearch/index/mapper/ObjectMapperTests.java
```diff
@@ -603,6 +612,94 @@ public void testNestedIsParent() throws Exception {
...
+            .startObject("properties")
+            .startObject("numeric_field")
+            .field("type", "long")
+            .endObject()
+            .startObject("field_1")
+            .startObject("properties")
+            .startObject("field_2")
+            .startObject("properties")
+            .startObject("keyword_field")
+            .field("type", "long")
```
sandeshkr419, line 650: A miss in test? `keyword_field` as `long`?

sandeshkr419 (commented):
Thanks @OVyshnevskyi for working on this. A few test gaps I'd suggest adding:

- Test more field types, not just numbers. Adding keyword, text, and boolean fields would exercise the different value-writing paths.
- Case for deeply nested sparse path like some.very.long.path.field_1 would reproduce the original case in your issue directly
- Case where an object has two sibling fields and only one of them has a value.
- Arrays / multi-value fields. They maybe be hitting a different code path, worth putting a test case maybe?

Also, please check codecov report to ensure we are not leaving open ends for bugs.

OVyshnevskyi:
@sandeshkr419 thanks for review!

I decided to compare the whole document (root level) with expected document and reworked test a bit.
There are more types as well. I also added null objects and empty array (should be missing in re-created `_source`) and also non-sorted array with duplicates -> sorted array with unique values only.

Not shore that i should test every `type`, my change affects mostly ObjectMapper.
At least not as part of this test, i would rather create totally dedicated test for it.

Also i am a bit concerned that my changed does not cover `dynamic` mappings, maybe makes sense to add it as well.

PR #21893, opened 2026-05-29
[Analytics Engine] Let constant filter predicates reach the backend

ahkcs:
### Description

A PPL `convert` over an `eval`-produced literal, followed by a `where` on the result, failed on the analytics-engine path:

```
... | eval s = '10/18/2003 20:07:13' | convert mktime(s) | where s > 1000000000
```

`mktime('<literal>')` is constant and deterministic, but the conversion UDF has no plan-time implementation (only the DataFusion backend evaluates it natively), so `ReduceExpressionsRule` cannot fold the comparison. The unfolded, column-less constant predicate then breaks two layers:

**1. Planner — `OpenSearchFilterRule`.** A deterministic predicate with no field references hit an assertion that `ReduceExpressionsRule` must have eliminated it, throwing `UnsupportedOperationException` (HTTP 500).

**2. Backend — indexed executor.** With the planner fix alone, the predicate reaches the parquet-indexed scan, which classified it as `FilterClass::None` (no index column) and — with `emit_row_ids = false` — rejected it: `execute_indexed_query called with no index_filter(...) in plan`. This only surfaced on the **distributed (indexed)** path; single-node (non-indexed) execution already succeeded, so the gap was invisible until a multi-node cluster ran the query.

### Fix

- **`OpenSearchFilterRule`** — a no-field-reference predicate carries no field, so backend selection is unconstrained by field-type support. Hand it to any backend viable for the child, exactly as the existing non-deterministic branch (`RAND() > 0`) already does.
- **`indexed_executor.rs`** — handle `FilterClass::None` uniformly regardless of `emit_row_ids`: run a predicate-only scan over the page-pruned universe and apply the predicate as a residual in `on_batch_mask`. `supports_filters_pushdown` reports `Exact`, so DataFusion drops the `FilterExec` and relies on the evaluator — which now applies the constant. Constant-true keeps every row; constant-false drops every row.

Both changes are additive: each branch is only reachable when the prior code already threw, so no previously succeeding query is affected.

### Testing

- **Rust e2e** (`indexed_table::tests_e2e::constant_predicate`) — drives a constant-true and constant-false predicate through `IndexedTableProvider` with `emit_row_ids = false`, asserting all 16 rows pass / 0 rows pass respectively (proves the residual is evaluated under `Exact` pushdown, not silently ignored).
- **QA** (`ConversionFunctionsIT`) — two new tests exercising `convert mktime(...) | where ...` with a passing and a filtered-out constant predicate, end-to-end via `POST /_analytics/ppl`.

PPL `convert` command through the analytics-engine path:

| Test suite | Before | After |
|---|---|---|
| `CalciteConvertCommandIT` (opensearch-project/sql, analytics route) | 28/29 | **29/29** |
| `ConversionFunctionsIT` (QA, distributed cluster) | 63/63 | **65/65** |

### Check List


sandeshkr419 (approved):
LGTM, the comments are although over-blown and emit bloated information. Can you please concise up comments.

Thanks!

ahkcs:
> LGTM, the comments are although over-blown and emit bloated information. Can you please concise up comments.
> 
> Thanks!

Trimmed comments, thanks for the suggestion!

PR #21898, opened 2026-05-29
Fsync data files before commit for crash-consistency & delete orphan files after refresh on merge

Bukhtawar:
### Description
Before the fsync change:

  1. LuceneCommitter.commit() calls indexWriter.commit()
  2. IndexWriter.commit() fsyncs only files it manages — files in its own Directory (<shard>/index/): segments_N, .si, .cfs, .doc, etc.
  3. Parquet files live in <shard>/parquet/ — a completely separate directory that Lucene has zero awareness of
  4. These Parquet files were written by RustBridge.finalizeWriter() during refresh → landed in OS page cache → never explicitly fsync'd

  The problem:

  The commit point (segments_N) contains the serialized CatalogSnapshot in its user data. That catalog references Parquet files by name (e.g., parquet/_parquet_file_generation_1.parquet). After indexWriter.commit() returns, the commit
   point is durable on disk, but the Parquet file it references may still only be in page cache.

  If the node crashes in that window:
  - Commit point exists on disk: "this shard's data includes _parquet_file_generation_1.parquet"
  - But that Parquet file is truncated/missing/corrupt (never made it from page cache to disk)
  - On recovery → engine opens → reads commit → tries to use Parquet file → data loss or corruption

Additionally merge on refresh resulted in original files(created by child index writers) prior to the merge to be orphaned and kept lingering on the disk, not freeing up storage, the change now fixes the orphan files

### Related Issues

### Check List


sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
```diff
@@ -142,6 +144,12 @@ public LuceneCommitter(CommitterConfig committerConfig) throws IOException {
     public synchronized CommitResult commit(CommitInput commitData) throws IOException {
         ensureOpen();
         indexWriter.setLiveCommitData(commitData.userData());
+        // Write-ahead fsync: data files durable before the commit point that references them.
+        // getFiles(false) excludes segments_N — IndexWriter.commit() handles that via rename + syncMetaData.
+        if (commitData.catalogSnapshot() != null) {
+            store.directory().sync(commitData.catalogSnapshot().getFiles(false));
+            store.directory().syncMetaData();
```
mgodwan, line 151: Will this apply metadata sync on all the sub directories as well?

server/src/main/java/org/opensearch/index/store/SubdirectoryAwareDirectory.java
```diff
@@ -76,9 +86,40 @@ public long fileLength(String name) throws IOException {
...
+                IOUtils.fsync(normalized, false);
+                subdirectoriesToSync.add(normalized.getParent());
+            }
+        }
+        if (indexFiles.isEmpty() == false) {
+            super.sync(indexFiles);
+        }
+        for (Path dir : subdirectoriesToSync) {
+            IOUtils.fsync(dir, true);
+        }
```
mgodwan, line 137: Do we need this separately? won't sync calling parseFilePath handle it to exactly sync the red files?

mgodwan: This i think we should keep to fsync sub dirs.

Bukhtawar: Agreed, I have moved it into syncMetaData() so the responsibility is clear: sync() handles file data, syncMetaData() handles all directory entries

Bukhtawar: Mostly to handle mock filesystems. For subdirectory files, parseFilePath resolves to an absolute path outside the FSDirectory root. FSDirectory.fsync() then tries to resolve that against its own root. This fails on Lucene's mock test filesystem because the path crosses filesystem provider boundaries, hence IOUtils.fsync() on the resolved path

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneCommitter.java
Bukhtawar: Subdirectory metadata was already handled by the sync() function, however now I have moved it to syncMetadata for better clarity

mgodwan (approved):

Bukhtawar:
```
  2> REPRODUCE WITH: ./gradlew ':server:internalClusterTest' --tests 'org.opensearch.cluster.coordination.RareClusterStateIT.testDisassociateNodesWhileShardInit' -Dtests.seed=4DCA9C92C40B677E -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=saq -Dtests.timezone=America/Manaus -Druntime.java=25
  2> java.lang.AssertionError
        at __randomizedtesting.SeedInfo.seed([4DCA9C92C40B677E:2CD80E305E1E5250]:0)
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertTrue(Assert.java:53)
        at org.opensearch.cluster.coordination.RareClusterStateIT.testDisassociateNodesWhileShardInit(RareClusterStateIT.java:483)
```

PR #21899, opened 2026-05-30
[Arrow Flight RPC] Producer-side back-pressure under slow consumers

rishabhmaurya:
## Summary

`FlightServerChannel.sendResponseBatch` now honours gRPC's `isReady()` contract: the producer thread parks on `BackpressureStrategy.waitForListener` before each batch is queued, and resumes only after gRPC reports the per-stream outbound buffer has drained below `setOnReadyThreshold`. Slow consumers throttle producer wall-clock instead of OOMing the flight pool.

## Context

3.1 used `BaseFlightProducer` with a synchronous `BackpressureStrategy.waitForListener` gate before every `putNext`. When we moved to the async/queue model (`flight-eventloop-N` + `FlightOutboundHandler.BatchTask`), the goal was to match the existing OpenSearch transport's fire-and-forget shape — producer enqueues quickly and returns, ordering is enforced by a single-threaded eventloop. That solved the concurrent-segment-search contention problem (multiple producer threads writing to one channel without serialising on a per-channel mutex), but in doing so we dropped the `isReady()` gate. Without it, the unguarded path ignored gRPC's flow-control signal and let buffers accumulate until the flight pool allocator threw `OutOfMemoryException`.

This PR restores the gate without giving up the async fire-and-forget shape. `awaitReadyOrThrow` runs **on the producer thread before** the `BatchTask` is submitted to the eventloop. The eventloop and its single-threaded ordering guarantees are unchanged.

## What changed

- `FlightServerChannel` registers a `CompositeBackpressureStrategy` and exposes `awaitReadyOrThrow()`. The strategy's cancel callback runs the channel's existing `onChannelCancelled` cleanup before notifying parked threads, so a thread waking from `waitForListener` always observes the cancelled state.
- `FlightOutboundHandler.sendResponseBatch` calls `awaitReadyOrThrow()` before `getExecutor().execute(...)`.
- `CompositeBackpressureStrategy` implements `BackpressureStrategy` directly rather than extending Arrow's `CallbackBackpressureStrategy`, which has a known race ([apache/arrow-java#346](https://github.com/apache/arrow-java/issues/346)): it reads `listener.isReady()` twice — once as the wait-loop predicate, then again post-loop to decide which `WaitResult` to return — and gRPC mutates that flag from Netty event-loop threads outside the strategy's lock, so the two reads can disagree (`"Invalid state when waiting for listener"`). This implementation reads `listener.isReady()` and `listener.isCancelled()` once per loop iteration inside the lock; the handlers' `notifyAll()` simply wakes the waiter so the next iteration reads fresh.
- New settings `arrow.flight.channel.outbound_buffer_threshold` (default 64 MiB) and `arrow.flight.channel.ready_timeout` (default 60s, 100ms minimum). The threshold is wired through to `OSFlightServer.Builder.backpressureThreshold`.

## Settings

| Setting | Default | Behaviour |
|---|---|---|
| `arrow.flight.channel.ready_timeout` | `60s` | Maximum time the producer parks before failing the batch with `StreamErrorCode.TIMED_OUT`. |
| `arrow.flight.channel.outbound_buffer_threshold` | `64mb` | Per-stream gRPC watermark. **Must be strictly smaller than `native.allocator.pool.flight.max`.** |

### Sizing the flight pool

`outbound_buffer_threshold` is **per stream**; `native.allocator.pool.flight.max` is **per node**, shared across all streams.

```
flight pool max  >=  N concurrent streams × (threshold + ~16 MiB headroom)
```

The threshold is a watermark, not a max-message-size — a single batch larger than the threshold ships in one shot, with the producer parking on the next `sendResponseBatch`. See [docs/backpressure.md](plugins/arrow-flight-rpc/docs/backpressure.md) for full sizing guidance.

### Thread-pool exhaustion under slow consumers

The producer thread **parks** under a slow consumer (up to `ready_timeout`). The action handler runs on whichever thread pool it was registered against; N concurrent slow streams hold N threads parked simultaneously, which can starve the pool.

Mitigations:
- Size `ready_timeout` to fail unresponsive streams within an acceptable window.
- Register stream actions on a pool sized for streaming workloads (not on shared CPU-bound pools): a pool of K threads both bounds concurrent streams to K and isolates the parking from unrelated traffic. A dedicated admission-control layer is a more flexible alternative, out of scope here.

### ⚠️ Virtual threads for park-bound producers

> **For workloads where the producer is mostly bottlenecked on `awaitReadyOrThrow` (lots of parking, little compute per batch), the code that registers the stream action can dispatch its handler on a virtual-thread executor it owns.** Park time then doesn't consume a platform thread. CPU-bound work (e.g. building each batch) should still run on a sized platform-thread pool to avoid pinning the carrier — typically by submitting that work to a separate executor and awaiting the result from the virtual thread. The framework itself does not assume virtual threads; this is a decision for the action's registrant.

## Tests

- `CompositeBackpressureStrategyTests` covers handler installation, fast path, ready-handler wakes a parked waiter, cancel-callback runs cleanup before notify, repeated waits return `READY` without a fresh handler firing (gRPC's `OnReadyHandler` is edge-triggered), and timeout.
- `FlightServerChannelTests` covers `awaitReadyOrThrow` fast path, parks-until-OnReady-fires, timeout, cancel-while-waiting + channel cleanup, already-cancelled, cancel-wakes-all-parked-producers (cleanup runs exactly once), post-cancel guard.
- `FlightOutboundHandlerTests` covers the `awaitReadyOrThrow` gating path.
- `ServerConfigTests` covers the new settings' parsing and bounds.
- `BackpressureProducerIT` exercises a slow consumer end-to-end; asserts the stream completes cleanly, all batches arrive, and producer wall-clock reflects consumer pacing (proves the producer was actually throttled).
- Concurrency-sensitive unit tests use `assertBusy` on `Thread.State.TIMED_WAITING` rather than fixed sleeps to avoid flakiness. Stable across multiple reruns.

Coverage on touched code: `CompositeBackpressureStrategy` 100%, `FlightServerChannel` 82%, `ArrowFlightProducer` 77%, `FlightOutboundHandler` 70%. Uncovered lines are pre-existing defensive paths.

## Test plan

- [x] `./gradlew :plugins:arrow-flight-rpc:test`
- [x] `./gradlew :plugins:arrow-flight-rpc:check -x internalClusterTest`
- [x] `./gradlew :plugins:arrow-flight-rpc:internalClusterTest --tests "*BackpressureProducerIT"` (stable across reruns)
- [x] `./gradlew :plugins:arrow-flight-rpc:jacocoTestReport`

### DIFF
--- a/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
+++ b/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
@@ -8,6 +8,7 @@
 
 package org.opensearch.arrow.flight.transport;
 
+import org.apache.arrow.flight.BackpressureStrategy;
 import org.apache.arrow.flight.CallStatus;
 import org.apache.arrow.flight.FlightProducer.ServerStreamListener;
 import org.apache.arrow.flight.FlightRuntimeException;
@@ -36,8 +37,14 @@ import java.util.concurrent.atomic.AtomicInteger;
 import static org.opensearch.arrow.flight.transport.FlightErrorMapper.mapFromCallStatus;
 
 /**
- * TcpChannel implementation for Arrow Flight. It is created per call in ArrowFlightProducer.
- * This implementation is not thread safe; consumer must ensure to invoke sendBatch serially and call completeStream() at the end
+ * TcpChannel implementation for Arrow Flight. Created per call in {@link ArrowFlightProducer}.
+ *
+ * <p>Honours gRPC's {@code isReady()} contract via {@link CompositeBackpressureStrategy}:
+ * producer threads call {@link #awaitReadyOrThrow()} before {@code sendBatch} is queued
+ * and park until gRPC's outbound buffer drains below {@code setOnReadyThreshold}.
+ *
+ * <p>This implementation is not thread safe; the producer must invoke {@code sendBatch}
+ * serially and call {@code completeStream()} at the end.
  */
 class FlightServerChannel implements TcpChannel, ArrowFlightChannel {
     private static final String PROFILE_NAME = "flight";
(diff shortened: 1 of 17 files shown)


plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java
```diff
@@ -56,29 +63,81 @@ class FlightServerChannel implements TcpChannel, ArrowFlightChannel {
...
+     *
+     * @throws StreamException with {@link StreamErrorCode#TIMED_OUT} if the consumer
+     *         remains not-ready longer than {@code readyTimeoutMillis}, or
+     *         {@link StreamErrorCode#CANCELLED} if the client cancelled.
+     */
+    public void awaitReadyOrThrow() {
+        if (cancelled || serverStreamListener.isCancelled()) {
+            throw StreamException.cancelled("stream cancelled before back-pressure wait");
+        }
+        BackpressureStrategy.WaitResult result = bp.waitForListener(readyTimeoutMillis);
```
mch2, line 113: in testing this diff, i'm hitting `cause=java.lang.RuntimeException: Invalid state when waiting for listener.` that causes one of the producers to fail immediately, and the failure isn't getting pushed to the consumer.

rishabhmaurya: good catch! I have addressed it in latest revision

mch2 (approved):
Thanks @rishabhmaurya lgtm! will be this out over the next few days in benchmarks.

PR #21903, opened 2026-05-30
Filter rule: respect nested scalar-function capabilities when computing viable backends

vinaykpud:
#### What this PR does

Updated `OpenSearchFilterRule` to look **inside** predicates when computing per-predicate viable backends. Previously it only checked the outer comparator (`EQUALS`, `>`, `IN`, …) against backend capabilities; nested scalar functions like `CHAR_LENGTH(str)`, `JSON_EXTRACT_ALL(doc)`, `ITEM(map, key)` were ignored. Now every nested `RexCall` is intersected against `scalarBackendsAnyFormat(fn, returnType)` so a backend stays viable only if it can evaluate the **whole predicate**, not just its outer operator.

#### Why

A predicate like:

```
=(ITEM(JSON_EXTRACT_ALL($0), 'user.name'), 'John')
```

would survive marking with `backends=[lucene, datafusion]` because Lucene legitimately declares `EQUALS` on `KEYWORD`. But Lucene has no scalar capability for `JSON_EXTRACT_ALL` or `ITEM-on-MAP` — so when CBO records Lucene as a perf-delegation peer and fragment conversion later calls `EqualsSerializer.buildQueryBuilder` on the original `RexCall`, it finds `(RexCall, RexLiteral)` instead of the expected `(RexInputRef, RexLiteral)` and throws:

```
java.lang.IllegalArgumentException:
  EQUALS performance-delegation requires (RexInputRef, RexLiteral);
  got array_element(map_extract(json_extract_all($0), 'user.name'), 1) = 'John'
```

The bug only surfaces on dual-format indices (parquet primary + lucene secondary). With this fix Lucene drops out at marking time for any predicate whose nested calls it can't evaluate, the perf-delegation arm never invokes the per-function serializer with an unsupported shape, and the runtime IAE becomes unreachable.

#### How

In `OpenSearchFilterRule.resolveViableBackends`:

1. **One tree walk** instead of two — collects field indices and nested scalar-function calls in a single pass into a `PredicateAnalysis` record.
2. **Field-storage intersection** (existing) — narrows on the outer comparator and each field's storage formats. Unchanged.
3. **Scalar-function intersection** (new) — for each nested call, narrows on `scalarBackendsAnyFormat(fn, returnType)`. A backend stays viable only if it can evaluate every nested function.
4. **Polymorphic-UDF fallback** — when a call's return type is `SqlTypeName.ANY` (e.g. `SCALAR_MAX`), infer the type from the first concrete operand. Mirrors `OpenSearchProjectRule`. Throws if even that fails.
5. **Better error messages** — both throw paths name the offending nested call and the enclosing predicate. A TODO marks the next granular-message improvement on the existing empty-set throw.

#### Tests

**Unit tests** — 5 new tests in `FilterRuleTests`:

- `testNoScalarFunctionsKeepsLuceneViable` — baseline; no nested RexCall ⇒ both backends survive.
- `testNestedScalarFunctionDropsLuceneFromPredicateAnnotation` — `UPPER(country) = 'US'` ⇒ Lucene drops.
- `testNestedScalarFunctionsDropLucene` — two-level nesting `UPPER(CONCAT(name, '_x')) = 'FOO'` ⇒ walk recurses.
- `testScalarFunctionNarrowingIsPerLeaf` — AND of two leaves; narrowing is leaf-local.
- `testUnrecognizedScalarFunctionThrows` — unknown operator throws with the function name.

**Integration tests** — removed 13 `@AwaitsFix` annotations across four classes that were skipped specifically for this gap. All previously-skipped tests now run and pass:

| File | Tests un-skipped |
|---|---|
| `WhereCommandIT.java` | 1 (`testWhereInnerLength`) |
| `ReplaceCommandIT.java` | 8 (literal/multi-pair/no-match/expected-rows/multi-field + 2 wildcard + 2 function-form) |
| `RexCommandIT.java` | 3 (sed literal, sed `/i`, sed backreference) |
| `SpathCommandIT.java` | 1 (`testSpathAutoExtractWithWhere`) |

The unused `LuceneTestCase.AwaitsFix` imports are dropped from each.

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
@@ -148,11 +148,11 @@ public class OpenSearchFilterRule extends RelOptRule {
         List<FieldStorageInfo> fieldStorageInfos,
         List<String> childViableBackends
     ) {
-        PredicateAnalysis analysis = new PredicateAnalysis(new HashSet<>(), new ArrayList<>());
+        PredicateContents contents = new PredicateContents(new HashSet<>(), new ArrayList<>());
         for (RexNode operand : predicate.getOperands()) {
-            analyze(operand, analysis);
+            collect(operand, contents);
         }
-        Set<Integer> fieldIndices = analysis.fieldIndices();
+        Set<Integer> fieldIndices = contents.fieldIndices();
 
         CapabilityRegistry registry = context.getCapabilityRegistry();
 
@@ -213,7 +213,7 @@ public class OpenSearchFilterRule extends RelOptRule {
         }
 
         // Every nested scalar function in the predicate must also be evaluable by a candidate backend
-        for (RexCall scalarFunctionCall : analysis.scalarFunctionCalls()) {
+        for (RexCall scalarFunctionCall : contents.scalarFunctionCalls()) {
             // Calcite-internal value constructors (named-parameter MAP/ARRAY/ROW used by full-text
             // operators like match() to pass `field`, `query`, etc.) aren't real scalar functions
             // they're parameter-passing scaffolding. Skip them
@@ -282,18 +282,25 @@ public class OpenSearchFilterRule extends RelOptRule {
      *
      * <p>{@code fieldIndices} — RexInputRef indices feeding the field-storage intersection.
      * <p>{@code scalarFunctionCalls} — nested RexCalls feeding the scalar-function capability intersection.
+     *
+     * <p>TODO: refactor as part of a wider {@link #resolveViableBackends} cleanup. Today the method
+     * juggles four concerns (operand walk, no-field branching, per-field intersection, scalar-function
+     * intersection); the unknown-operator throws at the outer comparator and the nested scalar function
+     * duplicate the same lookup-and-throw shape, and the walker here mirrors {@code OpenSearchProjectRule}'s
+     * operand traversal. A unified visitor over the predicate tree, shared between filter and project
+     * rules, would consolidate all of this.
      */
-    private record PredicateAnalysis(Set<Integer> fieldIndices, List<RexCall> scalarFunctionCalls) {
+    private record PredicateContents(Set<Integer> fieldIndices, List<RexCall> scalarFunctionCalls) {
     }
 
-    /** Recurses the operand subtree, populating {@code analysis} in-place. */
-    private void analyze(RexNode node, PredicateAnalysis analysis) {
+    /** Recurses the operand subtree, populating {@code contents} in-place. */
+    private void collect(RexNode node, PredicateContents contents) {
         if (node instanceof RexInputRef inputRef) {
-            analysis.fieldIndices().add(inputRef.getIndex());
+            contents.fieldIndices().add(inputRef.getIndex());
         } else if (node instanceof RexCall rexCall) {
-            analysis.scalarFunctionCalls().add(rexCall);
+            contents.scalarFunctionCalls().add(rexCall);
             for (RexNode operand : rexCall.getOperands()) {
-                analyze(operand, analysis);
+                collect(operand, contents);
             }
         }
     }
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
@@ -283,12 +283,9 @@ public class OpenSearchFilterRule extends RelOptRule {
      * <p>{@code fieldIndices} — RexInputRef indices feeding the field-storage intersection.
      * <p>{@code scalarFunctionCalls} — nested RexCalls feeding the scalar-function capability intersection.
      *
-     * <p>TODO: refactor as part of a wider {@link #resolveViableBackends} cleanup. Today the method
-     * juggles four concerns (operand walk, no-field branching, per-field intersection, scalar-function
-     * intersection); the unknown-operator throws at the outer comparator and the nested scalar function
-     * duplicate the same lookup-and-throw shape, and the walker here mirrors {@code OpenSearchProjectRule}'s
-     * operand traversal. A unified visitor over the predicate tree, shared between filter and project
-     * rules, would consolidate all of this.
+     * <p>TODO: ensure that the code for tagging and checking the scalar function of a predicate
+     * remains the same as the code for tagging and checking its nested inner expressions as much
+     * as possible.
      */
     private record PredicateContents(Set<Integer> fieldIndices, List<RexCall> scalarFunctionCalls) {
     }
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
@@ -214,6 +214,13 @@ public class OpenSearchFilterRule extends RelOptRule {
 
         // Every nested scalar function in the predicate must also be evaluable by a candidate backend
         for (RexCall scalarFunctionCall : analysis.scalarFunctionCalls()) {
+            // Calcite-internal value constructors (named-parameter MAP/ARRAY/ROW used by full-text
+            // operators like match() to pass `field`, `query`, etc.) aren't real scalar functions
+            // they're parameter-passing scaffolding. Skip them
+            SqlKind kind = scalarFunctionCall.getKind();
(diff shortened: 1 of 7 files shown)


mch2 (approved):

mch2:
@vinaykpud looks like a related test failure:
```
Suite: Test class org.opensearch.analytics.planner.dag.BackendPlanAdapterTests
  2> REPRODUCE WITH: ./gradlew ':sandbox:plugins:analytics-engine:test' --tests 'org.opensearch.analytics.planner.dag.BackendPlanAdapterTests.testSinAdapterInsertsCastForIntegerField' -Dtests.seed=E3BAB78821CE22B2 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=sr-Cyrl-BA -Dtests.timezone=Australia/Brisbane -Druntime.java=25
  2> java.lang.IllegalStateException: No backend can evaluate filter predicate [GREATER_THAN] on fields [status:integer]
        at __randomizedtesting.SeedInfo.seed([E3BAB78821CE22B2:D81C8B48B8352922]:0)
```

mch2 (changes requested):
test failure

vinaykpud:
Yes, related. I will take a look

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -209,6 +212,50 @@ private List<String> resolveViableBackends(
             viableSet.retainAll(fieldViable);
         }
 
+        // Every nested scalar function in the predicate must also be evaluable by a candidate backend
+        for (RexCall scalarFunctionCall : analysis.scalarFunctionCalls()) {
+            ScalarFunction scalarFunc = ScalarFunction.fromSqlOperatorWithFallback(scalarFunctionCall.getOperator());
+            if (scalarFunc == null) {
```
expani, line 225: Let's see how we can merge this with line Line 160 as some of things are duplicated.

vinaykpud: We can do that, but the role of those operators are different. One is outer operator and another is inner scalar function

vinaykpud: Its not worth the refactor now i feel. I'd like to address the duplication as part of that follow-up together with other refactors.

expani: IMO it has to be recursive irrespective of whether the operator is the outermost or the inner one in the predicate of a filter. 

Will take a look whether we can use the existing recursive cycle.

vinaykpud: we can do the refactor, i intentionally dint touch the existing code

expani: I am also fine with starting with a TODO for the same.

What happens if there are multiple levels of nesting in one predicate of a filter ? 

```
EQUALS(LEVEL2(LEVEL3(LEVEL4())))
```

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/FilterRuleTests.java
```diff
@@ -285,6 +286,144 @@ public void testFilterOnDerivedColumnPlansSuccessfully() {
...
+            annotated.getViableBackends().contains(MockLuceneBackend.NAME)
+        );
+    }
+
+    /**
+     * Two-level nesting: {@code UPPER(CONCAT(name, '_x')) = 'FOO'}. Confirms the walk
+     * looks past the outer scalar function into its operands; Lucene supports neither
+     * UPPER nor CONCAT, so it drops out either way.
+     */
+    public void testNestedScalarFunctionsDropLucene() {
```
expani: Can we write some cases with more than 2 levels of nesting ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
vinaykpud: `for (RexCall scalarFunctionCall : analysis.scalarFunctionCalls())` will take care, may be i can rename `PredicateAnalysis`, it actually collects FieldIndices and scalarFunctionCalls

vinaykpud: I have covered unit tests for this case

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/FilterRuleTests.java
vinaykpud: Sure

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
expani: The unit tests are only covering 1 level of nesting with `UPPER(CONCAT(name, '_x')) = 'FOO'` So, we still need to add for 2 or more levels of nesting

vinaykpud: Yes, just added, i will push

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/FilterRuleTests.java
```diff
@@ -285,6 +286,177 @@ public void testFilterOnDerivedColumnPlansSuccessfully() {
...
+
+        OpenSearchFilter result = runFilter(
+            "parquet",
+            Map.of("name", Map.of("type", "keyword", "index", true)),
+            new String[] { "name" },
+            new SqlTypeName[] { SqlTypeName.VARCHAR },
+            predicate
+        );
+
+        AnnotatedPredicate annotated = (AnnotatedPredicate) result.getCondition();
```
expani: We should be asserting that every inner function of the Predicate has been annotated. 

Looks like that's missing here.

vinaykpud: What do we want to assert here exactly?

say the predicate look like this
```
ANNOTATED_PREDICATE(id=0,
                      backends=[datafusion],
                      =(UPPER(CONCAT(UPPER(CONCAT($0, '_a':VARCHAR)), '_b':VARCHAR)), 'FOO':VARCHAR))

```

expani: Given that a predicate inside a Filter operator can only have one backend. These assertions are okay, if possible add some negative cases for the new exception being thrown `Unmapped return type`. Else LGTM

expani (approved):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
```diff
@@ -223,13 +277,30 @@ private List<String> resolveViableBackends(
...
+     * Result of a single walk over a predicate's operand subtree.
+     *
+     * <p>{@code fieldIndices} — RexInputRef indices feeding the field-storage intersection.
+     * <p>{@code scalarFunctionCalls} — nested RexCalls feeding the scalar-function capability intersection.
+     *
+     * <p>TODO: refactor as part of a wider {@link #resolveViableBackends} cleanup. Today the method
+     * juggles four concerns (operand walk, no-field branching, per-field intersection, scalar-function
+     * intersection); the unknown-operator throws at the outer comparator and the nested scalar function
+     * duplicate the same lookup-and-throw shape, and the walker here mirrors {@code OpenSearchProjectRule}'s
+     * operand traversal. A unified visitor over the predicate tree, shared between filter and project
```
expani: We shouldn't be unifying the visitor's for Filter and Project instead we need to combine any common things between them via Utility methods if required. 

I think this rule should instead have one common recursion into predicates and it's inner expression which can noted in the TODO

expani: Can you please change the TODO to be along these lines ? 

```
Ensure that the code for tagging and checking Scalar Function of a Predicate remains the same as the code for tagging and checking it's nested inner expressions as much as possible
```

mch2 (approved):

PR #21911, opened 2026-05-31
[analytics-engine] Wire FINAL aggregate filter drop and join-condition adapter dispatch

songkant-aws:
## Description

Two small fixes that together unblock \`CalciteTransposeCommandIT\` on the analytics-engine route. Both surface only when a single PPL command produces:
- (a) a non-prefix groupSet,
- (b) a FILTER aggCall, and
- (c) a Join whose condition carries PPL UDFs.

PPL transpose lowers via \`RelBuilder.unpivot()\` + \`RelBuilder.pivot()\` and is the only PPL command today that hits all three at once. The underlying bugs are general — any future command that lands in this shape will hit them.

## What goes wrong on main today

\`CalciteTransposeCommandIT\` against \`-Dtests.analytics.parquet_indices=true\` (analytics-engine route) on the merged-21804 main: 0/5 pass, 5/5 fail with \`IllegalArgumentException: filter must be BOOLEAN NOT NULL\`.

## Fix 1 — DistributedAggregateRewriter.buildOne: drop filterArg on FINAL

After PARTIAL+FINAL split, FINAL's input is PARTIAL output, laid out as \`[group keys, agg states]\`. The boolean column referenced by an aggCall's FILTER predicate exists only on the **ORIGINAL child input** that PARTIAL consumed; PARTIAL has already applied the filter while accumulating state. \`Aggregate.<init>\`'s \`isPredicate(input, filterArg)\` check fires when it reads \`filterArg=N\` against PARTIAL output that has fewer than \`N+1\` columns (or whose \`N\`th column is non-boolean).

Set \`filterArg = -1\` on the FINAL call. Calcite treats any \`filterArg < 0\` as "no FILTER" (see \`AggregateCall.hasFilter()\` and the \`Aggregate.<init>\` check \`aggCall.filterArg < 0 || isPredicate(...)\`); \`-1\` is the conventional value used throughout Calcite's own rule rewrites.

### Prior art (Calcite-blessed pattern)

This is exactly how Calcite and Flink themselves construct the merge-stage aggCall in two-phase aggregation:

- **Apache Calcite** [\`SqlSplittableAggFunction\`](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/SqlSplittableAggFunction.java) is the public abstraction for splittable aggregates. Every \`topSplit()\` implementation (\`CountSplitter\`, \`SumSplitter\`, \`MinMaxSplitter\`) hard-codes \`-1\` as filterArg when constructing the merge-stage aggCall — see e.g. \`CountSplitter.topSplit()\` returning \`AggregateCall.create(..., ImmutableList.of(ordinal), -1, ...)\` and \`CountSplitter.other()\` doing the same.
- **Apache Flink** [\`SplitAggregateRule\`](https://github.com/apache/flink/blob/master/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/plan/rules/logical/SplitAggregateRule.java) (used by Flink's MPP-style two-stage local/global aggregation) does the same thing structurally:
  - **STEP 2 partial aggregates**: keep \`aggCall.filterArg\` (the FILTER column still exists on PARTIAL's input).
  - **STEP 3 final aggregates**: build with \`AggregateCall.create(..., newArgList, -1, ...)\` — argList shifted to PARTIAL's output positions, filterArg dropped.

Both reflect the same semantic invariant: FILTER is consumed once during state accumulation; downstream merge stages see only states and have no rows left to gate.

## Fix 2 — BackendPlanAdapter: dispatch OpenSearchJoin for adapter rewrite

\`adaptNode()\` walks \`Filter\` / \`Project\` / \`Aggregate(FINAL)\` and runs each RexNode through the backend's \`ScalarFunctionAdapter\` chain (e.g. \`ToStringFunctionAdapter\` rewrites \`NUMBER_TO_STRING\` to a plain \`CAST\` that isthmus understands).

Calcite's \`FILTER_INTO_JOIN\` rule inlines an outer Filter's predicate into an inner Join's condition, so any PPL UDF that lived in the Project below the Filter rides into the Join condition. With \`OpenSearchJoin\` missing from \`adaptNode\`'s dispatch list, that copy of the UDF reaches isthmus unrewritten and trips:

\`\`\`
java.lang.IllegalArgumentException: Unable to convert call NUMBER_TO_STRING(fp64?).
  at io.substrait.isthmus.expression.RexExpressionConverter.visitCall(...)
\`\`\`

Add an \`OpenSearchJoin\` branch that runs the join's condition through \`adaptRex\` with concatenated left+right field storage — same convention \`OpenSearchJoin#getOutputFieldStorage()\` uses on the output side.

## Test coverage

New unit tests in \`BackendPlanAdapterTests\`:

- \`testJoinConditionAdapterInsertsCastForIntegerField\` — Join condition carrying \`SIN(integer_column)\` ends up with the integer operand wrapped in \`CAST(... AS DOUBLE)\` after adaptation, mirroring how \`testSinAdapterInProjectInsertsCastForIntegerField\` exercises the same adapter on a Project.
- \`testJoinConditionNoOpWhenNoAdaptedFunctions\` — Join with a plain equi-condition (no PPL UDF) passes through structurally unchanged. Guards against the dispatcher accidentally rebuilding Joins it doesn't need to touch.

## Verified results

\`CalciteTransposeCommandIT\` on 21804-merged main:

| Setup | Pass | Failure mode |
|---|---|---|
| Without these fixes | 0/5 | \`filter must be BOOLEAN NOT NULL\` |
| With Fix 1 only | 4/5 | \`testTransposeWithValueFieldNameCollision\` hits \`Unable to convert call NUMBER_TO_STRING(fp64?)\` |
| With Fix 1 + Fix 2 | **5/5** ✅ | — |

## Test plan

- [x] \`./gradlew :sandbox:plugins:analytics-engine:test --tests "org.opensearch.analytics.planner.dag.BackendPlanAdapterTests"\` — 8/8 (2 new + 6 existing all pass)
- [x] \`./gradlew :sandbox:plugins:analytics-engine:spotlessCheck\`
- [x] Local cluster via \`./gradlew run -x checkCompatibility\` with the sandbox plugin set, transpose IT 5/5 pass against analytics-engine route
- [ ] Full \`:integ-test:integTestRemote\` suite

## Related

- Builds on top of #21804 (LITERAL_AGG + JoinRule relaxation), which unblocks the marking-phase \`unmarked LogicalJoin\` error one layer earlier.
- Once merged, the corresponding sql-plugin workaround (opensearch-project/sql#5479 — axis literal padding + unbounded VARCHAR cast in transpose lowering) becomes redundant and can be closed.


sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
```diff
@@ -311,6 +311,111 @@ protected Map<ScalarFunction, ScalarFunctionAdapter> scalarFunctionAdapters() {
...
+            leftScan,
+            rightScan,
+            List.of(),
+            condition,
+            Set.of(),
+            org.apache.calcite.rel.core.JoinRelType.INNER
+        );
+
+        RelNode marked = runPlanner(join, context);
+        LOGGER.debug("Marked join:\n{}", RelOptUtil.toString(marked));
```
songkant-aws: Remove logger in test

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
```diff
@@ -311,6 +311,111 @@ protected Map<ScalarFunction, ScalarFunctionAdapter> scalarFunctionAdapters() {
...
+
+        QueryDAG dag = DAGBuilder.build(marked, context.getCapabilityRegistry(), mockClusterService(), TEST_RESOLVER);
+        PlanForker.forkAll(dag, context.getCapabilityRegistry());
+        BackendPlanAdapter.adaptAll(dag, context.getCapabilityRegistry());
+
+        StagePlan plan = findStagePlanByFragmentType(dag, org.opensearch.analytics.planner.rel.OpenSearchJoin.class);
+        org.opensearch.analytics.planner.rel.OpenSearchJoin adaptedJoin = (org.opensearch.analytics.planner.rel.OpenSearchJoin) plan
```
songkant-aws: Use import of OpenSearchJoin instead of full path

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
```diff
@@ -311,6 +311,111 @@ protected Map<ScalarFunction, ScalarFunctionAdapter> scalarFunctionAdapters() {
...
+        RelNode leftScan = stubScan(table);
+        RelNode rightScan = stubScan(table);
+
+        // Equi-join on left.$0 == right.$0 — no PPL UDF, nothing for adapter to rewrite.
+        RexNode condition = rexBuilder.makeCall(
+            SqlStdOperatorTable.EQUALS,
+            rexBuilder.makeInputRef(typeFactory.createSqlType(SqlTypeName.INTEGER), 0),
+            rexBuilder.makeInputRef(typeFactory.createSqlType(SqlTypeName.INTEGER), 2)
+        );
+        RelNode join = org.apache.calcite.rel.logical.LogicalJoin.create(
```
songkant-aws: Use import of OpenSearchJoin instead of full path

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
```diff
@@ -311,6 +311,111 @@ protected Map<ScalarFunction, ScalarFunctionAdapter> scalarFunctionAdapters() {
...
+            leftScan,
+            rightScan,
+            List.of(),
+            condition,
+            Set.of(),
+            org.apache.calcite.rel.core.JoinRelType.INNER
```
songkant-aws: Use import of JoinRelType.INNER full path

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
```diff
@@ -109,6 +113,49 @@ private static RelNode adaptNode(RelNode node, Map<ScalarFunction, ScalarFunctio
...
+     * a PPL transpose's {@code CAST(numeric AS VARCHAR)} CASE branch, picked up when the
+     * outer {@code IS NOT NULL(_value_transpose_)} filter is pushed down) get rewritten into
+     * substrait-compatible shapes before fragment conversion. Without this dispatch,
+     * isthmus's RexExpressionConverter sees the unadapted PPL UDF and fails with
+     * "Unable to convert call ...".
+     *
+     * <p>Field storage for adapter recursion concatenates left and right input storage,
+     * matching Calcite's join row type ordering (left fields first, then right) — same
+     * convention {@link OpenSearchJoin#getOutputFieldStorage()} uses on the output side.
+     */
```
songkant-aws, line 125: Concise up the comment. No need to explain so much

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
songkant-aws: Done in e754144c8d3752756d846c348755000f3c52b93a.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
songkant-aws: Done in e754144c8d3752756d846c348755000f3c52b93a.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
songkant-aws: Done in e754144c8d3752756d846c348755000f3c52b93a.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/dag/BackendPlanAdapterTests.java
songkant-aws: Done in e754144c8d3752756d846c348755000f3c52b93a.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
songkant-aws: Done in e754144c8d3752756d846c348755000f3c52b93a.

songkant-aws:
Addressed all review comments in the latest force-push (e754144c8d3):
- Removed the LOGGER call from `testJoinConditionAdapterInsertsCastForIntegerField`.
- Replaced full-path `org.apache.calcite.rel.logical.LogicalJoin` / `JoinRelType.INNER` / `org.opensearch.analytics.planner.rel.OpenSearchJoin` with explicit imports.
- Concised the `adaptJoin` Javadoc to three lines.
- Also picked up coderabbit's suggestion: early-return on empty `fieldStorage` so we don't kick off `adaptRex` with no storage info to consult — keeps the no-marked-input path identical to its previous behavior (just propagate `childrenChanged` through `join.copy(...)`).

Re the failing `sandbox-check`: the two `DatafusionReduceSinkTests.testCancel*UnwindsDrain` failures are unrelated to this PR — they live in `analytics-backend-datafusion` (this PR doesn't touch that module) and the failure mode is `RuntimeException: Execution error: Worker gone` from the native runtime cancellation path, which is a known race in the reduce-sink drain (not caused by my changes). The local UT (`BackendPlanAdapterTests` 8/8 pass, including the two new join-condition tests) and the local `CalciteTransposeCommandIT` end-to-end run (5/5 pass against the analytics-engine route) confirm the change itself is correct. I expect the next CI run to pass the relevant jobs; if `sandbox-check` keeps tripping on the same flaky reduce-sink tests we should rerun it.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/BackendPlanAdapter.java
```diff
@@ -114,6 +118,42 @@ private static RelNode adaptNode(RelNode node, Adapters adapters) {
...
+     * Adapts {@link OpenSearchJoin#getCondition()} so PPL UDFs inlined by
+     * Calcite's FILTER_INTO_JOIN reach the fragment converter in their adapted shape.
+     * Field storage is left ++ right output storage (Calcite join row-type ordering).
+     */
+    private static RelNode adaptJoin(OpenSearchJoin join, Adapters adapters, List<RelNode> adaptedChildren, boolean childrenChanged) {
+        RelNode left = childrenChanged ? adaptedChildren.get(0) : join.getLeft();
+        RelNode right = childrenChanged ? adaptedChildren.get(1) : join.getRight();
+        List<FieldStorageInfo> fieldStorage = new ArrayList<>();
+        if (RelNodeUtils.unwrapHep(left) instanceof OpenSearchRelNode l) {
+            fieldStorage.addAll(l.getOutputFieldStorage());
```
expani: FieldStorage is populated at marking. 
So, a Join should get them at marking and we need to just copy the same into the recreated `OpenSearchJoin` here.

expani: Basically, `join.getOutputFieldStorage()` should already have the values. so, we don't need to recreate from left and right childs again.

expani: If it doesn't have, then we can populate in `OpenSearchJoinRule`

songkant-aws: Yeah, checked that it should be populated during scan and the join's each child should have already deduced the subtree's FieldStorage.

expani (approved):

mch2 (approved):

PR #21913, opened 2026-05-31
Add datafusion.spill_directory setting

gaurav-amz:
## Description

Lets operators point DataFusion spill at a dedicated filesystem instead of inheriting the hardcoded `path.data[0]/../tmp` default. New setting is exposed to set the directory which will be picked by datafusion fallback to the disabled spill otherwise.

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
@@ -178,27 +178,30 @@ public class DataFusionPlugin extends Plugin
      * Spill directory used by DataFusion's {@code DiskManager} for intermediate state when
      * operators (HashAggregate, Sort, TopK) exceed {@link #DATAFUSION_MEMORY_POOL_LIMIT}.
      *
-     * <p><strong>Default: empty string</strong> — falls back to a {@code tmp/} sibling of
-     * {@code path.data[0]}, preserving the historical behaviour for self-managed clusters that
-     * never configured this setting. AOS optimized domain set this to
-     * {@code /mnt/analytics/backend/spill} via {@code opensearch.yml} so spill traffic lands on
-     * a dedicated EBS volume instead of contending with the data path.
+     * <p>Optional. When set, DataFusion uses {@code DiskManagerMode::Directories} to spill
+     * to the configured path. When unset (empty), DataFusion runs in
+     * {@code DiskManagerMode::Disabled} — spill is off and queries that exceed
+     * {@link #DATAFUSION_MEMORY_POOL_LIMIT} fail with a clear "DiskManager is disabled" error
+     * rather than silently spilling somewhere unexpected.
+     *
+     * <p>{@code Final} because DataFusion's {@code DiskManager} is built once at runtime
+     * startup; changing the directory mid-flight would orphan in-progress spill files.
      */
     public static final Setting<String> DATAFUSION_SPILL_DIRECTORY = new Setting<>(
         "datafusion.spill_directory",
         "",
         Function.identity(),
         DataFusionPlugin::validateSpillDirectory,
-        Setting.Property.NodeScope
+        Setting.Property.NodeScope,
+        Setting.Property.Final
     );
 
     /**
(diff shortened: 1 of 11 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -168,6 +170,58 @@ static String deriveMemoryPoolLimitDefault(Settings settings) {
...
+     * operators (HashAggregate, Sort, TopK) exceed {@link #DATAFUSION_MEMORY_POOL_LIMIT}.
+     *
+     * <p><strong>Default: empty string</strong> — falls back to a {@code tmp/} sibling of
+     * {@code path.data[0]}, preserving the historical behaviour for self-managed clusters that
+     * never configured this setting. AOS optimized domain set this to
+     * {@code /mnt/analytics/backend/spill} via {@code opensearch.yml} so spill traffic lands on
+     * a dedicated EBS volume instead of contending with the data path.
+     */
+    public static final Setting<String> DATAFUSION_SPILL_DIR = new Setting<>(
+        "datafusion.spill_dir",
```
Bukhtawar:
```suggestion
        "datafusion.spill_directory",
```

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -168,6 +170,58 @@ static String deriveMemoryPoolLimitDefault(Settings settings) {
...
+        if (value == null || value.isEmpty()) {
+            return value;
+        }
+        try {
+            Path.of(value).toAbsolutePath().normalize();
+        } catch (java.nio.file.InvalidPathException e) {
+            throw new IllegalArgumentException("Setting [datafusion.spill_dir] is not a valid path: [" + value + "]", e);
+        }
+        return value;
+    }
```
Bukhtawar: Should we validate if the directory is writable? Also should we add this to FsHealthCheck?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
gaurav-amz: Ah, yes. I missed updating this.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
gaurav-amz: As discussed, I'll address this comment in a follow-up PR. Boot-up script will handle making the disk writable.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
gaurav-amz: Done.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -172,6 +174,58 @@ static String deriveMemoryPoolLimitDefault(Settings settings) {
...
+     * uses it as-is; otherwise falls back to the legacy {@code path.data[0]/../tmp} default.
+     * The fallback preserves backward compatibility for self-managed clusters that never
+     * configured this setting.
+     */
+    static String resolveSpillDirectory(Settings settings, Environment environment) {
+        String configured = DATAFUSION_SPILL_DIRECTORY.get(settings);
+        if (configured != null && configured.isEmpty() == false) {
+            return configured;
+        }
+        return environment.dataFiles()[0].getParent().resolve("tmp").toAbsolutePath().toString();
```
Bukhtawar: Lets avoid a fallback

gaurav-amz: Done, If the directory is not set then the spill will be disabled.

Bukhtawar (approved):

PR #21917, opened 2026-05-31
analytics-engine: fix explain to enforce index permission check and run query under framework AnalyticsQueryTask

mch2:
### Description
This PR does two things:
1. Fix executeWithProfile to evaluate index permissions by invoking the local transport action.
2. Remove manual register/unregister of Analytics tasks.  This will now be handled through the local call / transport framework.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryRequest.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryRequest.java
@@ -14,10 +14,14 @@ import org.opensearch.action.ActionRequestValidationException;
 import org.opensearch.action.IndicesRequest;
 import org.opensearch.action.support.IndicesOptions;
 import org.opensearch.analytics.QueryRequestContext;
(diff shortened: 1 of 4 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryRequest.java
```diff
@@ -74,6 +97,16 @@ public IndicesOptions indicesOptions() {
         return IndicesOptions.strictExpandOpen();
     }
 
+    @Override
+    public Task createTask(long id, String type, String action, TaskId parentTaskId, Map<String, String> headers) {
+        // TODO: assign the real query id here instead of QUERY_ID_NOT_YET_ASSIGNED. The id is
+        // currently minted later in DAGBuilder.newQueryId() (random UUID), so the task and the
```
expani, line 103: Makes sense.

Also, I remember there was a TODO lying around to make it unique at least per node to avoid collisions if in future it becomes used for different things. 

But, those can come later.

sandeshkr419 (approved):

PR #21922, opened 2026-06-01
Analytics-Engine: Reduce-input early termination & fix reduce thread leak

mch2:
### Description

This PR fixes two issues caught while benchmarking:

1.  Upon cancellations/rejections of expensive queries, reduce threads were blocked indefinately, we were not tearing down on failure.
2. Early termination was not functional, we were never closing the backing streams for each sender input.

  The fix propagates a "done" signal back through the streaming response path so shards cancel their flight streams immediately, freeing resources for other queries.

  Changes

  - Native receiver-drop sentinel: df_sender_send returns SENDER_SEND_RECEIVER_DROPPED (positive, rides the success half of the FFM contract) when the consumer dropped the receiver. DatafusionPartitionSender
  latches the state, DatafusionReduceSink.feedToSender keys off it to short-circuit.
  - ExchangeSink.isConsumerDone(): best-effort signal that the downstream consumer will read no more batches. Default false; DatafusionReduceSink overrides based on the sender's receiver-dropped latch.
  - StreamingResponseListener: returns boolean keepReading from onStreamResponse. When false, AnalyticsSearchTransportService cancels the flight stream instead of draining to exhaustion.
  - ShardFragmentStageExecution: checks isConsumerDone() after each feed and returns false to cancel the shard stream.
  - closeChildInput on failure: the attachChildren cascade calls closeChildInput(childId) when a child transitions to FAILED, closing the sender so the reduce drain doesn't hang forever waiting for input from a
  dead child.
  - Idempotent closeImpl: simplified teardown guarded by a single torndown AtomicBoolean CAS. Closes outStream first (drops receiver, unblocks parked senders), then session. No state-machine dispatch — safe to
  call from any thread, any number of times.

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionReduceSink.java
```diff
@@ -319,52 +357,36 @@ protected void feedBatchUnderLock(VectorSchemaRoot batch) {
      */
     @Override
     protected Exception closeImpl() {
-        SinkState before = state.compareAndExchange(SinkState.READY, SinkState.DONE);
-        if (before == SinkState.REDUCING) {
-            // Drain parked — dropping senders/outStream now would panic in drop_in_place.
-            fireCancelQuery();
-            return null;  // reduce()'s finally calls closeImpl directly to tear down.
-        }
-        // before == READY (we just won) or DONE (reduce's finally calling us, or duplicate close).
```
aravindsagar, line 328: Does this cause a use-after-free issue? The docstring which still refers to the old code suggests so.

df_stream_next in the drain thread could be parked while holding a reference to stream_ptr. This close can run on another thread and drop the stream.

aravindsagar (commented):
`testCancelBeforeFirstBatchUnwindsDrain` and `testCancelAfterFirstBatchUnwindsDrain` are removed - perhaps we can keep those with the fix in https://github.com/opensearch-project/OpenSearch/pull/21924?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionReduceSink.java
mch2: you're right, adding this back, its also easier to reason about.  Though i think its only a double close from the finally block in `reduce` and not a use after free.

Bukhtawar (commented):
Need to add integ tests

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -1515,7 +1497,7 @@ pub unsafe fn sender_send(
     array_ptr: i64,
     schema_ptr: i64,
     io_handle: &tokio::runtime::Handle,
-) -> Result<(), DataFusionError> {
+) -> Result<crate::partition_stream::SendOutcome, DataFusionError> {
```
aravindsagar, line 1513: Should we update the [caller of this function in ffm.rs](https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-backend-datafusion/rust/src/ffm.rs#L561) to handle this change? There it's mapping to 0 irrespective of SendOutcome

mch2: updated - also added ITs to assert the outcome / cancel notice is sent on the stream so data nodes stop scanning.

mch2: Am hitting a leak on the java side per/query arrow allocator in SearchCancellationIT, during shutdown of the test - am investigating, only able to repro in CI.

mch2: fixed - needed to close the incoming batch iff we are unable to feed to the sink, else the per-query allocator closes at the end of cancel and throws a leak - caught by SearchCancellationIT.
```
 VectorSchemaRoot batch = it.next().getArrowRoot();
                boolean fed = false;
                try {
                    ctx.downstream().feed(batch);
                    fed = true;
                } finally {
                    if (!fed) {
                        batch.close();
                    }
                }
```

sandeshkr419 (approved):

PR #794, opened 2021-05-28
Adding release notes for 1.0.0-rc1

saratvemulapalli:
Signed-off-by: Sarat Vemulapalli <vemulapallisarat@gmail.com>

### Description
Adding release notes for OpenSearch 1.0.0-rc1.
This is generated using the tool: https://github.com/ariatemplates/git-release-notes
 
### Issues Resolved
#789 
 
### Check List
- [ ] New functionality includes testing.
  - [x] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #21925, opened 2026-06-01
Fix Time32-vs-Arrow mismatch on PPL time queries

vinaykpud:
### Problems

**1. Time32-vs-Time64 schema mismatch.** PPL time-of-day functions — `time(ts)`, `current_time()`, `curtime()`, `maketime(...)` — fail with:

```
Substrait error: Field has type Time32(Nanosecond), table column is Time64(Nanosecond)
```

`Time32(µs|ns)` isn't a real Arrow type — Arrow only allows `Time32(s|ms)` and `Time64(µs|ns)`. The plan describes something that can't round-trip cleanly to a real column, so all four functions are broken.

**2. `maketime` returns a different unit from its siblings.** `time(ts)`, `current_time()`, and `curtime()` return `Time64(Nanosecond)`; `maketime(...)` returns `Time64(Microsecond)`. Inconsistent and forces callers to special-case maketime when comparing or arithmetic-combining time values.

### Root cause (Problem 1)

substrait-java's `PrecisionTime` POJO is missing a setter for `typeVariationReference`, so high-precision time fields always serialize with that field at its proto default `0` (= Time32) regardless of precision. DataFusion's substrait consumer dutifully constructs `Time32(Nanosecond)` and the schema check rejects it against the real `Time64` column.

For low-precision values (seconds, ms) the default-0 encoding happens to coincide with Arrow's only legal Time32 layout — that's why `from_unixtime(...)` works and the bug only surfaces at µs/ns.

### Fix

* **`SubstraitPlanProtoRewriter`** — new proto-layer walker. Promotes `PrecisionTime` variation `0 → 1` (Time32 → Time64) for precision ∈ {6, 9}. Wired into both Substrait emit sites in `DataFusionFragmentConvertor` (`serializePlan` + the direct-proto schema-only path). Fixes Problem 1.
* **`maketime` UDF return type** — emit `Time64(Nanosecond)` instead of `Time64(Microsecond)`. Fixes Problem 2.
* **`SubstraitPlanRewriter` → `SubstraitPlanPojoRewriter`** — renamed.

Once substrait-java exposes `typeVariationReference` on `Type.PrecisionTime` (mirroring upstream PR [#794](https://github.com/substrait-io/substrait-java/pull/794) for `Type.UserDefined`), `SubstraitPlanProtoRewriter` can be deleted.

### Why two rewriters (POJO + proto), not one

**1. Why the new fix has to be at proto, not POJO** — substrait-java's POJO `io.substrait.plan.Plan` has no setter for `typeVariationReference` on `Type.PrecisionTime`. The field is only present on `io.substrait.proto.Type.PrecisionTime`, after `PlanProtoConverter` has serialized the POJO. So the only place to set it is post-serialization, on the proto.

**2. Why the existing POJO rewrites couldn't be moved into the proto rewriter** — `SubstraitPlanPojoRewriter` uses substrait-java's typed `RelCopyOnWriteVisitor` / `ExpressionCopyOnWriteVisitor`, which give recursion through every Rel variant (Filter, Project, Aggregate, Join, Sort, …) and every Expression case (function args, casts, if-then-else, scalar subqueries) for free. At the proto layer there's no equivalent typed visitor — you'd hand-roll all that recursion. A lot of code for no functional gain.

### Tests

* `SubstraitPlanProtoRewriterTests` — covers `PrecisionTime` precision 0/3/6/9 paths, nested `Struct/List/Map`, `ReadRel` base schema rewrite, no-op identity case.
* All four affected functions verified end-to-end on multi-shard PPL queries.

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanProtoRewriter.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanProtoRewriter.java
@@ -1,0 +1,327 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.be.datafusion;
+
+import io.substrait.proto.CrossRel;
+import io.substrait.proto.ExtensionMultiRel;
+import io.substrait.proto.HashJoinRel;
+import io.substrait.proto.JoinRel;
+import io.substrait.proto.MergeJoinRel;
+import io.substrait.proto.NamedStruct;
+import io.substrait.proto.NestedLoopJoinRel;
+import io.substrait.proto.Plan;
+import io.substrait.proto.PlanRel;
+import io.substrait.proto.ReadRel;
+import io.substrait.proto.Rel;
+import io.substrait.proto.RelRoot;
+import io.substrait.proto.SetRel;
+import io.substrait.proto.Type;
+
+/**
+ * Proto-layer Substrait plan rewriter. Runs after {@code SubstraitPlanPojoRewriter} serializes
+ * the POJO so it can patch fields the POJO API doesn't expose. Today it covers a single case:
+ * {@code Type.PrecisionTime} arrives with {@code type_variation_reference = 0}
+ * (TIME_32) regardless of the precision, because substrait-java's POJO has no
+ * setter for that field. For {@code precision ∈ {6, 9}} that yields
+ * {@code Time32(µs|ns)} — an Arrow type that doesn't exist; the DataFusion consumer
+ * builds it anyway and rejects the schema match against the real {@code Time64}
+ * column. Promoting to variation 1 (TIME_64) is unconditionally correct: the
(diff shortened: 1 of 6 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/SubstraitPlanProtoRewriter.java
```diff
@@ -0,0 +1,327 @@
...
+                }
+                return pr.toBuilder().setRoot(root.toBuilder().setInput(rewritten).build()).build();
+            }
+            case RELTYPE_NOT_SET:
+            default:
+                return pr;
+        }
+    }
+
+    private static Rel rewriteRel(Rel rel) {
```
expani, line 96: This doesn't feel right for handling each operator individually.

Why can't it move to Calcite layer ?

vinaykpud: Ideally yes, but we can't. Its a gap now with substrait-java, always set the `typeVariationReference` as 0, there is no java API to set this value and its defaulted to 0 always, this is required by the DF substrait consumer to decide the proper arrow datatype. 

This cannot be moved to calcite layer, even we had to do `SubstraitPlanProtoRewriter.java` earlier to earlier for another gap with substrait-java.

mch2 (approved):

PR #21929, opened 2026-06-01
Fix field resolution for dotted fields under disable_objects

mohit10011999:
### Description
Fixes field resolution and search for dotted field names under disable_objects. This PR fixes several issues that prevented these fields from being correctly searchable.

### Problems Fixed

1. Mapping parse ordering (ObjectMapper.java, RootObjectMapper.java): Pre-parse disable_objects before iterating the node map, so parseProperties knows to treat dotted names as literals regardless of JSON key order.
2. Dynamic mapping update (DocumentParser.java): The `handleDisableObjectsMapping` function handles both the first-mapper (add) and subsequent-mapper (merge) cases.
3. Field name reconstruction on merge (ParametrizedFieldMapper.java): Fixed parentPath() to use simpleName (which may contain dots under disable_objects) when computing the parent content path. Without this, a field like `attributes.address.city` with simpleName = "address.city" would incorrectly reconstruct as `attributes.address.address.city`.
4. Multi-field misclassification (MappingLookup.java): Fixed isMultiField() to check whether any ancestor has disableObjects() == true. Without this, `attributes.address.city` was incorrectly treated as a multi-field of `attributes.address` (since both exist in fieldMappers), hiding it from search.
### Related Issues

### Check List


mohit10011999:
Curl commands used for testing in local:-

```
# ============================================================
# Step 1: Create index with disable_objects on a nested object
# ============================================================
curl -X PUT "localhost:9200/test-disable-objects" -H 'Content-Type: application/json' -d '
{
  "mappings": {
    "properties": {
      "attributes": {
        "type": "object",
        "disable_objects": true,
        "dynamic": true
      }
    }
  }
}'

# ============================================================
# Step 2: Index Doc 1 — creates "attributes.address.city" as flat field
# ============================================================
curl -X PUT "localhost:9200/test-disable-objects/_doc/1?refresh=true" -H 'Content-Type: application/json' -d '
{
  "attributes": {
    "address.city": "Austin"
  }
}'

# ============================================================
# Step 3: Index Doc 2 — creates "attributes.address" as flat field
#          (coexists with "attributes.address.city" — no conflict)
# ============================================================
curl -X PUT "localhost:9200/test-disable-objects/_doc/2?refresh=true" -H 'Content-Type: application/json' -d '
{
  "attributes": {
    "address": "US"
  }
}'

# ============================================================
# Step 4: Verify mapping — both fields should be flat under "attributes"
#          No intermediate "address" ObjectMapper should exist
# ============================================================
curl -X GET "localhost:9200/test-disable-objects/_mapping?pretty"

# Expected: attributes.properties should contain:
#   "address.city": { "type": "text", ... }
#   "address": { "type": "text", ... }
# NOT a nested "address" object with "city" inside it.

# ============================================================
# Step 5: Qualified search — attributes.address.city:Austin
# ============================================================
curl -X GET "localhost:9200/test-disable-objects/_search?pretty" -H 'Content-Type: application/json' -d '
{
  "query": {
    "match": {
      "attributes.address.city": "Austin"
    }
  }
}'
# Expected: returns doc 1

# ============================================================
# Step 6: Qualified search — attributes.address:US
# ============================================================
curl -X GET "localhost:9200/test-disable-objects/_search?pretty" -H 'Content-Type: application/json' -d '
{
  "query": {
    "match": {
      "attributes.address": "US"
    }
  }
}'
# Expected: returns doc 2

# ============================================================
# Step 7: Verify address.city is NOT treated as multi-field of address
#          (the isMultiField fix) — _exists_ query should work
# ============================================================
curl -X GET "localhost:9200/test-disable-objects/_search?pretty" -H 'Content-Type: application/json' -d '
{
  "query": {
    "exists": {
      "field": "attributes.address.city"
    }
  }
}'
# Expected: returns doc 1

# ============================================================
# Step 8: Verify address.city is independently searchable via query_string
#          (unqualified-style search using query_string syntax)
# ============================================================
curl -X GET "localhost:9200/test-disable-objects/_search?pretty" -H 'Content-Type: application/json' -d '
{
  "query": {
    "query_string": {
      "query": "attributes.address.city:Austin"
    }
  }
}'
# Expected: returns doc 1

# ============================================================
# Step 9: Negative test — searching address.city for "US" should NOT match
# ============================================================
curl -X GET "localhost:9200/test-disable-objects/_search?pretty" -H 'Content-Type: application/json' -d '
{
  "query": {
    "match": {
      "attributes.address.city": "US"
    }
  }
}'
# Expected: returns 0 hits

# ============================================================
# Step 10: Term query to verify exact field resolution
# ============================================================
curl -X GET "localhost:9200/test-disable-objects/_search?pretty" -H 'Content-Type: application/json' -d '
{
  "query": {
    "term": {
      "attributes.address.city.keyword": "Austin"
    }
  }
}'
# Expected: returns doc 1 (if dynamic mapping created .keyword sub-field)

# ============================================================
# Cleanup
# ============================================================
curl -X DELETE "localhost:9200/test-disable-objects"

```

server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
```diff
@@ -3639,6 +3639,92 @@ public void testParseDocumentWithDocumentInputAndNestedFields() throws Exception
...
+        }
+        assertNotNull(mapperService.fieldType("attributes.address.city"));
+
+        ParsedDocument doc2 = mapperService.documentMapper()
+            .parse(source(b -> { b.startObject("attributes").field("address", "US").endObject(); }));
+        if (doc2.dynamicMappingsUpdate() != null) {
+            merge(mapperService, dynamicMapping(doc2.dynamicMappingsUpdate()));
+        }
+
+        assertNotNull("address should be resolvable", mapperService.fieldType("attributes.address"));
```
cwperks: Can we assert what the fieldType in these tests should be? I think that would be a stronger assertions than simply checking for non-null.

server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
```diff
@@ -3639,6 +3639,92 @@ public void testParseDocumentWithDocumentInputAndNestedFields() throws Exception
...
+
+        MapperService ms2 = createMapperService(topMapping(b -> {
+            b.field("dynamic", false);
+            b.startObject("properties");
+            b.startObject("attributes").field("type", "object").field("dynamic", true).field("disable_objects", true).endObject();
+            b.endObject();
+        }));
+
+        ParsedDocument docA = ms2.documentMapper().parse(source(b -> { b.startObject("attributes").field("address", "US").endObject(); }));
+        if (docA.dynamicMappingsUpdate() != null) merge(ms2, dynamicMapping(docA.dynamicMappingsUpdate()));
```
cwperks: Is this conditional always true or something true? If its always true, then would we need the check here?

cwperks (commented):
Do we have any integration tests for disable_objects? It would probably be a good idea to have those in addition to these unit tests.

server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
mohit10011999: The assertions are now stronger - they verify not just that the field type is non-null, but also that it's specifically a "text" field type, which confirms the dynamic mapping created the correct field type for dotted fields under disable_objects.

server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
mohit10011999: Replaced all if (doc.dynamicMappingsUpdate() != null) guards with explicit assertNotNull assertions + comments explaining why we expect a dynamic mapping update (the fields aren't pre-defined in the mapping).
This makes the test intent clear: we expect the dynamic update to be produced, and if it's unexpectedly null, we get a descriptive failure instead of silently skipping the merge

server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
cwperks: nit: can we combine the 2 lined assertions into a single assert?

i.e.

```
assertNotNull("address should be resolvable", mapperService.fieldType("attributes.address"));
assertEquals("text", mapperService.fieldType("attributes.address").typeName());
```

can that simply be `assertEquals` with a message like `address should be text` since that also implies its non-null.

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -362,17 +374,16 @@ private static void handleDisableObjectsMapping(List<ObjectMapper> parentMappers
 
         for (int i = 0; i < disableObjectsIndex; i++) {
             Mapper child = current.getMapper(fullNameParts[i]);
-            if (child instanceof ObjectMapper om) {
```
cwperks, line 365: nit: change this back

server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
mohit10011999: combine all the assertNotNull + assertEquals pairs into single assertEquals calls with descriptive messages. If fieldType() returns null, assertEquals will fail with an NPE or a clear mismatch - effectively covering the null check too.

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
mohit10011999: Done. Changed the two-line instanceof + cast back to the Java 16+ pattern matching syntax

mohit10011999:
this scenario would be likely to fail
Please add coverage for root-level disable_objects prefix conflicts:

```

- root disable_objects=true
- dynamically add `address.city`
- dynamically add `address`
- assert both `fieldType("address.city")` and `fieldType("address")` are resolvable
- assert `address.city` is not treated as a multi-field of `address`

```


The root-level disable_objects prefix conflict scenario was indeed failing. The isMultiField() method in MappingLookup wasn't accounting for root-level disable_objects because the root mapper is explicitly excluded from the objectMappers map.

Fix applied:


1. Added a rootDisableObjects boolean field to MappingLookup
2. Passed mapping.root().disableObjects() from the fromMapping factory method
3. Added an early return false in isMultiField() when rootDisableObjects is true


Added a UT testing this case and now succeeded, before the fix it failed

cwperks:
@mohit10011999 one other case to consider, what if attributes is marked `disabled_objects: true` and following is ingested:

```
{
  "attributes": {
    "address": {
      "city": "Austin"
    }
  }
}
```

Would it be expected to translated the nested object into flat object?

i.e. expect the mapping to have a field for `attributes.address.city?`

mohit10011999:
https://github.com/opensearch-project/OpenSearch/pull/21929#issuecomment-4676677544



This is documented in the disable_objects docs (https://docs.opensearch.org/latest/mappings/mapping-parameters/disable-objects/) under the "Example: Index-level configuration" section, which shows that nested JSON like `{ "metrics": { "cpu": { "usage": 0.65 } } }` is automatically flattened to `metrics.cpu.usage` when disable_objects is enabled

server/src/test/java/org/opensearch/index/mapper/DocumentParserTests.java
```diff
@@ -3639,6 +3639,144 @@ public void testParseDocumentWithDocumentInputAndNestedFields() throws Exception
...
+        assertEquals("address should be text", "text", mapperService.fieldType("address").typeName());
+
+        // Assert address.city is NOT treated as a multi-field of address
+        assertFalse(
+            "address.city should not be a multi-field of address",
+            mapperService.documentMapper().mappers().isMultiField("address.city")
+        );
+    }
+
+    public void testDisableObjectsMappingRecoveryWithPrefixConflict() throws Exception {
```
cwperks, line 3736: I think its possible to define these with json like so:

```
MapperService mapperService = createMapperService(topMapping(b -> {
   b.field("dynamic", false);
   b.startObject("properties");
   b.startObject("attributes").field("type", "object").field("dynamic", true).field("disable_objects", true).endObject();
   b.endObject();
}));
```
Can become:

```
MapperService mapperService = createMapperService(topMapping("""
   {
     "dynamic": false,
     "properties": {
       "attributes": {
         "type": "object",
         "dynamic": true,
         "disable_objects": true
       }
     }
   }
   """));
```

And the source(b -> {...}) calls can similarly use JSON text blocks:

```
ParsedDocument doc1 = mapperService.documentMapper().parse(source("""
   {
     "attributes": {
       "address.city": "Austin",
       "address.state": "Texas"
     }
   }
   """));
```

mohit10011999: Used JSON text blocks for test document sources

server/src/main/java/org/opensearch/index/mapper/MappingLookup.java
```diff
@@ -278,7 +282,21 @@ public Map<String, ObjectMapper> objectMappers() {
...
-        return sourceParent != null && fieldMappers.containsKey(sourceParent);
+        if (sourceParent == null || !fieldMappers.containsKey(sourceParent)) {
+            return false;
+        }
+        // When disable_objects is true, dotted fields are independent flat fields, not multi-fields.
+        // Check if the root or any ancestor ObjectMapper has disable_objects enabled.
+        if (rootDisableObjects) {
+            return false;
+        }
+        for (String ancestor = parentObject(field); ancestor != null; ancestor = parentObject(ancestor)) {
```
cwperks, line 293: For my own understanding, is the traversal up the parents necessary here?

cwperks (approved):

server/src/main/java/org/opensearch/index/mapper/MappingLookup.java
mohit10011999: You are right that the immediate parent ObjectMapper (the one with `disable_objects: true`) is always the relevant one. But `parentObject()` is a simple string operation - it strips the last dot segment from the field name. It doesn't know which segments are actual ObjectMappers. So for `attributes.address.city`, the immediate parent in string terms is `attributes.address`, which isn't an ObjectMapper.

The loop is needed because we have to walk up from the field's string-based parent until we find an actual ObjectMapper in the objectMappers map. The first ObjectMapper we encounter going up is the `disable_objects` parent.

PR #21930, opened 2026-06-01
Limit force-execution in search thread pool via configurable queue threshold

### LINKED ISSUE #19129: [Feature Request] Limit the force put on Search Queue

We recently saw an issue where the Search Queue size was set at `1000` but due to heavy usage of `_msearch` requests the queue size went upto `80_000` even after back-pressure kicking in and stopping requests.

This led us to investigate the cases in which the queue size can be increased beyond desired capacity from the code base.

The search queue is based on the `ResizableBlockingQueue` 

Starting with the method used for [force put into the queue here](https://github.com/opensearch-project/OpenSearch/blob/2.15.0/server/src/main/java/org/opensearch/common/util/concurrent/SizeBlockingQueue.java#L134)

This only happens from `OpenSearchAbortPolicy` if the execution of a [AbstractRunnable is rejected here](https://github.com/opensearch-project/OpenSearch/blob/2.15.0/server/src/main/java/org/opensearch/common/util/concurrent/OpenSearchAbortPolicy.java#L58) and it is eligible for forced execution. 

So, when can a Runnable be eligible for forced execution ? Multiple cases in OpenSearch Core as follows : 

### Indexing

- Acquiring Shard Operation Permit

Force execution is true for cases like [acquiring replica operation permit here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/shard/IndexShard.java#L4560) There can be others but not recording them at the moment.

- Whenever the Refresh Interval changes either via the Cluster Default or for an Index via it's setting. 

The thread submitted to refresh engine [is eligible for forced execution](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/IndexService.java#L1261)

- Bulk Shard request execution on a primary 

When a BulkShardRequest is rejected from the thread pool [then the outstanding request is force executed](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/bulk/TransportShardBulkAction.java#L517)

There are other instances like while [processing bulk index ingestion request](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/bulk/TransportBulkAction.java#L1012) AND during [shard operation on Replica here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/replication/TransportWriteAction.java#L314)

### Discovery 

When a new peer is found it updates the Found Peers [by force put to Generic Thread Pool ](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/discovery/PeerFinder.java#L334) Unsure if peer means only other cluster manager's or any node in the cluster. 

### Search

AbstractAsyncSearchAction is used to force put into the queue when [we execute search related tasks by forking here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/search/AbstractSearchAsyncAction.java#L410)

This can be triggered from different cases : 

1.  When all shards have been tried for execution, a `NoShardAvailableActionException` is thrown here by [forking with a runnable here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/search/AbstractSearchAsyncAction.java#L290)

2.  After a search phase completes or fails on a Shard OR there is an exception connecting to node containing the target shard, the next phase is [handled by forking here](https://github.com/opensearch-project/OpenSearch/blame/main/server/src/main/java/org/opensearch/action/search/AbstractSearchAsyncAction.java#L911) if and only if : 
         -   `max_concurrent_shard_requests` is less than the eligible shards for the request. 
         -    The response was received ( via callback ) on the same thread that submitted the request.  

All the nodes were consistently experiencing high CPU and GC pauses which would have led to increase in the queue size by force put due to point 2 in search.

Logs that indicate that queue really reached the size mentioned

```
Caused by: OpenSearchException[rejected execution of org.opensearch.common.util.concurrent.TimedRunnable@66c5f204 on 
QueueResizingOpenSearchThreadPoolExecutor[name = <NodeIdHere>/search, queue capacity = 1000, min queue 
capacity = 1000, max queue capacity = 1000, frame size = 2000, targeted response rate = 1s, task execution EWMA = 1.4s, adjustment amount 
= 50, org.opensearch.common.util.concurrent.QueueResizingOpenSearchThreadPoolExecutor@ce5417f[Running, pool size = 13, active threads 
= 13, queued tasks = 80692, completed tasks = 879184]]]; nested: OpenSearchRejectedExecutionException[rejected execution of 
org.opensearch.common.util.concurrent.TimedRunnable@66c5f204 on QueueResizingOpenSearchThreadPoolExecutor[name = 
a21da74cd90d08042953e1145adbaea3/search, queue capacity = 1000, min queue capacity = 1000, max queue capacity = 1000, frame size = 
2000, targeted response rate = 1s, task execution EWMA = 1.4s, adjustment amount = 50, 
org.opensearch.common.util.concurrent.QueueResizingOpenSearchThreadPoolExecutor@ce5417f[Running, pool size = 13, active threads = 13, 
queued tasks = 80692, completed tasks = 879184]]];
```

In such cases, an end user is forced to restart nodes or wait for the queue to clear up. 

Proposal is to put some limit on the SizeBlockingQueue only for Search ThreadPool in case for forcePut to account for `_msearch` requests so that such cases can be avoided by early rejecting the query.

### Related component

Search:Resiliency

_No response_

### Additional context

_No response_

pado0:
### Description
`AbstractSearchAsyncAction.fork()` unconditionally set `isForceExecution() = true`,
bypassing the search thread pool's queue capacity limit on every shard task.

Under heavy `_msearch` load, this allows unbounded queue growth (observed: 80,000+
vs. configured 1,000), effectively disabling the backpressure that
`OpenSearchThreadPoolExecutor` is designed to provide.

This PR introduces a dynamic cluster setting that gives operators control over when
force-execution is applied:

- **`search.force_execution_queue_threshold`** (default: `-1`, minimum: `-1`)
  - `-1` — always force-execute (preserves existing behavior, backward-compatible)
  - `N ≥ 0` — force-execute only when queue size `< N`;
    reject via `onRejection()` when queue size `≥ N`

The threshold check is scoped to `OpenSearchThreadPoolExecutor` only.
Any other `Executor` implementation retains unconditional force-execution.

### Approach
This PR implements **Option 2** from the issue discussion — keeping existing
`forcePut` semantics intact for all other subsystems (indexing, discovery,
refresh), while giving operators a configurable threshold specific to the
search execution path.
https://github.com/opensearch-project/OpenSearch/issues/19129#issuecomment-3555944694

The restriction is applied at `AbstractSearchAsyncAction.fork()` rather than
at the `SizeBlockingQueue` level, which keeps the change minimal and avoids
unintended side effects on non-search thread pools.

As the design direction has not yet been finalized in the issue discussion, this PR is opened to make the proposal more concrete and invite early feedback. Happy to adjust the approach based on maintainer input.

### Backward Compatibility
Default value is `-1`, which preserves unconditional force-execute behavior.
No behavior change unless the operator explicitly configures the threshold.

### Related Issues
Resolves #19129

### Check List


pado0:
Tested in local with script https://github.com/pado0/test-scripts/blob/main/opensearch/pr-21930-force-put-reproduce.py

Just to clarify the intended behavior: this threshold is a best-effort guardrail, not a strict hard limit. Since the queue size is checked before the task is submitted, concurrent submissions may temporarily push the queue above the configured threshold. The goal here is to prevent unbounded queue growth from unconditional force execution rather than guarantee an exact queue size cap.

1. set force_execution_queue_threshold: 1500 (queue size doesn't exceed 1500)
image

2. set force_execution_queue_threshold: -1
image

image

pado0:
Hi, @sandeshkr419 I was assigned this issue quite a while ago, and I've finally gotten around to opening a PR. I'd really appreciate it if you could take a look when you have some time. Thanks!

PR #21932, opened 2026-06-01
Fix array_length return type mismatch

vinaykpud:
## Description

Calcite types `array_length` as `INTEGER` (i32), but DataFusion's UDF returns `BIGINT` (i64). On a single shard the disagreement is invisible. On a multi-shard plan, the cross-fragment schema check sees `Int32 ≠ Int64` and fails the query with HTTP 500.

```
ArrayFunctionIT > testArrayLength FAILED
  Substrait error: Field 'len' has different type (Int32) than table schema (Int64)
```

## Fix

Mirror the `IntegerRoundingCastAdapter` pattern from #21914. New adapter `IntegerReturnWideningCastAdapter` rewrites:

```
array_length(arr) : INTEGER
```

into:

```
CAST( array_length(arr) : BIGINT  AS INTEGER )
```

- **Inner BIGINT** matches what DataFusion actually emits, so the cross-shard schema check passes.
- **Outer CAST back to INTEGER** preserves the original return type so parent operators see no change.

Registered against `ScalarFunction.ARRAY_LENGTH`. Generalised to `INTEGER`/`SMALLINT`/`TINYINT` returns, so future i32-vs-i64 mismatches are a one-line registration.

## Testing

- New unit tests in `IntegerReturnWideningCastAdapterTests` (wrap shape, nullability, no-op for BIGINT/non-integer, recursion safety).
- `ArrayFunctionIT.testArrayLength` passes on the 2-shard default tree.
- `MathFunctionIT.testFloorOnIntColumn` / `testCeilOnIntColumn` (the #21914 regression tests) still pass.

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
@@ -599,6 +599,7 @@ public class DataFusionAnalyticsBackendPlugin implements AnalyticsSearchBackendP
                 return Map.ofEntries(
                     Map.entry(ScalarFunction.ARRAY, new MakeArrayAdapter()),
                     Map.entry(ScalarFunction.ARRAY_JOIN, new ArrayToStringAdapter()),
+                    Map.entry(ScalarFunction.ARRAY_LENGTH, new IntegerReturnWideningCastAdapter()),
                     Map.entry(ScalarFunction.ARRAY_SLICE, new ArraySliceAdapter()),
                     Map.entry(ScalarFunction.ITEM, new ArrayElementAdapter()),
                     Map.entry(ScalarFunction.MVFIND, new MvfindAdapter()),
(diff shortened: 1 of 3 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java
```diff
@@ -34,6 +34,7 @@
 import org.opensearch.analytics.spi.FilterDelegationHandle;
 import org.opensearch.analytics.spi.FragmentConvertor;
 import org.opensearch.analytics.spi.FragmentInstructionHandlerFactory;
+import org.opensearch.analytics.spi.IntegerReturnWideningCastAdapter;
```
sandeshkr419, line 37: just an import statement here?

vinaykpud: No, we need to register, why it got removed, will add it back.

sandeshkr419 (approved):

PR #21935, opened 2026-06-02
[Analytics Engine] Add limit on max shards per query for alias, index patterns, multi index queries

mch2:
### Description
This adds a hard cap on the amount of shards that can be queried for alias, index pattern, and multi-index queries.  This is a temp work around in the absence of can-match phase.

This also makes pendingExecutions max per node queries a dynamic setting.

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/ShardTargetResolver.java
```diff
@@ -39,16 +42,22 @@ public class ShardTargetResolver extends TargetResolver {
...
+        this.maxShardsPerQuery = clusterService.getClusterSettings().get(AnalyticsQuerySettings.MAX_SHARDS_PER_QUERY);
         if (this.indexName == null) {
             throw new IllegalArgumentException("ShardTargetResolver: no OpenSearchTableScan found in fragment");
         }
     }
 
+    public void setMaxShardsPerQuery(int maxShardsPerQuery) {
+        this.maxShardsPerQuery = maxShardsPerQuery;
+    }
+
```
Bukhtawar: We need to wire in settings update consumer?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/ShardTargetResolver.java
```diff
@@ -58,6 +67,23 @@ public List<ExecutionTarget> resolve(ClusterState clusterState, @Nullable Object
         String[] concreteNames = resolution.concreteIndexNames().toArray(new String[0]);
         GroupShardsIterator<ShardIterator> shardIterators = clusterService.operationRouting()
             .searchShards(clusterState, concreteNames, null, null);
+        // TODO: Hard rejection in absence of a can-match phase. Without can-match to prune
+        // non-matching shards upfront, an unbounded fan-out can overload the coordinator.
+        // Once can-match is implemented, this limit can be relaxed or applied post-pruning.
+        int shardCount = shardIterators.size();
```
Bukhtawar, line 81: This might include replicas as well?

mch2: replicas would be inside of the individual iterators, this gives the shard count.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/ShardTargetResolver.java
mch2: the target resolver is re-constructed per query.  So added a simple setting lookup instead of storing the limit value and passing it down.

mch2: updated.

Bukhtawar (approved):
lgtm

PR #21937, opened 2026-06-02
Add search support for handling deleted documents

RS146BIJAY:
### Description

For a Composite format where one of the format is Lucene (or lucene only format), deleted documents are tracked by Lucene's liveDocs bitmap. When a query's filter contains at least one predicate naturally served by Lucene, that predicate's Weight.scorer(...).iterator() already excludes deletes — so deleted docs are filtered out. Adding support for updates/deltes by attaching a live docs delegation when there are deletes.

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
@@ -1237,7 +1237,6 @@ public final class NativeBridge {
             try {
                 result = call.invoke(EXECUTE_WITH_CONTEXT, sessionCtxPtr, plan, planLen);
             } finally {
-                // Rust took ownership via Box::from_raw; do not let doClose() double-free.
                 sessionContext.markConsumed();
             }
             listener.onResponse(result);
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
@@ -22,9 +22,11 @@ import org.opensearch.analytics.exec.action.FragmentExecutionRequest;
 import org.opensearch.analytics.exec.task.AnalyticsShardTask;
 import org.opensearch.analytics.spi.AnalyticsSearchBackendPlugin;
 import org.opensearch.analytics.spi.BackendExecutionContext;
+import org.opensearch.analytics.spi.DelegatedExpression;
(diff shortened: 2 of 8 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/nativelib/NativeBridge.java
expani, line 1240: Looks unintentional ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -448,9 +442,46 @@ private ShardScanExecutionContext buildContext(
...
+                break;
+            }
+        }
+
+        // In case registered format is parquet only format, return null.
+        if (luceneBackendId == null) {
+            return null;
+        }
+        try {
+            var matchAll = new org.opensearch.index.query.MatchAllQueryBuilder();
```
expani, line 473: This is leaking Lucene `MatchAllQuery` into core. 
We can make it a API to Lucene Backend.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -324,25 +325,22 @@ private FragmentResources startFragment(FragmentExecutionRequest request, Resolv
             // TODO: currently assumes single accepting backend. When multiple accepting backends exist
             // (e.g., Lucene + Tantivy), group expressions by acceptingBackendId and create one handle per group.
             DelegationDescriptor delegation = resolved.plan.getDelegationDescriptor();
+            // Inject synthetic MATCHALL delegation when the shard has soft-deleted documents
+            // but no real delegation exists. The MATCHALL scorer iterates all docs and
+            // collectDocs() checks liveDocs — filtering deleted rows via the existing FFM path.
+            if (delegation == null && ctx.hasDeletedDocs()) {
```
aasom143, line 331: MatchAll delegation will not be needed for SingleCollector shape considering there is at least one correctness delegation. 
Following are the cases where we need MatchAll
* With pure DF query, we need to AND with MatchAll.
* If SingleCollector contains all the performance delegation, we need to AND with MatchAll considering case if expression for any RG don't take the lucene delegation path.
* With Tree shape query, we need to AND with MatchAll.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
aasom143: Hi @expani , Can you please help with this comment?

PR #21940, opened 2026-06-02
Fix temp lucene_gen folder deletion

rayshrey:
### Description

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
@@ -28,6 +28,7 @@ import org.opensearch.be.lucene.LuceneFieldFactoryRegistry;
 import org.opensearch.be.lucene.LuceneReader;
 import org.opensearch.be.lucene.merge.LuceneMerger;
 import org.opensearch.common.annotation.ExperimentalApi;
+import org.opensearch.common.util.io.IOUtils;
 import org.opensearch.index.engine.dataformat.DataFormat;
 import org.opensearch.index.engine.dataformat.DocumentInput;
 import org.opensearch.index.engine.dataformat.IndexingExecutionEngine;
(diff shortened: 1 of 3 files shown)


sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
```diff
@@ -121,13 +121,16 @@ public LuceneIndexingExecutionEngine(
 
         this.luceneMerger = new LuceneMerger(sharedWriter, dataFormat, store.shardPath().resolveIndex());
 
-        // Create the lucene subdirectory if it doesn't exist
+        // Create the lucene subdirectory if it doesn't exist, or clear stale contents
+        // from a prior engine lifecycle. Any data here is either already hardlinked into
+        // index/ (via addIndexes) or will be replayed from the translog on recovery.
+        if (Files.isDirectory(baseDirectory)) {
+            tryDeleteDirectory(baseDirectory);
+        }
```
Bukhtawar, line 130: I am wondering if the deletion logic should sit in `LuceneWriter` instead?

LuceneWriter — owns creation AND deletion of its temp directory:
  - cleanup(): IOUtils.rm(tempDirectory) — called externally when safe

  LuceneIndexingExecutionEngine — knows when deletion is safe:
  - createWriter(): creates writer, adds to pendingCleanup
  - refresh(): after addIndexes hardlinks files → drains pendingCleanup, calls writer.cleanup() on each

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
```diff
@@ -180,6 +184,8 @@ public Writer<LuceneDocumentInput> createWriter(WriterConfig config) {
                 getChildWriterSortConfiguration(),
                 activeWriters
             );
+            pendingCleanup.put(config.writerGeneration(), writer);
```
mgodwan, line 187: Why not handle it directly in writer close?

rayshrey: The issue with close is that during rollback we call the writer's close first and later refresh is called for that writer, so if we cleanup during close, refresh fails due to that rolled back writer's segment files being missing.

Bukhtawar (approved):

PR #21948, opened 2026-06-02
[Analytics-Engine] Fix query shape and execution in case of multiple filter in the shape

alchemist51:
### Description

When an analytics fragment contains **more than one** `OpenSearchFilter`, the planner and the data-node executor disagreed on which filter drives the indexed scan. PR #21867 fixed the merge of two filters that the planner already knew were adjacent, but a `head N` (or any fetch-bearing `Sort`) between two `where` clauses blocks `CoreRules.FILTER_MERGE` and leaves two filters around the limit. Two symmetric off-by-one bugs in that case — one in Java, one in Rust.

### Reproducer

PPL:
```
source=calcs | where key='key01' | head 1 | where earliest('2004-07-26 00:00:00', `datetime0`) | stats count() as cnt
```

After marking, the fragment carries two filters around the `Sort fetch=1`:
```
OpenSearchAggregate(group=[{}], cnt=[COUNT()])
  OpenSearchFilter(condition=[ANNOTATED_PREDICATE(id=1, backends=[datafusion],
                              EARLIEST('2004-07-26 00:00:00':VARCHAR, $8))])
    OpenSearchSort(fetch=[1])
      OpenSearchFilter(condition=[ANNOTATED_PREDICATE(id=0, backends=[lucene, datafusion],
                                  =($14, 'key01'))])
        OpenSearchTableScan(table=[[calcs]])
```

The expected counts:
- `key='key01'` → datetime0 = 2004-07-26 22:48:48 → ≥ threshold → `cnt = 1`
- `key='key00'` → datetime0 = 2004-07-19 22:48:48 → < threshold → `cnt = 0`

### Bug 1 — planner (Java)

`FragmentConversionDriver.convertStage` derived the filter tree shape from one Filter via `RelNodeUtils.findNode` (single-input chain, **first** match — the upper one). The upper filter only contains the native `EARLIEST(...)` predicate, so the deriver returned `treeShape = NO_DELEGATION`.

`IntraOperatorDelegationBytes`, however, walks the **entire** fragment, so it picked up the lower filter's delegated `key='key01'` and reported `delegated.size() == 1`.

Pair sent to the data node:
```
treeShape = NO_DELEGATION (ordinal 0)
delegated = 1
```

Rust's classification mapping:
```rust
match (config.tree_shape, config.delegated_predicate_count) {
    (1, _) => FilterClass::SingleCollector,   // CONJUNCTIVE
    (2, _) => FilterClass::Tree,              // INTERLEAVED
    _      => FilterClass::None,              // ← (0, 1) falls through
}
```

Falls through to `FilterClass::None`, then with `emit_row_ids = false` errors:

```
java.io.IOException: Query execution with session context failed
  Caused by: java.lang.RuntimeException: Execution error: Execution error:
    execute_indexed_query called with no index_filter(...) in plan
```

#### Fix

- New helper `RelNodeUtils.findAllNodes` (sibling to `findNode`) walks the same single-input chain top-down and returns every match.
- New `FilterTreeShapeDeriver.deriveCombined(filters, drivingBackendId, fuseDualViable)` folds per-filter shapes:
  - any `INTERLEAVED` ⇒ `INTERLEAVED`
  - else any `CONJUNCTIVE` ⇒ `CONJUNCTIVE`
  - else `NO_DELEGATION`
- `convertStage` now uses both, so the shape covers exactly the same scope as `delegationBytes`.

For our example: upper filter → `NO_DELEGATION`, lower → `CONJUNCTIVE`, combined → `CONJUNCTIVE`. Pair `(CONJUNCTIVE, 1)` → `FilterClass::SingleCollector`.

### Bug 2 — executor (Rust)

After Bug 1 is fixed, the data node decodes the Substrait into:

```
Aggregate
  Filter: calcs.datetime0 >= TimestampMillisecond(...)   ← upper (native)
    Limit: skip=0, fetch=1
      Filter: delegated_predicate(Int32(0))              ← lower (Lucene-delegated)
        TableScan: calcs ...
```

`indexed_executor::execute_indexed_with_context_inner` calls `extract_filter_expr(&logical_plan)` to choose the Filter that drives the `IndexedTableProvider`'s row-selection callback. The original implementation:

```rust
pub fn extract_filter_expr(plan: &LogicalPlan) -> Option<Expr> {
    match plan {
        LogicalPlan::Filter(filter) => Some(filter.predicate.clone()),
        _ => {
            for child in plan.inputs() {
                if let Some(expr) = extract_filter_expr(child) {
                    return Some(expr);
                }
            }
            None
        }
    }
}
```

is **also** top-down-first-match — it picks the **upper** (`datetime0 >= ...`) Filter, builds a `BoolNode` with **no** `Collector` / `DelegationPossible` leaves, and `single_collector_id` returns `None`. The indexed path runs without any Lucene narrowing — `QueryShardExec` returns every row, `GlobalLimitExec(fetch=1)` keeps an arbitrary one, the upper `FilterExec` applies post-hoc.

For `key='key00'` the test asserts `cnt = 0`; with the bug live the wrong row picked by the limit makes it `1`.

#### Fix

Walk to the **deepest** Filter — the one immediately above the TableScan — because that is the filter whose `delegated_predicate(...)` markers pair with the IndexedTableProvider callbacks. Higher filters are still applied by their own `FilterExec`.

```rust
pub fn extract_filter_expr(plan: &LogicalPlan) -> Option<Expr> {
    let mut deepest: Option<Expr> = None;
    fn walk(plan: &LogicalPlan, deepest: &mut Option<Expr>) {
        if let LogicalPlan::Filter(filter) = plan {
            *deepest = Some(filter.predicate.clone());
        }
        for child in plan.inputs() {
            walk(child, deepest);
        }
    }
    walk(plan, &mut deepest);
    deepest
}
```

### Why the printed plans look identical with vs. without the fix

The fix changes which `Filter::predicate` the `IndexedTableProvider` is configured with. That callback is internal to `QueryShardExec`; it doesn't appear as a separate node in `LogicalPlan` / `PhysicalPlan` dumps. Diagnose at runtime via Lucene FFM `createCollector` / `collectDocs` upcalls firing for the delegated annotation id (present with the fix, absent without it).

### Related Issues

Surfaced via `ConditionalFunctionsIT.testEarliestAbsoluteLiteralSelectsSpecificRows`, which was previously gated with `@AwaitsFix`. This PR removes the gate.

### Check List


expani:
This seems to the case of Filter pushdown and Merging in Calcite. 

Essentially, we should only be worrying about one Filter operator in FragmentConversionDriver. 

The fix needs to be happen in the logical optimization of Calcite by adding a few more rules. 

```
FilterMergeRule -> Will merge such filters together. 
FilterProjectTransposeRule -> Pushes filter past project 
FilterJoinRule -> Pushes filter past the Join
FilterSortTransposeRule -> Pushes filter past the Sort
```

We need to add them before the Marking phase starts and way before Fragment Conversion.

This way we ensure that the existing logic always sees one filter.

expani (approved):

mch2 (dismissed):

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/MultiSourceJoinsPplIT.java
```diff
@@ -19,6 +20,7 @@ protected Dataset getDataset() {
         return MultiSourceJoinsTestHelper.DATASET;
     }
 
+    @AwaitsFix(bugUrl = "https://github.com/opensearch-project/OpenSearch/pull/21948")
     public void testMultiSourceJoinsPplQueries() throws Exception {
```
mch2, line 21: we need to mute only the failing queries and not the whole suite.

alchemist51: Addressed in ce0f7e93942 — switched all the affected suites from class-level @AwaitsFix to per-query getSkipQueries() entries. Only the queries that hit the multi-filter rejection are skipped now; the rest of each suite runs.

Failing query numbers per suite:
- AppLogs Q3, Q10
- ExtensiveCoverage Q1
- KubernetesLogs Q6
- MultiSourceJoins Q5
- OtelLogs Q19, Q20, Q38, Q39
- PerformanceMetrics Q7
- SecurityLogs Q10

For the two ConversionFunctions tests (testConvertMktimeThenWhereConstantPredicate*) and DateTimeScalarFunctionsIT.testDateAddInWherePredicate, I moved `head 1` to the end of the pipeline so FILTER_MERGE can collapse the two wheres — those are individual @Test methods so per-query muting doesn't apply. RareCommandIT.testRareCommandUseNullFalse and TopCommandIT.testTopCommandUseNullFalse are also single-method classes and stay on @AwaitsFix.

PR #21950, opened 2026-06-02
Only warn about disabled gateway.auto_import_dangling_indices when explicitly set

### LINKED ISSUE #21330: [BUG] warning for `gateway.auto_import_dangling_indices` with default settings
### Describe the bug

the following warning is printed to the log when starting a cluster with default settings:
```json
{"type": "server", "timestamp": "2026-04-23T07:17:31,166+0000", "level": "WARN", "component": "o.o.g.DanglingIndicesState", "cluster.name": "test-opensearch", "node.name": "test-opensearch-0", "message": "gateway.auto_import_dangling_indices is disabled, dangling indices will not be automatically detected or imported and must be managed manually" }
```

according to [the documentation](https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/discovery-gateway-settings/) the default for this setting is `true` => the warning shouldn't show up.
however, in reality, the default is `false`:
https://github.com/opensearch-project/OpenSearch/blob/dec02f901209d6b4b90bb4aeda76ad34f4ec13ba/server/src/main/java/org/opensearch/gateway/DanglingIndicesState.java#L79-L84

### Related component

Other

### To Reproduce

start a fresh cluster with the default config (i.e. `gateway.auto_import_dangling_indices` is not defined in the config) and check the logs

### Expected behavior

IMHO the default setting shouldn't lead to a warning. whether that means reducing the WARN to an INFO or changing the default i cannot judge.

also the documentation needs to be aligned with the code (or the code with the documentation 🙂)

### Additional Details

**Plugins**
n/a

**Screenshots**
n/a

 - OS: Linux
 - Version OpenSearch 3.6.0

**Additional context**
n/a

Lawson-Darrow:
### Description

`DanglingIndicesState` logs a `WARN` on startup whenever auto-import of dangling indices is disabled:

> gateway.auto_import_dangling_indices is disabled, dangling indices will not be automatically detected or imported and must be managed manually

The `gateway.auto_import_dangling_indices` setting is **deprecated** and **defaults to `false` (disabled)**, so this warning is emitted on *every* startup with the default configuration — pure noise about a deprecated, intentionally-off feature.

This change only emits the warning when the setting has been **explicitly set** by the user (`Setting.exists(settings)`). Operators who deliberately disable it still see the operational consequence; default configurations stay quiet.

An alternative would be to downgrade the message to `INFO`. I went with suppress-on-default because the reporter's concern is specifically that the *default* configuration shouldn't produce the warning — happy to switch to the downgrade if maintainers prefer it.

(Aside: the docs state the default is `true` while the code default is `false`; that documentation mismatch lives in the `documentation-website` repo and is out of scope here.)

### Related Issues

Resolves #21330

### Check List

  - Added `DanglingIndicesStateTests#testNoWarningWhenSettingUsesDefault` (no warning on default config) and `#testWarningWhenSettingExplicitlyDisabled` (warning still emitted when explicitly disabled).
- [x] All commits are signed off (DCO).
- [x] Commit changes are listed out in CHANGELOG.md file (the CHANGELOG is no longer used as of 3.6, so no entry is required).
- [x] Public documentation issue/PR created — N/A.

### DIFF
--- a/server/src/main/java/org/opensearch/gateway/DanglingIndicesState.java
+++ b/server/src/main/java/org/opensearch/gateway/DanglingIndicesState.java
@@ -42,6 +42,7 @@ import org.opensearch.cluster.metadata.Metadata;
 import org.opensearch.cluster.service.ClusterService;
 import org.opensearch.common.inject.Inject;
 import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
 import org.opensearch.common.util.concurrent.ConcurrentCollections;
 import org.opensearch.core.action.ActionListener;
 import org.opensearch.core.index.Index;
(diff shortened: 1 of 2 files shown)


Lawson-Darrow:
The `null` here is a gradle-check trigger/infra failure, not a test failure. The job fired with an empty gradle command and then logged `jq: parse error` 122 times over ~2 hours while polling Jenkins for a result that never came, ending in `Result: null`:

```
{"pr_from_sha": "4d8fa218...", "gradle_check_command": ""}
jq: parse error: Invalid numeric literal at line 1, column 7   (x122, 18:53 -> 20:54)
Result: null
```

No tests executed, so this is unrelated to the change (which only touches the dangling-indices startup warning and its test).

Issue #21900, opened 2026-05-30
[BUG] GET _aliases blocks transport_worker thread when many aliases carry filters


dzane17:
### Describe the bug

`GET _aliases` serializes the entire cluster's alias metadata on the Netty `transport_worker` thread that delivers the response. When many aliases carry filters (e.g. clusters using DLS), each filter is decompressed and re-parsed during serialization, and a single request can pin one event-loop thread for several seconds. While that thread is blocked, every connection it owns stalls — including unrelated `POST _aliases` calls — and the spike is invisible in node-level `CPUUtilization` because only one of N vCPUs is hot.

### Reproducing

Cluster with ~36k filtered aliases (e.g. ~12k tenants × 3 aliases each, each with a DLS filter). Issue `GET _aliases` from a polling client. Hot threads shows ~100% CPU on a single `transport_worker` thread inside `AliasMetadata$Builder.toXContent` → `XContentHelper.convertToMap` → `AbstractXContentParser.readMapEntries`.

### Where the cost is

Two issues stack:

1. **Serialization runs on the IO thread.** `TransportGetAliasesAction.executor()` returns `ThreadPool.Names.SAME` ([source](https://github.com/opensearch-project/OpenSearch/blob/8dadd37691b7db0589fcf63bb35ac8763c050835/server/src/main/java/org/opensearch/action/admin/indices/alias/get/TransportGetAliasesAction.java#L101-L104)), and `RestGetAliasesAction` builds the response inline via a plain `RestBuilderListener` ([source](https://github.com/opensearch-project/OpenSearch/blob/8dadd37691b7db0589fcf63bb35ac8763c050835/server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java#L203-L208)). The XContent build therefore runs on the transport worker.

2. **Each filter is decompressed + reparsed per call.** `AliasMetadata.Builder.toXContent` calls `XContentHelper.convertToMap(aliasMetadata.filter().uncompressed(), true)` ([source](https://github.com/opensearch-project/OpenSearch/blob/8dadd37691b7db0589fcf63bb35ac8763c050835/server/src/main/java/org/opensearch/cluster/metadata/AliasMetadata.java#L369)) for every alias, every request — O(aliases) decompress+parse+re-serialize per `GET _aliases`.

There's also no pagination on the API, so cost grows linearly with cluster alias count and clients have no way to chunk.

### Proposed fixes

[...]

PR #21954, opened 2026-06-02
Fix GET _aliases blocking transport thread on clusters with many filtered aliases

gingeekrishna:
## Summary

- `TransportGetAliasesAction.executor()` returns `SAME`, so alias metadata serialization (filter decompression + XContent build) runs on the Netty `transport_worker` thread
- On clusters with large numbers of filtered aliases (e.g. DLS tenants), a single `GET _aliases` call can pin one IO thread for several seconds and stall unrelated requests sharing that connection
- Mirrors the pattern already used by `RestGetMappingAction` (which has the same comment: *"the serialization of the raw mappings to XContent can be too slow to execute on an IO thread"*): fork the XContent build to `ThreadPool.Names.MANAGEMENT`

## Changes

- `RestGetAliasesAction`: inject `ThreadPool`, switch from `RestBuilderListener` to `RestActionListener`, fork serialization to `MANAGEMENT` thread pool
- `ActionModule`: pass `threadPool` to `RestGetAliasesAction` constructor

## Test plan

- [ ] Existing `RestGetAliasesActionTests` continue to pass (they test the static `buildRestResponse` method directly and are unaffected by the threading change)
- [ ] Manual verification: `GET _aliases` on a cluster with many filtered aliases no longer pins a `transport_worker` thread

Part of #21900


server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -200,10 +209,18 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
         // not always be available there (they may get replaced so retrieving request.aliases is not quite the same).
-        return channel -> client.admin().indices().getAliases(getAliasesRequest, new RestBuilderListener<GetAliasesResponse>(channel) {
+        return channel -> client.admin().indices().getAliases(getAliasesRequest, new RestActionListener<GetAliasesResponse>(channel) {
             @Override
-            public RestResponse buildResponse(GetAliasesResponse response, XContentBuilder builder) throws Exception {
-                return buildRestResponse(namesProvided, aliases, response.getAliases(), builder);
+            protected void processResponse(GetAliasesResponse response) {
+                // Serialization of alias metadata (filter decompression + XContent build) can be slow on clusters
+                // with many filtered aliases (e.g. DLS). Fork to MANAGEMENT to avoid blocking the transport thread.
+                threadPool.executor(ThreadPool.Names.MANAGEMENT)
```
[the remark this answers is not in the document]
gingeekrishna: Thanks for the suggestion! We're intentionally mirroring the pattern already established by `RestGetMappingAction`, which also uses `ThreadPool.Names.MANAGEMENT` for the same reason (serialization of raw metadata to XContent). Using the same pool keeps the two actions consistent. If the project decides to move mapping serialization to a different pool, aliases can follow the same change.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -200,10 +209,18 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
+                    .execute(ActionRunnable.wrap(this, l -> new RestBuilderListener<GetAliasesResponse>(channel) {
+                        @Override
+                        public RestResponse buildResponse(GetAliasesResponse r, XContentBuilder builder) throws Exception {
+                            return buildRestResponse(namesProvided, aliases, r.getAliases(), builder);
+                        }
+                    }.onResponse(response)));
```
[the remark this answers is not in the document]
gingeekrishna: Agreed the pattern looks unusual, but it's identical to what `RestGetMappingAction` uses — `ActionRunnable.wrap(this, l -> new RestBuilderListener<>(...).onResponse(response))`. The `l` parameter is the same listener as `this` (passed as the first argument to `wrap`); `ActionRunnable.wrap` ensures any exception thrown inside the lambda is routed to `this.onFailure()`. We've kept the same shape for consistency, but happy to refactor if the project prefers a different style.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -70,6 +73,12 @@
  */
 public class RestGetAliasesAction extends BaseRestHandler {
 
+    private final ThreadPool threadPool;
+
+    public RestGetAliasesAction(ThreadPool threadPool) {
```
dzane17: Removing the implicit no-arg constructor breaks `OpenSearchDashboardsModulePlugin.java:122`, which still calls `new RestGetAliasesAction()`. The build will fail.

Easiest fix: keep a no-arg constructor here that defers thread-pool resolution to request time (`client.threadPool()` inside `prepareRequest`), and have the new `(ThreadPool)` constructor delegate to it. That avoids touching the dashboards module and avoids breaking any out-of-tree plugin that compiles against this `@opensearch.api` class.

Alternative: update `OpenSearchDashboardsModulePlugin` to capture `ThreadPool` via `createComponents` and pass it in. More invasive and still leaves out-of-tree consumers broken.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -200,10 +209,22 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
+                    threadPool.executor(ThreadPool.Names.MANAGEMENT)
+                        .execute(ActionRunnable.wrap(this, l -> new RestBuilderListener<GetAliasesResponse>(channel) {
+                            @Override
+                            public RestResponse buildResponse(GetAliasesResponse r, XContentBuilder builder) throws Exception {
+                                return buildRestResponse(namesProvided, aliases, r.getAliases(), builder);
+                            }
+                        }.onResponse(response)));
+                } catch (Exception e) {
+                    onFailure(e);
+                }
```
dzane17: `try/catch` isn't needed. `RestActionListener.onResponse` already wraps `processResponse` in try/catch, and `OpenSearchThreadPoolExecutor` catches `RejectedExecutionException` internally for `AbstractRunnable` and routes it to `onRejection` without rethrowing.

So the catch block can never fire on rejection, and any other thrown exception was already handled. `RestGetMappingAction` (the action we're mirroring) doesn't have this wrapper either.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -200,10 +209,22 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
-        return channel -> client.admin().indices().getAliases(getAliasesRequest, new RestBuilderListener<GetAliasesResponse>(channel) {
+        return channel -> client.admin().indices().getAliases(getAliasesRequest, new RestActionListener<GetAliasesResponse>(channel) {
             @Override
-            public RestResponse buildResponse(GetAliasesResponse response, XContentBuilder builder) throws Exception {
-                return buildRestResponse(namesProvided, aliases, response.getAliases(), builder);
+            protected void processResponse(GetAliasesResponse response) {
+                // Serialization of alias metadata (filter decompression + XContent build) can be slow on clusters
+                // with many filtered aliases (e.g. DLS). Fork to MANAGEMENT to avoid blocking the transport thread.
+                try {
+                    threadPool.executor(ThreadPool.Names.MANAGEMENT)
```
dzane17: Before this PR, the response was serialized inline on the transport thread, so there was no place for stale work to queue — `cluster_manager_timeout` covered the only wait (the cluster-manager round trip).

After this PR, work can sit in the MANAGEMENT queue (1-5 threads, unbounded, never rejects) while the client times out and retries. Without a check, the pool will eventually drain every queued retry to completion against a closed connection — wasting CPU on serialization that no one will read.

Add a guard against this like in `RestGetMappingAction.java:121,129-131`:
```java
final long startTimeMs = threadPool.relativeTimeInMillis();
// ...
if (threadPool.relativeTimeInMillis() - startTimeMs > timeout.millis()) {
    throw new OpenSearchTimeoutException("Timed out getting aliases");
}
```

dzane17 (commented):
The existing tests only cover the static `buildRestResponse` helper, so the new dispatch path isn't exercised at all. The one IT that touches this class (`AdmissionForClusterManagerIT`) takes the admission-rejection path, which fires before `processResponse` runs.

Could you add a test that confirms `buildResponse` runs off the transport thread? That would catch regressions in the lambda capture / `ThreadContext` / dispatch wiring.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -200,10 +209,22 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
+                        .execute(ActionRunnable.wrap(this, l -> new RestBuilderListener<GetAliasesResponse>(channel) {
```
dzane17: Remove unused param `l ->`

dzane17:
Can you replace `Fixes #21900` from the PR description? Since this is just part 1 of the solution we don't want to auto-close the GitHub issue. `Part of #21900` or `Related to #21900` will work

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -198,12 +213,26 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
+        return channel -> client.admin().indices().getAliases(getAliasesRequest, new RestActionListener<GetAliasesResponse>(channel) {
             @Override
-            public RestResponse buildResponse(GetAliasesResponse response, XContentBuilder builder) throws Exception {
-                return buildRestResponse(namesProvided, aliases, response.getAliases(), builder);
+            protected void processResponse(GetAliasesResponse response) {
+                final long startTimeMs = tp.relativeTimeInMillis();
+                // Serialization of alias metadata (filter decompression + XContent build) can be slow on clusters
+                // with many filtered aliases (e.g. DLS). Fork to MANAGEMENT to avoid blocking the transport thread.
+                tp.executor(ThreadPool.Names.MANAGEMENT)
+                    .execute(ActionRunnable.run(this, () -> new RestBuilderListener<GetAliasesResponse>(channel) {
```
dzane17: `ActionRunnable.run` calls `listener.onResponse(null)` after the runnable finishes (see `ActionRunnable.java:60-63`). So on every successful request:
1. The inner `RestBuilderListener` sends the response.
2. Then `run` calls `this.onResponse(null)` on the outer listener — which re-enters `processResponse(null)` and dispatches another runnable to MANAGEMENT.
3. That builds against `null`, NPEs, fires the inner failure path, and tries to send a second response on the same channel.

Net effect: every successful `GET /_aliases` recursively redispatches and double-sends.

Cleanest fix is an explicit `AbstractRunnable` — no unused params, failure routing is obvious:
```java
return channel -> client.admin().indices().getAliases(getAliasesRequest, new RestActionListener<GetAliasesResponse>(channel) {
    @Override
    protected void processResponse(GetAliasesResponse response) {
        final long startTimeMs = tp.relativeTimeInMillis();
        final RestActionListener<GetAliasesResponse> outer = this;
        tp.executor(ThreadPool.Names.MANAGEMENT).execute(new AbstractRunnable() {
            @Override
            protected void doRun() throws Exception {
                new RestBuilderListener<GetAliasesResponse>(channel) {
                    @Override
                    public RestResponse buildResponse(GetAliasesResponse r, XContentBuilder builder) throws Exception {
                        // timeout check + buildRestResponse(...) here
                    }
                }.onResponse(response);
            }

            @Override
            public void onFailure(Exception e) {
                outer.onFailure(e);
            }
        });
    }
});
```

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -198,12 +213,26 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
         getAliasesRequest.indicesOptions(IndicesOptions.fromRequest(request, getAliasesRequest.indicesOptions()));
         getAliasesRequest.local(request.paramAsBoolean("local", getAliasesRequest.local()));
```
dzane17: Add this line to inject the user-supplied timeout into `getAliasesRequest`:
```java
getAliasesRequest.clusterManagerNodeTimeout(
    request.paramAsTime("cluster_manager_timeout", getAliasesRequest.clusterManagerNodeTimeout())
);
```
Without it, the `getAliasesRequest.clusterManagerNodeTimeout()` call below always returns the default timeout (30s). A client sending `?cluster_manager_timeout=5s` gets ignored.

dzane17:
Also please add/update tests for this change?

gingeekrishna:
> Also please add/update tests for this change?

Hi @dzane17 

Sorry, I was thinking to but I missed to updated this earlier. are we good now?

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -197,13 +221,48 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
         getAliasesRequest.indices(indices);
         getAliasesRequest.indicesOptions(IndicesOptions.fromRequest(request, getAliasesRequest.indicesOptions()));
         getAliasesRequest.local(request.paramAsBoolean("local", getAliasesRequest.local()));
+        // TODO: Remove the if condition and statements inside after removing MASTER_ROLE.
+        TimeValue clusterManagerTimeout = request.paramAsTime("cluster_manager_timeout", getAliasesRequest.clusterManagerNodeTimeout());
+        if (request.hasParam("master_timeout")) {
+            deprecationLogger.deprecate("get_aliases_master_timeout_parameter", MASTER_TIMEOUT_DEPRECATED_MESSAGE);
```
dzane17, line 212: `deprecationLogger` adds the `Warning` HTTP header on the REST `ThreadContext`, but `sendResponse` reads headers from whichever thread calls it (now MANAGEMENT thread). So the `Warning` header will never reach the client.

Easy fix — wrap the runnable in `preserveContext` (same pattern `ThreadPool.schedule` uses):
```java
tp.executor(ThreadPool.Names.MANAGEMENT).execute(
    tp.getThreadContext().preserveContext(new AbstractRunnable() {
        // ... existing body ...
    })
);
```

dzane17:
@gingeekrishna Added one more comment. Besides that and test coverage, it looks good.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
gingeekrishna: Fixed — wrapped the `AbstractRunnable` in `tp.getThreadContext().preserveContext(...)` so the transport-worker thread context (including the `Warning` header written by `deprecationLogger`) is restored on the MANAGEMENT thread before `sendResponse` is called. Pushed in the latest commit.

gingeekrishna:
> @gingeekrishna Added one more comment. Besides that and test coverage, it looks good.

Thank you @dzane17! Just pushed the changes. looks good?

dzane17:
@gingeekrishna The existing tests only cover the static `buildRestResponse` helper. None of the new behavior is exercised.

Could you add a simple test that confirms the response work is scheduled on the MANAGEMENT thread pool? A mocked `ThreadPool` returning a recording executor + asserting `execute()` was called once on `MANAGEMENT` should do it.

gingeekrishna:
> @gingeekrishna The existing tests only cover the static `buildRestResponse` helper. None of the new behavior is exercised.
> 
> Could you add a simple test that confirms the response work is scheduled on the MANAGEMENT thread pool? A mocked `ThreadPool` returning a recording executor + asserting `execute()` was called once on `MANAGEMENT` should do it.

Sure, will add that soon

gingeekrishna:
Hi @dzane17! 

I have Added `testProcessResponseForksToManagementThreadPool` in the latest commit. 

It mocks `ThreadPool` with a recording `ExecutorService` for `MANAGEMENT`, captures the `ActionListener` passed to `getAliases`, fires `onResponse`, and asserts `execute()` was called exactly once on the `MANAGEMENT` executor.

dzane17:
@gingeekrishna Can you check `CodeQL / Analyze (java)`? Some imports cannot be found.

gingeekrishna:
Thanks for flagging! Fixed in the latest push.

The CodeQL failures were caused by two issues in `RestGetAliasesActionTests.java`:

1. **Wrong package for `AdminClient` / `IndicesAdminClient`** — the test used `org.opensearch.client.AdminClient` and `org.opensearch.client.IndicesAdminClient`, but these classes have moved to `org.opensearch.transport.client`. Updated both references to match the current package.

2. **`CheckedConsumer.accept()` on a protected interface** — the test called `action.prepareRequest(request, nodeClient).accept(channel)`, but `RestChannelConsumer` is a `protected` interface inside `BaseRestHandler`, making the `accept()` call inaccessible to CodeQL's type analysis. Replaced with `action.handleRequest(request, channel, nodeClient)` which is the public API and handles the same flow internally.

dzane17 (approved):

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -70,6 +77,23 @@
  */
 public class RestGetAliasesAction extends BaseRestHandler {
 
+    private static final DeprecationLogger deprecationLogger = DeprecationLogger.getLogger(RestGetAliasesAction.class);
+    static final String MASTER_TIMEOUT_DEPRECATED_MESSAGE =
+        "Parameter [master_timeout] is deprecated and will be removed in 3.0. To support inclusive language, please use [cluster_manager_timeout] instead.";
```
cwperks:
```suggestion
        "Parameter [master_timeout] is deprecated and will be removed in 4.0. To support inclusive language, please use [cluster_manager_timeout] instead.";
```

cwperks:
@gingeekrishna @dzane17 if I'm not mistaken, I believe we can actually declare which threadpool to use directly on the transport action like https://github.com/opensearch-project/security/blob/b94282574b4261774f46b9bcd142ce4742d13aa3/src/main/java/org/opensearch/security/action/configupdate/TransportConfigUpdateAction.java#L84.

server/src/main/java/org/opensearch/rest/action/admin/indices/RestGetAliasesAction.java
```diff
@@ -197,13 +221,48 @@ public RestChannelConsumer prepareRequest(final RestRequest request, final NodeC
...
+                // Serialization of alias metadata (filter decompression + XContent build) can be slow on clusters
+                // with many filtered aliases (e.g. DLS). Fork to MANAGEMENT to avoid blocking the transport thread.
+                final RestActionListener<GetAliasesResponse> outer = this;
+                tp.executor(ThreadPool.Names.MANAGEMENT).execute(tp.getThreadContext().preserveContext(new AbstractRunnable() {
+                    @Override
+                    protected void doRun() throws Exception {
+                        new RestBuilderListener<GetAliasesResponse>(channel) {
+                            @Override
+                            public RestResponse buildResponse(GetAliasesResponse r, XContentBuilder builder) throws Exception {
+                                if (tp.relativeTimeInMillis() - startTimeMs > timeout.millis()) {
```
cwperks: Will this only check for timeout condition once?

gingeekrishna:
Thanks @cwperks!

**Deprecation message version ([comment](https://github.com/opensearch-project/OpenSearch/pull/21954#discussion_r3495649311)):** Applied — `MASTER_TIMEOUT_DEPRECATED_MESSAGE` now says 4.0. (Worth noting `RestGetMappingAction` and several other unrelated places in the codebase still say "removed in 3.0" — that's a pre-existing, repo-wide staleness issue beyond this PR's scope. Happy to file a follow-up if useful.)

**Timeout check frequency ([comment](https://github.com/opensearch-project/OpenSearch/pull/21954#discussion_r3495677414)):** Yes, it's checked exactly once — right when the dispatched work starts running on the MANAGEMENT thread, before serialization begins. This mirrors `RestGetMappingAction`'s identical single-check pattern exactly. It guards against time spent queued in MANAGEMENT while the client has already given up and retried; it does not re-check partway through `buildRestResponse`'s loop if serialization itself runs long. Keeping it consistent with the sibling action's established convention rather than introducing a different (more granular) timeout strategy just for aliases. Let me know if you'd rather see periodic checking during serialization — happy to add it as a deliberate deviation if so.

**Bonus finding:** While verifying these changes I found `testProcessResponseForksToManagementThreadPool` was actually broken — it mocked `NodeClient` directly and stubbed `admin()`, which is a `final` method Mockito can't stub (https://github.com/mockito/mockito/issues/146), so the test was never actually exercising the dispatch path it claimed to verify. Rewrote it using a real `NoOpNodeClient` (overriding `doExecute`) and a real `TestThreadPool` with a thread-name-capturing wrapper around the `MANAGEMENT` executor, so it now genuinely proves the response work runs off the calling thread. Ran it 3x in a loop with no flakiness before pushing.

cwperks:
@gingeekrishna did you see the comment from https://github.com/opensearch-project/OpenSearch/pull/21954#issuecomment-4839132580? I think we can declare the thread group in the transport action.

gingeekrishna:
Great call @cwperks. Changed `TransportGetAliasesAction.executor()` from `SAME` to `MANAGEMENT` — the transport framework then handles the dispatch off the Netty thread automatically, so all the manual `AbstractRunnable`/`preserveContext`/timeout-guard code in `RestGetAliasesAction` is gone. The REST handler is back to its original straightforward `RestBuilderListener` shape (keeping only the `cluster_manager_timeout`/`master_timeout` wiring and the deprecation-message version fix from earlier passes).

Added `TransportGetAliasesActionTests#testExecutorIsManagement` to pin the executor choice explicitly.

cwperks:
@gingeekrishna please also run `./gradlew spotlessApply`. Latest changes look good to me.

gingeekrishna:
@cwperks — ran `./gradlew spotlessApply`, which removed two unused imports (`GetAliasesResponse` and `Collections`) that were left over from the threading test deleted in the previous pass. Pushed.

cwperks (approved):

PR #21958, opened 2026-06-02
[analytics-engine] Push non-aggregate sort below exchange

sandeshkr419:
## What
  For non-aggregate `ORDER BY .. LIMIT [OFFSET]` queries, copy the bottom-most collated `Sort` to just below the `ExchangeReducer` so each shard ships only its local top-N already-sorted rows. The coordinator `Sort` is left intact to merge the streams and apply the final offset/limit.

  ## Why
  Cuts data-node → coordinator transport from all matching rows per shard to N per shard. Exact (no oversampling): a row in the global window is within its own shard's top-N.

  ## How
  - New `OpenSearchSortPushdownRewriter`: post-CBO rewriter registered after the TopK rewrite in `PlannerImpl`.
  - Handles both the SQL single-node shape (`Sort(collation, fetch)`) and the PPL two-node shape (outer pure-`fetch` over inner collated `Sort`), folding the enclosing fetch into the pushed shard `Sort`.
  - **Collation**: copied verbatim (no remap — ER preserves schema and Sort sits directly above it).
  - **Limit**: shard `fetch = effective fetch`.
  - **Offset**: never pushed; shard `fetch = offset + fetch`, offset kept only on the coordinator.
  - Skips the aggregate path (ER feeding a `PARTIAL` aggregate) — owned by the TopK rewriter.

  ## Testing
  - Unit (`SortPushdownPlanShapeTests`): SQL/PPL shapes, offset widening, and skip cases (no fetch, single-shard, aggregate path); updated 3 existing plan-shape expectations.
  - IT (`SortPushdownIT`, multi-shard): desc/asc global top-N, multi-key, sort+filter, pure-limit, large-head — assert count, ordering, and boundary == global/filtered max/min.
  - `ShardBucketOversamplingIT` 7/7 still pass (no interference). Unit + IT + spotless green.
### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
@@ -81,7 +81,7 @@ public class PlannerImpl {
      * Package-private so planner rule tests can inspect the marked+optimized tree.
      */
     public static RelNode runAllOptimizations(RelNode rawRelNode, PlannerContext context) {
-        LOGGER.info("Input RelNode:\n{}", RelOptUtil.toString(rawRelNode));
+        LOGGER.debug("Input RelNode:\n{}", RelOptUtil.toString(rawRelNode));
 
         RuleProfilingListener listener = context.isProfilingEnabled() ? new RuleProfilingListener() : null;
 
@@ -92,7 +92,7 @@ public class PlannerImpl {
         modifiedRelNode = pushdownRules(modifiedRelNode, listener);
         modifiedRelNode = decomposeAggregates(modifiedRelNode, listener);
         modifiedRelNode = mark(modifiedRelNode, context, listener);
-        LOGGER.info("After marking:\n{}", RelOptUtil.toString(modifiedRelNode));
+        LOGGER.debug("After marking:\n{}", RelOptUtil.toString(modifiedRelNode));
         // TODO(combine-delegated-predicates): a post-marking HEP rule should fuse same-backend
         // AND-sibling AnnotatedPredicates into one combined predicate per group, collapsing N
         // FFM round-trips per RG into one. Blocked on two open design points:
@@ -103,21 +103,21 @@ public class PlannerImpl {
         // Revisit once those are designed. The rule would also strip performance peers from
         // AnnotatedPredicates under OR/NOT (Lucene call buys nothing in those positions).
         modifiedRelNode = cbo(modifiedRelNode, rawRelNode, context, listener);
-        LOGGER.info("After CBO:\n{}", RelOptUtil.toString(modifiedRelNode));
+        LOGGER.debug("After CBO:\n{}", RelOptUtil.toString(modifiedRelNode));
         Optional<RelNode> lateMat = OpenSearchLateMaterializationRewriter.rewrite(modifiedRelNode);
         if (lateMat.isPresent()) {
             modifiedRelNode = lateMat.get();
-            LOGGER.info("After late-materialization:\n{}", RelOptUtil.toString(modifiedRelNode));
+            LOGGER.debug("After late-materialization:\n{}", RelOptUtil.toString(modifiedRelNode));
         }
         Optional<RelNode> topK = OpenSearchTopKRewriter.rewrite(modifiedRelNode, context);
         if (topK.isPresent()) {
             modifiedRelNode = topK.get();
-            LOGGER.info("After TopK rewrite:\n{}", RelOptUtil.toString(modifiedRelNode));
+            LOGGER.debug("After TopK rewrite:\n{}", RelOptUtil.toString(modifiedRelNode));
         }
         Optional<RelNode> sortPushdown = OpenSearchSortPushdownRewriter.rewrite(modifiedRelNode);
         if (sortPushdown.isPresent()) {
             modifiedRelNode = sortPushdown.get();
-            LOGGER.info("After sort pushdown:\n{}", RelOptUtil.toString(modifiedRelNode));
+            LOGGER.debug("After sort pushdown:\n{}", RelOptUtil.toString(modifiedRelNode));
         }
 
         if (listener != null) {
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
@@ -36,6 +36,7 @@ import org.opensearch.analytics.planner.rules.OpenSearchJoinRule;
 import org.opensearch.analytics.planner.rules.OpenSearchJoinSplitRule;
 import org.opensearch.analytics.planner.rules.OpenSearchLateMaterializationRewriter;
 import org.opensearch.analytics.planner.rules.OpenSearchProjectRule;
+import org.opensearch.analytics.planner.rules.OpenSearchSortPushdownRewriter;
 import org.opensearch.analytics.planner.rules.OpenSearchSortRule;
 import org.opensearch.analytics.planner.rules.OpenSearchSortSplitRule;
 import org.opensearch.analytics.planner.rules.OpenSearchTableScanRule;
@@ -112,6 +113,11 @@ public class PlannerImpl {
         if (topK.isPresent()) {
             modifiedRelNode = topK.get();
             LOGGER.info("After TopK rewrite:\n{}", RelOptUtil.toString(modifiedRelNode));
+        }
+        Optional<RelNode> sortPushdown = OpenSearchSortPushdownRewriter.rewrite(modifiedRelNode);
+        if (sortPushdown.isPresent()) {
+            modifiedRelNode = sortPushdown.get();
+            LOGGER.info("After sort pushdown:\n{}", RelOptUtil.toString(modifiedRelNode));
         }
 
         if (listener != null) {
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
@@ -15,6 +15,7 @@ import org.apache.calcite.sql.type.SqlTypeName;
 import org.opensearch.analytics.planner.rel.AggregateMode;
 import org.opensearch.analytics.planner.rel.OpenSearchAggregate;
 import org.opensearch.analytics.planner.rel.OpenSearchExchangeReducer;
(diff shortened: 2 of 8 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+        for (RelNode child : node.getInputs()) {
+            Match m = find(child, null);
+            if (m != null) return m;
+        }
+        return null;
+    }
+
+    /** Foldable when there's no offset, or both offset and fetch are literals we can sum. */
+    private static boolean foldable(OpenSearchSort bound) {
```
expani: Name is confusing what's the intent here ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -113,6 +114,11 @@ public static RelNode runAllOptimizations(RelNode rawRelNode, PlannerContext con
             modifiedRelNode = topK.get();
             LOGGER.info("After TopK rewrite:\n{}", RelOptUtil.toString(modifiedRelNode));
         }
+        Optional<RelNode> sortPushdown = OpenSearchSortPushdownRewriter.rewrite(modifiedRelNode);
+        if (sortPushdown.isPresent()) {
+            modifiedRelNode = sortPushdown.get();
+            LOGGER.info("After sort pushdown:\n{}", RelOptUtil.toString(modifiedRelNode));
```
expani: nit: can be debug let's swap the above one's as well.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/PlanShapeTests.java
```diff
@@ -294,13 +294,15 @@ public void testSortThenProjectThenLimit_multiShard() {
         RelNode result = runPlanner(limit, buildContext("parquet", 3, fields));
         // SORT_PROJECT_TRANSPOSE pushes the outer pure-LIMIT Sort below the identity Project,
         // producing the QTF-friendly two-Sort shape Project(identity) ← Sort(fetch) ← Sort(coll) ← ER.
+        // The sort-pushdown rewriter then copies the collated Sort (with the outer fetch) below the ER.
         assertPlanShape(
             """
                 OpenSearchProject(name=[$0], score=[$1], viableBackends=[[mock-parquet]])
                   OpenSearchSort(fetch=[3], viableBackends=[[mock-parquet]])
```
expani, line 301: Should be start using Calcite's `SortMergeRule` to avoid this from happening ? 

This query can get Top3 from all the shards then perform the final Top3 at the coordinator.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/SortPushdownIT.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * End-to-end correctness for non-aggregate sort pushdown on a multi-shard index.
+ * With the rewriter active, each shard ships only its local top-N already-sorted
+ * rows; the coordinator merges them. These tests assert the merged result is the
+ * correct, globally-ordered top-N (catches dropped rows / wrong ordering).
+ */
+public class SortPushdownIT extends AnalyticsRestTestCase {
```
expani, line 22: How is this asserting that the plan generated has sort pushed down ? 

IIUC All the changes are related to performance as each shard can sort in parallel after this.

One option is to perform assertion on the plan using the Explain API if an E2E IT is required.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
```diff
@@ -0,0 +1,135 @@
...
+                    return new Match(sort, bound, er);
+                }
+            }
+            // A pure-fetch Sort (no collation, has fetch) carries its offset+fetch to its direct child.
+            OpenSearchSort carry = (collated == false && sort.fetch != null) ? sort : null;
+            return find(sort.getInput(), carry);
+        }
+        for (RelNode child : node.getInputs()) {
+            Match m = find(child, null);
+            if (m != null) return m;
```
expani: Looks like if there are sort on the left and right child of a Join/Union we are short circuiting when the first one returns a match. 

Which means we won't split at each point which can hurt the performance of those queries. 

Also, we can look at introducing the following rules in Calcite RBO for pushing the sort down
```
SortJoinTransposeRule
SortUnionTransposeRule
```

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/SortPushdownPlanShapeTests.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.util.List;
+
+/**
+ * Plan-shape tests for {@link org.opensearch.analytics.planner.rules.OpenSearchSortPushdownRewriter}.
+ *
+ * <p>Fires for non-aggregate {@code ORDER BY .. LIMIT}: a copy of the collated Sort
+ * (with its effective fetch) is inserted below the ER so shards ship local top-N.
+ */
+public class SortPushdownPlanShapeTests extends PlanShapeTestBase {
```
expani, line 29: Let's add a few tests with Joins and Unions after we add Sort pushdown rule and handling it for Joins and Unions

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
```diff
@@ -0,0 +1,166 @@
...
+        RelNode[] newChildren = new RelNode[children.size()];
+        boolean changed = false;
+        for (int i = 0; i < children.size(); i++) {
+            newChildren[i] = replaceInTree(children.get(i), oldNode, newNode);
+            if (newChildren[i] != children.get(i)) changed = true;
+        }
+        return changed ? root.copy(root.getTraitSet(), List.of(newChildren)) : root;
+    }
+
+    private record Match(OpenSearchSort collated, OpenSearchSort bound, RelNode below) {
```
expani: This is not just the matched node so maybe SortPushdownRewriteContext or something shorter.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
```diff
@@ -0,0 +1,166 @@
...
+import org.opensearch.analytics.planner.rel.OpenSearchJoin;
+import org.opensearch.analytics.planner.rel.OpenSearchSort;
+import org.opensearch.analytics.planner.rel.OpenSearchUnion;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Post-CBO rewriter for non-aggregate TopK. Copies the bottom-most collated
```
expani, line 27: Can we start with a TODO as : 

```
We need to use the correct way of doing Trait Propagation in Calcite with Volcano and this entire rewrite is a temporary workaround until the same.
```

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
```diff
@@ -0,0 +1,166 @@
...
+                OpenSearchSort bound = sort.fetch != null ? sort : (sort.offset == null ? fetchSortAbove : null);
+                RelNode below = sort.getInput();
+                if (bound != null && canComputeShardFetch(bound) && isPushTarget(below)) {
+                    return new Match(sort, bound, below);
+                }
+            }
+            OpenSearchSort carry = (collated == false && sort.fetch != null) ? sort : null;
+            return find(sort.getInput(), carry);
+        }
+        return node.getInputs().size() == 1 ? find(node.getInputs().get(0), null) : null;
```
expani, line 83: `node.getInputs().get(0)` 

This is not recursing into other childs. Intentional ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * Eligible to push a Sort below: not the aggregate path, not already pushed, and not gathering a
+     * Join — joins are explicitly out of scope (a LIMIT cannot be safely pushed below a join).
+     */
+    private static boolean eligibleER(OpenSearchExchangeReducer er) {
+        RelNode input = er.getInput();
+        return isAggregatePath(er) == false && (input instanceof OpenSearchSort) == false && (input instanceof OpenSearchJoin) == false;
+    }
+
+    /** True when the shard fetch is computable: no offset, or offset and fetch are both literals to sum. */
```
expani: > offset and fetch are both literals to sum.

Ideally we can do with complex expressions as well but that is out of scope. 

Let's add a TODO for now.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
```diff
@@ -0,0 +1,166 @@
...
+            collated.getViableBackends()
+        );
+        return er.copy(er.getTraitSet(), List.of(shardSort));
+    }
+
+    /**
+     * Eligible to push a Sort below: not the aggregate path, not already pushed, and not gathering a
+     * Join — joins are explicitly out of scope (a LIMIT cannot be safely pushed below a join).
+     */
+    private static boolean eligibleER(OpenSearchExchangeReducer er) {
```
expani: nit:eligibleChildBelowER

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
sandeshkr419: Yes, added a comment: we follow only the single-input coordinator spine; multi-input Join is intentionally not traversed (pushing a sort into one join branch is unsafe), and UNION ALL is matched directly as a push target instead.

expani (approved):

mch2 (approved):

PR #21960, opened 2026-06-03
[analytics-engine] Fix multi-input single-shard backend resolution in PlanForker

mch2:
### Description
A multi-input operator (Union, Join) and its children compile into one exchange-free stage, so they must share a backend. PlanForker's multi-input branch took each child's *first* plan alternative (getFirst()), which is whichever backend leads the child's viable list — e.g. a dual-viable scan lists lucene first. A backend-restricted parent (a DataFusion-only Union/Join) then had viableBackends and produced zero plan alternatives, surfacing later as a NoSuchElementException at execution.

Fix PlanForker to mirror the single-input case: for each backend the parent can run, take each child's alternative on that backend and delegate to resolveOperator (which already does the viableBackends intersection). The parent is viable only on backends every child offers; one alternative is emitted per shared backend. This reconciles the arms on datafusion (the lucene-only filter predicate is delegated to Lucene at the scan), so union/join/semi-join shapes resolve and run.

Un-mutes that this enables, verified at 2 shards via /_plugins/_ppl:
- TwoShardJoinIT: all 6 join shapes (inner/left/semi/anti + join-then-group)
- TwoShardCommandIT: append, multisearch, regex, appendpipe
- StreamstatsCommandIT.testWhereInWithStreamstatsSubquery: where-in subquery now runs (asserts cardinality; result is nondeterministic by construction)

Also makes ExtensiveCoverage q78 deterministic (sort d.value before head 3; golden unchanged) so the join+head no longer returns a nondeterministic row subset.

Test: PlanForkerTests.testMultiInputForksChildToParentBackend.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanForker.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanForker.java
@@ -73,39 +73,39 @@ public class PlanForker {
             return results;
         }
 
-        // Multi-input: take the first alternative from each child. With a single backend
-        // (pure DataFusion), each child has exactly one alternative anyway. For correctness
-        // we require all children to agree on the chosen backend — a multi-input operator
-        // cannot straddle backends within a single stage.
-        // TODO: when multi-backend pipelines are added, fan out the Cartesian product of
-        // child alternatives and prune by backend agreement.
-        List<RelNode> resolvedChildren = new ArrayList<>(childAlternativeSets.size());
-        String agreedBackend = null;
-        for (List<Resolved> childAlts : childAlternativeSets) {
-            if (childAlts.isEmpty()) {
-                throw new IllegalStateException(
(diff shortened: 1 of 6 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/PlanForker.java
```diff
@@ -73,39 +73,46 @@ private static List<Resolved> resolve(RelNode node, CapabilityRegistry registry)
             return results;
         }
 
-        // Multi-input: take the first alternative from each child. With a single backend
-        // (pure DataFusion), each child has exactly one alternative anyway. For correctness
-        // we require all children to agree on the chosen backend — a multi-input operator
-        // cannot straddle backends within a single stage.
+        // Multi-input: a multi-input operator and all its children compile into one
+        // exchange-free stage, so they must share a backend. Mirror the single-input case —
```
expani: This might not always be true given we can Join different formats across each side.


nit: Can we make this 2-3 lines ?

expani (approved):

sandeshkr419 (approved):

PR #21962, opened 2026-06-03
Wire native store provider to warm-tier repos and fix DFA ITs + Foyer Tunning

vishwasgarg18:
### Description

**1. Fail-fast guard in `ParquetDataFormatStoreHandler` when native remote store is unavailable**

`ParquetDataFormatStoreHandler` creates a tiered object store (Parquet → Foyer cache → remote object store) in its constructor. On a warm shard, the remote object store pointer (`remotePtr`) must be non-zero — without it, every read would silently fall through to a null-pointer dereference inside native code.

Before this change, if `repo` was `null` or `repo.isLive()` returned `false`, `remotePtr` was `0L` and the code proceeded to call `TieredStorageBridge.createTieredObjectStore(0L, 0L, cachePtr)` — undefined behaviour in native Rust code that could produce silent data corruption, a JVM crash, or an opaque segfault rather than a clear actionable error.

The guard added:
```java
if (remotePtr == 0L) {
    throw new IllegalStateException(
        "[" + shardId + "] remote object store is not available for this warm shard. Cannot serve read requests."
    );
}
```
makes the failure mode explicit: the shard open fails with a descriptive Java exception, the cluster marks the shard as ALLOCATION_FAILED, retries allocation on another node, and the operator gets a clear error message. This is the correct contract — a warm shard without a remote store is misconfigured and must not silently serve incorrect data.

**2. Foyer block cache — scheduler task resilience (`foyer_cache.rs` + `tests.rs`)**

The sweep task (`reconcile_key_index`) and persist task (`key_index_store::save`) previously had no error boundary. A panic inside either would silently kill the background task for the node's lifetime, leaving the key_index permanently stale or never persisted again.

Both tasks now wrap their critical sections in `std::panic::catch_unwind(AssertUnwindSafe(...))` — panics are caught, logged as ERROR, and the task continues on the next tick. An outer restart loop with exponential back-off (1 s → 2 s → … capped at 60 s) ensures the scheduler never goes permanently dark if the inner select loop exits unexpectedly. `tests.rs` adds unit tests covering panic recovery, I/O error retry, and back-off restart behaviour.

**3. Native store provider wired through `ReloadableFsRepository` and `RepositoriesModule`**

`FsRepository` already supported a `NativeRemoteObjectStoreProvider` constructor (added earlier to support S3/GCS warm nodes). `ReloadableFsRepository` — a subclass used by test clusters and some production configurations — was missing this constructor, so its `nativeStore` field always defaulted to `NativeStoreRepository.EMPTY` (`isLive() == false`), which caused `remotePtr` to always be `0L`.

Changes:
- `ReloadableFsRepository`: adds the `NativeRemoteObjectStoreProvider` constructor that delegates to `super`, matching the existing `FsRepository` pattern.
- `RepositoriesModule`: passes `fsNativeProvider` to the `ReloadableFsRepository.TYPE` factory, closing the gap for any production node that registers a reloadable-fs repo.
- `MockFsMetadataSupportedRepository` (internalClusterTest): adds the same constructor so the test mock can be constructed with a native provider when the integration test requires it.

**4 Integration tests for warm-tier DFA remote store recovery (`DataFormatAwareReadonlyEngineBaseIT`)**

The ITs are needed because the above changes form a contract that requires end-to-end validation across the JVM↔native boundary: the native store must be live, the shard must open, data must be readable, and recovery from remote store must produce an identical data set. Unit tests cannot cover this — only a real cluster startup with real shard allocation, real Parquet file writes to remote store, and real shard recovery can prove the contract holds.

`RemoteStoreBaseIntegTestCase` randomly picks between two mock repository types (`fs_metadata_supported_repository` and `fs_multipart_repository`) via `metadataSupportedType = randomBoolean()`. Neither was wired with a `NativeRemoteObjectStoreProvider`, so the guard added in (1) fired on every test run. The fix replaces both mock plugins with native-provider-aware inner classes (`NativeAwareMockFsMetadataSupportedRepositoryPlugin` for `fs_metadata_supported_repository` and `NativeAwareMockFsRepositoryPlugin` for `fs_multipart_repository`) that wire `FsNativeObjectStorePlugin`, covering both random paths. `nodeSettings` pre-creates the remote store directories on disk so Rust's `canonicalize()` succeeds at node startup.

With the infrastructure correct, `DataFormatAwareReadonlyEngineRemoteStoreRecoveryIT` asserts two recovery contracts:
- **New replica recovery**: after scaling replicas from 0 to 1, the replica recovers from remote store. The test asserts that the recovered replica's Parquet catalog files (data files, excluding Lucene segment metadata) are an exact set-equality match with the primary's catalog — proving the full Parquet data set was replicated correctly via the remote store path.
- **Primary restart recovery**: after restarting the primary node, the test asserts that the re-opened engine is a `DataFormatAwareReadOnlyEngine` (not a fallback Lucene engine) and that the Parquet catalog is identical to the pre-restart snapshot — proving that warm shard state survives a node restart and is correctly recovered from remote store.

**5. `TransportHotToWarmTierAction` refactoring**

The DFA preflight flow for hot-to-warm tiering — capacity validation → add write block (persisted to index metadata) → pre-tiering sync with up to 3 retries → tier — was reorganised into cleaner extracted methods (`addWriteBlockAndPrepare`, `executePrepareTiering`, `removeWriteBlock`), with the non-DFA path bypassing the preflight entirely. `TransportHotToWarmTierActionTests` is updated to match the new mock structure and method signatures.

**6. `BlockCacheKeyIndexScaleIT` flakiness fix**

`testKeyIndexJsonWrittenOnShutdown` and `testEvictPrefixAfterRecovery` both call `restartNode()` followed by bare `ensureGreen()`. The default 30-second timeout is insufficient when 25 REMOTE_SNAPSHOT shards (5 indices × 5 shards) need to re-open after the warm node restarts under load. Both calls are raised to `ensureGreen(TimeValue.timeValueSeconds(120))`.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ToStringFunctionAdapter.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ToStringFunctionAdapter.java
@@ -67,7 +67,7 @@ class ToStringFunctionAdapter implements ScalarFunctionAdapter {
     private enum Format {
         HEX("hex", SqlTypeName.BIGINT),
         BINARY("binary", SqlTypeName.BIGINT),
-        COMMAS("commas", /* preserveFractional */ null),
+        COMMAS("commas", /* preserveFractional */null),
         DURATION("duration", SqlTypeName.BIGINT),
         DURATION_MILLIS("duration_millis", SqlTypeName.BIGINT);
 
--- a/sandbox/plugins/block-cache-foyer/src/internalClusterTest/java/org/opensearch/blockcache/foyer/BlockCacheKeyIndexScaleIT.java
+++ b/sandbox/plugins/block-cache-foyer/src/internalClusterTest/java/org/opensearch/blockcache/foyer/BlockCacheKeyIndexScaleIT.java
@@ -21,6 +21,7 @@ import org.opensearch.action.admin.cluster.snapshots.restore.RestoreSnapshotRequ
 import org.opensearch.action.support.PlainActionFuture;
 import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.unit.TimeValue;
 import org.opensearch.common.util.FeatureFlags;
 import org.opensearch.env.Environment;
 import org.opensearch.node.Node;
@@ -120,7 +121,9 @@ public class BlockCacheKeyIndexScaleIT extends AbstractSnapshotIntegTestCase {
 
         logger.info("[scale-01] restarting warm node '{}'", getWarmNodeName());
         internalCluster().restartNode(getWarmNodeName());
-        ensureGreen();
+        // Use an extended timeout: after restart, 25 REMOTE_SNAPSHOT shards (5 indices × 5 shards)
+        // must re-open before the cluster goes GREEN. The default 60s is insufficient under load.
+        ensureGreen(TimeValue.timeValueSeconds(120));
 
         Path cacheDir = findFoyerCacheDir();
         assertNotNull("foyer-block-cache directory must exist on warm node after restart", cacheDir);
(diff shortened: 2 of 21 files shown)


nishchay21 (approved):

Bukhtawar (approved):
Lets add benchmarks for the change, approving to unblock

PR #21951, opened 2026-06-02
Sandbox/qa - Add multi shard qa test suite

mch2:
### Description
Adds a test suite to sandbox/qa for 2 shard setups.  Also unmutes single shard tests that are now passing.
see [sandbox/qa/analytics-engine-rest/src/test/resources/datasets/merge_coverage/README.md](https://github.com/opensearch-project/OpenSearch/compare/main...mch2:OpenSearch:twoshard-merge-suite?expand=1#diff-96ee036550fb41359ecb054ca76b8aff3ef92e7ce8314df95b239fcb07fa6ec4).
 for current list of skipped/failing tests.

How

  merge_coverage is provisioned at 1 shard (oracle) and 2 shards; each %INDEX% query runs against both. Exact tier asserts the two results are equal (unordered, numeric-tolerant, multiset-normalized for
  values/list); curated shapes also pin a golden. Approximate tier (distinct_count/percentile) asserts the 2-shard result within tolerance of a required golden, since sketches drift across a merge. head/limit
  carry a unique id tie-breaker for determinism. Indices are composite parquet + lucene, so the planner's per-arm backend selection is exercised.

  Structure: one abstract TwoShardReduceTestCase (all machinery) + thin per-category ITs that only declare their dataset tier and known-issue set — TwoShardAggregationIT, TwoShardScalarIT, TwoShardShapeIT,
  TwoShardJoinIT, TwoShardCommandIT, plus IpFieldMultiShardIT.

Currently muted tests:

Here's the full list of currently-muted tests (@AwaitsFix) in analytics-engine-rest

```
ere's the complete failure inventory across the analytics-engine-rest module.

  1-shard suites

  Per-query failures (suite runs, these qN skip)

  AggregationsPplIT       q7, q8, q9, q10                              (4 of 10)
  AppLogsPplIT            q5, q9                                       (2 of 10)
  ComplexJoinsPplIT       q1,q2,q3,q4,q7,q8,q9,q10                     (8 of 10)
  FulltextWindowPplIT     q1,q6,q8,q12,q13,q14,q15,q17,q19            (9 of 20)
  FunctionsPplIT          q13                                          (1 of 18)
  KubernetesLogsPplIT     q9                                           (1 of 10)
  MultiIndexQueriesPplIT  q2, q7, q10                                  (3 of 14)
  MultiSourceJoinsPplIT   q2, q4                                       (2 of 5)
  RexCommandPplIT         q1,q5,q7,q8,q13,q18                         (6 of 20)
[...]

Issue #5525, opened 2022-12-12
Provide ZIP for repository-s3 plugin


Snoopy2077:
Unfortunately we cannot install the repository-s3 plugin currently (unless I missed something)

Neither on our build pipelines nor on the container itself do we have internet access, which means
We can't install through Maven because it doesn't support mirrors (which we don't have)
We can't install by name because again, no connectivity
We can't install by zip because we didn't find the zip hosted anywhere

Any suggestions on how to resolve this?

Snoopy2077:
found a way buried in some other issue ;)

dblock:
Link for the next person @Snoopy2077 ?

PR #21975, opened 2026-06-03
[analytics-engine] Enable now()/rand(seed)/sha2/span-month on the analytics-engine route + review fixes

mengweieric:
### Problem

On the analytics-engine (DataFusion) route, several PPL functions failed during Substrait conversion with `IllegalArgumentException: Unable to convert call ...`, or returned wrongly-rendered results:

- `now(fsp)` / `current_timestamp(fsp)` → `Unable to convert call now(i32)`
- `rand(seed)` → `Unable to convert call RAND(i32)`
- `sha2(x, <unsupported bitlen>)` → cryptic Substrait error
- `span(ts, N, month|quarter|year)` → unsupported variable-length buckets
- `date_format(..., '%c')` → month not zero-padded, diverging from the SQL/PPL reference
- a `rand()` predicate could be pushed below its defining projection and re-evaluated, drawing a fresh random value

### Solution

All engine-side (`analytics-framework` / `analytics-backend-datafusion` / `analytics-engine`). Valid SQL/PPL semantics match the SQL-plugin reference; unsupported AE-route shapes now fail earlier with a clearer AE-side error.

| Problem | Fix | Test |
|---|---|---|
| `now(fsp)` fails Substrait conversion | `NowFspAdapter` drops the optional fsp operand → niladic `now()` (the reference intentionally ignores fsp); only the 0/1-arg shapes are normalized | `NowFspAdapterTests` |
| `rand(seed)` fails Substrait conversion | `RandSeedAdapter` | `RandSeedAdapterTests` |
| `sha2(x, N)` | rewrite to `encode(digest(x, 'shaN'), 'hex')`; unsupported bit lengths fail with a clear error listing `[224, 256, 384, 512]` | `Sha2FunctionAdapterTests`, `CalcitePPLCryptographicFunctionIT` |
| `span(…, month/quarter/year)` unsupported | calendar month-index bucketing (bucket start at first-of-month); floored mod/div for pre-1970; `multiplyExact` overflow guard | `SpanAdapter`, `CalcitePPLAggregationIT` |
| `date_format` `%c` month not zero-padded | zero-padded month to match the merged ppl/sql `DateTimeFunctionIT` | `DatetimeCoverageIT` |
| `rand()` predicate pushed below its projection | `PlannerImpl`: don't push filters below non-deterministic projects (semantic-correctness guard; pushing would re-evaluate `RAND()` and draw a fresh value) | n/a |
| `strftime` token coverage | `os_strftime.rs` | (rust unit tests) |

`DataFusionFragmentConvertor` / `DataFusionAnalyticsBackendPlugin` add a shared Substrait preprocess entry point used by the adapters. `MinMaxBooleanAggregationIT` adds QA coverage for `min`/`max` over boolean fields.

### Review fixes

- Removed the late defensive `APPROX_COUNT_DISTINCT` fallback in `PplAggregateCallRewriter` so an unexpected marker fails loud rather than being silently rescued.
- Restored the Substrait-layer assessment `TODO` in `DataFusionFragmentConvertor`.
- Clarified that `fsp` is MySQL fractional-seconds precision in `NowFspAdapter`; added a `TODO` to move SHA2 bit-length validation to the frontend once the handoff carries function metadata.

### Out of scope

- `distinct_count_approx` on the analytics-engine route is handled upstream by #22120.
- `utc_date` / `utc_time` / `utc_timestamp` rendering and the Arrow DATE/TIME value conversion landed separately.
- `timestamp - timestamp` is handled upstream by the `MinusAdapter` (#21978).


sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateFunction.java
```diff
@@ -129,10 +129,21 @@ public static AggregateFunction fromSqlKind(SqlKind kind) {
         return null;
     }
 
-    /** Case-insensitive name lookup; throws if not recognized. */
+    /**
+     * Case-insensitive name lookup; throws if not recognized. Recognizes a small set of
+     * aliases for PPL aggregate names that don't match the enum constant spelling.
+     */
     public static AggregateFunction fromNameOrError(String name) {
+        String upper = name.toUpperCase(java.util.Locale.ROOT);
```
sandeshkr419: you might not need this now - can you give me an IT if it is still failing - I can look into it.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateFunction.java
```diff
@@ -137,10 +137,21 @@ public static AggregateFunction fromSqlKind(SqlKind kind) {
...
+        // PPL emits DISTINCT_COUNT_APPROX (dc/distinct_count_approx); it is the same approximate
+        // distinct-count as the enum's APPROX_COUNT_DISTINCT (DataFusion approx_distinct), just a
+        // different spelling. Map it to the existing constant so the proven sketch state +
+        // substrait mapping are reused.
+        if (upper.equals("DISTINCT_COUNT_APPROX")) {
```
sandeshkr419: We are sending `APPROX_COUNT_DISTINCT` from sql plugin itself now - might not need this. 

check https://github.com/opensearch-project/sql/pull/5525 and validate if this part is needed anymore

mengweieric: yea make sense to remove.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/NowFspAdapter.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexNode;
+import org.opensearch.analytics.spi.FieldStorageInfo;
+import org.opensearch.analytics.spi.ScalarFunctionAdapter;
+
+import java.util.List;
+
+/**
+ * Adapter for PPL {@code NOW([fsp])} (and its {@code CURRENT_TIMESTAMP} / {@code LOCALTIMESTAMP}
```
expani: It's unclear from reading what's `fsp` here. Please elaborate the same in the comments.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
```diff
@@ -510,15 +510,25 @@ public byte[] attachFragmentOnTop(RelNode fragment, byte[] innerBytes) {
         return serializePlan(SubstraitPlanPojoRewriter.rewrite(rewire(inner, wrapper, fieldNames(fragment))));
     }
 
-    private byte[] convertToSubstrait(RelNode fragment) {
-        // TODO: move rewriters that don't touch substrait-specific classes up to the analytics-engine
```
expani, line 537: This TODO shouldn't be removed for assesment later on whether all these Rewriter's really need handling at the Substrait Visitor layer or can be done in the Calcite RelNode itself.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/PplAggregateCallRewriter.java
```diff
@@ -155,6 +156,29 @@ private static AggregateCall rewriteCall(Aggregate agg, AggregateCall call) {
                     call.getName()
                 );
             }
+            // LATE, DEFENSIVE FALLBACK. The PPL `dc`/`distinct_count`/`distinct_count_approx` marker
+            // is normally remapped to the stock APPROX_COUNT_DISTINCT earlier, by the engine planner's
+            // OpenSearchDistinctCountRule (analytics-engine). This case stays as a Substrait-emission-time
+            // safety net for any path that reaches here still carrying the custom marker instance.
```
expani: >safety net for any path that reaches here still carrying the custom marker instance.

Then we should let those query fail and fix it upstream.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/Sha2FunctionAdapter.java
```diff
@@ -99,7 +99,13 @@ public RexNode adapt(RexCall original, List<FieldStorageInfo> fieldStorage, RelO
             return original;
         }
         if (!SUPPORTED_BIT_LENGTHS.contains(bitLen)) {
-            return original;
+            // The bit length is a concrete, unsupported literal (e.g. SHA2(x, 100)). Fail here
+            // with a clear message rather than returning the original call — otherwise the raw
+            // SHA2(string, i32) reaches Substrait conversion and surfaces the cryptic
+            // "Unable to convert call SHA2(string, i32)" instead of a user-actionable error.
+            throw new IllegalArgumentException(
+                "Unsupported SHA2 algorithm [" + bitLen + "]. Supported bit lengths are [224, 256, 384, 512]"
```
expani, line 113: I think this validation should happen way before backend plan adaption. 

Can we add a TODO to tackle this upstream later ? 

I understand the handoff between SQL/PPL plugin Frontend and AnalyticsEngine is only based on the Schema so would need enhancing.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -83,6 +85,25 @@ public class PlannerImpl {
 
     private static final Logger LOGGER = LogManager.getLogger(PlannerImpl.class);
 
+    /**
+     * Like {@link CoreRules#FILTER_PROJECT_TRANSPOSE} but refuses to push a Filter below a Project
+     * that computes any non-deterministic expression (e.g. {@code eval r = rand() | where r > 0}).
+     * The stock rule only guards against window functions ({@code !containsOver()}); pushing past a
+     * {@code rand()} project inlines {@code RAND()} into the predicate, turning a delegatable
+     * {@code ($ref > literal)} comparison into {@code RAND() > literal} on the scan. Lucene
+     * performance-delegation then rejects it (ComparisonSerializer requires (RexInputRef,
```
expani: >Lucene performance-delegation then rejects it 

Then the bug is why Lucene is getting marked as viable for `rand()` function.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchDistinctCountRule.java
```diff
@@ -12,29 +12,51 @@
...
+ *   <li>the PPL {@code dc}/{@code distinct_count_approx} marker UDAF
+ *       (DistinctCountApproxLogicalAggFunction). The SQL plugin lowers it to a user-defined
+ *       {@code SqlAggFunction} whose runtime name is {@code APPROX_COUNT_DISTINCT} (sql#5525;
+ *       the older spelling {@code DISTINCT_COUNT_APPROX} is still accepted defensively). That
+ *       custom operator instance is NOT Calcite's stock {@code APPROX_COUNT_DISTINCT}, so it has
+ *       no Substrait sig and is unknown to the {@link org.opensearch.analytics.spi.AggregateFunction}
+ *       enum; it must be remapped to the stock operator here, before any name-based enum resolution.
+ * </ul>
+ *
+ * <p>This is the primary rewrite path. {@code PplAggregateCallRewriter} (analytics-backend-datafusion)
```
expani: Similar comment as before, we should not have defensive rewrites as they might hide other genuine bugs

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchDistinctCountRule.java
```diff
@@ -12,29 +12,51 @@
...
 
 /**
- * Rewrites single-arg {@code COUNT(DISTINCT x)} → {@code APPROX_COUNT_DISTINCT(x)} before
- * the aggregate is marked by {@link OpenSearchAggregateRule}. Multi-arg distinct falls through
- * to coordinator-gather in {@link OpenSearchAggregateSplitRule}.
+ * Rewrites distinct-count aggregates to Calcite's stock {@code APPROX_COUNT_DISTINCT(x)} before
+ * the aggregate is marked by {@link OpenSearchAggregateRule}. Two shapes are normalized:
+ *
+ * <ul>
+ *   <li>single-arg {@code COUNT(DISTINCT x)} (SQL / implicit distinct count), and
```
expani: For my understanding, How is a user suppose to perform a true Count Distinct without approximation in PPL/SQL ? 

If we are rewriting the single arg `COUNT(DISTINCT <Column>)` to the approximate version.

expani (commented):
Thanks for taking this up @mengweieric

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateFunction.java
mengweieric: Good catch. main now covers `distinct_count_approx` on the analytics-engine route via #22120, so this is no longer needed here. I dropped my changes to this file, it is back to the upstream version.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/AggregateFunction.java
mengweieric: Checked #5525. main now covers `distinct_count_approx` on the analytics-engine route via #22120, so this enum path is no longer needed. I dropped my changes here, it is back to the upstream version.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
mengweieric: Restored the TODO. It now reads as an assessment note on whether each rewriter needs to run at the Substrait visitor layer or can be lifted into the Calcite RelNode level.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/NowFspAdapter.java
mengweieric: Clarified in the javadoc: `fsp` is the MySQL fractional seconds precision, an integer 0..6 selecting how many sub-second digits the timestamp carries (e.g. `now(3)`).

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/PplAggregateCallRewriter.java
mengweieric: Done. The defensive `APPROX_COUNT_DISTINCT` fallback is removed, and main (#22120) now handles the marker in the planner, so nothing rescues it late anymore. This file is back to the upstream version in this PR.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/Sha2FunctionAdapter.java
mengweieric: Added a TODO to move bit-length validation to the SQL/PPL analysis layer once the frontend-to-engine handoff carries function metadata. Kept the backend guard for now so an unsupported length gives a clear error instead of a cryptic Substrait failure.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
mengweieric: Reworded so this reads as what it is: a semantic-correctness guard, not a performance one. Pushing a filter below a projection that computes `rand()` would re-evaluate `RAND()` and draw a fresh value, changing results. The Lucene-viability question is separate and I am not conflating it here.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchDistinctCountRule.java
mengweieric: On this route they cannot today: `dc` / `distinct_count` / `distinct_count_approx` all map to the approximate HLL path, and single-arg `COUNT(DISTINCT)` is rewritten to it too for distributed correctness (summing per-shard exact counts is wrong). Exact distributed distinct-count would need a separate operator and is not implemented. This rule is now owned by main via #22120, so it is out of this PR scope.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchDistinctCountRule.java
mengweieric: Agreed. This is handled in main by #22120, which recognizes the marker properly. This PR no longer touches this file.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/MinMaxBooleanAggregationIT.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    private void createParquetIndex() throws IOException {
+        try {
+            client().performRequest(new Request("DELETE", "/" + INDEX));
+        } catch (Exception ignored) {}
+
+        String body = "{"
+            + "\"settings\": {"
+            + "  \"number_of_shards\": 1,"
```
expani, line 142: Can we cover both single and multi shard case here ? 
Multi shard does change the plan a lot.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -83,6 +85,30 @@ public class PlannerImpl {
...
+     *
+     * <p>This is a <b>semantic-correctness</b> guard, not a delegation/performance one. The stock
+     * rule only guards against window functions ({@code !containsOver()}); pushing a Filter past a
+     * {@code rand()} Project inlines {@code RAND()} into the predicate, so a reference to one
+     * already-computed random value ({@code $ref > literal}) becomes a fresh {@code RAND() > literal}
+     * evaluated again at scan time. That re-evaluates / duplicates the non-deterministic expression
+     * and changes results, so the Filter must stay above the Project regardless of backend.
+     *
+     * <p>(Aside: such an inlined {@code RAND() > literal} predicate is also what gets incorrectly
+     * marked Lucene-delegation-viable today — a separate filter-viability gap where a field-less
```
expani, line 100: >predicate is also what gets incorrectly marked Lucene-delegation-viable today

Okay this seems like a bug then will track it separately.

expani (approved):
Thanks for addressing the comments @mengweieric 

Just one comment around IT's with multi shard as well.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/MinMaxBooleanAggregationIT.java
mengweieric: Added 2-shard coverage. I put it in the `TwoShardReduceTestCase` suite (the #21951 multi-shard convention) rather than inline here: `min_bool` / `max_bool` golden cases over the `merge_coverage` boolean `flag` field now run at 1 shard and 2 shards with a differential equality check plus a pinned golden, so the boolean min/max binding is verified through the two-phase per-shard-partial then coordinator-merge reduce path. This file stays as the focused single-shard, grouped, and mixed-type coverage. Verified locally: `[2shard-val] agg/min_bool` and `agg/max_bool` run and pass at both shard counts.

sandeshkr419 (approved):

PR #21987, opened 2026-06-04
Add filter-delegation ShapeBased IT and fix bug in predicate combiner

expani:
## What
  - Added `FilterDelegationGoldenIT` — pins down filter-delegation behavior across leaf kinds (Dual / Native / Delegated), connectors (AND, OR, NOT, mixed), and the two settings that change delegation decisions.
  - Fixed `ClassCastException` in the predicate combiner on bubbled performance-delegated subtrees — surfaced by q4, q16, q33, q37 under `fuse_dual_viable=false`.

  ## How
  **IT.** `Shape` enum holds the expected `(chosen_backend, tree_shape)` per `(prefer × fuse)` cell; one JUnit method per shape, so `--tests` filters a single shape out of the box.

  **Fix.** Three sites cast a delegated subtree as a leaf, which broke when the combiner bubbled multi-leaf subtrees up. A single helper walks AND/OR/NOT preserving structure and applies the per-leaf op — leaf case unchanged, bubbled case correct.

  ## Add filter-delegation regression IT and fix bubbled-perf cast in combiner

  `FilterDelegationGoldenIT` enumerates leaf-kind × connector × cluster-setting for filter delegation. `Shape` enum owns expected `(chosen_backend, tree_shape)` per cell; one method per shape, so gradle's `--tests` targets a single shape with no plumbing.

  It surfaced a planner bug: when all children of an AND classify as performance-delegated under `fuse_dual_viable=false`, the combiner bubbles them as one delegated node whose subtree is a boolean call instead of a leaf. Three sites cast it as a leaf and threw `ClassCastException` (q4, q16, q33, q37).

  Fixed by routing those sites through a helper that walks AND/OR/NOT preserving structure and applies the per-leaf op. The performance placeholder for a bubbled subtree now carries the original boolean tree, so the driving backend evaluates it natively and still consults the peer per-RG via the annotation id.


sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/FilterDelegationGoldenIT.java
```diff
@@ -0,0 +1,393 @@
...
+                );
+            } else if (want.isPlaceholder() == false) {
+                Map<String, Object> stage = shardFragmentStage(ppl);
+                assertEquals(label + " — chosen_backend", want.chosenBackend(), stage.get("chosen_backend"));
+                assertEquals(label + " — tree_shape", want.treeShape(), stage.get("tree_shape"));
+            }
+        }
+    }
+
+    private Map<String, Object> shardFragmentStage(String ppl) throws Exception {
```
mch2: nice

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/FilterDelegationGoldenIT.java
```diff
@@ -0,0 +1,393 @@
...
+            if (CAPTURE_STAGES) {
+                Map<String, Object> stage = shardFragmentStage(ppl);
+                logger.info(
+                    "CAPTURE shape={} prefer={} fuse={} chosen_backend={} tree_shape={}",
+                    shape, key.prefer(), key.fuse(),
+                    stage.get("chosen_backend"), stage.get("tree_shape")
+                );
+            } else if (want.isPlaceholder() == false) {
+                Map<String, Object> stage = shardFragmentStage(ppl);
+                assertEquals(label + " — chosen_backend", want.chosenBackend(), stage.get("chosen_backend"));
```
mch2: where is this coming from?

mch2 (approved):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DelegatedPredicateCombiner.java
```diff
@@ -129,58 +134,75 @@ private Classified combine(RexCall call, List<Classified> kids, Function<Operato
             }
         }
 
-        if ((correctnessChildren.isEmpty() && performanceChildren.isEmpty()) || multiBackend) {
```
expani, line 132: This is purely due to a adding a debug logger. 
Conditions have been preserved in the same order as before, but don't trust my word for it :P

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/DelegatedPredicateCombiner.java
```diff
@@ -253,20 +275,38 @@ RexNode finalizeDelegated(Delegated d) {
...
         return DelegatedPredicateFunction.makeCall(rexBuilder, d.firstAnnotationId());
     }
 
+    /**
+     * Walks a delegated subtree — a leaf {@link AnnotatedPredicate} or an AND/OR/NOT of
+     * leaves bubbled up through {@link #combine} — applying {@code leafFn} to every leaf
+     * while preserving the boolean structure. The cast on the leaf is intentional: any
+     * other RexNode shape is a planner-invariant violation and should fail loudly.
+     */
+    private RexNode unwrapPreservingConnectors(RexNode node, Function<AnnotatedPredicate, RexNode> leafFn) {
```
expani, line 298: This is the real fix

alchemist51 (approved):

PR #21989, opened 2026-06-04
Use "thin" LTO instead of "fat" which takes substantially less time while still achieving performance gains

LantaoJin:
### Description
The lto setting controls the `-C lto flag` which controls LLVM's [link time optimizations](https://llvm.org/docs/LinkTimeOptimization.html). LTO can produce better optimized code, using whole-program analysis, at the cost of longer linking time.

The valid options are:

- **false**: Performs "thin local LTO" which performs "thin" LTO on the local crate only across its codegen units. No LTO is performed if codegen units is 1 or opt-level is 0.
- **true** or **"fat"**: Performs "fat" LTO which attempts to perform optimizations across all crates within the dependency graph.
- **"thin"**: Performs ["thin" LTO](http://blog.llvm.org/2016/06/thinlto-scalable-and-incremental-lto.html). This is similar to "fat", but takes substantially less time to run while still achieving performance gains similar to "fat".

### Related Issues

### Check List

### DIFF
--- a/sandbox/libs/dataformat-native/rust/Cargo.toml
+++ b/sandbox/libs/dataformat-native/rust/Cargo.toml
@@ -81,8 +81,11 @@ opensearch-repository-gcs = { path = "../../../plugins/native-repository-gcs/src
 opensearch-repository-azure = { path = "../../../plugins/native-repository-azure/src/main/rust" }
 opensearch-repository-fs = { path = "../../../plugins/native-repository-fs/src/main/rust" }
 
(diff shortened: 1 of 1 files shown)


alchemist51:
Datafusion uses the below one for the Cargo.toml:
```
[profile.release]
codegen-units = 1
lto = true
strip = true  
```
Can we run complete clickbench benchmark before making the change?

LantaoJin:
> Can we run complete clickbench benchmark before making the change?

Good view. Let's hold on this till sufficient benchmarking data supports.

PR #21994, opened 2026-06-04
[Bug fix] Remove CoalesceExec for chunk execution

alchemist51:
### Description

Within a single `QueryShardExec` partition, multiple `SegmentChunk`s (each backed by an `IndexedExec` over a contiguous subset of row groups from one segment) were being combined via `UnionExec` + `CoalescePartitionsExec`. `CoalescePartitionsExec` spawns one task per input child and merges their output through an mpsc channel — i.e., it runs the chunks **in parallel inside the partition**.

That parallelism is unnecessary and counter-productive on this path:

- Outer parallelism already exists at the `QueryShardExec` level — `compute_assignments` partitions row groups across `target_partitions`, and DataFusion drives one stream per partition concurrently. Adding a second layer of parallelism inside each partition just multiplies task count without adding useful concurrency.
- Each chunk in a partition shares the same segment, the same evaluator factory invocation pattern, and (per-RG) the same prefetch/decode pipeline that `IndexedStream` already overlaps internally. Running chunks in parallel within one partition contends for the same parquet object-store reads and the same FFM upcall path into Java without freeing up any genuinely independent work.
- The extra `CoalescePartitionsExec` hop adds an mpsc channel per partition and an additional spawn per chunk — pure overhead on the hot path.

### Change

`QueryShardExec::execute` now builds each chunk's `IndexedExec`, executes it immediately to obtain a `SendableRecordBatchStream`, and chains the per-chunk streams **sequentially** within the partition:

- 0 chunks → `EmptyExec` (unchanged)
- 1 chunk → return that stream directly (unchanged)
- N chunks → `futures::stream::iter(streams).flatten()` wrapped in `RecordBatchStreamAdapter`. One task, one stream, chunks consumed back-to-back.

Outer cross-partition parallelism is unchanged — DataFusion still drives N partitions concurrently. We just stop introducing a second layer of parallelism *inside* each partition where it wasn't buying us anything.

### Related Issues

### Check List
- [x] Functionality includes testing. (existing `indexed_table::tests_e2e` suite covers this path; `two_segments_single_partition_exercises_union_coalesce_path` exercises the multi-chunk-per-partition case)


bharath-techie (approved):
Lgtm

PR #22002, opened 2026-06-04
[analytics-engine] Route PPL grok command through the DataFusion backend

ahkcs:
### Description

Routes the PPL `grok <field> '<grok-pattern>'` command through the analytics-engine DataFusion backend. Before this change every grok query on the analytics route failed at the Layer-2 capability registry with:

```
IllegalStateException: No backend supports scalar function [GROK] among [datafusion]
```

`grok` is the grok-syntax sibling of `parse`: both lower (in the SQL plugin) through the same `ParseFunction` UDF, differing only by the third operand (`'grok'` vs `'regex'`), and both return `MAP<VARCHAR,VARCHAR>` consumed by a downstream `ITEM(...)` per named group. `parse` was already wired; `grok` was deliberately deferred (the Rust `parse` UDF's docstring and `ParseAdapter` both note grok "isn't on the analytics path yet").

Two things made grok harder than a one-line capability add, and shaped the design:

1. **Dictionary resolution.** Grok references like `%{COMMONAPACHELOG}` must be recursively expanded against a pattern dictionary before matching. The SQL plugin's grok library is not on OpenSearch core's classpath, so the 108-line default dictionary (forked from logstash, identical to the SQL plugin's) and the `GrokCompiler.compile` resolution algorithm are ported into the Rust UDF.
2. **Regex features.** Resolved grok patterns use negative lookbehind `(?<!…)`, lookahead `(?!…)`, and atomic groups `(?>…)` (e.g. `BASE10NUM`, `IPV4`) that the RE2-based `regex` crate behind `parse` cannot compile. The grok UDF matches with `fancy-regex` instead, whose backtracking engine supports the full `java.util.regex` feature set grok relies on. Matching is **unanchored**, mirroring Java grok's `Matcher.find()` (substring match), in contrast to `parse`'s anchored `^…$`.

This keeps the SQL-plugin lowering 100% unchanged — the same `GROK(field, pattern, 'grok')` RelNode reaches the backend; all grok-specific logic lives in the new Rust UDF.

### Changes

| Area | File | Change |
|---|---|---|
| Capability | `analytics-framework/.../ScalarFunction.java` | New `GROK` enum constant |
| Planner | `DataFusionAnalyticsBackendPlugin.java` | `GROK` in `MAP_RETURNING_PROJECT_OPS` + adapter registration |
| Adapter | `GrokAdapter.java` (new) | Validates literal pattern/method, gates `method=='grok'`, rewrites to local `grok` op |
| Convertor | `DataFusionFragmentConvertor.java` | `FunctionMappings.s(GrokAdapter.LOCAL_GROK_OP, "grok")` |
| Backend (Rust) | `udf/grok.rs` (new), `udf/grok_patterns.txt` (new), `udf/mod.rs`, `Cargo.toml` | grok UDF: dictionary + recursive resolution + `fancy-regex` match + `map<utf8,utf8>` output; `fancy-regex = "=0.14.0"` |
| Catalog | `opensearch_scalar_functions.yaml` | `grok` function entry |
| QA | `GrokCommandIT.java` (new) + `datasets/grok/{mapping,bulk}.json` (new) | Self-contained IT via `POST /_plugins/_ppl` |

### Testing

`CalcitePPLGrokIT` (opensearch-project/sql) on the analytics route, and the new core-side `GrokCommandIT`:

| Test | Pattern (exercises) | Before | After |
|---|---|---|---|
| `testGrokEmail` | `.+@%{HOSTNAME:host}` (RE2-clean; synthetic-name remap) | ❌ | ✅ |
| `testGrokAddressOverriding` | `%{NUMBER} %{GREEDYDATA:address}` (lookbehind + atomic group) | ❌ | ✅ |
| `testGrokApacheLog` | `%{COMMONAPACHELOG}` (recursion + lookbehind/lookahead/atomic) | ❌ | ✅ |
| **`CalcitePPLGrokIT` total** | | **0/3** | **3/3** |
| **QA `GrokCommandIT` total** | | n/a | **3/3** |

- Rust `cargoTest` (`opensearch-datafusion --lib`): **1019/1019** (incl. 6 new grok unit tests).
- `spotlessJavaCheck`, `forbiddenApisMain`, `licenseHeaders`, QA `precommit`: green.

### Check List
- [x] New functionality has been documented (function catalog YAML + javadoc/UDF docstrings).

### Unmuted grok-dependent ITs

Per review, did a pass over the QA suite for tests muted because grok was unsupported:

| IT | Was | Now |
|---|---|---|
| `ComplexRegexPplIT` | `@AwaitsFix` (5/5 queries grok) | unmuted — **5/5 pass** |
| `AppLogsPplIT` q5, q9 | in `getSkipQueries` | unskipped — **pass** |
| `LookupTableQueriesPplIT` q1 | `@AwaitsFix` | unmuted — **q1 passes** |
| `LookupTableQueriesPplIT` q7 | `@AwaitsFix` | still skipped — multi-line `.ppl` can't serialize into the `_ppl` request JSON + uses eventstats/dedup/percentile; **unrelated to grok** |

`GrokAdapter` simplified to drop the redundant method gate (the adapter is bound to the GROK operator by name, so it's only ever invoked for grok).


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GrokAdapter.java
```diff
@@ -0,0 +1,128 @@
...
+            throw new IllegalArgumentException("grok: expected 3 operands (input, pattern, method), got " + operands.size());
+        }
+        // Slot 1 (pattern) and slot 2 (method) must be non-null string literals.
+        // The Rust UDF resolves + compiles the grok pattern once at runtime entry,
+        // so a column-valued pattern would be ill-defined; the method literal is
+        // gated to 'grok' below.
+        requireStringLiteral(operands.get(1), "pattern");
+        String method = requireStringLiteral(operands.get(2), "method");
+
+        if (!"grok".equals(method.toLowerCase(Locale.ROOT))) {
```
mch2: don't think we need this check, we register these adapters up front by name.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/GrokCommandIT.java
```diff
@@ -0,0 +1,136 @@
...
+ * dictionary, recursively resolves {@code %{NAME:field}} references, and matches
+ * with {@code fancy-regex} (the resolved patterns use lookbehind / atomic groups
+ * the RE2-based {@code regex} crate behind {@code parse} cannot compile). Matching
+ * is unanchored, mirroring Java grok's {@code Matcher.find()}.
+ *
+ * <p>The data is a small index whose rows reproduce the {@code bank} (email,
+ * address) and {@code weblogs} (apache log line) values asserted by
+ * {@code CalcitePPLGrokIT}, sorted by {@code id} for deterministic ordering.
+ */
+public class GrokCommandIT extends AnalyticsRestTestCase {
```
mch2, line 39: I think there are also additional ITs in our suite that depend on Grok, can you do a quick pass and unmute those / see if they pass now?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/GrokAdapter.java
ahkcs: Good call — removed. Since the adapter is bound to the GROK operator by name in `scalarFunctionAdapters`, it's only ever invoked for grok, so the method gate was dead code. `GrokAdapter.adapt` now just validates the pattern operand is a string literal (the Rust UDF compiles it once per call) and passes the call through. Updated the class javadoc to match and dropped the now-unused `Locale` import.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/GrokCommandIT.java
ahkcs: Done — did a pass over the suite. Grok appears in three datasets, all previously muted because grok was unsupported:

- **ComplexRegexPplIT** (`@AwaitsFix`, all 5 queries grok) → unmuted, **5/5 pass**.
- **AppLogsPplIT** (q5, q9 grok, were in `getSkipQueries`) → unskipped, **both pass**.
- **LookupTableQueriesPplIT** (`@AwaitsFix`; q1 + q7 grok) → unmuted; **q1 passes**. q7 stays skipped for a reason unrelated to grok — its `.ppl` body is multi-line so the test harness can't serialize it into the `_ppl` request JSON ("Unterminated string"), and it also exercises eventstats/dedup/percentile. Left a comment to that effect.

All four (incl. the new GrokCommandIT, now 2 shards) are green on the QA cluster.

mch2 (approved):

sandbox/plugins/analytics-backend-datafusion/rust/Cargo.toml
```diff
@@ -75,6 +75,13 @@ jsonpath-rust = "=0.7.5"
 # `regex_match` operator (Operator::RegexMatch).
 regex = "=1.12.3"
 
+# grok UDF — the resolved grok dictionary patterns (e.g. BASE10NUM, IPV4) use
+# lookbehind `(?<!…)` and atomic groups `(?>…)` that the RE2-based `regex` crate
+# rejects. `fancy-regex` wraps `regex` with a backtracking engine that supports
+# those constructs (plus lookahead), matching java.util.regex's feature set that
+# the SQL plugin's grok library relies on. Only used by src/udf/grok.rs.
+fancy-regex = "=0.14.0"
```
mch2, line 83: lets pls update the cargo.lock file as well.

mch2 (approved):

sandbox/plugins/analytics-backend-datafusion/rust/Cargo.toml
ahkcs: Opened as a follow-up since this PR already merged: opensearch-project/OpenSearch#22057. It regenerates `sandbox/libs/dataformat-native/rust/Cargo.lock` to include `fancy-regex 0.14.0` (+ its `bit-set`/`bit-vec` transitive deps); no other locked versions change, verified with `cargo metadata --locked`.

PR #22005, opened 2026-06-04
Fix TopK oversampling to use the user's limit, not the system size cap

mch2:
### Description
A query like `... | stats count() by x | sort - c | head 10` was making each shard fetch way too many rows. The SQL plugin wraps every query in a system size-limit (default 10000), and the oversampling code was reading that 10000 instead of the user's `head 10` — so shards fetched ~30000 rows each instead of ~30.

The fix: when there are two stacked sorts (the system cap over the user's head), use the inner one (the real limit) to size the per-shard fetch.

### Related Issues

### Check List


expani (approved):

sandeshkr419 (approved):
Thanks for fixing this!

PR #22008, opened 2026-06-04
Sandbox - Fix crash on single-shard HAVING queries

mch2:
### Description
For `... | stats count() as c by x | where c > 5`, the scan was handed the HAVING filter (on the count result) instead of the WHERE. The scan has no "count" column, so it 500'd with "No field named count(...)"; working around that, the WHERE got dropped and queries returned too many rows. Same bug hit `streamstats ... | where rn <= N`.

The filter picker (Rust + Java) now skips filters above an aggregate or window and uses the WHERE beneath; the HAVING still runs after the aggregate.

Also add missing settings blocks to a few test mappings (parquet config was being silently skipped) and unmute the PPL tests these fixes pass.

### Related Issues

### Check List


mch2:
while this succeeds for HAVING, having similar issues with join & filter delegation on single shards - thinking we can disable delegation for single shard delegated queries until we have a proper fix.

PR #22010, opened 2026-06-05
[analytics-engine] Coerce temporal-vs-string comparisons in subqueries for DataFusion

LantaoJin:
### Description
A PPL temporal comparison such as `ts >= date('2004-01-01')` runs correctly at the top level of a query, but fails when the same comparison sits inside a subquery (scalar `[...]`, I`N [...]`, or `EXISTS [...]`). The query is rejected at Substrait conversion time with:

  Unable to convert call >=(precision_timestamp<0>?, string?).

  This blocks any PPL query that filters a timestamp/date field against a date literal inside a **subquery** -- e.g. TPC-H q15 and q20 on the analytics-engine (DataFusion) path.

  **Root cause**

  At the top level, a temporal comparison reaches the backend as `>=(timestamp, TIMESTAMP(varchar))` -- the PPL TIMESTAMP UDF wrapper carries the coercion, and the backend's scalar-function adapter pass lowers it to a timestamp literal.

  Inside a subquery, `PlannerImpl.removeSubQueries()` runs `RelDecorrelator`, whose expression simplification uses an implicit RexExecutor that executes the PPL TIMESTAMP UDF (its implementor returns a Java String), constant-folding the wrapper away. By the time the backend's adapter pass runs, there is no TIMESTAMP call left to adapt -- only a bare `>=(timestamp, varchar)` that isthmus cannot convert. The literal's surface form is irrelevant: both date('...') (PPL UDF) and DATE '...' (native Calcite literal) regress identically.

  **Fix**

  A new `ComparisonTemporalCoercionAdapter`, registered for the six comparison operators (`=, !=, <, <=, >, >=`) in `DataFusionAnalyticsBackendPlugin`. When one operand of a comparison is temporal (`DATE/TIME/TIMESTAMP`) and the other is a bare character operand, it casts the character side to `TIMESTAMP` (reusing the existing
  `DatePartAdapters.coerceCharacterOperandToTimestamp` helper, which also validates string literals at plan time).

  This recovers the coercion at the comparison itself, independent of how the string operand arrived (folded UDF, native `DATE '...'` literal, or a string column), so it fixes the subquery path without touching the shared decorrelation pipeline. Non-temporal comparisons (`gender = 'F'`, numeric ranges, etc.) pass through unchanged.

  **Testing**

  Added four ITs to DateTimeScalarFunctionsIT (existing calcs dataset, datetime0 timestamp field):

  ```
┌─────────────────────────────────────────────────────────────────┬────────────┬───────┐
  │                              Test                               │   Before   │ After │
  ├─────────────────────────────────────────────────────────────────┼────────────┼───────┤
  │ testTimestampGeDateLiteralTopLevel (baseline)                   │ pass       │ pass  │
  ├─────────────────────────────────────────────────────────────────┼────────────┼───────┤
  │ testTimestampGeDateLiteralInSubquery (IN subquery)              │ FAIL — 400 │ pass  │
  ├─────────────────────────────────────────────────────────────────┼────────────┼───────┤
  │ testTimestampGeDateLiteralScalarSubquery (scalar subquery)      │ FAIL — 400 │ pass  │
  ├─────────────────────────────────────────────────────────────────┼────────────┼───────┤
  │ testTimestampGeDateNativeLiteralInSubquery (DATE '...' literal) │ FAIL — 400 │ pass  │
  └─────────────────────────────────────────────────────────────────┴────────────┴───────┘
```
### Related Issues

### Check List

### CODE REVIEW

**Review by vinaykpud — APPROVED** [review]

**Review by mch2 — APPROVED** [review]

File: sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionAnalyticsBackendPlugin.java [hunk 1]
```diff
@@ -743,7 +745,16 @@ public Map<ScalarFunction, ScalarFunctionAdapter> scalarFunctionAdapters() {
                     Map.entry(ScalarFunction.WEEK, week),
                     Map.entry(ScalarFunction.WEEK_OF_YEAR, week),
                     Map.entry(ScalarFunction.WIDTH_BUCKET, new WidthBucketAdapter()),
-                    Map.entry(ScalarFunction.YEAR, DatePartAdapters.year())
+                    Map.entry(ScalarFunction.YEAR, DatePartAdapters.year()),
+                    // Coerce a bare character operand to TIMESTAMP when compared against a temporal
+                    // operand — recovers the coercion lost when RelDecorrelator folds the PPL
+                    // TIMESTAMP UDF wrapper to a string inside a subquery (see adapter javadoc).
+                    Map.entry(ScalarFunction.EQUALS, comparisonTemporalCoercion),
```
mch2, line 751: hm, i don't love how these common operators will all fall through this just for temporal, but we need to refactor the architecture to resolve this.

LantaoJin: In the final analysis, the `date` and `timestamp` types in the SQL plugin are `UDT`, not the native Calcite type, in order to support nanosecond operations in OpenSearch. However, Mustang now seems to have relaxed its support for nanoseconds.

PR #22023, opened 2026-06-06
feat: read oversampling factor from node opensearch.yml

sandeshkr419:
TopK rewriter was reading analytics.shard_bucket_oversampling_factor from clusterState.metadata().settings() which only contains dynamically-applied cluster settings. Settings from opensearch.yml are node-local and not in cluster metadata.

Fix: read initial value from clusterService.getSettings() (includes yml) at executor startup, subscribe to dynamic updates, and pass the value into PlannerContext for the TopK rewriter to consume.

### Description

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerContext.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerContext.java
@@ -29,6 +29,7 @@ public class PlannerContext {
     private final OpenSearchDistributionTraitDef distributionTraitDef;
     private final boolean profilingEnabled;
     private final boolean preferMetadataDriver;
+    private double oversamplingFactor;
     private int annotationIdCounter;
     private RuleProfilingListener.PlannerProfile lastProfile;
 
@@ -101,6 +102,14 @@ public class PlannerContext {
 
     public ClusterState getClusterState() {
         return clusterState;
+    }
+
+    public double getOversamplingFactor() {
+        return oversamplingFactor;
+    }
+
+    public void setOversamplingFactor(double oversamplingFactor) {
+        this.oversamplingFactor = oversamplingFactor;
     }
 
     public OpenSearchDistributionTraitDef getDistributionTraitDef() {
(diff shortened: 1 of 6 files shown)


mch2 (approved):

sandbox/qa/analytics-engine-rest/build.gradle
```diff
@@ -121,6 +121,7 @@ integTest {
     exclude '**/CoordinatorReduceMemtableIT.class'
     exclude '**/StreamingCoordinatorReduceIT.class'
     exclude '**/QueryCacheIT.class'
+    exclude '**/YmlOversamplingIT.class'
```
mch2, line 124: why excluded?

sandeshkr419: It's excluded from the default integTest task because it needs its own cluster configuration (analytics.shard_bucket_oversampling_factor: 2.0 baked in at startup). The default integTest cluster doesn't have that setting — if YmlOversamplingIT ran on the default cluster, it would test with oversampling=0 and defeat the purpose.

  It runs in its own dedicated task integTestYmlOversampling with its own cluster that has the setting preconfigured. The check.dependsOn(integTestYmlOversampling) ensures it still runs as part of the CI check target — just on a separate cluster instance.

  Same pattern as CoordinatorReduceMemtableIT (excluded from default, runs on integTestMemtable cluster with datafusion.reduce.input_mode=memtable) and QueryCacheIT (runs on integTestQueryCache cluster with its own config).

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerContext.java
```diff
@@ -29,6 +29,7 @@ public class PlannerContext {
     private final OpenSearchDistributionTraitDef distributionTraitDef;
     private final boolean profilingEnabled;
     private final boolean preferMetadataDriver;
+    private double oversamplingFactor;
```
expani, line 32: This should have moved to an inner class that contains `PlannerSettings` 

Otherwise this class will be filled with all the other settings very soon

PR #22025, opened 2026-06-06
Surface spill stats via /_plugins/_analytics_backend_datafusion/stats/spill

gaurav-amz:
## Description

Two related changes to DataFusion spill observability and operability:

  1. Stats — adds a spill section to /_plugins/_analytics_backend_datafusion/stats/spill exposing the configured directory plus disk-volume telemetry per node. Builds on #21913 (datafusion.spill_directory).
  2. Writability validation — implements the deferred reviewer comment from #21913 ("validate if the directory is writable? add to FsHealthCheck?"). Hard-fails node boot when datafusion.spill_directory is set but unwritable, and surfaces a runtime writability flag through the same stats endpoint.

```
  GET /_plugins/_analytics_backend_datafusion/stats/spill
  {
    "cluster_name": "...",
    "nodes": {
      "<id>": {
        "spill": {
          "directory":            "/mnt/spill",
          "disk_total_bytes":     ...,
          "disk_available_bytes": ...,
          "disk_used_bytes":      ...,
          "disk_reserved_bytes":  ...
        }
      }
    }
  }
```

  ## Implementation

  ### Stats endpoint

  - New SpillStats (Writeable + ToXContentFragment) under stats/, mirrors the PartitionGateStats shape.
  - SpillStatsCollector.collect(dir, limit, writable) reads filesystem capacity via Files.getFileStore (one syscall per call). disk_used_bytes is computed as total - available; correct for dedicated spill volumes. When the dir is empty or
  missing, preserves directory + disk_reserved_bytes and zeros the rest.
  - DataFusionStats gains a nullable spillStats field, serialized via writeOptionalWriteable.
  - REST: stat=spill is a valid filter value; transport action filters the section per request.

  ### Writability validation

  - Boot probe — DataFusionPlugin.createComponents runs a one-shot write+delete probe before DataFusionService.start(). On any failure (missing directory, RO mount, permission denied, ENOSPC) it throws IllegalStateException and halts node start. Empty/null spill_directory is a no-op.
  - Runtime monitor — SpillDirectoryHealthMonitor schedules a write+delete probe every 60s via ThreadPool.scheduleWithFixedDelay. A volatile boolean writable flag flows into SpillStats.directory_writable. WARN log on healthy → unhealthy transition, INFO on recovery; silent steady state.


gaurav-amz:
> ## PR Code Analyzer ❗
> **AI-powered '_Code-Diff-Analyzer_' found issues on commit [d1a028e](https://github.com/opensearch-project/OpenSearch/commit/d1a028ea1234b013341fd36fc9628a94f57dfb8a).**
> 
> Path	Line	Severity	Description
> sandbox/plugins/analytics-backend-datafusion/src/main/plugin-metadata/plugin-security.policy	22	medium	The security policy grants `java.io.FilePermission "<>"` with read/readlink/write/delete access. While the comment explains this is necessary because static policy files cannot reference dynamic cluster settings, this effectively allows the plugin to read/write/delete any file on the system (subject only to OS-level permissions). The three documented doPrivileged call sites are narrow and legitimate, but the policy-level grant is maximally broad. Reviewers should confirm this cannot be further scoped (e.g., using a system property set at node startup to represent the spill directory) and that no other code paths in this plugin or its dependencies can exploit the elevated FilePermission.
> The table above displays the top 10 most important findings. **Total: 1 | Critical: 0 | High: 0 | Medium: 1 | Low: 0**
> 
> **Pull Requests Author(s)**: Please update your Pull Request according to the report above.
> 
> **Repository Maintainer(s)**: You can `bypass diff analyzer` by adding label `skip-diff-analyzer` after reviewing the changes carefully, then `re-run failed actions`. To re-enable the analyzer, remove the label, then `re-run all actions`.
> 
> ⚠️ Note: The _Code-Diff-Analyzer_ helps protect against potentially harmful code patterns. Please ensure you have thoroughly reviewed the changes beforehand.
> 
> Thanks.

Spill directory will only holds DataFusion's transient files (intermediate state during sorts/aggregations) — no code, no executables, no long-lived data. Java security policy works at the granularity of who can do what to which path, not what the path stores. Even though spill_directory is just scratch space, we still need write/delete permission on it for DataFusion to spill at all, and for our probe to verify that capability. The grant is plugin-scoped and reachable only via three doPrivileged blocks that all target spill_directory.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/action/stats/RestDataFusionStatsAction.java
```diff
@@ -56,7 +56,8 @@ public class RestDataFusionStatsAction extends BaseRestHandler {
         "stream_next",
         "plan_setup",
         "datanode_gate",
-        "coordinator_gate"
+        "coordinator_gate",
+        "spill"
```
Bukhtawar: Rename to disk_spill

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/health/SpillDirectoryHealthMonitor.java
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * <p>Probe is intentionally minimal — write+delete only, no fsync. Spill is
+ * transient working state, not durable cluster state, so detecting permission /
+ * mount / capacity failures is sufficient.
+ *
+ * <p>Lifecycle: constructed by {@code DataFusionPlugin.createComponents} when
+ * {@code datafusion.spill_directory} is non-empty; cancelled in
+ * {@code DataFusionPlugin.close()}.
+ */
+public final class SpillDirectoryHealthMonitor implements Runnable {
```
Bukhtawar: Can we not extend FsHealthService to probe plugin paths?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -689,11 +725,50 @@ public Supplier<AnalyticsBackendNativeMemoryStats> getAnalyticsBackendNativeMemo
...
     }
 
+    /**
+     * One-shot writability probe for the configured spill directory. Runs once at
+     * {@link #createComponents} when {@code datafusion.spill_directory} is non-empty.
+     * Throws {@link IllegalStateException} on any failure so node boot halts.
+     *
+     * <p>Empty/null input is a no-op — spill is disabled and there is nothing to probe.
+     */
+    static void probeSpillDirectoryWritable(String spillDirectory) {
```
Bukhtawar: Move this logic outside `DatafusionPlugin`

sandbox/plugins/analytics-backend-datafusion/src/main/plugin-metadata/plugin-security.policy
```diff
@@ -0,0 +1,24 @@
...
+  // it only resolves system-property variables (${java.io.tmpdir}, ${codebase.*}),
+  // not cluster settings. Granting <<ALL FILES>> with read/readlink/write/delete is the
+  // narrowest scope we can express at policy-load time. Actual reach is limited by
+  // SpecialPermission.check() + AccessController.doPrivileged at the three call sites
+  // that touch the spill directory:
+  //   - DataFusionPlugin.probeSpillDirectoryWritable (boot probe)
+  //   - SpillDirectoryHealthMonitor.probeOnce        (runtime probe, 60s cadence)
+  //   - SpillStatsCollector.collect                  (stats endpoint, on-demand)
+  // No execute permission is granted; no other plugin code uses doPrivileged to access files.
+  permission java.io.FilePermission "<<ALL FILES>>", "read,readlink,write,delete";
```
Bukhtawar: restrict to spill directory

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/action/stats/RestDataFusionStatsAction.java
gaurav-amz: Sure, Will do.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
gaurav-amz: Yes, we can move it to the existing FsHealthCheck. The idea was to keep this logic abstracted from the core implementation. If you're okay with adding it there, I'll make that change in the next revision.

sandbox/plugins/analytics-backend-datafusion/src/main/plugin-metadata/plugin-security.policy
gaurav-amz: Ack.

Bukhtawar (commented):
```
// Before creating DiskManager, clear stale spill files
  let spill_path = PathBuf::from(spill_dir);
  if spill_path.exists() {
      let _ = fs::remove_dir_all(&spill_path);
  }
  fs::create_dir_all(&spill_path).ok();
```

We might need something like this on start up

gaurav-amz:
> ```
> // Before creating DiskManager, clear stale spill files
>   let spill_path = PathBuf::from(spill_dir);
>   if spill_path.exists() {
>       let _ = fs::remove_dir_all(&spill_path);
>   }
>   fs::create_dir_all(&spill_path).ok();
> ```
> 
> We might need something like this on start up

Done with fail fast on failure in cleanup.

Bukhtawar (approved):
Thanks lgtm

PR #22027, opened 2026-06-06
fix(analytics-engine): zero null text cells before passing results to…

mch2:
### Description
A QTF query could crash randomly when results had null text values. The null cells held leftover garbage memory, and the native engine read that garbage as if it were a real string and crashed.

Fix: clear the null cells before handing results to the native engine.

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/ViewVectorSanitizer.java
```diff
@@ -0,0 +1,42 @@
...
+final class ViewVectorSanitizer {
+
+    private ViewVectorSanitizer() {}
+
+    static void sanitize(VectorSchemaRoot root) {
+        int rows = root.getRowCount();
+        for (FieldVector vec : root.getFieldVectors()) {
+            if (!(vec instanceof BaseVariableWidthViewVector view) || view.getNullCount() == 0) continue;
+            ArrowBuf viewBuffer = view.getDataBuffer();
+            for (int row = 0; row < rows; row++) {
```
expani: This could get expensive for performance given we are going through each row ?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionReduceSink.java
```diff
@@ -222,6 +222,7 @@ private void feedToSender(DatafusionPartitionSender sender, VectorSchemaRoot bat
                     + batch.getSchema()
             );
         }
+        ViewVectorSanitizer.sanitize(batch);
```
expani: If the bug is arising in QTF path why does it require changes in the generic Reduce Sink ? Shouldn't a decorator specific to QTF be doing it instead ?

mch2: you're right - this should move to the stitcher instead of the sink.  Moving

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/ViewVectorSanitizer.java
mch2: yep, moving this to the stitcher mitigates some of this, one we  only have topk in stitcher so row size is small, and two we check for null count to short circuit cols without nulls.

sandeshkr419 (approved):

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/exec/stage/ViewVectorSanitizerTests.java
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * <p>Regression for the QTF late-materialization {@code interleave len=0, index=<garbage>} panic.
+ * A Java-built VSR fed across the Arrow C Data Interface can carry view columns (Utf8View /
+ * BinaryView) whose null slots were never written — Arrow unsets the validity bit but leaves the
+ * 16-byte view as uninitialized garbage. DataFusion's interleave reads the view length prefix
+ * unconditionally; a garbage length &gt; 12 dereferences a variadic data buffer that does not exist
+ * on an all-null column, panicking. The sanitizer must zero every null view slot before export so
+ * it reads back as a valid length-0 inline view.
+ */
+public class ViewVectorSanitizerTests extends OpenSearchTestCase {
```
expani: Should we add a QTF based E2E test to catch this ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/stage/ViewVectorSanitizer.java
expani: Yes if it's stiticher then should be okay unless a User does a large limit. 

nit: This method can be moved to `sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/VectorUtils.java`

expani (approved):

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/exec/stage/ViewVectorSanitizerTests.java
mch2: i was only able to repro this reliably with higher doc counts with gradle run ~300k docs, given the allocator buffers are clean in the IT with lower doc counts.

PR #22029, opened 2026-06-07
Push bare LIMIT down to shards (head N without ORDER BY)

mch2:
### Description
A head N with no sort left the limit only on the coordinator, so every shard streamed its full scan and the coordinator trimmed to N. Push a collation-free fetch below the exchange so each shard caps at N locally. Safe without an order: the coordinator's N rows are a subset of the union of each shard's N rows.

Extends the existing collated-sort pushdown to the bare-fetch case, reusing its offset handling. Plan-shape tests cover the pushed shape plus single-shard, offset-widening, and UNION ALL. Updates testPureLimit_2shard, which had pinned the old shape.

Also tested with sql's default 10k limit applied:
```
OpenSearchProject(@timestamp=[$0], viableBackends=[[datafusion]])
  OpenSearchSort(fetch=[10000], viableBackends=[[datafusion]])
    OpenSearchExchangeReducer(viableBackends=[[datafusion]], exchange=[ExchangeInfo[distributionType=SINGLETON, partitionKeyIndices=[]]])
      OpenSearchSort(fetch=[10000], viableBackends=[[datafusion]])
        OpenSearchTableScan(table=[[lim_test]], viableBackends=[[lucene, datafusion]])
```

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
@@ -73,6 +73,11 @@ public final class OpenSearchSortPushdownRewriter {
                 if (bound != null && canComputeShardFetch(bound) && isPushTarget(below)) {
                     return new RewriteContext(sort, bound, below);
                 }
+            } else if (sort.fetch != null && canComputeShardFetch(sort) && isPushTarget(sort.getInput())) {
+                // Bare LIMIT, no ORDER BY (`head N`): push a fetch-only Sort below the ER so each
+                // shard caps at N instead of streaming its whole scan. Safe with no order — the
+                // coordinator just needs some N rows, and each shard's local N supplies them.
+                return new RewriteContext(sort, sort, sort.getInput());
             }
             OpenSearchSort carry = (collated == false && sort.fetch != null) ? sort : null;
             return find(sort.getInput(), carry);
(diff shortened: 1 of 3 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortPushdownRewriter.java
```diff
@@ -73,6 +73,11 @@ private static RewriteContext find(RelNode node, OpenSearchSort fetchSortAbove)
                 if (bound != null && canComputeShardFetch(bound) && isPushTarget(below)) {
                     return new RewriteContext(sort, bound, below);
                 }
+            } else if (sort.fetch != null && canComputeShardFetch(sort) && isPushTarget(sort.getInput())) {
+                // Bare LIMIT, no ORDER BY (`head N`): push a fetch-only Sort below the ER so each
+                // shard caps at N instead of streaming its whole scan. Safe with no order — the
```
expani, line 78: `Safe with no order` Which check is satisfying this ?

mch2: we are in the else block, we only reach this if theres no collation:
```java
            if (collated) {
                // Bound = this Sort if it carries a fetch; else the enclosing pure-fetch Sort, but
                // only when this Sort has no offset of its own (which the enclosing one can't honor).
                OpenSearchSort bound = sort.fetch != null ? sort : (sort.offset == null ? fetchSortAbove : null);
                RelNode below = sort.getInput();
                if (bound != null && canComputeShardFetch(bound) && isPushTarget(below)) {
                    return new RewriteContext(sort, bound, below);
                }
            } else if (sort.fetch != null && canComputeShardFetch(sort) && isPushTarget(sort.getInput())) {
```

expani (approved):

sandeshkr419 (approved):

PR #22030, opened 2026-06-07
[Analytics-Engine] Bias plan cost toward aggregate-split via row-count estimates

mch2:
### Description
Bias plan cost toward aggregate-split via row-count estimates

Volcano can't pick between split (PARTIAL→ER→FINAL) and non-split (SINGLE-at-coordinator) plans when every operator reports Calcite's default row count of 100. Both look cost-equivalent, the tie-break picks one randomly, and PPL frontend shapes (auto-NOT-NULL filter on synthesized columns) routinely land on the wrong one. q19 went from ~2s to 80+s.

Two estimate overrides fix it:

  - OpenSearchTableScan.estimateRowCount returns shards × 10M (placeholder until real stats land via the new StatisticsCollector entry point).
  - OpenSearchAggregate.estimateRowCount returns input / 100 for grouped aggregates.

 Result: split ships ~100× fewer rows than non-split, cost arithmetic prefers split for any multi-shard input. Single-shard input still passes the existing SINGLETON gate and stays SINGLE.

  StatisticsCollector + TableStatistics(indexName, rowCount, shardCount) are the entry point for future IndicesStats-cache integration; today they return shard count from cluster metadata and rowCount=0 (fallback path).

  Tests: AggregateSplitCostTests covers 1-shard / 2-shard plan shape and direct row-count arithmetic. Full planner test suite + IT pass.

### Related Issues

### Check List


expani:
Okay I again went into the rabbit hole of Calcite Volcano with this issue. 

Some useful things came out : 

- The fix is working in conjunction with infiniteCost emitted at `OpenSearchAggregate#computeSelfCost` which is not implicit. Maybe we make it explicit with a comment. 

- Another way to fix this is overriding `getKind` in `AnnotatedPredicate` 
```
+    @Override
+    public SqlKind getKind() {
+        return original.getKind();
+    }
```
But this breaks other predicate functions, which requires unwrapping using a custom `RelMdSelectivity` handler without touching row count estimates. 


- Proper distribution-trait propagation would make SINGLE-over-RANDOM unsatisfiable by construction and is still the right long term fix. 


Approving this as it is the right temporary workaround until Trait Propagation is done across the Planner.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -172,10 +186,59 @@ public void onMatch(RelOptRuleCall call) {
...
+     * <p>The walk stops at the first {@link OpenSearchExchangeReducer}: a window <em>below</em> a
+     * shuffle doesn't gather <em>this</em> aggregate's input, so a partitioned aggregate sitting
+     * above that exchange (e.g. the lower {@code stats} in {@code stats ... | eval w=..OVER() |
+     * stats ..}) still splits correctly. Only the aggregate directly above the window-induced
+     * gather skips. Mirrors the chain walk in {@code OpenSearchLateMaterializationRewriter}.
+     */
+    private static boolean childGatheredByWindow(RelNode node) {
+        RelNode cur = RelNodeUtils.unwrapHep(node);
+        while (cur != null) {
+            if (cur instanceof OpenSearchExchangeReducer) {
```
expani: This seems a bit weird because exchange is inserted by the Split rule and the Split rule is also looking for an exchange. Can you help understand this a bit better ?

expani (approved):

sandeshkr419 (approved):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
expani:
```
private static RelNode unwrapForWalk(RelNode n) {
  ▎     n = RelNodeUtils.unwrapHep(n);
  ▎     return (n instanceof RelSubset s) ? s.getBestOrOriginal() : n;
  ▎ }
```

We need this for queries like 
```
source = <index> | eventstats sum(ResolutionWidth) as w | where w > 0 | stats count() by SearchPhrase
```

Otherwise we get a redundant split like
```
  Aggregate(mode=FINAL)              ← \
    Aggregate(mode=PARTIAL)          ←  } redundant split
      Project(SearchPhrase)
        Filter(w > 0)
          Project(w = SUM OVER ())
            ExchangeReducer(SINGLETON)
              TableScan

  WITH fix (getBestOrOriginal() unwrap):
  Aggregate(mode=SINGLE)             ← single pass
    Project(SearchPhrase)
      Filter(w > 0)
        Project(w = SUM OVER ())
          ExchangeReducer(SINGLETON)
            TableScan
```

mch2: this branch actually never fires, the window check only runs after we've confirmed the input is random, and an exchange always outputs singleton. So there can never be an exchange to find on that path anyway.

PR #22032, opened 2026-06-07
[Analytics-Engine] Route text-field filter to the keyword subfield

mch2:
### Description
Equality on a `text` field returned zero rows: the delegated predicate built a term query on the analyzed text field, but a term query does not analyze its input, so it never matched the analyzed (e.g. lowercased) indexed tokens. Keyword fields were unaffected.

Capture a text field's keyword multifield name in FieldStorageInfo (populated by FieldStorageResolver from the mapping's `fields` block), and have EqualsSerializer target `field.<keyword>` when present. Text fields without a keyword subfield keep the base name. Mirrors the SQL engine's OpenSearchTextType.convertTextToKeyword.

### Related Issues

### Check List

### DIFF
--- a/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldStorageInfo.java
+++ b/sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldStorageInfo.java
@@ -31,12 +31,10 @@ public class FieldStorageInfo {
     private final boolean derived;
     private final LinkedHashSet<String> dependsOnPhysicalCols;
     /**
-     * For a {@code text} field that declares a {@code keyword} multifield, the subfield's name
-     * (e.g. {@code "keyword"}); {@code null} otherwise. Exact-equality predicates must target
-     * this subfield — a term query on the analyzed text field does not match (the analyzer
-     * lowercases/tokenizes the indexed value but a term query does not analyze its input).
+     * Subfield to target for exact-match (term) predicates — e.g. a text field's keyword
+     * multifield — or {@code null} when the field is queried directly. Resolved from the mapping.
      */
-    private final String keywordSubfield;
+    private final String exactMatchSubfield;
 
     public FieldStorageInfo(
         String fieldName,
@@ -52,7 +50,7 @@ public class FieldStorageInfo {
         this(fieldName, mappingType, fieldType, docValueFormats, indexFormats, storedFieldFormats, derived, new LinkedHashSet<>());
     }
 
-    /** Physical-field ctor with an explicit keyword multifield name (or {@code null} if none). */
+    /** Physical-field ctor with an explicit exact-match subfield name (or {@code null} if none). */
     public FieldStorageInfo(
         String fieldName,
         String mappingType,
@@ -61,7 +59,7 @@ public class FieldStorageInfo {
         List<String> indexFormats,
         List<String> storedFieldFormats,
         boolean derived,
-        String keywordSubfield
+        String exactMatchSubfield
     ) {
         this(
             fieldName,
@@ -72,7 +70,7 @@ public class FieldStorageInfo {
             storedFieldFormats,
             derived,
             new LinkedHashSet<>(),
-            keywordSubfield
+            exactMatchSubfield
         );
     }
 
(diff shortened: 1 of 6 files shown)


sandeshkr419 (approved):

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/serializers/EqualsSerializer.java
```diff
@@ -58,7 +58,14 @@ public QueryBuilder buildQueryBuilder(RexCall call, List<FieldStorageInfo> field
             );
         }
 
-        String fieldName = FieldStorageInfo.resolve(fieldStorage, columnRef.getIndex()).getFieldName();
+        FieldStorageInfo field = FieldStorageInfo.resolve(fieldStorage, columnRef.getIndex());
+        // Exact equality on a text field must target its keyword multifield: a term query on the
+        // analyzed text field never matches (the analyzer lowercases/tokenizes the indexed value,
+        // but a term query does not analyze its input). Mirrors the SQL engine's convertTextToKeyword.
+        // Text fields without a keyword subfield fall back to the base name (analyzed term — same as SQL).
+        String fieldName = field.getKeywordSubfield() != null
```
expani: Can't this instead check based on the FieldType ? Given the suffix will always be `.keyword` and we don't need to store it as a separate string in `FieldStorageInfo` ?

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldStorageInfo.java
```diff
@@ -30,6 +30,13 @@ public class FieldStorageInfo {
     private final List<String> storedFieldFormats;
     private final boolean derived;
     private final LinkedHashSet<String> dependsOnPhysicalCols;
+    /**
+     * For a {@code text} field that declares a {@code keyword} multifield, the subfield's name
+     * (e.g. {@code "keyword"}); {@code null} otherwise. Exact-equality predicates must target
+     * this subfield — a term query on the analyzed text field does not match (the analyzer
+     * lowercases/tokenizes the indexed value but a term query does not analyze its input).
+     */
+    private final String keywordSubfield;
```
expani: This kinda ties the whole class to Keyword field whereas it was linked to type via `FieldType`

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/serializers/EqualsSerializer.java
mch2: keyword is the default suffix, but it can be named something else.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldStorageInfo.java
mch2: actually i suppose i could update the Enum and decouple this.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/serializers/EqualsSerializer.java
expani: I meant instead of checking for `getKeywordSubfield` being null. We can instead check that the FieldType is Text and has a sub mapping of Keyword, then we attach the suffix `.keyword` 

Hmm, does Vanilla OpenSearch handle it the same way ? Is it not going to the Mapper and getting handled automatically ?

mch2: vanilla OpenSearch doesn't route this to the keyword subfield automatically, sql is rewriting the field name to the keyword subfield manually before building the term query.

sandbox/libs/analytics-framework/src/main/java/org/opensearch/analytics/spi/FieldStorageInfo.java
mch2: this got more convoluted typing to our FieldType enum - i've renamed this to "exactMatchSubfield" vs keyword itself.

PR #22039, opened 2026-06-07
Reorder DynamicLimitPool::try_grow and split RSS-pressure error type

gaurav-amz:
## What

Reorders `DynamicLimitPool::try_grow` and splits RSS-pressure errors out from pool-fill errors. Also lowers the default jemalloc decay from 30s to 1s.

## Why

`try_grow` is the allocation hot path for the DataFusion `MemoryPool`. It used to check jemalloc-resident bytes against `pool × 85%` **before** the pool-limit CAS — and reject when over the threshold. The intent was "trigger spill," but the spill path itself calls `try_grow` for its sort buffer, hits the same gate, and fails.

[Testing] On warm tier the cdylib's idle resident is ~130 MB, so any pool below ~155 MB had the gate trip on every allocation; under sustained concurrent load it triggered on production-default sizes too. Net effect: spill-capable operators (HashAggregate, Sort, TopK) couldn't recover when working set pushed RSS past the threshold.

## Fix

`try_grow` now resolves in this order:

1. **Pool CAS** — succeeds when accounting has room.
2. **Override on stale fill** — CAS-fail + jemalloc shows headroom → allow (handles accounting drift).
3. **RSS-critical** — CAS-fail + jemalloc above critical → cancel via new `rss_critical_error`.
4. **Pool-fill** — otherwise return `pool_limit_error`, which DataFusion converts into `operator.spill()`.

The pre-CAS RSS gate is removed; `execution_spill_threshold` now gates the override path only.

### Error split

Three rejection paths previously emitted the same `"Failed to allocate ... 0 available out of N limit"` message with `available=0` hard-coded for the RSS paths. Now:

- **Pool fill** → `pool_limit_error` with real `available = limit - used`. Java parser unchanged.
- **RSS critical** → new `rss_critical_error` carrying real `pool_used` / `pool_limit` / `rss_bytes`. `NativeErrorConverter` registers the new key phrase and converts to `CircuitBreakingException` (HTTP 429) — same retry surface, distinct message for triage.

### Decay

`MALLOC_CONF` decay reduced from `30000` to `1000` for both `dirty` and `muzzy`, with matching cluster-setting defaults. RSS now reflects current usage within ~1s instead of carrying a 60s lookback. `native.jemalloc.{dirty,muzzy}_decay_ms` remain runtime-tunable, so operators can revert without redeploy if a workload regresses.

## Behaviour changes

| Scenario | Before | After |
|---|---|---|
| Pool empty, RSS at 85–95% × pool | Reject pre-CAS | Allocate via CAS |
| Pool full, RSS at 85–95% × pool | (unreachable) | CAS fail → pool-fill error → operator spills |
| Pool full, RSS > 95% × pool | (unreachable) | CAS fail → RSS-critical cancel |
| Cancel-path Java conversion | RuntimeException → HTTP 500 | CircuitBreakingException → HTTP 429 |
| `tripped_count` | RSS-spill + RSS-critical + pool-fill | RSS-critical + pool-fill |

`execution.spill_threshold` is now an override gate, not a rejection threshold. `execution.critical_threshold` is now a post-CAS cancel only. Both stay dynamic.

## Test plan

**Unit (Rust):**
- `test_rss_critical_rejects_after_cas_fail_when_rss_exceeds_critical` — pre-fills pool, asserts rejection comes via `rss_critical_error`.
- `test_rss_critical_passes_when_rss_below_critical` — large pool, allocation succeeds via CAS.
- `test_pool_fill_rejection_emits_pool_limit_error_not_rss_error` — locks the error-message contract for the Java parser.

**Cluster (3 warm nodes, 1M docs / 200k unique groups, `count avg max min sum by host`):**

| Pool | Concurrency | Result | Spill captured |
|---|---|---|---|
| 256 MB | 1 | 200, 2.1s | 0 (working set fits) |
| 15 MB | 8 | 4×200 + 4×429 | 74 files / 180 MB |
| 128 MB | 8 | 8×200 | 51 files / 230 MB |

Captured spill files verified as real Arrow IPC streams. The 128 MB run is the dispositive case — pre-fix the RSS gate would have fired at 109 MB but cdylib idle is ~130 MB, so every allocation would have been rejected. Post-fix all 8 queries spill and complete.

A no-spill control at pool=1 GB / threshold=0.99 produced 8/8 success, 0 files, `tripped_count` delta 0 — confirming spill engages only when the pool genuinely fills.

### DIFF
--- a/sandbox/libs/dataformat-native/rust/lib/src/lib.rs
+++ b/sandbox/libs/dataformat-native/rust/lib/src/lib.rs
@@ -35,8 +35,10 @@
 /// zero CPU overhead and negligible memory overhead. Profiling is activated on-demand
 /// via cluster settings at runtime. Dev/test builds (cargo build without --release)
 /// do not include profiling support.
+// Decay set to 1s so jemalloc resident reflects current usage within ~1s of free().
+// Runtime-tunable via cluster settings `native.jemalloc.{dirty,muzzy}_decay_ms`.
 #[export_name = "malloc_conf"]
-pub static MALLOC_CONF: &[u8] = b"dirty_decay_ms:30000,muzzy_decay_ms:30000,lg_tcache_max:16\0";
+pub static MALLOC_CONF: &[u8] = b"dirty_decay_ms:1000,muzzy_decay_ms:1000,lg_tcache_max:16\0";
 
 #[global_allocator]
 static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
@@ -112,61 +112,10 @@ impl MemoryPool for DynamicLimitPool {
         reservation: &MemoryReservation,
         additional: usize,
     ) -> Result<(), DataFusionError> {
-        // Two-tier RSS guard for in-flight queries:
-        //
-        // - operator (85%): reject to trigger spill. The operator will flush its hash
-        //   table to disk and retry. This is recoverable — the query continues.
-        // - critical (95%): hard reject to prevent OOM. Last resort — even spill can't
-        //   save the node at this pressure level.
-        //
-        // Between 85-95%, the override (below) ensures spill sort buffers can still
-        // allocate: when the CAS fails and RSS has dropped below operator threshold
-        // (because the hash table was freed during spill), the override fires and
-        // allows the sort allocation through.
-        //
-        // The adaptive cache (cached_resident_bytes) bypasses the 100ms cache when
-        // the last value was above operator threshold — so the override sees the
-        // fresh (lower) RSS after spill frees memory, not a stale peak.
-        let limit = self.dynamic_limit.load(Ordering::Acquire);
-        let resident = crate::memory_guard::cached_resident_bytes();
-        if resident > 0 && limit >= 16 * 1024 * 1024 {
-            let thresholds = crate::memory_guard::get_thresholds();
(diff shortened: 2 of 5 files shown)


sandbox/libs/dataformat-native/rust/lib/src/lib.rs
```diff
@@ -35,8 +35,10 @@
 /// zero CPU overhead and negligible memory overhead. Profiling is activated on-demand
 /// via cluster settings at runtime. Dev/test builds (cargo build without --release)
 /// do not include profiling support.
+// Decay set to 1s so jemalloc resident reflects current usage within ~1s of free().
+// Runtime-tunable via cluster settings `native.jemalloc.{dirty,muzzy}_decay_ms`.
 #[export_name = "malloc_conf"]
-pub static MALLOC_CONF: &[u8] = b"dirty_decay_ms:30000,muzzy_decay_ms:30000,lg_tcache_max:16\0";
+pub static MALLOC_CONF: &[u8] = b"dirty_decay_ms:1000,muzzy_decay_ms:1000,lg_tcache_max:16\0";
```
Bukhtawar, line 41: Might need benchmarks for this change

sandbox/plugins/analytics-backend-datafusion/rust/src/memory.rs
```diff
@@ -112,61 +112,10 @@ impl MemoryPool for DynamicLimitPool {
...
-                    additional,
-                    reservation.consumer().name(),
-                    reservation.size(),
-                    0,
-                    limit,
-                ));
-            }
-        }
-
+        // Decision order: CAS → override (stale fill) → RSS-critical → pool-fill.
```
Bukhtawar, line 115: This is critical for stability, why is this being removed

PR #22044, opened 2026-06-07
[sandbox] Delegate LIKE predicates to Lucene + add per-backend delegation block-list

sandeshkr419:
## What

  Extends DataFusion→Lucene filter delegation beyond `EQUALS`:

  - **`LIKE` / `ILIKE` delegation.** A new `LikeSerializer` emits a Lucene `PrefixQuery` for `'p%'` patterns (cheap term-dict scan) and `WildcardQuery` otherwise (`%`/`_` → `*`/`?`). `ILIKE` (PPL default `like`) is matched case-insensitively. Shared `SqlLikePattern` classifier picks the shape.
  - **Per-backend delegation block-list** — an operational kill-switch: `analytics.delegation.<backend>.blocked_predicates` (affix, dynamic). Lets an operator stop a predicate from being delegated to a backend without a deploy.

  ## Decisions (to ease review)

  - **Delegation is a performance pre-filter.** DataFusion re-verifies the exact predicate (`Collector ∧ residual`), so the Lucene query only needs to be a  **superset** of matches. This is why `LIKE` on `text` (per-token) vs `keyword`  (whole-value) can't change results — keyword and text are treated together.
  - **No new `PREFIX` op.** `ScalarFunction.PREFIX` is orphaned (nothing emits it); prefix is handled as a *shape of `LIKE`*, not a separate serializer.
  - **Negation needs no handling.** Calcite forbids a negated `LIKE` as a `RexCall`, so `NOT LIKE` arrives as `NOT(LIKE(...))` and is composed by the existing boolean path; the serializer only sees positive `LIKE`.
  - **Block-list reuses `ScalarFunction`** (not a parallel enum). Validated against the **runtime registry** at setting-update time: the namespace must be a FILTER-delegation acceptor (today only `lucene` — any other is rejected), and the predicate must have a serializer on that backend.
  - **Block-list is safe-by-construction.** Enforced at marking (`OpenSearchFilterRule`); a blocked backend is dropped from a predicate's viable set only when a non-blocked backend survives — blocking never makes a predicateunexecutable.

  ## Notes

 - **3 commits** (block-list / LIKE / a tiny unrelated `-Werror` redundant-cast fix in `WindowPlanShapeTests`), each reviewable alone.
  - **Tested at every layer:** unit (pattern classifier, serializer dispatch,  registry wiring, marking-viability, block-list validation + marking) **and end-to-end** — `LikeDelegationIT` runs each LIKE shape on a live cluster and asserts results are identical with delegation on vs. blocked via the setting  (same answer, different execution path), plus that a non-`lucene` namespace is rejected.
  - Follow-up: `IN`/`RANGE` delegation (arrive Sarg-packed) in a separate PR.

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerContext.java
```diff
@@ -32,6 +33,9 @@ public class PlannerContext {
     private double oversamplingFactor;
     private int annotationIdCounter;
     private RuleProfilingListener.PlannerProfile lastProfile;
+    // Per-backend delegation block-list. Defaults to empty (nothing blocked); DefaultPlanExecutor
+    // injects the live, settings-backed instance via setDelegationBlockList before planning.
+    private DelegationBlockList delegationBlockList = DelegationBlockList.empty();
```
expani: Can we please keep all the settings in a Top wrapper class that's called `PlannerSettings` ? 
```
class PlannerSettings {
 
     private double oversamplingFactor;
      private DelegationBlockList delegationBlockList
 
     public PlannerSettings(ClusterSettings settings) {
            oversamplingFactor = .....
            delegationBlockList = .....
      }

      public DelegationBlockList getDelegationBlockList();

      public double getOversamplingFactor();

}
````

We should take this in the current PR given it was punted during TopK

sandeshkr419: Refactored.

sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_table/substrait_to_tree.rs
sandeshkr419, line 1: **Problem:** extract_filter_expr walked the logical plan looking for the first Filter node and returned its predicate. But when the plan has a HAVING clause (e.g. ... | stats count() as c by x | where c > 0), the HAVING filter sits above the Aggregate. Its predicate references aggregate outputs (like count(Int64(1))), not scan columns. The indexed executor tried to lower this against the raw scan schema → "No field named count(Int64(1))" error.

  **Fix:** A Filter is only returned if its input reaches_scan_without_schema_change — i.e., the filter's predicate is over scan-schema columns (goes through only passthrough nodes like Projection/SubqueryAlias/Sort/Limit down to a TableScan). If the input hits an Aggregate/Window/Join/Union (schema-changing nodes), the Filter is a HAVING predicate and is skipped. The function continues descending past the Aggregate to find the real WHERE filter below it.

New helper reaches_scan_without_schema_change: Returns true only if the plan reaches a TableScan through schema-preserving nodes. Returns false if an Aggregate (or any schema-changing node) is in the path.

expani: This seems like a parallel issue not related to adding support for LIKE predicate function. Can we take this up in a different PR ? I think Join and Union queries are also affected by this

sandeshkr419:
Breaking this monolith to small PRs:


- Setting : https://github.com/opensearch-project/OpenSearch/pull/22071
- is null, is not null, not equal: https://github.com/opensearch-project/OpenSearch/pull/22075
- bare comparisons (>, >=, <, <=): https://github.com/opensearch-project/OpenSearch/pull/22076
- in / range: https://github.com/opensearch-project/OpenSearch/pull/22077

PR #22045, opened 2026-06-07
[analytics-engine] PPL date/time/timestamp fixes — format, overloads, span types, invalid literals

vinaykpud:
## Summary

Fixes a batch of PPL date/time/timestamp gaps on the analytics-engine (parquet-primary) path: format rendering, missing function overloads, invalid-literal error surface, span() column type preservation, and sub-second precision preservation.

Pairs with [opensearch-project/sql#5521](https://github.com/opensearch-project/sql/pull/5521). Tests that depend on the sql siblings are gated with `@AwaitsFix` and listed at the bottom of this description.

### `datetime stringify uses space, time has no epoch prefix`
- Timestamp cells render as `yyyy-MM-dd HH:mm:ss` (space separator instead of ISO-T).
- Time-only cells render as `HH:mm:ss` (no `1970-01-01` epoch prefix).
- `cast(boolean AS varchar)` rewrites to upper-case `TRUE` / `FALSE` to match PPL semantics.

### `PPL date function gap fixes: week modes, format tokens, tz NULL, cast TIME`
- `WEEK()` / `WEEK_OF_YEAR()` default to mode 0 (Sunday-start, MySQL semantics) instead of ISO.
- `STR_TO_DATE` parses `%b` (Jan/Feb/.../Dec) abbreviations.
- `DATE_FORMAT` adds `%c %U %u %V %v %P` tokens.
- `UNIX_TIMESTAMP` accepts numeric `YYYYMMDDhhmmss` literals.
- `CONVERT_TZ` returns NULL on out-of-range tz instead of throwing.
- `CAST(<datetime-string> AS TIME)` strips the date prefix.
- New Rust UDFs: `os_week` (8-mode MySQL `WEEK`) and `os_strftime` (shared parse / render engine for `DATE_FORMAT` and `STR_TO_DATE`).

### `PPL invalid date literals reject with typed format-hint message`
- Pre-isthmus validators surface malformed string literals as `IllegalArgumentException` with the typed message `"<type>:<value> in unsupported format, please use '<pattern>'"`, replacing the opaque Arrow Parser error / 500 StreamException.
- Covers `CAST(<lit> AS DATE/TIME/TIMESTAMP)`, `TIMESTAMPADD/DIFF` string args, DATE/TIME/DATE_FORMAT/TIME_FORMAT/DATE_PART operands, and silent-success cases (`HOUR('99:99:99')`, `STR_TO_DATE` out-of-range month/day).
- `DATETIME(<invalid>)` folds to NULL TIMESTAMP at plan time.

### `Add missing PPL date function overloads`
Ten function overloads that previously failed isthmus lowering with "Unable to convert call ..." now bind:
- `DATETIME(string, tz)` — plan-time fold for literals; column input rewrites to `convert_tz`.
- `now(fsp)` / `sysdate(fsp)` — drops the FSP argument.
- `DATE_ADD(TIME, interval)` and `DATE_SUB(TIME, interval)` — anchors TIME to today UTC.
- `TIMESTAMPADD` standalone — lowers to `DATETIME_PLUS` + interval.
- `TIMESTAMPDIFF` standalone — `to_unixtime` delta with unit scaling.
- `DATE_PART(unit, TIME)` — today-UTC anchor on TIME operands.
- `FROM_UNIXTIME(epoch, format)` — composes `date_format` around 1-arg UDF.
- `SPAN(timestamp, N, 'M' | 'q' | 'y')` — calendar-unit `date_bin` rewrite.
- `CONVERT_TZ` invalid timestamp literal — folds to typed NULL.
- `microsecond()` — single CAST to `TIMESTAMP(6)` so the µs fraction survives `date_part`.

### `PPL span(date) and span(time) preserve column type`
- Adds `DateOnlyType` / `TimeOnlyType` UDT markers in the schema builder.
- A `date` / `date_nanos` field with a date-only or time-only `format` mapping now surfaces with the right user-visible label.
- `span(<date col>, …)` returns a date-typed bucket column (not widened to timestamp); `span(<time col>, …)` returns a time-typed bucket.
- Substrait wire stays `Timestamp(ms)` — only the schema label changes.

### `PPL TIMESTAMP(date, time) combine and HOUR-of-TIME-literal fold`
Two capabilities lifted from #22014:
- `TIMESTAMP(<date col>, <time col>)` 2-arg combine via `from_unixtime(to_unixtime(date) + to_unixtime(time))`.
- `HOUR / MINUTE / SECOND / MICROSECOND` on a TIME literal folds to `BIGINT` at plan time (avoids the `precision_time` substrait gap).

### `test(sandbox): backfill CAST(time-string AS DATE) rejection test`
- Backfills a single missing unit test in `CastTemporalLiteralValidatorTests`.

### `test(sandbox): add DatetimeCoverageIT for cluster-all regression gate`
- End-to-end IT — 59 methods covering every shape this PR fixes plus min/max boundary values for DATE / TIME / TIMESTAMP.
- One-to-one mapping between methods and commits — any cluster regression surfaces as a single named failure.

### `test(sandbox): @AwaitsFix on tests that require sql cluster A/D`
- 18 tests temporarily annotated `@AwaitsFix` because they depend on the sql/core siblings (cluster A schema label + cluster D list-element formatter).
- To be reverted once the sql PR merges.

### `fix(sandbox): DateOnly/TimeOnly UDTs carry source-type precision (date→3, date_nanos→9)`

[#22049](https://github.com/opensearch-project/OpenSearch/pull/22049) added sub-second precision (`date→3`, `date_nanos→9`) on the plain-TIMESTAMP path. The `DateOnlyType` / `TimeOnlyType` UDT branch short-circuited before that switch, so a `date_nanos` field with a date-only or time-only `format` produced `TIMESTAMP(0)` and broke multi-shard sort. Mirror the same precision selection on the UDT factories.

### Format / output rendering

```ppl
source=calcs | head 1 | eval ts = cast('2024-01-15 12:00:00' as timestamp) | fields ts
# was: "2024-01-15T12:00:00"     now: "2024-01-15 12:00:00"

source=calcs | head 5 | stats list(time1) as l
# was: ["1970-01-01T19:36:22", ...]   now: ["19:36:22", ...]

source=calcs | head 1 | eval s = cast(true as string) | fields s
# was: "true"     now: "TRUE"

source=calcs | head 1 | eval m = microsecond('2024-01-15 12:00:00.123456') | fields m
# was: 123000     now: 123456

source=calcs | head 1 | eval f = date_format('2024-01-15 12:00:00.123456', '%f') | fields f
# was: "123000"   now: "123456"

source=calcs | head 1 | eval ts = cast('2023-10-01 12:00:00.123456' as timestamp) | fields ts
# was: "2023-10-01 12:00:00.123"   now: "2023-10-01 12:00:00.123456"
```

### Function overloads — previously failed with `Unable to convert call ...`

```ppl
source=calcs | head 1 | eval f = DATETIME('2008-01-01 02:00:00', '+10:00') | fields f
# now: "2008-01-01 12:00:00"

source=calcs | head 1 | eval f = DATE_ADD(TIME('08:40:00'), INTERVAL 1 DAY) | fields f
source=calcs | head 1 | eval f = DATE_SUB(TIME('08:40:00'), INTERVAL 1 DAY) | fields f

source=calcs | head 1 | eval f = TIMESTAMPADD(DAY, 5, '2016-03-01 00:00:00') | fields f
# now: "2016-03-06 00:00:00"

source=calcs | head 1 | eval f = TIMESTAMPDIFF(DAY, '2020-01-01 00:00:00', '2020-01-05 00:00:00') | fields f
# now: 4

source=calcs | head 1 | eval h = HOUR(TIME('17:30:00')) | fields h
# now: 17

source=calcs | head 1 | eval f = FROM_UNIXTIME(1611657712, '%Y-%m-%d') | fields f
# now: "2021-01-26"

source=calcs | head 1 | eval f = sysdate(0) | fields f      # now: current ts (FSP arg dropped)

source=calcs | stats count() as c by span(datetime0, 3month) as bucket
# now: 200 + buckets at 3-month boundaries

source=calcs | head 1 | eval ts = TIMESTAMP(date1, time1) | fields ts   # 2-arg date+time combine
```

### Format tokens — date_format / str_to_date

```ppl
source=calcs | head 1 | eval f = date_format(timestamp('1998-01-31 13:14:15'), '%c %U %u %V %v') | fields f
# now: "01 4 4 4 4"   (was: missing tokens or wrong values)

source=calcs | head 1 | eval f = str_to_date('1-May-13', '%d-%b-%y') | fields f
# was: "2013-01-01" (silent month=1)   now: "2013-05-01"
```

### Mode / range corrections

```ppl
source=calcs | head 1 | eval w = week(date('2008-02-20')) | fields w
# was: 8 (ISO)   now: 7 (MySQL mode 0, Sunday-start)

source=calcs | head 1 | eval u = unix_timestamp(20771122143845) | fields u
# was: pass-through 20771122143845   now: parsed as YYYYMMDDhhmmss → 3404817525

source=calcs | head 1 | eval f = convert_tz('2021-05-12 11:34:50', '+09:00', '+15:00') | fields f
# was: HTTP 400 IllegalArgumentException   now: null (out-of-range tz)

source=calcs | head 1 | eval t = cast('1985-10-09 12:00:00' as TIME) | fields t
# was: HTTP 500 parse error   now: "12:00:00" (date prefix stripped)
```

### date_nanos sub-second precision (multi-shard sort)

```ppl
# date_nanos field with a date-only format on a 2-shard parquet-backed index:
source=ix | sort d | fields d
# was: HTTP 500 — "Timestamp(ms) got Timestamp(ns)" at coordinator RowConverter
# now: HTTP 200 — rows surface in ascending order
```

### Span column type preservation

```ppl
source=date_formats | stats count() as c by span(basic_date, 1day) as date_span
# was: schema date_span:timestamp, value "1984-04-12 00:00:00"
# now: schema date_span:date,      value "1984-04-12"

source=date_formats | stats count() as c by span(time1, 1minute) as time_span
# was: schema time_span:timestamp, value "1970-01-01 09:00:00"
# now: schema time_span:time,      value "09:00:00"
```

### Invalid literal rejection

```ppl
source=calcs | head 1 | eval a = DATE('2025-13-02')                                  | fields a
source=calcs | head 1 | eval a = TIME('99:99:99')                                    | fields a
source=calcs | head 1 | eval a = TIMESTAMP('2025-15-26 25:00:00')                    | fields a
source=calcs | head 1 | eval a = cast('09:07:42' as DATE)                            | fields a
source=calcs | head 1 | eval a = HOUR('99:99:99')                                    | fields a   # was: silent 200 with 0
source=calcs | head 1 | eval a = STR_TO_DATE('01,13,2013', '%d,%m,%Y')               | fields a   # was: silent month=13

# All now: HTTP 400 IllegalArgumentException
# {"details": "<type>:<value> in unsupported format, please use '<pattern>'"}
# (replacing prior HTTP 500 / opaque Arrow Parser error / silent wrong row)

source=calcs | head 1 | eval a = DATETIME('2025-13-02 10:00:00') | fields a
# now: HTTP 200, returns null (legacy SQL DATETIME(<invalid>) semantics)
```

## Tests gated `@AwaitsFix` until sql PR merges

The following pass when the sql/core sibling PR is installed; without it they fail with substrait schema mismatches or unbound function-call lowerings.

- `DatetimeCoverageIT` — 14 methods (cluster A span tests, cluster B 2-arg DATETIME, cluster E µs preservation, cluster F format/cast/maketime, max-time / max-timestamp boundaries).
- `DateTimeScalarFunctionsIT.testDateAddMillisecondIntervalOnTimestampColumn`
- `TimestampFunctionIT.testShapeCTimeLiteralFoldsWithTodayUtc`
- `TimestampFunctionIT.testTimeEqualsDateDoesNotCrash`
- `TwoShardCommandIT.testReduceCorrectnessAcrossTwoShards`

Sites are searchable via `grep -rn "pull/<TBD>" sandbox/qa/` and will be unmuted in a follow-up commit once the sql PR merges.

### DIFF
--- a/sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/DateFormatClassifier.java
+++ b/sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/DateFormatClassifier.java
@@ -1,0 +1,197 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.analytics.schema;
+
+import java.util.HashSet;
+import java.util.Locale;
+import java.util.Set;
+
+/**
+ * Classifies an OpenSearch {@code date} field's {@code format} as date-only, time-only,
+ * or full timestamp. Mirrors the SQL plugin's {@code OpenSearchDateType} rules
+ * (named-format allow-lists + custom-pattern symbol scan); sandbox sits below sql-core in
+ * the dependency graph so the lists are duplicated here. Format syntax is the OpenSearch
+ * {@code DateFormatter} grammar — components separated by {@code ||}.
+ */
+public final class DateFormatClassifier {
+
+    /** Classification result for a {@code date} mapping's {@code format}. */
+    public enum Kind {
+        DATE_ONLY,
+        TIME_ONLY,
+        TIMESTAMP
+    }
+
+    /** Named formats with year/month/day only. */
+    private static final Set<String> NAMED_DATE_FORMATS = setOf(
+        "basic_date",
+        "basic_ordinal_date",
+        "date",
+        "strict_date",
+        "year_month_day",
+        "strict_year_month_day",
+        "ordinal_date",
+        "strict_ordinal_date",
+        "week_date",
+        "strict_week_date",
+        "weekyear_week_day",
+        "strict_weekyear_week_day"
+    );
+
+    /** Named formats with hour/minute/second only. */
+    private static final Set<String> NAMED_TIME_FORMATS = setOf(
+        "basic_time",
+        "basic_time_no_millis",
+        "basic_t_time",
+        "basic_t_time_no_millis",
+        "time",
+        "strict_time",
+        "time_no_millis",
+        "strict_time_no_millis",
+        "hour_minute_second_fraction",
+        "strict_hour_minute_second_fraction",
+        "hour_minute_second_millis",
+        "strict_hour_minute_second_millis",
+        "hour_minute_second",
+        "strict_hour_minute_second",
+        "hour_minute",
+        "strict_hour_minute",
+        "hour",
+        "strict_hour",
+        "t_time",
+        "strict_t_time",
+        "t_time_no_millis",
+        "strict_t_time_no_millis"
+    );
+
+    /** Named formats forcing TIMESTAMP: full datetime + numeric epochs + incomplete date (year, year_month, …). */
+    private static final Set<String> NAMED_TIMESTAMP_FORMATS = setOf(
+        // datetime
+        "iso8601",
+        "basic_date_time",
+        "basic_date_time_no_millis",
+        "basic_ordinal_date_time",
+        "basic_ordinal_date_time_no_millis",
+        "basic_week_date_time",
+        "strict_basic_week_date_time",
+        "basic_week_date_time_no_millis",
+        "strict_basic_week_date_time_no_millis",
+        "basic_week_date",
+        "strict_basic_week_date",
+        "date_optional_time",
+        "strict_date_optional_time",
+        "strict_date_optional_time_nanos",
+        "date_time",
+        "strict_date_time",
+        "date_time_no_millis",
+        "strict_date_time_no_millis",
+        "date_hour_minute_second_fraction",
+        "strict_date_hour_minute_second_fraction",
+        "date_hour_minute_second_millis",
+        "strict_date_hour_minute_second_millis",
+        "date_hour_minute_second",
+        "strict_date_hour_minute_second",
+        "date_hour_minute",
+        "strict_date_hour_minute",
+        "date_hour",
+        "strict_date_hour",
+        "ordinal_date_time",
+        "strict_ordinal_date_time",
+        "ordinal_date_time_no_millis",
+        "strict_ordinal_date_time_no_millis",
+        "week_date_time",
+        "strict_week_date_time",
+        "week_date_time_no_millis",
+        "strict_week_date_time_no_millis",
+        // numeric epochs
+        "epoch_millis",
+        "epoch_second",
+        "epoch_micros",
+        // incomplete date — can't be downgraded to DATE
+        "year_month",
+        "strict_year_month",
+        "year",
+        "strict_year",
+        "week_year",
+        "week_year_week",
+        "strict_weekyear_week",
+        "weekyear",
+        "strict_weekyear"
+    );
+
+    /** Custom-pattern symbols for time components (matches SQL plugin's CUSTOM_FORMAT_TIME_SYMBOLS). */
+    private static final String CUSTOM_TIME_SYMBOLS = "nNASsmHkKha";
+
+    /** Custom-pattern symbols for date components (matches SQL plugin's CUSTOM_FORMAT_DATE_SYMBOLS). */
+    private static final String CUSTOM_DATE_SYMBOLS = "FecEWwYqQgdMLDyuG";
+
+    private DateFormatClassifier() {}
+
+    /** Null / empty maps to {@link Kind#TIMESTAMP} (OpenSearch default). */
+    public static Kind classify(String format) {
+        if (format == null || format.isEmpty()) {
+            return Kind.TIMESTAMP;
+        }
+        boolean sawDate = false;
+        boolean sawTime = false;
+        for (String component : format.split("\\|\\|")) {
+            String trimmed = component.trim();
+            if (trimmed.isEmpty()) {
+                continue;
+            }
+            String stripped = strip8Prefix(trimmed);
+            String lower = stripped.toLowerCase(Locale.ROOT);
+            if (NAMED_TIMESTAMP_FORMATS.contains(lower)) {
+                return Kind.TIMESTAMP;
+            }
+            if (NAMED_DATE_FORMATS.contains(lower)) {
+                sawDate = true;
+            } else if (NAMED_TIME_FORMATS.contains(lower)) {
+                sawTime = true;
+            } else {
+                // unknown name → treat as custom pattern; scan for symbols
+                boolean[] flags = scanCustomPattern(stripped);
+                sawDate |= flags[0];
+                sawTime |= flags[1];
+            }
+            if (sawDate && sawTime) {
+                return Kind.TIMESTAMP;
+            }
+        }
+        if (sawDate && !sawTime) return Kind.DATE_ONLY;
+        if (sawTime && !sawDate) return Kind.TIME_ONLY;
+        return Kind.TIMESTAMP;
+    }
+
+    /** Returns {[hasDateSymbol, hasTimeSymbol]} for a custom DateTimeFormatter pattern. */
+    private static boolean[] scanCustomPattern(String pattern) {
+        boolean date = false;
+        boolean time = false;
+        for (int i = 0; i < pattern.length(); i++) {
+            char c = pattern.charAt(i);
+            if (!date && CUSTOM_DATE_SYMBOLS.indexOf(c) >= 0) date = true;
+            if (!time && CUSTOM_TIME_SYMBOLS.indexOf(c) >= 0) time = true;
+            if (date && time) break;
+        }
+        return new boolean[] { date, time };
+    }
+
+    /** Strips leading {@code 8} prefix used by OpenSearch's joda compatibility shim. */
+    private static String strip8Prefix(String s) {
+        return s.startsWith("8") ? s.substring(1) : s;
+    }
+
+    private static Set<String> setOf(String... values) {
+        Set<String> set = new HashSet<>();
+        for (String v : values) {
+            set.add(v.toLowerCase(Locale.ROOT));
+        }
+        return set;
+    }
+}
--- a/sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/DateFormatClassifier.java
+++ b/sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/DateFormatClassifier.java
@@ -28,6 +28,9 @@ public final class DateFormatClassifier {
         TIMESTAMP
     }
 
+    // TODO: replace with constants from OpenSearch core's DateFormatters registry once they're
+    // exposed for external use; today we mirror the well-known names locally to avoid coupling
+    // to internal APIs.
     /** Named formats with year/month/day only. */
     private static final Set<String> NAMED_DATE_FORMATS = setOf(
         "basic_date",
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/CastTemporalLiteralValidator.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/CastTemporalLiteralValidator.java
@@ -18,6 +18,7 @@ import org.apache.calcite.sql.SqlKind;
 import org.apache.calcite.sql.type.SqlTypeName;
 
 /** Pre-isthmus pass that rejects malformed CAST(string-lit AS DATE/TIME/TIMESTAMP) and TIMESTAMPADD/DIFF string args. */
+// TODO: fold into BackendPlanAdaption alongside CastToVarcharRewriter (unified rejection/coercion model).
 final class CastTemporalLiteralValidator {
 
     private CastTemporalLiteralValidator() {}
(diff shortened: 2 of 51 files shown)


sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/CastTemporalLiteralValidator.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.calcite.rel.RelHomogeneousShuttle;
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rex.RexCall;
+import org.apache.calcite.rex.RexLiteral;
+import org.apache.calcite.rex.RexNode;
+import org.apache.calcite.rex.RexShuttle;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.type.SqlTypeName;
+
+/** Pre-isthmus pass that rejects malformed CAST(string-lit AS DATE/TIME/TIMESTAMP) and TIMESTAMPADD/DIFF string args. */
```
expani, line 20: Rejects or coerces ? Either way seems like an extension to BackendPlanAdaption if it doesn't neatly fit the current model. We can start with a TODO for this and `CastToVarchar`

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConvertTzAdapter.java
```diff
@@ -84,13 +61,46 @@ class ConvertTzAdapter implements ScalarFunctionAdapter {
...
+        }
+        if (!(unwrapped instanceof RexLiteral literal)) return null;
+        SqlTypeName typeName = literal.getType().getSqlTypeName();
+        if (typeName != SqlTypeName.CHAR && typeName != SqlTypeName.VARCHAR) return null;
+        String value = literal.getValueAs(String.class);
+        if (value == null) return null;
+        try {
+            java.time.LocalDateTime.parse(value.replace(' ', 'T'));
+            return null;
+        } catch (java.time.format.DateTimeParseException ignored) {}
```
expani: Is it okay to silently ignore, what's the end user response ? 
Please add a comment if this is expected.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
```diff
@@ -459,6 +459,8 @@ private byte[] convertToSubstrait(RelNode fragment) {
         preprocessed = PplAggregateCallRewriter.rewrite(preprocessed);
```
expani, line 462: Okay was not aware we have accumulated tons of rewriters at the backend layer. 

Please add a TODO to move all of these up unless they are using some Substrait specific classes

sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/DateFormatClassifier.java
```diff
@@ -0,0 +1,197 @@
...
+
+    /** Classification result for a {@code date} mapping's {@code format}. */
+    public enum Kind {
+        DATE_ONLY,
+        TIME_ONLY,
+        TIMESTAMP
+    }
+
+    /** Named formats with year/month/day only. */
+    private static final Set<String> NAMED_DATE_FORMATS = setOf(
```
expani, line 35: Are these not exposed by OpenSearch today ? 
We can add a TODO to open them up for use.

sandbox/plugins/analytics-backend-datafusion/src/main/resources/opensearch_scalar_functions.yaml
```diff
@@ -952,6 +936,35 @@ scalar_functions:
           - { value: "varchar<L2>", name: "format" }
         return: precision_timestamp<6>
 
+  - name: "os_week"
```
expani, line 975: nit: We can create opensearch_date_functions.yaml

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatetimeLiteralValidator.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.calcite.rex.RexLiteral;
+import org.apache.calcite.rex.RexNode;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.time.format.DateTimeParseException;
+import java.util.Locale;
+
+/** Plan-time validation for string-literal datetime operands; rejects malformed input with the format-hint message. */
```
expani, line 20: This should be happening in the Marking phase. 
Let's add a TODO for the same.

sandbox/qa/analytics-engine-coordinator/src/internalClusterTest/java/org/opensearch/be/datafusion/QtfDerivedAboveProjectIT.java
```diff
@@ -137,9 +135,9 @@ public void testQtfFires_outerProjectRunsDerivedUpperOnStage3() throws Exception
                     String expectedUrl = "HTTPS://EXAMPLE.COM/PAGE" + i;
                     assertEquals("row " + i + " UPPER(URL) mismatch", expectedUrl, row[0]);
 
-                    // EventDate is returned as a LocalDateTime by the executor (TIMESTAMP at
-                    // midnight on the seeded date).
-                    LocalDateTime expectedDate = LocalDate.of(2026, 5, i + 1).atStartOfDay();
+                    // EventDate is returned by the executor as a formatted string
```
expani, line 138: Shouldn't this be dependent on the mapping of field in the index ? 

Given the only change is this, it feels like there is a different bug elsewhere ?

expani (commented):
Thanks for taking this up @vinaykpud 

Most of my comments are around making the code short which could be follow-ups.

Just have a couple of things that seems odd specially the QTF test fix

sandbox/plugins/analytics-backend-datafusion/src/main/resources/opensearch_scalar_functions.yaml
vinaykpud: Sure, will take up as followup

sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/DateFormatClassifier.java
vinaykpud: Added a TODO. Most of these names map to constants inside `org.opensearch.common.time.DateFormatters`, but they're not currently exposed for plugin use left a marker to swap to upstream constants if/when they're opened up.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionFragmentConvertor.java
vinaykpud: Added a TODO. moving them up to analytics-engine for cross-backend reuse is the right cleanup, leaving the marker for follow-up.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatetimeLiteralValidator.java
vinaykpud: Added a TODO to move this into the Marking phase as a follow-up.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/ConvertTzAdapter.java
vinaykpud: silent NULL is the documented MySQL CONVERT_TZ contract for unparseable timestamps. Verified against `ConvertTZFunctionIT.nullFromGarbageInput2`, `nullDateTimeInvalidDateValueFebruary/April/Month` which all assert null rows or invalid inputs. Throwing here would break the contract. Added an explanatory comment at the catch + reference to the IT, and tidied the imports.

sandbox/qa/analytics-engine-coordinator/src/internalClusterTest/java/org/opensearch/be/datafusion/QtfDerivedAboveProjectIT.java
vinaykpud: The rendering does depend on the mapping. EventDate is type: date with the default format (strict_date_optional_time, which is TIMESTAMP-shaped), so the value flows through `ArrowValues#formatTimestamp` and gets the 00:00:00 suffix.

A field with an explicitly date-only format (e.g., format: yyyy-MM-dd) would be classified by DateFormatClassifier as DateOnlyType and return "2026-05-01" without the time suffix.

The reason this commit touched the assertion: the IT was previously hardcoded to expect LocalDateTime object equality, which broke when `ArrowValues#formatTimestamp` switched scalar temporal cells from raw LocalDateTime to formatted strings. The fix is just updating the assertion to the new return contract.

vinaykpud:
Thanks @expani , replied to the comments

expani (approved):

mch2 (approved):

PR #22047, opened 2026-06-08
[Analytics-Engine] Add Lucene-side serializers for LIKE, !=, IN, IS NULL, range, REGEXP

rishabhmaurya:
## Summary

Wires up `DelegatedPredicateSerializer` implementations for the standard SQL comparison predicates that today fall through to DataFusion even when the shard's Lucene secondary could resolve them on the term dictionary.

| Class | Operator | Lucene QueryBuilder |
|---|---|---|
| `LikeSerializer` | SQL `LIKE` | `WildcardQueryBuilder` (SQL `%`→`*`, `_`→`?`, backslash escapes) |
| `NotEqualsSerializer` | `col != x` | `BoolQuery.mustNot(TermQuery)` |
| `InSerializer` | `IN (...)` | `TermsQueryBuilder` |
| `IsNullSerializer` | `IS NULL` / `IS NOT NULL` | `BoolQuery.mustNot(ExistsQuery)` / `ExistsQuery` |
| `RangeSerializer` | `<`, `<=`, `>`, `>=` | `RangeQueryBuilder` (mirrors operator if literal is on the left) |
| `RegexpSerializer` | `REGEXP` | `RegexpQueryBuilder` |

`LuceneSubtreeConvertor` already wraps `NOT`/`AND`/`OR` connectives, so `NOT LIKE` and similar shapes compose without convertor changes. Each new serializer follows the `EqualsSerializer` template — accepts the `(RexInputRef, RexLiteral)` shape (and the literal-on-the-left mirror where it makes sense), routes the literal through `CalciteToOSMapperConversionUtils.literalToOpenSearchValue`, and lets the field's mapper do final type coercion at `toQuery(qsc)` time.

`LikeSerializer` accepts both 2- and 3-operand forms — Calcite supplies the SQL default escape `\` as a third operand even when the user didn't write an `ESCAPE` clause. A custom escape character is rejected explicitly so a future caller doesn't get silent backslash semantics.

A DEBUG-level breadcrumb in `LuceneSubtreeConvertor` (`[analytics.delegated] op=[<fn>] queryBuilder=[<json>]`) lets the next person flipping a capability on confirm engagement end-to-end without attaching a debugger.

## Why advertised capability stayed at `EQUALS` only

The capability advertisement in `LuceneAnalyticsBackendPlugin.STANDARD_OPS` is unchanged. Two upstream gaps make unconditional delegation regress when validated end-to-end on a 100 M-doc ClickBench cluster:

1. **No shard-level cost model** — declaring a backend viable for an operator picks Lucene whenever it can answer the predicate, even when DataFusion would scan the parquet column faster (e.g. `LIKE '%substr%'` on a 30 M-cardinality `URL` field).
2. **`approx_distinct` partial→final mismatch** — the partial→final aggregate split fails with `reduce_eval(approx_distinct): expected Binary state` when filter delegation rearranges the operator chain on `dc(...)` queries.

Both are tracked separately. When either lands, flip `STANDARD_OPS` on per-operator (or gate it behind an `analytics.delegation.lucene_filter_predicates` cluster setting). The serializer code is in place and ready.

## Verification

End-to-end engagement was verified by temporarily re-advertising `NOT_EQUALS` in `STANDARD_OPS` and running:

```
source = clickbench | where MobilePhoneModel != '' | head 5
```

The coordinator log captured the breadcrumb:

```
[2026-06-08T06:31:31,242][INFO ][o.o.b.l.LuceneSubtreeConvertor] [fsst-coord]
  [analytics.delegated] op=[NOT_EQUALS] queryBuilder=[
    { "bool" : { "must_not" : [ { "term" : { "MobilePhoneModel" :
                                             { "value" : "", "boost" : 1.0 } } } ],
                 "adjust_pure_negative" : true, "boost" : 1.0 } }
  ]
```

The serialized bytes were shipped over Flight to the data node, deserialized, and executed against the Lucene secondary. The query returned 5 rows correctly.

## ClickBench scope numbers — analytics path before vs. after this PR

3-node r5.2xlarge cluster (1 coord + 2 data, 32 GB heap each, 100 M ClickBench docs, 2 primary shards, 0 replicas). Median of 3 PPL runs against the analytics path on `composite-parquet` index. **Before** = current `main` build. **After** = this PR (capability still `EQUALS`-only — serializer code dormant on the hot path). Stock OpenSearch + opensearch-sql DSL latency on a hardware-matched cluster is included for context only, not as the optimization target.

| qid | predicate family | before (ms) | after (ms) | Δ | stock DSL (ms) |
|---|---|---:|---:|---:|---:|
| q11 | `MobilePhoneModel != ''` | 306 | 272 | -11 % | 345 |
| q12 | `MobilePhoneModel != ''` (more group keys) | 349 | 315 | -10 % | 585 |
| q13 | `SearchPhrase != ''` | 937 | 911 | -3 % | 661 |
| q14 | `SearchPhrase != '' \| dc(UserID)` | OOM | OOM | — | 2,381 |
| q15 | `SearchPhrase != '' \| count by SearchEngineID, SearchPhrase` | OOM* | OOM | — | 22,097 |
| q21 | `LIKE '%google%' \| count` | 2,076 | 2,011** | flat | 3,143 |
| q22 | `LIKE '%google%' AND SearchPhrase != ''` | 2,219 | — | — | 3,105 |
| q23 | `LIKE '%Google%' AND NOT LIKE '%.google.%' AND SearchPhrase != ''` | 12,045 | — | — | 4,239 |
| q24 | `LIKE '%google%' \| sort EventTime` | 3,509 | — | — | 3,114 |
| q25 | `SearchPhrase != '' \| sort EventTime` | 10,109 | 10,664 | +5 % | 58 |
| q26 | `SearchPhrase != '' \| sort SearchPhrase` | 11,906 | 12,345 | +4 % | 510 |
| q27 | `SearchPhrase != '' \| sort EventTime, SearchPhrase` | 10,623 | 10,779 | +1 % | 45 |
| q28 | `URL != '' \| stats avg(length(URL)), count by CounterID` | 1,661 | 1,821 | +10 % | err*** |
| q31 | `SearchPhrase != ''` (count/sum/avg by …) | 914 | 921 | flat | 2,414 |
| q32 | `SearchPhrase != ''` (count/sum/avg by …) | 1,042 | 1,071 | flat | 2,846 |
| q37 | `CounterID=62 AND date range AND URL != ''` | 773 | 886 | +15 %† | 216 |
| q38 | `CounterID=62 AND date range AND Title != ''` | 683 | 655 | flat | 119 |

\* Pre-existing OOM in DataFusion's `dc(...)` over 30 M+ groups (`Resources exhausted: Memory Exhausted while SpillPool (DiskManager is disabled)`); independent of this PR.
\*\* q21 re-ran cleanly after restarting d1; q22–q24 hit the same OOM during the rerun and weren't re-collected because DataFusion crashed the data node before they finished. Their predicate shapes are identical to operators in q21 / q11–q15 which were re-collected.
\*\*\* Stock DSL `q28` fails with `all shards failed` (a script-runtime issue in the DSL form) — not a fair comparison.
† Within run-to-run noise on this cluster (cold caches; shard merges in flight).

**Reading this:** the "after" column matches the "before" column within noise, because `STANDARD_OPS` still advertises only `EQUALS`. That's by design for this PR — the new serializer code is in tree but not on the hot path until a separate commit flips the capability. Once both upstream blockers above land, the queries that benefit most from this PR are q11/q12 (`!=`), q21/q22/q23 (`LIKE`), and q37/q38 (range + `!=`), where the analytics path today either falls through to DataFusion or already does, and stock-DSL Lucene completes in a fraction of the time.

## Test plan

- [x] `./gradlew -Dsandbox.enabled=true sandbox:plugins:analytics-backend-lucene:assemble` succeeds.
- [x] Existing `QuerySerializerRegistryTests` pass (no behaviour change for advertised operators).
- [ ] Add unit tests for the new serializers — round-trip serialize → deserialize → assert `QueryBuilder` shape, mirroring the existing `WildcardQuerySerializer` tests. Out of scope for the draft; will follow before un-drafting.
- [x] End-to-end engagement verified on a 3-node r5.2xlarge ClickBench cluster (see Verification section).

### DIFF
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneAnalyticsBackendPlugin.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneAnalyticsBackendPlugin.java
@@ -63,9 +63,22 @@ public class LuceneAnalyticsBackendPlugin implements AnalyticsSearchBackendPlugi
     // registered in QuerySerializerRegistry — declaring a capability without a matching
     // DelegatedPredicateSerializer makes the marking layer pick Lucene as viable for
     // operators it can't actually translate, and the failure surfaces at convert time as
-    // an IllegalStateException ("No Lucene serializer for [..]"). Today only EQUALS has
-    // a serializer; range ops, NOT_EQUALS, IS_NULL, IS_NOT_NULL, IN, LIKE are deferred
-    // until their serializers land.
+    // an IllegalStateException ("No Lucene serializer for [..]").
+    //
+    // Serializers for NOT_EQUALS, LIKE, IN, IS_NULL, IS_NOT_NULL, the four range comparisons,
+    // and REGEXP are wired up in QuerySerializerRegistry so the convertor path is ready, but
+    // they are intentionally NOT advertised here yet because two upstream gaps make
+    // unconditional delegation regress:
+    //   1. The planner has no shard-level cost model — declaring a backend viable for an
+    //      operator picks Lucene whenever it can answer the predicate, even when DataFusion
+    //      would scan the parquet column faster (e.g. LIKE '%substr%' on a 30 M-cardinality
+    //      URL field).
+    //   2. The partial→final aggregate split for {@code approx_distinct} fails with
+    //      "expected Binary state" when filter delegation rearranges the operator chain on
+    //      {@code dc(...)} queries.
+    // When either lands, flip the advertisement on per-operator (or gate it behind an
+    // {@code analytics.delegation.lucene_filter_predicates} cluster setting).
+    //
     // TODO: have CapabilityRegistry intersect declared FilterCapability against the
     // backend's serializer keyset at startup so this list can't drift again. The TODO in
     // OpenSearchFilterRule.resolveViableBackends references the same constraint.
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneSubtreeConvertor.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneSubtreeConvertor.java
@@ -10,6 +10,8 @@ package org.opensearch.be.lucene;
 
 import org.apache.calcite.rex.RexCall;
 import org.apache.calcite.rex.RexNode;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import org.opensearch.analytics.planner.rel.AnnotatedPredicate;
 import org.opensearch.analytics.spi.DelegatedPredicateSerializer;
 import org.opensearch.analytics.spi.DelegatedSubtreeConvertor;
@@ -31,6 +33,8 @@ import java.util.Map;
  * @opensearch.internal
  */
 final class LuceneSubtreeConvertor implements DelegatedSubtreeConvertor {
+
+    private static final Logger LOGGER = LogManager.getLogger(LuceneSubtreeConvertor.class);
 
     private final Map<ScalarFunction, DelegatedPredicateSerializer> leafSerializers;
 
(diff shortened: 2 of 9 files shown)


alchemist51:
I have seen for the text fields, if we have a != operator and we do the != '' check for example in q25. I saw wrong results being returned. For text field the term query with "" end up selecting all of the docs due to the analyser

alchemist51:
Can we also take correctness into account in the serializer change?

PR #22048, opened 2026-06-08
Fix min_target_partitions overriding configured target_partitions

aravindsagar:
### Description
The `datafusion.min_target_partitions` setting was being used as `configured.max(setting)`, which bumped target_partitions UP when the setting exceeded the configured value derived from `search.concurrent.max_slice_count`. The setting is meant to be a floor for adaptive reduction under memory pressure — a no-op when the caller already requests fewer partitions.

Clamp the floor to the configured value per call so it can only stop reduction below the floor and never raise the starting value. Add regression tests for both directions: no-op when configured <= floor, still acts as a reduction floor when configured > floor.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/query_budget.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/query_budget.rs
@@ -75,8 +75,11 @@ use datafusion::common::DataFusionError;
 use datafusion::execution::memory_pool::{MemoryConsumer, MemoryLimit, MemoryPool, MemoryReservation};
 use parquet::file::metadata::ParquetMetaData;
 
-/// Configurable minimum target_partitions floor. Updated from Java when the
-/// cluster setting `datafusion.min_target_partitions` changes.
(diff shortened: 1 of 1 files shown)


aravindsagar:
> [**Logic Error**](https://github.com/opensearch-project/OpenSearch/pull/22048/files#diff-b3f1956383d79315af7b7b733f9e44a49bd384bd3e8c8d92523d04d1a7914719R215-R216)
> Line 215 computes `min_partitions` but it is never used in the subsequent logic. The adaptive reduction code that should respect this floor is missing from the visible diff. Without using `min_partitions` to clamp reductions, the floor constraint is not enforced, making the fix incomplete.
> 
> ```rust
> let min_partitions = get_min_target_partitions().min(configured_target_partitions.max(1));
> let mut target_partitions = configured_target_partitions.max(1);
> ```

This is not true. `min_partitions` is used later on in lines 252 and 308 for setting `target_partitions`

alchemist51 (approved):
LGTM!

Bukhtawar (approved):
thanks lgtm

PR #22049, opened 2026-06-08
Fix support for != in search command and Date Nanos support

expani:
### First Problem
PPL `search source=idx field!="v"` (and any `_exists_:field` clause) 500s on a composite parquet-primary / lucene-secondary index:

FieldExistsQuery requires that the field indexes doc values, norms or vectors, but field 'log_level' exists and indexes neither of these data structures

The frontend sends `!=` into a `query_string` carrying `_exists_:field`. `_exists_` for Search command and compiles to a Lucene `FieldExistsQuery`, which needs doc_values/norms/vectors on the segment it runs against. On a composite index the keyword doc_values live in the **parquet primary**, while the query executes on the **lucene-secondary** segment (postings only) — so `FieldExistsQuery.rewrite()` throws and the query fails.

### Second Problem

OpenSearchSchemaBuilder.buildLeafType mapped date/date_nanos to TIMESTAMP with no precision (defaults to 0 → millis downstream), but parquet reads date_nanos as nanos → coordinator reduce RowConverter throws Timestamp(ms) got Timestamp(ns) on multi-shard sort. Fix: set precision explicitly — date→3, date_nanos→9 via a switch that throws on unknown TIMESTAMP types. Nanos preserved end-to-end

### Fix
Rewrite `FieldExistsQuery` → a postings-only `TermRangeQuery(field, null, null)` ("field has ≥1 indexed term") before execution. This is semantically equivalent to "field exists" for the postings-indexed keyword/text/match_only_text types that filter delegation supports.

New helper `LuceneQueryConversionUtils.rewriteFieldExistsForSecondary(Query)` recurses through the Boolean / ConstantScore / Boost / DisjunctionMax wrappers the `query_string` parser emits, rebuilds only changed branches, and fails fast (via `QueryVisitor`) on any unhandled container still hiding a `FieldExistsQuery`.

Applied at **both** Lucene execution entry points:
  - `LuceneFilterDelegationHandle` — filter-delegation path (DataFusion-driven scans).
  - `LuceneScanInstructionHandler` — Lucene-driver scan/count path (`LuceneSearchExecEngine`).

### Tests
  - **Unit** (`LuceneQueryConversionUtilsTests`): rewrite across each wrapper, nested wrappers, unchanged-query reference-equality, and detection.
  - **Integration** (`SearchFieldExistsDelegationIT`, `app_logs_filter_delegation`, 2 shards):
    `search log_level!="INFO"` count + value (no INFO) assertions, a value-returning `| fields`
    variant (scan path), plus `NOT` / `OR` / `=` controls. Reverting the rewrite reproduces the 500.

### Follow-up
Interim fix. Long-term, the lucene-secondary should read doc_values from the parquet primary, after which `FieldExistsQuery` resolves natively and this rewrite can be removed.

### DIFF
--- a/sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/OpenSearchSchemaBuilder.java
+++ b/sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/OpenSearchSchemaBuilder.java
@@ -253,14 +253,22 @@ public class OpenSearchSchemaBuilder {
         if (sqlType == null) {
             return null;
         }
-        // date / date_nanos both map to TIMESTAMP, but their sub-second precision differs and must
-        // be carried on the Calcite type (fractional-seconds digits): date → millis (3), date_nanos
-        // → nanos (9). Without an explicit precision, createSqlType(TIMESTAMP) defaults to 0, which
-        // downstream lowers to Timestamp(Second)→Millisecond; the parquet read of a date_nanos field
-        // then produces Timestamp(Nanosecond), and the reduce-stage RowConverter rejects the mismatch.
+        // TIMESTAMP must carry sub-second precision: date → millis (3), date_nanos → nanos (9).
+        // Without it the type defaults to precision 0 (→ millis downstream) and clashes with the
+        // parquet read's Timestamp(Nanosecond) for date_nanos. Switch (not default) so an unforeseen
+        // type mapping to TIMESTAMP fails loudly rather than silently inheriting 3.
         RelDataType base;
         if (sqlType == SqlTypeName.TIMESTAMP) {
-            int precision = "date_nanos".equals(opensearchType) ? 9 : 3;
+            int precision = switch (opensearchType) {
+                case "date" -> 3;
+                case "date_nanos" -> 9;
+                default -> throw new IllegalStateException(
+                    "OpenSearch type '"
+                        + opensearchType
+                        + "' maps to TIMESTAMP but has no declared sub-second "
+                        + "precision; add a case in buildLeafType"
+                );
+            };
             base = typeFactory.createSqlType(sqlType, precision);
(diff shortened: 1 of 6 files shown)


sandbox/libs/analytics-api/src/main/java/org/opensearch/analytics/schema/OpenSearchSchemaBuilder.java
```diff
@@ -253,7 +253,19 @@ public static RelDataType buildLeafType(String opensearchType, RelDataTypeFactor
...
             return null;
         }
-        return typeFactory.createTypeWithNullability(typeFactory.createSqlType(sqlType), true);
+        // date / date_nanos both map to TIMESTAMP, but their sub-second precision differs and must
+        // be carried on the Calcite type (fractional-seconds digits): date → millis (3), date_nanos
+        // → nanos (9). Without an explicit precision, createSqlType(TIMESTAMP) defaults to 0, which
+        // downstream lowers to Timestamp(Second)→Millisecond; the parquet read of a date_nanos field
+        // then produces Timestamp(Nanosecond), and the reduce-stage RowConverter rejects the mismatch.
+        RelDataType base;
+        if (sqlType == SqlTypeName.TIMESTAMP) {
```
mch2, line 261: @vinaykpud i think you are doing something similar here in https://github.com/opensearch-project/OpenSearch/pull/22045/changes - FYI.

vinaykpud: No, this is different, it actually keeps the precision intact with the type instead of defaulting, But once this PR lands i can rebase and check

vinaykpud (approved):

mch2 (approved):

PR #22041, opened 2026-06-07
[Sandbox] Propagate index.sort.field to DataFusion ListingTable

alchemist51:
### Description

Wires the index sort settings (`index.sort.field`, `index.sort.order`) end-to-end into DataFusion's `ListingOptions::with_file_sort_order(...)`. The OpenSearch parquet writer enforces this sort within each segment, so the per-file claim is verifiable-by-construction. Once advertised on `DataSourceExec`, downstream optimizers can fire:

- `output_ordering` propagates upward, letting `SortPreservingMergeExec` replace `SortExec(TopK)` when the query's ORDER BY matches.
- `sort_prefix` enables partial-sort early termination on TopK.
- `split_file_groups_by_statistics=true` (set when sort fields are present) keeps file boundaries intact for the bin-packer; combined with `target_partitions ≥ num_files`, the optimizer produces single-file groups with `output_ordering` validated by DataFusion's per-file min/max chain check.

**Java side.** `ReaderManagerConfig` now carries `IndexSettings`. `DataFusionPlugin.createReaderManager` reads `INDEX_SORT_FIELD_SETTING` / `INDEX_SORT_ORDER_SETTING` and passes them through `DatafusionReaderManager` → `DatafusionReader` → `ReaderHandle` → `NativeBridge.createDatafusionReader` as parallel `String[]` arrays plus a count. `SortOrder` enum is converted to its lowercase string form (`asc` / `desc`) for the wire.

**Rust side.** `df_create_reader` decodes the new FFM args. `ShardView` carries `pub sort_fields: Vec<String>` / `pub sort_orders: Vec<String>`. The session-context build site builds a `Vec<SortExpr>` and calls `with_file_sort_order(...)` on the `ListingOptions`. Uses `Expr::Column(Column::from_name(name.clone()))` rather than `col(&str)` — the latter goes through SQL identifier normalization and silently lowercases column names, producing an empty ordering at plan time when the Arrow schema is case-sensitive (e.g. `EventTime` vs `eventtime`). Sets `split_file_groups_by_statistics = true` and `with_target_partitions(effective_partitions)` on the `ListingOptions` when sort fields are present.

### Verification

End-to-end on a sorted ClickBench segment (`index.sort.field=[EventTime] asc`, single-segment shard) running

```
source = clickbench_mini | where SearchPhrase != '' | sort EventTime | head 10 | fields SearchPhrase
```

DataFusion physical plan from `RustLoggerBridge` DEBUG:

```
[...]

PR #22056, opened 2026-06-08
Reverse indexed-parquet segment iteration for opposite-direction sorts

alchemist51:
### Description

Mirror of Lucene's `ContextIndexSearcher.shouldUseTimeSeriesDescSortOptimization` for the **indexed-parquet** scan path (`IndexedTableProvider` / `QueryShardExec`). When `index.sort.field` is configured and the query's leading `ORDER BY` runs counter to the catalog's stored direction, reverse the per-shard segment iteration so a `TopK` above the scan pulls the highest-priority segment first and parquet page-stats pruning kicks in for the rest.

**Reuses PR #22041's plumbing.** That PR threads `index.sort.field` / `index.sort.order` from `IndexSettings` through `DatafusionReaderManager` → `DatafusionReader` → `ReaderHandle` → `df_create_reader` to `ShardView.sort_fields` / `ShardView.sort_orders`. This PR adds a second consumer of those same fields, on the indexed path, with **no new Java or FFM surface area**.

### Where the optimization fires

`indexed_executor.rs::execute_indexed_with_context_inner`, after `from_substrait_plan` decodes the `LogicalPlan`:

```rust
let mut segments = segments;
if should_reverse_segments(analyze_top_sort(&logical_plan).as_ref(), &sort_fields, &sort_orders) {
    reverse_segment_iteration_order(&mut segments);
}
```

Three module-local helpers:

- **`analyze_top_sort(&LogicalPlan) -> Option<TopSort>`** — walks `Projection` / `Limit` / `SubqueryAlias` / `Distinct` / `Filter` to find the first `LogicalPlan::Sort`. Returns the leading sort key's bare column name and direction. Returns `None` if no top-level Sort, or if the leading sort key is anything but a plain `Expr::Column` (function on the sort key breaks catalog monotonicity).
- **`should_reverse_segments(top_sort, sort_fields, sort_orders) -> bool`** — true iff the catalog has a sort, the query's leading sort column matches `sort_fields[0]` exactly (case-sensitive, matching #22041's `Column::from_name`), and the directions differ. Multi-key sorts use only `sort_fields[0]` — segments aren't monotonic on a non-leading key.
- **`reverse_segment_iteration_order(&mut [SegmentFileInfo])`** — flips the slice in place. Per-segment `global_base` is **deliberately not** recomputed; see below.

### QTF correctness — why `global_base` must be preserved

The `__row_id__` emitted by the indexed scan is `segment.global_base + position`. QTF's fetch phase (`api::fetch_by_row_ids`) goes through a **different code path** (`ShardTableProvider`, not `IndexedTableProvider`) and rebuilds segments via `build_segments(ShardView.object_metas, ShardView.writer_generations)` — always in catalog order. Fetch then `partition_point`s the cached row IDs against catalog-order `global_base`s to find the owning segment.

For the round trip to hold, query-phase IDs must match catalog-order bases. `reverse_segment_iteration_order` therefore flips iteration order **only** — each segment retains its catalog-order `global_base`. Reversal becomes purely a scheduling decision; the row-id space is unchanged. The optimization composes safely with QTF.

### What this PR does not change

- **Vanilla path** (`ListingTable` / `DataSourceExec`) — covered by #22041 via `ListingOptions::with_file_sort_order(...)` (different mechanism: advertise `output_ordering` so the planner can drop `SortExec`).
- **`QueryShardExec.properties.equivalence_properties`** — no output-ordering claim. The TopK / SortPreservingMergeExec above the scan still re-sorts. Reversal is purely a pruning aid for sort-with-limit shapes; it doesn't try to make the scan's output-ordering claim load-bearing. That's a separate, more invasive change.

### Tests

**Rust unit tests** in `indexed_executor::tests`:
- `analyze_top_sort_*` — top-level desc, descends through Limit/Projection, None for no-sort, None for function sort key.
- `should_reverse_segments_*` — matching field opposite direction, matching same direction, catalog desc query asc, no query sort, no catalog sort, non-leading catalog field, case-sensitive name.
- `reverse_segments_preserves_global_base`, `reverse_segments_empty_is_noop`, `reverse_segments_single_keeps_its_base`.

**Rust e2e tests** in `indexed_table/tests_e2e/sort_reverse_row_id.rs` — pin the load-bearing invariant directly: build 3- and 4-segment fixtures with `emit_row_ids: true`, run the query in natural order vs `reverse_segment_iteration_order`, assert the **set** of emitted `__row_id__` values is identical (sorted set equality). If `global_base` cumulation drifted under reversal, fetch would return wrong rows; this catches it locally.

**Java IT** in `sandbox/qa/analytics-engine-coordinator/.../SortReverseQtfIT.java` (parquet primary + Lucene secondary, 2 shards, 3 flushes → 2-4 parquet segments per shard, `index.sort.field=ts asc`):
- `testReversalTriggers_orderByTsDescLimit_correctResults` — captures the reversal log line via `MockLogAppender` on `RustLoggerBridge` (proves the optimization fired) and asserts top-5 rows by `ts DESC` are correct via the QTF query→fetch round-trip.
- `testNoReversal_orderByTsAscLimit_correctResults` — `UnseenEventExpectation` confirms reversal does NOT fire when query direction matches catalog; results still correct.
- `testReversal_keywordFilterDelegationPath_correctResults` — `WHERE category = 'C0' ORDER BY ts DESC LIMIT 4` exercises the keyword-filter delegation path (predicate shipped to Lucene perf peer) + reversal; asserts QTF row-id round-trip composes with delegation.

### Verification

- `cargo test --lib`: 1045 / 1045 pass (1043 pre-existing + 22 new unit tests + 2 new e2e tests, after dedupe).
- `:server`, `:sandbox:plugins:composite-engine`, `:sandbox:plugins:analytics-backend-datafusion`, `:sandbox:plugins:analytics-backend-lucene`, `:sandbox:qa:analytics-engine-coordinator` all `compileJava` / `compileTestJava` / `compileInternalClusterTestJava` clean.

### Benchmark results — full clickbench corpus

Measured on EC2 `r8g.2xlarge` (ARM64, 8 vCPU, 64 GB) with the full 100M-row OSB clickbench corpus (parquet primary + Lucene secondary, OSB sort `[CounterID desc, EventDate desc, UserID desc, EventTime desc, WatchID desc]`).

Same EC2 box, same data, same script, same 13 queries (10 reversal-eligible + 3 controls), 3 passes per query. Build A = upstream main without this PR; Build B = upstream main + this PR. **min_ms** is reported (warm cache, ignores pass-1 cold-cache outliers; mean is misleading because Q01 cold-cache is ~2300 ms before page cache warms).

All queries: `where <field> = <literal> | sort + CounterID | head 1000` (catalog leading is `CounterID desc` → `+` triggers reversal). Selectivity is the number of rows matching the filter against the 100M corpus.

| Query | Filter (selectivity) | Build A (no fix) | **Build B (with fix)** | Δ ms | Δ % |
|---|---|---:|---:|---:|---:|
| Q01 | `BrowserLanguage = 'S0'` (~92M) | 478 | **388** | −90 | **−18.8%** |
| Q02 | `HitColor = '5'` (~80M) | 402 | **339** | −63 | **−15.7%** |
| Q03 | `BrowserCountry = 'h1'` (~28M) | 324 | **210** | −114 | **−35.2%** |
| Q04 | `MobilePhoneModel = 'iPad'` (~5M) | 328 | **186** | −142 | **−43.3%** |
| Q07 | `UTMSource = 'feedburner'` (~572K) | 67 | 67 | 0 | flat |
| Q08 | `MobilePhoneModel = 'iPhone'` (~229K) | 89 | 88 | −1 | flat |
| Q10 | compound (~thousands) | 73 | 75 | +2 | flat |
| **C1** | same direction as catalog (`sort - CounterID`) | 243 | 242 | −1 | flat ✓ |
| **C2** | non-leading catalog key (`sort + EventDate`) | 234 | 233 | −1 | flat ✓ |

(Q05/Q06/Q09 returned 0 rows on this corpus slice — literal absent from the data — and are omitted; their flat timings are uninformative.)

**Reading the result:**

- **High-selectivity queries (Q01–Q04, 5M–92M matching rows): 16% → 43% latency reduction.** These are the cases where `head 1000` against tens of millions of matching rows actually has work to do — TopK has to scan many segments before its threshold tightens. With reversal, the highest-`CounterID` segment is hit first, the threshold tightens fast, and parquet page stats prune most of the remaining segments.
- **Low-selectivity queries (Q07–Q10, ≤600K matching rows): flat.** Finding the top 1000 is cheap regardless of scheduling order — work is dominated by the keyword-filter scan, not segment iteration. The optimization correctly does no harm where it can't help.
- **Control queries (C1 same direction, C2 non-leading sort key): identical timings between builds.** Negative confirmation that the change is neutral on shapes where reversal must not fire.

Variance across warm passes is ~3–5 ms, well below the 60–140 ms deltas on Q01–Q04.

### Related Issues / PRs

Builds on #22041 (vanilla-path index.sort.field propagation).

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
@@ -346,9 +346,14 @@ pub struct ShardView {
     /// Index sort fields, in priority order. Sourced from the index's
     /// `index.sort.field` setting on the Java side. Empty when the index has
     /// no `index.sort.field` configured. Parallel to `sort_orders`.
-    /// Used to build `ListingOptions.with_file_sort_order(...)` so DataFusion
-    /// advertises `output_ordering` from the scan and enables the
-    /// `sort_prefix` optimization on TopK / SortPreservingMerge.
+    ///
+    /// Two consumers today:
+    ///   - Vanilla path: `ListingOptions.with_file_sort_order(...)` so DataFusion advertises
+    ///     `output_ordering` from the scan and the `sort_prefix` optimization fires on
+    ///     TopK / SortPreservingMerge.
+    ///   - Indexed path (`indexed_executor`): when the query's leading ORDER BY runs counter
+    ///     to catalog-snapshot order, the per-shard segment iteration is reversed so a TopK
+    ///     above us can prune via parquet page statistics.
     pub sort_fields: Vec<String>,
     /// Index sort directions per field — values: `"asc"` or `"desc"`.
     /// Parallel to `sort_fields`. Sourced from `index.sort.order`.
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
@@ -201,6 +201,101 @@ pub async fn execute_indexed_query(
 
 // ── Helpers ───────────────────────────────────────────────────────────
 
+/// Result of walking a logical plan looking for the leading top-of-plan ORDER BY.
+///
+/// `column` is the bare column name (no qualifier — we compare against `index.sort.field`
+/// which is also unqualified). `descending` is `true` for `ORDER BY x DESC`.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct TopSort {
+    pub column: String,
+    pub descending: bool,
+}
+
+/// Walk through the top of a logical plan to find the leading sort expression.
+///
+/// Descends through `Projection`, `Limit`, `SubqueryAlias`, `Distinct`, and `Filter` —
+/// nodes that don't reorder rows or rewrite sort key columns. On reaching
+/// `LogicalPlan::Sort`, returns the leading sort key's bare column name and direction.
+/// Returns `None` if the plan has no top-level Sort, or if the first Sort key is not
+/// a plain `Expr::Column` (e.g. `ORDER BY lower(x)` — we can't claim catalog monotonicity
+/// after a function).
+pub(crate) fn analyze_top_sort(plan: &datafusion::logical_expr::LogicalPlan) -> Option<TopSort> {
+    use datafusion::logical_expr::{Expr, LogicalPlan};
+    let mut current = plan;
+    loop {
+        match current {
+            LogicalPlan::Sort(s) => {
+                let leading = s.expr.first()?;
+                let col = match &leading.expr {
+                    Expr::Column(c) => c.name.clone(),
+                    _ => return None,
+                };
+                return Some(TopSort {
+                    column: col,
+                    descending: !leading.asc,
+                });
+            }
+            LogicalPlan::Projection(p) => current = p.input.as_ref(),
+            LogicalPlan::Limit(l) => current = l.input.as_ref(),
+            LogicalPlan::SubqueryAlias(a) => current = a.input.as_ref(),
+            LogicalPlan::Distinct(d) => match d {
+                datafusion::logical_expr::Distinct::All(input) => current = input.as_ref(),
+                datafusion::logical_expr::Distinct::On(on) => current = on.input.as_ref(),
+            },
+            LogicalPlan::Filter(f) => current = f.input.as_ref(),
+            _ => return None,
+        }
+    }
+}
+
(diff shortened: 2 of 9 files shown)


bharath-techie (approved):
LGTM

PR #22062, opened 2026-06-08
Surface date-only and time-only fields as SQL DATE / TIME on the analytics-engine route

ahkcs:
### Description

OpenSearch stores every `date` field as an epoch-millisecond timestamp regardless of its mapping `format`. On the analytics-engine route, `OpenSearchSchemaBuilder.mapFieldType` typed every `date`/`date_nanos` column as `TIMESTAMP` and discarded the `format`, so a **date-only** field (e.g. `basic_date`, `yyyy-MM-dd`) surfaced as `1984-04-12 00:00:00` and a **time-only** field (e.g. `hour`, `HH:mm:ss`) as `1970-01-01 09:00:00`, instead of the SQL `DATE` (`1984-04-12`) / `TIME` (`09:00:00`) the v2 / Calcite path produces.

This closes both gaps (DATE + TIME).

### Approach

- Add `DateType` and `TimeType` Calcite UDTs, each backed by `SqlTypeName.TIMESTAMP` and carrying the same precision a plain timestamp does — mirroring the existing `IpType` / `BinaryType` UDTs. Because the backing type and precision are identical to a plain timestamp, **storage, Substrait serialization, and the DataFusion scan are unchanged**: the column is still `Timestamp(MILLISECOND)` on disk and binds against the parquet column exactly as before. Only the planner-side logical type changes, so the SQL plugin can render it as `DATE` / `TIME`.
- Add `DateFormatClassifier`, which narrows a `date`/`date_nanos` field to `DATE` or `TIME` by its `format` (named formats, custom patterns like `yyyy-MM-dd` / `HH:mm:ss`, and `||`-combined alternatives). It mirrors `OpenSearchDateType.getExprTypeFromFormatString` in the SQL plugin.
- `OpenSearchSchemaBuilder.buildLeafType` becomes format-aware and returns `DateType` / `TimeType` accordingly; datetime/epoch formats keep `TIMESTAMP`.

> Companion SQL-plugin change (consumes the UDTs for the response schema label and renders the value as a `LocalDate` / `LocalTime`) is in opensearch-project/sql and depends on this change being published, exactly as the `IpType`/`BinaryType` consumers do.

### Failure this fixes (cluster log, before)

A naive UDT without an explicit precision hit:
```
This feature is not implemented: Unsupported Substrait precision -1, for PrecisionTimestamp
```
The fix carries the plain-timestamp precision so the type serializes to an identical Substrait `PrecisionTimestamp`.

### Testing

- New unit tests: `DateFormatClassifierTests` (named date/time/datetime formats, custom patterns, `'T'`-literal stripping, combined alternatives, null/blank) — all pass.
- Verified end-to-end against a local 9-plugin analytics cluster with the companion SQL change, `-Dtests.analytics.parquet_indices=true`:

| SQL-plugin IT (analytics route) | before | after |
|---|---|---|
| `CalcitePPLAggregationIT.testCountByDateTypeSpanForDifferentFormats` | ❌ | ✅ |
| `CalcitePPLAggregationIT.testCountByDateTypeSpanWithDifferentUnits` | ❌ | ✅ |
| `CalcitePPLAggregationIT.testCountByTimeTypeSpanForDifferentFormats` | ❌ | ✅ |
| `CalcitePPLAggregationIT.testCountByTimeTypeSpanWithDifferentUnits` | ❌ | ✅ |
| `CalcitePPLAggregationIT.testCountByNullableTimeSpan` | ❌ | ✅ |
| `CalcitePPLAggregationIT.testCountBySpanForCustomFormats` | ❌ | ✅ |
| `CalciteAnalyticsDatetimeWireFormatIT` (date + time cols) | ❌ | ✅ |

No regressions: the change only alters columns whose mapping format is date-only or time-only; datetime/epoch columns are untouched. Other pre-existing analytics-route failures (percentile-approx, count-distinct-approx, nested fields, `date_add`) are date/time-independent.

- `./gradlew :sandbox:libs:analytics-api:{spotlessJavaCheck,forbiddenApisMain,licenseHeaders}` pass.

### Check List


ahkcs:
Closing as superseded.

The date-only / time-only DATE/TIME surfacing this PR adds has already landed in `main` via #22045 ("[analytics-engine] PPL date/time/timestamp fixes", merged 2026-06-09, 47a44bf3db4). There the UDTs were renamed `DateType` → `DateOnlyType` and `TimeType` → `TimeOnlyType`, with `DateFormatClassifier` and the format-aware `OpenSearchSchemaBuilder.buildLeafType(type, format, factory)` in place. Rebasing this branch would only re-add duplicate `DateType` / `TimeType` classes alongside the merged `*OnlyType` ones (which is what the last `Merge branch 'main'` here turned red on).

Verified against current `main` on the analytics-engine (parquet-primary) route: `CalcitePPLAggregationIT` and `CalcitePPLAggregationPaginatingIT` both pass all 15 date/time-span tests (90/100 overall; the 10 remaining failures are pre-existing and unrelated — approx-percentile, approx-count-distinct, and nested-field aggregations).

SQL-plugin consumer side is covered by the merged opensearch-project/sql#5521.

PR #22064, opened 2026-06-08
Fix OpenSearchTimeoutException to return HTTP 504 instead of 500

### LINKED ISSUE #22060: [BUG] OpenSearchTimeoutException returns HTTP 500 instead of 504
### Describe the bug

`OpenSearchTimeoutException` doesn't override `status()`, so it falls through to `OpenSearchException`'s default of `INTERNAL_SERVER_ERROR` (500). It should return `GATEWAY_TIMEOUT` (504) instead — when any of the ~12 callers in the codebase throw it across a REST boundary, the client sees a 500, which looks like a server bug when the actual condition is a timeout. Retry logic that backs off on 500 won't retry the request even though it would likely succeed.

Affected call sites include:
- `RestGetMappingAction.java:130`
- `RestGetAliasesAction.java:254` (added in #21954)
- `TaskManager.java:572` (`_tasks/_wait_for_completion`)
- `TransportAddVotingConfigExclusionsAction.java:195`
- `TransportClearVotingConfigExclusionsAction.java:149`
- `ListenerTimeouts.java:72`
- `DecommissionService.java:284`, `DecommissionController.java:141`
- `PeerRecoveryTargetService.java:523`
- `Node.java:2110`
- `PlainTransportFuture.java:83`, `FutureUtils.java:96`

### Suggested fix

One-line override on `OpenSearchTimeoutException`:
```java
@Override
public RestStatus status() {
    return RestStatus.GATEWAY_TIMEOUT;  // 504
}
```

This is consistent with how other timeout-shaped exceptions already work (`ProcessClusterEventTimeoutException.java:59`, `ClusterManagerNotDiscoveredException.java:62`, `NoNodeAvailableException.java:62` all override `status()` to return `SERVICE_UNAVAILABLE`).

### Compatibility notes

- Wire format unaffected — `status()` is computed by the receiving JVM, not serialized.
- The only `instanceof OpenSearchTimeoutException` check in-tree (`AsyncShardFetchCache.java:225`) is for retryability, unaffected by status.
- Tests asserting `status == 500` for a timeout-triggered response would need updating — worth grepping before landing.

gingeekrishna:
## Description

`OpenSearchTimeoutException` was not overriding `status()`, so all ~12 call sites that throw it across a REST boundary returned HTTP 500 (`INTERNAL_SERVER_ERROR`) to the client. This makes timeouts indistinguishable from server bugs and breaks retry logic that backs off on 500 but should retry on 504.

## Changes

- Override `status()` in `OpenSearchTimeoutException` to return `RestStatus.GATEWAY_TIMEOUT` (504)
- Add `OpenSearchTimeoutExceptionTests` asserting the new status

## Consistency

This aligns with how other timeout-shaped exceptions already behave:
- `ProcessClusterEventTimeoutException` → 503
- `ClusterManagerNotDiscoveredException` → 503

504 is more precise for `OpenSearchTimeoutException` since it represents a gateway-level timeout (no response received within the allowed window), not general service unavailability.

## Compatibility

- Wire format is unaffected — `status()` is computed by the receiving JVM, not serialized
- No existing tests assert HTTP 500 for a timeout-triggered response (verified by grep)
- The only `instanceof OpenSearchTimeoutException` check in-tree (`AsyncShardFetchCache.java:225`) is for retryability logic, unaffected by status

Fixes #22060

## Test plan
- [ ] `OpenSearchTimeoutExceptionTests` — asserts `GATEWAY_TIMEOUT` for message and cause constructors


dzane17 (approved):

gingeekrishna:
Hi @dzane17 

Could you help me to merge this PR?

dzane17:
@jainankitk @andrross Can you help review?

cwperks (approved):

server/src/test/java/org/opensearch/OpenSearchTimeoutExceptionTests.java
```diff
@@ -0,0 +1,25 @@
...
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch;
+
+import org.opensearch.core.rest.RestStatus;
+import org.opensearch.test.OpenSearchTestCase;
+
+public class OpenSearchTimeoutExceptionTests extends OpenSearchTestCase {
```
cwperks, line 14: I don't think these unit test add a great deal of value, is it possible to do another test for this (an integ test) to show an actual timeout returning 504?

gingeekrishna:
Good call — added a true end-to-end test: `TasksWaitForTimeoutIT` in `:modules:transport-netty4` (that module is where the real, non-mock HTTP transport lives for `internalClusterTest`).

It registers a minimal blocking transport action that holds a task open indefinitely, then issues a real HTTP request through the low-level REST client:

```
GET /_tasks/<id>?wait_for_completion=true&timeout=100ms
```

and asserts `response.getStatusLine().getStatusCode() == 504`.

Ran it locally and confirmed the full path:

```
OpenSearchTimeoutException[Timed out waiting for completion of [org.opensearch.tasks.Task@...]]
	at org.opensearch.tasks.TaskManager.waitForTaskCompletion(TaskManager.java:572)
	at org.opensearch.action.admin.cluster.node.tasks.get.TransportGetTaskAction$1.doRun(TransportGetTaskAction.java:171)
	...

testGetTaskWaitForCompletionReturnsGatewayTimeoutOverHttp PASSED
```

— a real timeout fires server-side and the client genuinely receives a 504, not just a unit-level check of `status()` in isolation. Kept the existing unit tests too since they're cheap and pin down the exact mapping per constructor.

cwperks:
@gingeekrishna I don’t think this needs a Netty-module IT. We already have `TasksIT#testGetTaskWaitForTimeout` / `testListTasksWaitForTimeout` covering the task wait timeout path using the existing `TestTaskPlugin`. The status change itself is generic REST exception mapping, so a smaller `BytesRestResponseTests` assertion for new OpenSearchTimeoutException(...) producing status 504 would cover the REST conversion without tying the test to Netty or duplicating a blocking action plugin.

so maybe an IT is too heavyweight here and let's instead add a test in `BytesRestResponseTests`

gingeekrishna:
@cwperks Good point, that's a much better fit. Removed \`TasksWaitForTimeoutIT\` and added \`BytesRestResponseTests#testResponseWhenTimeoutException\`, mirroring the existing \`testResponseWhenInternalServerError\` pattern in that file:

\`\`\`java
public void testResponseWhenTimeoutException() throws IOException {
    final RestRequest request = new FakeRestRequest();
    final RestChannel channel = new DetailedExceptionRestChannel(request);
    final BytesRestResponse response = new BytesRestResponse(channel, new OpenSearchTimeoutException("simulated timeout"));
    assertEquals(RestStatus.GATEWAY_TIMEOUT, response.status());
    assertNotNull(response.content());
    final String content = response.content().utf8ToString();
    assertThat(content, containsString("\"reason\":\"simulated timeout\""));
    assertThat(content, containsString("\"status\":" + 504));
}
\`\`\`

This exercises the actual \`BytesRestResponse\` REST conversion rather than just \`OpenSearchTimeoutException#status()\` in isolation, without tying the test to Netty or a custom blocking action plugin. Kept the existing simple \`OpenSearchTimeoutExceptionTests\` too since it's cheap and still useful as a baseline.

cwperks:
@gingeekrishna latest changes look good. TY for addressing!

PR #22065, opened 2026-06-08
[Analytics-Engine] Fix multi-shard aggregate split for non-prefix group keys

mch2:
### Description
stats <agg> by span(field, N) returned wrong, un-coalesced buckets on multi-shard indices (e.g. avg(price) by span(amount,5) gave 11 buckets across 2 shards vs 7 on one) — affecting every aggregate except count().

Cause: the per-shard PARTIAL stage moves group keys to the front (Calcite contract), but the coordinator FINAL stage was built with the original groupSet. For a non-prefix groupSet like {1} (agg arg projected before the span key), FINAL grouped on the partial-state column instead of the key. count() worked only because its groupSet was already {0}.

Fix: build FINAL's groupSet as range(groupCount) to match the keys-fronted PARTIAL output (no-op for the common prefix case). Added aggregate-by-span multi-shard tests with hand-verified goldens.

### Related Issues

### Check List


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchAggregateSplitRule.java
```diff
@@ -162,20 +162,29 @@ public void onMatch(RelOptRuleCall call) {
         List<IntermediateField> intermediateFields = FinalAggCallBuilder.classify(aggregate.getAggCallList());
 
         // Build FINAL's aggCalls against gathered's row type so typeMatchesInferred passes.
+        int groupCount = aggregate.getGroupSet().cardinality();
```
expani: We can ensure that `OpenSearchAggregate` contains this while creation from `LogicalAggregate` then the change here is not required.

mch2: updated

expani (approved):

sandeshkr419 (approved):

PR #22055, opened 2026-06-08
Dynamic filter support for indexed path

alchemist51:
## Summary

Adds **runtime dynamic-filter pushdown** to the indexed-parquet scan in `analytics-backend-datafusion`.

### What dynamic filters do

When a query has an ordering + LIMIT (e.g. `ORDER BY ts DESC LIMIT N`), DataFusion's `SortExec`-TopK pushes a `DynamicFilterPhysicalExpr` down into `QueryShardExec`. The filter starts as the trivial `true` placeholder and is **tightened at runtime** as the TopK heap fills (e.g. `ts > <current heap-min>`). Per row group, the scan checks the parquet statistics against the latest snapshot of that filter and skips the entire row group if it can prove no row in it can reach the top-k. The filter only references the SORT columns, so RG min/max stats span a strict superset of WHERE-passing candidates — pruning is **conservative** (we may miss a skip, never drop a qualifying row).

### What this PR adds

- `dynamic_filter.rs` — `DynamicRgPruner` that watches the dynamic filter's snapshot generation, builds a `PruningPredicate` from the latest snapshot, and decides per-RG whether stats prove the RG cannot satisfy the predicate.
- `dynamic_filter_probe.rs` — Helper that probes whether the planner-pushed filter is in fact a `DynamicFilterPhysicalExpr` and extracts it for the pruner.
- Wires the pruner into `indexed_table/stream.rs` at two points: (a) at prefetch-time when scheduling a row group, and (b) as a poll-phase backstop just before parquet decode (since the filter may have tightened further during the ~1-RG prefetch lead).
- New metric `dynamic_filter_rg_pruned_at_poll` to count pruned RGs at the poll-phase backstop.
- New `IndexedTableProvider` integration to accept the pushed filter via DataFusion's physical filter-pushdown API.
- New cluster setting `datafusion.indexed.dynamic_filter_pushdown` (default `true`, `NodeScope` + `Dynamic`) — turn off to A/B the perf impact; when off, the scan declines the pushdown and behaves exactly as before.
- Threads the new flag through `DatafusionSettings` and `WireConfigSnapshot` to the Rust side.
- End-to-end Rust tests in `tests_e2e/dynamic_filter_pushdown.rs`.

[...]

PR #22066, opened 2026-06-08
[Analytics-Engine] Pre-compute peer bitmap per segment for delegation

rishabhmaurya:
## Summary

- Pre-computes the performance-delegated peer bitmap once per segment (shared across all partitions via `Arc<OnceLock>`) instead of creating a new Lucene scorer per row-group
- Eliminates repeated `weight.scorer(leaf)` calls that caused O(row_groups) × O(scorer_build_time) cost for WildcardQuery delegation

## Motivation

When `LIKE '%substr%'` is delegated to Lucene, the `WildcardQuery` scorer construction walks the entire term FST (~3s per call on a 30M-term field). With 7 row-groups per shard, this meant 7 × 3s = 21s wasted on repeated scorer builds — making delegation 2.4× slower than the scan-only path.

## Results (ClickBench q23, 100M docs)

| Configuration | Before | After |
|---|---|---|
| Delegation OFF (DataFusion scan) | 12s | 12s |
| Delegation ON (LIKE + !=) | **29s** | **11.5s** |
| `index_query_time` per shard | 21.6s | 2.2s |

Delegation now slightly outperforms the scan-only path (Lucene prunes 100M → 7K rows, saving parquet decode time).

## Design

See: [design doc](https://github.com/rishabhmaurya/OpenSearch/blob/mustang-peer-bitmap-pr/docs/peer-bitmap-precompute.md) (will link separately)

On first access (when any partition's RG passes the selectivity gate), compute the full-segment bitmap in one shot — one scorer + one forward pass. Store in a `RoaringBitmap` shared via `OnceLock`. Subsequent partitions slice the pre-computed bitmap (microseconds, no FFM call).

## Test plan

- [x] `single_collector::tests` — 5 passed
- [x] `tests_e2e::multi_segment` — 16 passed
- [x] `tests_e2e::page_pruning` — 17 passed
- [x] `tests_e2e::fuzz` (including `fuzz_delegation_passthrough` and `fuzz_delegation_sloppy`) — 48 passed
- [x] Deployed to 100M-doc ClickBench cluster, verified q23 delegation ON: 29s → 11.5s


bharath-techie:
Aren't we materialising the complete bitset for the segment to gain perf advantage ? One of the reasons we choose RG by RG is to avoid materialising bitsets beyond a single RG.

sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_table/eval/single_collector.rs
```diff
@@ -163,14 +163,17 @@ pub struct SingleCollectorEvaluator {
...
     /// a slot do so once per (query × annotation_id) — not per chunk.
     performance_provider_locks: Arc<HashMap<i32, Arc<OnceLock<ProviderHandle>>>>,
-    /// Writer generation identifying the segment this evaluator was bound to at
-    /// factory time. Captured so `prefetch_rg` can build a per-call
-    /// `FfmSegmentCollector` lazily without re-deriving the segment from
-    /// `RowGroupInfo` (which doesn't carry it).
+    /// Pre-computed full-segment peer bitmap cache. Shared across all partitions of
+    /// the same segment via Arc<OnceLock>. Keyed by annotation_id. Lazily populated
+    /// on first access (when any partition's RG passes the selectivity gate).
+    peer_bitmap_cache: Arc<HashMap<i32, Arc<OnceLock<Option<RoaringBitmap>>>>>,
```
bharath-techie, line 169: 1. Avoiding materializing complete bitsets is one of the main reasons why we built RG by RG pipeline - and with dynamic filters , in runtime we'll be able to prune RGs and exit early, so we never need to collect/score all the docs . https://github.com/opensearch-project/OpenSearch/pull/22055

2. And similarly prefetch/candidates phase will skip collecting docs per RG if we are able to prune based on stats.

3. For single partition and majority of the queries where create scorer is cheap, this will not help. Existing patterns are memory and pruning friendly.

Can we rather evaluate based on cardinality of terms etc ?

rishabhmaurya: @bharath-techie thanks for reviewing. Its still WIP so I will keep you posted with the progress. Feel free to share ideas -  I believe current way of creating scorer per RG isn't great to begin with and we shouldn't be building this logic in a way which works for one type of query and causing significant (>2x in this case) regression for others.

bharath-techie: We don't create scorer per RG right? We create scorer per partition.

bharath-techie: So essentially for 1 partition, there wouldn't be any perf difference at all. Difference comes with multiple partition, the same might happen for lucene concurrent segment too right? Correct me if I'm wrong. 

I think we can try offloading this from hot path, we do have a prefetch flow that happens in spawn blocking tokio pool which doesn't affect the hot path except for the first segment and slow scorer creation queries such as LIKE. So creating parellel scorers for successive files might help but again it's a tradeoff with compute.

PR #22069, opened 2026-06-08
Return 400 for deeply nested queries exceeding protobuf recursion limit

finnegancarroll:
### Description

Queries with 31+ nested function calls (e.g. `ceil(floor(ceil(...)))`) cause the prost protobuf decoder on the data node to hit its recursion limit when decoding the Substrait plan. Previously this surfaced as an opaque 500 `Stage 0 failed` error. Two fixes added:

- (`AnalyticsSearchService`): detects `recursion limit reached` in the native exception cause chain and throws `IllegalArgumentException` (HTTP 400) with a user-friendly message before the error crosses the gRPC boundary.

- (`NativeErrorConverter`): adds pattern for `Query too deeply nested` so the coordinator-side `convertException()` maps these to `IllegalArgumentException` (400).

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSearcher.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DatafusionSearcher.java
@@ -73,6 +73,13 @@ public class DatafusionSearcher implements EngineSearcher<DatafusionContext> {
         try {
             streamPtr = future.join();
         } catch (Exception exception) {
+            if (containsRecursionLimit(exception)) {
+                throw new IllegalArgumentException(
+                    "Query too deeply nested: the expression exceeds the maximum nesting depth supported by the execution engine. "
+                        + "Simplify the query by reducing nested function calls.",
+                    exception
+                );
+            }
             throw new IOException("Query execution with session context failed", exception);
         }
         // NativeBridge#executeWithContextAsync has already marked the handle consumed (which
@@ -130,5 +137,18 @@ public class DatafusionSearcher implements EngineSearcher<DatafusionContext> {
     public void close() {
         // ReaderHandle lifecycle is owned by DatafusionReader / EngineReaderManager,
         // not by the searcher. Do not close it here.
+    }
+
+    // TODO: Handle recursion limit on the Rust side (e.g. increase prost decode limit or
+    // return a structured error code) so Java doesn't need to pattern-match on error messages.
+    private static boolean containsRecursionLimit(Throwable t) {
+        for (int depth = 0; t != null && depth < NativeErrorConverter.MAX_CAUSE_CHAIN_DEPTH; depth++) {
+            String msg = t.getMessage();
+            if (msg != null && msg.contains("recursion limit reached")) {
+                return true;
+            }
+            t = t.getCause();
+        }
+        return false;
     }
 }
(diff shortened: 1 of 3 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -141,10 +141,29 @@ private ResolvedExecution executeFragmentStreamingResolved(
...
+                    "Query too deeply nested: the expression exceeds the maximum nesting depth supported by the execution engine. "
+                        + "Simplify the query by reducing nested function calls.",
+                    e
+                );
+            }
             throw new RuntimeException("Failed to start streaming fragment on " + shard.shardId(), e);
         }
     }
 
+    private static boolean isQueryTooComplex(Exception e) {
```
mch2: lets add a TODO here to return a better exception rust side, this is brittle.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -141,10 +141,29 @@ private ResolvedExecution executeFragmentStreamingResolved(
             throw e;
         } catch (Exception e) {
             listener.onFragmentFailure(resolved.queryId, resolved.stageId, resolved.shardIdStr, e);
```
mch2, line 143: we should return the same IE to the listener

finnegancarroll: I'm not sure i understand, in both case here the exception is propagated to `onFragmentFailure` listener before re-throwing.

finnegancarroll:
Two flaky sandbox check failures related to correctness:
```
REPRODUCE WITH: ./gradlew ':sandbox:qa:analytics-engine-rest:integTest' --tests 'org.opensearch.analytics.qa.TwoShardScalarIT.testReduceCorrectnessAcrossTwoShards' -Dtests.seed=C6B028C74C8DA470 -Dtests.security.manager=false -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=fr-BJ -Dtests.timezone=America/Argentina/La_Rioja -Druntime.java=25

TwoShardScalarIT > testReduceCorrectnessAcrossTwoShards FAILED

    java.lang.AssertionError: 1 of 30 2-shard reduce checks failed:
Suite: Test class org.opensearch.analytics.qa.TwoShardScalarIT
      - scalar/sc_str_to_date [1-shard vs golden] row 0 col 1: Value mismatch - expected 2024-01-01T00:00:00, got 2024-01-01 00:00:00
```

```
REPRODUCE WITH: ./gradlew ':sandbox:qa:analytics-engine-rest:integTest' --tests 'org.opensearch.analytics.qa.TwoShardScalarIT.testReduceCorrectnessAcrossTwoShards' -Dtests.seed=C6B028C74C8DA470 -Dtests.security.manager=false -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=fr-BJ -Dtests.timezone=America/Argentina/La_Rioja -Druntime.java=25
  2> java.lang.AssertionError: 1 of 30 2-shard reduce checks failed:
      - scalar/sc_str_to_date [1-shard vs golden] row 0 col 1: Value mismatch - expected 2024-01-01T00:00:00, got 2024-01-01 00:00:00
        at __randomizedtesting.SeedInfo.seed([C6B028C74C8DA470:5F791D1F186B48AB]:0)
        at org.opensearch.analytics.qa.TwoShardReduceTestCase.testReduceCorrectnessAcrossTwoShards(TwoShardReduceTestCase.java:132)
```

mch2 (approved):

PR #22074, opened 2026-06-09
_cat/nodes API shows negative values in cpu stats

### LINKED ISSUE #19120: [BUG] (v3.2.0) `_cat/nodes` API shows negative values in cpu stats
### Describe the bug

I ran into the below issue in 3.2.0 version that cpu usages could be negative and the same value (-1)

```
$ curl "https://dev-opensearch-test-250821-os.osp01.abc.com/_cat/nodes?v" -ku admin:test

ip             heap.percent ram.percent cpu load_1m load_5m load_15m node.role node.roles      cluster_manager name
10.251.175.203           38          60  -1    0.44    0.41     0.49 m         cluster_manager *               dev-opensearch-test-250821-master-1
10.251.117.144           20          60  -1    0.58    0.54     0.60 m         cluster_manager -               dev-opensearch-test-250821-master-0
10.251.225.190           67          90  -1    1.79    1.14     1.03 di        data,ingest     -               dev-opensearch-test-250821-data-0
10.251.143.196           24          60  -1    0.40    0.68     0.61 m         cluster_manager -               dev-opensearch-test-250821-master-2
```

### Related component

Clients

### To Reproduce

1. deploy opensearch cluster using 3.2.0 version
2. After your cluster runs successfully, `curl "https://dev-opensearch-test-250821-os.osp01.abc.com/_cat/nodes?v" -ku admin:test`
3. It also happens wrongly when you try `_cluster/stats` API

### Expected behavior

.

### Additional Details

**Plugins**
I just added the prometheus-exporter plugin by building new Docker image
```
FROM harbor.abc.com/library/opensearchproject/opensearch:3.2.0

COPY prometheus-exporter-3.2.0.0.zip /tmp/
RUN ./bin/opensearch-plugin install file:/tmp/prometheus-exporter-3.2.0.0.zip
```

 - OS: All hosts are worker nodes of kubernetes v1.31.4 and based on RHEL 8.10 (4.18.0 kernel)
 - Version: OpenSearch 3.2.0

**Additional context**
- Air-gapped environment but I believe it doesn't matter with this issue.

reta:
### Description
The way OpenSearch fetches CPU stats (for CPU load in particular), is using `com.sun.management.internal.OperatingSystemImpl::getCpuLoad`. As per official documentation,
the `OperatingSystemImpl::getCpuLoad` may return -1 if perf counters are disabled or container stats are not available fe (or any other reason), this is documented https://docs.oracle.com/en/java/javase/21/docs/api/jdk.management/com/sun/management/OperatingSystemMXBean.html#getCpuLoad()

To mitigate and detect the issue early, we decided to:
 - on bootstrap, do a check to report if CPU load is not available (but do not fail the bootstrap procedure)
 - if exception occurred while `com.sun.management.internal.OperatingSystemImpl::getCpuLoad` is called, log it only once (ideally, it should happen during bootstrap only)

Also, we should document that on some systems JVM may not report getCpuLoad() correctly, so -1 will be returned y the `/_cat/nodes`.

### Related Issues
Closes https://github.com/opensearch-project/OpenSearch/issues/19120

### Check List


server/src/main/java/org/opensearch/monitor/Probes.java
```diff
@@ -49,6 +55,12 @@ public static short getLoadAndScaleToPercent(Method method, OperatingSystemMXBea
                     return (short) (load * 100);
                 }
             } catch (Exception e) {
+                // Only log an exception once, ideally it should happen during the bootstrap check only
+                // (see please BootstrapChecks.ProbesCheck)
+                if (shouldLogException == true) {
```
reta, line 60: Using simple volatile here, if there is a race and we log exception more than once, non issue

cwperks (approved):

PR #22079, opened 2026-06-09
[analytics-engine] Don't drop user head/limit over grouped aggregates

mch2:
### Description
Gate OpenSearchSortRule's limit-drop on the input being provably <=1 row (Calcite getMaxRowCount), so a no-op limit over a scalar aggregate is dropped but a user `stats ... by k | head N` is preserved. Add CoreRules.LIMIT_MERGE so a user head and the system size-cap collapse to one Sort.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortRule.java
@@ -10,15 +10,10 @@ package org.opensearch.analytics.planner.rules;
 
 import org.apache.calcite.plan.RelOptRule;
 import org.apache.calcite.plan.RelOptRuleCall;
-import org.apache.calcite.rel.RelFieldCollation;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.core.Sort;
-import org.apache.calcite.rex.RexInputRef;
-import org.apache.calcite.rex.RexNode;
 import org.opensearch.analytics.planner.PlannerContext;
 import org.opensearch.analytics.planner.RelNodeUtils;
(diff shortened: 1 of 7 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchSortRule.java
```diff
@@ -58,30 +53,13 @@ public void onMatch(RelOptRuleCall call) {
...
         }
 
-        // Drop a pure-Fetch Sort over an Aggregate (possibly via a Project chain). Aggregate
-        // output cardinality is bounded by group count, so a JOIN_SUBSEARCH_MAXOUT-style 50000
-        // limit is structurally never hit. The Sort/Fetch shape here also triggers a DataFusion
-        // hang for composite-group keys. Removing it keeps the safety contract intact for
-        // unbounded children but avoids the redundant Fetch when the aggregate already bounds
-        // output.
-        if (sort.getCollation().getFieldCollations().isEmpty() && sort.offset == null && hasAggregateUnderProjects(child)) {
+        // A collation-less, offset-less LIMIT over a provably single-row input (e.g. a scalar
```
expani: We should not be dropping anything here given it's purely for Marking/Annotating the operators.

Should we make the SortPushdownRule handle this instead ?

mch2: we need this in hep, split and pushdown rewriter run during cbo and post cbo.  Will add a new rule.

mch2: nevermind, CoreRules can do this

expani (approved):

sandeshkr419 (approved):

PR #22082, opened 2026-06-09
Projection pushdown below Exchange + QTF adaptation

expani:
### What
Push `Project` below the Exchange so DataFusion reads only referenced columns from parquet on each shard, instead of scanning all ~105 columns and projecting at the coordinator.

### Why
PPL queries like `source=idx | where x | fields body` were reading every column off parquet and shipping them across the gather. Wasteful I/O and network.

### Changes
  1. **`PlannerImpl.pushdownRules`** — remove `SORT_PROJECT_TRANSPOSE`. Keeps the SELECT `Project` below the Sort/Exchange so DataFusion's `OptimizeProjections` prunes the scan.
  2. **`OpenSearchExchangeReducer.computeSelfCost`** — scale cost by output column count (`× getRowType().getFieldCount()`). Volcano now prefers placing the Exchange where fewer columns cross it.
  3. **`OpenSearchLateMaterializationRewriter`** — adapt QTF, which previously assumed `SORT_PROJECT_TRANSPOSE` lifted the Project above the Sort:
     - Targeted detection gate: decline only on a derived sort key or a window function below the anchor (was: any expression).
     - Relocate the SELECT projection above the late-mat wrapper, refs rebased scan→wrapper space (aliases + expressions preserved).
     - Collapse the resulting adjacent Projects into one.

  ### Verified
  - DF physical plan reads only referenced columns (QTF declines) / narrowed sort-key scan + post-fetch projection (QTF fires) - confirmed on a live 2-shard node.
  - LateMaterialization suite green, incl. new tests for alias preservation, mixed-direction multi-key sort, and expressions where an operand is also the sort key.
  - No regression in Aggregate / Window / Join / Union / Subquery plan-shape suites.

  ### Notes for reviewers
  The QTF change is a **consequence** of removing `SORT_PROJECT_TRANSPOSE`, not a feature — it preserves QTF behavior (same fire set, same scan narrowing, same deferred-fetch set), not improves it.


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -286,21 +288,13 @@ private static RelNode pushdownRules(RelNode input, RuleProfilingListener listen
...
             .addRuleCollection(
                 List.of(
                     CoreRules.FILTER_PROJECT_TRANSPOSE,
                     CoreRules.FILTER_AGGREGATE_TRANSPOSE,
                     CoreRules.FILTER_INTO_JOIN,
-                    CoreRules.SORT_PROJECT_TRANSPOSE,
-                    CoreRules.PROJECT_MERGE,
-                    CoreRules.LIMIT_MERGE,
-                    CoreRules.SORT_REMOVE_REDUNDANT
+                    // FIXME [TempChangeForExperiment] CoreRules.SORT_PROJECT_TRANSPOSE,
```
alchemist51: Maybe unintended?

alchemist51 (changes requested):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
expani: Removed

sandeshkr419 (approved):

sandeshkr419:
@expani I think the sandbox errors are related to this change: https://github.com/opensearch-project/OpenSearch/actions/runs/27251790133/job/80483842313?pr=22082

Can you please check once if any tests need modifications.

PR #22083, opened 2026-06-09
Add periodic jemalloc arena purge for native memory reclamation

himshikhagupta:
### Description
 Adds a background purge mechanism that reclaims jemalloc-retained physical memory (RSS) after query execution completes. Without this fix, jemalloc holds dirty/muzzy pages indefinitely because its decay timer doesn't tick without allocation activity.

  - Adds `native_jemalloc_purge()` FFI function (Rust) that calls `mallctl("arena.N.purge")` on all arenas
  - Adds `NativeArenaPurger` — a dedicated daemon thread that periodically checks resident memory and purges if above threshold
  - Exposes `purge_count` in `_nodes/stats` native memory response
  - Configurable via dynamic cluster settings:
    - `native.jemalloc.purge_interval` (default: 5s) — check frequency
    - `native.jemalloc.purge_threshold_percent` (default: 85%) — percentage of `node.native_memory.limit`

### Related Issues

### Check List

### DIFF
--- a/sandbox/libs/dataformat-native/rust/common/src/allocator.rs
+++ b/sandbox/libs/dataformat-native/rust/common/src/allocator.rs
@@ -13,6 +13,7 @@
 //!   - `< 0`  → error pointer. Negate and pass to `native_error_message` / `native_error_free`.
 
 use crate::error::{ffm_wrap, into_error_ptr};
+use crate::log_info;
 use std::sync::OnceLock;
 use tikv_jemalloc_ctl::{epoch, epoch_mib, stats, stats::allocated_mib, stats::resident_mib};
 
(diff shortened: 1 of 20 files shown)


himshikhagupta:
> ❌ Gradle check result for [c0bedf2](https://github.com/opensearch-project/OpenSearch/commit/c0bedf2134be434629116bba370bb6fc0ec885ec): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/80188/)
> 

flaky test https://github.com/opensearch-project/OpenSearch/issues/15822

himshikhagupta:
> ❌ Gradle check result for [093549f](https://github.com/opensearch-project/OpenSearch/commit/093549fb9df3b1417efa01760070795a3d679a7a): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/80238/)
> 

Flaky test https://github.com/opensearch-project/OpenSearch/issues/16145

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/NativeArenaPurger.java
```diff
@@ -0,0 +1,147 @@
...
+            lookup.find("native_jemalloc_purge").orElseThrow(),
+            FunctionDescriptor.of(ValueLayout.JAVA_LONG)
+        );
+        purgeScheduler = Executors.newSingleThreadScheduledExecutor(r -> {
+            Thread t = new Thread(r, "opensearch[jemalloc_purge]");
+            t.setDaemon(true);
+            return t;
+        });
+        reschedule();
+    }
```
Bukhtawar, line 79: Why can we run this directly on the Rust side?

himshikhagupta: The major reason to keep the scheduler in Java was to reuse existing threadpool constructs.

Bukhtawar (commented):
Should we have a way to differentiate active arena purge vs idle to not compete against a running query?

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
```diff
@@ -111,6 +111,157 @@ fn set_all_arenas(suffix: &[u8], ms: i64) -> Result<i64, String> {
...
+use std::time::Duration;
+
+static PURGE_THRESHOLD_BYTES: AtomicI64 = AtomicI64::new(i64::MAX);
+static PURGE_INTERVAL_MS: AtomicU64 = AtomicU64::new(5000);
+static PURGE_COUNT: AtomicU64 = AtomicU64::new(0);
+static PURGE_THREAD_STARTED: std::sync::Once = std::sync::Once::new();
+static PURGE_THREAD: OnceLock<std::thread::Thread> = OnceLock::new();
+
+/// Starts the background purge thread. Idempotent — only the first call spawns the thread.
+fn start_purge_thread() {
```
Bukhtawar: When does this loop terminate, do we have a graceful shutdown?

himshikhagupta:
> Should we have a way to differentiate active arena purge vs idle to not compete against a running query?

purge only reclaims pages already freed back to jemalloc — actively allocated query buffers are never touched. The only risk is between queries: if purge fires after query A frees but before query B allocates, B pays page-fault cost instead of reusing dirty pages. This is bounded by the 85% threshold gate and 5s interval, so only when resident is high purge will be triggered

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
himshikhagupta: Didn't add explicit shutdown since the thread holds no external resources and dies with the process. Added now for completeness, called from NativeBridgeModule.close() during node shutdown.

sandbox/modules/native-bridge/src/main/java/org/opensearch/nativebridge/NativeBridgeModule.java
```diff
@@ -64,6 +70,35 @@ public class NativeBridgeModule extends Plugin {
...
+        TimeValue.ZERO,
+        Setting.Property.NodeScope,
+        Setting.Property.Dynamic
+    );
+
+    /**
+     * Purge fires only when jemalloc resident exceeds this percentage of
+     * {@code node.native_memory.limit}. Default: 85%.
+     */
+    public static final Setting<Integer> JEMALLOC_PURGE_THRESHOLD_PERCENT = Setting.intSetting(
```
gaurav-amz, line 89: Why do we need this setting? Purge should be fired when the resident - allocated > 0? Because the trigger is 85% of the total limit, node that finishes the query and goes idle at say 60–65% never reaches the threshold and will never purges? Is my understanding correct here?

sandbox/libs/dataformat-native/rust/common/src/allocator.rs
```diff
@@ -111,6 +112,133 @@ fn set_all_arenas(suffix: &[u8], ms: i64) -> Result<i64, String> {
...
+        }
+        std::thread::park_timeout(Duration::from_millis(interval_ms));
+
+        let threshold = PURGE_THRESHOLD_BYTES.load(Ordering::Relaxed);
+        let resident = match refresh_stats() {
+            Ok((_, res)) => res,
+            Err(_) => continue,
+        };
+
+        if resident <= threshold {
```
gaurav-amz, line 159: The decision is based on total resident memory, but a purge can only give back pages that were already freed (dirty/muzzy) — i.e. `resident − allocated`.

himshikhagupta: If allocated is close to resident, which means there are actually no dirty pages, then purge will have little to no impact.

sandbox/modules/native-bridge/src/main/java/org/opensearch/nativebridge/NativeBridgeModule.java
himshikhagupta: Default jemalloc decay should take care of active purging. background thread is for supporting scenarios not covered like when there are no active requests in the cluster to trigger jemalloc decay check. In those cases, we only want to purge if we see high resident, as a protection against possible issues.

Bukhtawar (approved):

PR #22085, opened 2026-06-09
[analytics-engine] Wipe spill directory contents instead of the directory itself

gaurav-amz:
## Summary

`create_global_runtime` previously called `fs::remove_dir_all(spill_dir)`, which `rmdir`s the spill root as its final step. That requires write permission on the spill directory's parent. When `spill_dir` is a pre-existing mount point and the JVM user does not own its parent, boot fails with `EACCES`:

```
org.opensearch.bootstrap.StartupException: java.lang.RuntimeException:
  Invalid or Unsupported Configuration: Failed to clear leaked spill files in
  /mnt/analytics/backend/spill (io kind=PermissionDenied):
  Permission denied (os error 13).
  Verify the OpenSearch process owns the spill directory and the volume is
  mounted read-write before restarting.

  at org.opensearch.be.datafusion.DataFusionService.doStart(DataFusionService.java:101)
  at org.opensearch.be.datafusion.nativelib.NativeBridge.createGlobalRuntime(NativeBridge.java:695)
```

## Fix

Wipe the directory's *contents* — every immediate child — but leave the spill root in place:

- Subdirectories: `fs::remove_dir_all(child)` (recursive, as before)
- Files and symlinks at top level: `fs::remove_file(child)` (the symlink itself is unlinked; its target is never touched)
- The spill root: untouched

DataFusion's `create_local_dirs` sees the existing root, skips `std::fs::create_dir(root)`, and provisions a fresh `datafusion-XXXXXX/` via `tempdir_in(root)`. All operations now require write only on the spill directory itself, which the JVM owns.

The contract change: `spill_dir` is treated as a pre-existing mount point. Callers wanting auto-creation must ensure the JVM has write on the parent. The rustdoc on `create_global_runtime` is updated accordingly.

## Test plan

**Rust unit tests** (all in `sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs`)

- [x] `create_global_runtime_succeeds_when_jvm_does_not_own_spill_parent` (new) — chmods the parent tempdir to `0o555` to reproduce the mount-point topology, seeds leaked entries (datafusion subtree + loose file), asserts boot succeeds, all leaked entries are removed, and the spill root is preserved. Pre-fix code would fail at the final `rmdir`.
- [x] `create_global_runtime_unlinks_top_level_symlink_without_following` (new) — places a symlink in the spill dir pointing at a file outside; asserts the symlink is unlinked but its target is untouched. Pins the symlink-defense behavior.
- [x] `create_global_runtime_clears_leaked_spill_files_recursively` (updated) — assertion now reflects that the spill root is preserved by cleanup (previously asserted DataFusion recreated it after `remove_dir_all`).
- [x] `create_global_runtime_fails_fast_when_cleanup_fails` (updated) — error-message assertion accepts either of the two new failure paths (`enumerate spill directory` or `clear leaked spill entry`).
- [x] Existing tests `create_global_runtime_with_empty_spill_dir_disables_disk_manager`, `create_global_runtime_with_spill_dir_enables_disk_manager`, and `create_global_runtime_with_empty_spill_dir_does_not_touch_filesystem` all unchanged and pass.

`cargo test --lib api::tests::create_global_runtime` — 7/7 passing.

**Integration test**

- [x] `SpillCleanupOnBootIT` (new, in `sandbox/qa/analytics-engine-rest/`) — `integTestSpillCleanup` Gradle task pre-seeds a `datafusion-XXXX/tmp_spill_001.arrow` subtree and a loose `stray.txt` in the spill directory at config-phase before cluster boot. The test then asserts: (1) cluster booted healthy (200 from spill stats endpoint), (2) seeded entries are gone, (3) spill root still exists, (4) DataFusion provisioned at least one fresh `datafusion-XXXXXX/` whose name does not collide with the seeded leaked name. Excluded from default `integTest`.

**Manual reproduction on the failing cluster**

- Verified diagnosis on a test cluster.


sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -410,24 +411,53 @@ pub fn create_global_runtime(
...
+                } else {
+                    fs::remove_dir_all(&path)
+                };
+                if let Err(e) = result {
+                    let msg = format!(
+                        "Failed to clear leaked spill entry {} (io kind={:?}): {}. \
+                         Verify the process owns the spill directory before restarting.",
+                        path.display(), e.kind(), e
+                    );
+                    log::error!("{}", msg);
```
Bukhtawar, line 492: Do we know how long this would take to clean up maybe a 50G disk with many small files?

bharath-techie (approved):
LGTM

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
gaurav-amz: I tested it with 30 GB of data, and it was taking around 130 seconds. I'll make the deletion asynchronous/non-blocking for stale entries.

sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
```diff
@@ -410,24 +430,143 @@ pub fn create_global_runtime(
...
+            // AND any *.stale leftovers from prior boots whose cleanup didn't
+            // finish. Errors inside the thread are logged only — the cluster
+            // has not yet joined, but stragglers are recoverable next boot.
+            //
+            // Spawn failure itself is rare (EAGAIN/ENOMEM under extreme system
+            // pressure) and not boot-fatal: the disk is already in a valid
+            // state for DataFusion to start; the *.stale entries just persist
+            // until a future boot with a healthy thread library reaps them.
+            // Failing boot here would be more disruptive than the wasted disk.
+            let spill_path_for_gc = spill_path.clone();
```
Bukhtawar, line 508: What happens if there is a failure between the two phases, would operations be idempotent?

gaurav-amz: Then next startup will cleanup the stale data, there won't be any `.stale.stale` entries. Only `.stale` files will be there which will be cleanup in next startup.

Let me know if you concerns is something different.

Bukhtawar (approved):

PR #22094, opened 2026-06-10
fix(analysis): Add path boundary validation in resolveAnalyzerPath

nagarajg17:
### Description
Validate that the resolved path remains within configDir() after normalization. Previously, the method only normalized the path without checking boundaries, allowing relative paths to resolve outside the intended directory.

Also add tests verifying both valid paths and rejected traversal attempts.

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/analysis/Analysis.java
+++ b/server/src/main/java/org/opensearch/index/analysis/Analysis.java
@@ -343,6 +343,9 @@ public class Analysis {
     }
 
     public static Path resolveAnalyzerPath(Environment env, String wordListPath) {
+        if (isUnderConfig(env, wordListPath) == false) {
+            throw new IllegalArgumentException("Resource path must be inside config directory: " + wordListPath);
+        }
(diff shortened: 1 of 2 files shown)


cwperks (approved):
Thank you @nagarajg17! I didn't know about this `isUnderConfig` method and we used a different approach for Hunspell Dictionaries here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java#L161-L190

We should probably adapt that other code to fit this convention.

nagarajg17:
> Thank you @nagarajg17! I didn't know about this `isUnderConfig` method and we used a different approach for Hunspell Dictionaries here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/analysis/HunspellTokenFilterFactory.java#L161-L190
> 
> We should probably adapt that other code to fit this convention.

Yeah Hunspell checks looks complex and can be simplified

PR #22095, opened 2026-06-10
Fix QTF stitcher precision mismatch on date_nanos fields

vinaykpud:
### Description

`ArrowCalciteTypes.toArrow()` hardcoded `TimeUnit.MILLISECOND` for every Calcite `TIMESTAMP`. Shards emit `Timestamp(NANOSECOND)` for `date_nanos`, but the late-materialization Stitcher pre-allocated a `TimestampMilliVector` for the output. Arrow's `copyFromSafe` requires identical fixed-width subclasses, so it threw an empty `IllegalArgumentException` — surfacing to clients as HTTP 400 "Invalid Query".

**Fix:** branch on Calcite precision — `precision == 9 → NANOSECOND`, else `MILLISECOND`.

### Query shapes this fixes

Triggered on a multi-shard parquet-backed index with a `date_nanos` field when the plan includes:
- projection + sort on the `date_nanos` column
- `head N` (or other top-K)
- at least one fetch-only field above the anchor

Example PPL:
```
source = idx | where match(body, 'failed') and service = 'checkout'
             | sort - ts | fields ts, severity, body | head 4
```

Drop any of those and the LM rewriter declines QTF or the bug doesn't fire.

### Tests

- `ArrowCalciteTypesTests` — 3 new cases covering precision 3, 9, and default.
- `LateMaterializationDateNanosIT` — new 2-shard parquet+lucene IT running the q01-style plan. Pre-fix: HTTP 400. Post-fix: 4 rows in DESC order with sub-millisecond precision preserved.

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/ArrowCalciteTypes.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/ArrowCalciteTypes.java
@@ -22,12 +22,6 @@ import org.apache.calcite.rel.type.RelDataType;
  * schema. The {@code AggregateFunction.ArrowToCalciteTypeMapper} (in the SPI module) handles
  * the inverse direction for {@code IntermediateField} resolution; this class is kept as the
  * single authority for the Calcite→Arrow direction needed outside that resolver.
- *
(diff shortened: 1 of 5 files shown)


mch2 (approved):

vinaykpud:
https://github.com/opensearch-project/sql/pull/5498

PR #22097, opened 2026-06-10
Fix NPE for search-only indices without primaries

HUSTERGS:
### Description
This PR fixes `NullPointerException` failures when operating on search-only indices whose routing tables do not contain primary shards.

After an index is scaled down to search-only mode, its routing table can contain only search-only shard copies. Some management and routing paths still assume that `primaryShard()` is always non-null, which can cause NPEs when those paths are used on a search-only index.

This change adds explicit handling for routing tables without primary shards:
  - Cluster allocation explain now returns a clear `IllegalArgumentException` when a primary shard is requested but no primary exists in the routing table, instead of dereferencing `null`.
  - Bulk adaptive shard selection now tolerates shard routing tables without primaries.
    - The single-shard path returns `null` when no primary exists.
    - The multi-shard path skips shard routing tables without active primaries.
  - Index block verification now treats a missing primary shard the same as an unassigned primary, avoiding an NPE when adding blocks to search-only indices.

### Related Issues

as said above

### Check List


prudhvigodithi:
Thanks @HUSTERGS, may I know how was this failure know when using any API or during enabling and disabling search only mode?

HUSTERGS:
> Thanks @HUSTERGS, may I know how was this failure know when using any API or during enabling and disabling search only mode?

@prudhvigodithi Thanks for taking a look.

The issue happens after search-only mode is enabled. The enable/disable operation itself is not the only concern here; the broader issue is that once an index is in search-only mode, its routing table may not contain a primary shard. Some code paths still assume `primaryShard()` is always non-null.

The concrete API path where I first reproduced the NPE was adding an index block after enabling search-only mode:

  1. Create an index with search replicas.
  2. Enable search-only mode.
  3. Call the add index block API on that index.

That path dereferences `shardRoutingTable.primaryShard().unassigned()`, which fails when `primaryShard()` returns `null`.

During the investigation, I also found the same no-primary routing table assumption in allocation explain and bulk adaptive shard selection. The added tests cover these paths separately so the fix is not limited to the add-index-block API; it covers the shared condition introduced by search-only routing tables without primaries.

prudhvigodithi:
> 3\. Call the add index block API on that index.

Got it, today search only mode adds a `search_only` block https://docs.opensearch.org/latest/api-reference/index-apis/blocks/, so calling index block API on top this is where the error was seen.

prudhvigodithi:
@HUSTERGS LGTM, can you check the Codecov CI?

HUSTERGS:
> @HUSTERGS LGTM, can you check the Codecov CI?

@prudhvigodithi  I've added some tests, waiting for the Codecov CI result : )

prudhvigodithi (approved):

PR #22102, opened 2026-06-10
[analytics-datafusion] Apply aggregate mode in indexed executor, remove dc() delegation block gate

sandeshkr419:
## Summary

The indexed executor now applies `aggregate_mode` stripping (Mode::Partial) after building its physical plan, producing Binary HLL state instead of Int64 for engine-native-merge aggregates (dc/APPROX_COUNT_DISTINCT). This fixes the schema mismatch at the reduce sink without needing to suppress filter delegation at marking time.

- **indexed_executor.rs**: Apply `apply_aggregate_mode` after physical plan build when `aggregate_mode != Mode::Default`
- **substrait_to_tree.rs**: `delegation_possible` UDF body returns first arg (transparent pass-through) instead of panicking — allows prepared_plan FilterExec to evaluate normally
- **OpenSearchFilterRule.java**: Remove engine-native-merge gate (no longer needed)
- **PlannerContext/PlannerImpl**: Remove dead `hasEngineNativeMergeAggregate` flag and `containsEngineNativeMergeAggregate` method
- **FilterDelegationExtendedIT**: Expand with stddev_pop/samp, percentile, additional dc() shapes (29 exact + 6 double-tolerance cases on 2 shards)

## Context

When filter predicates are dual-viable (can execute on both DataFusion and Lucene), they get wrapped in `delegation_possible(...)` in the substrait, which routes execution to the indexed executor. Previously, the indexed executor always re-decoded substrait with Mode::Default — producing Int64 (final aggregate output) instead of Binary HLL intermediate state. The coordinator expected Binary (from `derive_schema_from_partial_plan`), causing a type mismatch.

 The prior workaround suppressed dual-viable marking for dc() queries entirely (the "gate"), preventing filter delegation from benefiting dc() queries. This PR removes that gate by teaching the indexed executor to respect aggregate mode.

## Testing

- `FilterDelegationExtendedIT` — 35 cases: included more count/sum/avg/min/max/dc/percentile/stddev_pop/stddev_samp/fields (2-shard dataset)
- `PplClickBenchIT` — all 42 queries pass (including Q23: LIKE + dc + correctness delegation)
- `FilterDelegationGoldenIT` — all 38 routing shapes pass

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
@@ -500,6 +500,7 @@ async unsafe fn execute_indexed_with_context_inner(
 
     let query_config = Arc::new(handle.query_config);
     let num_partitions = query_config.target_partitions.max(1);
+    let aggregate_mode = handle.aggregate_mode;
     let ctx = handle.ctx;
     let table_name = handle.table_name;
     let table_path = handle.table_path;
@@ -880,6 +881,13 @@ async unsafe fn execute_indexed_with_context_inner(
     // types. The target is schema_coerce::coerce_inferred_schema(physical_schema) — same
     // narrowing the partition-stream registration uses, so consumer-side StreamingTable
     // and producer-side batches agree by construction (see crate::relabel_exec).
+    // Apply aggregate mode stripping when prepare_partial_plan was called (engine-native-merge).
+    // This makes the indexed executor produce Binary HLL state (Partial) instead of Int64 (Final).
+    let physical_plan = if aggregate_mode != crate::agg_mode::Mode::Default {
+        crate::agg_mode::apply_aggregate_mode(physical_plan, aggregate_mode)?
+    } else {
+        physical_plan
+    };
     let target_schema = crate::schema_coerce::coerce_inferred_schema(physical_plan.schema());
     let physical_plan = crate::relabel_exec::wrap_if_relabel_needed(physical_plan, target_schema)?;
     log_debug!("DataFusion physical plan:\n{}", displayable(physical_plan.as_ref()).indent(true));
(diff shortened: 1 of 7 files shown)


sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
```diff
@@ -880,6 +881,13 @@ async unsafe fn execute_indexed_with_context_inner(
     // types. The target is schema_coerce::coerce_inferred_schema(physical_schema) — same
     // narrowing the partition-stream registration uses, so consumer-side StreamingTable
     // and producer-side batches agree by construction (see crate::relabel_exec).
+    // Apply aggregate mode stripping when prepare_partial_plan was called (engine-native-merge).
+    // This makes the indexed executor produce Binary HLL state (Partial) instead of Int64 (Final).
+    let physical_plan = if aggregate_mode != crate::agg_mode::Mode::Default {
```
vinaykpud, line 886: Do we have UT for this? if now can we add?

sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_table/substrait_to_tree.rs
```diff
@@ -460,13 +460,19 @@ impl ScalarUDFImpl for DelegationPossibleUdf {
...
-        _args: ScalarFunctionArgs,
+        args: ScalarFunctionArgs,
     ) -> datafusion::common::Result<ColumnarValue> {
-        Err(datafusion::common::DataFusionError::Internal(format!(
-            "{} UDF body invoked — expr_to_bool_tree did not unwrap the marker; \
-                 treat as a serious correctness bug",
-            DELEGATION_POSSIBLE_FUNCTION_NAME
-        )))
+        // Transparent pass-through: return the original predicate's value (first arg).
+        // This fires when the prepared_plan path executes a FilterExec containing
```
expani: Then this would hide correctness issues in the DataNode path based on the previous comment. 

Can we differentiate b/w which path is calling AND accordingly run it ? 

`prepared_plan` should return here BUT data node should not

expani: If the change to pass along that metadata is too complex, we can add a TODO and contine

expani (approved):

vinaykpud (approved):

mch2 (approved):

PR #22103, opened 2026-06-10
Harden hot-to-warm tiering and file registry warm read routing

vishwasgarg18:
### Description

This change fixes two issues in the writable-warm (hot-to-warm tiering) path:

**1. Non-blocking merge drain — fixes a thread-pool deadlock during tiering prepare**

The prepare-tiering step used to block a thread (on the `MANAGEMENT` pool) while
waiting for in-flight merges to drain. Under load this could exhaust the pool and
deadlock, since the merges themselves need threads to complete. This change makes
the prepare path fully non-blocking:

- `TransportBroadcastByNodeAction` gains an opt-in async mode (`shardOperationAsync`)
  that dispatches each shard operation to a thread pool and sends the node response
  only after all shards complete — no thread is held while waiting.
- The DFA engine/merge scheduler expose a `freeze`/`onDrained` model: the prepare
  action freezes the engine (blocking new merges and primary writes), registers a
  merge-drain listener, and completes the flush/refresh/remote-sync on the callback.
  In-flight merges drain to completion; pending merges are not re-registered while frozen.
- A configurable, dynamic timeout `cluster.tiering.prepare_timeout` (default 90s) bounds
  the wait. On timeout the shard fails with a typed `MergeDrainTimeoutException` carrying
  the active/pending merge counts, and the coordinator surfaces targeted, actionable
  error messages.
- Cancel now also resolves the tiering service from the persisted tiering state, so a
  mid-migration index remains cancellable after a cluster-manager failover.

**2. Per-query object-store registry isolation (DataFusion backend)**

Each per-query `RuntimeEnv` now gets its own `DefaultObjectStoreRegistry`. Previously,
concurrent queries registering different shard stores under the bare `file://` scheme
could clobber each other, routing one shard's parquet read through another shard's store
and failing with "No such file or directory."

### Related Issues
Resolves #[issue number]

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
@@ -26,7 +26,6 @@ use datafusion::{
     physical_plan::displayable,
     physical_plan::execute_stream,
     execution::SessionStateBuilder,
-    execution::runtime_env::RuntimeEnvBuilder,
     execution::context::SessionContext,
     common::DataFusionError,
     prelude::*,
@@ -34,7 +33,7 @@ use datafusion::{
     catalog::Session,
     common::tree_node::{TreeNode, TreeNodeRecursion},
     datasource::{TableProvider, TableType},
-    execution::cache::cache_manager::{CacheManagerConfig, CachedFileList},
+    execution::cache::cache_manager::CachedFileList,
     execution::cache::{CacheAccessor, DefaultListFilesCache, TableScopedPath},
     execution::memory_pool::MemoryPool,
     execution::object_store::ObjectStoreUrl,
@@ -109,20 +108,7 @@ pub async fn execute_indexed_query(
     };
     list_file_cache.put(&table_scoped_path, CachedFileList::new(shard_view.object_metas.as_ref().clone()));
 
-    let mut runtime_env_builder = RuntimeEnvBuilder::from_runtime_env(&runtime.runtime_env)
-        .with_cache_manager(
-            CacheManagerConfig::default()
-                .with_list_files_cache(Some(list_file_cache))
-                .with_file_metadata_cache(Some(
-                    runtime.runtime_env.cache_manager.get_file_metadata_cache(),
-                ))
-                .with_metadata_cache_limit(
-                    runtime.runtime_env.cache_manager.get_metadata_cache_limit(),
-                )
-                .with_file_statistics_cache(
-                    runtime.runtime_env.cache_manager.get_file_statistic_cache(),
-                ),
-        );
+    let mut runtime_env_builder = crate::query_executor::query_runtime_env_builder(runtime, list_file_cache);
     if let Some(pool) = query_memory_pool {
         runtime_env_builder = runtime_env_builder.with_memory_pool(pool);
     }
--- a/server/src/main/java/org/opensearch/OpenSearchServerException.java
+++ b/server/src/main/java/org/opensearch/OpenSearchServerException.java
@@ -25,6 +25,7 @@ import static org.opensearch.Version.V_2_7_0;
 import static org.opensearch.Version.V_3_0_0;
 import static org.opensearch.Version.V_3_2_0;
 import static org.opensearch.Version.V_3_7_0;
+import static org.opensearch.Version.V_3_8_0;
 
 /**
  * Utility class to register server exceptions
@@ -1248,6 +1249,14 @@ public final class OpenSearchServerException {
                 org.opensearch.index.engine.dataformat.merge.MergeFailedEngineException::new,
                 178,
                 V_3_7_0
+            )
+        );
+        registerExceptionHandle(
+            new OpenSearchExceptionHandle(
+                org.opensearch.storage.action.tiering.MergeDrainTimeoutException.class,
+                org.opensearch.storage.action.tiering.MergeDrainTimeoutException::new,
+                179,
+                V_3_8_0
             )
         );
     }
--- a/server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
+++ b/server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
@@ -76,6 +76,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReferenceArray;
 
@@ -102,6 +103,7 @@ public abstract class TransportBroadcastByNodeAction<
     private final ClusterService clusterService;
     private final TransportService transportService;
     private final IndexNameExpressionResolver indexNameExpressionResolver;
+    private final ThreadPool threadPool;
 
     final String transportNodeBroadcastAction;
 
@@ -132,6 +134,7 @@ public abstract class TransportBroadcastByNodeAction<
         this.clusterService = clusterService;
         this.transportService = transportService;
         this.indexNameExpressionResolver = indexNameExpressionResolver;
+        this.threadPool = transportService.getThreadPool();
 
         transportNodeBroadcastAction = actionName + "[n]";
 
@@ -234,6 +237,23 @@ public abstract class TransportBroadcastByNodeAction<
     protected abstract ShardOperationResult shardOperation(Request request, ShardRouting shardRouting) throws IOException;
 
     /**
+     * Async shard operation with ActionListener callback.
+     * Subclasses that opt into async execution (via {@link #isAsyncShardOperation()}) MUST
+     * override this method with their non-blocking implementation.
+     * The default throws {@link UnsupportedOperationException} — async subclasses should
+     * never fall through to the sync path.
+     *
+     * @param request      the indices-level request
+     * @param shardRouting the shard on which to execute the operation
+     * @param listener     callback to invoke with the result or failure
+     */
+    protected void shardOperationAsync(Request request, ShardRouting shardRouting, ActionListener<ShardOperationResult> listener) {
+        throw new UnsupportedOperationException(
+            "shardOperationAsync must be overridden by subclasses that return isAsyncShardOperation()=true"
+        );
+    }
+
+    /**
      * Determines the shards on which this operation will be executed on. The operation is executed once per shard.
      *
      * @param clusterState    the cluster state
@@ -249,6 +269,33 @@ public abstract class TransportBroadcastByNodeAction<
      * @param accumulatedExceptions List of any exceptions thrown by the shard-level operations.
      */
     protected void nodeOperation(List<ShardOperationResult> results, List<BroadcastShardOperationFailedException> accumulatedExceptions) {}
+
+    /**
+     * Returns true if shard operations should execute asynchronously on the data node.
+     * When true, each shard operation is submitted to the thread pool in parallel and the
+     * transport response is sent only after all shards complete (via CountDownLatch).
+     * This prevents long-running shard operations (e.g., flush + remote sync) from blocking
+     * the thread pool thread that received the transport request.
+     * <p>
+     * Default is false (synchronous, sequential execution) for backwards compatibility.
+     *
+     * @return true if shard operations should be dispatched asynchronously
+     */
+    protected boolean isAsyncShardOperation() {
+        return false;
+    }
+
+    /**
+     * Returns the thread pool name to use when dispatching async shard operations on the data node.
+     * Only consulted when {@link #isAsyncShardOperation()} is true.
+     * Default is {@link ThreadPool.Names#GENERIC} — a scaling pool, so per-shard tasks do not queue
+     * behind each other on a bounded executor. Override to redirect to a dedicated pool if needed.
+     *
+     * @return the thread pool name for async per-shard execution
+     */
+    protected String asyncShardOperationThreadPool() {
+        return ThreadPool.Names.GENERIC;
+    }
 
     /**
      * Executes a global block check before polling the cluster state.
@@ -480,12 +527,109 @@ public abstract class TransportBroadcastByNodeAction<
             }
             final Object[] shardResultOrExceptions = new Object[totalShards];
 
-            int shardIndex = -1;
-            for (final ShardRouting shardRouting : shards) {
-                shardIndex++;
-                onShardOperation(request, shardResultOrExceptions, shardIndex, shardRouting);
+            if (isAsyncShardOperation()) {
+                logger.trace("[{}] executing async operation on [{}] shards", actionName, totalShards);
+                onAsyncShardOperation(request, channel, shards, totalShards, shardResultOrExceptions);
+            } else {
+                // Sync mode: original sequential execution (backwards compatible)
+                int shardIndex = -1;
+                for (final ShardRouting shardRouting : shards) {
+                    shardIndex++;
+                    onShardOperation(request, shardResultOrExceptions, shardIndex, shardRouting);
+                }
+                sendNodeResponse(request, channel, totalShards, shardResultOrExceptions);
             }
+        }
 
+        /**
+         * Dispatches each shard's {@link #shardOperationAsync} as its own task on
+         * {@link #asyncShardOperationThreadPool()} so a slow/blocking shard prefix (e.g. a permit wait)
+         * does not serialize the others on the single node-handler thread. The listener is wrapped in
+         * {@link ActionListener#notifyOnce} so a shard implementation that completes more than once
+         * (e.g. a racing drain-callback and timeout) still counts as exactly one shard.
+         * If the pool rejects a per-shard dispatch, that shard is recorded as a failure via the same
+         * listener so {@code remaining} still reaches zero and the node response is sent.
+         */
+        private void onAsyncShardOperation(
+            final NodeRequest request,
+            TransportChannel channel,
+            List<ShardRouting> shards,
+            int totalShards,
+            Object[] shardResultOrExceptions
+        ) {
+            final AtomicInteger remaining = new AtomicInteger(totalShards);
+            final String asyncPool = asyncShardOperationThreadPool();
+            int idx = 0;
+            for (final ShardRouting shardRouting : shards) {
+                final int shardIndex = idx++;
+                ActionListener<ShardOperationResult> shardListener = ActionListener.notifyOnce(new ActionListener<ShardOperationResult>() {
+                    @Override
+                    public void onResponse(ShardOperationResult result) {
+                        shardResultOrExceptions[shardIndex] = result;
+                        onShardOperationComplete(request, channel, totalShards, shardResultOrExceptions, remaining);
+                    }
+
+                    @Override
+                    public void onFailure(Exception e) {
+                        BroadcastShardOperationFailedException failure = new BroadcastShardOperationFailedException(
+                            shardRouting.shardId(),
+                            "operation " + actionName + " failed",
+                            e
+                        );
+                        failure.setShard(shardRouting.shardId());
+                        shardResultOrExceptions[shardIndex] = failure;
+                        onShardOperationComplete(request, channel, totalShards, shardResultOrExceptions, remaining);
+                    }
+                });
+                try {
+                    threadPool.executor(asyncPool).execute(() -> {
+                        try {
+                            shardOperationAsync(request.indicesLevelRequest, shardRouting, shardListener);
+                        } catch (Exception inner) {
+                            // shardOperationAsync threw synchronously on the pool worker (e.g. an
+                            // IndexNotFoundException raced index deletion). Without this, the throw
+                            // would escape onto the pool's uncaught-exception handler — the listener
+                            // would never fire, `remaining` would never decrement, and the coordinator
+                            // would only get a response after its transport timeout. notifyOnce keeps
+                            // this safe even if the impl partially completed the listener before throwing.
+                            shardListener.onFailure(inner);
+                        }
+                    });
+                } catch (Exception e) {
+                    // Pool rejected at dispatch time (shutdown / saturated) — record as a shard failure
+                    // on the same (notifyOnce-wrapped) listener so the remaining counter still reaches
+                    // zero and the node response is sent.
+                    RejectedExecutionException rejection = new RejectedExecutionException(
+                        "rejected by thread pool [" + asyncPool + "] for shard " + shardRouting.shardId(),
+                        e
+                    );
+                    shardListener.onFailure(rejection);
+                }
+            }
+        }
+
+        private void onShardOperationComplete(
+            NodeRequest request,
+            TransportChannel channel,
+            int totalShards,
+            Object[] shardResultOrExceptions,
+            AtomicInteger remaining
+        ) {
+            if (remaining.decrementAndGet() == 0) {
+                try {
+                    sendNodeResponse(request, channel, totalShards, shardResultOrExceptions);
+                } catch (IOException e) {
+                    logger.warn("[{}] failed to send response after async shard operations", actionName);
+                }
+            }
+        }
+
+        private void sendNodeResponse(
+            final NodeRequest request,
+            TransportChannel channel,
+            int totalShards,
+            Object[] shardResultOrExceptions
+        ) throws IOException {
             List<BroadcastShardOperationFailedException> accumulatedExceptions = new ArrayList<>();
             List<ShardOperationResult> results = new ArrayList<>();
             for (int i = 0; i < totalShards; i++) {
--- a/server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
+++ b/server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
@@ -1387,6 +1387,7 @@ public class DataFormatAwareEngine implements Indexer {
      *
      * @param listener the callback to fire when merges are drained
      */
+    @Override
     public void onMergesDrained(Runnable listener) {
         mergeScheduler.onDrained(listener);
     }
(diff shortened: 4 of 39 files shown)


server/src/main/java/org/opensearch/OpenSearchServerException.java
```diff
@@ -1250,5 +1250,13 @@ public static void registerExceptions() {
                 V_3_7_0
             )
         );
+        registerExceptionHandle(
+            new OpenSearchExceptionHandle(
+                org.opensearch.storage.action.tiering.MergeDrainTimeoutException.class,
+                org.opensearch.storage.action.tiering.MergeDrainTimeoutException::new,
+                179,
+                V_3_7_0
```
nishchay21: this should be new version now

vishwasgarg18: Going ahead with IOException for now. As we are planning to backport to 3.5 an there is no node attribute for exceptions.

sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
```diff
@@ -110,6 +110,7 @@ pub async fn execute_indexed_query(
     list_file_cache.put(&table_scoped_path, CachedFileList::new(shard_view.object_metas.as_ref().clone()));
 
     let mut runtime_env_builder = RuntimeEnvBuilder::from_runtime_env(&runtime.runtime_env)
+        .with_object_store_registry(Arc::new(datafusion::execution::object_store::DefaultObjectStoreRegistry::new()))
```
mgodwan: can this be done at a common place? I see multiple places with same override

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
```diff
@@ -250,6 +272,32 @@ protected abstract Response newResponse(
...
+     * Returns the thread pool name to use for async shard operations.
+     * Only used when {@link #isAsyncShardOperation()} returns true.
+     * Default is {@link ThreadPool.Names#GENERIC}.
+     *
+     * @return the thread pool name for async shard execution
+     */
+    protected String asyncShardOperationThreadPool() {
+        return ThreadPool.Names.GENERIC;
+    }
+
```
mgodwan, line 299: This should use the same executor as one used for shard operation?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -6380,4 +6380,68 @@ public DataFormatAwareEngine getCompositeEngine() {
...
+     *
+     * @param listener the callback to fire when merges are drained
+     * @return true if already drained, false if listener was registered
+     */
+    public boolean onMergesDrained(Runnable listener) {
+        final Indexer engine = getIndexerOrNull();
+        if (engine instanceof DataFormatAwareEngine dfa) {
+            return dfa.onMergesDrained(listener);
+        }
+        return true; // non-DFA engines have no merge scheduler — consider drained
```
mgodwan: This assumption is wrong. Even merges can be happening on non DFA engines as well

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
```diff
@@ -480,12 +528,98 @@ public void messageReceived(final NodeRequest request, TransportChannel channel,
             }
             final Object[] shardResultOrExceptions = new Object[totalShards];
 
-            int shardIndex = -1;
+            if (isAsyncShardOperation()) {
+                logger.info("[{}] executing async operation on [{}] shards", actionName, totalShards);
```
mgodwan: Make this a trace log

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
```diff
@@ -480,12 +528,98 @@ public void messageReceived(final NodeRequest request, TransportChannel channel,
...
+                    );
+                    failure.setShard(shardRouting.shardId());
+                    shardResultOrExceptions[shardIndex] = failure;
+                    if (remaining.decrementAndGet() == 0) {
+                        try {
+                            sendNodeResponse(request, channel, totalShards, shardResultOrExceptions);
+                        } catch (IOException ioEx) {
+                            logger.warn("[{}] failed to send response after rejection", actionName);
+                        }
+                    }
```
mgodwan: Common in onResponse/onFailure. move to a commont method

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
```diff
@@ -480,12 +528,98 @@ public void messageReceived(final NodeRequest request, TransportChannel channel,
...
+            final AtomicInteger remaining = new AtomicInteger(totalShards);
+            final String asyncPool = asyncShardOperationThreadPool();
+            int idx = 0;
             for (final ShardRouting shardRouting : shards) {
-                shardIndex++;
-                onShardOperation(request, shardResultOrExceptions, shardIndex, shardRouting);
+                final int shardIndex = idx++;
+                try {
+                    threadPool.executor(asyncPool).execute(() -> {
+                        shardOperationAsync(request.indicesLevelRequest, shardRouting, new ActionListener<ShardOperationResult>() {
```
mgodwan: Should we ensure this is run only once per shard?

mgodwan: Maybe delegate the cointdown to a RunOnce block?

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
```diff
@@ -233,6 +235,26 @@ protected abstract Response newResponse(
...
+     * @param shardRouting the shard on which to execute the operation
+     * @param listener     callback to invoke with the result or failure
+     */
+    protected void shardOperationAsync(Request request, ShardRouting shardRouting, ActionListener<ShardOperationResult> listener) {
+        try {
+            listener.onResponse(shardOperation(request, shardRouting));
+        } catch (Exception e) {
+            listener.onFailure(e);
+        }
+    }
```
mgodwan, line 254: Better to throw unsupported exception rather than have a sync implementation for async method

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
mgodwan: Given this method is already called on a configured thread pool, do we need to further delegate it?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -6380,4 +6380,68 @@ public DataFormatAwareEngine getCompositeEngine() {
...
+     * tiering. Calling this explicitly makes the frozen state independent of cluster-state
+     * propagation timing. The operation is idempotent and is a no-op for engines that do not
+     * support tiering.
+     */
+    public void freezeForTiering() {
+        final Indexer engine = getIndexerOrNull();
+        if (engine instanceof DataFormatAwareEngine dfa) {
+            dfa.freezeForTiering();
+        }
+    }
```
mgodwan: In general, we should avoid any method in IndexShard which is specific to DFAE. We need to ensure this layer can handle both InternalEngine and DFAE ops

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -6380,4 +6380,68 @@ public DataFormatAwareEngine getCompositeEngine() {
...
+     * Freezes this shard's engine in preparation for tiering.
+     *
+     * <p>Once frozen, no new merges can start, any in-flight merges are drained to completion, and
+     * late-arriving merge results are discarded, leaving the shard's segment state stable for
+     * tiering. Calling this explicitly makes the frozen state independent of cluster-state
+     * propagation timing. The operation is idempotent and is a no-op for engines that do not
+     * support tiering.
+     */
+    public void freezeForTiering() {
+        final Indexer engine = getIndexerOrNull();
```
mgodwan: assert primary routing here

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -2096,4 +2198,24 @@ Exception getFailedEngine() {
...
+     *   <li>Primary indexing operations are rejected with "frozen for tiering"</li>
+     *   <li>{@link #triggerPossibleMerges()} becomes a no-op</li>
+     * </ul>
+     */
+    public void freezeForTiering() {
+        if (!frozenForTiering) {
+            frozenForTiering = true;
+            mergeScheduler.freeze();
+            logger.info("Engine explicitly frozen for tiering by prepare action");
+        }
```
mgodwan: If freeze and unfreeze are expected, please ensure this has check for previous state. i.e. only allow freeze if not already frozen, and vice versa. Also, ensure thread safety for parallel tier/cancel if possible

mgodwan: Lets add concurrency tests around this

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1259,6 +1277,33 @@ public void onSettingsChanged(TimeValue translogRetentionAge, ByteSizeValue tran
...
+        if (shouldFreeze && !frozenForTiering) {
+            logger.info("Freezing engine for tiering — blocking merges, refresh, flush, and catalog commits");
+            frozenForTiering = true;
+            mergeScheduler.freeze();
+        } else if (!shouldFreeze && frozenForTiering) {
+            logger.info("Unfreezing engine — tiering cancelled, resuming normal operations");
+            frozenForTiering = false;
+            mergeScheduler.unfreeze();
+        }
```
mgodwan, line 1349: Move this to a private method

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
mgodwan: Also, shouldn't onSettingsChanged already have applied this? Should we just validate here?

server/src/main/java/org/opensearch/storage/common/tiering/TieringUtils.java
```diff
@@ -125,6 +126,22 @@ private TieringUtils() {}
         Setting.Property.NodeScope
     );
 
+    /**
+     * Timeout for the prepare tiering action (merge drain + flush + sync + remote upload).
+     * Controls how long each shard waits for in-flight merges to complete before timing out.
+     * Default is 90 seconds — large indices with many segments may need significant time
```
mgodwan, line 132: We can keep this 300s

server/src/main/java/org/opensearch/storage/action/tiering/TransportHotToWarmTierAction.java
```diff
@@ -186,13 +189,66 @@ public void onResponse(BroadcastResponse broadcastResponse) {
                         executePrepareTiering(request, state, listener, attempt + 1);
                         return;
                     }
```
mgodwan, line 179: extract common retry code out in onFailure/onResponse to a single method

server/src/main/java/org/opensearch/storage/action/tiering/TransportHotToWarmTierAction.java
```diff
@@ -232,6 +288,26 @@ public void onFailure(Exception e) {
...
+     * (not {@code instanceof}) on purpose: the exception type is not registered for serialization, so
+     * the concrete class does not survive transport, but its message always does. This keeps detection
+     * working across mixed-version clusters with no version/registry coupling. Returns the matching
+     * message, or {@code null} if no merge-drain timeout is present in the chain.
+     */
+    private static String findMergeDrainTimeoutMessage(Throwable t) {
+        int depth = 0;
+        while (t != null && depth++ < 10) {
+            final String message = t.getMessage();
+            if (message != null && message.contains(MergeDrainTimeoutException.MERGE_DRAIN_TIMEOUT_MARKER)) {
```
mgodwan, line 317: Instead of relying on the exception message, please create a specific exception type and check for the same

server/src/main/java/org/opensearch/storage/action/tiering/TransportHotToWarmTierAction.java
```diff
@@ -186,13 +189,66 @@ public void onResponse(BroadcastResponse broadcastResponse) {
...
+                            + MAX_PREPARE_RETRIES
+                            + " attempts: "
+                            + broadcastResponse.getFailedShards()
+                            + " shard(s) failed. "
+                            + "First failure: "
+                            + firstFailure
+                            + ". Please retry.";
+                    }
                     logger.error(errorMsg);
                     removeWriteBlock(request.getIndex());
```
mgodwan: We should have write block removal in a finally block

server/src/main/java/org/opensearch/storage/action/tiering/TransportPrepareTieringAction.java
```diff
@@ -109,22 +145,102 @@ protected PrepareTieringRequest readRequestFrom(StreamInput in) throws IOExcepti
...
+            if (completed.compareAndSet(false, true)) {
+                int activeMerges = indexShard.getActiveMergeCount();
+                int pendingMerges = indexShard.getPendingMergeCount();
+                permit.close();
+                listener.onFailure(
+                    new MergeDrainTimeoutException(shardRouting.shardId(), activeMerges, pendingMerges, mergeTimeout.toString())
+                );
+            }
+        }, effectiveTimeout, ThreadPool.Names.GENERIC);
+
```
mgodwan, line 209: Can we leverage ClusterAdminTask here?

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
```diff
@@ -188,13 +197,85 @@ public synchronized double getIORateLimitMBPerSec() {
...
+     * thread when the last merge finishes.
+     * <p>
+     * Multiple listeners can be registered concurrently (thread-safe via CopyOnWriteArrayList).
+     *
+     * @param listener the callback to fire when merges are drained
+     * @return true if already drained (caller can proceed), false if listener was registered
+     */
+    public boolean onDrained(Runnable listener) {
+        if (activeMerges.get() == 0 && !mergeHandler.hasPendingMerges()) {
+            return true; // already drained
```
mgodwan: Instead of true/false, can we have the listener fired directly here and reduce the additional client side handling? This would reduce the duplicate code in TransportPrepareTieringAction

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
```diff
@@ -258,6 +339,18 @@ private void submitMergeTask(OneMerge oneMerge) {
...
+                    for (Runnable listener : listeners) {
+                        try {
+                            listener.run();
+                        } catch (Exception ex) {
+                            logger.warn("Exception in onDrained listener", ex);
+                        }
+                    }
+                }
                 // A completed merge may free up capacity for new merges, so check again.
                 executeMerge();
```
mgodwan, line 351: Should this be done in finally block for the above added code?

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
```diff
@@ -258,6 +339,18 @@ private void submitMergeTask(OneMerge oneMerge) {
...
+                    List<Runnable> listeners = List.copyOf(onDrainedListeners);
+                    onDrainedListeners.clear();
+                    for (Runnable listener : listeners) {
+                        try {
+                            listener.run();
+                        } catch (Exception ex) {
+                            logger.warn("Exception in onDrained listener", ex);
+                        }
+                    }
+                }
```
mgodwan, line 349: If this is satisfied, it would mean that executeMerges should not trigger any new merges, right? Can we add an assertion around this as well?

server/src/main/java/org/opensearch/storage/action/tiering/MergeDrainTimeoutException.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Exception thrown when a shard's merge drain times out during tiering preparation.
+ * <p>
+ * This is intentionally a plain {@link IOException} that is <b>not</b> registered with the
+ * {@code OpenSearchException} serialization registry. Detection on the coordinator is done by
+ * matching {@link #MERGE_DRAIN_TIMEOUT_MARKER} in the exception message rather than by
+ * {@code instanceof}. The message is always preserved across the wire (an unregistered exception
+ * is serialized as a {@code NotSerializableExceptionWrapper}), so detection works in any
+ * mixed-version cluster without needing a {@code versionAdded} guard or a registry id — which
+ * keeps it safe to backport into an already-released line.
```
mgodwan: Message based checks are fragile. We should avoid

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
```diff
@@ -188,13 +197,85 @@ public synchronized double getIORateLimitMBPerSec() {
...
+    public void freeze() {
+        frozen.set(true);
+    }
+
+    /**
+     * Unfreezes the merge scheduler, allowing merges to resume.
+     * Called when tiering is cancelled.
+     */
+    public void unfreeze() {
+        frozen.set(false);
```
mgodwan: The way we freeze is that we stop finding new merges. Any unfreeze should then allow the handler to be able to findAndRegister new merges as well to ensure we can reach the equilibrium state for files

mgodwan:
https://github.com/opensearch-project/OpenSearch/pull/22103#issuecomment-4675026151

These need to be addressed as well

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
mgodwan: Also, this is not safe to execute in cases when freeze has not happened as this will never be thread safe (other threads add more merges while onDrained execute). We should keep the scope limited to frozen state and wire accordingly

server/src/main/java/org/opensearch/storage/action/tiering/TransportPrepareTieringAction.java
```diff
@@ -109,22 +145,102 @@ protected PrepareTieringRequest readRequestFrom(StreamInput in) throws IOExcepti
...
+        // Non-blocking merge wait
+        boolean alreadyDrained = indexShard.onMergesDrained(() -> {
+            if (completed.compareAndSet(false, true)) {
+                timeout.cancel();
+                try {
+                    completeSyncAndFlush(indexShard, shardRouting);
+                    permit.close();
+                    listener.onResponse(EmptyResult.INSTANCE);
+                } catch (Exception e) {
+                    permit.close();
```
mgodwan: Close permits in finally block

sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
vishwasgarg18: Yes, makes sense.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
vishwasgarg18: Ack, updated. Firing the listener immediately for non-dfa.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
vishwasgarg18: Ack

server/src/main/java/org/opensearch/index/shard/IndexShard.java
vishwasgarg18: Done

server/src/main/java/org/opensearch/storage/common/tiering/TieringUtils.java
vishwasgarg18: We already have 3 retries and ISM timeout is 5 mins. Hence keeping it the same. Let me know if there are any concerns.

server/src/main/java/org/opensearch/storage/action/tiering/TransportPrepareTieringAction.java
vishwasgarg18: Yes, done!

server/src/main/java/org/opensearch/storage/action/tiering/TransportHotToWarmTierAction.java
vishwasgarg18: sure

server/src/main/java/org/opensearch/storage/action/tiering/MergeDrainTimeoutException.java
vishwasgarg18: ack, avoided it earlier since we have to port the change to 3.5 as well and there is no node attribute associated with exceptions. 
Anyway, added it. Will make the managed side change.

server/src/main/java/org/opensearch/storage/action/tiering/TransportHotToWarmTierAction.java
vishwasgarg18: ack

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
vishwasgarg18: ack, added triggermerges to findAndRegister any new merges post `frozen.set(false);`

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
vishwasgarg18: Updated

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
vishwasgarg18: Sure, added the `isfrozen` condition.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
vishwasgarg18: ack

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
vishwasgarg18: Ack

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
vishwasgarg18: We are already catching Exception above and in the finally block.

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
vishwasgarg18: done

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
vishwasgarg18: Ack

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
vishwasgarg18: ack

server/src/main/java/org/opensearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java
vishwasgarg18: Yes, we still need it. The handler runs on a single thread per node and shardOperationAsync blocks on acquirePrimaryPermits().actionGet() (up to 30s) before it returns and also registering the drain listener. Without delegating, the shards would run one after another and one slow shard would block the rest on that node. Dispatching each shard separately lets them run in parallel and frees the handler thread immediately.

vishwasgarg18: Good point. 
Wrapped each shard's listener in ActionListener.notifyOnce(...) so a double-completion (drain vs timeout) can't over-decrement the countdown. Each shard counts once, node response fires once. Same idea as RunOnce, but on the listener itself. 

Added a UT that completes a shard twice and asserts counts stay correct.

server/src/main/java/org/opensearch/storage/action/tiering/TransportHotToWarmTierAction.java
vishwasgarg18: Ack, done!

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
vishwasgarg18: > Also, shouldn't onSettingsChanged already have applied this? Should we just validate here?

No, we freeze the engine before applying any settings change. This change is required. OnSettingsChanged we do validate and update the values as per the current state.

>Lets add concurrency tests around this

Added

>If freeze and unfreeze are expected, please ensure this has check for previous state. i.e. only allow freeze if not already frozen, and vice versa. Also, ensure thread safety for parallel tier/cancel if possible

Makes sense, done!

server/src/main/java/org/opensearch/storage/action/tiering/TransportPrepareTieringAction.java
vishwasgarg18: Created an issue - https://github.com/opensearch-project/OpenSearch/issues/22190 to do it as a follow up.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
mgodwan: No, even invoking the listener is wrong. 
Also, instead of these instanceOf checks, this should be delegated to specific implementation

mgodwan (approved):

server/src/main/java/org/opensearch/index/shard/IndexShard.java
vishwasgarg18: Delegated to individual engines, rather than having instanceOf check.

andrross:
@vishwasgarg18 @mgodwan A test introduced here is very flaky: #22216 

Can we get someone to revert/mute/fix this ASAP?

PR #22104, opened 2026-06-10
Use IPC for sorted chunk staging with streaming k-way merge at finalize

Bukhtawar:
## Summary

When sorted columns are configured, the sorted write path (`SortingChunkedWriter`) flushes intermediate chunks as **Arrow IPC** instead of Parquet. This eliminates the per-chunk encode-decode cycle — chunks are written as raw Arrow buffers (~5ms vs ~70ms Parquet encode per chunk), and at finalize the streaming k-way merge reads IPC directly without decompression. The Parquet encoding happens exactly once, at the final output stage.

### Where the gain is

The `DataFormatAwareEngine` uses cooperative flush via `preIndex()` — indexing threads help drain pending flush work. The flush path calls `writer.write()` (which triggers `flush_and_sort_chunk`) and `writer.flush()` (which triggers `finalize_sorted_chunks`). The finalize step dominates total flush time (~2000ms merge+encode), so the per-chunk savings (~65ms) are proportionally small on the indexing thread itself.

The primary benefit is on the **`vsrRotationThread`** — the background thread that handles VSR rotation writes. By reducing `flush_and_sort_chunk` from ~70ms to ~5ms per chunk:

- **Rotation throughput**: background writer finishes N × 65ms faster per segment, preventing the frozen slot from backing up
- **VSR accumulation**: when the background thread keeps up, rotations aren't skipped → VSRs stay bounded → predictable heap usage
- **Flush/refresh latency**: ~65ms saved per pending chunk on the cooperative flush path

### CPU savings

Per segment (assuming 4 chunks at 32MB threshold):

| Work eliminated | CPU saved |
|---|---|
| N × Parquet encode (compression + page encoding + dict + bloom) | 4 × ~65ms = **~260ms** |
| N × Parquet decode at finalize (decompression + page reconstruction) | 4 × ~15ms = **~60ms** |
| **Total CPU freed per segment** | **~320ms** |

Work added (negligible):

| Work added | CPU cost |
|---|---|
| N × IPC write (memcpy of Arrow buffers) | 4 × ~5ms = ~20ms |
| N × CRC32 write + verify (hardware-accelerated) | 4 × ~2ms = ~8ms |
| **Total added per segment** | **~28ms** |

**Net: ~290ms CPU freed per segment.** On a loaded node with many shards flushing concurrently, this reduces CPU contention — the freed cycles are available for indexing, merges, and queries.

### How each operation changes

| Operation | Before (Parquet chunks) | After (IPC chunks) |
|-----------|------------------------|---------------------|
| **write_data** | Batch → IPC staging (unchanged) | Same |
| **flush_and_sort_chunk** | Read IPC → sort → **Parquet encode (~70ms)** | Read IPC → sort → **IPC write (~5ms)** |
| **finalize (single chunk)** | rename (0ms) | Stream IPC → Parquet encode (~70ms) |
| **finalize (multi-chunk)** | Parquet decode + k-way merge + encode | **Memcpy read** + k-way merge + encode |
| **segment merge** | Unchanged (operates on final Parquet) | Same |

Net: encode cost moves from N flush points to 1 finalize point. Eliminates 2N codec passes → 1.

### Thread model

```
Indexing thread → addDocument → VSR field append → return
                                  ↓ (when VSR full)
                      maybeRotateActiveVSR()
                                  ↓
              vsrRotationThread → writer.write() → flush_and_sort_chunk ← SAVED HERE
                                  
Indexing thread → preIndex() → flushQueue.poll()
                                  ↓ (cooperative flush if pending)
                      writer.write() → flush_and_sort_chunk  ← ~65ms saved
                      writer.flush() → finalize_sorted_chunks ← ~2000ms (dominates)
```

### Key design decisions

- **Feature flag** (`ipc_sorted_chunks` in NativeSettings, default `false`) — opt-in, no behavior change unless enabled
- **Shared merge loop** — `run_merge_loop()` is called by both `merge_sorted` (Parquet inputs) and `merge_sorted_from_ipc` (IPC inputs), eliminating code duplication
- **No unsafe** — IPC reader lives on a dedicated thread, sends projected+sliced batches through a bounded `sync_channel(1)`. `ChannelIterator` wraps the receiver as a standard Send iterator
- **Bounded memory** — per-cursor memory is 2 × batch_size rows (one in channel + one active), matching the Parquet reader's streaming profile
- **CRC32 integrity** — `CrcWriter` wraps IPC output; verified via `compute_file_crc32` before merge at finalize
- **Multi-batch IPC write** — chunks are written in `batch_size` slices, ensuring the reader thread yields small batches
- **Independent row counts** — `chunk_row_counts` tracked separately from `chunk_row_ids`, correct even when `__row_id__` column is absent
- **Streaming merge via MergeContext** — Rayon parallel column encoding + rate-limited IO (same infrastructure as segment merge)

### Benchmark results (1.5M rows, 32MB threshold, 7-col schema)

Measures the native write path in isolation (write_data + finalize, no Java/VSR overhead):

```
Near-sorted data:
  Write phase:     644ms → 460ms  (+28.6% faster)
  Finalize:       2161ms → 2113ms (+2.2%, no regression)
  Total:          2805ms → 2573ms (+8.3%)

Random data:
  Write phase:     729ms → 451ms  (+38.2% faster)
  Finalize:       3512ms → 3521ms (-0.2%, within noise)
  Total:          4241ms → 3972ms (+6.4%)
```

The write-phase gain reflects what the `vsrRotationThread` and cooperative flush path (`preIndex`) experience. The total gain is diluted by finalize which dominates and is unchanged.

### Files changed

- `native_settings.rs` — Add `ipc_sorted_chunks: Option<bool>` flag
- `writer.rs` — IPC chunk write with CRC + multi-batch slicing, CRC verification at finalize, `build_single_chunk_permutation` / `build_multi_chunk_permutation` helpers, independent `chunk_row_counts`
- `merge/cursor.rs` — `FileCursor::new_from_ipc()` with dedicated reader thread + `ChannelIterator`
- `merge/sorted.rs` — Extract `run_merge_loop()`, add `merge_sorted_from_ipc()` entry point
- `merge/mod.rs` — Export `merge_sorted_from_ipc`
- `benches/sorted_write_bench.rs` — Criterion benchmark (near-sorted + random distributions)
- `Cargo.toml` — Add criterion dev-dependency

### Risks and mitigations

| Risk | Mitigation |
|------|-----------|
| Disk space (IPC uncompressed ~32MB vs ~10MB Parquet) | Ephemeral: deleted after finalize succeeds |
| Single-chunk overhead (IPC→Parquet at finalize vs rename) | Break-even: saves same time at flush |
| Row ID permutation correctness | Same merge algorithm, same mapping logic, dedicated helpers |
| Crash leaves orphaned .arrow_ipc files | Same cleanup responsibility as existing .parquet chunks |
| Silent corruption | CRC32 computed at write, verified before merge |
| Native memory pressure | Bounded: channel-based reader, 2 × batch_size per cursor |

### Check List
- [x] Functionality includes testing (16 unit/integration tests pass)


mgodwan:
We had earlier tried a similar approach, but parked it because of the overall additional IOPS usage this have. (CPU vs I/O)
We saw intermediate files reaching upto 128MBs per segment on 30/60s refresh intervals as well, where-as parquet files were <20MB

Can we measure the additional transient storage cost, and change in iops/throughput load as well?

cc @chaitanya588

Bukhtawar:
> We had earlier tried a similar approach, but parked it because of the overall additional IOPS usage this have. (CPU vs I/O)

That's right, a middle ground is use light-weight lz4 compression on IPC, the other thing I was thinking was write to /tmpfs and spill as needed

PR #22113, opened 2026-06-11
Add missing physical plan debug logs for pure-Parquet and coordinator paths

finnegancarroll:
The indexed executor (delegation path) logs both logical and physical DataFusion plans at DEBUG level, but the query executor (pure Parquet) only logs the logical plan, and the local executor (coordinator reduce) logs neither.

This adds the missing `log_debug!` statements so all three execution paths consistently log both plans:

| Executor | Before | After |
|----------|--------|-------|
| `indexed_executor.rs` (delegation) | ✅ logical + ✅ physical | no change |
| `query_executor.rs` (pure Parquet) | ✅ logical + ❌ physical | ✅ logical + ✅ physical |
| `local_executor.rs` (coordinator) | ❌ neither | ✅ logical + ✅ physical |

Enable logger
```
PUT /_cluster/settings
  {
    "transient": {
      "logger.org.opensearch.nativebridge.spi.RustLoggerBridge": "DEBUG"
    }
  }
```

### Thread leaks in test runs

Our tests configure 2 DF worker threads which persist beyond the lifetime of the test. When we add an up call from DF these threads become visible to the JVM and tracked by the test infrastructure. As these thread pools persist beyond the lifetime of the test teardown our test harness considers them "leaked".

To help identify these "leaked" threads we tag them with names in our logging up call.
```
static void log(int level, MemorySegment msgPtr, long msgLen) {
        Thread t = Thread.currentThread();
        if (t.getName().startsWith("Thread-")) {
            t.setName("datafusion-native-" + t.getId());
        }
```

And observer that this is the reported to leak by test infrastructure.
```
 DatafusionMemtableReduceSinkTests > classMethod FAILED
      com.carrotsearch.randomizedtesting.ThreadLeakError: 1 thread leaked from SUITE scope at org.opensearch.be.datafusion.DatafusionMemtableReduceSinkTests:
         1) Thread[id=45, name=datafusion-native-45, state=RUNNABLE, group=main]
              at (empty stack)
```

Datafusion in our tests delegates to a fixed worker thread pool and we should not need to worry about these threads making upcall leaking.
(In `DatafusionMemtableReduceSinkTests` which previously reported the above thread leaks)
```
        NativeBridge.initTokioRuntimeManager(2);
```

To resolve these i've added some annotations to various tests impacted by these new upcalls which cause them to report thread leaks.

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/local_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/local_executor.rs
@@ -184,11 +184,13 @@ impl LocalSession {
             DataFusionError::Execution(format!("Failed to decode Substrait plan: {}", e))
         })?;
         let logical_plan = from_substrait_plan(&self.ctx.state(), &plan).await?;
+        log_debug!("DataFusion logical plan:\n{}", logical_plan.display_indent());
         let dataframe = self.ctx.execute_logical_plan(logical_plan).await?;
         let physical_plan = dataframe.create_physical_plan().await?;
 
         let target_schema = crate::schema_coerce::coerce_inferred_schema(physical_plan.schema());
         let physical_plan = crate::relabel_exec::wrap_if_relabel_needed(physical_plan, target_schema)?;
+        log_debug!("DataFusion physical plan:\n{}", displayable(physical_plan.as_ref()).indent(true));
         datafusion::physical_plan::execute_stream(physical_plan, self.ctx.task_ctx())
             .map_err(|e| DataFusionError::Execution(format!("execute_substrait: {}", e)))
     }
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
@@ -350,6 +350,7 @@ pub async fn execute_with_context(
 
         let target_schema = crate::schema_coerce::coerce_inferred_schema(physical_plan.schema());
         let physical_plan = crate::relabel_exec::wrap_if_relabel_needed(physical_plan, target_schema)?;
+        log_debug!("DataFusion physical plan:\n{}", displayable(physical_plan.as_ref()).indent(true));
 
         let df_stream = execute_stream(physical_plan.clone(), handle.ctx.task_ctx()).map_err(|e| {
             error!("execute_with_context: failed to create stream: {}", e);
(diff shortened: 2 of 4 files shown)


finnegancarroll:
testConvInvalidRadixThrows appears flaky
```
MathScalarFunctionsIT > testConvInvalidRadixThrows FAILED
    java.lang.AssertionError: Expected 'less than Character.MIN_RADIX' in error body but got: method [POST], host [http://[::1]:42959], URI [/_plugins/_ppl], status line [HTTP/1.1 500 Internal Server Error]
    {
      "error": {
        "reason": "Error occurred in OpenSearch engine: engine is closed",
        "details": "StreamException[errorCode=INTERNAL, message=engine is closed, metadata={opensearch.content-type=[application/grpc], opensearch.raw-header=[RVMAAAA2AAAAAAAAArABCC7/owAAACUBHF9zeXN0ZW1faW5kZXhfYWNjZXNzX2FsbG93ZWQFZmFsc2UAAAAAAA==], opensearch.correlation-id=[1867763330193555706]}]\nFor more details, please send request for Json format to see the raw response from OpenSearch engine.",

        "type": "StreamException"
      },
      "status": 500
    }
        at __randomizedtesting.SeedInfo.seed([99AC427744C95E97:911DC4F371A4A4E1]:0)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.opensearch.analytics.qa.MathScalarFunctionsIT.testConvInvalidRadixThrows(MathScalarFunctionsIT.java:387)
REPRODUCE WITH: ./gradlew ':sandbox:qa:analytics-engine-rest:integTest' --tests 'org.opensearch.analytics.qa.MathScalarFunctionsIT.testConvInvalidRadixThrows' -Dtests.seed=99AC427744C95E97 -Dtests.security.manager=false -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=yo-BJ -Dtests.timezone=Africa/Freetown -Druntime.java=25
```

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
```diff
@@ -45,6 +45,10 @@ public class RustLoggerBridge {
      * Signature must match: {@code void(int level, MemorySegment msgPtr, long msgLen)}
      */
     static void log(int level, MemorySegment msgPtr, long msgLen) {
+        Thread t = Thread.currentThread();
```
expani: Can you add a comment on why is this required ?

finnegancarroll: Ah, it's not. This was me trying to identify thread leaks reported in CI. Removing.

mch2 (approved):

PR #22116, opened 2026-06-11
Deprecate REST client (finally)

### LINKED ISSUE #5424: Deprecate REST client

reta:
### Description
Deprecate REST client ( `RestClient`). The internal HTTP client (see please https://github.com/opensearch-project/OpenSearch/issues/20634) has been introduced and designed for OpenSearch internal use only. The `RestHighLevelClient` is still going to be supported for now but migrated away from `RestClient`.

### Related Issues
Closes https://github.com/opensearch-project/OpenSearch/issues/5424

### Check List

### DIFF
--- a/client/rest/src/main/java/org/opensearch/client/RestClient.java
+++ b/client/rest/src/main/java/org/opensearch/client/RestClient.java
@@ -136,7 +136,11 @@ import static java.util.Collections.singletonList;
  * Requests can be either synchronous or asynchronous. The asynchronous variants all end with {@code Async}.
  * <p>
  * Requests can be traced by enabling trace logging for "tracer". The trace logger outputs requests and responses in curl format.
+ *
+ * @deprecated the {@code RestClient} is deprecated is going to be removed in future releases, please consider using official
+ * OpenSearch Java Client or {@code RestHighLevelClient}
  */
+@Deprecated
 public class RestClient implements Closeable {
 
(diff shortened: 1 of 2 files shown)


cwperks (approved):

marko-bekhta:
hey @reta 👋🏻 
I noticed this deprecation during the upgrade and am trying to understand: what is an actual alternative to it?

we have an OpenSearch-based client in Hibernate Search (https://github.com/hibernate/hibernate-search/tree/main/backend/elasticsearch-client/opensearch-rest-client) that users can use when connecting to OpenSearch clusters. Neither high-level (my understanding is that you plan to deprecate it anyway, and it's not suitable in the first place for the use case 🙂) nor the opensearch-Java clients, which are suggested, look like a good alternative here ...

Why the current RestClient was a good fit -- we need a client that can talk to any OpenSearch version without bringing version specifics with it. Hibernate Search can handle versions internally, but it needs something to handle the node management + sending requests.

PR #22117, opened 2026-06-12
[Arrow Flight RPC] Fail the stream when a batch send fails instead of hanging the consumer

LantaoJin:
### Description

When `FlightOutboundHandler.processBatchTask` fails to serialize/send a batch (e.g. a malformed Arrow batch with no field vectors, or any transport error), it only notified the message listener and **dropped** the batch. The stream was then closed via the normal `completeStream` path, but no schema frame ever reached the client -- so the consumer's `FlightStream.getRoot()` **blocked forever** waiting for the first frame, turning a recoverable send error into an indefinite hang.

**Fix:** On a batch-send failure, fail the stream via `FlightServerChannel.sendError` so the error surfaces to the consumer as an exception instead, then release the channel (mirroring the isError BatchTask close path) so a subsequently-enqueued `completeStream` no-ops on its open guard. The stream is failed before the listener is notified so a listener exception cannot leave it un-terminated. Release is idempotent. Also close the handler-created first-frame stream root when `sendBatch` fails before the channel adopts it (`forNativeArrow` close is a no-op), fixing a buffer leak on the failure path.

Adds regression tests for the existing-root and first-frame (no schema sent) failure shapes, asserting `sendError` is invoked, the channel is released, and the stream is failed before the listener is notified.

### Related Issues

### Check List


plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
```diff
@@ -180,15 +189,84 @@ private void processBatchTask(BatchTask task) {
...
+            // listener that throws cannot leave the stream un-terminated and the consumer hanging.
+            failStream(flightChannel, task, handlerCreatedRoot, e);
             messageListener.onResponseSent(task.requestId(), task.action(), FlightErrorMapper.fromFlightException(e));
         } catch (Exception e) {
+            failStream(flightChannel, task, handlerCreatedRoot, e);
             messageListener.onResponseSent(task.requestId(), task.action(), e);
         }
     }
 
+    /**
```
rishabhmaurya, line 181: There is a bunch of redundant comments in this function, could you compress it into a sentence. Just keep the high level idea here i.e. close the root to avoid leak if it isn't handled by the FlightServerChannel

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
```diff
@@ -153,6 +157,10 @@ private void processBatchTask(BatchTask task) {
             return;
         }
 
+        // Root the handler allocates for the first frame (when the channel has none yet). On success
+        // the channel adopts it (sendBatch -> serverStreamListener.start) and closes it at stream
+        // end; on failure we must close it ourselves if the channel never adopted it — see catch.
+        VectorSchemaRoot handlerCreatedRoot = null;
```
rishabhmaurya: I think this error handling is a bit complex than it should be. If I understand it correct, the root isn't closed when there is some issue while creating root or right after while transferring ownership. I'm guessing once we are past `flightChannel.sendBatch() `, flightChannel handles it gracefully. If that's true, we can move this exception handling prior to flightChannel.sendBatch() and also this way we don't have to worry about the double close logic you have added.

LantaoJin: You're right that it reads as more complex than it should, and your comment is almost exactly correct -- the one subtlety is where adoption happens.
The channel adopts the root inside `sendBatch`, at `root = output.getRoot()` (FlightServerChannel:171/174), which is after the entry guards (cancelled / !open.get(), lines 161–167) but before `serverStreamListener.start(root)` / `putNext()`. So a single `sendBatch` call straddles both ownership regimes:

- throws at the entry guards (e.g. client cancelled between our `getRoot()==null` check and the call) → channel never assigned our root → handler must close it (and `NativeArrow.close()` is a no-op, so otherwise it leaks);
- throws at `start()`/`putNext()` → `channel.root == ourRoot` → channel owns it and frees it in `close()` → handler must not close (double-close).

That's the only reason for the `getRoot() == unownedRoot` probe -- moving the handling before `sendBatch` doesn't help, because the failing call is `sendBatch` and it can fail on either side of that assignment.

But I think we can get the simplification you're after by pushing ownership entirely into the channel instead of probing from the handler — have `sendBatch` adopt the root before its guard checks so the channel always owns it on any throw:

```java
 public void sendBatch(ByteBuffer header, VectorStreamOutput output, byte[] metadata) {
     boolean firstFrame = (root == null);
     root = output.getRoot();            // adopt up front; channel.close() now always frees it
     if (cancelled) { throw ... }
     if (!open.get()) { throw ... }
     if (firstFrame) { middleware.setHeader(header); serverStreamListener.start(root); }
     ...
 }
```

 Then the handler drops `handlerCreatedRoot`, the probe, and the double-close logic completely — failStream just does `sendError` + `releaseChannel`. The existing adopt-then-throw test still holds (it asserts the channel owns and closes the root).

LantaoJin: I tried moving adoption above the guards so `close()` always owns it, but that reintroduces a leak in the client-cancel race: if the consumer cancels during the first frame, `onChannelCancelled()`→`close()` can run while root is still null (frees nothing, flips `open=false`), then `sendBatch` adopts the `root` and the guard throws -- and `close()` is one-shot, so the now-adopted `root` never gets freed. The current `getRoot() == unownedRoot` probe avoids this precisely because the field is set only once the channel commits to the batch, and the handler reconciles ownership afterward on its own thread.

Open to a cleaner ownership model if you have one in mind, but I think the post-hoc probe is the race-safe minimum here.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
LantaoJin: done

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
rishabhmaurya: i do not like this ownership check between FlightServerChannel and OutBoundHandler. One of them should be owning it, see if we can move this transfer/create root logic there itself and handle it? or maybe some other way. 
I will spend some time tomorrow if we aren't able to find some good solution

LantaoJin: > i do not like this ownership check between FlightServerChannel and OutBoundHandler.

Done. 549e10d Moved the **create/transfer/adopt/free** logic into `FlightServerChannel.sendArrowBatch`, so the channel owns the `root` end to end. The handler no longer tracks the `root` or probes `getRoot()`; the adoption check is now a channel-internal local comparison (`root != created`) inside the method that does the assignment, so there's no cross-object ownership reasoning and the cancel-race leak from the earlier "adopt above the guards" idea doesn't apply. `failStream` is now just `sendError` + `releaseChannel`. Added `FlightServerChannelTests` coverage for **first-frame/reuse/adopt-then-throw/orphan-root** cases.

rishabhmaurya:
Thanks @LantaoJin for finding and root cause of the issue here. We also encountered other issues with the leaks in this part of code. Since it was urgent to handle them in our day time, we decided to open another PR https://github.com/opensearch-project/OpenSearch/pull/22244. We can close this one, feel free add your review there and I should be able to take them as follow ups.

PR #22127, opened 2026-06-12
Fix disable_objects array parsing to preserve dotted field names

mohit10011999:
### Description
```
final String[] paths = mapper.disableObjects() ? new String[] { lastFieldName } : splitAndValidatePath(lastFieldName);
```
This ensures that when parseNonDynamicArray is called from a disable_objects context (e.g., via processFlattenedTokenForDisableObjects for a dotted field whose value is an array), the dotted field name is kept as a single literal path segment. This prevents parseValue - getDynamicParentMapper from splitting it further and creating rogue intermediate ObjectMappers that lack disable_objects=true.

Tests Added
1. testDisableObjectsArrayWithDottedFieldNameNoIntermediateObjectMapper
```
Simulates the TextBench LogAttributes scenario at nested ObjectMapper level
Indexes {"LogAttributes": {"app": {"payment": ["visa", "mastercard"]}}}
Asserts that LogAttributes.app.payment is a single flat FieldMapper
Asserts that NO intermediate ObjectMapper "app" is created in the dynamic mapping update
```
2. testDisableObjectsRootArrayWithDottedFieldNameNoIntermediateObjectMapper
```
Same scenario at root level with disable_objects=true
Indexes {"metrics": {"cpu": {"cores": [1, 2, 3, 4]}}}
Asserts that metrics.cpu.cores is a single flat FieldMapper
Asserts that NO intermediate ObjectMappers "metrics" or "metrics.cpu" are created
```
### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
+++ b/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
@@ -1192,7 +1192,10 @@ final class DocumentParser {
                 )
             );
         }
-        final String[] paths = splitAndValidatePath(lastFieldName);
+        // When disable_objects is enabled, dotted field names must be treated as literal single-segment
+        // names. Splitting them would cause getDynamicParentMapper to create intermediate ObjectMappers
+        // without disable_objects=tru.
+        final String[] paths = mapper.disableObjects() ? new String[] { lastFieldName } : splitAndValidatePath(lastFieldName);
         while ((token = parser.nextToken()) != XContentParser.Token.END_ARRAY) {
             if (token == XContentParser.Token.START_OBJECT) {
                 parseObject(context, mapper, lastFieldName, paths);
--- a/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
+++ b/server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
@@ -281,6 +281,10 @@ final class DocumentParser {
         return new MapperParsingException("failed to parse", e);
     }
 
+    private static String[] resolvePathForParsing(ObjectMapper mapper, String fieldName) {
+        return mapper.disableObjects() ? new String[] { fieldName } : splitAndValidatePath(fieldName);
+    }
+
     private static String[] splitAndValidatePath(String fullFieldPath) {
         if (fullFieldPath.contains(".")) {
             String[] parts = fullFieldPath.split("\\.");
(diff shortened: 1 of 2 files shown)


cwperks:
@mohit10011999 can you fix the spotless failures?

mohit10011999:
https://github.com/opensearch-project/OpenSearch/pull/22127#issuecomment-4690347726
Done

server/src/main/java/org/opensearch/index/mapper/DocumentParser.java
```diff
@@ -1192,7 +1192,10 @@ private static void parseNonDynamicArray(ParseContext context, ObjectMapper mapp
                 )
             );
         }
-        final String[] paths = splitAndValidatePath(lastFieldName);
+        // When disable_objects is enabled, dotted field names must be treated as literal single-segment
+        // names. Splitting them would cause getDynamicParentMapper to create intermediate ObjectMappers
+        // without disable_objects=tru.
+        final String[] paths = mapper.disableObjects() ? new String[] { lastFieldName } : splitAndValidatePath(lastFieldName);
```
cwperks: @mohit10011999 should we consider moving this logic into `splitAndValidatePath` so that all callers get the right expected return value?

mohit10011999: Decided against it for a few reasons:

1. splitAndValidatePath is a pure string-validation utility with no access to mapper context. We'd need to thread a disableObjects boolean through, which breaks its single-responsibility of "validate and split a dotted path."
2. Not all callers want the disable_objects behavior. The call sites in buildDynamicUpdate always need the full split because they're constructing the mapper hierarchy from fully-qualified mapper names - even for fields under a disable_objects parent. Only the document-parsing call sites (lines 640 and 1198) need the "treat as literal" ,i.e, disable_objects behavior. 
3. The decision is context-dependent (are we inside a disable_objects parent?) rather than a property of the string itself, so it belongs at the caller where that context is available.

cwperks: What about making a small helper like so:

```
private static String[] resolvePathForParsing(ObjectMapper mapper, String fieldName) {
    return mapper.disableObjects() ? new String[] { fieldName } : splitAndValidatePath(fieldName);
}
```

This logic is now duplicated in 2 places

cwperks (approved):

PR #22131, opened 2026-06-12
Support field collapsing + rescore

### LINKED ISSUE #7484: Support field collapsing + rescore
OpenSearch will error "cannot use`collapse` in conjunction with `rescore`", if I have both collapse and rescore clause in the query. In ecommerce space, my existing query rely on collapse (collapse on the same parent ID) to deduplicate the same variations of a product. Because of the limitation, I cannot use [learn to rank plugin](https://github.com/o19s/elasticsearch-learning-to-rank) which need rescore to re-rank to improve my search relevancy.

No error issues when using `collapse` in conjunction with `rescore`. And `rescore` should happen first to ensure the correct ranking of the face out item.

**Describe alternatives you've considered**
N/A

**Additional context**
The same request in Elastic Search https://github.com/elastic/elasticsearch/issues/27243

ackintosh:
First of all, thank you to all the maintainers and contributors of OpenSearch. We use OpenSearch for the search system of our e-commerce platform, and it has been a huge help — this is my first attempt to give something back to the project.

To be upfront: as this is my first contribution, I used AI assistance to navigate and understand the codebase. That said, all of the source code changes in this PR were written by my own hand. I'd be grateful for your patience and guidance, and I'm happy to address any feedback.

> [!IMPORTANT]
> This PR is essentially a port of the `collapse` + `rescore` feature implemented in Elasticsearch (elastic/elasticsearch#107779).
>
> While porting it, I found a **cross-shard correctness issue**: when a collapse group's documents are spread across multiple shards, the group representative can be chosen incorrectly (see `Known limitation: groups spread across shards` below). 
> It has a clear mitigation (route by the collapse field, as the feature already recommends), and single-shard / non-rescore cases are unaffected.
>
> **I'd appreciate maintainer guidance before finalizing this:** is the known limitation (with the routing mitigation + a tracked follow-up) acceptable for this PR, or should the cross-shard fix land first / in the same PR?

### Description

This change adds support for using `rescore` in conjunction with `collapse` in search requests, which was previously rejected with `cannot use 'collapse' in conjunction with 'rescore'`.

Rescoring is applied to the top-ranked document per collapsed group on each shard. After rescoring, the collapse group information is reconstructed so that the coordinator-level merge (deduplication of groups across shards) and the `fields` output keep working as before.

This implements the same functionality that Elasticsearch added in elastic/elasticsearch#107779.

#### Semantics

- Rescoring runs per shard, on the top document of each collapsed group selected by the first-pass query.
- To guarantee that only one top document per collapse key is rescored globally, it is recommended to index documents using the collapse field value as the routing key (same recommendation as Elasticsearch documents for this feature).
- Rescorers are not applied to `inner_hits`.
- Since `rescore` cannot be combined with `sort` (existing restriction), `collapse` + `rescore` always operates on score-sorted hits.

#### Known limitation: groups spread across shards

When the documents of a single collapse group are **spread across multiple shards** (i.e. the collapse field is not used as the routing key), the group representative may be chosen incorrectly. This is reproduced by the integration test `testRescoreAfterCollapseGroupAcrossShards`.

**Root cause.** After rescoring, `RescoreProcessor` rebuilds the `CollapseTopFieldDocs` with the **rescored score** stored as the `FieldDoc` sort value, and the coordinator-level merge (`CollapseTopFieldDocs.merge`) deduplicates groups by that same value. So `CollapseTopFieldDocs` ends up using a single sort field for *both* the cross-shard group dedup (which should use the first-pass score) *and* the final ordering (which should use the rescored score). When a group spans shards, the representative is therefore picked by the rescored (second-pass) score instead of the first-pass score.

### Related Issues
Resolves #7484 

### Check List

### DIFF
--- a/rest-api-spec/src/main/resources/rest-api-spec/test/search/110_field_collapsing.yml
+++ b/rest-api-spec/src/main/resources/rest-api-spec/test/search/110_field_collapsing.yml
@@ -246,25 +246,6 @@ setup:
           collapse: { field: numeric_group }
 
 ---
-"field collapsing and rescore":
-
-  - do:
-      catch:      /cannot use \`collapse\` in conjunction with \`rescore\`/
-      search:
-        allow_partial_search_results: false
-        rest_total_hits_as_int: true
-        index: test
-        body:
-          collapse: { field: numeric_group }
-          rescore:
-            window_size: 20
-            query:
-              rescore_query:
-                match_all: {}
-              query_weight: 1
-              rescore_query_weight: 2
-
----
 "no hits and inner_hits":
 
   - skip:
--- a/rest-api-spec/src/main/resources/rest-api-spec/test/search/111_field_collapsing_with_rescore.yml
+++ b/rest-api-spec/src/main/resources/rest-api-spec/test/search/111_field_collapsing_with_rescore.yml
@@ -1,0 +1,109 @@
+setup:
+  - skip:
+      version: " - 3.7.99"
+      reason: "field collapsing with rescore was added later"
+  - do:
+      indices.create:
+        index: products
+        body:
+          mappings:
+            properties:
+              product_id: { type: keyword }
+              description: { type: text }
+              popularity: { type: integer }
+  - do:
+      bulk:
+        index: products
+        refresh: true
+        body:
+          - '{"index": {"_id": "1", "routing": "0"}}'
+          - '{"product_id": "0", "description": "flat tv 4K HDR", "score": 2, "popularity": 30}'
+          - '{"index": {"_id": "2", "routing": "10"}}'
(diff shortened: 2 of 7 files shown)


ackintosh:
I've marked this PR as ready for review.

One thing I'd especially appreciate feedback on is the known cross-shard limitation described in [the **Important** section](https://github.com/opensearch-project/OpenSearch/pull/22131#issue-4647788700). The current implementation works correctly when documents are routed by the collapse field (the recommended mitigation), but the cross-shard case can still select the wrong group representative, and I've included a failing integration test that demonstrates the issue.

Would maintainers prefer to:

1. merge this feature with the documented limitation and address the cross-shard correctness issue in a follow-up, or
2. require the cross-shard fix before this feature can be merged?

I'd appreciate guidance on the preferred direction before investing further work into the fix.

PR #22136, opened 2026-06-12
Remove AgentPolicy startup log

### LINKED ISSUE #22133: [BUG] warning log entries (stderr) from agent policy on startup
### Describe the bug

```
{"type": "server", "timestamp": "2026-06-12T09:32:51,911+0000", "level": "WARN", "component": "stderr", "cluster.name": "test-opensearch", "node.name": "test-opensearch-0", "message": "Jun 12, 2026 9:32:51 AM org.opensearch.javaagent.bootstrap.AgentPolicy setPolicy" }
{"type": "server", "timestamp": "2026-06-12T09:32:51,977+0000", "level": "WARN", "component": "stderr", "cluster.name": "test-opensearch", "node.name": "test-opensearch-0", "message": "INFO: Policy attached successfully: org.opensearch.bootstrap.OpenSearchPolicy@4e933cf1" }
```

### Related component

Other

### To Reproduce

start a fresh cluster (i checked using the minimal distribution and the normal distribution) and check the log

### Expected behavior

### Additional Details

**Plugins**
n/a

**Screenshots**
n/a

 - OS: tested on windows & linux
 - Version n/a

**Additional context**
maybe related to #21323 which is also about `stderr` - so possibly #21359 would also solve this here? (CC @Chocochip101)

Arin016:
## Description

Fixes #22133

`AgentPolicy.setPolicy()` uses `java.util.logging` to log an INFO message when the security policy is attached. Because the agent's `premain()` runs before `LogConfigurator.configure()` installs the JUL-to-Log4j2 bridge, JUL's default `ConsoleHandler` writes to `System.err`. OpenSearch redirects `System.err` to Log4j2 at WARN level, causing a harmless startup confirmation to appear as:

```
{"level": "WARN", "component": "stderr", "message": "INFO: Policy attached successfully: ..."}
```

## Changes

Change the log level from `INFO` to `FINE`. The default `ConsoleHandler` threshold is `INFO`, so `FINE` messages are suppressed unless explicitly enabled for debugging.

## Issues Resolved

Closes #22133

### DIFF
--- a/libs/agent-sm/bootstrap/src/main/java/org/opensearch/javaagent/bootstrap/AgentPolicy.java
+++ b/libs/agent-sm/bootstrap/src/main/java/org/opensearch/javaagent/bootstrap/AgentPolicy.java
@@ -18,7 +18,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Set;
 import java.util.function.BiFunction;
-import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -27,7 +26,6 @@ import java.util.stream.Stream;
  */
 @SuppressWarnings("removal")
 public class AgentPolicy {
-    private static final Logger LOGGER = Logger.getLogger(AgentPolicy.class.getName());
     private static volatile Policy policy;
     private static volatile Set<String> trustedHosts;
     private static volatile Set<String> trustedFileSystems;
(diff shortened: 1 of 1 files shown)


msfroh (commented):
Can we just remove the log line?

IMO, a log line that is below INFO threshold (especially on startup) will never be used. I think this was useful when the agent was being developed, but now that it's well-tested, we don't need this log line.

@reta -- you built the agent. Do you think this log line is still useful?

Arin016:
@msfroh Happy to remove it entirely, agree it's not useful in production. Will wait for @reta 's input and update accordingly.

reta:
> @msfroh Happy to remove it entirely, agree it's not useful in production. Will wait for @reta 's input and update accordingly.

I think we could drop it folks, not really useful thanks @Arin016 @msfroh !

Arin016:
Updated — removed the log line and the Logger import entirely. Thanks @msfroh @reta for the quick feedback.

reta (approved):

reta:
GA Checks are broken now, we would need to wait for https://github.com/opensearch-project/OpenSearch/pull/22138 very likely

PR #22142, opened 2026-06-13
Fix flaky MatchLikeParityIT: NPE when a fast query completes inline

mch2:
### Description
DefaultPlanExecutor's profiling rows-listener reads execRef.get().getGraph() on the success path, but execRef is set only after scheduler.execute() returns. A fast query (e.g. a scalar count) can fire its terminal callback inline — before execRef.set runs — so execRef.get() is null and the request 500s. The failure path already null-guards execRef; apply the same guard to the success path.

### Related Issues

### Check List


sandeshkr419 (approved):
nit: Please check if we need a dedicated test to validate this change. 

Thanks!

PR #22145, opened 2026-06-13
Remove arrow module reference from jvm.options --add-opens

### LINKED ISSUE #22132: [BUG] warning log entry on startup: "Unknown module: org.apache.arrow.memory.core specified to --add-opens"
### Describe the bug

starting an OpenSearch 3.7.0 distribution logs the following error at the very beginning:

```
WARNING: Unknown module: org.apache.arrow.memory.core specified to --add-opens
```

### Related component

Other

### To Reproduce

start an OpenSearch 3.7.0 distribution and check the log.

### Expected behavior

no such error.

### Additional Details

**Plugins**
n/a (tested both with the minimal distribution and the standard distribution)

**Screenshots**
n/a

 - OS: irrelevant (tested on linux & windows)
 - Version: n/a

**Additional context**
i presume that this is the culprit:
https://github.com/opensearch-project/OpenSearch/blob/723abcb9fd4d70354e0ab78250e7199e40a2a3e2/distribution/src/config/jvm.options#L88

a [code search](https://github.com/search?q=repo%3Aopensearch-project%2FOpenSearch%20org.apache.arrow.memory.core&type=code) shows several plugins having this in their configuration. but these plugins are not installed by default

Arin016:
## Description

Fixes #22132

The global `jvm.options` contains:
```
21-:--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED
```

This references the `org.apache.arrow.memory.core` named module, which only exists when the `arrow-flight-rpc` plugin is installed. On default installations (minimal or standard distribution without the plugin), the JVM emits:

```
WARNING: Unknown module: org.apache.arrow.memory.core specified to --add-opens
```

## Changes

Drop the named module target from the `--add-opens` directive, keeping only `ALL-UNNAMED`. The Arrow Flight plugin loads in the unnamed module space (plugin classloader), so `ALL-UNNAMED` is sufficient for its runtime access to `java.nio.DirectByteBuffer` internals.

## Issues Resolved

Closes #22132

### DIFF
--- a/distribution/src/config/jvm.options
+++ b/distribution/src/config/jvm.options
@@ -85,7 +85,9 @@ ${error.file}
 23:-XX:CompileCommand=dontinline,java/lang/invoke/MethodHandle.asTypeUncached
 
 21-:-javaagent:agent/opensearch-agent.jar
-21-:--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED
+# Required by Arrow memory-core (DirectByteBuffer access); targets ALL-UNNAMED
+# because Arrow loads via plugin classloader (classpath), not module path
+21-:--add-opens=java.base/java.nio=ALL-UNNAMED
 
 # For cases with high memory-mapped file counts, a lower value can improve stability and
 # prevent issues like "leaked" maps or performance degradation. A value of 1 effectively


reta (changes requested):
Thanks @Arin016 

> This references the org.apache.arrow.memory.core named module, which only exists when the arrow-flight-rpc plugin is installed. On default installations (minimal or standard distribution without the plugin), the JVM emits:

So here is the issue: unfortunately, plugins could not contribute JVM options to the OpenSearch distribution, but the `arrow-flight-rpc` needs some, which leaves us with two options:
 - existing approach: add to `jvm.options` by default, with minor inconvenience (arguably) of single warning message on startup
 - ask users to themselves update `jvm.options` if and when they decide to install `arrow-flight-rpc` plugin, this is undoubtedly major undertaking

So what is the exact problem that this warning causes? Thank you.

Arin016:
@reta , thanks for the context, I wasn't aware that plugins can't contribute JVM options to the distribution and that makes sense.
your'e right that the warning doesn't cause functional harm.
the concern from the reporter was log noise, it appears as a WARNING level message which could confuse operators scanning startup logs for real issues. but it does not break anything.
given that constraint, would targeting ALL-UNNAMED instead of the named module work here? Since the arrow-flight-rpc plugin loads via the plugin classloader (classpath), arrow-memory-core ends up in the unnamed module at runtime thus  ALL-UNNAMED should provide the same access without the JVM warning.

But if the preference is to keep the explicit named module reference for clarity purposes and accept the warning, I'm happy to close this PR, totally understand the tradeoff.
do let me know, Thanks.

reta:
> But if the preference is to keep the explicit named module reference for clarity purposes and accept the warning, I'm happy to close this PR, totally understand the tradeoff.
> do let me know, Thanks.

Thanks @Arin016 , I think it is borrowed directly from the Apache Arrow documentation: https://arrow.apache.org/java/main/install.html

krugerLucas:
I opened #22152 as a more complete variant of this fix, mainly to address the review concern that Arrow Flight still needs this JVM option and should keep working after removing the named module target.

The additional coverage in #22152 does three things beyond changing `distribution/src/config/jvm.options`:

- updates the other `org.apache.arrow.memory.core` `--add-opens` usages in Arrow Flight, examples, sandbox, and analytics QA so the repo does not keep mixed guidance/configuration;
- adds an `internalClusterTest` assertion that the Arrow API class, runtime allocator class, and plugin allocator class are loaded from unnamed modules, followed by a real Arrow buffer allocation;
- validates the packaged distribution path with `:qa:smoke-test-arrow-plugins:integTest`, confirming the test cluster starts with `--add-opens=java.base/java.nio=ALL-UNNAMED` and no longer logs `Unknown module: org.apache.arrow.memory.core specified to --add-opens`.

I do not want to duplicate effort unnecessarily. If maintainers prefer to continue with this PR, the test/config cleanup from #22152 can be reused here. I opened the separate PR so the expanded evidence is reviewable as a complete patch.

Arin016:
@reta @krugerLucas Makes sense, the comprehensive approach in #22152 with the test coverage and cross-config cleanup is the right way to address the review concern properly. Happy to close this in favor of that. Thanks for the quick feedback!

PR #22146, opened 2026-06-13
Make DataFusion metadata/statistics caches dynamic + scan-resistant (S3-FIFO)

alchemist51:
### Description

Makes the DataFusion analytics backend's **metadata** and **statistics** caches fully runtime-configurable, fixes their size limits being silently reverted, and replaces plain LRU with a scan-resistant eviction policy. Previously these cache settings were declared `Property.Dynamic` but were effectively read once at bootstrap, and the eviction policy could be polluted by a single wide query.

The PR is organized as three commits:

#### 1. Support dynamic metadata/statistics cache settings
The settings were read once at node bootstrap (`CacheUtils.createCacheConfig` → `DataFusionService.doStart()`) and never re-applied. This wires them through to the live native caches:
- **Size limits** (`datafusion.metadata/statistics.cache.size.limit`) now resize the live caches via a new `df_cache_manager_update_size_limit` FFM export (dispatching by cache type to the existing Rust resize functions) + `addSettingsUpdateConsumer`, instead of being restart-required.
- **Enable/disable** (`datafusion.metadata/statistics.cache.enabled`) toggles caches at runtime. Disabling clears the cache so its memory is freed immediately; caches are always constructed so a boot-disabled cache can be re-enabled later.
- Fixes the **statistics cache limit being clobbered to the 20MiB DataFusion default at startup** — `build_cache_manager_config` now sets `with_file_statistics_cache_limit`.
- Fixes a **u64 overflow in the memory-guard spill threshold** (`saturating_mul`) that panicked in debug/CI builds when the pool limit is `i64::MAX`.
- Disabled-cache lookups no longer count as cache misses.

#### 2. Preserve cache size limits across per-query config rebuilds
Each per-query path (`session_context`, `indexed_executor`, `query_executor`, plus the test session) rebuilds a `CacheManagerConfig` and re-attaches the shared caches. DataFusion's `CacheManager::try_new` then resets each cache's limit to the config's default (20MiB statistics / 50MiB metadata) unless the limit is re-supplied — silently shrinking the configured `*.cache.size.limit` on the hot path. Each rebuild now re-asserts both limits from the live `CacheManager` (`get_file_statistic_cache_limit` / `get_metadata_cache_limit`).

#### 3. Scan-resistant S3-FIFO eviction (new default for both caches)
Plain LRU is not scan-resistant: a single wide one-off query evicts the hot footer/statistics working set. This adds an **S3-FIFO** (Yang et al., SOSP'23) policy — small/main/ghost queues, quick demotion of one-hit-wonders, ghost re-admission — and makes it the default `eviction.type` for both caches (LRU/LFU still selectable). `small_ratio` defaults to `0.50` for an SLRU-like probation/protected split (matching ClickHouse's `cache_size_ratio`). To honor the policy on the metadata cache, `MutexFileMetadataCache` now owns its store and drives the pluggable policy directly (DataFusion's `DefaultFilesMetadataCache` is hardcoded LRU, so `eviction.type` was previously a no-op for metadata).

### Related Issues

N/A (sandbox/experimental analytics backend).

### Check List
- [x] Functionality includes unit tests — `DataFusionPluginSettingsTests` (Java: settings dynamic/node-scope, consumers are safe no-ops before service start), plus Rust unit + end-to-end tests for cache resize, enable/disable, and S3-FIFO scan resistance through both caches.
- [x] New functionality has been documented (internal KB).

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/cache.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/cache.rs
@@ -8,7 +8,7 @@
 
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, RwLock};
 
 use datafusion::execution::cache::cache_manager::{
     CachedFileMetadataEntry, FileMetadataCache, FileMetadataCacheEntry,
@@ -33,10 +33,14 @@ fn entry_size(entry: &CachedFileMetadataEntry) -> usize {
     entry.file_metadata.memory_size()
 }
 
-/// Inner store for [`MutexFileMetadataCache`]: the keyed entries plus running byte total.
+/// Inner store for [`MutexFileMetadataCache`]: the keyed entries.
+///
+/// Held behind an `RwLock` so concurrent reads (`get`/`contains_key`/`len`/`list_entries`) run in
+/// parallel; only mutations (`put`/`remove`/`clear`/eviction) take the write lock. The running
+/// byte total is tracked separately in the lock-free `memory_used` atomic (see the struct field),
+/// so size queries never touch this lock at all.
 struct MetaState {
     map: HashMap<Path, CachedFileMetadataEntry>,
-    memory_used: usize,
 }
 
 /// Metadata cache for parquet footers, keyed by file path.
@@ -48,10 +52,16 @@ struct MetaState {
 /// of metadata entries (and a one-off wide scan no longer pollutes the hot footer set
 /// under S3-FIFO). Staleness is still the caller's job via `CachedFileMetadataEntry::is_valid_for`.
 pub struct MutexFileMetadataCache {
-    state: Mutex<MetaState>,
+    state: RwLock<MetaState>,
     /// Pluggable eviction policy (interior-mutable; shared shape with the statistics cache).
     policy: Arc<Mutex<Box<dyn CachePolicy>>>,
     size_limit: AtomicUsize,
+    /// Running byte total of resident entries, tracked lock-free alongside the map. Updated on
+    /// every mutation while the `state` write lock is held (so it stays consistent with the map),
+    /// but read without any lock — size queries (`memory_used`, eviction's budget check) never
+    /// contend on `state`. Reads may momentarily trail an in-flight mutation; that's fine, the
+    /// total need not be exact (eviction re-checks in a loop).
+    memory_used: AtomicUsize,
     hit_count: AtomicUsize,
     miss_count: AtomicUsize,
     /// When false, the cache serves every lookup as a miss and drops all writes.
@@ -64,9 +74,10 @@ impl MutexFileMetadataCache {
     /// Create with an explicit eviction policy, byte limit, and initial enabled state.
     pub fn with_enabled(policy_type: PolicyType, size_limit: usize, enabled: bool) -> Self {
         Self {
-            state: Mutex::new(MetaState { map: HashMap::new(), memory_used: 0 }),
+            state: RwLock::new(MetaState { map: HashMap::new() }),
             policy: Arc::new(Mutex::new(create_policy(policy_type))),
             size_limit: AtomicUsize::new(size_limit),
+            memory_used: AtomicUsize::new(0),
             hit_count: AtomicUsize::new(0),
             miss_count: AtomicUsize::new(0),
             enabled: AtomicBool::new(enabled),
@@ -105,9 +116,9 @@ impl MutexFileMetadataCache {
     }
 
     pub fn clear_cache(&self) {
-        if let Ok(mut s) = self.state.lock() {
+        if let Ok(mut s) = self.state.write() {
             s.map.clear();
-            s.memory_used = 0;
+            self.memory_used.store(0, Ordering::Relaxed);
         }
         if let Ok(mut p) = self.policy.lock() {
             p.clear();
@@ -115,9 +126,10 @@ impl MutexFileMetadataCache {
         self.reset_stats();
     }
 
-    /// Current bytes held by metadata entries.
+    /// Current bytes held by metadata entries. Lock-free read of the running atomic total — may
+    /// momentarily trail an in-flight `put`/`remove`, which is fine (callers tolerate inexactness).
     pub fn memory_used(&self) -> usize {
-        self.state.lock().map(|s| s.memory_used).unwrap_or(0)
+        self.memory_used.load(Ordering::Relaxed)
     }
 
     pub fn update_cache_limit(&self, new_limit: usize) {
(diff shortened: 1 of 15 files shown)


sandbox/plugins/analytics-backend-datafusion/rust/src/cache.rs
```diff
@@ -78,6 +100,11 @@ impl MutexFileMetadataCache {
 
 impl CacheAccessor<Path, CachedFileMetadataEntry> for MutexFileMetadataCache {
     fn get(&self, k: &Path) -> Option<CachedFileMetadataEntry> {
+        // Disabled cache is always a miss (and holds nothing, since put drops writes).
+        if !self.is_enabled() {
+            self.miss_count.fetch_add(1, Ordering::Relaxed);
+            return None;
+        }
```
HarishNarasimhanK: Should we skip bumping miss_count here when the caches are disabled?

alchemist51:
## Benchmark: metadata-cache hit-rate, S3-FIFO vs LRU (TextBench)

Measured on a 1B-doc `textbench` index (OTel logs, ~30 parquet files, ~55MB footer-metadata
each → ~1.6GB working set), same data both arms, policy switched via
`datafusion.metadata.cache.eviction.type` (restart per arm). Workload = 5 simple textbench
queries + ClickHouse TextBench Q1–Q7 (`hasToken`/`hasAllTokens`/`hasAnyTokens` → PPL
`match(Body, …, operator=AND/OR)`), 3 rounds. `final_hr` = metadata-cache hit-rate from the
`_plugins/_analytics_backend_datafusion/stats` endpoint over the run.

### Final metadata-cache hit-rate by cache size

| cache size | LRU hit-rate | S3-FIFO hit-rate | Δ |
|-----------:|-------------:|-----------------:|----:|
| 256mb  | 0.161 | 0.527 | +0.366 |
| 512mb  | 0.232 | 0.615 | +0.383 |
| 768mb  | 0.313 | 0.657 | +0.344 |
| 1024mb | 0.550 | 0.747 | +0.197 |
| 1536mb | 0.686 | 0.914 | +0.228 |
| 2048mb | 1.000 | 1.000 | +0.000 |

S3-FIFO retains the hot footer set better at every cap where the working set doesn't fully
fit; the two converge at 2048mb where all ~1.6GB of footers fit. At 256mb the resident-entry
count was LRU=2 vs S3-FIFO=16 footers (LRU evicts by recency under the recurring wide scans).

### Per-query warm latency @ 512mb (ms, best-of-2)

| query | LRU | S3-FIFO |
|---|---:|---:|
| `… \| stats count()` | 27 | 25 |
| `… where ServiceName = 'checkout' \| stats count()` | 28 | 24 |
| `… where SeverityNumber >= 17 \| stats count()` | 2867 | 2550 |
| `… stats count() by ServiceName` | 10614 | 10321 |
| `… sort - Timestamp \| head 100` | 1292 | 1142 |
| Q2 `match(Body,'connection reset timeout',OR) … sev>=13` | 2849 | 2444 |
| Q3 `match(Body,'error exception failed timeout refused',OR) \| sort \| head 100` | 73681 | 75494 |
| Q4 `match(Body,'timeout') \| stats count()` | 24 | 23 |
| Q5 `match(Body,'error exception failed timeout refused',OR) \| stats count()` | 158 | 157 |
| Q6 `match(Body,'exception timeout failed',OR) \| stats count() by ServiceName` | 6737 | 6140 |
| Q7 `match(Body,'connection reset',AND) \| stats count() by ServiceName` | 3951 | 3481 |

S3-FIFO is faster on the cacheable token/group-by queries (up to ~1.17× here); the giant
full-corpus OR+sort (Q3, ~74s) is data-scan-bound, so cache policy is ~neutral there.

_Notes: latency is client-side wall-clock (`date +%s%3N` around the `curl`), so it includes
result transfer; numbers are raw, unmodified. The ClickHouse-Q1 (`ServiceName + match AND +
fields + head`) errored identically on both policies (a PPL translation issue), so it's
omitted above._

bharath-techie (approved):
LGTM

sandbox/plugins/analytics-backend-datafusion/rust/src/cache.rs
```diff
@@ -56,128 +105,325 @@ impl MutexFileMetadataCache {
...
-                    self.hit_count.fetch_add(1, Ordering::Relaxed);
-                } else {
-                    self.miss_count.fetch_add(1, Ordering::Relaxed);
+        // Disabled cache holds nothing (put drops writes); skip hit/miss accounting since
+        // the cache isn't participating.
+        if !self.is_enabled() {
+            return None;
+        }
+        let hit = {
+            let s = match self.state.lock() {
```
bharath-techie: lock in get ?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache.rs
```diff
@@ -56,128 +105,325 @@ impl MutexFileMetadataCache {
...
+        }
+        if let Ok(mut p) = self.policy.lock() {
+            p.clear();
         }
+        self.reset_stats();
+    }
+
+    /// Current bytes held by metadata entries.
+    pub fn memory_used(&self) -> usize {
+        self.state.lock().map(|s| s.memory_used).unwrap_or(0)
```
bharath-techie: do we need to acquire lock ? can we overall look at locks in this file and see how to get reads avoid contention

alchemist51: Nice catch, this is not needed and was able to remove this

sandbox/plugins/analytics-backend-datafusion/rust/src/cache.rs
alchemist51: Changed it to Read/Write locking, but internally the policy do a lock to update the queue so, even if I changed it. I was not able to see any perf diff, attaching micro benchmark run below.

alchemist51:
### Metadata-cache lock micro-benchmark — 3-arm (main vs PR vs PR+RwLock)

Contention micro-bench on `MutexFileMetadataCache` (working set 512 × 4KB entries, 500ms/point), throughput in **M ops/s**, threads 1→8:

**read_heavy (100% get — hot query path)**

| threads | main (`Mutex<DefaultFilesMetadataCache>`) | PR baseline (`Mutex` + S3-FIFO) | PR + RwLock/atomic |
|--------:|------:|------:|------:|
| 1 | 1.79 | 2.99 | 2.99 |
| 2 | 0.73 | 2.87 | 2.93 |
| 4 | 0.59 | 2.93 | 2.66 |
| 8 | **0.55** | **2.51** | **2.29** |

**write_heavy (100% put)**

| threads | main | PR baseline | PR + RwLock/atomic |
|--------:|------:|------:|------:|
| 1 | 1.43 | 1.10 | 1.12 |
| 8 | **0.50** | **0.77** | **0.97** |

**memory_used (size query)** — _main has no such method; PR-only comparison_

| threads | PR baseline (Mutex) | PR + atomic |
|--------:|------:|------:|
| 1 | 83.7 | 1,183 |
| 8 | **15.0 (collapses, 0.18×)** | **9,550 (scales, 8.07×)** |

**Takeaways**
- **The PR's custom cache already fixed read concurrency vs main**: main's `DefaultFilesMetadataCache` read throughput *collapses* under threads (1.79→0.55 M/s); the PR's cache holds ~2.5 M/s at 8 threads (**~4.5×** better) — independent of the lock change.
- **`memory_used` → lock-free atomic is a decisive win**: collapses under the Mutex (15 M/s @ 8 thr) → scales linearly (9,550 M/s, ~630× at 8 threads). Also lifts `write_heavy` (+26% @ 8 thr) since `enforce_limit` reads it on every put.
- **`RwLock` for the map is currently ~neutral on reads** (even slightly lower at 8 threads): `get()` still takes the policy `Mutex` for `on_access`, so the read path is bottlenecked on the *policy* lock, not the map lock. RwLock is the right foundation; making `on_access` lock-free is the follow-up that unlocks read scaling.

_ClickBench end-to-end A/B (hit-rate + warm latency) in progress; will post numbers when complete._

PR #22148, opened 2026-06-14
Fix writer lock leak on flush failure in refresh path

Bukhtawar:
## Summary
- Close the writer on flush failure in the refresh thread's `flushQueue` drain loop, preventing Lucene file lock leaks into the JVM-level `LOCK_HELD` set
- Complements #22021 which fixed the `CompositeWriter` constructor/close paths but missed this flush path

## Root Cause

When the refresh thread checks out writers from the pool and flushes them via `flushQueue`:

1. Writers are checked out of the pool → added to `flushQueue`
2. Refresh thread polls and calls `writer.flush()` on each
3. **If `flush()` throws** (e.g. Lucene `forceMerge` fails), the writer is NOT added to `toClose`
4. The writer becomes an orphan — not in the pool (checked out), not in `toClose`, not in `pendingWritersToClose`
5. `LuceneWriter`'s `IndexWriter` still holds the `NativeFSLock`, path stays in Lucene's static `LOCK_HELD`
6. Engine fails and restarts (same JVM), `writerGenerationCounter` re-initializes from `maxGenFromCommit`
7. Counter produces the same generation → same path → `LockObtainFailedException: Lock held by this virtual machine`

## Fix

Wrap the flush call in try/catch. On failure, `IOUtils.closeWhileHandlingException(writer)` ensures `LuceneWriter.close()` → `indexWriter.rollback()` → `NativeFSLock.close()` → `LOCK_HELD.remove(path)` before the exception propagates.

## Test plan
- [x] `./gradlew :server:compileJava` passes
- [ ] Existing `Composite*IT` suite
- [ ] Verify on staging cluster that was hitting the `LockObtainFailedException` (after JVM restart to clear existing leaked entries)

### DIFF
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
@@ -23,12 +23,14 @@ import org.apache.lucene.misc.store.HardlinkCopyDirectoryWrapper;
 import org.apache.lucene.search.Sort;
 import org.apache.lucene.store.Directory;
 import org.apache.lucene.store.MMapDirectory;
+import org.apache.lucene.store.NativeFSLockFactory;
 import org.opensearch.be.lucene.LuceneDataFormat;
 import org.opensearch.be.lucene.LuceneFieldFactoryRegistry;
 import org.opensearch.be.lucene.LuceneReader;
 import org.opensearch.be.lucene.merge.LuceneMerger;
 import org.opensearch.be.lucene.stats.LuceneShardStatsTracker;
 import org.opensearch.be.lucene.stats.LuceneStatsProvider;
+import org.opensearch.common.SuppressForbidden;
 import org.opensearch.common.annotation.ExperimentalApi;
 import org.opensearch.common.util.io.IOUtils;
 import org.opensearch.index.engine.dataformat.DataFormat;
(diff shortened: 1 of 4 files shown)


sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
```diff
@@ -456,6 +462,59 @@ public void close() throws IOException {
...
+                for (var it = lockHeld.iterator(); it.hasNext();) {
+                    Path lockPath = Path.of(it.next());
+                    if (lockPath.startsWith(realBase)) {
+                        it.remove();
+                        removed++;
+                    }
+                }
+            }
+            if (removed > 0) {
+                logger.warn(
```
mgodwan, line 516: Thanks for adding this. Can we add a metric as a follow up as well to help fix the close issues

mgodwan (approved):

mgodwan:
Would it make sense to directly clean all tracked writers from here: https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java#L99 ? This can be the common registry for clearing writers when engine close is called, keeping it agnostic of all failure modes DFAE may have?

cc: @rayshrey

Bukhtawar:
Re: metrics follow-up — will add a static node-level counter (`stale_locks_cleared_total`) exposed via `/_plugins/lucene/_nodes/{nodeId}/_stats` in a follow-up PR. Since `clearStaleLocks` runs before the per-shard tracker is initialized, it'll be a node-level metric rather than shard-level.

PR #22150, opened 2026-06-14
Fix QTF query then fetch race by refcounting ReaderContext

alchemist51:
### Description

The QTF (Query-Then-Fetch / late-materialization) reader handoff held a `GatedCloseable<Reader>` open across the query and fetch phases via `ReaderContext`, tracking in-use state with a **single-permit `AtomicBoolean`** acquired through `markInUse()` → `compareAndSet(false, true)`.

This made the query→fetch handoff racy:

- The query phase releases the flag only in its streaming `FragmentResources.close()`, which runs **after** `onComplete()` has already been sent to the coordinator (it's the closing brace of the try-with-resources that drains the stream).
- The coordinator dispatches the phase-B fetch as soon as it has the query rows, on a **separate thread/executor**.
- Under load, the fetch's `acquireContext()` → `markInUse()` could reach the shard **before** the query-phase release settled. `compareAndSet` then lost, `acquireContext` returned `null`, and the query failed with HTTP 500:

  ```
  No ReaderContext for queryId=<uuid> on [index][shard] — query phase missing or context expired
  ```

The context was never actually missing or expired — it was alive in the `ReaderContextStore` the entire time (and later swept as "expired" ~5 min later, confirming it was registered and never consumed). `acquireContext` returning `null` for both "absent" and "present-but-already-in-use" is what made a live context look missing.

#### Fix

Replace the single-permit `AtomicBoolean` with an `AbstractRefCounted`, mirroring core `org.opensearch.search.internal.ReaderContext` (which solves this exact query→fetch overlap with refcounting):

- `markInUse()` is now an additive `tryIncRef()` — it never fails on contention with another phase, so overlapping query + fetch references are a non-event (refcount goes 1→2→3→…→0).
- `markDone()` is a `decRef()`.
- `isExpired()` checks `refCount() > 1` (a phase still holds a use-reference on top of the base ref).
- `close()` drops the store's base reference; the underlying reader is closed only once **every** phase has released — so the reaper can never close a reader out from under an in-flight phase. This is strictly safer for the off-heap/native reader handed across the FFM boundary, where a use-after-free is a crash rather than a clean exception.

No change was needed in `ReaderContextStore`, `AnalyticsSearchService`, or `FragmentResources` — the call sites are identical; the behavior change comes entirely from `markInUse` becoming additive.

### Related Issues
N/A — found via staging investigation (`source=textbench | sort -Timestamp | head N` reproduced consistently under load).

### Check List
  - Flipped the two unit tests that encoded the old exclusive-lock semantics (`markInUse` is now additive; acquire-while-in-use now succeeds).
  - Added `testReaperWaitsForOverlappingQueryAndFetch` asserting the reaper waits for both overlapping query+fetch references before closing, even past keepAlive.
- [ ] API changes companion pull request created, if applicable. (No API change.)
- [ ] Public documentation issue/PR created, if applicable. (Internal sandbox plugin, no docs.)

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
@@ -227,6 +227,8 @@ public class AnalyticsSearchService implements AutoCloseable {
                         stats
                     );
                 } catch (Exception e) {
+                    // Query phase failed: no fetch will follow, so free the reader eagerly (no-op if already freed).
+                    readerContextStore.freeContext(request.getQueryId(), shard.shardId());
                     responseHandler.onFailure(e);
                 }
             });
@@ -308,7 +310,9 @@ public class AnalyticsSearchService implements AutoCloseable {
         assert assertFetchInvariants(readerContext, request.getQueryId());
         AnalyticsSearchBackendPlugin backend = backends.get(request.getBackendId());
         if (backend == null) {
+            // Fetch is terminal: free the reader eagerly (release this acquire, then free the base ref).
             readerContextStore.releaseContext(request.getQueryId(), shard.shardId());
+            readerContextStore.freeContext(request.getQueryId(), shard.shardId());
             responseHandler.onFailure(
                 new IllegalStateException(
                     "No backend registered for backendId="
@@ -345,7 +349,9 @@ public class AnalyticsSearchService implements AutoCloseable {
             return;
         } catch (Exception e) {
             if (rowIdVector != null) rowIdVector.close();
+            // Fetch is terminal: free the reader eagerly (release this acquire, then free the base ref).
             readerContextStore.releaseContext(request.getQueryId(), shard.shardId());
+            readerContextStore.freeContext(request.getQueryId(), shard.shardId());
             responseHandler.onFailure(new RuntimeException("Failed to execute fetch-by-row-ids on " + shard.shardId(), e));
             return;
         }
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
@@ -487,7 +487,7 @@ public class AnalyticsSearchService implements AutoCloseable {
      * fetch phase will reuse this reader. Derived from the {@link ShardScanInstructionNode} the
      * coordinator put in the plan (the same flag that makes the backend emit row ids).
      */
-    private static boolean requestsRowIds(List<InstructionNode> instructions) {
+    static boolean requestsRowIds(List<InstructionNode> instructions) {
         for (InstructionNode node : instructions) {
             if (node instanceof ShardScanInstructionNode scan) {
                 return scan.requestsRowIds();
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
@@ -343,7 +343,8 @@ public class AnalyticsSearchService implements AutoCloseable {
             resources = new FragmentResources(readerContextStore, readerContext, null, stream, null, rowIdVector, false);
         } catch (OpenSearchException e) {
             if (rowIdVector != null) rowIdVector.close();
-            readerContextStore.releaseContext(request.getQueryId(), shard.shardId());
+            // Fetch is terminal: free the reader eagerly.
+            readerContextStore.releaseAndFree(request.getQueryId(), shard.shardId());
             responseHandler.onFailure(e);
             return;
         } catch (Exception e) {
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
@@ -384,12 +384,12 @@ public class AnalyticsSearchService implements AutoCloseable {
     private FragmentResources startFragment(FragmentExecutionRequest request, ResolvedFragment resolved, IndexShard shard, Task task)
         throws IOException {
         GatedCloseable<Reader> gatedReader = resolved.readerProvider.acquireReader();
-        // A fetch phase follows only for QTF, detected from the ShardScan instruction's
-        // requestsRowIds flag. When it does, close() keeps the reader in the store for the fetch;
-        // otherwise close() frees it immediately instead of waiting for the reaper.
+        // A query that requested top-N docs (row-ids) will be followed by a fetch phase that reuses
+        // this reader. When it does, close() keeps the reader in the store for the fetch; otherwise
+        // close() frees it immediately instead of waiting for the reaper.
         // TODO: the coordinator (which knows the query shape) should tell us whether a fetch
         // follows, rather than us inferring it from the row-id signal here.
-        boolean fetchFollows = requestsRowIds(resolved.plan.getInstructions());
+        boolean requiresTopDocs = requestsRowIds(resolved.plan.getInstructions());
         ReaderContext readerContext = readerContextStore.createContext(request.getQueryId(), shard.shardId(), gatedReader);
         assert assertReaderInvariants(gatedReader, readerContext, request.getQueryId(), shard);
         SearchExecEngine<ShardScanExecutionContext, EngineResultStream> engine = null;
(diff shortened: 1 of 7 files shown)


bharath-techie:
Lets borrow what is already working from `server/src/main/java/org/opensearch/search/internal/ReaderContext.java` - otherwise LGTM.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -369,8 +377,16 @@ private FragmentResources startFragment(FragmentExecutionRequest request, Resolv
...
+        // FragmentResources holds a reference to the ReaderContext; for a multi-session (QTF) query
+        // phase close() releases it back to the store (the reaper closes after keepAlive); for a
+        // single-session query close() frees it immediately so the reader is not pinned until the
+        // reaper sweeps it. A query is multi-session only for QTF, which we detect from the
+        // ShardScan instruction's requestsRowIds flag (the same signal that tells the backend to
+        // emit __row_id__).
+        // TODO: ideally the coordinator (which knows the query shape) should tell us explicitly
+        // whether this is a single-session query or a QTF query, rather than us inferring it from
+        // the row-id signal here.
+        boolean singleSession = !requestsRowIds(resolved.plan.getInstructions());
```
bharath-techie: i don't think single session means this - anything other than PIT and scroll needs to be considered as single session.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -303,7 +306,9 @@ private void drainFetchByRowIds(
         assert assertFetchInvariants(readerContext, request.getQueryId());
         AnalyticsSearchBackendPlugin backend = backends.get(request.getBackendId());
         if (backend == null) {
+            // Fetch is terminal: free the reader eagerly (release this acquire, then free the base ref).
             readerContextStore.releaseContext(request.getQueryId(), shard.shardId());
```
mch2: What's the difference here?

alchemist51: releaseContext decRef the count. Free Context will now free it up as well post the fetch call.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
alchemist51: Good catch! There is a bit of semantics difference now since in vanilla Opensearch, it's always `QUERY_THEN_FETCH` while for us we need to handle the fetch case also. Have created a new flag for it

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
mch2: This should happen automatically with refcount 0?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/FragmentResources.java
```diff
@@ -37,15 +40,26 @@ public final class FragmentResources implements AutoCloseable {
      * pull memory out from under the FFM call.
      */
     private final BigIntVector rowIdVector;
+    /** True when a fetch phase will reuse this reader (QTF query phase); close then keeps it for the fetch. */
+    private final boolean fetchFollows;
```
expani: Can't we use the non-null `rowIdVector` to determine that a fetch will follow instead of a separate flag ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/FragmentResources.java
```diff
@@ -37,15 +40,26 @@ public final class FragmentResources implements AutoCloseable {
...
+
+    /**
+     * Whether this reader serves a single request rather than living across many (scroll/PIT style).
+     * Always true today — analytics-engine has no scroll/PIT-equivalent that outlives one request.
+     * TODO: the coordinator (which knows the query shape) should set this per request rather than it
+     *  defaulting to true; a future multi-session reader would not be freed on close here.
+     */
+    private final boolean singleSession = true;
```
expani: I think we can defer introducing this properly with other refactoring for Scroll/PIT like queries.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/FragmentResources.java
alchemist51: No that is a fetch phase property, while you need to take the call in the query phase itself.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/FragmentResources.java
alchemist51: Ack will remove it for now

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/ReaderContext.java
```diff
@@ -18,34 +21,40 @@
...
     private final String queryId;
     private final ShardId shardId;
     private final GatedCloseable<Reader> gatedReader;
-    private final AtomicBoolean inUse = new AtomicBoolean(false);
     private final AtomicLong lastAccessTime;
     private volatile long keepAliveMillis;
-    private volatile boolean closed;
+    private final AtomicBoolean closed = new AtomicBoolean(false);
+    // Use-count: starts at 1 (base ref), reader closes when it hits 0.
+    private final AbstractRefCounted refCounted;
```
expani, line 44: I am assuming the 2 ref counts is one for query phase and one for fetch phase ? If so, ref counting seems a bit excessive here.

Can't we ensure that Query phase will not close the reader when there is fetch involved instead ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/FragmentResources.java
expani: I meant that rowIdVector being non-null is already the replacement for this new boolean flag since it will be null for non QTF cases.

See `AnalyticsSearchService#drainFetchByRowIds`

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/ReaderContext.java
alchemist51: I think this in inline with the Opensearch current ReaderContext. Later on when we are extending this for PIT or scroll, we will need the RefCounter so it will help there as well.

alchemist51: Also we need a marker to say how many phases/queries are using the context for the reaper. We might see edge cases in that as well

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/FragmentResources.java
alchemist51: But `AnalyticsSearchService#drainFetchByRowIds` is in fetch phase right? 

```
    /**
     * Acquires the per-shard {@link ReaderContext}, materialises the rowId vector, invokes
     * the backend, drains the stream into {@code responseHandler}, and releases all resources
     * in a single try-with-resources scope. Runs on the caller's executor — exhausting the
     * iterator here lets the native engine apply backpressure when the channel is slow.
     */
```

How will you know that when the Query Phase lands on the node since the context is created in QueryPhase?

alchemist51: https://github.com/opensearch-project/OpenSearch/pull/22150#discussion_r3411222915

are you saying to not close and take the call in fetch phase? I think we should keep the logic simple, create a context, keep refCounters , clean it when:
1. RefCounter is zero
2. Idle for 5mins

We will make the logic hard to follow and open to bugs if we start keeping context open or close based on our implementation.

expani: >are you saying to not close and take the call in fetch phase?

Yes but only for the happy path. Are you saying that we will always close using the Reaper instead ? 

Then, we keep context open for 5 mins instead of actively closing when fetch is done. This will also not be optimal for High QPS 


>RefCounter is zero
> open to bugs if we start keeping context open or close based on our implementation.

So, if someone misses to do proper ref counting you are still open to bugs ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/ReaderContext.java
bharath-techie: +1, would like to use the patterns that are well tested unless there is a strong need otherwise.

bharath-techie (approved):
LGTM

PR #22151, opened 2026-06-14
Harden force merge, VSR dedup, and refreshLock safety in DataFormatAwareEngine

mgodwan:
Summary

  This PR hardens several correctness and concurrency issues in the DataFormat-aware engine's merge and indexing paths:

  Bug Fixes:

  - VSRManager multi-value dedup — The dedup Set was created but never populated (add() was missing), making the duplicate-field
  rejection dead code. Documents with repeated fields would silently overwrite Arrow vector values.
  - Force merge fire-and-forget — forceMerge dispatched tasks to the thread pool but never waited for completion, so the transport
  layer returned success before merges finished. Now runs synchronously on the caller's FORCE_MERGE thread.

  Improvements:

  - Force merge serialization — Semaphore(1) ensures only one force merge runs per shard at a time; concurrent callers block
  rather than throwing IllegalStateException.
  - Flush after force merge — Calls flush(true, true) when the flush parameter is set (default), ensuring all changes are visible. This is in line with existing guarantees. 
  - Reduced refreshLock fragility — Extracted runMerge() as a single method encapsulating the doMerge→apply/cleanup lifecycle.
  Both background and force merges funnel through it, keeping the lock release contract in one place.
  - RowIdFieldMapper — Registered as a MetadataFieldMapper so user-provided __row_id__ values in documents are rejected at parse
  time via inherited parseCreateField().
  - Debug logging — Safety-net refreshLock unlock now logs at DEBUG level for diagnosing merge-failure paths in production.

  Testing

  - Existing MergeTests and VSRManagerTests cover basic flows
  - Integration validated via composite engine suite (CompositeMergeIT)

### Description

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
+++ b/sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
@@ -18,6 +18,7 @@ import org.opensearch.index.IndexSettings;
 import org.opensearch.index.engine.dataformat.DocumentInput;
 import org.opensearch.index.engine.dataformat.RowIdMapping;
 import org.opensearch.index.mapper.MappedFieldType;
+import org.opensearch.index.mapper.MapperParsingException;
 import org.opensearch.nativebridge.spi.ArrowExport;
 import org.opensearch.parquet.ParquetDataFormatPlugin;
 import org.opensearch.parquet.bridge.NativeParquetWriter;
@@ -33,6 +34,9 @@ import org.opensearch.parquet.writer.ParquetDocumentInput;
 import org.opensearch.threadpool.ThreadPool;
 
 import java.io.IOException;
+import java.util.Collections;
+import java.util.IdentityHashMap;
+import java.util.Set;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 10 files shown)


sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
```diff
@@ -204,8 +215,14 @@ public void addDocument(ParquetDocumentInput doc) throws IOException {
...
         }
         ManagedVSR activeVSR = managedVSR.get();
+        Set<MappedFieldType> dedup = Collections.newSetFromMap(new IdentityHashMap<>());
         for (FieldValuePair pair : doc.getFinalInput()) {
             MappedFieldType fieldType = pair.getFieldType();
+            if (dedup.add(fieldType) == false) {
+                throw new MapperParsingException(
+                    "Cannot accept multiple values for field: [" + fieldType.name() + "] of type: [" + fieldType.typeName() + "]."
+                );
+            }
```
Bukhtawar, line 225: Will this result in a document write failure?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1201,7 +1221,13 @@ public void forceMerge(
...
     ) throws EngineException, IOException {
-        mergeScheduler.forceMerge(1);
+        if (upgrade || upgradeOnlyAncientSegments || onlyExpungeDeletes) {
+            return;
+        }
+        mergeScheduler.forceMerge(maxNumSegments);
+        if (flush) {
+            flush(true, true);
+        }
     }
```
Bukhtawar, line 1231: Sorry I don't see a flush before the merge?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
mgodwan: Yes

mgodwan: Given we don't have a construct in mappings to allow/reject arrays, this is handled during ingestion itself rather than create index. I believe we should make our schema definition (i.e. mappings) stricter as well so that we can allow the array on an opt-in basis rather than default (esp. for formats like parquet)

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
mgodwan: This is meant for after only. Let me update the PR description as well

Bukhtawar (approved):
thanks lgtm

Issue #1588, opened 2021-11-19
Jigsaw integration


rishabhmaurya:
# Jigsaw integration OpenSearch

[RFC: Modularization in OpenSearch](https://quip-amazon.com/4KCYAPfSHwkP) gives some background on why we should consider modularizing OpenSearch codebase.  
**This proposal addresses following problems related to OpenSearch codebase and its plugins framework -** 

1. Strong encapsulation: the server module in OpenSearch is weakly encapsulated and all packages are exported both for direct access and through reflections. 
2. Lack of separation of concern and monolith module: The server modules needs to be split down into components depending on the use-case they are solving in the system. Also, the plugin developer or consumer of these modules shouldn’t be concerned about the business logic or changes in unrelated modules. For e.g. AnalyzerPlugin should only depend on modules of its concern, maybe - analyzer-module, search-module, index-module etc. 

Note: this proposal assumes breaking changes are allowed in OpenSearch 2.0 and that’s when these changes can go in all at once.

**Solving above two problems using Java 9 modules will inherently address other problems such as -** 

1. OpenSearch version agnostic of individual plugins versions - Plugins will depend on the versions of modules on which they depend on and not the OpenSearch version. If some module doesn’t have any major version bump in OpenSearch release, the dependent plugin should continue to run as it is.  
2. Code hygiene and better maintainability: modular codebase is more maintainable and easy to contribute to as it will have better visualization and dependency graph, which would make more sense to the opensource contributors who want to focus on specific part. 
3. Unblock customers using Jigsaw modular architecture: Customers who have already migrated to Java 9 modules are currently blocked as OpenSearch rest client doesn’t supports Java 9 modules. More details here - https://github.com/elastic/elasticsearch/issues/38299

**Benefits of  Java 9 modules?** 

* Strong encapsulation: Java 9 modules enables strong encapsulation of internal classes of modules if the modules are properly configured. By default, all public types are not exported and needs to be explicitly exported by exporting corresponding packages. 
* Reliable configuration: The dependencies missing can be found at compile or launch time instead of delaying it to the runtime of application. 
* Security: Strong encapsulation ensures hiding internal packages not meant to be exposed to consumers. 
* Better migration support: Concepts like automatic, named, unnamed modules are designed to make it possible for large codebases such as OpenSearch to migrate to modular architecture incrementally. More details in following sections. All modules are backward compatible with JDK-8 as Java runtime environment, so applications running on JDK-8, will not break. 
[...]

PR #22152, opened 2026-06-14
[Arrow] Use ALL-UNNAMED for java.nio add-opens

krugerLucas:
### Description

The default distribution currently includes this JVM option:

```text
21-:--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED
```

When the `org.apache.arrow.memory.core` named module is not present at JVM startup, the JVM emits:

```text
WARNING: Unknown module: org.apache.arrow.memory.core specified to --add-opens
```

This happens even on nodes that do not install Arrow plugins, because the option is part of the global distribution `jvm.options`.

This PR keeps the required `java.nio` access for classpath/plugin code by targeting `ALL-UNNAMED` only:

```text
21-:--add-opens=java.base/java.nio=ALL-UNNAMED
```

The important compatibility point is that the Arrow classes used by the OpenSearch plugins are loaded from the unnamed module space in the current plugin/test runtime. To make that explicit, this PR adds an internal cluster test that checks the Arrow API class, runtime allocator class, and plugin allocator class are all unnamed modules, then performs a real Arrow buffer allocation.

### What changed

- Update the global distribution `jvm.options` entry to use `ALL-UNNAMED` only.
- Align the same Arrow `java.nio` add-opens option in Arrow Flight, example, sandbox, and analytics QA Gradle configurations.
- Remove duplicate named-module `add-opens` entries where an `ALL-UNNAMED` entry already existed.
- Update the Arrow Flight README manual startup guidance.
- Add coverage in `NativeAllocatorBoundaryIT` to prove the relevant Arrow/runtime/plugin classes are unnamed modules and that Arrow allocation still works.

### Context

This is related to #22132 and is a more complete variant of the direction proposed in #22145. The extra changes are intended to address the review concern that Arrow Flight still needs the JVM option and should continue to function after removing the named module target.

### Test plan

- [x] Minimal JVM reproduction:

```bash
java --add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED -version
# emits: WARNING: Unknown module: org.apache.arrow.memory.core specified to --add-opens

java --add-opens=java.base/java.nio=ALL-UNNAMED -version
# no Unknown module warning
```

- [x] Arrow unit tests:

```bash
./gradlew :plugins:arrow-base:test :plugins:arrow-flight-rpc:test --no-build-cache -Dbuild.snapshot=false
```

- [x] Arrow Flight internal cluster tests:

```bash
./gradlew :plugins:arrow-flight-rpc:internalClusterTest --no-build-cache -Dbuild.snapshot=false
```

This covers the current `FlightClient.getStream(...)` to `ArrowFlightProducer.getStream(...)` path, native allocator behavior, metadata streaming, batch streaming, and stream search/aggregation coverage.

- [x] Sandbox Arrow/DataFusion/Parquet consumers:

```bash
./gradlew \
  :sandbox:plugins:analytics-backend-datafusion:test \
  :sandbox:plugins:parquet-data-format:test \
  :sandbox:plugins:analytics-engine:test \
  --no-build-cache -Dbuild.snapshot=false -Dsandbox.enabled=true
```

- [x] Analytics coordinator QA:

```bash
./gradlew :sandbox:qa:analytics-engine-coordinator:internalClusterTest \
  --no-build-cache -Dbuild.snapshot=false -Dsandbox.enabled=true
```

- [x] Stream transport example default run:

```bash
./gradlew :example-plugins:stream-transport-example:internalClusterTest \
  --no-build-cache -Dbuild.snapshot=false
```

The default run succeeds; the example tests are annotated with `@AwaitsFix` and are skipped by default. With `-Dtests.awaitsfix=true`, the native Arrow example tests pass, while the non-native example still fails because it starts `FlightStreamPlugin` without `ArrowBasePlugin`. That failure is pre-existing and unrelated to this JVM option change.

- [x] Arrow plugin smoke test using packaged distribution config:

```bash
./gradlew :qa:smoke-test-arrow-plugins:integTest --no-build-cache -Dbuild.snapshot=false
```

The test cluster started with the packaged option:

```text
--add-opens=java.base/java.nio=ALL-UNNAMED
```

and logs/test reports did not contain:

```text
Unknown module: org.apache.arrow.memory.core specified to --add-opens
```

- [x] Targeted precommit/format checks:

```bash
./gradlew \
  :plugins:arrow-flight-rpc:spotlessJavaCheck \
  :plugins:arrow-flight-rpc:precommit \
  :example-plugins:stream-transport-example:precommit \
  :qa:smoke-test-arrow-plugins:precommit \
  --no-build-cache -Dbuild.snapshot=false

./gradlew \
  :sandbox:plugins:analytics-backend-datafusion:precommit \
  :sandbox:plugins:analytics-engine:precommit \
  :sandbox:plugins:composite-engine:precommit \
  :sandbox:plugins:parquet-data-format:precommit \
  :sandbox:qa:analytics-engine-coordinator:precommit \
  --no-build-cache -Dbuild.snapshot=false -Dsandbox.enabled=true
```

### Related Issues

Related to #22132
Related to #22145

### Check List
- [x] API changes companion pull request: not applicable.
- [x] Public documentation issue/PR: not applicable.

### DIFF
--- a/distribution/src/config/jvm.options
+++ b/distribution/src/config/jvm.options
@@ -85,7 +85,7 @@ ${error.file}
 23:-XX:CompileCommand=dontinline,java/lang/invoke/MethodHandle.asTypeUncached
 
 21-:-javaagent:agent/opensearch-agent.jar
-21-:--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED
+21-:--add-opens=java.base/java.nio=ALL-UNNAMED
 
 # For cases with high memory-mapped file counts, a lower value can improve stability and
 # prevent issues like "leaked" maps or performance degradation. A value of 1 effectively
--- a/plugins/arrow-flight-rpc/README.md
+++ b/plugins/arrow-flight-rpc/README.md
@@ -33,7 +33,7 @@ opensearch.experimental.feature.transport.stream.enabled: true
 -Dio.netty.noUnsafe=false
 -Dio.netty.tryUnsafe=true
 -Dio.netty.tryReflectionSetAccessible=true
---add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED
+--add-opens=java.base/java.nio=ALL-UNNAMED
 ```
 
 3. Install and run the plugin manually
--- a/plugins/arrow-flight-rpc/build.gradle
+++ b/plugins/arrow-flight-rpc/build.gradle
@@ -84,7 +84,7 @@ test {
   systemProperty 'io.netty.noUnsafe', 'false'
   systemProperty 'io.netty.tryUnsafe', 'true'
   systemProperty 'io.netty.tryReflectionSetAccessible', 'true'
-  jvmArgs += ["--add-opens", "java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED"]
+  jvmArgs += ["--add-opens", "java.base/java.nio=ALL-UNNAMED"]
 }
 
 internalClusterTest {
@@ -92,7 +92,7 @@ internalClusterTest {
   systemProperty 'io.netty.noUnsafe', 'false'
   systemProperty 'io.netty.tryUnsafe', 'true'
   systemProperty 'io.netty.tryReflectionSetAccessible', 'true'
-  jvmArgs += ["--add-opens", "java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED"]
+  jvmArgs += ["--add-opens", "java.base/java.nio=ALL-UNNAMED"]
 }
 
 spotless {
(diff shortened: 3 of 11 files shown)


distribution/src/config/jvm.options
```diff
@@ -85,7 +85,7 @@ ${error.file}
 23:-XX:CompileCommand=dontinline,java/lang/invoke/MethodHandle.asTypeUncached
 
 21-:-javaagent:agent/opensearch-agent.jar
-21-:--add-opens=java.base/java.nio=org.apache.arrow.memory.core,ALL-UNNAMED
+21-:--add-opens=java.base/java.nio=ALL-UNNAMED
```
reta, line 88: @rishabhmaurya wondering what do you think about going just with `--add-opens=java.base/java.nio=ALL-UNNAMED` across the board? It is unlikely we are getting JPMS compliant OpenSearch any time soon [1] etc.

[1] https://github.com/opensearch-project/OpenSearch/issues/1588

rishabhmaurya: right! this change looks good to me.

rishabhmaurya:
LGTM! thanks @krugerLucas 
We have explicit setup [steps](https://github.com/opensearch-project/OpenSearch/blob/main/plugins/arrow-flight-rpc/README.md) to be performed for arrow flight plugin to be functional with other jvm properties too, so this is the right thing to do to remove it from global jvm options.

rishabhmaurya (approved):

PR #22155, opened 2026-06-15
Fix ObjectMapper.deriveSource to skip empty nested objects

### LINKED ISSUE #21886: [BUG] [Derived source] ObjectMapper adds empty objects {} into reconstructed _source

Richard-coco:
### Description
Fix `ObjectMapper.deriveSource` to skip writing empty objects `{}` 
when a document has no values for nested object fields.

Previously, `ObjectMapper.deriveSource` would always call 
`builder.startObject(simpleName())` and `builder.endObject()`, 
resulting in empty `{}` objects being added to the reconstructed 
`_source` even when the document contained no data for those fields.

This fix uses a temporary `XContentBuilder` to collect child field 
content first, and only writes the object to the main builder if 
there is actual content.

### Related Issues
Resolves #21886

### Check List
- [ ] API changes companion pull request, if applicable.
- [ ] Public documentation issue/PR, if applicable.

### DIFF
--- a/server/src/main/java/org/opensearch/index/mapper/ObjectMapper.java
+++ b/server/src/main/java/org/opensearch/index/mapper/ObjectMapper.java
@@ -40,9 +40,12 @@ import org.opensearch.cluster.metadata.IndexMetadata;
 import org.opensearch.common.Explicit;
 import org.opensearch.common.Nullable;
 import org.opensearch.common.annotation.PublicApi;
+import org.opensearch.core.common.bytes.BytesArray;
+import org.opensearch.core.common.bytes.BytesReference;
 import org.opensearch.common.collect.CopyOnWriteHashMap;
 import org.opensearch.common.logging.DeprecationLogger;
 import org.opensearch.common.settings.Settings;
+import org.opensearch.common.xcontent.XContentFactory;
 import org.opensearch.common.xcontent.support.XContentMapValues;
 import org.opensearch.core.xcontent.ToXContent;
 import org.opensearch.core.xcontent.XContentBuilder;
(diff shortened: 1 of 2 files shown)


Richard-coco:
I noticed that PR #21889 addresses the same issue with a more elegant solution using LazyXContentBuilder. I'll close this PR to avoid duplication. Thanks!

PR #22157, opened 2026-06-15
fix: report bloom filter row group pruning metrics correctly

cocosz:
# Description

The `row_groups_pruned_bloom_filter` metric in the query profile always reported `0`, even when bloom filter pruning was actively skipping row groups during Parquet scans.

Two bugs existed in `collect_metrics` (`api.rs`):

**1. `PruningMetrics::as_usize()` always returns `0`**

`PruningMetrics` is a compound DataFusion metric type that stores two independent atomic counters: `pruned` and `matched`. It cannot be collapsed into a single number. DataFusion's `as_usize()` implementation explicitly returns `0` for this variant:

```rust
// From DataFusion source:
Self::PruningMetrics { .. } => 0,
```

The old implementation called `as_usize()` uniformly across all metric types. As a result, all pruning-related metrics — bloom filter, row group statistics, and page index — were always serialized as `0` in the profile output.

**2. Metrics from multiple partitions were overwritten instead of aggregated**

DataFusion registers independent metric instances for each scan partition. With `N` partitions, `collect_metrics` iterates over all `N` instances of the same metric name. The old implementation called:

```rust
map.insert(name, value);
```

This unconditionally replaced the existing entry, retaining only the last partition's value.

# Fix

Match on `MetricValue` variant to handle `PruningMetrics` explicitly:

- For `PruningMetrics`: extract `.pruned()` and `.matched()` directly and emit as `<name>_pruned` and `<name>_matched`.
- For all other types: use `as_usize()` as before.

Both paths accumulate values across partitions (read-then-add) instead of overwriting.

# Verification

With `index: false` and Parquet bloom filters enabled on the `URL` field:

| Query | `bloom_pruned` | `bloom_matched` |
|-------|--------------|----------------|
| `where URL = 'http://kinopoisk.ru'` | 7 | 98 |
| `where URL like '%kinopoisk%'` | 0 | 106 |

`LIKE` predicates cannot be evaluated by bloom filters — they require exact equality. The `bloom_pruned = 0` on the `LIKE` query confirms the metric correctly reflects actual pruning behavior rather than always returning `0`.

# Testing

Verified on ClickBench (100M rows) with `index: false` and `parquet_bloom_filter_enabled: true` on: `WatchID`, `UserID`, `Referer`, `URL`, `Title`, `OriginalURL`, `SearchPhrase`, `SocialSourcePage`, `OpenstatAdID`.

```

| Field | Matching Rows | Row Groups Pruned | Row Groups Matched | Execution Node | Time (ms) |
|-------|--------------:|------------------:|-------------------:|----------------|----------:|
| WatchID | 2 | 4 | 102 | DataSourceExec | 1,665 |
| UserID | 29,097 | 72 | 27 | DataSourceExec | 157 |
| URL | 1,625,250 | 7 | 98 | DataSourceExec | 2,643 |
| SearchPhrase | 16,746 | 44 | 62 | DataSourceExec | 486 |
| OpenstatAdID | 5,527 | 86 | 1 | DataSourceExec | 16 |
```


bharath-techie (approved):
LGTM

PR #22164, opened 2026-06-15
[analytics-engine] Fix TopK sort key mismatch

sandeshkr419:
## Summary
  - Fix TopK rewriter pushing wrong sort column to shards. It was sorting by group key instead of measure due to PPL's column-swap Project between Sort and Aggregate
  - Add `NOT_EQUALS` to lucene delegation blocklist defaults
  - ~Add `PlanShapeITs` with golden-file plan assertions for all 42 ClickBench queries (topk on/off)~

  ## Root Cause
  Commit `03283ee45b6` removed `SORT_PROJECT_TRANSPOSE`, leaving PPL's column-swap Project between Sort and FINAL Aggregate. The TopK rewriter copied the Sort's column index verbatim without remapping through this Project — causing shards to sort alphabetically by key, truncating high-count results.

  ## Fix
  `findFinalAgg` now captures the intermediate Project. Before pushing the Sort to shards, collation indices are remapped through the Project's inverse mapping.

  ## Test plan
  - [ ] 25 unit tests in `TopKRewriterPlanShapeTests` (edge cases, PPL swap shapes, AVG exclusion)

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
@@ -88,6 +88,9 @@ public final class OpenSearchTopKRewriter {
         // PPL emits measures-first output via a column-swapping Project (e.g. Project(c=$1, key=$0)).
         // The outer Sort's indices reference the Project's output; the shard Sort must reference
         // the PARTIAL Aggregate's output (group-keys first, measures second).
+        // If the Project contains computed expressions (literals, casts, arithmetic) rather than
+        // simple column references, we bail — the sort key has no direct mapping to the aggregate
+        // output and TopK cannot be safely applied.
         RelCollation adjustedCollation = sort.getCollation();
         if (match.intermediateProject != null) {
             List<RelFieldCollation> remapped = new ArrayList<>();
@@ -211,6 +214,17 @@ public final class OpenSearchTopKRewriter {
         return findFinalAgg(node, null);
     }
 
+    /**
+     * Walks toward the FINAL aggregate, capturing at most one intermediate Project.
+     * The {@code seenProject == null} guard assumes CoreRules.PROJECT_MERGE has collapsed
+     * adjacent Projects during RBO — if that rule is ever removed, multiple Projects may
+     * appear and this method will skip the second one (returning null for the project field),
+     * which safely disables the collation remapping (TopK still fires with the raw collation,
+     * correct when no reordering exists between the two).
+     *
+     * <p>Multi-input nodes (joins, unions) terminate the walk — TopK only applies to
+     * single-path Sort → Aggregate chains.
+     */
     private static PathToFinal findFinalAgg(RelNode node, OpenSearchProject seenProject) {
         if (node instanceof OpenSearchAggregate agg && agg.getMode() == AggregateMode.FINAL) {
             return new PathToFinal(seenProject, agg);
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
@@ -14,6 +14,7 @@ import org.apache.calcite.rel.RelFieldCollation;
 import org.apache.calcite.rel.RelNode;
 import org.apache.calcite.rel.type.RelDataType;
 import org.apache.calcite.rex.RexBuilder;
+import org.apache.calcite.rex.RexInputRef;
 import org.apache.calcite.rex.RexLiteral;
 import org.apache.calcite.rex.RexNode;
 import org.apache.calcite.sql.type.SqlTypeName;
@@ -83,10 +84,27 @@ public final class OpenSearchTopKRewriter {
             true
         );
 
-        RelNode sortInput = partial;
+        // Remap collation through the intermediate Project (if any) between Sort and FINAL Aggregate.
+        // PPL emits measures-first output via a column-swapping Project (e.g. Project(c=$1, key=$0)).
+        // The outer Sort's indices reference the Project's output; the shard Sort must reference
+        // the PARTIAL Aggregate's output (group-keys first, measures second).
         RelCollation adjustedCollation = sort.getCollation();
+        if (match.intermediateProject != null) {
+            List<RelFieldCollation> remapped = new ArrayList<>();
+            for (RelFieldCollation fc : sort.getCollation().getFieldCollations()) {
+                RexNode expr = match.intermediateProject.getProjects().get(fc.getFieldIndex());
+                if (expr instanceof RexInputRef ref) {
+                    remapped.add(new RelFieldCollation(ref.getIndex(), fc.getDirection(), fc.nullDirection));
+                } else {
+                    return Optional.empty();
+                }
+            }
+            adjustedCollation = RelCollations.of(remapped);
+        }
 
-        if (hasEngineNativeMergeMeasure(partial, sort.getCollation())) {
+        RelNode sortInput = partial;
+
+        if (hasEngineNativeMergeMeasure(partial, adjustedCollation)) {
             RelDataType partialRowType = partial.getRowType();
             int fieldCount = partialRowType.getFieldCount();
             int groupCount = partial.getGroupSet().cardinality();
@@ -99,7 +117,7 @@ public final class OpenSearchTopKRewriter {
             }
 
             List<RelFieldCollation> newCollations = new ArrayList<>();
-            for (RelFieldCollation fc : sort.getCollation().getFieldCollations()) {
+            for (RelFieldCollation fc : adjustedCollation.getFieldCollations()) {
                 int sortIdx = fc.getFieldIndex();
                 if (sortIdx >= groupCount && AggregateFunction.isEngineNativeMerge(partial.getAggCallList().get(sortIdx - groupCount))) {
                     AggregateFunction aggFunc = AggregateFunction.fromSqlAggFunction(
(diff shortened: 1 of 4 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -83,10 +84,27 @@ public static Optional<RelNode> rewrite(RelNode root, PlannerContext context) {
...
+        // the PARTIAL Aggregate's output (group-keys first, measures second).
         RelCollation adjustedCollation = sort.getCollation();
+        if (match.intermediateProject != null) {
+            List<RelFieldCollation> remapped = new ArrayList<>();
+            for (RelFieldCollation fc : sort.getCollation().getFieldCollations()) {
+                RexNode expr = match.intermediateProject.getProjects().get(fc.getFieldIndex());
+                if (expr instanceof RexInputRef ref) {
+                    remapped.add(new RelFieldCollation(ref.getIndex(), fc.getDirection(), fc.nullDirection));
+                } else {
+                    return Optional.empty();
```
expani, line 102: What happens if there is a literal in the Project that's not a column reference a.k.a. `RexInputRef` ?

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/TopKRewriterPlanShapeTests.java
```diff
@@ -356,4 +358,320 @@ private PlannerContext contextWithOversampling(double factor) {
...
+    public void testRewrite_pplShape_q11_dcWithSwap_sortKeyRemapped() {
+        RelOptTable table = mockTable("test_index", "status", "size");
+        RelNode scan = stubScan(table);
+        RelNode result = runPlanner(
+            buildPplSwapPlan(ImmutableBitSet.of(0), List.of(approxCountDistinctCall(scan)), 10),
+            contextWithOversampling(2.0)
+        );
+        assertPlanShape(
+            """
+                OpenSearchSort(sort0=[$0], dir0=[DESC], fetch=[10000], viableBackends=[[mock-parquet]])
```
expani, line 424: What does the Datafusion Physical plan look like here ? Does the optimizer collapses these to one ? 

We can use `SORT_REMOVE_REDUNDANT` or a similar CoreRule from Calcite to ensure there's only one of these left with a fetch of `10` Maybe add a TODO if it's too complicated to do now.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -183,15 +201,24 @@ private static SortAboveFinal findSortAboveFinal(RelNode node) {
...
-        if (node.getInputs().size() == 1) return findFinalAgg(node.getInputs().get(0));
+    private static PathToFinal findFinalAgg(RelNode node) {
+        return findFinalAgg(node, null);
+    }
+
+    private static PathToFinal findFinalAgg(RelNode node, OpenSearchProject seenProject) {
+        if (node instanceof OpenSearchAggregate agg && agg.getMode() == AggregateMode.FINAL) {
+            return new PathToFinal(seenProject, agg);
+        }
+        if (node instanceof OpenSearchProject proj && seenProject == null) {
```
expani, line 232: `seenProject == null` 

This is depending on the in-built Calcite rule of `CoreRules.PROJECT_MERGE` which is applied during RBO. 

Can we write a UT to ensure that someone removing it in future knows that the rewrite depends on it ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -183,15 +201,24 @@ private static SortAboveFinal findSortAboveFinal(RelNode node) {
...
+            return findFinalAgg(proj.getInput(), proj);
+        }
+        if (node.getInputs().size() == 1) return findFinalAgg(node.getInputs().get(0), seenProject);
```
expani, line 235:
```
node.getInputs().get(0)
```
Are we still rejecting Joins/Unions from TopK rewrite ?

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/TopKRewriterPlanShapeTests.java
```diff
@@ -356,4 +358,320 @@ private PlannerContext contextWithOversampling(double factor) {
...
+        for (int i = 0; i < measureCount; i++) {
+            swapExprs.add(rexBuilder.makeInputRef(agg, groupCount + i));
+            swapNames.add(agg.getRowType().getFieldNames().get(groupCount + i));
+        }
+        for (int i = 0; i < groupCount; i++) {
+            swapExprs.add(rexBuilder.makeInputRef(agg, i));
+            swapNames.add(agg.getRowType().getFieldNames().get(i));
+        }
+
+        RelNode swapProject = LogicalProject.create(agg, List.of(), swapExprs, swapNames);
```
expani, line 726: Can we add cases where there can be multiple projects as well ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
sandeshkr419: We bail, return Optional.empty() disables TopK for that query. Added a comment.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/TopKRewriterPlanShapeTests.java
sandeshkr419: Yes, DF's optimizer collapses the two coordinator Sorts (fetch=10000 + fetch=10) into a single TopK(fetch=10) in the physical plan. At the Calcite level we don't register SORT_REMOVE_REDUNDANT (which would eliminate the outer Sort), but it's harmless since DF handles it downstream. Added a TODO comment noting this.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
sandeshkr419: Correct. If PROJECT_MERGE is ever removed, multiple adjacent Projects would appear. The guard (seenProject == null) means we only capture the first Project — the second is skipped, so intermediateProject stays null and we don't remap (TopK fires with the raw collation, which is correct when no reordering exists between the two identity Projects since PROJECT_MERGE would have collapsed them anyway). Added a javadoc explaining this + a test (testDetection_multipleProjects_topKStillFires).

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
sandeshkr419: Yes — node.getInputs().size() == 1 on line 221 terminates the walk for any multi-input node (join, union). If a join/union sits between Sort and FINAL Aggregate, findFinalAgg returns null and TopK doesn't fire. Added this to the javadoc.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/TopKRewriterPlanShapeTests.java
sandeshkr419: Added testDetection_multipleProjects_topKStillFires — two identity Projects between Sort and Aggregate. PROJECT_MERGE collapses them during RBO, so TopK fires normally.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
expani: Can we not bail and support those cases ? I am assuming it should be simple to do so.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/PlanShapeIT.java
```diff
@@ -0,0 +1,223 @@
...
+ *       {@code resources/datasets/clickbench/ppl/expected_plans/{topk_on,topk_off}/}.
+ *   <li><b>Result correctness</b> (real {@code /_plugins/_ppl}): runs each TopK-eligible query
+ *       with factor=0 (baseline) and factor=2, asserts results match.
+ * </ol>
+ *
+ * <p>To regenerate golden files after an intentional plan change, run with
+ * {@code -Dtests.topk.generate=true}. This writes plans to stdout in a format you can
+ * copy into the resource files.
+ */
+public class PlanShapeIT extends AnalyticsRestTestCase {
```
expani: Should we iterate on this in a followup PR  ? Let's keep the edge cases restricted to `TopKRewriterPlanShapeTests` 

Thinking because we need to consider a lot of other factors like SQL/PPL settings, TopK setting, Slice Count setting, Single/Multi shard, workload type, asserting on the DF Logical and Physical plan; etc. 

`-Dtests.topk.generate=true` We should be manually updating this instead of generating ?

sandeshkr419: Sure, moving it as follow-up for now. Removed from this PR.

expani (approved):

mch2 (approved):

PR #22168, opened 2026-06-16
Support COUNT(DISTINCT x) OVER(...) via os_count_distinct UDAF in the DataFusion analytics backend

vinaykpud:
<table><tr><td width="1400">

## What changed

Adds an `os_count_distinct(x)` aggregate UDF to the DataFusion analytics backend, plus a planner-side adapter that rewrites `COUNT(DISTINCT x) OVER(...)` into `os_count_distinct(x) OVER(...)` before substrait emission. Non-distinct `COUNT(x) OVER(...)` passes through unchanged.

## Why

DataFusion 54.x's `datafusion-substrait` consumer hardcodes `distinct: false` on window-aggregate calls (`datafusion/substrait/src/logical_plan/consumer/expr/window_function.rs:116`), and the producer destructure-discards the same field (`producer/expr/window_function.rs:45,99`, with a `// TODO: fix`). Result: a vanilla `COUNT(DISTINCT x) OVER(...)` reaches the executor as a plain non-distinct `COUNT`, returning the wrong rows.

Every JVM-side layer threads DISTINCT to the wire correctly; the bug is purely in DataFusion's substrait crate. Vendoring or forking that crate would work but adds a sustained maintenance tax. Encoding DISTINCT in the operator *name* — by routing through a sandbox-namespace UDAF — sidesteps the consumer's broken `invocation` field and ships entirely inside the sandbox tree.

## Semantics

`os_count_distinct(x)` returns the count of distinct, non-null values:

| Input | Output |
|---|---|
| `{}` (empty) | 0 |
| `{NULL, NULL}` | 0 |
| `{1, 1, 2}` | 2 |
| `{1, NULL, 2, NULL, 1}` | 2 |

Cross-shard correctness: state is `List<element>` (the dedup'd value set). PARTIAL emits the per-shard set; FINAL unions the lists and recounts on the coordinator. NULLs are skipped at every stage, matching SQL `COUNT(DISTINCT x)` semantics.

## Query shapes supported

| Shape | Example |
|---|---|
| `OVER(ORDER BY ...)` running cumulative distinct count | `SELECT k, COUNT(DISTINCT g) OVER(ORDER BY k) FROM t` |
| `OVER()` global distinct count repeated on every row | `SELECT k, COUNT(DISTINCT g) OVER() FROM t` |
| `OVER(PARTITION BY ... ORDER BY ...)` per-partition running distinct count | `SELECT k, COUNT(DISTINCT g) OVER(PARTITION BY p ORDER BY k) FROM t` |
| Non-distinct passthrough — adapter is a no-op | `SELECT k, COUNT(g) OVER(ORDER BY k) FROM t` |

## PPL Calcite tests in opensearch-project/sql unblocked

- `WindowFunctionIT.testDistinctCountOver`
- `WindowFunctionIT.testDistinctCountOverNull`
- `WindowFunctionIT.testDistinctCountPartition`

These are the three failing cases that previously returned plain non-distinct counts because the substrait window invocation field was dropped.

## Tests

- `udaf::os_count_distinct::tests` — 7 Rust unit tests cover dedup, null skipping, multi-batch accumulation, cross-shard merge, and the FINAL list-input shape.
- `WindowFunctionAdaptersTests` — 2 new Java unit tests pin the rewrite contract: `COUNT(DISTINCT)` rebinds to `LOCAL_OS_COUNT_DISTINCT_OP` with the distinct flag cleared, and non-distinct `COUNT` passes through.
- `WindowCountDistinctIT` — sandbox QA IT covering the three failing test shapes plus null handling and non-distinct passthrough on the shared `calcs` dataset (`/_plugins/_sql` route).

## Test plan

- [x] Rust UDAF unit tests pass (`cargo test --lib udaf::os_count_distinct`)
- [x] Java adapter unit tests pass (`./gradlew :sandbox:plugins:analytics-backend-datafusion:test --tests WindowFunctionAdaptersTests`)
- [x] Live verified on Variant B (2-node, `-PrustDebug`) with `calcs` ingested:
  - `COUNT(DISTINCT str0) OVER(ORDER BY key)` → 1,1,2,2,...,3
  - `COUNT(DISTINCT str0) OVER()` → 3 on every row
  - `COUNT(DISTINCT str0) OVER(PARTITION BY str0 ORDER BY key)` → 1 within each partition
  - `COUNT(DISTINCT str3) OVER(ORDER BY key)` (str3 has nulls) → 1 once first non-null seen
  - `COUNT(str0) OVER(ORDER BY key)` (passthrough) → 1..17
- [x] `:precommit` clean on `analytics-backend-datafusion` and `analytics-engine`

</td></tr></table>

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/udaf/mod.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/udaf/mod.rs
@@ -14,10 +14,12 @@ use datafusion::execution::context::SessionContext;
 
 pub mod internal_pattern;
 pub mod list_merge;
+pub mod os_count_distinct;
 pub mod take;
 
 pub fn register_all(ctx: &SessionContext) {
     take::register_all(ctx);
     list_merge::register_all(ctx);
     internal_pattern::register_all(ctx);
+    os_count_distinct::register_all(ctx);
 }
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/udaf/os_count_distinct.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/udaf/os_count_distinct.rs
@@ -1,0 +1,333 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+//! `os_count_distinct(x)` — exact distinct count for use in a window context.
+//!
+//! Why this UDAF exists: DataFusion 54.x's substrait crate drops the
+//! `AggregationInvocation::DISTINCT` bit on window aggregates (the consumer
+//! hardcodes `distinct: false` and the producer writes `invocation: 0`). A
+//! plain `count(DISTINCT x) OVER(...)` therefore reaches the executor as a
+//! plain `count(x)`. By encoding DISTINCT in the *operator name* instead of
+//! the invocation enum, we sidestep the bug without vendoring a fork of
+//! `datafusion-substrait`.
+//!
+//! Semantics: emits the count of distinct, non-null values per group / per
+//! frame. NULL inputs are skipped — matches `count(DISTINCT x)` SQL semantics
+//! and the running cumulative behavior expected by `OVER(ORDER BY)`.
+//!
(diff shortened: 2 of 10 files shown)


mch2 (approved):

mch2:
We are doing a lot of hacks here vs updating the consumer, we should contribute updates here going fwd.

PR #22177, opened 2026-06-16
[analytics-engine] Emit SETUP_PARTIAL_AGGREGATE for all PARTIAL aggregate stages

sandeshkr419:
## Summary
  - Emit `SETUP_PARTIAL_AGGREGATE` instruction for all PARTIAL aggregate stages, not just engine-native-merge (dc)
  - Without this, DF uses `FinalPartitioned` in shards with slice_count > 1, causing double-counted groups at the coordinator

  ## Test plan
  - 4 unit tests: instruction emitted for COUNT, SUM, filtered aggregate, APPROX_COUNT_DISTINCT
  - 8 IT combinations: slice(1, N) × topk(on, off) × scan(vanilla, delegation) — asserts shard plan has only `mode=Partial`, coordinator has `Final`/`FinalPartitioned`
  - 
### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
```diff
@@ -288,7 +288,8 @@ pub async fn execute_with_context(
         // and fall through to the standard decode + execute below.
         if let Some(prepared) = handle.prepared_plan.as_ref() {
             let physical_plan = std::sync::Arc::clone(prepared);
-            let df_stream = execute_stream(physical_plan, handle.ctx.task_ctx()).map_err(|e| {
+            log_debug!("DataFusion physical plan:\n{}", displayable(physical_plan.as_ref()).indent(true));
```
expani: We already log the Datafusion physical plan below at line 352

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/dag/FragmentConversionDriver.java
```diff
@@ -249,6 +249,16 @@ private static List<InstructionNode> assembleInstructions(
     }
 
     /** Tree-walks for an engine-native-merge aggregate in the given mode. */
+    private static boolean containsPartialAggregate(RelNode root) {
```
expani, line 252: ***Purely clean code related comment***

There are 2 methods across the codebase to find a Partial Aggregate :  

- `isAggregatePath` in SortPushdownRewriter
- `findBuriedPartialAggregate` in FragmentConversionDriver 

We can add this in `RelNodeUtils`

If the other 2 places are not hard to migrate, we can do so. Else add a TODO so it's not missed out

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/ExplainApiIT.java
```diff
@@ -304,6 +304,143 @@ public void testProfileReturnsPhysicalPlan() throws IOException {
...
+                    );
+                    assertFalse(
+                        "coordinator plan must NOT contain mode=Partial (shards handle that). Plan:\n" + reducePlan,
+                        reducePlan.contains("mode=Partial")
+                    );
+                }
+            });
+    }
+
+    private void setSliceCount(String mode) throws IOException {
```
expani: Seems to be the incorrect name as it's setting the search mode. 

Can we add slice count as well ?

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/ExplainApiIT.java
```diff
@@ -304,6 +304,143 @@ public void testProfileReturnsPhysicalPlan() throws IOException {
             physicalPlan.contains("QueryShardExec"));
     }
 
+    /** slice=1, topk=off, vanilla scan */
+    @SuppressWarnings("unchecked")
+    public void testShardPlan_slice1_topkOff_vanilla() throws Exception {
```
expani: All these tests are related to TopKRewrite whereas this class is specific to ExplainAPI

Can we have a different IT dedicated to TopK Rewrite instead ?

expani: Another refactoring to make this less lines of code is to extract the 4 lines into a common test method like 

```
private void runTopKTest(String concurrentMode, int sliceCount, int oversamplingFactor, String pplQuery){
        ensureClickBenchProvisioned();
        setSliceCount(concurrentMode, sliceCount);
        setOversampling(oversamplingFactor);
        assertShardPlanPartialOnly(pplQuery);
}
```

expani (commented):
Thanks for taking this up @sandeshkr419 Good find 👏

sandeshkr419:
Thanks @expani for the review, addressed the comments.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/PartialAggregateModeIT.java
```diff
@@ -0,0 +1,141 @@
...
+ * Tests all combinations of slice count (1, N), TopK (on, off), and scan type
+ * (vanilla, delegation with filter).
+ */
+public class PartialAggregateModeIT extends AnalyticsRestTestCase {
+
+    private static volatile boolean provisioned = false;
+    private static final Dataset CLICKBENCH = ClickBenchTestHelper.DATASET;
+    private static final String VANILLA = "source=" + CLICKBENCH.indexName + " | stats count() as c by RegionID | sort - c | head 10";
+    private static final String DELEGATION = "source=" + CLICKBENCH.indexName
+        + " | where URL != '' | stats count() as c by URL | sort - c | head 10";
```
expani: This won't undergo delegation unless we remove from BlockList which the tests is not doing ?

expani: Or we can use a Predicate Function which can undergo delegation like EQUALS

expani (approved):

mch2 (approved):

PR #22179, opened 2026-06-16
[Analytics-Engine] Fix delegation panic on multi-index queries

mch2:
### Description
The indexed executor registered IndexedTableProvider under the per-shard table_name, but the substrait scan binds to the plan's NamedTable name to support multi-index queries.
Ex. The alias/pattern "tb-1,tb-2" for multi-index
queries.
DataFusion never consulted supports_filters_pushdown, kept the delegated_predicate FilterExec, and executed the marker UDF which panics.

Register under first_named_table_name(plan) so the provider binds and the Lucene-delegated filter is pushed down as a single-index.

Adds match()-delegation coverage to MultiIndexQueryShapesIT across comma-list, alias, and wildcard sources with count/dc/group-by/fields.

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
@@ -804,10 +804,19 @@ async unsafe fn execute_indexed_with_context_inner(
     // callback to the correct per-query FilterDelegationHandle and DelegationThreadTracker.
     let context_id = query_context.context_id();
 
+    // The substrait scan binds to the plan's NamedTable name (the alias/pattern like "tb-1,tb-2"
+    // for multi-index queries, the concrete name otherwise), not the per-shard table_name.
+    // create_session_context registered the provider under that name, so re-register under the
+    // same name. Using table_name for a multi-index query leaves the scan unbound, so DataFusion
+    // never consults supports_filters_pushdown and keeps the delegated_predicate FilterExec —
+    // which then executes the marker UDF and errors.
+    let register_name = crate::api::first_named_table_name(substrait_bytes.as_slice())
+        .unwrap_or_else(|| table_name.clone());
+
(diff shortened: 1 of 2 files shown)


sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
```diff
@@ -804,10 +804,19 @@ async unsafe fn execute_indexed_with_context_inner(
     // callback to the correct per-query FilterDelegationHandle and DelegationThreadTracker.
     let context_id = query_context.context_id();
 
+    // The substrait scan binds to the plan's NamedTable name (the alias/pattern like "tb-1,tb-2"
+    // for multi-index queries, the concrete name otherwise), not the per-shard table_name.
+    // create_session_context registered the provider under that name, so re-register under the
+    // same name. Using table_name for a multi-index query leaves the scan unbound, so DataFusion
+    // never consults supports_filters_pushdown and keeps the delegated_predicate FilterExec —
+    // which then executes the marker UDF and errors.
+    let register_name = crate::api::first_named_table_name(substrait_bytes.as_slice())
```
expani, line 813: So, we use the first named table as the table_name AND depend on the schema widening to handle different schema's ? 

Shouldn't the table name be the one associated with this Shard ?

expani (approved):

sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_executor.rs
mch2: yeah I don't love it, but the reason we do this is so we don't have to generate/update a unique plan per index at the coordinator.  Shard local planner could likely take care of this, but more substrait parsing.

schema widening is unavoidable here to support dynamic mapping.

bharath-techie (approved):
LGTM

PR #22187, opened 2026-06-16
[analytics-datafusion] Memory guard fix for spill

gaurav-amz:
### Description

Five fixes in the DataFusion analytics backend. They all affect native (Rust-side)
memory on data nodes when queries run concurrently, get cancelled, or spill to disk.

**1. Let a spilling query allocate the buffer it needs to finish spilling** (`memory.rs`, `memory_guard.rs`)

*Background — how spilling is driven here.* The memory pool watches the node's actual
resident memory (RSS, read from jemalloc) rather than only its own bookkeeping, because some
operators allocate native memory before they reserve it in the pool, so the pool's own counter
can understate reality. There are two RSS lines:

- **85% — the spill line.** Above this, the pool starts rejecting memory requests. A rejection
  is not an error to the operator: spillable operators (sort, grouped aggregate) treat a
  rejected request as the signal *"out of room — spill to disk now."* So rejecting *is* how the
  pool tells a query to spill.
- **95% — the critical line.** Above this, every request is rejected outright to protect the
  node from running out of memory.

*The problem.* Spilling is not free — to write its in-memory data out, an operator first frees
that data and then needs a small amount of memory to run the sort/merge and write the spill
file. The catch is timing: right after the operator frees its data, the freed pages have not
been returned to the OS yet (the allocator still holds them), so RSS is still above 85% at the
exact moment the spill needs its working memory. The old code rejected that request too — so
the spill it had just ordered could never finish, and the query got stuck with the
"Failed to reserve memory for sort during spill" error.

*The fix.* In the band between 85% and 95%, the pool now distinguishes two cases:

- A request from a **non-spillable** consumer is still rejected (it has no disk fallback, so
  letting it grow would only add pressure).
- A request from a **spillable** consumer is allowed through, because that consumer is in the
  middle of doing exactly what we want — spilling — and this request is part of finishing it.

To keep this safe, the allowance is bounded by a fixed **byte budget** (`SPILL_EXEMPT_CAP_BYTES`,
default 512 MB), not by a second RSS line. This matters: an RSS-based limit would not work,
because RSS lags behind the frees (the same staleness that caused the original bug). The byte
budget is logical accounting that updates the instant memory is freed, so it stays accurate
during a spill. A running total (`exempt_outstanding`) tracks how much has been allowed through
but not yet freed; once it reaches the budget, further spillable requests are rejected again
(which just makes them spill more aggressively). The budget is only charged *after* an
allocation actually succeeds, so a request that is allowed through but then fails for another
reason does not permanently consume budget; it is returned as memory is freed.

The 95% critical line is checked first and ignores this allowance entirely, so even with the
full budget granted to concurrent spills, the node can never be pushed past the hard limit.

*Flow — one spillable query as RSS climbs:*

1. RSS below 85%: every request is checked only against the pool limit and allowed. The query buffers in memory.
2. RSS crosses 85%: the next request from the spillable consumer is allowed through (it fits the budget) and charged to `exempt_outstanding`.
3. The budget fills: the next request is rejected. The operator reads this as "spill now", writes its data to disk, and frees it.
4. Freeing calls `shrink`, which lowers `exempt_outstanding` immediately — so the budget reopens even though RSS has not dropped yet.
5. The spill's own sort/merge buffer is now allowed through the reopened budget, so the spill finishes and the query continues.
6. If RSS ever reaches 95%, step 2 onward stops: every request is rejected regardless of the budget, protecting the node.

**4. Make runtime updates to the memory-guard thresholds actually take effect** (`DataFusionPlugin.java`)

The four memory-guard thresholds were each registered with a callback that re-read the
settings from the cluster service. During a settings update, that read returns the previous
value (the new values aren't visible to that read until after all callbacks have run), so
changing a threshold at runtime silently kept using the old value. The fix uses a single
callback that reads the four values from the updated settings it is handed, so the new values
take effect.

**5. Default the spill size limit to 80% of the spill disk, and reject invalid overrides** (`api.rs`, `DataFusionPlugin.java`, tests)

`datafusion.spill_memory_limit_bytes` now defaults to 80% of the total capacity of the disk
the spill directory is on (with an 8 GiB fallback if that capacity can't be read). An explicit
value set by an operator is rejected if it is larger than the disk's capacity.

### Check List
- [x] Functionality includes unit tests (the spill decision rule; `DataFusionPluginSettingsTests`).

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
@@ -104,7 +104,7 @@ public class DataFusionPlugin extends Plugin
     private static final Logger logger = LogManager.getLogger(DataFusionPlugin.class);
 
     /** Fraction of the spill volume's total capacity used as the default cap. */
-    static final double SPILL_LIMIT_FRACTION = 0.80;
+    static final double SPILL_LIMIT_FRACTION = 0.90;
 
     /** Fallback when the spill volume's capacity cannot be probed. 8 GiB. */
     static final long SPILL_LIMIT_FALLBACK_BYTES = 8L * 1024 * 1024 * 1024;
--- a/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
+++ b/sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
@@ -49,7 +49,6 @@ import org.opensearch.index.engine.exec.EngineReaderManager;
 import org.opensearch.index.engine.exec.IndexReaderProvider;
 import org.opensearch.index.get.DocumentLookupResult;
 import org.opensearch.indices.breaker.BreakerSettings;
-import org.opensearch.monitor.os.OsProbe;
 import org.opensearch.nativebridge.spi.NativeMemoryFetcher;
 import org.opensearch.nativebridge.spi.RustLoggerBridge;
 import org.opensearch.node.resource.tracker.ResourceTrackerSettings;
@@ -72,10 +71,12 @@ import org.opensearch.transport.client.Client;
 import org.opensearch.watcher.ResourceWatcherService;
 
 import java.io.IOException;
+import java.nio.file.FileStore;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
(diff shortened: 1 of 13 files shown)


Bukhtawar (commented):
Can we write a test to demonstrate spill kicks in?

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -94,6 +95,67 @@ public class DataFusionPlugin extends Plugin
 
     private static final Logger logger = LogManager.getLogger(DataFusionPlugin.class);
 
+    /** Fraction of the spill volume's total capacity used as the default cap. */
+    static final double SPILL_LIMIT_FRACTION = 0.80;
+
+    /** Fallback when the spill volume's capacity cannot be probed. 8 GiB. */
+    static final long SPILL_LIMIT_FALLBACK_BYTES = 8L * 1024 * 1024 * 1024;
+
```
Bukhtawar, line 111: Will this lead to wasted disk space? Can we reduce the 8GB buffer based on some benchmarks

gaurav-amz: This buffer is a fallback and based on minimum instance ram size (50% of 16GB). Ideally the `SPILL_LIMIT_FRACTION ` will be used which is 80% of the spill disk size. Above this threshold the rejection will be triggered for spill.

gaurav-amz:
>Can we write a test to demonstrate spill kicks in? 

Have added a integ test for spill test.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
Bukhtawar: For 50G spill volume, 10G would go unused?

gaurav-amz: Yes. Should I increase it to 90% or 95%? We can't set it to 100% because the OS only see about 99% of the memory.

Bukhtawar (approved):

PR #22191, opened 2026-06-16
Only carry relevant trailers into StreamException metadata in FlightErrorMapper

rishabhmaurya:
### Description

`FlightErrorMapper.fromFlightException` copied every Arrow Flight gRPC trailer onto the `StreamException` metadata using the `opensearch.` prefix. Most of these trailers (e.g. `content-type`, `raw-header`) are internal stream-transport framing that is not useful to the caller and just adds noise to the error metadata.

This change restricts the copied trailers to an allow-list of relevant keys (`SAFE_METADATA_KEYS`, currently just `correlation-id`), so the correlation id is preserved for debugging/tracing while the irrelevant transport trailers are not carried over. This also aligns the inbound mapping with the outbound `toFlightException` path, which already does not copy this metadata.

### Related Issues
N/A

### Check List


mch2 (approved):
lgtm

PR #22195, opened 2026-06-16
Add parent action name to ProcessorGenerationContext for system gener…

### LINKED ISSUE #22165: [RFC] Expose Parent Action Name in System Generated Processor Context
## Summary

Extend `ProcessorGenerationContext` to include the parent action name so that `SystemGeneratedFactory.shouldGenerate()` can make decisions based on which transport action triggered the search request (e.g., distinguish a direct `_search` from a search triggered by `_msearch`, `_async_search`, or a plugin action like SQL).

## Motivation

System generated processors need to conditionally generate based on the calling action. For example, a processor may only apply when the search is invoked directly via `TransportSearchAction` but not when triggered as a child of `_async_search` or `_reindex`. Today, `ProcessorGenerationContext` only exposes `SearchRequest`, which does not carry information about the parent action that initiated it.

Ref: https://github.com/opensearch-project/k-NN/issues/3365

### Approach 1: Pass Parent Task Through `resolvePipeline` (Recommended)

Thread the parent `Task` object (resolved from `SearchRequest.getParentTask()`) through `SearchPipelineService.resolvePipeline()` and extract the action name into `ProcessorGenerationContext`.

**Changes:**
- `ProcessorGenerationContext`: Add `String parentAction` field (nullable).
- `SearchPipelineService.resolvePipeline()`: Add `Task parentTask` parameter. Resolve `parentTask.getAction()` (null-safe) and pass it into the system-generated pipeline config map.
- `TransportSearchAction`: Resolve parent task from `taskManager.getTask(parentTaskId)` before calling `resolvePipeline`.
- `SearchPipelinePlugin.SystemGeneratedSearchPipelineConfigKeys`: Add `PARENT_ACTION` constant.
- `SystemGeneratedPipelineWithMetrics.generateProcessors()`: Extract parent action from config and pass to `ProcessorGenerationContext`.

**Pros:**
- Minimal blast radius: changes are limited to the search pipeline path.
- No serialization/BWC concerns since `Task` is a local in-memory object and the action name is extracted at the coordinator node.
- `ProcessorGenerationContext` is an internal DTO with no wire format.
- Clear extension point: additional context fields can be added to the record in the future without any transport-layer changes.

**Cons:**
- `parentTask` can be `null` when there is no parent (direct user request with no parent task ID). Callers of `shouldGenerate` must handle a nullable `parentAction`.
- The parent task must still be alive in the local `TaskManager` at the time of resolution. If the parent completes before the child reaches `resolvePipeline`, the lookup returns `null`. In practice this is unlikely since the parent awaits the child's response.

### Approach 2: Embed Parent Action in `TaskId` on `SearchRequest`

Extend `TaskId` (or add a parallel field on `SearchRequest`) to carry the parent action name so it is available generically anywhere `SearchRequest` is accessible.

**Changes:**
- `TaskId`: Add an optional `String parentAction` field.
- `SearchRequest` serialization: Write/read the new field.
- All transport actions that create child search requests: Populate the parent action.

**Pros:**
- Generically available: any code with access to `SearchRequest` can inspect the parent action without needing a `Task` reference.
- Works across nodes in cross-cluster search scenarios where the parent task is on a different node.

**Cons:**
- **BWC risk**: `TaskId` implements `Writeable` and is serialized on the wire (`streamIn`/`streamOut`). Adding a field changes the wire format. This requires version-gating (e.g., `if (version.onOrAfter(X))`) in serialization, and mixed-cluster rolling upgrades must handle both formats.
- **Large blast radius**: `TaskId` is used across the entire codebase (bulk, reindex, persistent tasks, etc.). Any change to its wire format affects all task-bearing requests, not just search pipelines.
- **Complexity of version gating**: Even with version gating, integration tests must cover mixed-cluster scenarios. The overhead is disproportionate to the feature's scope.
- **Semantic mismatch**: `TaskId` identifies a task; overloading it with action metadata conflates identity with context.

## Decision

**Approach 1** is recommended. The feature is scoped to system generated search processors and does not warrant a wire-format change to a foundational class like `TaskId`. The nullable `parentAction` on `ProcessorGenerationContext` is sufficient for the use case and avoids any BWC concerns.

## API Changes

```java
// Before
public record ProcessorGenerationContext(SearchRequest searchRequest) {}

// After
public record ProcessorGenerationContext(SearchRequest searchRequest, String parentAction) {}
```

```java
// Before
public PipelinedRequest resolvePipeline(
    SearchRequest searchRequest,
    IndexNameExpressionResolver indexNameExpressionResolver
) throws Exception;

// After
public PipelinedRequest resolvePipeline(
    SearchRequest searchRequest,
    Task parentTask,
    IndexNameExpressionResolver indexNameExpressionResolver
) throws Exception;
```

## Usage Example

```java
public class MyProcessorFactory implements SystemGeneratedFactory<SearchRequestProcessor> {

    @Override
    public boolean shouldGenerate(ProcessorGenerationContext context) {
        // Only generate when search is triggered directly, not from async_search
        String parentAction = context.parentAction();
        if (parentAction != null && parentAction.contains("async_search")) {
            return false;
        }
        return context.searchRequest().source().size() < 10;
    }
}
```

### Related component

Search

shatejas:
…ated processors

This exposes the parent transport action (e.g., _msearch, _async_search) to SystemGeneratedFactory.shouldGenerate() so processors can conditionally generate based on the calling action.

### Related Issues
Resolves [#22165](https://github.com/opensearch-project/OpenSearch/issues/22165)

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
+++ b/server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
@@ -485,7 +485,8 @@ public class TransportSearchAction extends HandledTransportAction<SearchRequest,
             PipelinedRequest searchRequest;
             ActionListener<SearchResponse> listener;
             try {
-                searchRequest = searchPipelineService.resolvePipeline(originalSearchRequest, indexNameExpressionResolver);
+                final Task parentTask = extractParentTask(originalSearchRequest);
+                searchRequest = searchPipelineService.resolvePipeline(originalSearchRequest, parentTask, indexNameExpressionResolver);
                 listener = searchRequest.transformResponseListener(updatedListener);
             } catch (Exception e) {
                 updatedListener.onFailure(e);
@@ -514,6 +515,14 @@ public class TransportSearchAction extends HandledTransportAction<SearchRequest,
             }, listener::onFailure);
             searchRequest.transformRequest(requestTransformListener);
         }
+    }
+
+    private Task extractParentTask(final SearchRequest searchRequest) {
+        TaskId taskId = searchRequest.getParentTask();
+        if (taskId != null && taskId != TaskId.EMPTY_TASK_ID) {
+            return taskManager.getTask(taskId.getId());
+        }
+        return null;
     }
 
     private ActionListener<SearchSourceBuilder> buildRewriteListener(
--- a/server/src/main/java/org/opensearch/search/pipeline/SearchPipelineService.java
+++ b/server/src/main/java/org/opensearch/search/pipeline/SearchPipelineService.java
@@ -47,6 +47,7 @@ import org.opensearch.index.analysis.AnalysisRegistry;
 import org.opensearch.ingest.ConfigurationUtils;
 import org.opensearch.plugins.SearchPipelinePlugin;
 import org.opensearch.script.ScriptService;
+import org.opensearch.tasks.Task;
 import org.opensearch.threadpool.ThreadPool;
 import org.opensearch.transport.client.Client;
 
@@ -66,6 +67,7 @@ import java.util.stream.Collectors;
 
 import static org.opensearch.cluster.service.ClusterManagerTask.DELETE_SEARCH_PIPELINE;
 import static org.opensearch.cluster.service.ClusterManagerTask.PUT_SEARCH_PIPELINE;
+import static org.opensearch.plugins.SearchPipelinePlugin.SystemGeneratedSearchPipelineConfigKeys.PARENT_ACTION;
 import static org.opensearch.plugins.SearchPipelinePlugin.SystemGeneratedSearchPipelineConfigKeys.SEARCH_REQUEST;
 
 /**
@@ -458,8 +460,11 @@ public class SearchPipelineService implements ClusterStateApplier, ReportingServ
         return newState.build();
     }
 
-    public PipelinedRequest resolvePipeline(SearchRequest searchRequest, IndexNameExpressionResolver indexNameExpressionResolver)
-        throws Exception {
+    public PipelinedRequest resolvePipeline(
+        SearchRequest searchRequest,
+        Task parentTask,
+        IndexNameExpressionResolver indexNameExpressionResolver
+    ) throws Exception {
         Pipeline pipeline = Pipeline.NO_OP_PIPELINE;
         if (searchRequest.source() != null && searchRequest.source().searchPipelineSource() != null) {
             // Pipeline defined in search request (ad hoc pipeline).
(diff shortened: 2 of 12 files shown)


server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
```diff
@@ -516,6 +517,14 @@ void executeRequest(
         }
     }
 
+    private Task extractParentTask(final SearchRequest searcRequest) {
+        TaskId taskId = searcRequest.getParentTask();
```
kotwanikunal: nit: `searchRequest.getParentTask`

server/src/main/java/org/opensearch/index/get/ShardGetService.java
```diff
@@ -227,7 +227,8 @@ private GetResult innerGet(
 
         try (
             Engine.GetResult get = indexShard.get(
-                new Engine.Get(realtime, true, id, uidTerm).version(version)
+                new Engine.Get(realtime, true, id, uidTerm)
```
kotwanikunal: nit: Not needed for this change.

server/src/main/java/org/opensearch/search/pipeline/SearchPipelineService.java
```diff
@@ -520,7 +525,10 @@ public PipelinedRequest resolvePipeline(SearchRequest searchRequest, IndexNameEx
             }
         }
         // Resolve system generated search pipeline
-        final Map<String, Object> config = Map.of(SEARCH_REQUEST, searchRequest);
+        final String parentAction = parentTask != null ? parentTask.getAction() : null;
```
kotwanikunal, line 528: Should we use structured task types here instead of raw strings? I was thinking of going one step ahead and adding in the necessary details instead of just the task name

server/src/main/java/org/opensearch/index/get/ShardGetService.java
shatejas: will revert

server/src/main/java/org/opensearch/action/search/TransportSearchAction.java
shatejas: will change

server/src/main/java/org/opensearch/search/pipeline/SearchPipelineService.java
shatejas: The idea is to limit access to whats required. Having said that we can add type, id is available from searchRequest. I am not sure if requestHeaders are needed

kotwanikunal: That sounds reasonable. We can potentially add additional structured params when needed.

kotwanikunal (approved):

kotwanikunal:
@owaiskazi19 / @andrross - mind taking a look here?

owaiskazi19 (approved):
LGTM!

PR #22198, opened 2026-06-17
Cancellation cleanup for analytics fragments: cooperative scan + parent-deregister cascade

aravindsagar:
### Description

Two complementary changes that reduce the time data-node analytics fragments hold `fragment_executor_gate` permits after the query they belong to is cancelled or completes. Each is independently meaningful and the test results below report the combined effect on a heavy-load workload.

---

#### 1. Cooperatively cancel indexed scan when the query task is cancelled

The indexed scan had no cancellation checkpoints. When a query task was cancelled, `IndexReader` kept dispatching and evaluating every remaining row group to completion, holding its native (Arrow/DataFusion) memory the entire time. The per-row-group evaluator runs inside `tokio::task::spawn_blocking`, which cannot be aborted once started.

This commit threads the per-query `CancellationToken` (from `QUERY_REGISTRY`) down through `IndexedTableConfig → IndexedExec → IndexReader` and adds three cooperative checkpoints:

- **`IndexReader::poll_next_row_group`** — bails before dispatching the next row group, so no new `spawn_blocking` jobs are admitted after cancellation.
- **`IndexReader::fetch_row_group`** — checks before the evaluator call, so a queued job that starts after cancellation skips its work.
- **`IndexedStream::poll_inner`** — stops draining, so decoded batches in the current segment are released immediately.

Cancellation surfaces as a `DataFusionError("query cancelled")` and propagates as a clean fragment failure — never partial results.

Queries with `context_id == 0` (untracked) and all unit tests pass `None` for the token, leaving existing behaviour unchanged.

#### 2. Cancel dispatched fragments before `AnalyticsQueryTask` deregisters

`TaskManager.setBan(parentTaskId)` resolves children by filtering the data-node-local `cancellableTasks` for tasks whose `parent_task_id` matches. When an `AnalyticsQueryTask` finishes (success, failure, or already-cancelled) and `TransportAction`'s framework wrapper calls `taskManager.unregister(task)`, that lookup no longer finds the data-node `AnalyticsShardTask` fragments dispatched from it. Under sustained heavy load this leaves orphan fragments alive on the data node holding `fragment_executor_gate` permits — they run to natural completion (~11–12 minutes for heavy LIKE-with-groupby queries).

This commit wraps the user listener in `QueryScheduler.execute` with `ActionListener.runBefore` that invokes `taskManager.cancelTaskAndDescendants(queryTask, …)` immediately before the framework deregisters the coord task. The cascade fires while the parent is still in the registry, so `setBan` resolves to its alive children and cancels them. Fires uniformly on success, failure, and cancelled terminals.

Idempotent against the already-cancelled path: `task.cancel` short-circuits when the task is already cancelled (`SetOnce.trySet`), and a duplicate `setBan` to a data node is a no-op at the receiving end (`banedParents` is a `ConcurrentHashMap` keyed by parent `TaskId`).

Required wiring change: `QueryScheduler` accepts `TransportService` via Guice and reads `taskManager` from it. No new cluster setting or configurability is introduced.

---

### Testing

**Unit tests (commit 1):**
- `cancel_stops_row_group_dispatch`: verifies that after cancellation, at most one already-in-flight `spawn_blocking` job (non-abortable) completes, the reader terminates via the cancellation error path, and fewer than all 8 row groups are evaluated.
- All existing e2e indexed table tests pass unchanged with `cancellation_token: None`.

**Unit tests (commit 2):**
- `QuerySchedulerTests` and `OperationListenerCoverageTests` updated for the new constructor signature; the full analytics-engine test suite passes.

**End-to-end results, commit 1 only (`stats count() by UserID` on a 100M-doc ClickBench cluster, ec2 r8g.2xlarge):**

| Cancel source | Behaviour |
| --- | --- |
| Manual `_tasks/_cancel` (5 trials, fired 166–675 ms into a 1.4 s query) | Query returns 8–10 ms after the cancel call. `tasks_cancelled=3`. `HTTP 500 TaskCancelledException[query cancelled]`. Never partial results. |
| SBP `enforced`, `elapsed_time_millis_threshold=500ms`, native-memory-duress workload | Baseline (SBP disabled) ~8.5 s. Under enforced SBP, all 10 queries cancelled and returned in 743–1001 ms — one SBP poll cycle after becoming eligible. |

**End-to-end results, both commits applied (14 QPS × 90 s of `where like(URL, "%google%") | stats count() by Title` against a 2-shard, 87M-doc ClickBench beta-stage AOS domain, `curl -m 60`, default SBP, no manual cancel):**

| Configuration | Drain after load end (gate `active_permits` → 0, sustained ≥30 s) |
| --- | --- |
| Stock | 11–12 min |
| Stock + manual `_tasks/_cancel?actions=*analytics*,cluster:admin/opensearch/ppl` every 1 s | 1–6 s |
| This PR (HTTP-disconnect cascade only — no manual cancel) | **61 s** |

The combined effect brings the HTTP-disconnect cancel path within an order of magnitude of explicit manual cancel, eliminating the multi-minute pin on `fragment_executor_gate` and the orphan-fragment accumulation in `_tasks`.

### Related Issues

### Check List


himshikhagupta:
How are we ensuring that all data is getting dropped in this flow when query is cancelled? Did we do any memory leak analysis with this?

aravindsagar:
> How are we ensuring that all data is getting dropped in this flow when query is cancelled? Did we do any memory leak analysis with this?

This is not targeted towards fixing memory leak, but to ensure that queries stop execution when cancelled. Did some test runs where the memory leak wasn't any worse with this change.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/QueryScheduler.java
```diff
@@ -59,7 +65,28 @@ public QueryExecution execute(QueryContext context, ActionListener<Iterable<Vect
...
-        }), () -> executions.remove(queryId));
+        }), () -> {
+            executions.remove(queryId);
+            // Cascade a cancel to dispatched data-node fragments before the framework unregisters
+            // the parent task. Otherwise TaskManager.setBan, which matches children by parent
+            // TaskId in the cancellable-tasks registry, cannot reach fragments after the parent
+            // has been unregistered — they survive into orphan state and run to natural
+            // completion. Idempotent against the already-cancelled path.
+            try {
+                taskManager.cancelTaskAndDescendants(
```
mch2, line 76: This is going to run for all queries as a runBefore, even those not impacted by SBP?  What is the cost here?  Does it short circuit? should we add this in failure listener path?

aravindsagar: Yes, it'll run before the framework unregisters any query. AnalyticsShardTask orphans were not exclusive to the failure/cancel path, hence did not add it just in the failure listener path.

Cost on the success / no-orphan path should be cheap; childTasksPerNode for such tasks will be an emptyMap, so a hash lookup is the extra cost in these cases.

I'm running a small load test to measure if there's a significant change.

mch2 (approved):

bharath-techie (approved):
LGTM

PR #22199, opened 2026-06-17
Fix flaky NativeArenaPurgerTests.testPurgeOnlyFiresAboveThreshold

### LINKED ISSUE #22669: Flaky Stabilize NativeArenaPurger threshold test
## Description

`NativeArenaPurgerTests.testPurgeOnlyFiresAboveThreshold` is timing-sensitive. It intermittently observes a purge after setting the threshold to `Long.MAX_VALUE`.

## CI evidence

- Failing test: `NativeArenaPurgerTests.testPurgeOnlyFiresAboveThreshold`
- Assertion: `Purge should not fire when below threshold expected:<0> but was:<1>`
- Failing task: `:sandbox:libs:dataformat-native:test`
- CI job: https://github.com/opensearch-project/OpenSearch/actions/runs/31149795016/job/92776901007?pr=22609

The job ran 45 tests in this task; this was the sole failure.

## Likely cause

The test suite starts a process-global native purge worker with a threshold of `0` and a 200 ms interval. The failing test then sets the threshold to `Long.MAX_VALUE`, samples the purge counter, and waits. There is no acknowledgement that the worker has observed the new configuration or that an earlier iteration has completed. A worker iteration that read the old threshold can therefore complete after the baseline sample and increment the counter, producing the observed `0 -> 1` failure.

The worker is backed by a native OS thread and its lifetime/configuration spans test methods, which increases cross-test timing sensitivity.

## Proposed stabilization

Add deterministic synchronization between the test and the native worker rather than relying on fixed sleeps. Preferred options:

1. Expose a test-only worker/configuration-generation acknowledgement so the test can wait until the new threshold is applied and any prior check is finished.
2. Alternatively, pause the worker, wait deterministically for an in-flight check to drain, establish the baseline, then resume it with the high threshold before asserting that the count is unchanged.

Use a latch, polling helper, or explicit native acknowledgement instead of `Thread.sleep` for the synchronization point.

## Scope

This failure is unrelated to the DataFusion coordinator-reduce teardown changes in PR #22609; that PR does not modify `sandbox/libs/dataformat-native` or the purger implementation.

himshikhagupta:
### Description
The test had a race condition: setUp() starts the purge thread with threshold=0, which can trigger an in-flight purge. When the test then sets threshold=MAX_VALUE and immediately records the baseline count, the in-flight purge can finish and increment the counter after the baseline is captured, causing a spurious assertion failure.

Fix: add a drain sleep after changing the threshold to let any in-flight purge complete before recording the baseline. Also reduce the check interval from 200ms to 50ms and tighten all sleeps proportionally to speed up the test suite.

### Related Issues
Resolves #22669 

### Check List

### DIFF
--- a/sandbox/libs/dataformat-native/src/test/java/org/opensearch/nativebridge/spi/NativeArenaPurgerTests.java
+++ b/sandbox/libs/dataformat-native/src/test/java/org/opensearch/nativebridge/spi/NativeArenaPurgerTests.java
@@ -21,11 +21,12 @@ import org.opensearch.test.OpenSearchTestCase;
 @ThreadLeakScope(ThreadLeakScope.Scope.NONE)
 public class NativeArenaPurgerTests extends OpenSearchTestCase {
 
+    private static final long INTERVAL_MS = 50;
+
     @Override
     public void setUp() throws Exception {
         super.setUp();
-        // Start with threshold=0 (always purge) and short interval for fast test feedback
-        NativeArenaPurger.init(0, 200);
+        NativeArenaPurger.init(0, INTERVAL_MS);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


sandbox/libs/dataformat-native/src/test/java/org/opensearch/nativebridge/spi/NativeArenaPurgerTests.java
```diff
@@ -36,25 +39,24 @@ public void tearDown() throws Exception {
...
     }
 
     public void testPurgeOnlyFiresAboveThreshold() throws Exception {
-        // Set threshold very high — purge should never fire
         NativeArenaPurger.setThresholdBytes(Long.MAX_VALUE);
+        Thread.sleep(DRAIN_MS); // let any in-flight purge (started under old threshold=0) finish
         long before = NativeArenaPurger.getPurgeCount();
-        Thread.sleep(500);
+        Thread.sleep(OBSERVE_MS);
         assertEquals("Purge should not fire when below threshold", before, NativeArenaPurger.getPurgeCount());
```
Bukhtawar: Better have a wait loop to check for a trigger condition, any time delay is generally flaky

sandbox/libs/dataformat-native/src/test/java/org/opensearch/nativebridge/spi/NativeArenaPurgerTests.java
```diff
@@ -36,25 +37,24 @@ public void tearDown() throws Exception {
...
+            assertEquals("No purges should fire when interval=0", countBefore, NativeArenaPurger.getPurgeCount());
+        });
     }
 
     public void testPurgeFiresPeriodically() throws Exception {
-        long before = NativeArenaPurger.getPurgeCount();
-        Thread.sleep(500);
-        assertTrue("Purge should have fired at least once", NativeArenaPurger.getPurgeCount() > before);
+        long countBefore = NativeArenaPurger.getPurgeCount();
+        assertBusy(() -> assertTrue("Purge should have fired at least once", NativeArenaPurger.getPurgeCount() > countBefore));
```
alchemist51, line 49: Can this be flaky? What is the condition for the getPurgeCount() to change? Is there a fixed interval then we can do the check post the interval

alchemist51:
Issue for this: #22669

sandbox/libs/dataformat-native/src/test/java/org/opensearch/nativebridge/spi/NativeArenaPurgerTests.java
himshikhagupta: setUp starts the purge thread with threshold 0 and a 50ms interval, so a purge fires every cycle and the count keeps rising. assertBusy polls for that with backoff up to 10s and exits on the first success — a large margin over a 50ms cadence. Checking once after a fixed interval is what the test did before, and that's the flake: a single hard deadline can be missed on a loaded host. The wait loop avoids it.

alchemist51 (approved):

himshikhagupta:
> ❌ Gradle check result for [881fbbd](https://github.com/opensearch-project/OpenSearch/commit/881fbbd5be4e87dd3dbdd1abb75329c6aab118c2): [FAILURE](https://build.ci.opensearch.org/job/gradle-check/83124/)
> 

Flaky test https://github.com/opensearch-project/OpenSearch/issues/21034

himshikhagupta:
Unrelated test case failing for sandbox check, gradle checks passing. The fix in this PR is for a flaky test so should not be the reason for other test failures. Should be safe to merge.

alchemist51:
Thanks @himshikhagupta  for the fix, merged the PR

PR #22206, opened 2026-06-17
Flush CPU runtime deferred drops on query cancellation

Bukhtawar:
## Summary

- After `cancel_query` aborts the CPU task via `AbortHandle::abort()`, DataFusion's `RepartitionExec` `pull_from_input` tasks (holding `GroupedHashAggregateStream` with GB-scale `GroupValues` buffers) remain in tokio's deferred drop queue until a worker thread processes them. On an idle runtime post-cancellation, this leaves hundreds of MB to GB of native memory allocated indefinitely.
- This change stores the `DedicatedExecutor`'s tokio runtime `Handle` in `QueryTracker`, and after aborting, spawns lightweight flush tasks across multiple workers that yield repeatedly — giving the scheduler opportunities to poll the aborted tasks and drop their captured futures (freeing GroupValues buffers).
- The calling Java thread blocks via `std::sync::mpsc` until all flush tasks complete or a 500ms timeout fires.

## Details

The abort cascade after `cancel_query` is:
1. CrossRtStream task aborted → polled → cancelled → future dropped (synchronous on one worker)
2. `PerPartitionStream` drops → `Arc<Vec<SpawnedTask>>` refcount hits 0 → `SpawnedTask::drop` aborts N `pull_from_input` tasks (synchronous)  
3. Each `pull_from_input` task must be **scheduled** (woken → polled → cancelled → future dropped → GroupValues freed)

Step 3 requires tokio workers to be actively cycling. The flush tasks ensure workers are active and processing the aborted tasks' deferred drops.

### Key design choices:
- **4 flush tasks × 3 yields each**: Spreads work across multiple runtime workers instead of serializing through one worker's yield loop. Covers the 2-level abort cascade with headroom.
- **500ms timeout**: Safety net. In practice the flush completes in < 1ms. If it times out, memory is still freed eventually (just not synchronously).
- **Zero hot-path cost**: Only runs on the cancellation path — no overhead on normal query execution.

## Test plan

- [x] `cargo check` — compiles cleanly (no new warnings)
- [x] `cargo test -- cancel_query` — all 5 existing cancel tests pass
- [x] `cargo test -- cross_rt_stream` — all 4 stream tests pass  
- [x] `cargo test -- local_executor` — cancel_query_fires_token test passes
- [ ] Deploy to staging cluster, run ClickBench workload with search cancellation, capture heap profile before/after to verify RSS drop

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/api.rs
@@ -1737,10 +1737,14 @@ pub async unsafe fn execute_local_plan(
     // shape as `execute_query`, so existing `stream_next` / `stream_close`
     // drain this handle unchanged. Use the cancellable variant so the CPU
     // task can be aborted mid-execution when cancel_query fires.
+    let cpu_exec = manager.cpu_executor();
     let (cross_rt_stream, abort_handle) =
-        CrossRtStream::new_with_df_error_stream_cancellable(df_stream, manager.cpu_executor());
+        CrossRtStream::new_with_df_error_stream_cancellable(df_stream, cpu_exec.clone());
     if let Some(h) = abort_handle {
         query_tracker::set_abort_handle(context_id, h);
+    }
+    if let Some(rt) = cpu_exec.handle() {
+        query_tracker::set_cpu_runtime_handle(context_id, rt);
     }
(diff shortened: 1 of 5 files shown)


Bukhtawar:
Closing - wrong base branch. Sandbox plugin files don't exist on upstream main. Reopening against feature branch.

Bukhtawar:
Superseded by #22207 (clean single commit on upstream/main)

PR #22208, opened 2026-06-17
Fix NPE in bitmap terms lookup when stored field is missing

### LINKED ISSUE #21995: [BUG] Bitmap terms lookup throws NPE when stored lookup field is missing
### Describe the bug

A `terms` query using terms lookup with `value_type: bitmap` and `store: true` throws a server-side exception when the lookup document exists but the stored field at `path` is missing.

This should not fail with `null_pointer_exception`; it should behave as empty terms (no matches).

### Related component

Search:Query Capabilities

### To Reproduce

```bash
# cleanup
curl -s -X DELETE "http://localhost:9200/target_idx_bitmap_lookup,lookup_idx_bitmap_lookup"

# target index
curl -s -X PUT "http://localhost:9200/target_idx_bitmap_lookup" -H "Content-Type: application/json" -d '{
  "mappings": {
    "properties": {
      "user_id": { "type": "integer" }
    }
  }
}'

# lookup index (stored bitmap field)
curl -s -X PUT "http://localhost:9200/lookup_idx_bitmap_lookup" -H "Content-Type: application/json" -d '{
  "mappings": {
    "properties": {
      "bitmap_payload": { "type": "keyword", "store": true },
      "note": { "type": "keyword" }
    }
  }
}'

# lookup doc intentionally missing bitmap_payload
curl -s -X PUT "http://localhost:9200/lookup_idx_bitmap_lookup/_doc/1?refresh=true" -H "Content-Type: application/json" -d '{
  "note": "field intentionally missing"
}'

# target docs
curl -s -X POST "http://localhost:9200/target_idx_bitmap_lookup/_bulk?refresh=true" \
  -H "Content-Type: application/x-ndjson" \
  --data-binary $'{"index":{"_id":"1"}}\n{"user_id":1}\n{"index":{"_id":"2"}}\n{"user_id":2}\n'

# trigger terms lookup bitmap query
curl -s -X GET "http://localhost:9200/target_idx_bitmap_lookup/_search?pretty" -H "Content-Type: application/json" -d '{
  "query": {
    "terms": {
      "user_id": {
        "index": "lookup_idx_bitmap_lookup",
        "id": "1",
        "path": "bitmap_payload",
        "store": true
      },
      "value_type": "bitmap"
    }
  }
}'
```

### Expected behavior

If the lookup field is missing, query rewrite should treat it as empty terms and return no hits (equivalent to `match_none`) without throwing NPE.

### Additional Details

**Plugins**
All default plugins in AWS opensearch when encountering the issue.

- OS: macOS
  - OpenSearch version/commit: AWS Opensearch 3.5. Tested also on latest opensearch, commit d5756a7f7fad279b85b39d23fe4d37db97e0b857

kshitizj03:
### Description
A `terms` query using terms lookup with `value_type: bitmap` and `store: true` throws a
`NullPointerException` when the looked-up document exists but the stored field at `path` is
missing.

Root cause: in `TermsQueryBuilder#fetch`, `getResponse.getField(termsLookup.path())` returns
`null` for an absent stored field, and `getValues()` was called on it unconditionally.

This change guards the null `DocumentField` and treats a missing stored field as no terms
(no matches), which is consistent with how a missing source field is already handled. Existing
single-value bitmap validation is preserved for the present-field case.

Added a REST yaml test in `search/380_bitmap_filtering.yml`: a lookup document missing the stored
bitmap field now returns `hits.total: 0` instead of erroring.

### Related Issues
Resolves #21995

### Check List

### DIFF
--- a/rest-api-spec/src/main/resources/rest-api-spec/test/search/380_bitmap_filtering.yml
+++ b/rest-api-spec/src/main/resources/rest-api-spec/test/search/380_bitmap_filtering.yml
@@ -48,6 +48,8 @@ setup:
           -  { "enrolled": "OjAAAAEAAAAAAAAAEAAAAE0B" }     # 333
           -  { "index": { "_index": "classes", "_id": "104" } }
           -  { "enrolled": "OjAAAAEAAAAAAAEAEAAAAN4ATQE=" } # 222,333
+          -  { "index": { "_index": "classes", "_id": "105" } }
+          -  {} # "enrolled" stored field intentionally missing
   - do:
       cluster.health:
         wait_for_status: green
@@ -182,3 +184,24 @@ setup:
   - match: { hits.hits.1._source.student_id: 111 }
   - match: { hits.hits.2._source.name: John Doe }
   - match: { hits.hits.2._source.student_id: 333 }
+
+---
+"Terms lookup on a missing stored bitmap field returns no matches":
+  - do:
+      search:
+        rest_total_hits_as_int: true
+        index: students
+        body: {
+          "query": {
+            "terms": {
+              "student_id": {
+                "index": "classes",
+                "id": "105",
+                "path": "enrolled",
+                "store": true
+              },
+              "value_type": "bitmap"
+            }
+          }
+        }
+  - match: { hits.total: 0 }
--- a/server/src/main/java/org/opensearch/index/query/TermsQueryBuilder.java
+++ b/server/src/main/java/org/opensearch/index/query/TermsQueryBuilder.java
@@ -42,6 +42,7 @@ import org.opensearch.Version;
 import org.opensearch.action.get.GetRequest;
 import org.opensearch.action.search.SearchRequest;
 import org.opensearch.common.SetOnce;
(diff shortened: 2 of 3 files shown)


kshitizj03:
The gradle-check failure looks like an unrelated CI flake rather than a real test failure.

The failing task is `:plugins:analysis-phonetic:test`, and the cause is a Gradle worker connection problem, not a test assertion:

```
org.gradle.internal.remote.internal.ConnectException: Could not connect to server [...127.0.0.1...]. Connection refused
MessageIOException: Could not read message from '/127.0.0.1:49564'
```

The Jenkins run also reports "Test Result (no failures)", and `:server:test` passed in the same run. This change only touches bitmap terms lookup in `:server`, and the bitmap REST suite (`search/380_bitmap_filtering`, including the new missing-field case) passes locally.

Could a maintainer please re-run gradle-check? Thanks!

kshitizj03:
Rebased on latest main and added a unit test for the missing-stored-field case (covers the null-guard branch directly). gradle-check is green now. Could a maintainer take a look when there's a chance? Happy to address any feedback.

PR #22209, opened 2026-06-17
Schema alignment fix for dynamic mapping

alchemist51:
### Description

Row-group / page statistics pruning in the composite indexed-parquet scan resolved `StatisticsConverter` against the **full union table schema**. `StatisticsConverter` maps a column name to a parquet leaf **positionally** (parquet crate `parquet_column`): it finds the column's ordinal in the arrow schema it is handed and indexes into the parquet file by that ordinal.

Under **dynamic-mapping schema drift** — where a segment's own parquet schema is narrower or reordered relative to the union schema — that ordinal lands on the wrong leaf, or off the end of the file. The converter then returns all-null / wrong-column statistics, and an always-true residual such as `severity >= 0` evaluates to *"cannot match"*, so the whole row group / segment is wrongly pruned and matching rows are **silently dropped** (e.g. `match('…') AND severity >= 0` returned far fewer rows than `match('…')` alone).

### Fix

Resolve statistics against each **segment's own** parquet schema, reconstructed from its footer via `parquet_to_arrow_schema(descr, kv)`, at all three stats-resolution sites:

- `page_pruner.rs` `eval_leaf` — RG-level / whole-segment prune
- `page_pruner.rs` `PagePruner::prune_rg` — page-level prune
- `dynamic_filter.rs` `RgPruningContext::rg_provably_excluded` — dynamic-filter prefetch prune

The `PruningPredicate` is still built against the table schema (its column refs are table-schema relative); only the stats-extraction step — mapping a name to a physical leaf — uses the segment schema. Falls back to the table schema if footer conversion fails.

### Testing

- Two Rust unit tests (one per fixed file) build a fixture where the filter column resolves positionally onto a different real column and assert the row group is not pruned.
- `SchemaDriftPrunePplIT`: end-to-end REST IT over a composite index with a Lucene secondary and dynamic mapping, run under a new `integTestNoMerge` cluster variant. Asserts numeric-filter-only, sort+limit, `match() + numeric residual`, and `match() + filter + sort + limit` all return correct counts under schema drift.

### Check List
- [x] New functionality has been documented.

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_table/dynamic_filter.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/indexed_table/dynamic_filter.rs
@@ -204,10 +204,19 @@ impl RgPruningContext {
         let Some(rg_meta) = metadata.row_groups().get(rg_idx) else {
             return false;
         };
+        // Resolve stats against the segment's OWN parquet schema, not the full table
+        // schema: StatisticsConverter maps name→parquet-column positionally, so the full
+        // schema reads the wrong/no column under dynamic-mapping schema drift.
+        let descr = metadata.file_metadata().schema_descr();
+        let seg_schema = datafusion::parquet::arrow::parquet_to_arrow_schema(
+            descr,
+            metadata.file_metadata().key_value_metadata(),
+        )
+        .unwrap_or_else(|_| self.schema.as_ref().clone());
         let stats = SingleRowGroupStatistics {
-            parquet_schema: metadata.file_metadata().schema_descr(),
+            parquet_schema: descr,
             rg_meta,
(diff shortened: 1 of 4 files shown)


alchemist51:
The `gradle-check` failure is an unrelated flaky test, not caused by this change. This PR is **sandbox-only** (all changes under `sandbox/`); the core/server suite that `gradle-check` runs does not exercise any modified code. The `sandbox-check` job — which builds and tests the changed Rust crates and runs the new integration test — passes consistently, including after a clean rebase onto the latest `main` (head `86ffb30a`). Could a maintainer re-run `gradle-check`? Happy to rebase again if it helps.

bharath-techie (approved):
schema_drift.rs also needs to be updated. But otherwise LGTM

PR #22210, opened 2026-06-17
[analytics-engine] Conditionally include CombinePartialFinalAggregate via FFI flag for single-shard correctness

sandeshkr419:
On a single-shard index, the Java planner correctly keeps the aggregate as SINGLE (no PARTIAL/FINAL split, no coordinator reduce). But `CombinePartialFinalAggregate` is unconditionally removed from the Rust session optimizer rules, so DataFusion produces:

```
      AggregateExec: mode=FinalPartitioned
        RepartitionExec: Hash([group_key], 4)
          AggregateExec: mode=Partial
            DataSourceExec
```

  With target_partitions > 1, this hash-repartition produces incorrect row counts for group-by queries with HAVING filters on large datasets.

  ## Why the rule was removed

  `prepare_partial_plan` (multi-shard PARTIAL path) needs Partial+Final intact so `apply_aggregate_mode(Partial)` can strip the Final half:

      WITH rule:    P+F --> Single --> strip(Partial) --> REMOVES ENTIRE AGG (broken)
      WITHOUT rule: P+F (intact) --> strip(Partial) --> keeps Partial only (correct)

  ## Fix

  Thread `hasPartialAggregate` from Java to Rust at session creation time:

```
      FragmentConversionDriver.containsPartialAggregate()
          |
          v
      ShardScanInstructionNode(hasPartialAggregate)
          |
          v
      ShardScanInstructionHandler --> NativeBridge.createSessionContext(..., hasPartialAggregate)
          |
          v (FFI)
      session_context::create_session_context(..., has_partial_aggregate: bool)
          |
          +-- true  (multi-shard) --> exclude CombinePartialFinalAggregate
          +-- false (single-shard) --> use default rules (P+F collapses to Single)
```
  Single point of control, no post-hoc fixup. Alternative I considered was to decide to add the rule back if agg mode is default: https://github.com/sandeshkr419/OpenSearch/commit/a9cdfe5ae07e2840721d502efee768378ce920c5 but discarded the approach so that we don't have to remove a rule to add it back making the structure flaky.

  ## Test Plan

  - [x] `SingleShardAggregationIT` -- 12 tests covering full matrix:
    - Partitions: {slice=1, slice=4}
    - TopK: {off, on}
    - Scan type: {vanilla DF ListingTable, delegation IndexedTableProvider}
    - Plan shape: no COORDINATOR_REDUCE on 1-shard index

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
@@ -142,6 +142,7 @@ pub async unsafe fn create_session_context(
     shard_view_ptr: i64,
     table_name: &str,
     context_id: i64,
+    has_partial_aggregate: bool,
     query_config: DatafusionQueryConfig,
     plan_bytes: &[u8],
 ) -> Result<i64, DataFusionError> {
@@ -209,7 +210,11 @@ pub async unsafe fn create_session_context(
         .with_config(config)
         .with_runtime_env(Arc::from(runtime_env))
         .with_default_features()
-        .with_physical_optimizer_rules(crate::agg_mode::physical_optimizer_rules_without_combine());
+        .with_physical_optimizer_rules(if has_partial_aggregate {
+            crate::agg_mode::physical_optimizer_rules_without_combine()
+        } else {
+            datafusion::physical_optimizer::optimizer::PhysicalOptimizer::new().rules
+        });
 
     // For ListingTable query strategy:
     // 1. Add ProjectRowIdAnalyzer (logical) — ensures __row_id__ survives pruning.
@@ -377,10 +382,11 @@ pub async unsafe fn create_session_context_indexed(
     tree_shape: i32,
     delegated_predicate_count: i32,
     requests_row_ids: bool,
+    has_partial_aggregate: bool,
     query_config: DatafusionQueryConfig,
     plan_bytes: &[u8],
 ) -> Result<i64, DataFusionError> {
-    let ptr = create_session_context(runtime_ptr, shard_view_ptr, table_name, context_id, query_config, plan_bytes).await?;
+    let ptr = create_session_context(runtime_ptr, shard_view_ptr, table_name, context_id, has_partial_aggregate, query_config, plan_bytes).await?;
 
     // Augment with indexed config. The delegation marker UDFs (index_filter, delegation_possible)
     // are now registered for every session by udf::register_all (via create_session_context above);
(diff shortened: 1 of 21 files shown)


sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
```diff
@@ -209,7 +210,11 @@ pub async unsafe fn create_session_context(
         .with_config(config)
         .with_runtime_env(Arc::from(runtime_env))
         .with_default_features()
-        .with_physical_optimizer_rules(crate::agg_mode::physical_optimizer_rules_without_combine());
+        .with_physical_optimizer_rules(if has_partial_aggregate {
+            crate::agg_mode::physical_optimizer_rules_without_combine()
```
expani, line 214: Can we look at never removing this unless there is a Partial or a Final Aggregate instruction passed from Core ?

expani: Basically trying to see if we can avoid setting a flag of `hasPartialAggregate` on the ShardScanInstruction

sandeshkr419: I like this idea, let me iterate on this quickly, thanks!

sandeshkr419: I moved this to execution context, please check the latest changes. Let me know if you had something else in mind entirely?

expani: IIUC We are using the flag of `private boolean hasPartialAggregate;` to set/unset the Physical Optimizer of  `CombinePartialFinalAggregate` 

I am thinking the set/unset is already handled by `PartialAggregateInstructionHandler` so we can depend on it

expani: Phew I think this will get messy as SessionContext is immutable I am fine with the current approach.

expani (approved):

sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
sandeshkr419: The SessionState optimizer rules are immutable after build(). PartialAggregateInstructionHandler runs after session creation so it can't retroactively remove the rule. The flag tells session construction upfront whether to exclude it.

mch2 (approved):

PR #22211, opened 2026-06-17
Gate Rust log macros on level to avoid format! on hot path

finnegancarroll:
## Summary

Rust `log_debug!`/`log_info!` macros unconditionally call `format!(...)` before the FFM upcall to Java, paying string allocation + formatting cost even when the log level is suppressed on the Java side. On hot paths (per-row-group, per-segment), this is a measurable overhead.

## Changes

These changes register  a cluster settings listener via `addAffixUpdateConsumer(Loggers.LOG_LEVEL_SETTING, ...)`. Whenever the log level is modified in cluster settings we use rust down call `RustLoggerBridge.pushLevel()` to update atomic static `MAX_LEVEL`. When log level is higher than `MAX_LEVEL` skip `format!` call entirely and short circuit the log.

## Testing

Manually tested.

1. Start single-node cluster with analytics-backend-datafusion plugin, delegation-enabled index (keyword field with Lucene secondary format).
2. Query at default log level (INFO). Confirm no debug logs.
3. Enable debug logs.
```
 PUT /_cluster/settings
  {"transient": {"logger.org.opensearch.nativebridge.spi.RustLoggerBridge": "DEBUG"}}
```
4. Re-run query, observe debug logs in output.
```
  [2026-06-17T20:40:42,804][INFO ][o.o.c.s.ClusterSettings  ] [integTest-0] updating [logger.org.opensearch.nativebridge.spi.RustLoggerBridge] from [INFO] to [DEBUG]
  [2026-06-17T20:40:45,440][DEBUG][o.o.n.s.RustLoggerBridge ] [integTest-0] create_session_context: registered table 'logtest_delegation' with file_sort_order_keys=0
  [2026-06-17T20:40:45,441][DEBUG][o.o.n.s.RustLoggerBridge ] [integTest-0] DataFusion logical plan:
  [2026-06-17T20:40:45,441][DEBUG][o.o.n.s.RustLoggerBridge ] [integTest-0] DataFusion physical plan:
  [2026-06-17T20:40:45,442][DEBUG][o.o.n.s.RustLoggerBridge ] [integTest-0] [scf-rust] lazy provider initialized context_id=49 annotation_id=0 provider_key=1
  [2026-06-17T20:40:45,443][DEBUG][o.o.n.s.RustLoggerBridge ] [integTest-0] [scf-segment-done] file=_parquet_file_generation_1.parquet row_groups=1 elapsed=1.073266ms
```

### DIFF
--- a/sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
+++ b/sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
@@ -8,6 +8,7 @@
 
 package org.opensearch.nativebridge.spi;
 
+import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
@@ -27,18 +28,23 @@ import java.nio.charset.StandardCharsets;
  * <p>At startup, {@link NativeLibraryLoader} registers {@link #log} as a function pointer
  * with the native library. Rust calls it directly — no JNI.
  *
- * <p>TODO: the Rust {@code log_debug!} / {@code log_info!} macros currently format
- * their message ({@code format!(...)}) unconditionally and only filter by level here,
- * after the FFM call lands. Hot-path debug sites pay the formatting + allocation cost
- * regardless of whether debug is enabled on the Java side. Expose the active log
- * level from Java back to Rust (e.g. an atomic refreshed on level changes) so the
- * macros can short-circuit before {@code format!} runs. Until then, hot-path Rust
- * call sites either need to be commented out (see {@code single_collector.rs}) or
- * gated with a manual level check.
+ * <p>Level short-circuit: the Rust {@code log_debug!}/{@code log_info!} macros gate on a
+ * Rust-side atomic level BEFORE {@code format!}, so a suppressed log costs only an atomic
+ * load + branch (no string allocation, no FFM crossing). Java keeps that atomic in sync
+ * via {@link #pushLevel()} — called at registration and re-pushed on every upcall if the
+ * level has drifted (detecting dynamic changes from {@code _cluster/settings}).
  */
 public class RustLoggerBridge {
 
     private static final Logger logger = LogManager.getLogger(RustLoggerBridge.class);
+
+    /** Level codes shared with Rust's {@code LogLevel}: 0=Debug, 1=Info, 2=Error. */
+    private static final int LEVEL_DEBUG = 0;
+    private static final int LEVEL_INFO = 1;
+    private static final int LEVEL_ERROR = 2;
+
+    /** Bound downcall to Rust's {@code native_logger_set_level(int)}; null until registered. */
+    private static volatile MethodHandle setLevelHandle;
 
     /**
      * Called from Rust via the registered function pointer.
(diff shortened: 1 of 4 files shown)


sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
```diff
@@ -74,5 +88,42 @@ static void register(Linker linker, MemorySegment initLoggerSymbol) {
...
+        int level = currentJavaLevel();
+        lastPushedLevel = level;
+        try {
+            handle.invokeExact(level);
+        } catch (Throwable t) {
+            logger.error("Failed to push native log level", t);
+        }
+    }
+
+    private static int currentJavaLevel() {
```
expani: Logger has a `getLevel().intLevel()` which seems more appropriate here.

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
```diff
@@ -27,24 +27,38 @@
...
+
+    /** Last level we pushed to Rust, used to detect drift without an extra FFM call. */
+    private static volatile int lastPushedLevel = LEVEL_INFO;
+
     /**
      * Called from Rust via the registered function pointer.
      * Signature must match: {@code void(int level, MemorySegment msgPtr, long msgLen)}
      */
     static void log(int level, MemorySegment msgPtr, long msgLen) {
+        // Detect level drift: if Java's level changed since last push, re-sync Rust.
```
expani: Can we not depend on the `addAffixUpdateConsumer` to propagate the level on updates instead of actively detecting the drift ?

finnegancarroll: Yes this code is not needed. I think only way this can happen is if someone sets log level manually in code. Removing.

sandbox/libs/dataformat-native/src/main/java/org/opensearch/nativebridge/spi/RustLoggerBridge.java
finnegancarroll: Added this change. It looks like log4j logger int levels are in the range of 100, 200, ect so mapping these onto rust log levels as well.

sandbox/plugins/analytics-backend-datafusion/src/main/java/org/opensearch/be/datafusion/DataFusionPlugin.java
```diff
@@ -448,6 +450,11 @@ public Collection<Object> createComponents(
             .addSettingsUpdateConsumer(DATAFUSION_MEMORY_GUARD_ADMISSION_THROTTLE_THRESHOLD, v -> updateMemoryGuardThresholds());
         clusterService.getClusterSettings()
             .addSettingsUpdateConsumer(DATAFUSION_MEMORY_GUARD_ADMISSION_REJECT_THRESHOLD, v -> updateMemoryGuardThresholds());
+
+        // Push Rust log level whenever any logger.* cluster setting changes, so Rust macros
```
expani, line 454: Should we gate this at `org.opensearch.nativebridge.spi.RustLoggerBridge` ? 

I am assuming this will make Rust start emitting Debug loggers even if the Level is set at `org.opensearch.parquet.* = DEBUG`

finnegancarroll: Rust will only emit debug logs when exactly the the `RustLoggerBridge` logger is enabled. Our cluster setting listener triggers on all cluster setting updates, but `pushLevel` only ever syncs the rust log level with the `RustLoggerBridge` logger log level.

expani (approved):

sandeshkr419 (approved):

PR #22227, opened 2026-06-17
Integrate rust side changes for memory pool

rayshrey:
### Description

### Related Issues

### Check List

### DIFF
--- a/sandbox/libs/dataformat-native/rust/common/src/memory_pool.rs
+++ b/sandbox/libs/dataformat-native/rust/common/src/memory_pool.rs
@@ -26,13 +26,15 @@ pub const DEFAULT_WAIT_TIMEOUT: Duration = Duration::from_secs(300);
 /// Merge operations can wait longer (600 seconds).
 pub const MERGE_WAIT_TIMEOUT: Duration = Duration::from_secs(600);
 
-/// Controls whether an allocation blocks or rejects immediately.
+/// Controls how a reservation reacts when the pool is full.
 #[derive(Debug, Clone)]
 pub enum PoolBehavior {
     /// Block until memory is available, up to the given timeout.
     Wait(Duration),
     /// Fail immediately if pool is full.
     Reject,
+    /// Never block and never fail: account via infallible `grow`, allowing the pool to over-commit.
+    IgnoreLimit,
 }
 
 /// Error returned when a pool cannot satisfy an allocation request.
@@ -253,6 +255,11 @@ impl MemoryReservation {
                 self.size += bytes;
                 Ok(())
             }
+            PoolBehavior::IgnoreLimit => {
+                self.pool.grow(bytes);
+                self.size += bytes;
+                Ok(())
+            }
         }
     }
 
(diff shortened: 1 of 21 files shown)


sandbox/plugins/parquet-data-format/src/main/rust/src/writer.rs
```diff
@@ -462,17 +478,29 @@ impl NativeParquetWriter {
...
+                            writer.write(&record_batch, &mut state.reservation)?;
                         }
                         WriterVariant::Parquet(writer_arc) => {
                             log_debug!("Writing RecordBatch to Parquet file");
+                            let batch_bytes = record_batch.get_array_memory_size();
+                            // Reserve 2× batch as estimate — ArrowWriter encoding may temporarily
+                            // hold dictionary, compressed pages, and data page buffers.
+                            let estimated = batch_bytes * 2;
+                            let writer_arc = writer_arc.clone();
+                            state.reservation.reserve_estimated(estimated)?;
```
mgodwan, line 496: what is the behavior if we cannot reserve?

rayshrey: We return a PoolTimeOut Error from Rust side, which will ultimately lead to shard failures on Java.

sandbox/libs/dataformat-native/rust/common/src/memory_pool.rs
```diff
@@ -269,6 +269,28 @@ impl MemoryReservation {
         self.size -= actual;
     }
 
+    /// Reserve an estimated amount. Returns the estimated amount for later use with `reconcile()`.
+    pub fn reserve_estimated(&mut self, estimated: usize) -> Result<usize, Box<dyn std::error::Error>> {
+        self.request(estimated)?;
+        Ok(estimated)
+    }
```
Bukhtawar, line 283: This might hold lock and wait on grow(when close to the pool limit?) which can contend with other threads needing the same shard lock or trying to shrink?

Bukhtawar: I would prefer we apply backpressure or rejection early at the Java layer(based on native limits) and avoid locks/awaits at the Rust layer(yes this will might overshoot slightly but avoids deadlocks altogether).

rayshrey: request (based on PoolBehaviour - which can be either Wait or Reject) will either wait for some time or simply reject the request if capacity is not available. 
Rejections will simply cause us to throw, and Wait will also eventually cause us to throw (after the specified duration). As a result of this, the reservation instance will be dropped as soon as we go out of scope causing the pool to release the memory as well.

Also in both these behaviours we ultimately call try_and_grow (which rejects reservations if memory is pool is above limit) and wait_and_grow (which does not run indefinitely but only till a max duration which we have configured to be 5 mins). So none of these calls are blocking. The grow method that we have separately is also non-blocking in the sense that we just increase the counters in that call (since the reservation has already been made there)

server/src/main/java/org/opensearch/indices/IndexingMemoryController.java
```diff
@@ -115,10 +115,10 @@ public class IndexingMemoryController implements IndexingOperationListener, Clos
 
     /** How much native (off-heap) memory we allow for indexing buffers across all shards.
      *  Accepts either a percentage (of available native memory = total physical - JVM heap) or an absolute byte value.
-     *  Default is 10%. */
+     *  Default is 7%. */
     public static final Setting<ByteSizeValue> NATIVE_INDEX_BUFFER_SIZE_SETTING = Setting.nativeMemorySizeSetting(
         "indices.memory.native_index_buffer_size",
-        "10%",
+        "7%",
```
Bukhtawar, line 121: This might have perf impact, can we lower it only marginally and run benchmarks

rayshrey: So this is the initial value we begin with, which was previously 10% of the (total-jvm) memory
In IMC we are not aware of the various pools that we have (since we do not want any dependency on any plugin here while construction)

Eventually after the plugins load, when pools are rebalanced, we reset this limit to 70% of the Indexing pool limit through listeners. The 7% value that we have modified this to is basically the corrected value that we eventually set this to after the plugins are loaded and rebalancer runs.

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -1353,6 +1353,18 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
 
             final RestController restController = actionModule.getRestController();
 
+            // Discover the native-allocator stats supplier from any plugin that publishes a
+            // NativeAllocatorStatsRegistry component (today: ArrowBasePlugin). Lookup mirrors
+            // the SearchRequestOperationsListener instanceof filter on pluginComponents elsewhere
+            // in this file. Server has no compile-time dependency on arrow-base.
```
Bukhtawar, line 1359: nit:remove

Bukhtawar (approved):
Thanks lgtm

sandbox/libs/dataformat-native/rust/common/src/memory_pool.rs
Bukhtawar: Thanks for the changes

PR #22237, opened 2026-06-19
[Sandbox] Add TieredBlockCache with separate metadata and data SSD caches

Bukhtawar:
## Summary

Introduces a two-tier SSD block cache (`TieredBlockCache`) backed by Foyer that physically isolates Parquet metadata from column data on separate SSDs. Metadata and data have independent LRU eviction — data scan pressure never causes metadata eviction.

## Architecture

```
TieredBlockCache
├── metadata Foyer: separate SSD, LRU eviction (metadata-only entries)
└── data Foyer: separate SSD, LRU eviction (column data entries)
```

**Memory model:**
- **Heap** (`file_metadata_cache`): lightweight footer only (schema + RG stats). No page/offset indexes — avoids heap bloat.
- **Metadata Foyer** (SSD): raw bytes of footer + page/offset indexes as ONE global merged range. Survives restart via Foyer disk recovery. Evicts oldest metadata when full (own LRU).
- **Data Foyer** (SSD): column data chunks. Populated at query time. Evicts oldest data when full (own LRU).

## Init Flow

```
df_cache_manager_add_files_with_store(runtime, store_ptr, files)
  │
  ├── fetch_footer_to_heap(): footer via TieredObjectStore → parse → heap cache
  ├── compute_page_index_range(): ONE global merged range (matches parquet crate's
  │   range_for_page_index — folds ALL columns × ALL RGs into single range)
  ├── fetch_ranges_via_store(): page index bytes via TieredObjectStore
  └── ts_put_metadata(): footer + page index bytes → metadata Foyer

On restart: Foyer recovers metadata SSD → get_ranges probes metadata Foyer first → HIT → zero S3.
```

## Query Flow

```
DataFusion ParquetOpener:
  get_metadata() → heap (lightweight footer, zero I/O)
  Row group pruning → RG stats from heap

  Page pruning (if needed):
    load_page_index() → store.get_ranges([global_page_index_range])
      → TieredObjectStore → probe → metadata Foyer HIT

  Column data:
    get_byte_ranges(ranges) → store.get_ranges()
      → probe: metadata Foyer MISS → data Foyer HIT or S3
      → populate: cache.put() → data Foyer
```

## Key Design Decisions

| Decision | Rationale |
|----------|-----------|
| Separate SSDs, both LRU | Metadata and data evict independently — no cross-pressure |
| No routing state | `get()` probes metadata Foyer first, then data Foyer |
| No auto-populate in `get_opts` | Only probes (serves hits). Prevents column data from polluting metadata Foyer |
| Single global page index range | Matches parquet crate's `range_for_page_index()` — ONE range across all RGs |
| `HybridCache::close()` on drop (5s timeout) | Flushes partial blocks to SSD so entries survive restart |
| Footer-only in heap | `PageIndexPolicy::Skip` during warmup. Page indexes in Foyer SSD only |
| `put_metadata()` on BlockCache trait | Default falls back to `put()`; TieredBlockCache routes to metadata Foyer |
| `buffer_pool_size` must be adequately sized | Undersized buffer causes Foyer to silently drop entries |

## Safety Guarantees

| Guarantee | How enforced |
|-----------|-------------|
| Data never enters metadata Foyer | `put()` routes to data Foyer; only `put_metadata()` writes to metadata Foyer |
| Metadata eviction isolated from data | Separate SSDs — data fills/evictions never touch metadata cache |
| Metadata has its own LRU | When metadata SSD fills, oldest metadata evicted (not data) |
| Page indexes not in heap | Warmup uses `PageIndexPolicy::Skip`; indexes only in Foyer SSD |
| Key alignment | Global page index range matches parquet crate's `range_for_page_index()` exactly |
| Restart durability | `close()` on drop flushes partial blocks; Foyer recovers from SSD |

## Test Coverage

**90 tests total, all passing.**

High-confidence end-to-end:
- `production_warmup_then_query_from_cache_only` — warmup → SQL → delete file → SQL from cache
- `restart_with_file_deleted_query_succeeds_from_foyer_only` — cross-session Foyer recovery
- `datafusion_query_succeeds_from_cache_after_local_file_deleted` — key alignment proof

Cache isolation and LRU:
- `data_pressure_does_not_evict_metadata` — fill data cache, metadata untouched
- `metadata_cache_lru_evicts_oldest` — metadata LRU evicts oldest when full
- `data_cache_lru_evicts_oldest_metadata_unaffected` — data LRU evicts, metadata unaffected
- `get_opts_probe_does_not_pollute_metadata_foyer` — column data never enters metadata

Page index correctness:
- `page_index_key_alignment_warmup_matches_query_time` — global range serves from cache
- `page_index_ranges_match_parquet_crate_computation` — fold covers all column ranges

Robustness:
- `concurrent_shard_warmup_does_not_corrupt` — 5 parallel warmups safe
- `metadata_foyer_capacity_breach_graceful_degradation` — no panics under pressure
- `metadata_served_from_ssd_not_local_fs` — cache is the source, not local FS

## Java Integration (follow-up, ~10 lines)

Java warmup calls `df_cache_manager_add_files_with_store` + `ts_put_metadata` instead of `df_cache_manager_add_files`.

### DIFF
--- a/sandbox/plugins/block-cache-foyer/src/main/rust/src/tiered_block_cache.rs
+++ b/sandbox/plugins/block-cache-foyer/src/main/rust/src/tiered_block_cache.rs
@@ -26,7 +26,7 @@
 //! and finds the recovered entries — zero S3 calls for metadata.
 
 use std::sync::Arc;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicU64, Ordering};
 
 use bytes::Bytes;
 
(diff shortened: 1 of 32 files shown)


sandbox/plugins/block-cache-foyer/src/main/rust/src/tiered_block_cache.rs
```diff
@@ -0,0 +1,132 @@
...
+            "[tiered-block-cache] created: data_disk={}B, metadata_disk={}B",
+            data_cache.disk_bytes,
+            metadata_cache.disk_bytes
+        );
+        Self { data_cache, metadata_cache }
+    }
+
+    /// Put bytes into the metadata cache. Called during shard warmup only.
+    ///
+    /// Entries in the metadata cache are never evicted (AdmitAll reinsertion)
```
bharath-techie: should we have a bound for this ? there could be edge cases with wide column parquet files where metadata could be high but i understand this logic goes for a toss
```
 Box::pin(async move {
            // Metadata cache first — small SSD, fast probe, never evicts.
            // On warm restart, Foyer recovers these from disk — instant hit.
            if let Some(bytes) = self.metadata_cache.get(key).await {
                return Some(bytes);
            }
            // Fall through to data cache.
            self.data_cache.get(key).await
        })
```

PR #22241, opened 2026-06-19
Introduce a new collector for filter only queries with size=N

### LINKED ISSUE #22235: [Feature Request] Filter-context queries with size > 0 are unnecessarily score all matching documents(and is slow)

For one of the query below which I was trying to analyze and profile, 

```
```json
{
  "size": 1,
  "track_total_hits": true,
  "query": {
    "bool": {
      "filter": [{
        "bool": {
          "should": [
            { "term": { "field1": "value1" } },
            { "term": { "field2": "value2" } }
          ]
        }
      }]
    }
  }
}
```

Sample profiled response showed all docs were scored which seems unnecessary to me for filter only queries.
Below example only shows the top level query, the underlying children queries didn't have score_count >0

```
{
    "took" : 33,
    "hits" : {
      "total" : { "value" : 48820, "relation" : "eq" },
      "max_score" : 0.0
    },
    "profile" : {
      "shards" : [{
        "searches" : [{
          "query" : [{
            "type" : "BoostQuery",
            "description" : "(ConstantScore(field_a:term_1 field_b:term_2 field_c:term_3))^0.0",
            "breakdown" : {
              "next_doc_count" : 48823,
              "score_count" : 48820, <-- all docs were scored!
              "score" : 498740,
              "set_min_competitive_score_count" : 0
            }
          }],
          "collector" : [{ "name" : "TopScoreDocCollector", "reason" : "search_top_hits" }]
        }]
      }]
    }
  }

```

Seems like it uses `TopScoreDocCollector` which keeps collecting docs with the same score, and does unnecessary work here. 

When I passed size=0, OpenSearch then uses `TotalHitCountCollector` which was very fast(20x faster with the same query) as it does two things:
1. Obviously use `ScoreMode.COMPLETE_NO_SCORES`, so no scoring needed.
2. Overrides  `collect(DocIdStream stream)` to call` stream.count()` ie bulk collecting matching docs instead of individually.

I have suggested one approach below, but happy to listen to other feedback as well.

I think for such queries, we should instead use a different collector to just collect first N docs, and then switch to the mode where we bulk collect the remaining matching documents like size=0 does?

I did some POC by writing a custom collector, index 50k docs and running queries:

```
Results (50K docs, cold first-run, profile:true):

  ┌─────────────────────┬─────────┬───────────┐
  │                │   Before    │   After(custom collector)   
  ├────────────────────────┼─────┼───────────┤
  │ took           │ 33ms        │ 5ms       │
  ├────────────────┼─────────────┼───────────┤
  │ score_count    │ 48,820      │ 0         │
  ├────────────────┼─────────────┼───────────┤
  │ score_time     │ 498,740ns   │ 0ns       │
  ├────────────────┼─────────────┼───────────┤
  │ next_doc_time  │ 5.4ms       │ 4.5ms     │
  ├────────────────┼─────────────┼───────────┤
  │ collector_time │ 1,055,517ns │ 635,467ns │
  └────────────────┴─────────────┴───────────┘

```

### Related component

_No response_

_No response_

### Additional context

_No response_

sgup432:
### Description
- Adds early termination to FilterOnlyCollector so filter-only queries with size=N don't iterate every matching doc.
- Terminates once numHits are collected and the track_total_hits threshold is met 
- Shares a LongAdder across concurrent-search slices so the decision is global, not per-slice.
- Short-circuits later leaves via CollectionTerminatedException in getLeafCollector, avoiding re-collection per segment.

Example where this change helps:

The query below, on an index where 5,000,000 docs match. Note the shape: a single filter clause wrapping a should disjunction, this is what rewrites to BoostQuery(ConstantScore(...), 0f) and reaches FilterOnlyCollector
```
{
  "size": 10,
  "track_total_hits": 10000,
  "query": { "bool": { "filter": [
    { "bool": { "should": [
      { "term": { "field1": "value1" } },
      { "term": { "field2": "value2" } }
    ] } }
  ] } }
}
```

- Before this change: FilterOnlyCollector had no canTerminate(). It collected the first 10 docs, then kept counting the remaining matches via DocIdStream.count() which for a multi-term disjunction iterates across all 5,000,000 matches to compute totalHits, even though the response caps the reported total at 10,000 ("gte").
- After this change: once ~10 docs are collected and 10,000 have been counted, canTerminate() returns true and the collector throws CollectionTerminatedException, skipping the remaining ~4,990,000 matches and any later segments. total is reported as {value: 10000, relation: "gte"}.

### Note:
- This fires only for queries that rewrite to a top-level BoostQuery(ConstantScore(...), 0f) 
- The largest win for this PR is with `track_total_hits: true` (or a large threshold) and many matches, where the full match set would otherwise be counted/scored.

#### Benchmark/Testing:

Ran a local opensearch cluster and indexed different numbers docs on single shard to compare.

Sample query which was run multiple times to compare against this and the baseline:
```
{ "size": 1, "track_total_hits": 10000, "profile": true,
  "query": { "bool": { "filter": [
      { "bool": { "should": [
          { "term": { "field1": "value1" } },
          { "term": { "field2": "value2" } } ] } } ] } } }
```

```
┌────────────┬─────────┬────────────────────────────────────────────────────────┬───────────────────────────────────────────────────┬────────────────────────────────────────┐
│ index size │ matches │ pre-PR (TopScoreDocCollector) score_count / query_time │ PR (FilterOnlyCollector) score_count / query_time │ scoring eliminated (pre-PR score_time) │
├────────────┼─────────┼────────────────────────────────────────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────┤
│ 100K       │ 46,667  │ 46,667 / 3.39ms                                        │ 0 / 2.80ms                                        │ 0.60ms                                 │
├────────────┼─────────┼────────────────────────────────────────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────┤
│ 300K       │ 140,000 │ 140,000 / 9.82ms                                       │ 0 / 8.05ms                                        │ 1.92ms                                 │
├────────────┼─────────┼────────────────────────────────────────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────┤
│ 600K       │ 280,000 │ 280,000 / 20.64ms                                      │ 0 / 17.34ms                                       │ 3.60ms                                 │
├────────────┼─────────┼────────────────────────────────────────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────┤
│ 1M         │ 466,667 │ 466,667 / 36.75ms                                      │ 0 / 30.20ms                                       │ 6.21ms                                 │
├────────────┼─────────┼────────────────────────────────────────────────────────┼───────────────────────────────────────────────────┼────────────────────────────────────────┤
│ 2M         │ 933,333 │ 933,333 / 72.70ms                                      │ 0 / 70.17ms                                       │ 12.13ms                                │
└────────────┴─────────┴────────────────────────────────────────────────────────┴───────────────────────────────────────────────────┴────────────────────────────────────────┘
```

So overall I saw ~20% again in per shard latency with this change.

### Related Issues
Resolves https://github.com/opensearch-project/OpenSearch/issues/22235

### Check List


sgup432:
@rishabhmaurya  Can you take a look at this?

PR #22243, opened 2026-06-19
Add indexing throttle on merge pressure for DataFormatAwareEngine

Shailesh-Kumar-Singh:
When outstanding merges (active + pending) exceed maxMergeCount, the MergeScheduler now activates indexing throttle via the engine's existing IndexingThrottler, serializing write threads to a single thread until merge pressure subsides. This mirrors the behavior already present in InternalEngine's EngineMergeScheduler.

### Description

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/engine/dataformat/MergeInput.java
+++ b/server/src/main/java/org/opensearch/index/engine/dataformat/MergeInput.java
@@ -23,14 +23,14 @@ import java.util.Objects;
  * @opensearch.experimental
  */
 @ExperimentalApi
-public record MergeInput(List<Segment> segments, RowIdMapping rowIdMapping, long newWriterGeneration) {
+public record MergeInput(List<Segment> segments, RowIdMapping rowIdMapping, long newWriterGeneration, double rateLimitMBPerSec) {
 
     public MergeInput {
         segments = List.copyOf(segments);
     }
 
     private MergeInput(Builder builder) {
-        this(new ArrayList<>(builder.segments), builder.rowIdMapping, builder.newWriterGeneration);
+        this(new ArrayList<>(builder.segments), builder.rowIdMapping, builder.newWriterGeneration, builder.rateLimitMBPerSec);
     }
 
     /**
@@ -60,6 +60,7 @@ public record MergeInput(List<Segment> segments, RowIdMapping rowIdMapping, long
         private List<Segment> segments = new ArrayList<>();
         private RowIdMapping rowIdMapping;
         private long newWriterGeneration;
+        private double rateLimitMBPerSec = Double.POSITIVE_INFINITY;
 
         private Builder() {}
 
@@ -104,6 +105,18 @@ public record MergeInput(List<Segment> segments, RowIdMapping rowIdMapping, long
          */
         public Builder newWriterGeneration(long newWriterGeneration) {
             this.newWriterGeneration = newWriterGeneration;
+            return this;
+        }
+
+        /**
+         * Sets the IO rate limit in MB per second for the merge.
+         * A value of Double.POSITIVE_INFINITY means no rate limiting.
+         *
+         * @param rateLimitMBPerSec the rate limit in MB/s
+         * @return this builder
+         */
+        public Builder rateLimitMBPerSec(double rateLimitMBPerSec) {
+            this.rateLimitMBPerSec = rateLimitMBPerSec;
             return this;
         }
 
--- a/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
+++ b/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
@@ -415,6 +415,7 @@ public class MergeScheduler {
                 try {
                     activateThrottling.run();
                 } catch (Exception e) {
+                    isThrottling.set(false);
                     logger.warn("exception in activateThrottling callback", e);
                 }
             }
@@ -424,6 +425,7 @@ public class MergeScheduler {
                 try {
                     deactivateThrottling.run();
                 } catch (Exception e) {
+                    isThrottling.set(true);
                     logger.warn("exception in deactivateThrottling callback", e);
                 }
             }
--- a/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
+++ b/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
@@ -48,8 +48,11 @@ public class MergeScheduler {
     private final MergeHandler mergeHandler;
     private final BiConsumer<MergeResult, OneMerge> applyMergeChanges;
     private final Runnable onMergeFailureCleanup;
+    private final Runnable activateThrottling;
+    private final Runnable deactivateThrottling;
     private final ThreadPool threadPool;
     private final AtomicInteger activeMerges = new AtomicInteger(0);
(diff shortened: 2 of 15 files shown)


server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
rayshrey, line 74: Can we check if these are being used or are they dead code ?
If yes, then let's remove them to avoid any confusion.

rayshrey, line 200: Same comment as the above.

server/src/main/java/org/opensearch/index/engine/dataformat/MergeInput.java
```diff
@@ -23,14 +23,14 @@
  * @opensearch.experimental
  */
 @ExperimentalApi
-public record MergeInput(List<Segment> segments, RowIdMapping rowIdMapping, long newWriterGeneration) {
+public record MergeInput(List<Segment> segments, RowIdMapping rowIdMapping, long newWriterGeneration, double rateLimitMBPerSec) {
```
rayshrey, line 26: How are we handling this value in the Rust side ?
Also does it need to be a doube, shouldn't long/int be enough ?

server/src/test/java/org/opensearch/index/engine/DataFormatAwareEngineTests.java
```diff
@@ -890,6 +890,45 @@ public void testThrottling() throws IOException {
         }
     }
 
+    @SuppressForbidden(reason = "access private fields to simulate merge pressure and verify throttle integration")
+    public void testMergeThrottleIntegration() throws Exception {
```
rayshrey, line 896: Can we take care of the full class path names being used here (import and use just the class names)

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
```diff
@@ -318,6 +343,7 @@ private void executeMerge() {
             try {
                 submitMergeTask(oneMerge);
             } catch (Exception e) {
+                activeMerges.decrementAndGet();
```
rayshrey, line 346: Is this needed ?
We are already handling this in submitMergeTask by catching Exceptions thrown there.

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
```diff
@@ -343,6 +369,7 @@ private void submitMergeTask(OneMerge oneMerge) {
                 // uncaught exception on the merge thread pool.
             } finally {
                 activeMerges.decrementAndGet();
+                evaluateThrottle();
```
rayshrey, line 372: `submitMergeTask` will only be executed in case when the merge is being executed
Will it make sense to move this check to `executeMerge` instead so that this can be called as soon as merges are being added to queue ?

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
Shailesh-Kumar-Singh: Yes it is needed. Because the `activeMerges.incrementAndGet()` at the top of submitMergeTask runs before threadPool.execute(lambda). If execute() throws, the lambda never runs so its finally block never decrements. The catch in executeMerge() compensates for that orphaned increment.

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
Shailesh-Kumar-Singh: No, we need this check here as well.

 >submitMergeTask will only be executed in case when the merge is being executed

We are doing `evaluateThrottle()` in the finally block, which means that merge has already completed and we might need throttle deactivating, hence the check.

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
Shailesh-Kumar-Singh: was used in merge stats and tests, let me check & refractor

server/src/main/java/org/opensearch/index/engine/dataformat/MergeInput.java
Shailesh-Kumar-Singh: it is double in Lucene as well - https://github.com/apache/lucene/blob/main/lucene/core/src/java/org/apache/lucene/index/ConcurrentMergeScheduler.java#L109
Also, we use Double.POSITIVE_INFINITY as the sentinel for "unlimited" everywhere.

sandbox/plugins/parquet-data-format/src/main/rust/src/merge/context.rs
```diff
@@ -95,8 +96,13 @@ impl MergeContext {
         let parquet_root = build_parquet_root_schema(parquet_descriptors)?;
 
         let output_file = File::create(output_path)?;
+        let effective_rate = if rate_limit_mb_per_sec.is_infinite() || rate_limit_mb_per_sec <= 0.0 {
+            0.0
```
rayshrey: Let's add a comment to properly specify that 0 mean rate limiting is disabled.

server/src/main/java/org/opensearch/indices/ClusterMergeSchedulerConfig.java
```diff
@@ -109,10 +109,20 @@ public class ClusterMergeSchedulerConfig {
         Property.NodeScope
     );
 
+    public static final Setting<Double> CLUSTER_IO_RATE_LIMIT_MB_PER_SEC_SETTING = Setting.doubleSetting(
```
rayshrey, line 112: Let's properly document here as well that though this is a cluster level setting, the limit applies per merge only. Also that changing this setting dynamically will only restrict newer merges

mgodwan:
all of this applies only for parquet.
we need to ensure lucene limits are honored as well

server/src/test/java/org/opensearch/index/engine/dataformat/merge/MergeSchedulerTests.java
```diff
@@ -171,7 +186,76 @@ public void testForceMergeAbortsRemainingMergesOnShutdown() throws Exception {
...
+                .put(IndexMetadata.SETTING_VERSION_CREATED, Version.CURRENT)
+                .put(MergeSchedulerConfig.MAX_THREAD_COUNT_SETTING.getKey(), "1")
+                .put(MergeSchedulerConfig.MAX_MERGE_COUNT_SETTING.getKey(), "2")
+                .build()
+        );
+
+        MergeHandler mergeHandler = mock(MergeHandler.class);
+        // After findAndRegisterMerges, 3 pending merges exceed maxMergeCount=2
+        when(mergeHandler.getPendingMergeCount()).thenReturn(3);
+        when(mergeHandler.hasPendingMerges()).thenReturn(false);
```
rayshrey, line 200: Not a valid mocking - mergeHandler will never return a posiive merge count and false for hasPendingMerges.

Let's remove this line. Now that we have added evaluateThrottle in executeMerge, tests should pass without this as well.

rayshrey (approved):
LGTM, left a comment on one of the tests, once that is fixed should be good to merge

PR #22244, opened 2026-06-19
[Arrow Flight RPC] Make FlightServerChannel the sole owner of stream lifecycle and buffers

rishabhmaurya:
### Description

Server-side Arrow Flight streaming frees off-heap Arrow buffers and terminates the gRPC call from **two unsynchronized threads**: the producer runs serially on the flight executor, while a client cancellation (`onChannelCancelled` → `close()`) is delivered concurrently on a gRPC thread. Split ownership between `FlightOutboundHandler` and `FlightServerChannel`, combined with a non-atomic `close()`, made several interleavings unsafe:

- **Double-close → node crash.** `close()` used a non-atomic get-then-set, so two callers (the gRPC cancel path and a `release()`-driven close) could both pass the open-check and both run `notifyCloseListeners()`, double-firing the TaskManager cancellable-channel untrack listener → `AssertionError` (with `-ea`) that kills the node.
- **Use-after-free / leak.** The handler transferred the producer root into the reused stream root and called `putNext` while a concurrent `close()` could free that same root (gRPC may still hold a zero-copy reference), or the producer refilled the reused root after `close()` had already done its single free → stranded buffers.
- **Source-root orphan.** A batch built but never sent (the back-pressure gate threw, or the executor rejected the task) freed nothing.
- **Consumer hang.** A failed batch send dropped the batch with no error frame, so the consumer's `FlightStream.getRoot()` blocked forever.

### Fix — single-writer ownership model

Make `FlightServerChannel` the **sole owner** of every Arrow buffer and of the gRPC stream lifecycle, and keep that ownership safe by **confining the stream root to a single thread** rather than synchronizing access to it:

- **The stream root and every `serverStreamListener` call (`start`/`putNext`/`completed`/`error`) happen only on the channel's single-threaded flight executor.** The executor serializes batch sends, `completeStream`/`sendError`, and the root free among themselves, so none can interleave.
- **`close()` is fire-once and a signal, not a buffer operation.** Via `open.compareAndSet(true, false)`, exactly one caller (any thread) proceeds; it fires close listeners and **posts the root free onto the flight executor**. Because the executor is FIFO, that free runs only after any in-flight send completes. `close()` never touches Arrow buffers or gRPC from the caller (cancel) thread, so it cannot race an in-flight `putNext` and never blocks. **This is deadlock-free by construction** — nothing is ever held across a gRPC/Arrow/blocking call.
- **Single `sendBatch(response, headerSupplier)` entry.** The channel decides native vs byte-serialized internally and builds the header itself; the handler no longer touches a stream root. The native source root is freed exactly once on every exit (`finally`), and a never-handed-off batch is freed via `releaseUnsent` (a `handedOff` discipline keeps the two free sites mutually exclusive). Batches queued behind a cancel reject at the `cancelled` guard and free their own source root.
- **Byte-serialized stream root** is reused across batches and freed once at `close()`, never per batch (fixes a zero-copy use-after-free).
- **`completeStream`/`sendError` are terminal and idempotent** (no-op after close/cancel/terminal), so a failed batch send fails the stream instead of hanging the consumer, and a late terminal op after a cancel is benign.
- Close-listener registration and firing are atomic (a small leaf mutex) and fault-isolated.

**Trade-off (documented):** `close()` frees the stream root asynchronously (on the executor) rather than inline; on node shutdown a dropped free task leaks the root until process exit. Both are called out in `docs/channel-lifecycle-ownership.md` §10.

The full ownership/lifecycle contract — threads, invariants, per-method behavior, and a scenario walk-through (S1–S9) — is documented in `plugins/arrow-flight-rpc/docs/channel-lifecycle-ownership.md`.

### Related Issues

Supersedes the consumer-hang fix in #22117 (that scenario is the S2 path here) and additionally closes the double-close crash, the transfer-vs-cancel use-after-free/leak, and the source-root orphan that #22117 leaves open.

### Testing

- New unit tests: cancel-vs-in-flight-send race (root free runs behind the in-flight send on the executor), **batches queued behind a cancel (each rejects and frees its own source; stream root freed last; no leak)**, fire-once `close()` under concurrent callers, source-root freed on every exit (reject / closed / empty-vectors / header-throw), byte-path reuse freed once, terminal-op idempotency, and the hand-off discipline (gate-throw / executor-reject release).
- `./gradlew :plugins:arrow-flight-rpc:check` — green (tests, internalClusterTest, precommit: spotless, checkstyle, forbiddenApis, loggerUsage, license, javadoc).
- `internalClusterTest`: `NativeArrowTransportIT`, `BackpressureProducerIT`, `StreamMetadataIT`, `FlightTransportIT` — green.

### Check List

### DIFF
--- a/plugins/arrow-flight-rpc/docs/channel-lifecycle-ownership.md
+++ b/plugins/arrow-flight-rpc/docs/channel-lifecycle-ownership.md
@@ -1,0 +1,448 @@
+# FlightServerChannel Lifecycle & Buffer-Ownership Contract
+
+## Purpose
+
+Server-side Arrow Flight streaming frees off-heap Arrow buffers and terminates a gRPC
+call from **multiple unsynchronized threads at once**. Without a single, explicit
+ownership model this produces several classes of bug:
+
+1. **Double-close crash** — `close()` runs its teardown twice, double-firing the
+   TaskManager cancellable-channel untrack listener → `AssertionError` (with `-ea`) that
+   kills the node.
+2. **Stream-root use-after-free / double-free** — a concurrent cancel frees the reused
+   stream root while the producer is mid-`transferRoot` or mid-`putNext` (gRPC still holds
+   a zero-copy reference).
+3. **Stream-root / source-root leak** — buffers transferred into the reused root after a
+   concurrent close, or a source root for a batch whose send never runs, are owned by
+   nobody.
+4. **Consumer hang** — a batch send fails, the batch is dropped, no schema frame ever
+   reaches the client, and the consumer's `FlightStream.getRoot()` blocks forever.
+
+This document is the **authoritative model**. It was validated by an exhaustive
+case-sweep (98 enumerated cases across buffer / thread-interleaving / method-exit /
+lifecycle-state axes; 64 confirmed defects in the current code, collapsing to **12
+distinct root-cause classes** — see §9). The implementation in `FlightServerChannel` and
+`FlightOutboundHandler` MUST conform to it, and §9's coverage matrix MUST stay true.
+
+---
+
+## 1. Threads
+
+All concurrency reduces to **three** thread classes per channel. The distinction between the
+first two is load-bearing: the gap between them is where source roots can orphan (§6 S7).
+
+| Name | What it is | What it runs |
+|------|------------|--------------|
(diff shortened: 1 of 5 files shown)


mch2 (approved):
lgtm, thanks @rishabhmaurya

PR #22247, opened 2026-06-20
[Sandbox] Fix coordinator cancel-teardown leaks in analytics-engine

mch2:
### Description
Three coordinator-side cancellation/teardown leaks, each unit-tested (verified to fail without the fix, pass with it):

- StageExecution cascade: on a CANCELLED child, still closeChildInput(childId) so a parent COORDINATOR_REDUCE drain blocked on streamNext sees EOF and unwinds — without it the reduce thread hangs forever and its borrowed batches/GroupValues are stranded (the QTF nested-reduce leak: a mid-flight shard scan cancelled never EOFs its sender). Cancel is NOT propagated to the parent's state (owner-driven, unchanged). Test: AttachChildrenTests.
- QueryExecution.close: cancel EVERY stage before closing the per-query context, not just the root sink, so non-root reduce stages run their teardown first. Test: QueryExecutionTests.
- RowProducingSink: a feed() racing close() closes the batch instead of buffering it (close already freed+cleared the list and won't run again). Test: RowProducingSinkTests.
- AnalyticsPlugin: analytics.coordinator.buffer_limit default 0 (no per-query child allocator; queries share the coordinator allocator).

### Related Issues

### Check List

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
@@ -39,6 +39,9 @@ import java.util.List;
  * {@code QueryPhaseResultConsumer} for coordinator-reduce in the core
  * search path.
  */
+// TODO: refactor this push-based flow — it's brittle with multiple failure points (feed racing
+// close, partial buffering on early exit). Revisit the locking as part of that (e.g. tryLock with
+// timeout); the current single-monitor synchronization is correct but worth reconsidering then.
 public class RowProducingSink implements ExchangeSink, ExchangeSource {
 
     /**
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
@@ -54,6 +54,8 @@ public class RowProducingSink implements ExchangeSink, ExchangeSource {
     private final List<String> fieldNames = new ArrayList<>();
     private final long maxRows;
     private long totalRows;
+    /** Set by {@link #close}; a {@link #feed} after this frees the batch instead of buffering it. */
+    private boolean closed;
 
     /**
      * Creates a sink with the default row limit.
@@ -71,6 +73,11 @@ public class RowProducingSink implements ExchangeSink, ExchangeSource {
 
     @Override
     public synchronized void feed(VectorSchemaRoot batch) {
+        // Feed racing close(): buffering now would strand the batch, so free it here instead.
+        if (closed) {
+            batch.close();
+            return;
+        }
         if (fieldNames.isEmpty() && batch.getSchema().getFields().isEmpty() == false) {
             for (Field f : batch.getSchema().getFields()) {
                 fieldNames.add(f.getName());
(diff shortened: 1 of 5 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
```diff
@@ -54,6 +54,8 @@ public class RowProducingSink implements ExchangeSink, ExchangeSource {
     private final List<String> fieldNames = new ArrayList<>();
     private final long maxRows;
     private long totalRows;
+    /** Set by {@link #close}; a {@link #feed} after this frees the batch instead of buffering it. */
+    private boolean closed;
```
Bukhtawar, line 61: Should this be volatile?

Bukhtawar (approved):
Thanks lgtm

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
```diff
@@ -71,6 +73,11 @@ public RowProducingSink(long maxRows) {
 
     @Override
     public synchronized void feed(VectorSchemaRoot batch) {
```
expani, line 78: I think we can use explicit ReadWriteLocks instead of the Object level monitor `synchronized`.
It can allow us to avoid deadlocks by not indefinitely waiting on locks to be acquired. 

Maybe we add a TODO and take it as a followup.

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/exec/RowProducingSinkTests.java
```diff
@@ -203,6 +203,26 @@ public void testUnlimitedSinkRetainsAllBatches() {
         sink.close();
     }
 
+    // ─── Feed racing close (leak fix) ────────────────────────────────────
```
expani, line 206: Is it possible to reproduce this using a E2E IT and tracking leaked resources after a query is complete/cancelled/failed ?

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
mch2: closed is read/written only inside synchronized methods

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/exec/RowProducingSinkTests.java
mch2: To a degree yes, its hard to repro the exact leak case in an IT but we can cancel and assert empty.  We have some in SearchCancellationIT that are actually muted bc they are flaky.  Planning on adding more once this set of fixes land, so we can ensure they are non flaky.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/RowProducingSink.java
mch2: We could with the benefit of tryLock with a timeout but I don't think there's a deadlock to avoid here. That said, I'd rather refactor most of this flow, the push-based model is brittle with multiple failure points.

mch2: added a todo

PR #22248, opened 2026-06-20
Wire cooperative cancellation into the reduce and QTF fetch native streams

mch2:
### Description
A cancelled coordinator-reduce or QTF fetch was either abort()-killed mid-send (skipping drop+drain, leaking the aggregate's GroupValues) or left with no cancel signal at all (stranding its stream_next task and DataFusion pool reservation). Make cancel cooperative.

- cross_rt_stream: optional CancellationToken on the producer loop; a cancel select-breaks the loop and falls through to drop(stream)+drain instead of a JoinSet abort. Existing call sites pass None (unchanged). Covered by 3 cargo tests.
- api.rs reduce paths: pass the token, do not register the abort handle.
- query_executor.rs wrap_stream_as_handle (QTF fetch-by-rowid): build with the cancellable variant and register the abort + CPU runtime handles, matching execute_query.
- AnalyticsSearchBackendPlugin.cancelByContext SPI (default no-op) + datafusion impl; AnalyticsSearchService registers a fetch task cancellation listener that fires it.

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/cross_rt_stream.rs
```diff
@@ -90,12 +96,35 @@ impl CrossRtStream {
...
                 }
             }
+            // Drop the inner stream while this future is still polled, then yield so the runtime
+            // reclaims its child producer tasks (drop-time JoinSet::abort only schedules cancel).
+            // This wakes parked sends and lets the children run to completion, freeing GroupValues.
+            drop(stream);
+            for _ in 0..2 {
+                tokio::task::yield_now().await;
+            }
         };
```
Bukhtawar, line 125: would this be sufficient across varying cpu cores?

Bukhtawar (approved):

sandbox/plugins/analytics-backend-datafusion/rust/src/cross_rt_stream.rs
mch2: Its not, this is best effort.  Going to remove this actually its redundant when we have flush_cpu_runtime.  The critical add here is the drop.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/AnalyticsSearchService.java
```diff
@@ -354,6 +354,13 @@ private void drainFetchByRowIds(
             responseHandler.onFailure(new RuntimeException("Failed to execute fetch-by-row-ids on " + shard.shardId(), e));
             return;
         }
+        // On cancel, a fetch parked in the native pull must be released. The fetch path returns an
+        // opaque stream, so fire the backend's cooperative cancellation here (not stream.close(),
+        // which would race the in-flight native pull).
+        final long fetchContextId = task != null ? task.getId() : 0L;
```
expani: Shouldn't we throw `IllegalStateException` if task is null and code reaches here ?

mch2: removed this, task can't ever be null here, added an assert higher up.

PR #22250, opened 2026-06-20
[DFAE] Fix races between merge and force merge

mgodwan:
## Problem

A force merge running concurrently with a background merge could pick the same segments, causing the Lucene secondary to merge a different segment set than Parquet. This produced a row count mismatch between formats:

```
IllegalStateException: Segment [gen=4466] has inconsistent row counts across data formats:
686758681 vs 668676254 (format=lucene)
```

## Root Causes

1. `MergeHandler.findForceMerges()` did not filter out segments already being merged by background merges
2. `LuceneMerger` silently returned an empty result when `IndexWriter.segmentInfos` was transiently empty, allowing parquet to merge successfully while lucene produced nothing
3. `CompositeMergeExecutor` skipped the cross-format row count check when the secondary returned null
4. `MergeScheduler.shutdown()` was never called during engine close, allowing new merges to start during shutdown
5. `forceMerge` had no shutdown check, potentially running long merges on a closing engine

## Fixes

**Race prevention:**
- `MergeHandler.findForceMerges()` now filters out `currentlyMergingSegments` and atomically registers selected segments within the same synchronized block

**Fail-fast integrity checks (production-grade, not assertions):**
- `LuceneMerger`: throws `IOException` when segmentInfos is empty (was: silent return)
- `LuceneMerger`: throws `IllegalStateException` when matched segments < requested (partial match = concurrent consumption)
- `CompositeMergeExecutor`: throws `IllegalStateException` when primary has output but secondary returns null
- `CompositeMergeExecutor`: throws `IllegalStateException` on cross-format row count mismatch
- `CatalogSnapshotManager`: converts three merge invariants from assertions to `IllegalStateException` (source segments exist, generation no-collision, row count conservation)

**Graceful shutdown:**
- `DataFormatAwareEngine.closeNoLock()` now calls `mergeScheduler.shutdown()` as the first step
- `MergeScheduler.forceMerge()` checks `isShutdown` before starting and between merge iterations

## Testing

| Test | Validates |
|------|-----------|
| `MergeHandlerTests.testFindForceMergesExcludesAlreadyMergingSegments` | Race fix: **fails without fix**, passes with it |
| `MergeHandlerTests.testFindForceMergesReturnsNonConflictingCandidates` | Non-conflicting groups still returned |
| `CompositeMergerTests.testExecutorThrowsWhenSecondaryReturnsNullButPrimaryHasOutput` | Null secondary detection |
| `CompositeMergerTests.testExecutorThrowsOnRowCountMismatch` | Cross-format divergence caught |
| `MergeSchedulerTests.testForceMergeSkipsAfterShutdown` | Force merge respects shutdown |
| `MergeSchedulerTests.testForceMergeAbortsRemainingMergesOnShutdown` | Mid-merge shutdown abort |
| `DataFormatAwareEngineTests.testGracefulCloseUnderConcurrentLoad` | Concurrent index/refresh/flush + close |

### Check List

### DIFF
--- a/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMergeExecutor.java
+++ b/sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMergeExecutor.java
@@ -58,7 +58,36 @@ public class CompositeMergeExecutor {
                 : null;
 
             for (DataFormat secondary : plan.secondaryFormats()) {
-                completed.add(mergeFormat(plan, secondary, mapping));
+                FormatMergeResult secondaryResult = mergeFormat(plan, secondary, mapping);
+                // Verify secondary produced output when primary did
+                if (primaryResult.mergedFiles() != null && secondaryResult.mergedFiles() == null) {
+                    throw new IllegalStateException(
+                        "Primary format ["
+                            + plan.primaryFormat().name()
+                            + "] produced merged output but secondary format ["
+                            + secondary.name()
+                            + "] returned null — possible concurrent merge consumed segments"
+                    );
+                }
+                // Verify secondary merged row count matches primary
+                if (primaryResult.mergedFiles() != null && secondaryResult.mergedFiles() != null) {
+                    long primaryRows = primaryResult.mergedFiles().numRows();
+                    long secondaryRows = secondaryResult.mergedFiles().numRows();
+                    if (primaryRows != secondaryRows) {
+                        throw new IllegalStateException(
+                            "Row count mismatch after merge: primary format ["
+                                + plan.primaryFormat().name()
+                                + "] has "
+                                + primaryRows
+                                + " rows but secondary format ["
+                                + secondary.name()
+                                + "] has "
+                                + secondaryRows
+                                + " rows"
+                        );
+                    }
+                }
+                completed.add(secondaryResult);
             }
 
             return toMergeResult(completed, mapping);
--- a/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
+++ b/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
@@ -107,7 +107,22 @@ public class MergeHandler {
             List<Segment> segmentList = catalogSnapshotRef.get().getSegments();
             List<List<Segment>> mergeCandidates = mergePolicy.findForceMergeCandidates(segmentList, maxSegmentCount);
             for (List<Segment> mergeGroup : mergeCandidates) {
-                oneMerges.add(new OneMerge(mergeGroup));
+                boolean hasConflict = false;
+                synchronized (this) {
+                    for (Segment seg : mergeGroup) {
+                        if (currentlyMergingSegments.contains(seg)) {
+                            hasConflict = true;
+                            break;
+                        }
+                    }
+                    if (!hasConflict) {
+                        OneMerge oneMerge = new OneMerge(mergeGroup);
+                        if (registerMerge(oneMerge, false)) {
+                            oneMerges.add(oneMerge);
+                        }
+                    }
+                }
+
             }
         } catch (Exception e) {
             logger.warn("Failed to acquire snapshots", e);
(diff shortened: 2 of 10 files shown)


server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
```diff
@@ -107,7 +107,22 @@ public Collection<OneMerge> findForceMerges(int maxSegmentCount) {
             List<Segment> segmentList = catalogSnapshotRef.get().getSegments();
             List<List<Segment>> mergeCandidates = mergePolicy.findForceMergeCandidates(segmentList, maxSegmentCount);
             for (List<Segment> mergeGroup : mergeCandidates) {
-                oneMerges.add(new OneMerge(mergeGroup));
+                boolean hasConflict = false;
```
RS146BIJAY, line 110: Shouldn't this be done at both the force merge and regular segment merges? As segment merge triggered via force merge should also not be included in regular segment merges.

mgodwan: Its already there in regular merges

RS146BIJAY: Got it. It seems we are calling `findAndRegisterMerges` instead of `findMerges`. But in that case should we make `findMerges` private as we do not intend to expose this outside and any external call to find merges will not exclude currently merging segment as well: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java#L83

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
RS146BIJAY, line 47: Just wondering if we should make this Atomic? Though I can see this is called under only synchronised block, but just wondering if any later changes does make the same assumption.

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
```diff
@@ -155,8 +155,16 @@ public void forceMerge(int maxNumSegment) throws IOException {
...
+                logger.debug("MergeScheduler is shutdown, skipping force merge");
+                return;
+            }
             Collection<OneMerge> oneMerges = mergeHandler.findForceMerges(maxNumSegment);
             for (OneMerge oneMerge : oneMerges) {
+                if (isShutdown.get()) {
+                    logger.debug("MergeScheduler shutdown during force merge, aborting remaining merges");
+                    break;
+                }
                 runMerge(oneMerge);
```
Bukhtawar, line 168: Should we add assertion on the row count inconsistency due to merges to catch similar issues in future

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMergeExecutor.java
```diff
@@ -57,8 +57,42 @@ public MergeResult execute(MergePlan plan) {
                     .orElseThrow(() -> new IllegalStateException("Primary merge did not produce row-ID mapping required by secondaries"))
                 : null;
 
+            // Verify row-ID mapping was produced by primary for secondary consumption
+            if (mapping == null && plan.hasSecondaries()) {
+                throw new IllegalStateException("Primary merge did not produce row-ID mapping required by secondaries");
+            }
+
```
Bukhtawar: nit: seems redundant from the previous check

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
mgodwan: Yes, added at 2 places with this change:
1. CatalogSnapshotManager#applyMergeChanges
2. CompositeMergeExecutor

sandbox/plugins/composite-engine/src/main/java/org/opensearch/composite/merge/CompositeMergeExecutor.java
mgodwan: Ack, removing this

Bukhtawar (approved):
Thanks lgtm, let just verify if the new exception handling would cause 5xx for the force merges, does it retry today?

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
mgodwan: That is used in tests (CompositeMergeTests) which are in different pkg. Hence kept public for now

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
mgodwan: I thought about this while making amendments to the class but didn't do the changes to avoid additional overhead as synchronized block should already ensure the memory effects take place. Let me know your thoughts

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
RS146BIJAY: Got it.

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeHandler.java
RS146BIJAY: Agree this will add additional overhead within synchronised and may not be necessary in current use.

RS146BIJAY (approved):

PR #22252, opened 2026-06-20
Fix FilterDelegationGoldenIT for opensearch-sql nested profile envelope

alchemist51:
### Description

`FilterDelegationGoldenIT` NPEs with `Cannot invoke "java.util.List.iterator()" because "stages" is null` at `shardFragmentStage`.

**Root cause:** the `opensearch-sql` plugin restructured its PPL `profile` response envelope to `{summary, plan, phases}` (SQL #5044, "operator-level metrics"), nesting the analytics `QueryProfile` — including the `stages` array — under the **`plan`** key. The SHARD_FRAGMENT stage list therefore now lives at `profile.plan.stages`, while the test read top-level `profile.stages` → got `null` → NPE on iteration.

The profiling data is **not** lost — it is fully present, just one level deeper. Verified by dumping the actual response:
```
profile = { summary={total_time_ms=...}, plan={ ..., stages=[{execution_type=SHARD_FRAGMENT, chosen_backend=lucene, ...}] }, phases={...} }
```

**Fix:** `shardFragmentStage()` reads `profile.plan.stages`, falling back to the legacy top-level `profile.stages` (the `test-ppl-frontend` explain path serializes the analytics `QueryProfile` directly, without the sql-plugin envelope). This makes the test resilient to both envelope shapes.

**Why it surfaced now:** `sandbox/qa/analytics-engine-rest/build.gradle` pulls `opensearch-sql-plugin:3.8.0.0-SNAPSHOT@zip` as a floating snapshot. Older snapshots emitted the flat shape; newer ones emit the nested envelope, so the failure depends on which snapshot a given build resolves. Pinning that dependency is a sensible follow-up.

### Check List
- [x] Functionality includes testing — `FilterDelegationGoldenIT.testOrDelegatedDelegated` passes locally with the nested-envelope sql-plugin snapshot (BUILD SUCCESSFUL); fallback preserves the flat-shape path.
- [ ] API changes companion pull request created, if applicable. (N/A — test-only change.)
- [ ] Public documentation issue/PR created, if applicable. (N/A.)

### DIFF
--- a/sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/FilterDelegationGoldenIT.java
+++ b/sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/FilterDelegationGoldenIT.java
@@ -266,7 +266,12 @@ public class FilterDelegationGoldenIT extends AnalyticsRestTestCase {
             throw new AssertionError("No 'profile' block in response — request must set profile=true");
         }
         @SuppressWarnings("unchecked")
-        List<Map<String, Object>> stages = (List<Map<String, Object>>) profile.get("stages");
+        Map<String, Object> plan = (Map<String, Object>) profile.get("plan");
+        if (plan == null) {
+            throw new AssertionError("No 'plan' block in profile: " + profile);
+        }
+        @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


bharath-techie (commented):
LGTM

bharath-techie (approved):

PR #22254, opened 2026-06-20
core page index cache changes

bharath-techie:
This change reduces the dominant native heap consumer + compute on wide-schema Parquet files: the full ColumnIndex + OffsetIndex decoded for every column on every query. For example in textbench, 1 billion rows dataset - the full parquet metadata was 1750 MB.

So,  DataFusion's parquet opener decodes the entire page index (per-page string min/max for all columns × all row groups) before caching, then caches the result. That amounts to 1750 MB.

So we try to only cache the parquet metadata footer which consists of file stats and RG stats by default and lazily load and cache the offset index and column index.

**`datafusion.scoped_page_index.enabled` - can be used to switch off column and offset index, after testing, once we confirm that there are no regressions, we can remove this setting altogether - although in few cases, users can also choose to not do this for no performance drops during query phase [happens when all indices are loaded into cache during refresh] **
  
  **Two-cache scoped page index (column/offset index cache)**
  - COLUMN_INDEX_CACHE — cell-granular (file, col, rg) key. Stores only the decoded ColumnIndexMetaData cells for predicate columns; NONE placeholders elsewhere. The per-page string min/max (the heap hog) is
  decoded and stored once per file, shared across every query whose predicate touches that column regardless of projection or literal.
  - OFFSET_INDEX_CACHE — cell-granular (file, col) key. Stores OffsetIndexMetaData (fixed-width page offsets) for predicate ∪ projection ∪ {col 0} columns; single-page safe placeholders elsewhere. Must span
  all row groups (arrow-rs push decoder indexes directly by file-global RG).
  - Backed by BoundedCache<K,V> with Box<dyn CachePolicy> — eviction policy is pluggable (LRU now, S3-FIFO follow-up).
  
  Footer-only level-1 metadata cache
  - load_parquet_metadata now uses PageIndexPolicy::Skip on cache miss — fetches only footer bytes, never page index bytes.
  - strip_page_index in MutexFileMetadataCache::put — safety net for any DataFusion internal path (infer_schema, fetch_statistics) that calls DFParquetMetadata::fetch_metadata with the cache. Strips before
  storing so the level-1 LRU only ever holds footer-only entries.
  - Pre-warming loop in create_session_context — populates the metadata cache footer-only before infer_schema fires, so every DataFusion planning call is a cache hit and never triggers the
  PageIndexPolicy::Optional code path.
  
  Listing-table scan path (scoped_page_index_reader.rs, scoped_index_optimizer.rs)
  - ScopedPageIndexOptimizer — PhysicalOptimizerRule that walks the physical plan and installs ScopedPageIndexReaderFactory on every DataSourceExec backed by a ParquetSource. Provider-agnostic.
  - ScopedPageIndexReaderFactory / ScopedPageIndexReader — on get_metadata, loads footer from the shared cache then augments with a scoped page index. Gated on predicate OR projection being non-empty (fixes
  the match()-only 2.5× bytes_scanned regression).
  
  Indexed-table scan path (indexed_executor.rs)
  - After build_segments, resolves predicate + projection column names → per-file parquet leaf indices in one pass (resolve_predicate_parquet_columns_pair) and augments each segment's metadata with a scoped
  page index via load_scoped_page_index_cols.
  - Both paths share the same process-global caches — a listing query seeds entries that an indexed query hits.
  
Cache size after these changes : 

Level 1 cache : 157 MB [ parquet footer ].
Offset index : 158 MB even after select * type of queries .
Column index : 20-30 MB when 20 columns are used in various queries as filter.

So down to 300 MB from 1750 MB and lazy loading causes evictions to be less painful since only part of the offset index and column index - ~5 MB needs to be loaded back.

  ---
  Follow-ups (not in this PR)
  
  1. Wire surviving_rgs into both scan paths (Step 2) — ScopedPageIndexReaderFactory already carries predicate: Option<Arc<dyn PhysicalExpr>> for this. Indexed executor has it in scope. Both would call
  surviving_row_groups() → load_scoped_page_index_scoped(). Currently load_scoped_page_index_cols (no RG scoping) is used.
  2. S3-FIFO eviction policy — add PolicyType::S3Fifo to cache/eviction_policy.rs, swap the PolicyType::Lru constant in cache/page_index/mod.rs. Zero changes to BoundedCache.
  3. Cache the derived Arrow schema on SegmentFileInfo — parquet_to_arrow_schema is called O(N×RGs) times per file per query across page_pruner.rs, dynamic_filter.rs, and resolve.rs. Store it once on
  SegmentFileInfo at build_segments time and thread it through.
  4. Stats integration
  5. Clear cache integration

### Related Issues

### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_keys.rs
```diff
@@ -0,0 +1,65 @@
...
+/// that column — regardless of the predicate-column *combination* or the
+/// surviving-row-group *set*. (The prior set-keyed design re-decoded and
+/// re-stored a column for every distinct predicate/RG combination — storage grew
+/// with query diversity, not schema width.)
+///
+/// Both scan paths resolve the same `(file, col, rg)` for the same logical
+/// request, so cells are shared across paths → cross-path sharing.
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+pub(crate) struct CiCellKey {
+    pub(crate) path: Arc<str>,
```
alchemist51, line 31: Is this absolute path?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// Byte-bounded cache with a pluggable eviction policy.
+///
+/// `DashMap` shards the value store so concurrent `get` calls on different keys
+/// never contend. The eviction policy (ordering metadata + victim selection) sits
+/// behind a `Mutex` — it is only touched on `insert` and `set_limit`, not on the
+/// read path. Counters are atomics so `stats()` is always lock-free.
+pub(super) struct BoundedCache<K, V>
+where
+    K: Eq + Hash + Clone + Display + Send + Sync + 'static,
```
alchemist51, line 202: Why Hash?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/column_schema_resolver.rs
```diff
@@ -0,0 +1,116 @@
...
+    // union table schema and is intentionally NOT used for index resolution —
+    // see the doc comment.)
+    let file_arrow_schema = match parquet_to_arrow_schema(
+        parquet_schema,
+        metadata.file_metadata().key_value_metadata(),
+    ) {
+        Ok(s) => Arc::new(s),
+        // If we can't derive the file schema, fall back to the union schema; the
+        // caller still falls back to footer-only on any downstream mismatch.
+        Err(_) => return resolve_with_schema(_arrow_schema, metadata, predicate_column_names),
```
alchemist51: Why are we resolving with union here? Isn't that wrong?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/column_schema_resolver.rs
```diff
@@ -0,0 +1,116 @@
...
+/// the same file in one pass. Deriving the per-file arrow schema
+/// (`parquet_to_arrow_schema`) is the dominant cost of name→leaf resolution on
+/// wide schemas (it rebuilds the whole file's Schema); the two callers in the
+/// indexed setup loop previously each rebuilt it, so doing it once here removes a
+/// full schema reconstruction per file per query. Pure refactor — each returned
+/// Vec is identical to calling `resolve_predicate_parquet_columns` separately.
+pub fn resolve_predicate_parquet_columns_pair(
+    union_schema: &SchemaRef,
+    metadata: &ParquetMetaData,
+    names_a: &[String],
```
alchemist51: nit: could be named better

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/column_schema_resolver.rs
```diff
@@ -0,0 +1,116 @@
...
+        Ok(s) => {
+            let file_arrow_schema = Arc::new(s);
+            (
+                resolve_with_schema(&file_arrow_schema, metadata, names_a),
+                resolve_with_schema(&file_arrow_schema, metadata, names_b),
+            )
+        }
+        // Same fallback as the single-name path: resolve against the union schema.
+        Err(_) => (
+            resolve_with_schema(union_schema, metadata, names_a),
```
alchemist51: Not sure again if this is right, let's throw a warning log atleast?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+//!
+//! - **OffsetIndex — projection-driven.** Read at *scan* time for **projected**
+//!   columns (`InMemoryRowGroup::fetch_ranges`, `projection.leaf_included(idx)`),
+//!   and at prune time for the predicate column, and at column 0 for the
+//!   page-skip metric. Key: `(file, projection_cols)` where
+//!   `projection_cols = predicate ∪ projection ∪ {0}`. This is the cheap, fixed-width
+//!   index (no per-page string stats). Built for **all row groups** (an empty
+//!   OffsetIndex on a row group DataFusion scans panics / breaks reads, and
+//!   DataFusion chooses the scanned set itself, after our load — see
+//!   HANDOFF_step2_rg_scoping.md §1e).
```
alchemist51: nit: please give reference of HANDOFF_step2 md file here? or update the comment?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+    surviving_row_groups,
+};
+pub use column_schema_resolver::{
+    resolve_predicate_parquet_columns,
+    resolve_predicate_parquet_columns_pair,
+};
+
+// Process-global caches
+
+pub(crate) static COLUMN_INDEX_CACHE: Lazy<BoundedCache<CiCellKey, ColumnIndexMetaData>> =
```
alchemist51, line 113: Tie them with the global runTime rather?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
```diff
@@ -0,0 +1,1626 @@
...
+/// same footer-stats pruning plus bloom/range/limit, which only remove more), so
+/// scoping the predicate-column ColumnIndex to this set is safe. Returns all row
+/// groups if the predicate can't be lowered or stats are missing. Deterministic
+/// in `(footer_meta, schema, predicate)` → both scan paths agree.
+pub fn surviving_row_groups(
+    footer_meta: &ParquetMetaData,
+    arrow_schema: &SchemaRef,
+    predicate: &Arc<dyn datafusion::physical_expr::PhysicalExpr>,
+) -> Vec<usize> {
+    use arrow::array::{ArrayRef, BooleanArray, UInt64Array};
```
alchemist51: nit: we can move them up?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
```diff
@@ -0,0 +1,1626 @@
...
+        Some(proj_cols) => {
+            let mut set: HashSet<usize> = HashSet::new();
+            set.insert(0); // metric reads column 0
+            for &c in predicate_cols {
+                set.insert(c);
+            }
+            for &c in proj_cols {
+                set.insert(c);
+            }
+            debug_assert!(
```
alchemist51, line 293: Will this take CPU cycles in prod code also?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
```diff
@@ -0,0 +1,1626 @@
...
+                set.iter().all(|&c| c < num_cols),
+                "column index out of bounds (num_cols={num_cols}): {set:?}"
+            );
+            set.into_iter().filter(|&c| c < num_cols).collect()
+        }
+    };
+    if off_cols.is_empty() {
```
alchemist51, line 300: Why are we returning here? Do we not have to place the placeholders?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
```diff
@@ -0,0 +1,1626 @@
...
+fn column_index_union(chunks: &[ColumnChunkMetaData]) -> Option<Range<u64>> {
+    range_union(chunks, |c| {
+        let off = u64::try_from(c.column_index_offset()?).ok()?;
+        let len = u64::try_from(c.column_index_length()?).ok()?;
+        Some(off..off + len)
+    })
+}
+
+/// Union of `offset_index` byte ranges across the given column chunks.
+fn offset_index_union(chunks: &[ColumnChunkMetaData]) -> Option<Range<u64>> {
```
alchemist51, line 460: both column_index_union and this one looks same

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
```diff
@@ -0,0 +1,1626 @@
...
+    let mut fetch_ranges: Vec<Range<u64>> = Vec::with_capacity(num_rgs);
+    for rg_idx in 0..num_rgs {
+        let rg = footer_meta.row_group(rg_idx);
+        let chunks: Vec<ColumnChunkMetaData> = cols.iter().map(|&i| rg.column(i).clone()).collect();
+        let range = offset_index_union(&chunks)?;
+        plans.push(RgPlan { chunks, range_start: range.start });
+        fetch_ranges.push(range);
+    }
+
+    let buffers = store.get_ranges(location, &fetch_ranges).await.ok()?;
```
alchemist51, line 417: How are we stich the query cancellation framework with this? Just want to understand in case the store is lagging or stuck, we cancel the query, how will it propagate to here? Or that is something already handled in the task managers?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
```diff
@@ -0,0 +1,195 @@
...
+use crate::cache::eviction_policy::{CachePolicy, PolicyType, create_policy};
+
+/// Default byte budget for EACH scoped cache, used until the caller sets one from
+/// the runtime's configured limit (see [`set_column_index_cache_limit`] /
+/// [`set_offset_index_cache_limit`]). The two caches are budgeted independently:
+/// the ColumnIndex (per-page string min/max) is the heavy one and the OffsetIndex
+/// (fixed-width page offsets) is tiny, so they get separate, separately-tunable
+/// limits rather than sharing one number.
+///
+/// TODO : configure via settings
```
expani: IIUC This is global for the node which is at par with DocValueSkipList OR any other Index file stored in OS File page cache for Vanilla OpenSearch. Can you record this in a comment explicitly ?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_keys.rs
bharath-techie: its store relative path - but key here is arc<str> - its just a reference to the object_path

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
bharath-techie: Required by dashmap - needs both eq and hash for keys.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/column_schema_resolver.rs
bharath-techie: there is followup for resolve in general, we can handle it there.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
expani: Only if prod code ( dylib ) is built using the dev profile. Release profile has `debug-assertions = false` by default if not explicitly set in Cargo.toml

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/mod.rs
bharath-techie: This is a custom cache right? Global runtime doesn't have support for caches other than the list files, stats and metadata cache ?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
bharath-techie: No in release builds these are removed.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_keys.rs
expani: Will this be an issue when Warm Nodes refer to the files locally v/s in S3 ? Or is it abstracted away by S3 Filesystem ?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
bharath-techie: None falls back to footer only. This only happens when there are technically no columns in parquet which shouldn't happen at all.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
bharath-techie: column_index_offset vs offset_index_offset but otherwise the same only

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
bharath-techie: Here, we either insert to cache or don't right ? There is no possibility of leak here

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_keys.rs
bharath-techie:
```
Indexed path:
  Java passes file paths via ShardView.object_metas (ObjectMeta[])
    ↓
  build_segments() — reads meta.location for:
    - load_parquet_metadata()         → metadata cache key
    - segment.object_path             → stored on SegmentFileInfo
    ↓ 
  SegmentFileInfo.object_path used by:
    - load_scoped_page_index_cols()   → page index cache key (Arc<str>)
    - CachedMetadataReaderFactory     → passed to ParquetSource for actual data reads
    - IndexedExec.object_path         → passed to RowGroupStreamConfig for parquet IO
  
  Listing path:
  DataFusion list_files_cache (pre-populated from shard_view.object_metas)
    ↓
  ListingTable.list_files_for_scan() → PartitionedFile.object_meta.location
    ↓
  Same location used by:
    - DFParquetMetadata.fetch_metadata()  → metadata cache key
    - ScopedPageIndexReader.location      → page index cache key (Arc<str>)
    - CachedParquetFileReader.location    → actual parquet data reads
    - ParquetFileMetrics                  → metrics labeling
```
So for overall query we are already using the file location, here just reusing with an arc reference.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/column_schema_resolver.rs
bharath-techie: returning empty

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/custom_cache_manager.rs
```diff
@@ -351,42 +352,20 @@ impl CustomCacheManager {
...
-
-            // fetch_metadata() performs the cache put operation internally
-            df_metadata.fetch_metadata().await
-                .map_err(|e| format!("Failed to fetch metadata: {}", e))
+        // Warm the level-1 metadata cache footer-only. `load_parquet_metadata`
+        // fetches with PageIndexPolicy::Skip — only footer bytes, no page index IO.
+        // On success the entry is in the cache; on failure the error propagates.
+        let location = object_meta.location.clone();
+        rt_handle.block_on(async {
+            parquet_bridge::load_parquet_metadata(store, &location, metadata_cache).await
```
Bukhtawar: metadata_cache_put already fetched and parsed the footer, can the same IO be reused for populating statistics cache?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
```diff
@@ -0,0 +1,195 @@
...
+use std::sync::Mutex;
+
+use dashmap::DashMap;
+
+use crate::cache::eviction_policy::{CachePolicy, PolicyType, create_policy};
+
+/// Default byte budget for EACH scoped cache, used until the caller sets one from
+/// the runtime's configured limit (see [`set_column_index_cache_limit`] /
+/// [`set_offset_index_cache_limit`]). The two caches are budgeted independently:
+/// the ColumnIndex (per-page string min/max) is the heavy one and the OffsetIndex
```
Bukhtawar: Should column index for non-string have a separate limit, given they would normally get used in predicates and their sizes would usually be small? Do we need to protect them from getting thrashed by string column index

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
```diff
@@ -0,0 +1,195 @@
...
+impl<K, V> BoundedCache<K, V>
+where
+    K: Eq + Hash + Clone + Display + Send + Sync + 'static,
+    V: Clone + Send + Sync + 'static,
+{
+    pub(super) fn new(limit: usize, policy_type: PolicyType) -> Self {
+        Self {
+            map: DashMap::new(),
+            reverse: DashMap::new(),
+            policy: Mutex::new(create_policy(policy_type)),
```
Bukhtawar: Good to avoid map and reverse with a generic `CachePolicy<K>`

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
```diff
@@ -0,0 +1,1630 @@
...
+use object_store::ObjectStore;
+use parquet::file::page_index::offset_index::OffsetIndexMetaData;
+use prost::bytes::{buf, Buf, Bytes};
+
+use super::cache_keys::{CiCellKey, OiCellKey, OiColumn};
+use super::{COLUMN_INDEX_CACHE, OFFSET_INDEX_CACHE};
+
+/// Load + graft a scoped page index: ColumnIndex for `predicate_cols` (all RGs),
+/// OffsetIndex for all columns/all RGs. The Step-1 baseline.
+pub async fn load_scoped_page_index(
```
Bukhtawar: Sorry where is this wired outside tests?

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
expani: Should the setting be done in this PR itself ? Otherwise, some benchmarks might show regressions depending on the time it takes to load the cache during a miss.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub limit_bytes: usize,
+}
+
+/// Byte-bounded cache with a pluggable eviction policy.
+///
+/// `DashMap` shards the value store so concurrent `get` calls on different keys
+/// never contend. The eviction policy (ordering metadata + victim selection) sits
+/// behind a `Mutex` — it is only touched on `insert` and `set_limit`, not on the
+/// read path. Counters are atomics so `stats()` is always lock-free.
+pub(super) struct BoundedCache<K, V>
```
expani: I think using the Foyer In-Memory cache has potential here.

`evict_by_prefix` and Custom Cache stats especially for `evictions` can be surfaced using a Wrapper around `foyer::Cache` and we can use `foyer::EventListener` for any custom tracking that is NOT provided out of the box. 

Foyer would also give us S3-FIFO out of the box.

`eviction_policy.rs` can be removed AND Statistics Cache can use Foyer as well.
Also, the `select_for_eviction` is not optimal as it takes O(NLogN) compared to what Foyer does using Linked lists which is O(1)

This might be a bigger migration, but worth taking a look if not done already.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
```diff
@@ -0,0 +1,195 @@
...
+            evictions: AtomicU64::new(0),
+            used_bytes: AtomicUsize::new(0),
+        }
+    }
+
+    pub(super) fn get(&self, key: &K) -> Option<V> {
+        match self.map.get(key) {
+            Some(entry) => {
+                let size = entry.1;
+                if let Ok(mut p) = self.policy.lock() {
```
expani: The policy lock will make all the threads block here for get, insert and evictions which means the concurrency of DashMap gets NULLIFIED by this global lock.

Can we make the eviction policy thread safe instead since it already uses a DashMap + AtomicUsize internally? 

The &mut self on the CachePolicy methods is what forces the outer Mutex and the only thing that actually needs &mut is set_policy swapping the whole box, and that's only used in TESTS

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/metadata_cache.rs
```diff
@@ -0,0 +1,344 @@
...
+use log::error;
+use object_store::path::Path;
+
+// Cache type constants
+pub const CACHE_TYPE_METADATA: &str = "METADATA";
+pub const CACHE_TYPE_STATS: &str = "STATISTICS";
+
+// Helper function to log cache operations
+fn log_cache_error(operation: &str, error: &str) {
+    error!("[CACHE ERROR] {} operation failed: {}", operation, error);
```
expani: Why are we not using the Logger Bridge instead ?

sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
```diff
@@ -229,6 +230,15 @@ pub async unsafe fn create_session_context(
             );
     }
 
+    // Install the scoped page-index reader factory on every parquet scan.
+    // Registered AFTER ProjectRowIdOptimizer so it sees the final DataSourceExec.
```
expani, line 235: Please also mention that THIS SHOULD be always last given Projection Pushdown is another essential thing it depends on.

If that happens after, we end up trying to cache all columns.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
```diff
@@ -0,0 +1,1630 @@
...
+    // them in the matrix, and populate the cache.
+    if !missing_col_rg_matrix.is_empty() {
+        let built = build_column_index_cells(store, location, footer_meta, &missing_col_rg_matrix).await?;
+        for cell in built {
+            debug_assert!(
+                cell.rg < col_index_matrix.len() && cell.col < col_index_matrix[cell.rg].len(),
+                "cell ({}, {}) out of matrix bounds ({num_rgs} rgs, {num_cols} cols)",
+                cell.col, cell.rg,
+            );
+            COLUMN_INDEX_CACHE.insert(
```
expani: Can we do a pre-check here based on the number of columns ? 

Say a query has 10 columns ( Predicate + Projection ) and only a few can eventually fit in the cache.  By just plain inserting, we will evict one entry after another until the last column is inserted. 

We can instead try to insert only a few columns that neatly fit into the current available space. They can contend with some other parallel query but will still be a best effort.

expani (commented):
Thanks for taking this up @bharath-techie 

Still going through the changes but dropping some early comments. 

Consider if we need a feature gate for this change AND we should track exposing the time spend in cache related things in the final query profile as well.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
bharath-techie: This is column index used for min,max stats, and so we will insert only the predicate columns , and its lightweight

expani: My point is when those predicate columns are a lot for just one query then we end up thrashing the cache with inserts and evictions. 
Maybe we need a Bulk put to avoid locking per predicate OR remove the lock altogether.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
bharath-techie: The rust docs are not correct here , will correct it. Basically col index is light weight and string cols if they get delegated are not cached, and also its only for predicates. offset index is heavy as they contain , predicates, projections and first column.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
bharath-techie: i mean that's the plan , we've merged the stats, settings PR first so that we can integrate here and have ITs as well. will update shortly.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
bharath-techie: I think batch makes sense for both caches, implemented it.

bharath-techie: Even without a lock, it helps in doing one eviction call, so batch helps.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
bharath-techie: Will evaluate as part of https://github.com/opensearch-project/OpenSearch/pull/22146 cc : @alchemist51

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/custom_cache_manager.rs
bharath-techie: nice idea man. implemented it.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/cache_store.rs
bharath-techie: Moved to FIFO, implemented arc swap for cache swap , removed mut self everywhere.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/page_index_io.rs
bharath-techie: Removed redundant methods and code.

sandbox/plugins/analytics-backend-datafusion/rust/src/cache/page_index/mod.rs
bharath-techie: I've grouped them all together under cacheType.

Bukhtawar (approved):
Thanks lgtm, lets see how we could upstream the changes

expani (approved):
Thanks for the changes @bharath-techie This is a great start to lazily load the Metadata cache only for columns required by the queries. 

It was a great learning experience on DF internals while reviewing it. Let's keep this going 🚀

PR #22259, opened 2026-06-21
Fix: OpenSearch process does not exit when startup fails due to StartupException

aparajita31pandey:
## Description

When a plugin or bootstrap component  can throws a `RuntimeException` during startup, `OpenSearch.init()` wraps it in a `StartupException` and rethrows it. `StartupException` was never caught in the call chain — it escaped `OpenSearch.main(String[])` entirely, bypassing the `exit(status)` call. 

This lefts the JVM process hanging indefinitely after startup failure.

## Resolves 
https://github.com/opensearch-project/OpenSearch/issues/22260

## Root Cause

`StartupException` is a `RuntimeException`. The CLI framework in `Command.main()` only handles `OptionException` and `UserException` — so `StartupException` propagates uncaught all the way through `main(String[], OpenSearch, Terminal)` and escapes `main(String[])`. When non-daemon threads are still alive (as they are during a partial bootstrap), the JVM does not terminate automatically, leaving the process hanging.

```java
// main(String[]) — StartupException escapes here, exit() is never reached
int status = main(args, opensearch, Terminal.DEFAULT);
if (status != ExitCodes.OK) {
    ...
    exit(status);  // never reached
}
```

## Fix

`StartupException` **designed** to escape to `main()` — its `printStackTrace()` override has a comment: *"This logic actually prints the exception to the console, its what is invoked by the JVM when we throw the exception from main()"*. The formatted output was already correct. The only missing piece was `exit()`.

Catch `StartupException` in `main(String[])` — the correct level where `System.err` is appropriate and process exit decisions belong — call `e.printStackTrace(System.err)` to preserve the existing custom-formatted output (which truncates guice frames, etc.), then `exit(CODE_ERROR)`:

```java
try {
    status = main(args, opensearch, Terminal.DEFAULT);
} catch (StartupException e) {
    e.printStackTrace(System.err);
    exit(ExitCodes.CODE_ERROR);
    return;
}
```

## Stack trace (reproducer)

```
org.opensearch.bootstrap.StartupException: java.lang.RuntimeException: java.lang.RuntimeException: dummy
    at org.opensearch.bootstrap.OpenSearch.init(OpenSearch.java:173)
    at org.opensearch.bootstrap.OpenSearch.execute(OpenSearch.java:160)
    at org.opensearch.common.cli.EnvironmentAwareCommand.execute(EnvironmentAwareCommand.java:110)
    at org.opensearch.cli.Command.mainWithoutErrorHandling(Command.java:146)
    at org.opensearch.cli.Command.main(Command.java:101)
    at org.opensearch.bootstrap.OpenSearch.main(OpenSearch.java:126)
```

Signed-off-by: Aparajita Pandey <aparajita31pandey@gmail.com>

### DIFF
--- a/server/src/main/java/org/opensearch/bootstrap/OpenSearch.java
+++ b/server/src/main/java/org/opensearch/bootstrap/OpenSearch.java
@@ -159,6 +159,8 @@ class OpenSearch extends EnvironmentAwareCommand {
             init(daemonize, pidFile, quiet, env);
         } catch (NodeValidationException e) {
             throw new UserException(ExitCodes.CONFIG, e.getMessage());
+        } catch (StartupException e) {
+            throw new UserException(ExitCodes.CODE_ERROR, e.getMessage());
         }
     }
 
--- a/server/src/main/java/org/opensearch/bootstrap/OpenSearch.java
+++ b/server/src/main/java/org/opensearch/bootstrap/OpenSearch.java
@@ -88,7 +88,17 @@ class OpenSearch extends EnvironmentAwareCommand {
 
         LogConfigurator.registerErrorListener();
         final OpenSearch opensearch = new OpenSearch();
-        int status = main(args, opensearch, Terminal.DEFAULT);
+        int status;
+        try {
+            status = main(args, opensearch, Terminal.DEFAULT);
+        } catch (StartupException e) {
+            // StartupException has custom printStackTrace formatting (truncates guice frames, etc.).
+            // Catch it here so the process exits rather than hanging, while preserving that output.
+            e.printStackTrace(System.err);
+            exit(ExitCodes.CODE_ERROR);
+            return;
+        }
+
         if (status != ExitCodes.OK) {
             final String basePath = System.getProperty("opensearch.logs.base_path");
             // It's possible to fail before logging has been configured, in which case there's no point
@@ -159,8 +169,6 @@ class OpenSearch extends EnvironmentAwareCommand {
             init(daemonize, pidFile, quiet, env);
         } catch (NodeValidationException e) {
             throw new UserException(ExitCodes.CONFIG, e.getMessage());
-        } catch (StartupException e) {
-            throw new UserException(ExitCodes.CODE_ERROR, e.getMessage(), e);
         }
     }
 
--- a/server/src/main/java/org/opensearch/bootstrap/OpenSearch.java
+++ b/server/src/main/java/org/opensearch/bootstrap/OpenSearch.java
@@ -88,16 +88,7 @@ class OpenSearch extends EnvironmentAwareCommand {
 
         LogConfigurator.registerErrorListener();
         final OpenSearch opensearch = new OpenSearch();
-        int status;
-        try {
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/opensearch/bootstrap/OpenSearch.java
```diff
@@ -88,7 +88,17 @@ public static void main(final String[] args) throws Exception {
...
+        int status;
+        try {
+            status = main(args, opensearch, Terminal.DEFAULT);
+        } catch (StartupException e) {
+            // StartupException has custom printStackTrace formatting (truncates guice frames, etc.).
+            // Catch it here so the process exits rather than hanging, while preserving that output.
+            e.printStackTrace(System.err);
+            exit(ExitCodes.CODE_ERROR);
+            return;
+        }
```
msfroh: [Nitpick] Would it make sense to do something more like:

```
try {
  status = main(args, opensearch, Terminal.DEFAULT);
} catch (StartupException e) {
  // StartupException has custom printStackTrace formatting (truncates guice frames, etc.).
  // Catch it here so the process exits rather than hanging, while preserving that output.
  e.printStackTrace(System.err);
  status = ExitCodes.CODE_ERROR;
  // Continue to the next if-statement to exit().
}
```

What do you think? It's essentially the same thing (and the same number of lines), but part of me likes having exactly one place where we call `exit()`.

I guess another option would be to modify [`OpenSearch.main(String[], OpenSearch, Terminal)`](https://github.com/opensearch-project/OpenSearch/blob/18b0d1c7dbb756e52274cad2098664bdef7cacc5/server/src/main/java/org/opensearch/bootstrap/OpenSearch.java#L124) to be:

```
    static int main(final String[] args, final OpenSearch opensearch, final Terminal terminal) throws Exception {
        try {
            return opensearch.main(args, terminal);
        } catch (StartupException e) {
            e.printStackTrace(terminal.getErrorWriter());
            return ExitCodes.CODE_ERROR;
        }
    }
```

Incidentally, I think the `e.printStackTrace(System.err)` in your solution should at least be `e.printStackTrace(Terminal.DEFAULT.getErrorWriter())`. I just read up on `System.console()` (since the `ConsoleTerminal` is the one that differentiates from `SystemTerminal`, which just delegates to `System.out` and `System.err`), since I was unfamiliar with it. I *think* the essential piece is that `Console` synchronizes its output (and input) methods, so at least a `println` is guaranteed to be atomic. (I guess lines of a stack trace could get interleaved with other lines.) It looks like localization may also be affected (if the console's locale differs from the system's locale).

aparajita31pandey: went with option 2. moved the catch into the 3-arg main so it uses the injected terminal instead of `Terminal.DEFAULT` and `main(String[])` keeps a single `exit()` call. 

Didn't know about the Console sync and locale difference between `System.err` and `terminal.getErrorWriter() . Good to know

msfroh (approved):
Nice! Thank you @aparajita31pandey!

It's so much better to fail explosively when a plugin fails to initialize properly than just hanging.

PR #22269, opened 2026-06-22
Remove duplicate versionMap after refresh call.

thorkous:
### Description

`DataFormatAwareEngine` registers `versionMap` as a `RefreshListener` in its constructor, so `notifyRefreshListenersAfter(refreshed)` already drives `versionMap.afterRefresh(refreshed)` as part of the listener fan-out. The `refresh()` method then called `versionMap.afterRefresh(refreshed)` a second time inline in the `finally` block, double-invoking `afterRefresh` on every successful refresh.

This change removes the redundant inline `versionMap.afterRefresh(refreshed)` call, leaving the single invocation that flows through the standard refresh-listener notification path. This matches `InternalEngine`, where the version map's `afterRefresh` is driven solely through the registered listener rather than being called directly.

No behavior change beyond eliminating the duplicate call; the version map is still advanced exactly once per refresh via the listener.

### Related Issues
N/A

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
+++ b/server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
@@ -1112,7 +1112,6 @@ public class DataFormatAwareEngine implements Indexer {
                 }
             } finally {
                 notifyRefreshListenersAfter(refreshed);
-                versionMap.afterRefresh(refreshed);
                 IOUtils.close(toClose);
                 refreshLock.unlock();
             }


server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1106,7 +1106,6 @@ public void refresh(String source) throws EngineException {
                     }
                 }
             } finally {
-                versionMap.afterRefresh(refreshed);
```
RS146BIJAY, line 1115: Why this was not caught by test cases? Should we add a test case for this scenario?

thorkous: because the duplicate is idempotent, no observable behavior changes, so a black-box test can't distinguish one call from two.

RS146BIJAY: > no observable behavior changes

I am not sure this is correct. Consider the case that afterRefresh is called by `notifyRefreshListenersAfter`. Then there are some entry made in version map due to ingestion. Then you again called [`versionMap.afterRefresh(refreshed)`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/LiveVersionMap.java#L288) which will [replace new maps while setting old map as empty](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/LiveVersionMap.java#L182).  This will loose all the new entries that came in between the two afterRefresh calls.

RS146BIJAY: Also just thinking if any assertion like this will also make sense here: https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/shard/IndexShard.java#L5520

thorkous: > I am not sure this is correct. Consider the case that afterRefresh is called by notifyRefreshListenersAfter. Then there are some entry made in version map due to ingestion. Then you again called [versionMap.afterRefresh(refreshed)](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/LiveVersionMap.java#L288) which will [replace new maps while setting old map as empty](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/index/engine/LiveVersionMap.java#L182). This will loose all the new entries that came in between the two afterRefresh calls.

After refresh only invalidates old map. In first afterRefresh calls, only the old map is marked empty and until we call beforeRefresh the old map stays empty. So second afterRefresh calls just keeps it empty and no data will be lost.

RS146BIJAY: Got it. My bad. Actually rotation happens in beforeRefresh. Make sense.

RS146BIJAY (approved):

thorkous:
Thanks @RS146BIJAY for merging the PR.

PR #22257, opened 2026-06-20
Changes in Datafusion plugin to stitch clear cache rest action, cache settings and cache stats

bharath-techie:
#### Dynamic settings for cache limits
  
Adjusts the native memory partitioning to carve out a dedicated 3% budget for DataFusion's parquet metadata caches:
  
```
  node.native_memory.limit = 80% of off-heap  (was 79%, +1% from unmanaged headroom)
    ├── 71% → datafusion.memory_pool_limit_bytes   (operator pool, was 74%)
    ├──  3% → datafusion.metadata_index_cache.total_size_bytes  (new)
    │     ├── 48% → footer metadata cache  (parquet footers + row-group stats)
    │     ├── 34% → offset index cache     (projection-driven page offsets)
    │     └── 13% → column index cache     (predicate-driven page min/max)
    |    └── 5% → file stats cache     (predicate-driven page min/max)
    ├──  8% → Arrow ingest pool
    ├──  5% → Arrow flight pool
    ├──  5% → Arrow query pool
    ├──  5% → Parquet write pool
    └──  3% → Parquet merge pool
    = 100%
```
  
  The 3% is funded by reducing the datafusion pool from 74→71% and expanding node.native_memory.limit from 79→80% of off-heap. Percentages within the cache budget must sum to ≤100 (unused headroom is accepted, matching Arrow's sum(max) ≤ budget model).  
Cluster setting updates get rejected if the new sum exceeds 100.

```
Total RAM:    64 GB
  JVM heap:     16 GB
  Off-heap:     48 GB

  OLD:
    node.native_memory.limit = 79% × 48 = 37.92 GB
    Unmanaged                = 21% × 48 = 10.08 GB
    DataFusion: 74% × 37.92 = 28.06 GB

  NEW:
    node.native_memory.limit = 80% × 48 = 38.40 GB  (+0.48 GB, 1% of off-heap from unmanaged)
    Unmanaged                = 20% × 48 =  9.60 GB (-0.48 GB)
    DataFusion: 71% × 38.40 = 27.26 GB  (-0.80 GB)

    71% → DataFusion pool  = 27.26 GB
    3%  → cache total      =  1.15 GB
    8%  → Arrow ingest     =  3.07 GB
    5%  → Arrow flight     =  1.92 GB
    5%  → Arrow query      =  1.92 GB
    5%  → Parquet write    =  1.92 GB
    3%  → Parquet merge    =  1.15 GB
    = 100%

  Unmanaged: 20% × 48 = 9.60 GB
[...]

PR #22273, opened 2026-06-22
Reducing the ingestion/write memory limit by 50% of original value

nssuresh2007:
### Description
Originally, Arrow ingest pool had 8% and Parquet write pool had 5% of the `node.native_memory.limit` setting. Since warm nodes would not get more write traffic, we are reducing both the above limits by 50% and dedicate that additional size to `datafusion.metadata_index_cache.total_size` which would increase from 3% to 9%.
Refer #22257 for the original size limits

### Related Issues
None

### Check List

### DIFF
--- a/plugins/arrow-base/src/main/java/org/opensearch/arrow/allocator/ArrowBasePlugin.java
+++ b/plugins/arrow-base/src/main/java/org/opensearch/arrow/allocator/ArrowBasePlugin.java
@@ -11,6 +11,7 @@ package org.opensearch.arrow.allocator;
 import org.opensearch.arrow.spi.NativeAllocatorPoolConfig;
 import org.opensearch.arrow.spi.PoolGroup;
 import org.opensearch.cluster.metadata.IndexNameExpressionResolver;
+import org.opensearch.cluster.node.DiscoveryNode;
 import org.opensearch.cluster.node.DiscoveryNodes;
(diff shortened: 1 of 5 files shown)


plugins/arrow-base/src/main/java/org/opensearch/arrow/allocator/ArrowBasePlugin.java
```diff
@@ -125,19 +125,19 @@ public ArrowBasePlugin() {}
         Setting.Property.Dynamic
     );
 
-    /** Minimum guaranteed bytes for the ingest pool. Default is 4% of budget. */
+    /** Minimum guaranteed bytes for the ingest pool. Default is 2% of budget. */
     public static final Setting<Long> INGEST_MIN_SETTING = new Setting<>(
         NativeAllocatorPoolConfig.SETTING_INGEST_MIN,
-        s -> derivePoolMinDefault(s, 4),
+        s -> derivePoolMinDefault(s, 2),
```
nishchay21: We should have this only for warm correct ?

nssuresh2007: Updated

Bukhtawar (approved):
Thanks Lgtm

PR #22149, opened 2026-06-14
Changes to spawn object store reads in IO runtime in datafusion plugin

bharath-techie:
### Description
[Describe what this change achieves]

### Related Issues
Resolves #[Issue number to be closed when this PR is merged]

### Check List
- [ ] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #22280, opened 2026-06-23
Switch to IO runtime for warm

KhishorekumarBS:
### Description
Rebased PR https://github.com/opensearch-project/OpenSearch/pull/22149

Moves object-store network IO (S3, GCS, Azure) off the CPU executor threads and onto the dedicated IO runtime. This prevents HTTP response-body streaming from blocking DataFusion's CPU-bound query processing threads.

1. native_bridge_common::io_runtime — New process-global IO runtime handle slot (set_io_handle / clear_io_handle / io_handle). Lives in native-bridge-common so all native-repository-* crates can access it without depending on opensearch-datafusion.
2. RuntimeManager — Publishes the IO runtime handle on startup and clears it on shutdown. Adds a separate coordinator concurrency gate to avoid deadlock between shard streams and coordinator reduce.
3. native-repository-s3/gcs/azure — Each object-store builder checks for the global IO handle and, when present, installs a SpawnedReqwestConnector that dispatches HTTP IO onto the IO runtime.
4. Cleanup — Removes the old thread-local register_io_runtime / io.rs / SpawnIoStore wrapper approach in favor of the simpler global handle pattern.

### Check List
- [Yes] Functionality includes testing.
- [NA] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [NA] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.


sandbox/plugins/analytics-backend-datafusion/rust/src/runtime_manager.rs
```diff
@@ -69,6 +71,9 @@ impl RuntimeManager {
 
     pub fn shutdown(&self) {
         info!("Shutting down RuntimeManager");
+        // Clear the published IO handle so a torn-down runtime is never handed
+        // out to a remote object-store builder after shutdown.
+        native_bridge_common::io_runtime::clear_io_handle();
```
alchemist51, line 79: Why native bridge common is tied to analytics backend plugin?

sandbox/plugins/analytics-backend-datafusion/rust/src/runtime_manager.rs
```diff
@@ -33,23 +32,26 @@ impl RuntimeManager {
                 .expect("Failed to create IO runtime"),
         );
 
-        register_io_runtime(Some(io_runtime.handle().clone()));
+        // Publish the IO runtime handle in native_bridge_common so the separate
+        // native-repository-* crates (s3/gcs/azure) can install a
+        // SpawnedReqwestConnector that runs HTTP IO on this runtime.
+        native_bridge_common::io_runtime::set_io_handle(io_runtime.handle().clone());
```
bharath-techie, line 41: Can we document the order of initialization here - basically as @alchemist51 has pointed out below, we are setting io runtime handle here which is getting used in a different plugin.

KhishorekumarBS: Added and revised.

sandbox/plugins/analytics-backend-datafusion/rust/src/runtime_manager.rs
KhishorekumarBS: From my understanding, native-bridge-common is the only shared crate that both the analytics plugin and the native-repository-* plugins depend on. Added a code comment explaining this dependency

alchemist51 (approved):

bharath-techie (approved):
LGTM

Bukhtawar (commented):
Curious how does global queue/local queue distribution looks with this change?

PR #22301, opened 2026-06-24
Integrated RelFieldTrimmer for Projection Pusdown in different query shapes

expani:
### What
Narrow the per-shard scan to only the columns a query needs by inserting a Project using RelFieldTrimmer, instead of reading every column and eliminating at the coordinator. This let's Datafusion's Physical Optimizer kick in and reduce the amount of columns scanned.

### Why
ClickBench q25/q27 where the output is a subset of the sort/filter columns) read all ~100 columns on each shard in the multi-shard case and shipped them across the gather — a large I/O and network waste (~13s → ~0.6s once fixed). Late-materialization (QTF) correctly declines these shapes (there is no fetch-only column to defer), so nothing was narrowing the scan. Single-shard was already fine; the cliff was multi-shard only.

### How
Run Calcite's field trimmer as the first planner step, on the logical tree, so each operator is slimmed to the columns its consumer needs and a narrowing projection reaches the scan; DataFusion then prunes the parquet read. 

Trimming is a pure optimization, so it falls back to the untrimmed tree on query shapes the trimmer can't handle. No changes to the late-materialization path. 

Plan-shape tests were re-baselined for the narrowed/inserted projection keeping a column where it carries real meaning (window, join, filter-project) and accepting the narrowing for count-by-key shapes.

Thanks to @rishabhmaurya for reporting the issue.

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
@@ -396,9 +396,7 @@ public class PlannerImpl {
      */
     static RelNode trimFields(RelNode input) {
         RelBuilder relBuilder = RelBuilder.proto(Contexts.empty()).create(input.getCluster(), null);
-        // Trimming is a pure optimization; the untrimmed tree is always a correct fallback. The
-        // trimmer's stock handlers reject some valid shapes via IllegalArgumentException (e.g.
-        // RelBuilder.sortLimit on a non-literal OFFSET) — fall back rather than fail the query.
+        // Trimming is a pure optimization; any trimmer failure must fall back, never fail the query.
         try {
             RelNode trimmed = new RelFieldTrimmer(null, relBuilder).trim(input);
             // trim() asserts an identity ref-mapping at the root, so field count/order are preserved
@@ -407,7 +405,7 @@ public class PlannerImpl {
             trimmed = relBuilder.push(trimmed).rename(input.getRowType().getFieldNames()).build();
             RelNodeUtils.logPlan(LOGGER, "After field trimming", trimmed);
             return trimmed;
-        } catch (IllegalArgumentException e) {
+        } catch (RuntimeException | AssertionError e) {
             LOGGER.warn("RelFieldTrimmer skipped (falling back to untrimmed tree): {}", e.toString());
             return input;
         }
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
@@ -400,7 +400,13 @@ public class PlannerImpl {
         // trimmer's stock handlers reject some valid shapes via IllegalArgumentException (e.g.
         // RelBuilder.sortLimit on a non-literal OFFSET) — fall back rather than fail the query.
         try {
-            return new RelFieldTrimmer(null, relBuilder).trim(input);
+            RelNode trimmed = new RelFieldTrimmer(null, relBuilder).trim(input);
+            // trim() asserts an identity ref-mapping at the root, so field count/order are preserved
+            // but names can drift (it drops alias-only top Projects, e.g. transpose's RENAME). Re-impose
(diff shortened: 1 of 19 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -124,6 +125,8 @@ public static RelNode runAllOptimizations(RelNode rawRelNode, PlannerContext con
 
         RelNode modifiedRelNode = rawRelNode;
         modifiedRelNode = removeSubQueries(modifiedRelNode, listener);
+        modifiedRelNode = trimFields(modifiedRelNode);
+        LOGGER.debug("After field trimming:\n{}", RelOptUtil.toString(modifiedRelNode));
```
bharath-techie: nit : please remove if its not intentional

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
```diff
@@ -372,6 +375,24 @@ private static RelNode decomposeAggregates(RelNode input, RuleProfilingListener
...
+     * projections, so the trimmer is only exercised in isolation on a deliberately-wide tree).
+     */
+    static RelNode trimFields(RelNode input) {
+        RelBuilder relBuilder = RelBuilder.proto(Contexts.empty()).create(input.getCluster(), null);
+        // Trimming is a pure optimization; the untrimmed tree is always a correct fallback (the
+        // trimmer's stock handlers throw on some valid shapes, e.g. a non-literal OFFSET).
+        try {
+            return new RelFieldTrimmer(null, relBuilder).trim(input);
+        } catch (RuntimeException e) {
+            LOGGER.debug("RelFieldTrimmer skipped (falling back to untrimmed tree): {}", e.toString());
```
alchemist51: Do we need warn?

bharath-techie (approved):
LGTM

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
bharath-techie: Here should be okay i think ? exception is rare

expani: WARN makes sense as this means it will skip an optimization and scan all columns from Parquet instead.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
expani: Yes it's intentional. 
Maybe I can move this inside `trimFields` to keep it cleaner

vinaykpud (approved):

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
expani: Done

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/PlannerImpl.java
expani: Moved it.

sandeshkr419:
@mch2 Verified the results are correct. Performance gains:


| Query | Pre P50 | Post P50 | Speedup | All Pre (ms) | All Post (ms) |
  |-------|---------|----------|---------|--------------|---------------|
  | Q25 | 6,729 ms | 293 ms | **23×** | [10116, 6668, 6659, 6729, 6788] | [425, 295, 293, 291, 289] |
  | Q27 | 21,596 ms | 794 ms | **27×** | [22422, 21596, 20975, 20944, 21846] | [809, 794, 794, 791, 788] |
  | q2 | 130 ms | 127 ms | — | — | — |
  | q20 | 581 ms | 580 ms | — | — | — |
  | q21 | 6,207 ms | 6,225 ms | — | — | — |


q2, q20, q21 were already fast with lucene delegation so no changes. change in planning only.

PR #22303, opened 2026-06-24
Fix dynamic native-memory settings that are accepted but not applied (rebalancer interval/thresholds, spill limit)

gaurav-amz:
#### Fix 1 — Changing the rebalancer interval silently turned the rebalancer off

**Setting:** `native.allocator.rebalance.interval_seconds` (`arrow-base`)

`updateRebalanceInterval()` first called `cancelRebalanceTask()`, which sets the `rebalancer` and `rebalancerScheduler` fields to `null`. It then tried to reschedule — but guarded that reschedule on those same fields being **non-null**. They were just nulled, so the reschedule branch was unreachable dead code.

The visible effect: you could `PUT` a new interval and get a `200`, but the rebalancer never ran again. It stayed off until someone toggled the `enabled` flag or restarted the node.

**Fix:** rebuild the rebalancer through `startRebalancer()`, using the allocator and budget supplier that were captured at startup. An interval of `0` still means "disabled", as before.

#### Fix 2 — The rebalancer ignored thresholds set in `opensearch.yml`

**Settings:** `native.allocator.rebalance.{pressure_threshold, idle_threshold, shrink_factor}` (`arrow-base`)

`startRebalancer()` built the `NativeMemoryRebalancer` using `getDefault(Settings.EMPTY)` for these three thresholds. `Settings.EMPTY` has nothing in it, so this always returned the hard-coded defaults — any value you put in `opensearch.yml` was thrown away at boot, and a disable/enable cycle reset live values back to defaults.

**Fix:** seed the thresholds from the live `ClusterSettings` (which includes both `opensearch.yml` values and any applied dynamic overrides) instead of from an empty settings object.

#### Fix 3 — A valid spill limit was rejected on a node where spill is actually enabled

**Setting:** `datafusion.spill_memory_limit_bytes` (`analytics-backend-datafusion`)

This setting's validator checks the requested limit against the spill volume's real capacity. To find the volume, it depends on `datafusion.spill_directory`. The catch: `spill_directory` is `NodeScope + Final`, so it lives only in `opensearch.yml` and **never appears in cluster state**.

When you do a dynamic `PUT _cluster/settings`, OpenSearch validates each setting scope in isolation, so the validator's dependency on `spill_directory` came back empty — even on a node where spill is fully configured. The validator then assumed "spill is disabled" and rejected any non-zero limit with a misleading *"spill_directory is unset"* error.

**Fix:** capture the node-scoped spill directory once at startup, and have the validator fall back to it when the dependency map is empty. Now the capacity check runs against the real configured volume, so valid limits are accepted and only genuinely-too-large values are rejected (with the correct "exceeds spill volume capacity" message).

---

### Related Issues

N/A — internal audit findings.

### Check List
- [x] All tests pass.


plugins/arrow-base/src/main/java/org/opensearch/arrow/allocator/ArrowBasePlugin.java
```diff
@@ -421,10 +461,21 @@ private synchronized void cancelRebalanceTask() {
...
         cancelRebalanceTask();
-        if (newInterval > 0 && rebalancerScheduler != null && rebalancer != null) {
-            rebalanceTask = rebalancerScheduler.scheduleAtFixedRate(rebalancer, newInterval, newInterval, TimeUnit.SECONDS);
+        if (newInterval <= 0) {
+            return;
+        }
+        ArrowNativeAllocator a = this.rebalancerAllocator;
+        Supplier<Long> budget = this.rebalancerBudgetSupplier;
+        if (a != null && budget != null) {
+            startRebalancer(a, budget, newInterval);
```
Bukhtawar, line 478: I am not clear how are we shutting down a previously scheduled rebalancer

PR #22306, opened 2026-06-24
Fix QTF fetch routing to wrong node after shard retry

alchemist51:
## Summary

- Fix Late Materialization fetch phase routing to the wrong node after a shard query retries on a replica
- After successful retry, update `resolvedTargetsByStage` so fetch goes to the node that actually executed the query

## Problem

When a shard's query phase fails on the primary (e.g., task cancelled by `SearchBackpressureService`), the `QueryScheduler` retries on a replica via `retargetForRetry()`. The replica succeeds and sends its batch to the reduce. However, `QueryContext.resolvedTargetsByStage` still points to the original primary node (populated at plan time in `ShardFragmentStageExecution.materializeTasks()`).

When the LM fetch stage dispatches via `scatterFetchAndStitch()`, it reads the stale target and sends the fetch to the primary — which freed the context on failure. Result: `No ReaderContext for queryId=... — query phase missing or context expired`.

## Root Cause

`recordResolvedTargets()` is called once at stage creation. `retargetForRetry()` creates a new `ShardExecutionTarget` with the replica's node but the same ordinal — this new target is never written back to the side-table.

## Fix

After a successful retry (`currentAttempt != task`), update the resolved target in `QueryContext` so the LM fetch phase routes to the correct node.

## Test plan

- [ ] Existing unit tests pass
- [ ] Verified with OSB load test (60 clients, q03 with sort+head) against a 40-shard index — confirmed the `No ReaderContext` error no longer occurs when shards retry on replicas
- [ ] Integration test to be added in follow-up

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/QueryContext.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/QueryContext.java
@@ -19,9 +19,9 @@ import org.opensearch.analytics.settings.AnalyticsQuerySettings;
 import org.opensearch.common.settings.Settings;
 import org.opensearch.threadpool.ThreadPool;
 
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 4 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/QueryScheduler.java
```diff
@@ -112,13 +114,26 @@ protected ActionListener<Void> handleFor(StageExecution stage, StageTask task) {
             @Override
             public void onResponse(Void unused) {
                 currentAttempt.transitionTo(StageTaskState.FINISHED);
+                // After a successful retry on a different shard copy, update the resolved target
+                // so downstream stages (LM fetch) route to the node that actually ran the query.
+                if (currentAttempt != task && stage instanceof ShardFragmentStageExecution shardStage) {
+                    if (currentAttempt instanceof ShardStageTask shardTask) {
+                        shardStage.updateResolvedTarget(shardTask);
```
mch2: can we do this inside the stage's retargetForRetry?

alchemist51: Removed it!

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/QueryContext.java
```diff
@@ -153,6 +153,18 @@ public void recordResolvedTargets(int stageId, List<ShardExecutionTarget> target
         resolvedTargetsByStage.put(stageId, byOrdinal);
     }
 
+    /**
+     * Updates a single resolved target after a successful shard retry on a different copy.
+     * This ensures downstream stages (e.g. LM fetch) route to the node that actually
+     * executed the query, not the original primary that failed.
+     */
+    public void updateResolvedTarget(int stageId, int ordinal, ShardExecutionTarget target) {
+        Map<Integer, ShardExecutionTarget> byOrdinal = resolvedTargetsByStage.get(stageId);
```
mch2, line 163: we need to update resolvedTargetsByStage to a concurrent collection, prior to this it was a single threaded dump on completion.

mch2 (approved):

PR #22309, opened 2026-06-24
Switch to slice() for low-priority large file uploads to avoid part corruption due to close

Bukhtawar:
## Summary

Two bugs in the S3 multipart upload path for large segment files (>15GB), all diagnosed from production heap profiles and async-profiler data on `otel-traces-000001`.

---

### Commit 1: Fix NPE when `provideStream()` throws

**Root cause:** `AsyncPartsHandler.uploadParts()` catch block swallowed exceptions from `provideStream()` and added no future, leaving `inputStreamContainers` slots null. `allOfExceptionForwarded()` completed on the shorter futures list, then `mergeAndVerifyChecksum()` NPE'd on the null slot — surfaced as `SdkClientException: Failed to send multipart upload requests`.

**Secondary effect:** `allOfExceptionForwarded()` completing early triggered `completionListener → indexInput.close()` while the `uploadParts()` loop was still blocked on `maybeAcquireSemaphore()` for later parts (~12.7s/part under contention). This raced with `indexInput.clone()` calls for those parts.

**Fix:** Add a pre-failed `CompletableFuture` in the catch block so `futures.size()` always equals `numberOfParts`.

---

### Commit 2: Fix `Arrays.fill(segments, null)` shared array corruption

**Root cause (confirmed by async-profiler):** `MemorySegmentIndexInput.clone()` passes the master's `segments[]` array **by reference** to `MultiSegmentImpl`. When any part's clone closes, `Arrays.fill(segments, null)` fills the shared array with nulls. All subsequent `provideStream()` calls hit `AlreadyClosedException` in the `OffsetRangeIndexInputStream` constructor at `indexInput.seek(position)` — because `seek()` accesses `segments[index]` which is now null.

This was confirmed by:
- Async-profiler: 79.84% of `MemorySegmentIndexInput.close()` calls came from `OffsetRangeRefCount.decRef` (normal part upload complete path)
- The `AlreadyClosedException` stack shows `FilterIndexInput.seek ← OffsetRangeIndexInputStream.<init>` — failure in the constructor, not during reads

**Fix:** Replace `indexInput.clone()` with `indexInput.slice("part@" + position, position, size)`. For non-full-range slices, `buildSlice()` calls `ArrayUtil.copyOfSubArray()` — each part gets its own independent `MemorySegment[]` copy. Closing one part only fills its own private array. ~79 KB extra heap for 1129 parts; no extra mmap, no data copy.

Also wires an `AtomicInteger` ref count on the master `IndexInput` so `indexInput.close()` is deferred until all part slices have closed AND the completion listener has fired.

---

## Check List
- [x] Functionality includes unit tests (failing without the fix, passing with it).

### DIFF
--- a/server/src/main/java/org/opensearch/index/store/remote/DataFormatAwareRemoteDirectory.java
+++ b/server/src/main/java/org/opensearch/index/store/remote/DataFormatAwareRemoteDirectory.java
@@ -48,7 +48,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.UnaryOperator;
 
 /**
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/opensearch/index/store/remote/DataFormatAwareRemoteDirectory.java
```diff
@@ -379,19 +380,75 @@ private void uploadBlob(
...
+                if (indexInputClosed.get()) {
+                    logger.warn(
+                        () -> new ParameterizedMessage(
+                            "IndexInput.slice() attempted on already-closed IndexInput for [{}] on thread [{}];"
+                                + " the master was closed before all parts completed",
+                            src,
+                            Thread.currentThread().getName()
+                        )
+                    );
+                }
```
mgodwan, line 418: should we throw here? clone should not be called on closed, right?

mgodwan (approved):

PR #22314, opened 2026-06-25
Fix multi index resolution on analytics engine

finnegancarroll:
### Description

`RelNodeUtils.extractIndices()` returned comma-delimited table names as a single array element (e.g. `["idx1,idx2"]`). The security filter's `IndexNameExpressionResolver` tried to resolve this as one index name, was unable to, and resolved no concrete indices.

Due to the default setting of [strict expand open](https://github.com/opensearch-project/OpenSearch/blob/main/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/exec/action/AnalyticsQueryRequest.java#L97) on analytics query request [index resolution errors are not fatal](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/metadata/IndexNameExpressionResolver.java#L329).

### Fix

Split comma-delimited table names in `collectIndices()` using `Strings.splitStringByCommaToArray()` (same utility used by `IndexResolution` in the planner). Each index is now a separate entry in the `String[]` array, so security evaluates permissions on each independently.

### DIFF
--- a/sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/RelNodeUtilsTests.java
+++ b/sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/RelNodeUtilsTests.java
@@ -151,6 +151,74 @@ public class RelNodeUtilsTests extends OpenSearchTestCase {
         assertTrue(e.getMessage().contains("maximum depth"));
     }
 
+    // --- Multi-index comma-separated table name tests (FGAC bypass fix) ---
+
+    public void testCommaDelimitedIndicesSplit() {
+        RelBuilder b = builderWithTable("logs-2024-01,secrets-2024-01");
+        RelNode plan = b.scan("logs-2024-01,secrets-2024-01").build();
+        assertArrayEquals(new String[] { "logs-2024-01", "secrets-2024-01" }, RelNodeUtils.extractIndices(plan));
+    }
+
+    public void testCommaDelimitedThreeIndices() {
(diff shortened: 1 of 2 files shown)


mch2 (approved):

sandbox/plugins/analytics-engine/src/test/java/org/opensearch/analytics/planner/RelNodeUtilsTests.java
```diff
@@ -151,6 +151,74 @@ public void testDepthGuardThrowsOnExcessiveDepth() {
...
+    }
+
+    public void testCommaDelimitedThreeIndices() {
+        RelBuilder b = builderWithTable("a,b,c");
+        RelNode plan = b.scan("a,b,c").build();
+        assertArrayEquals(new String[] { "a", "b", "c" }, RelNodeUtils.extractIndices(plan));
+    }
+
+    public void testDoubleCommaProducesEmptyStringFiltered() {
+        RelBuilder b = builderWithTable("index1,,index2");
```
expani, line 169: So, Vanilla allows incorrectly parsed names ?

finnegancarroll: some of these might be rejected up front by sql/ppl before making it into AE. I'm unsure. Adding SQL plugin integ tests for this PR here https://github.com/opensearch-project/sql/pull/5581 and will cover these cases.

PR #22318, opened 2026-06-25
Handle missing source for no-op tombstones

### LINKED ISSUE #22317: Lucene changes snapshot can fail when no-op tombstone source is missing
## Description

`LuceneChangesSnapshot` can throw a `NullPointerException` when it reads a no-op tombstone whose stored `_source` is unavailable.

The no-op tombstone path currently builds a `Translog.NoOp` by dereferencing `fields.source()` directly. The regular index operation path already handles a missing source by either throwing `MissingHistoryOperationsException` when a full range is required, or skipping the operation when a full range is not required. The no-op tombstone path should use the same missing-source handling.

## Impact

History snapshots can fail with an NPE instead of reporting or skipping missing history according to the `requiredFullRange` setting.

## Expected behavior

- If `requiredFullRange` is `true`, a missing no-op tombstone source should throw `MissingHistoryOperationsException`.
- If `requiredFullRange` is `false`, the snapshot should skip the operation and increment the skipped operation count.

## Actual behavior

The no-op tombstone path dereferences `fields.source()` without checking for `null`, so a missing source can abort the snapshot with `NullPointerException`.

## Reproduction

A unit test can reproduce this by creating a no-op tombstone, reading a `LuceneChangesSnapshot` through a reader that hides the stored `_source` field, and calling `next()`.

aneesh-db:
### Description

Fixes #22317.

`LuceneChangesSnapshot` already handles missing source for index operations, but no-op tombstones were still dereferencing `fields.source()` directly. When a stored `_source` field is missing for a no-op tombstone, this change reconstructs the no-op with a fallback reason instead of failing with a `NullPointerException`.

Index operations continue to use the existing missing-source handling.

The regression tests cover missing no-op tombstone `_source` for both recovery modes by reading a no-op tombstone through a reader that hides the stored `_source` field. They also verify that the normal path preserves the stored no-op reason.

### Testing

- `server:test --tests "org.opensearch.index.engine.LuceneChangesSnapshotTests"`
- `server:spotlessJavaCheck`


andrross:
> When requiredFullRange is true and the no-op's source is missing, silently substituting a fallback reason may hide a real data-loss issue in peer recovery. 

^ from the review bot. Looks like a valid question. Should the code unconditionally substitute a fallback reason?

aneesh-db:
@andrross Yes, for the no-op tombstone branch I think the code should substitute a fallback reason whenever the stored `_source` reason is missing, regardless of `requiredFullRange`.

The reason is that a no-op tombstone does not contain document source needed to replay an operation. Its `_source` stores only the no-op reason string, so losing that field means losing diagnostic text, not losing the operation itself. The sequence number, primary term, and operation type are still available, so the history range can still be reconstructed.

This is intentionally different from index operations: missing index `_source` still goes through `handleMissingSource(seqNo)` and can throw when `requiredFullRange` is true, because an index operation cannot be replayed without its source.

I pushed 2e5d04f to make that distinction explicit in the code and added coverage for the normal path that preserves the stored no-op reason.

PR #22327, opened 2026-06-26
[analytics-engine] Fix flaky rust test: concurrent_shard_warmup_does_not_corrupt

LantaoJin:
### Description
`FoyerCache` is configured disk-only (`.memory(1)` — in-memory tier is 1 byte). `put_metadata` → `insert()` returns immediately, but the actual write goes through Foyer's async storage flusher. This test fires 5 concurrent `put_metadata` tasks (the most write-concurrent test in the file), then issues `get()` immediately — so under heavy load an entry can still be queued in the write buffer when read, intermittently tripping `assert!(cached.is_some())`.

Fix: Added `cache.wait_for_flush().await` after the warmup tasks join and before the verification reads. That method exists for exactly this purpose — its documented post-condition is "all previously put() entries are on SSD and findable via get()." The production cache path is unaffected; this only makes the test's read-after-write deterministic.

### Related Issues

### Check List


LantaoJin:
@mch2 @sandeshkr419

gaobinlong (approved):

PR #22328, opened 2026-06-26
Validate base_path in FsRepository to prevent path.repo containment bypass

kh3ra:
### Description
The fs repository base_path setting was read verbatim from REST input with no validation. An absolute base_path causes Path.resolve to discard the path.repo-validated location, redirecting all blob-store operations outside the repository (CWE-22) and enabling arbitrary filesystem deletion via the snapshot _cleanup API.

Add two layers of defense:
 - BASE_PATH_SETTING validator rejects absolute and upward-escaping (..) values after normalization (benign interior '..' that cancels out is allowed).
 - validateBasePathWithinRepo() resolves base_path against the location and verifies the result stays within a configured path.repo directory.

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/repositories/fs/FsRepository.java
+++ b/server/src/main/java/org/opensearch/repositories/fs/FsRepository.java
@@ -102,7 +102,19 @@ public class FsRepository extends BlobStoreRepository {
         Property.Deprecated
     );
 
-    public static final Setting<String> BASE_PATH_SETTING = Setting.simpleString("base_path");
+    public static final Setting<String> BASE_PATH_SETTING = Setting.simpleString("base_path", value -> {
+        // CWE-22 hardening (string-only fail-fast; avoids constructing a Path from a raw user-supplied string):
+        // reject absolute base_path values (POSIX "/", Windows "\\"/UNC, drive-letter "C:..") and any parent-directory
+        // ("..") segment. An absolute base_path makes Path.resolve discard the path.repo-validated location, escaping
+        // containment; a ".." segment lets the resolved path climb above it. The authoritative, location-aware
+        // containment check is performed in validateBasePathWithinRepo().
+        if (Strings.hasLength(value)
+            && (value.startsWith("/") || value.startsWith("\\") || value.matches("^[A-Za-z]:.*") || value.contains(".."))) {
+            throw new IllegalArgumentException(
+                "[base_path] must be a relative path that does not contain '..' segments; got [" + value + "]"
+            );
+        }
+    });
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/opensearch/repositories/fs/FsRepository.java
```diff
@@ -102,7 +102,19 @@ public class FsRepository extends BlobStoreRepository {
...
     );
 
-    public static final Setting<String> BASE_PATH_SETTING = Setting.simpleString("base_path");
+    public static final Setting<String> BASE_PATH_SETTING = Setting.simpleString("base_path", value -> {
+        // CWE-22 hardening (string-only fail-fast; avoids constructing a Path from a raw user-supplied string):
+        // reject absolute base_path values (POSIX "/", Windows "\\"/UNC, drive-letter "C:..") and any parent-directory
+        // ("..") segment. An absolute base_path makes Path.resolve discard the path.repo-validated location, escaping
+        // containment; a ".." segment lets the resolved path climb above it. The authoritative, location-aware
+        // containment check is performed in validateBasePathWithinRepo().
+        if (Strings.hasLength(value)
```
cwperks, line 111: @kh3ra is this supposed to be under `config/` as well? I believe we have an existing convention that you can find in https://github.com/opensearch-project/OpenSearch/pull/22094.

kh3ra: `path.config` and `path.repo` are separate. I don't believe moving this inside config would be right.

kh3ra:
The build is red, seems to be a known flaky test - https://github.com/opensearch-project/OpenSearch/issues/21378

server/src/main/java/org/opensearch/repositories/fs/FsRepository.java
cwperks: Should we create a similar `isUnderRepo` then?

cwperks (dismissed):

server/src/main/java/org/opensearch/repositories/fs/FsRepository.java
```diff
@@ -102,7 +102,19 @@ public class FsRepository extends BlobStoreRepository {
...
+            && (value.startsWith("/") || value.startsWith("\\") || value.matches("^[A-Za-z]:.*") || value.contains(".."))) {
```
cwperks, line 112: This is good, but I think we have similar logic in the HunspellService as well. I think we should consider creating a special setting for these relative paths and move the validation there.

cwperks (approved):

PR #22334, opened 2026-06-27
Fix default replica count from node settings

### LINKED ISSUE #20389: [BUG] `cluster.default_number_of_replicas` setting is ignored when set in `opensearch.yml`
### Describe the bug

According to the [docs](https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/index-settings/), there is a cluster wide setting that allows changing the default number of replicas for new indices (I want 0 by default). However, it does not work when I add the setting to the `opensearch.yml` file. However, the setting does work if I perform a request to the API.

From the [docs](https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/index-settings/):

> `cluster.default_number_of_replicas` (Integer): Controls the default number of replicas for indexes in the cluster. The index-level index.number_of_replicas setting defaults to this value if not configured. Default is 1

Is this the expected behavior? Tested in a single node cluster.

### Related component

Other

### To Reproduce

**Setting in `opensearch.yml`**

```yml
network.host: "0.0.0.0"
node.name: "node-1"
```

-> Restart

Image

**Setting in API**

Image

### Expected behavior

The setting `cluster.default_number_of_replicas` is taken into account regardless of the application method.

### Additional Details

**Additional context**
Tested in a single node cluster, v3.3.2.

krugerLucas:
### Description

`cluster.default_number_of_replicas` is a dynamic node-scoped setting and can be supplied through `opensearch.yml`. The create-index path, however, resolved the default replica count from `currentState.metadata().settings()`, which only contains cluster-state settings (`persistent` and `transient`). As a result, indices created without an explicit `index.number_of_replicas` ignored the value loaded from node settings.

This PR resolves the default replica count through `ClusterSettings` when creating and validating index settings. That keeps the expected effective-setting precedence:

```text
transient > persistent > node settings > hardcoded default
```

### What changed

- Use `ClusterSettings` when applying the default `index.number_of_replicas` during index creation.
- Use the same effective default when validating index settings.
- Add unit coverage for hardcoded default, node setting, persistent setting, transient setting, explicit request setting, template setting, and validation behavior.
- Add an internal cluster test that configures `cluster.default_number_of_replicas` through node settings and verifies a newly created index uses it.

### Related Issues

Resolves #20389

### Test plan

- [x] Metadata create-index unit tests:

```bash
JAVA_HOME=/Users/lucaskruger/Library/Java/JavaVirtualMachines/openjdk-24.0.1/Contents/Home \
  ./gradlew :server:test --tests org.opensearch.cluster.metadata.MetadataCreateIndexServiceTests
```

- [x] Node-settings internal cluster coverage:

```bash
JAVA_HOME=/Users/lucaskruger/Library/Java/JavaVirtualMachines/openjdk-24.0.1/Contents/Home \
  ./gradlew :server:internalClusterTest --tests org.opensearch.action.admin.indices.create.CreateIndexDefaultReplicaCountWithNodeSettingsIT
```

- [x] Server precommit:

```bash
JAVA_HOME=/Users/lucaskruger/Library/Java/JavaVirtualMachines/openjdk-24.0.1/Contents/Home \
  ./gradlew :server:precommit
```

### Check List
- [x] API changes companion pull request: not applicable.
- [x] Public documentation issue/PR: not applicable; this fixes existing documented behavior.

### DIFF
--- a/server/src/internalClusterTest/java/org/opensearch/action/admin/indices/create/CreateIndexDefaultReplicaCountWithNodeSettingsIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/action/admin/indices/create/CreateIndexDefaultReplicaCountWithNodeSettingsIT.java
@@ -1,0 +1,52 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.action.admin.indices.create;
+
+import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.cluster.metadata.Metadata;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.test.OpenSearchIntegTestCase;
+import org.opensearch.test.OpenSearchIntegTestCase.ClusterScope;
+import org.opensearch.test.OpenSearchIntegTestCase.Scope;
+
+import static org.opensearch.test.hamcrest.OpenSearchAssertions.assertAcked;
+import static org.hamcrest.Matchers.equalTo;
+
+@ClusterScope(scope = Scope.TEST)
+public class CreateIndexDefaultReplicaCountWithNodeSettingsIT extends OpenSearchIntegTestCase {
+
+    private static final String INDEX_NAME = "test-index";
+    private static final int DEFAULT_NUMBER_OF_REPLICAS = 0;
+
+    @Override
+    protected Settings nodeSettings(int nodeOrdinal) {
+        return Settings.builder()
+            .put(super.nodeSettings(nodeOrdinal))
+            .put(Metadata.DEFAULT_REPLICA_COUNT_SETTING.getKey(), DEFAULT_NUMBER_OF_REPLICAS)
+            .build();
+    }
+
+    @Override
+    protected void randomIndexTemplate() {
+        // The base template sets index.number_of_replicas and would override the cluster-level default verified here.
+    }
+
+    public void testCreateIndexUsesDefaultNumberOfReplicasFromNodeSettings() {
+        assertAcked(
+            client().admin()
+                .indices()
+                .prepareCreate(INDEX_NAME)
+                .setSettings(Settings.builder().put(IndexMetadata.INDEX_NUMBER_OF_SHARDS_SETTING.getKey(), 1))
+                .get()
+        );
+
+        Settings indexSettings = client().admin().indices().prepareGetSettings(INDEX_NAME).get().getIndexToSettings().get(INDEX_NAME);
+        assertThat(indexSettings.getAsInt(IndexMetadata.SETTING_NUMBER_OF_REPLICAS, null), equalTo(DEFAULT_NUMBER_OF_REPLICAS));
+    }
+}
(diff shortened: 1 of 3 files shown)


server/src/internalClusterTest/java/org/opensearch/action/admin/indices/create/CreateIndexDefaultReplicaCountWithNodeSettingsIT.java
```diff
@@ -0,0 +1,47 @@
...
+import org.opensearch.common.settings.Settings;
+import org.opensearch.test.OpenSearchIntegTestCase;
+import org.opensearch.test.OpenSearchIntegTestCase.ClusterScope;
+import org.opensearch.test.OpenSearchIntegTestCase.Scope;
+
+import static org.opensearch.test.hamcrest.OpenSearchAssertions.assertAcked;
+import static org.hamcrest.Matchers.equalTo;
+
+@ClusterScope(scope = Scope.TEST)
+public class CreateIndexDefaultReplicaCountWithNodeSettingsIT extends OpenSearchIntegTestCase {
```
andrross, line 22: The base test class installs a random index template that uses a random number of replicas: https://github.com/opensearch-project/OpenSearch/blob/916ef741090c3408f8b2f22f03bddf6aa0b51f0c/test/framework/src/main/java/org/opensearch/test/OpenSearchIntegTestCase.java#L449

Will that possibly override the default and make this test flaky?

krugerLucas: Good catch. Yes, the base `random_index_template` uses pattern `*` and can set `index.number_of_replicas`, which would override or mask the cluster-level default this IT is verifying.

I disabled the inherited random index template for this test so the created index exercises `cluster.default_number_of_replicas` from node settings directly.

server/src/test/java/org/opensearch/cluster/metadata/MetadataCreateIndexServiceTests.java
```diff
@@ -1003,6 +1004,118 @@ public void testSettingsFromClusterState() {
...
+        assertThat(aggregatedIndexSettings.get(SETTING_NUMBER_OF_REPLICAS), equalTo("0"));
+    }
+
+    public void testDefaultNumberOfReplicasUsesPersistentSettingOverNodeSetting() {
+        Settings nodeSettings = Settings.builder().put(Metadata.DEFAULT_REPLICA_COUNT_SETTING.getKey(), 0).build();
+        ClusterSettings clusterSettings = new ClusterSettings(nodeSettings, ClusterSettings.BUILT_IN_CLUSTER_SETTINGS);
+        Metadata metadata = Metadata.builder()
+            .persistentSettings(Settings.builder().put(Metadata.DEFAULT_REPLICA_COUNT_SETTING.getKey(), 2).build())
+            .build();
+        ClusterState clusterState = ClusterState.builder(org.opensearch.cluster.ClusterName.CLUSTER_NAME_SETTING.getDefault(Settings.EMPTY))
```
andrross: Please use an import instead of a fully qualified class name

andrross (commented):
Looks good, just one annoying nitpick in the test :)

server/src/test/java/org/opensearch/cluster/metadata/MetadataCreateIndexServiceTests.java
krugerLucas: Done. I replaced the fully qualified `org.opensearch.cluster.ClusterName` references introduced by this PR with the existing `ClusterName` import.

andrross (approved):

krugerLucas:
thanks @andrross for the review, CI is green and the PR is approved. can you please merge it in when you have a chance?

Mmuzaf:
@krugerLucas 
We encountered this issue in prod. It appears that this fix is only available from version 3.8 onwards. Do we also need to cherry-pick it for the latest LTS version (I suppose it's 3.6)? I can prepare a patch for that if need.

https://opensearch.org/long-term-support/

PR #22347, opened 2026-06-29
Add cross-setting validator to prevent balance factor zero crash loop

### LINKED ISSUE #22305: Setting balance factors to zero causes unrecoverable cluster-manager crash loop
### Describe the bug

Setting `cluster.routing.allocation.balance.shard` and `cluster.routing.allocation.balance.index` both to `0.0` causes an unrecoverable cluster-manager crash loop. The cluster becomes completely inoperable and cannot be recovered via the API.

**Affected versions:** All versions of OpenSearch (1.0 through current). The code was inherited from Elasticsearch and has never had a cross-setting validator for the balance factor sum constraint.

### Root Cause

`BalancedShardsAllocator.WeightFunction` constructor throws `IllegalArgumentException` when balance factors sum to 0:

```java
float sum = indexBalance + shardBalance;
if (sum <= 0.0f) {
    throw new IllegalArgumentException("Balance factors must sum to a value > 0 but was: " + sum);
}
```

The individual settings allow `0.0` as minimum (`Setting.floatSetting(..., 0.0f, ...)`), so each passes validation independently. But the cross-setting constraint (sum > 0) is only checked inside the `WeightFunction` constructor, which runs during `ClusterApplierService.applyChanges()` -> `clusterSettings.applySettings()` -> the registered settings consumer `updateIndexBalanceFactor()` -> `updateWeightFunction()` -> throws.

No cross-setting validator exists at the settings API layer to prevent this combination from entering the cluster state.

### How the crash loop works

1. User sets both balance factors to 0 via `PUT /_cluster/settings`
2. `TransportClusterUpdateSettingsAction.execute()` builds new ClusterState (validates each setting individually - passes)
3. `MasterService.runTasks()` -> `calculateTaskOutputs()` -> `publish()` -> state sent to all nodes
4. Quorum acks -> commit sent to all nodes
5. On every node (including master), `ClusterApplierService.applyChanges()` runs:
   - `clusterSettings.applySettings(incomingSettings)` iterates all registered `SettingUpdater`s
   - The `INDEX_BALANCE_FACTOR_SETTING` updater calls `BalancedShardsAllocator.updateIndexBalanceFactor(0.0)`
   - -> `updateWeightFunction()` -> `new WeightFunction(0.0, 0.0)` -> throws `IllegalArgumentException`
6. On the master, exception propagates: `applyChanges()` -> `runTask()` -> `task.listener.onFailure()` -> `Coordinator.removePublicationAndPossiblyBecomeCandidate()` -> master steps down
7. New election -> same state (with bad transient setting in memory) -> same crash -> infinite loop

### Impact

- **Cluster-manager**: Crashes every ~150ms. Observed terms incrementing by thousands per minute.
- **Data nodes**: Also fail to apply every incoming cluster state (same exception). `BalancedShardsAllocator` is instantiated on ALL nodes and registered as a settings consumer, even though only the cluster-manager uses it for allocation decisions.
- **API access impossible**: The cluster-manager steps down before it can process any queued tasks. The only tasks that run are the `elected-as-cluster-manager` join tasks, which immediately trigger the crash.
- **`GET /_cluster/settings` still works**: Because it reads from `preCommitState` (set before apply runs), not the applied `state` (which is never updated since `state.set()` is never reached).
- **`lastSettingsApplied` never updates**: Since `applySettings()` throws before `lastSettingsApplied = newSettings`, the same failed apply retries on every subsequent cluster state.

### Recovery

- **Transient settings**: Restart ALL cluster-manager-eligible nodes simultaneously. Transient settings are not written to disk. On restart, nodes load from Lucene `_state/` -> no transient settings -> clean boot.
- **Persistent settings**: Requires `opensearch-node unsafe-bootstrap` or wiping `_state/` directory.
- **API fix impossible**: Cannot issue `PUT /_cluster/settings` because the master never stabilizes long enough to process queued tasks.

### Reproduction

```bash
curl -X PUT "localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d '{
  "transient": {
    "cluster.routing.allocation.balance.shard": "0.0",
    "cluster.routing.allocation.balance.index": "0.0"
  }
}'
```

Master logs show:

```
[INFO ][o.o.c.s.ClusterSettings] updating [cluster.routing.allocation.balance.index] from [0.55] to [0.0]
[WARN ][o.o.c.s.ClusterSettings] failed to apply settings
java.lang.IllegalArgumentException: Balance factors must sum to a value > 0 but was: 0.0
    at o.o.c.r.a.a.BalancedShardsAllocator$WeightFunction.<init>
    at o.o.c.r.a.a.BalancedShardsAllocator.updateWeightFunction
    at o.o.c.r.a.a.BalancedShardsAllocator.updateIndexBalanceFactor
    at o.o.c.s.Setting$Updater.apply
    at o.o.c.s.AbstractScopedSettings.applySettings
    at o.o.c.s.ClusterApplierService.applyChanges
[INFO ][o.o.c.s.ClusterApplierService] cluster-manager node changed {previous [...], current []}
    reason: becoming candidate: clusterApplier#onNewClusterState
```

### Proposed Fix

**1. Cross-setting validator (prevents poison pill from entering cluster state):**

Add a `Setting.Validator` (same pattern as `DiskThresholdSettings.LowDiskWatermarkValidator`) that validates the sum of both balance factors > 0 before the setting is committed. This runs during `SettingsUpdater.updateSettings()` -> `clusterSettings.validate()` - before the state is published.

**2. Defensive handling in `WeightFunction` (prevents crash if bad values somehow get through):**

```java
if (sum <= 0.0f) {
    indexBalance = 0.45f;
    shardBalance = 0.55f;
    sum = 1.0f;
}
```

Dhanwani:
## Summary

Setting both `cluster.routing.allocation.balance.shard` and `cluster.routing.allocation.balance.index` to `0.0` causes an unrecoverable cluster-manager crash loop. Each setting independently allows `0.0` as minimum, but the `WeightFunction` requires their sum to be > 0. When both are set to zero, the cluster-manager crashes on every election attempt and becomes unrecoverable via the API.

### Changes

- Add `IndexBalanceFactorValidator` and `ShardBalanceFactorValidator` (same pattern as `DiskThresholdSettings.LowDiskWatermarkValidator`) that validate the sum of both balance factors > 0 before settings enter cluster state
- Add unit tests for both valid and invalid setting combinations

### Impact

- **Before**: Setting both balance factors to 0 causes permanent cluster-manager crash loop requiring manual node restart to recover
- **After**: The API rejects the invalid combination with a clear error message (HTTP 400)

## Test Plan

- [x] Unit tests verify invalid combinations are rejected on construction and dynamic update
- [x] Unit tests verify valid combinations (one factor at 0 with the other > 0) are accepted
- [x] Manually tested on a 6-node cluster (3 master + 3 data nodes): confirmed the API returns HTTP 400 with clear error message when attempting to set both factors to 0
- [x] Confirmed existing clusters with bad values recover gracefully (defensive fallback resets to defaults)

### Manual test output (post-fix):

```bash
curl -X PUT "http://localhost:9200/_cluster/settings" -H 'Content-Type: application/json' -d '{
    "transient": {
      "cluster.routing.allocation.balance.shard": "0",
      "cluster.routing.allocation.balance.index": "0"
    }
  }'
```

```json
{
  "error": {
    "root_cause": [
      {
        "reason": "Balance factors [cluster.routing.allocation.balance.index] and [cluster.routing.allocation.balance.shard] must sum to a value greater than zero but was [0.0]"
      }
    ],
    "reason": "Balance factors [cluster.routing.allocation.balance.index] and [cluster.routing.allocation.balance.shard] must sum to a value greater than zero but was [0.0]"
  },
  "status": 400
}
```

## Issues Resolved

Fixes https://github.com/opensearch-project/OpenSearch/issues/22305


server/src/main/java/org/opensearch/cluster/routing/allocation/allocator/BalancedShardsAllocator.java
```diff
@@ -563,8 +630,17 @@ static class WeightFunction {
             boolean preferPrimaryShardRebalance
         ) {
             float sum = indexBalance + shardBalance;
+            // Defensive fallback: if invalid values reach here (e.g., from a cluster state written by an older version
+            // without the cross-setting validator), reset to defaults to prevent a division-by-zero crash loop that
+            // would make the cluster unrecoverable.
```
gbbafna: Lets not have this fallbacks , as it makes system unpredictable. Its okay to do full cluster restart to fix this instead .

Dhanwani: restart can fix transient setting but if the persistent setting is changed, there the defensive check will help, right? Do we have any other way of recovery in case persistent setting is impacted?

server/src/main/java/org/opensearch/cluster/routing/allocation/allocator/BalancedShardsAllocator.java
```diff
@@ -518,6 +524,67 @@ public boolean getPreferPrimaryBalance() {
         return preferPrimaryShardBalance;
     }
 
+    /**
+     * Validates that the index balance factor, combined with the shard balance factor, sums to a value greater than zero.
+     *
+     * @opensearch.internal
+     */
+    static final class IndexBalanceFactorValidator implements Setting.Validator<Float> {
```
alchemist51, line 532: We don't need individual classes for them, In the constructor we can pass the setting name to be fetched

alchemist51 (changes requested):

alchemist51:
Spotless is failing for the new files

server/src/main/java/org/opensearch/cluster/routing/allocation/allocator/BalancedShardsAllocator.java
Dhanwani: `SHARD_BALANCE_FACTOR_SETTING` is not initialized, so we can not pass it during class loading.

Dhanwani:
Reopening - branch was fixed after an accidental force push.

server/src/main/java/org/opensearch/cluster/routing/allocation/allocator/BalancedShardsAllocator.java
alchemist51: We can pass a supplier ?

Dhanwani: `SHARD_BALANCE_FACTOR_SETTING` is not initialized at the point where `INDEX_BALANCE_FACTOR_SETTING` is declared — it gets initialized after. 

  ```
static final Setting<Float> INDEX_BALANCE_FACTOR_SETTING = Setting.floatSetting(..., new Validator(() -> SHARD_BALANCE_FACTOR_SETTING), ...);  // ERROR: illegal forward reference
  static final Setting<Float> SHARD_BALANCE_FACTOR_SETTING = Setting.floatSetting(..., new Validator(() -> INDEX_BALANCE_FACTOR_SETTING), ...);  // Fine - INDEX exists

```
  We could use a boolean to switch between index vs shard setting in a single class, but two separate classes are more readable.

PR #22351, opened 2026-06-29
Modernize logger usage check

andrross:
Accept a trailing throwable with parameters as opposed to requiring a clunkly lambda to use this syntax.

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/transport/TransportService.java
+++ b/server/src/main/java/org/opensearch/transport/TransportService.java
@@ -34,7 +34,6 @@ package org.opensearch.transport;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
-import org.apache.logging.log4j.message.ParameterizedMessage;
 import org.opensearch.OpenSearchServerException;
 import org.opensearch.Version;
 import org.opensearch.action.ActionListenerResponseHandler;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/opensearch/transport/TransportService.java
```diff
@@ -432,25 +431,13 @@ protected void doStop() {
                     public void onRejection(Exception e) {
                         holderToNotify.handler().handleRejection(e);
                         // if we get rejected during node shutdown we don't wanna bubble it up
-                        logger.debug(
```
andrross, line 435: This class was arbitrarily chosen just to demonstrate the syntax the previous logger usage checked forced us to use and the simpler alternative.

test/logger-usage/src/main/java/org/opensearch/test/loggerusage/OpenSearchLoggerUsageChecker.java
```diff
@@ -627,4 +708,66 @@ public BasicValue ternaryOperation(AbstractInsnNode insnNode, BasicValue value1,
...
+        }
+
+        private static boolean isReference(Type type) {
+            return type != null && (type.getSort() == Type.OBJECT || type.getSort() == Type.ARRAY);
+        }
+
+        private boolean isThrowable(Type type) {
+            return THROWABLE_CACHE.computeIfAbsent(type.getClassName(), className -> {
+                try {
+                    Class<?> clazz = Class.forName(className, false, classLoader);
```
cwperks, line 762: I'm trying to understand why we need to do all the convoluted logic with classloader here, is this because of the plugin ecosystem?

andrross: It's because the final argument may be an exception type defined in the OpenSearch code base, so we need to be able to load those classes in order to do the check.

andrross: Previously the check could operate with just the JDK and log4j classes on the classpath

cwperks (approved):

PR #22355, opened 2026-06-30
Push NestedQueryBuilder filters to inner query

venkateshwaracholan:
## Summary

Override `NestedQueryBuilder.filter()` to apply filters to the inner query instead of wrapping the nested query in a parent-level `bool` query.

This fixes incorrect filter semantics when callers invoke `QueryBuilder.filter()` on nested queries.

## Problem

`NestedQueryBuilder` inherits the default `AbstractQueryBuilder.filter()` implementation, which wraps the query as:

```json
{
  "bool": {
    "must": [ <nested query> ],
    "filter": [ <filter> ]
  }
}
```

This behavior is correct for most query types but not for nested queries.

Nested queries do not support a top-level `filter` clause in their DSL. The filter must be applied to the inner query so it executes in the nested document context rather than at the parent document level after the nested join.

## Solution

Override `NestedQueryBuilder.filter()` to delegate filter application to the inner query:

```java
QueryBuilder filteredInnerQuery = query.filter(filter);
```

This approach:

- Applies filters in the nested document context.
- Preserves query metadata (`path`, `scoreMode`, `innerHitBuilder`, `ignoreUnmapped`, `boost`, and `queryName`).
- Works recursively for nested-in-nested queries.
- Automatically supports inner query types that override `filter()`.

## Tests

Added unit tests covering:

- Filter propagation to the inner query
- Randomized nested query builders
- Null filter handling
- Immutability of the original query
- Metadata preservation
- Nested-in-nested query behavior

### Validation

```bash
./gradlew :server:test --tests "org.opensearch.index.query.NestedQueryBuilderTests"

./gradlew :server:test --tests "org.opensearch.index.query.*Tests"
```

## Related
- Addresses the root cause discussed in https://github.com/opensearch-project/neural-search/issues/1759
- Implements the design proposed by @bzhangam in https://github.com/opensearch-project/neural-search/pull/1874

### Check List


bzhangam:
LGTM. Thank you for taking care of this change.

Issue #17885, opened 2025-04-10
[Docs] Documentation gap: Max response size


Swiddis:
### Is your feature request related to a problem? Please describe

In side-channels, multiple people have asked on different occasions what the maximum size for a single OpenSearch HTTP response is. There's the `http.max_content_length` setting which applies to the request size. No similar option for the response size seems to exist.

Here's my attempt at an answer based on combining a few sources and envelope math:

The response body doesn't have a hard limit[^1] since it's in principle enforced by Jetty (which supports arbitrarily large responses). But it can be bound by several things:
- The cluster needs to have enough heap memory to compute and store the response, otherwise you'll get a `CircuitBreakerException`.
- If the cluster is servicing multiple requests at once, they'll interfere with each other's memory usage, leading to earlier circuit breakers.
- It also likely depends on the exact query: something like computing a large histogram will take more memory than simply listing a bunch of documents.
- Other background operations on the cluster (e.g. shard management, especially if there's many shards) also will take some of that heap.

_Very roughly,_ you can expect a max response size of about 10% of the cluster's available heap, up to the exact usage profile. The most surefire way to know is to test the cluster with realistic request load for your use case.

### Describe the solution you'd like

Determine a conclusive answer (even this one if it's sufficiently accurate) and document it somewhere.

This question should be avoidable with [pagination](https://opster.com/guides/opensearch/opensearch-how-tos/opensearch-pagination-which-techniques/), but the topic comes up often enough regardless, so there may be use case gaps. We can link this or a similar page from the answer.

### Related component

Search

### Describe alternatives you've considered

N/A

### Additional context

N/A

[^1]: We can compute a hard upper limit for listing records: there's a 10k records response limit, and each record is limited by OpenSearch to be at most 1 MB. So the largest response you could get is 10GB. In practice, this will be significantly reduced by the above factors.

mch2:
@jainankitk @kaushalmahi12 Wondering if either of you have experience with this from the wlm work, do you know how the system responds here so we can document correctly?

kaushalmahi12:
As per my understanding response size is indirectly limited by following
*  Maximum number of docs returned in the single response. (i,e; 10000)
* `max_search_buckets`: for aggs

PR #7963, opened 2023-06-08
Check UTF16 string size before converting to String to avoid OOME

imRishN:
### Description
Node can drop out for OutOfMemory errors other than the VM OutOfMemory which comes from the implementation limits of StringUTF16.

```
java.lang.OutOfMemoryError: UTF16 String size is 1089861360, should be less than 1073741823
```

This PR attempts to pre check the size and throw an exception proactively before trying a conversion.

### Related Issues
Resolves #1651 

### Check List
- [x] New functionality includes testing.
  - [x] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [x] Commits are signed per the DCO using --signoff
- [x] Commit changes are listed out in CHANGELOG.md file (See: [Changelog](../blob/main/CONTRIBUTING.md#changelog))

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

Issue #1651, opened 2021-12-03
Node drops out with OutOfMemoryError for reasons other than the VM having run out of memory


Bukhtawar:
**Describe the bug**
Node can drop out for OutOfMemory errors other than the VM OutOfMemory like 
```
java.lang.OutOfMemoryError: UTF16 String size is 1089861360, should be less than 1073741823
```
This is coming from the implementation limits of StringUTF16.

```
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:362) ~[?:?]
    at org.elasticsearch.transport.netty4.Netty4MessageChannelHandler.channelRead(Netty4MessageChannelHandler.java:53) ~[?:?]
    at org.elasticsearch.transport.TcpTransport.inboundMessage(TcpTransport.java:763) ~[elasticsearch-6.8.0.jar:6.8.0]
    at org.elasticsearch.transport.TcpTransport.messageReceived(TcpTransport.java:952) ~[elasticsearch-6.8.0.jar:6.8.0]
    at org.elasticsearch.transport.TcpTransport.handleResponse(TcpTransport.java:977) ~[elasticsearch-6.8.0.jar:6.8.0]
    at org.elasticsearch.common.util.concurrent.EsExecutors$DirectExecutorService.execute(EsExecutors.java:216) ~[elasticsearch-6.8.0.jar:6.8.0]
    at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:37) ~[elasticsearch-6.8.0.jar:6.8.0]
    at org.elasticsearch.transport.TcpTransport$1.doRun(TcpTransport.java:985) ~[elasticsearch-6.8.0.jar:6.8.0]
    at org.elasticsearch.transport.TransportService$ContextRestoreResponseHandler.handleResponse(TransportService.java:1104) ~[elasticsearch-6.8.0.jar:6.8.0]
    at org.elasticsearch.action.search.SearchTransportService$ConnectionCountingHandler.handleResponse(SearchTransportService.java:454) ~[elasticsearch-6.8.0.jar:6.8.0]
    at org.elasticsearch.action.ActionListenerResponseHandler.handleResponse(ActionListenerResponseHandler.java:54) ~[elasticsearch-6.8.0.jar:6.8.0]
[...]

PR #22356, opened 2026-06-30
Prevent REST response overflow and double-close failures

### LINKED ISSUE #22311: [BUG] ArithmeticException integer overflow in BytesRestResponse causes silent 504 hang for responses >= ~715 MB
## Describe the bug

Two stacked bugs in the REST response path combine to produce a **silent 5-minute hang** (zero bytes flushed to client, 504 from the load balancer) whenever a search response body reaches ~715 MB.

### Bug A — integer overflow in `BytesRestResponse(String)` serialization path

`BytesRestResponse(RestStatus, String contentType, String content)` calls `new BytesArray(content)`, which calls `new BytesRef(text)`, which calls `UnicodeUtil.maxUTF8Length(text.length())`. That method returns `Math.multiplyExact(utf16Length, MAX_UTF8_BYTES_PER_CHAR)` where `MAX_UTF8_BYTES_PER_CHAR == 3`. For ASCII-dominant JSON this overflows once `text.length() > Integer.MAX_VALUE / 3 = 715_827_882` (~715 MB). The resulting `ArithmeticException` is thrown **after** the channel is half-closed, so no error response is flushed.

Upstream precedent: OS#1651 / PR #7963 fixed the same overflow on the **request** (ingest) side. The response side was never patched.

### Bug B — non-idempotent `close()` in `ResourceHandlingHttpChannel`

`RestController$ResourceHandlingHttpChannel.close()` uses `AtomicBoolean.compareAndSet(false, true)` and throws `IllegalStateException("Channel is already closed")` when called a second time. When Bug A's exception is thrown mid-`sendResponse`, the first `close()` call half-closes the channel. `RestActionListener.onFailure` then tries to send the error response by calling `sendResponse` again, which re-enters `close()` and throws again. This leaves the underlying Netty channel dangling until the upstream load balancer times out (e.g. 300 s).

## Related component

Search

### Bug A

```java
// Reproducer (no large buffer allocation — overflow fires before any char access)
int overflowingLength = (Integer.MAX_VALUE / 3) + 1; // 715_827_883
CharSequence text = new CharSequence() {
    @Override public int length() { return overflowingLength; }
    @Override public char charAt(int i) { throw new AssertionError("not reached"); }
    @Override public CharSequence subSequence(int s, int e) { throw new UnsupportedOperationException(); }
};
assertThrows(ArithmeticException.class, () -> new BytesArray(new BytesRef(text)));
```

### Bug B

```java
// Proves that a second sendResponse throws instead of no-op'ing
restController.registerHandler(GET, "/repro-bug-b", (req, channel, c) -> {
    channel.sendResponse(new BytesRestResponse(OK, TEXT_CONTENT_TYPE, BytesArray.EMPTY));
    // second call mirrors RestActionListener.onFailure after Bug A fires
    assertThrows(IllegalStateException.class, () ->
        channel.sendResponse(new BytesRestResponse(INTERNAL_SERVER_ERROR, TEXT_CONTENT_TYPE, BytesArray.EMPTY)));
});
```

Both reproducer tests are included in the companion PR.

## Expected behavior

**Bug A:** Pre-check the response size before crossing into Lucene's `UnicodeUtil.maxUTF8Length`. If `(long) content.length() * 3 > Integer.MAX_VALUE`, fail with a typed exception that maps to a clean HTTP 413/507 error response — not a raw `ArithmeticException` thrown mid-write. Mirror the pattern from PR #7963 (request side).

**Bug B:** `ResourceHandlingHttpChannel.close()` should be idempotent — a second call should be a no-op, not throw `IllegalStateException`. This allows the error path in `RestActionListener.onFailure` to reach the client gracefully even if the success path already closed the channel.

## Additional Details

**Observed failure mode (production stack trace, OS 3.1):**

```
[WARN ][r.suppressed] path: /<index>/_search
java.lang.ArithmeticException: integer overflow
  at java.lang.Math.multiplyExact(Math.java:992)
  at org.apache.lucene.util.UnicodeUtil.maxUTF8Length(UnicodeUtil.java:676)
  at org.apache.lucene.util.BytesRef.<init>(BytesRef.java:80)
  at org.opensearch.core.common.bytes.BytesArray.<init>(BytesArray.java:56)
  at org.opensearch.rest.BytesRestResponse.<init>(BytesRestResponse.java:89)
  at ...HttpResponseChannel.sendResponse(HttpResponseAdapter.java:140)
  at AbstractSearchAsyncAction.sendSearchResponse(AbstractSearchAsyncAction.java:773)
  at ExpandSearchPhase.run(ExpandSearchPhase.java:132)

[ERROR][o.o.r.a.RestResponseListener] failed to send failure response
java.lang.IllegalStateException: Channel is already closed
  at RestController$ResourceHandlingHttpChannel.close(RestController.java:648)
  at RestController$ResourceHandlingHttpChannel.sendResponse(RestController.java:641)
  at RestActionListener.onFailure(RestActionListener.java:88)
  ...
  Suppressed: java.lang.ArithmeticException: integer overflow  (Bug A, above)
```

**Cliff value:** `utf16Length > Integer.MAX_VALUE / 3 = 715_827_882` (~715 MB ASCII JSON). Confirmed on cluster with 74x m7i.4xlarge data nodes; responses <=650 MB succeed (~20 s), responses >=700 MB hang the full 300 s load-balancer timeout with zero bytes flushed to the client.

**Version:** OS 3.1 (stack trace); both code paths are present in current `main`.

**Related issues:**
- #17885 — documentation gap on max response size (related, not a fix)
- #1651 / PR #7963 — identical overflow fixed for the request (ingest) path in 2023; this is the response-path equivalent
- Apache Lucene JIRA LUCENE-7538 — same overflow class in the indexing path (fixed 2016)

venkateshwaracholan:
## Summary

Fix two related bugs in the REST response path that can cause a silent ~5-minute hang (zero bytes flushed, LB 504) when search response bodies exceed ~715 MB.

- **Bug A:** Prevent integer overflow during UTF-8 encoding of oversized response strings.
- **Bug B:** Make `ResourceHandlingHttpChannel` and `StreamHandlingHttpChannel` `close()` idempotent so failure responses can still be sent after a partial `sendResponse()` failure.

Fixes #22311.

---

## Problem

Large string responses are encoded through:

```text
BytesRestResponse
  → BytesArray
    → BytesRef
      → UnicodeUtil.maxUTF8Length()
```

When the UTF-16 string length exceeds `Integer.MAX_VALUE / 3`, Lucene throws an `ArithmeticException`.

Because the response channel has already been partially closed, `RestActionListener.onFailure()` attempts another `sendResponse()`, which fails with `IllegalStateException("Channel is already closed")`, leaving the connection hanging until the load balancer times out.

---

### Bug A

Validate the UTF-16 length before constructing `BytesRef` and return a typed `OpenSearchStatusException` (HTTP **413**) instead of exposing an uncaught `ArithmeticException`.

> The issue discussion mentioned either **413** or **507**. This PR currently uses **413** and can be updated if maintainers prefer **507**.

### Bug B

Make channel `close()` idempotent so repeated calls become a no-op instead of throwing, allowing the failure response to be sent correctly.

---

## Evidence / Tests

| Test | Verifies |
|------|----------|
| `BytesArrayTests.testOverflowGuardPreventsBytesRefIntegerOverflow` | Prevents integer overflow without allocating a ~715 MB string |
| `BytesRestResponseTests.testToBytesArrayMapsOverflowGuardFailureToRequestEntityTooLarge` | Maps overflow to `OpenSearchStatusException` (HTTP 413) |
| `RestControllerTests.testResourceHandlingChannelCloseIsIdempotent` | Multiple `sendResponse()` calls no longer fail due to repeated `close()` |

---

## Validation

```bash
./gradlew :libs:opensearch-core:test --tests "org.opensearch.core.common.bytes.BytesArrayTests"

./gradlew :server:test --tests "org.opensearch.rest.BytesRestResponseTests.testToBytesArrayMapsOverflowGuardFailureToRequestEntityTooLarge"

./gradlew :server:test --tests "org.opensearch.rest.RestControllerTests.testResourceHandlingChannelCloseIsIdempotent"
```

All tests pass locally.

---

## Related

- Fixes #22311
- Mirrors the request-side fix in #7963
- Related: #1651
- Related: #17885

---

### Checklist

- [x] Functionality includes tests.
- [x] API changes companion PR not required.
- [x] Documentation changes not required.

### Check List


sandeshkr419:
[Search Triage] @asimmahmood1 Can you please help with the review.

asimmahmood1:
Thanks for picking this up @venkateshwaracholan while I was out. 

Overall this looks good to me. I was trying to find if there's a way to add an integ test to make sure its working as expected end to end, could you see if can add something in https://github.com/opensearch-project/OpenSearch/tree/main/qa/smoke-test-http/src/test/java/org/opensearch/http ?

kkhatua:
@venkateshwaracholan thanks for fixing this long lingering bug!
Could you please add an integ test for sake of completeness of this PR? 

@asimmahmood1  / @bowenlan-amzn  / @jainankitk  can do the final review and merge then.

PR #22359, opened 2026-06-30
Add cross-cluster streaming support to the remote cluster client

rishabhmaurya:
### Description

Adds **cross-cluster streaming support to the remote cluster client**: a plugin can open an Arrow Flight stream to a *remote cluster* (mirroring the existing regular remote-client request path), with a poll-able cancellation signal so a long-lived sender doesn't leak when the consumer goes away, plus an opt-in synchronous send for callers that need a hard bound on outstanding batches.

This is transport infrastructure underneath cross-cluster replication's streaming GetChanges, but it carries **no replication-specific code** and is reusable by any streaming cross-cluster action. Three pieces:

#### 1. Stream-aware remote client
`RemoteClusterAwareClient.sendStreamRequest(...)` sends a streaming request to a remote cluster over the stream (Arrow Flight) transport, with the same connect + node-selection as the existing `doExecute` request/response path; `RemoteClusterService.getRemoteClusterClient(threadPool, alias, streamTransportService)` builds such a client. No fallback to the regular transport — a missing stream transport throws. The class is `public @ExperimentalApi` because streaming has no polymorphic home on the `Client` interface (streaming is initiated on `StreamTransportService`, not `Client`), so a cross-module caller holds this concrete type to reach `sendStreamRequest`.

#### 2. Optional synchronous send (`sendResponseBatch(response, sync)`)
By default a batch send is dispatched to the channel's send executor and the producer returns once it is queued. The new opt-in overload lets a caller instead **block until the batch is on the wire**:
- `sendResponseBatch(response)` — async (default, unchanged).
- `sendResponseBatch(response, true)` — the send still runs on the channel's send executor, but the caller blocks until it completes, so it cannot queue the next batch ahead of this one. Outstanding batches are bounded to one (useful when batches hold large buffers that must not accumulate).

The send always runs on the same single-threaded executor that the channel's `close()` posts its stream-root free to, so a batch send and a concurrent `close()` are FIFO-serialized on that executor and cannot race on the stream root. A caller that opts into `sync` must drive a single stream from a single thread and must not call from the send-executor thread.

#### 3. Cancellation + liveness
`TransportChannel.isCancelled()` (default method, forwarded by `TaskTransportChannel`) lets a parked sender poll for consumer cancellation instead of leaking until its next send throws; plus gRPC server keepalive on the Flight server (`flight.keepalive.time`/`timeout`, defaulting to gRPC's own 2h/20s so behaviour is unchanged out of the box) to free senders on half-open connections.

### Related Issues
N/A — extracted from internal cross-cluster-replication streaming work for upstream review.

### Check List
- [x] Functionality includes testing (stream client connect/route + no-fallback contract, sync send runs on the executor and blocks, sync failure surfaces as StreamException, async dispatch, `TaskTransportChannel` forwarding, keepalive defaults + validator + applied to the gRPC builder).
- [x] Public APIs annotated (`@ExperimentalApi`).
- [x] Docs updated (`backpressure.md`).

### DIFF
--- a/plugins/arrow-flight-rpc/docs/backpressure.md
+++ b/plugins/arrow-flight-rpc/docs/backpressure.md
@@ -108,7 +108,7 @@ doesn't formally bound this.
 
 A byte-aware bounded queue at the eventloop entry — that parks (or rejects)
 the producer when the sum of queued batch sizes crosses a per-channel cap —
-would close the gap. Open design questions:
+would close the gap for the async path. Open design questions:
 
 - **Cap dimension**: byte-aware (sum of retained sizes) vs depth-aware (count).
   Bytes is correct for OOM protection.
@@ -119,6 +119,19 @@ would close the gap. Open design questions:
   budget tied to `flight pool max`.
 
 Tracked separately; not addressed in this change.
+
+### Synchronous send (bounded, opt-in)
+
+A caller that needs a hard bound *today* can send synchronously:
+`channel.sendResponseBatch(response, /* sync */ true)`. The batch is still
+serialized and written on the channel's send executor (the same executor the
+async path uses), but the calling thread **blocks until the batch is on the
+wire**, so it cannot queue the next batch ahead of this one — outstanding batches
+are bounded to one, with no eventloop queue to grow.
+
+A caller that opts in **must drive a single stream from a single thread** and must
+not call from the send-executor thread itself. The default
+`sendResponseBatch(response)` is unchanged — async, via the eventloop.
 
 ### ⚠️ Virtual threads for park-bound producers
 
--- a/plugins/arrow-flight-rpc/docs/backpressure.md
+++ b/plugins/arrow-flight-rpc/docs/backpressure.md
@@ -125,9 +125,11 @@ Tracked separately; not addressed in this change.
 A caller that needs a hard bound *today* can send synchronously:
 `channel.sendResponseBatch(response, /* sync */ true)`. The batch is still
 serialized and written on the channel's send executor (the same executor the
-async path uses), but the calling thread **blocks until the batch is on the
-wire**, so it cannot queue the next batch ahead of this one — outstanding batches
-are bounded to one, with no eventloop queue to grow.
+async path uses), but the calling thread **blocks until the batch has been
+pushed to gRPC's per-stream outbound buffer**, so it cannot queue the next batch
+ahead of this one — outstanding batches are bounded to one, with no eventloop
+queue to grow. When that outbound buffer is full, the `isReady()` back-pressure
+gate (above) throttles the producer.
 
 A caller that opts in **must drive a single stream from a single thread** and must
 not call from the send-executor thread itself. The default
--- a/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
+++ b/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
@@ -168,8 +168,9 @@ class FlightOutboundHandler extends ProtocolOutboundHandler {
                     messageListener.onResponseSent(requestId, action, e);
                 }
             });
-            // sync blocks the caller until the batch is on the wire (bounding outstanding batches to one);
-            // async returns once it is queued. Both run on the channel executor, and both transfer source
+            // sync blocks the caller until the batch has been pushed to gRPC's outbound buffer (bounding
+            // outstanding batches to one; a full buffer is throttled by the readiness gate above); async
+            // returns once it is queued. Both run on the channel executor, and both transfer source
             // ownership to that task the moment it is accepted — so handedOff is set before any blocking
             // wait, ensuring the finally's releaseUnsent never double-frees a source the task already owns.
             if (sync) {
@@ -189,9 +190,10 @@ class FlightOutboundHandler extends ProtocolOutboundHandler {
 
     /**
      * Blocks the caller until the executor has run the submitted send, so the caller cannot submit the
-     * next batch until this one is on the wire. The send itself runs on the channel's flight executor (so
-     * it is serialized with the stream-root free that {@code close()} posts to the same executor); this
-     * only parks the caller for the result. The caller must be a producer thread, not the flight executor
+     * next batch until this one has been pushed to gRPC's outbound buffer (a full buffer is throttled by
+     * the readiness back-pressure gate). The send itself runs on the channel's flight executor (so it is
+     * serialized with the stream-root free that {@code close()} posts to the same executor); this only
+     * parks the caller for the result. The caller must be a producer thread, not the flight executor
      * thread itself (blocking on the result from that thread would deadlock the single-threaded executor).
      */
     private void awaitSend(Future<?> future) {
--- a/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
+++ b/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
@@ -26,11 +26,14 @@ import org.opensearch.transport.TransportMessageListener;
 import org.opensearch.transport.TransportRequest;
 import org.opensearch.transport.TransportRequestOptions;
 import org.opensearch.transport.nativeprotocol.NativeOutboundMessage;
+import org.opensearch.transport.stream.StreamErrorCode;
 import org.opensearch.transport.stream.StreamException;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 
 /**
  * Outbound handler for Arrow Flight streaming responses.
@@ -120,7 +123,8 @@ class FlightOutboundHandler extends ProtocolOutboundHandler {
         final String action,
         final TransportResponse response,
         final boolean compress,
-        final boolean isHandshake
+        final boolean isHandshake,
+        final boolean sync
     ) throws IOException {
         BatchTask task = new BatchTask(
             nodeVersion,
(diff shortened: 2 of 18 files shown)


andrross:
@rishabhmaurya Are the three things listed in the description here related? I don't understand what items 2 and 3 have to do with a stream-aware remote client.

rishabhmaurya:
@andrross we can decouple them into separate PR. They are related here due to the nature of the long lived connections we will be creating in CCR. Synchronous send is important for integration with virtual threads and backpressure in CCR. Cancellation and keepalive are needed for long lived nature of stream and better control on it instead of relying on grpc defaults.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportChannel.java
```diff
@@ -35,6 +36,11 @@ class FlightTransportChannel extends TcpTransportChannel implements ArrowFlightC
     private static final Logger logger = LogManager.getLogger(FlightTransportChannel.class);
 
     private final AtomicBoolean streamOpen = new AtomicBoolean(true);
+    // Whether batch sends on this stream run synchronously on the calling thread (vs the channel
+    // executor). Latched by the first batch from (caller-requested sync || the response type requires
```
andrross: What does "latched by the first batch" mean here? I see this boolean being set on every invocation of sendResponseBatch()

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
```diff
@@ -145,25 +146,33 @@ public void sendResponseBatch(
...
+            // not our queue depth — see docs/backpressure.md "Known limitation: unbounded eventloop queue".
             flightChannel.awaitReadyOrThrow();
 
-            flightChannel.getExecutor().execute(threadPool.getThreadContext().preserveContext(() -> {
+            // A cancelled/stalled stream surfaces to the caller from awaitReadyOrThrow() above (it
+            // throws StreamException directly on the calling thread); processBatchTask itself terminates
+            // the stream and notifies the listener on a send failure, identically on both lanes.
+            if (sync) {
                 try (BatchTask ignored = task) {
                     processBatchTask(task);
```
andrross, line 166: I think the sync path might break the expectation here:

https://github.com/opensearch-project/OpenSearch/blob/3bff5f574d18faca44ba9ad7cbd56a67aa32b566/plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightServerChannel.java#L482-L492

As I understand it, that cancellation event can be triggered independently (e.g. if the other end of the connection is closed) and it expects to be able to serialize the resulting `root.close();` with respect to all other operations. That holds if all these operations are posted to the same thread, but the sync path now lets the send operations happen directly on any arbitrary calling thread.

andrross:
> Synchronous send is important for integration with virtual threads

@rishabhmaurya Another way to solve this problem is to let the producer manage it's own permit system for how many in-flight sends it can allow, with something like:

```
semaphore.acquire();
channel.sendResponseBatch(batch, ActionListener.wrap(semaphore::release, e -> { semaphore.release(); fail(e); }));
```

It looks like sendResponseBatch() doesn't have a callback to notify the caller that the batch has been pulled off of its executor queue and written to the outbound buffer. That would allow the caller to limit how many in-flight requests there are queued up.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransportChannel.java
rishabhmaurya: this comment looks inaccurate. The main idea is, if someone is using sync API, they must continue to use sync and not change in between batches as that may break the contract of FIFO. Also, sync API is meant to be single threaded to guarantee FIFO. This is covered as part of `sendResponseBatch` API contract in `TransportChannel`.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
rishabhmaurya: this is a great catch! it may delay the close and also breaks the invariant for FlightServerChannel. Let me think more on how to handle this.

andrross: To be clear, the specific concern is that the sync calling thread could be inside FlightServerChannel.transferIntoStreamRoot() while an async close event happens and does VectorSchemaRoot.close() concurrently on another thread. Here is an AI analysis of the possible outcomes:

  Failure modes, by interleaving:

  1. Entry check wins (narrow, benign). close() completes and the caller then enters sendBatch and sees cancelled/!open → throws StreamException/IllegalStateException, source freed by the finally. This is the only genuinely graceful outcome, and it's the small window.
  2. First-batch, close lambda runs before root = created is published (leak). On the first batch root is still null when the caller is inside transferIntoStreamRoot building created. The posted lambda if (root !=
  null) { root.close(); } sees null and does nothing. The caller then publishes root = created and calls start/putNext on an already-closed/cancelled gRPC stream. close() already consumed the CAS and will never run
  again, and sendBatch's finally frees only sourceRoot, not created. So the stream root leaks (the documented free-at-close never happens).
  3. Later batch (or first-batch after publish) — the dangerous one. The caller is inside VectorTransfer.transferRoot(sourceRoot, root) (moving buffers into root) or serializeIntoStreamRoot (writing/reallocating a VarBinary vector in root) or putNext(root) while the executor thread runs root.close() on the same vectors. Now:
    - The vector's internal structure (its buffer references, value counts, the allocator's accounting) is mutated by two threads with no synchronization →
  NullPointerException/ArrayIndexOutOfBoundsException/IllegalStateException ("RefCnt has gone negative"), or a double-free. These might be caught by processBatchTask's catch (Throwable), but by then the
  allocator/refcount state can already be corrupted, and failStream then touches the same channel/root again.
    - Worse, if close() releases the last reference and returns the off-heap memory to the OS while the caller is still reading/writing it → use-after-free: a JVM SIGSEGV/hard crash, or silent heap corruption that
  surfaces arbitrarily later somewhere unrelated.

rishabhmaurya: breaking invariant of `FlightServerChannel `to be operated by thread other than its `executor` is a bad idea.
In the new revision, I have ensured this invariant isn't broken and all mutations to `FlightServerChannel` and vectorSchemaRoot is single threaded. 
We are now creating a future and wait on its completion but its still executed on `FlightServerChannel`'s executor.

plugins/arrow-flight-rpc/docs/backpressure.md
```diff
@@ -120,6 +120,19 @@ would close the gap. Open design questions:
 
 Tracked separately; not addressed in this change.
 
+### Synchronous send (bounded, opt-in)
+
+A caller that needs a hard bound *today* can send synchronously:
+`channel.sendResponseBatch(response, /* sync */ true)`. The batch is still
+serialized and written on the channel's send executor (the same executor the
+async path uses), but the calling thread **blocks until the batch is on the
+wire**, so it cannot queue the next batch ahead of this one — outstanding batches
```
andrross: I think "until the batch is on the wire" is not quite right. I believe the call returns once the batch is enqueued into gRPC's outbound buffer, not on the wire. If that outbound buffer fills up then the other back pressure mechanism kicks in.

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightTransport.java
```diff
@@ -251,6 +251,17 @@ private List<InetSocketAddress> bindToPort(InetAddress[] hostAddresses) {
...
+                // Server-side gRPC keepalive (see ServerConfig.FLIGHT_KEEPALIVE_TIME). NOTE: only the
+                // server pings today; adding a client keepalive also requires
+                // permitKeepAliveTime/permitKeepAliveWithoutCalls here, else the server GOAWAYs the
+                // client with "too_many_pings".
+                final long keepAliveTimeMs = ServerConfig.getGrpcKeepAliveTime().millis();
+                final long keepAliveTimeoutMs = ServerConfig.getGrpcKeepAliveTimeout().millis();
+                builder.transportHint("grpc.builderConsumer", (java.util.function.Consumer<io.grpc.netty.NettyServerBuilder>) b -> {
+                    b.keepAliveTime(keepAliveTimeMs, java.util.concurrent.TimeUnit.MILLISECONDS);
+                    b.keepAliveTimeout(keepAliveTimeoutMs, java.util.concurrent.TimeUnit.MILLISECONDS);
+                });
```
andrross: No need for the fully qualified class names

plugins/arrow-flight-rpc/src/main/java/org/opensearch/arrow/flight/transport/FlightOutboundHandler.java
```diff
@@ -157,21 +161,56 @@ public void sendResponseBatch(
...
+            final Runnable sendBatch = threadPool.getThreadContext().preserveContext(() -> {
                 try (BatchTask ignored = task) {
                     processBatchTask(task);
                 } catch (Exception e) {
                     messageListener.onResponseSent(requestId, action, e);
                 }
-            }));
-            handedOff = true;
+            });
+            // sync blocks the caller until the batch is on the wire (bounding outstanding batches to one);
```
andrross: Same nitpick about "on the wire".

plugins/arrow-flight-rpc/docs/backpressure.md
rishabhmaurya: that's exactly the right behaviour, let me correct.

andrross (approved):

PR #22360, opened 2026-06-30
[analytics-engine] Improve TopK correctness with CSS: replace Final with PartialReduce using topk check

sandeshkr419:
Co-Authored-By: @mch2 & @expani 

## Problem

  When concurrent segment search (CSS) is active, DataFusion's physical optimizer inserts `RepartitionExec(RoundRobin, N) → AggregateExec(Partial, ×N) → AggregateExec(FinalPartitioned)` on the shard fragment. `prepare_partial_plan` calls `force_aggregate_mode(Partial)` which strips `FinalPartitioned` and returns the `Partial(×N)` subtree directly.

  Each of the N CSS partitions then runs the TopK `SortExec` independently, truncating to the TopK fetch limit before the coordinator merge. Groups split across partitions are dropped, producing incorrect counts (e.g. q31 on clickbench dataset returns `c=313` instead of `c=1633`).

  ## Fix

  Propagate a `hasTopK` boolean from `OpenSearchTopKRewriter` through the planner pipeline:

  `PlannerContext` → `FragmentConversionDriver` → `PartialAggregateInstructionNode` → `ShardScanExecutionContext` → `NativeBridge` → Rust `create_session_context` → `SessionContextHandle`

  In `prepare_partial_plan`, when `has_topk=true`, `force_aggregate_mode` replaces `FinalPartitioned` with `PartialReduce` instead of stripping it. `PartialReduce` keeps `agg.input()` intact — `RepartitionExec(Hash) → Partial(×N)` — so CSS partitions are merged by group key before the TopK `SortExec` truncates. The hash repartition ensures each group key lands in exactly one partition, so `PartialReduce` produces complete per-key partial state.

  CSS scan parallelism is fully preserved: the scan + partial agg still run ×N in parallel. Only a merge step is added before TopK. Non-TopK queries (`has_topk=false`) are unaffected.

 
  ## Changes

  - **Java**: `PlannerContext.setTopKApplied` set by `PlannerImpl` after `OpenSearchTopKRewriter` fires; threaded through `FragmentConversionDriver` → `PartialAggregateInstructionNode.hasTopK` → `ShardScanExecutionContext.hasTopK` →`NativeBridge.createSessionContext`
  - **Rust `agg_mode.rs`**: `force_aggregate_mode` accepts `has_topk: bool`; when true and hitting `Final/FinalPartitioned`, returns `PartialReduce(input=agg.input())` instead of stripping. Removes Marc's `partition_count() > 1` check — `has_topk` from the planner is the authoritative signal
  - **Rust `session_context.rs`**: `SessionContextHandle.has_topk` field; `prepare_partial_plan` passes it to `apply_aggregate_mode`
  - **Plan shape goldens**: 29 queries (q8–q43, prod2s) updated to show `AggregateExec(PartialReduce)` above `RepartitionExec(Hash) → AggregateExec(Partial)`

  ## Testing

  - `integTest`: full QA suite passes
  - `integTestPlanShape`: all 29 affected golden files verified
  - Unit tests pass

### Check List


sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
expani: This needs to be gated by TopK presence not for all Partial Aggregates.

It might be hamper the performance by inserting into HashTable and possibly make query more memory intensive.

expani: nit: duplicated twice.

sandbox/plugins/analytics-backend-datafusion/rust/src/agg_mode.rs
```diff
@@ -71,32 +75,45 @@ fn force_aggregate_mode(
...
         }
         // Mode mismatch — strip this node
         match target {
             AggregateMode::Partial => {
-                // Current node is Final; find the Partial subtree below
+                // Current node is Final/FinalPartitioned.
+                // When TopK is active, replace with PartialReduce instead of stripping.
+                // PartialReduce keeps agg.input() (RepartitionExec(Hash) → Partial(×N))
+                // so CSS partitions are merged by group key before TopK truncation.
+                if has_topk {
```
expani: Even if TopK exist we can also gate on the target_partitions of the Child operator being > 1 otherwise PartialReduce becomes redundant.

sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
```diff
@@ -448,20 +461,57 @@ pub async fn prepare_partial_plan(
...
+
+/// Returns true if the Substrait plan bytes contain a FetchRel (Sort+Limit node).
+/// A FetchRel in a shard fragment means `OpenSearchTopKRewriter` inserted a per-shard
+/// Sort+Limit — TopK is active. Used in `create_session_context` to detect TopK before
+/// the DataFusion physical plan is built, so the result can be stored on the handle and
+/// reused in `prepare_partial_plan` without re-scanning the physical plan.
+///
+/// Single-shard (SINGLE aggregate mode) never has `has_partial_aggregate=true` so this
+/// function is only called for multi-shard partial-aggregate fragments.
+fn substrait_has_fetch_rel(plan_bytes: &[u8]) -> bool {
```
expani, line 500: Let's add a TODO to do this via AnalyticsCore using a Flag later. 
Any maybe add a one liner mentioning the Upgrade path issue of adding new fields.

mch2 (approved):

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/TopKCssCorrectnessIT.java
```diff
@@ -0,0 +1,297 @@
...
+
+    private static volatile boolean provisioned = false;
+    private static final String INDEX = "parquet_hits";
+
+    private void ensureProvisioned() throws Exception {
+        if (!provisioned) {
+            DatasetProvisioner.provision(client(), ClickBenchTestHelper.DATASET, 2);
+            Request req = new Request("PUT", "/_cluster/settings");
+            req.setJsonEntity(
+                "{\"persistent\":{\"analytics.shard_bucket_oversampling_factor\": 2.0}}"
```
expani: Are we able to simulate incorrect result with a higher oversampling factor ? 
I was able to only reproduce it with 0.1

Do these tests fail without fix ?

sandbox/qa/analytics-engine-rest/src/test/resources/planshape/clickbench/q15.plan.yaml
```diff
@@ -34,18 +34,22 @@ plans:
...
         SortPreservingMergeExec: [count(Int64(1))@2 DESC NULLS LAST, SearchEngineID@0 ASC, SearchPhrase@1 ASC], fetch=30
           SortExec: TopK(fetch=30), expr=[count(Int64(1))@2 DESC NULLS LAST, SearchEngineID@0 ASC, SearchPhrase@1 ASC], preserve_partitioning=[true]
-            AggregateExec: mode=Partial, gby=[SearchEngineID@0 as SearchEngineID, SearchPhrase@1 as SearchPhrase], aggr=[count(Int64(1))]
-              FilterExec: SearchPhrase@1 != 
-                RepartitionExec: partitioning=RoundRobinBatch(4), input_partitions=1
-                  DataSourceExec: file_groups={<scrubbed>}, projection=[SearchEngineID, SearchPhrase], file_type=parquet, predicate=SearchPhrase@63 != , pruning_predicate=SearchPhrase_null_count@2 != row_count@3 AND (SearchPhrase_min@0 !=  OR  != SearchPhrase_max@1), required_guarantees=[SearchPhrase not in ()]
+            AggregateExec: mode=PartialReduce, gby=[SearchEngineID@0 as SearchEngineID, SearchPhrase@1 as SearchPhrase], aggr=[count(Int64(1))]
+              RepartitionExec: partitioning=Hash([SearchEngineID@0, SearchPhrase@1], 4), input_partitions=4
```
expani, line 38: We might need to see why Datafusion physical planner is inserting this layer of `RepartitionExec` again given all `shard_physical_1seg` already operate on 1 segment. 

Can you please add a TODO across a bunch of queries for this ? It might turn out to be expensive for performance and we can tackle as a fast followup.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchTopKRewriter.java
```diff
@@ -229,8 +230,17 @@ private static PathToFinal findFinalAgg(RelNode node, OpenSearchProject seenProj
         if (node instanceof OpenSearchAggregate agg && agg.getMode() == AggregateMode.FINAL) {
             return new PathToFinal(seenProject, agg);
         }
-        if (node instanceof OpenSearchProject proj && seenProject == null) {
-            return findFinalAgg(proj.getInput(), proj);
+        // Anything between the Sort and the FINAL that consumes its full grouped output makes
+        // the pushdown unsafe — refuse to match at all.
+        if (node instanceof OpenSearchAggregate) return null;                        // nested stats
```
expani, line 237: I think we should not be rejecting nested aggregates as the result being off for Correctness is purely due to lower default limits. 

Can you add a TODO to revisit this once we have TopK oversampling factor as an Execution hint ?

sandbox/plugins/analytics-backend-datafusion/rust/src/session_context.rs
```diff
@@ -448,20 +461,57 @@ pub async fn prepare_partial_plan(
...
+    use substrait::proto::rel::RelType;
+
+    fn rel_has_fetch(rel: &substrait::proto::Rel) -> bool {
+        match rel.rel_type.as_ref() {
+            Some(RelType::Fetch(f)) => f.count_mode.is_some(),
+            Some(RelType::Sort(s)) => s.input.as_ref().map_or(false, |r| rel_has_fetch(r)),
+            Some(RelType::Project(p)) => p.input.as_ref().map_or(false, |r| rel_has_fetch(r)),
+            Some(RelType::Filter(f)) => f.input.as_ref().map_or(false, |r| rel_has_fetch(r)),
+            Some(RelType::Aggregate(a)) => a.input.as_ref().map_or(false, |r| rel_has_fetch(r)),
+            _ => false,
```
expani: This is silently working in case of Single Shard say if there is a Join/Union then we return false and correctly assume that no TopK rewrite was done. 

I think we can be more explicit in those cases please add a comment for the same

expani: On a Second thought, let's explicitly `false` for All such operators and default case should panic with ERROR instead of returning false

sandeshkr419: adding a debug log, not panicking for now. added a todo as well.

sandbox/qa/analytics-engine-rest/src/test/java/org/opensearch/analytics/qa/TopKCssCorrectnessIT.java
sandeshkr419: made it 0.1 to make the tests tightened up - thanks for this.

expani (approved):

PR #22364, opened 2026-07-01
Add extensible payload decoder framework for pull-based ingestion

### LINKED ISSUE #22363: [Feature Request]  Support Avro-encoded messages in the ingestion-kafka plugin

The pull-based ingestion Kafka plugin currently requires messages to be UTF-8 JSON.
Users whose Kafka topics carry **Avro-binary payloads** — a very common serialization
format in Kafka ecosystems — cannot use pull-based ingestion without adding an external
preprocessing step to re-encode messages to JSON before they reach OpenSearch.

Add an opt-in Avro decoder that is activated by `avro.*` keys in the
`ingestion_source.param` map. When present, the plugin decodes each Avro-binary
message to JSON before handing it to the indexing pipeline. If no `avro.*` keys are
configured, existing behaviour is unchanged.

The schema can be provided inline or fetched from a schema registry URL at index
creation time. A `skip_bytes` option handles framing headers (e.g. the 5-byte
Confluent wire format prefix). An optional `msg_field` allows extracting a nested
field as the document source, supporting envelope-style message layouts.

**Example configuration:**

```json
PUT /my-index
{
  "settings": {
    "ingestion_source": {
      "type": "kafka",
      "pointer": { "init": { "reset": "earliest" } },
      "param": {
        "topic": "my-avro-topic",
        "bootstrap_servers": "localhost:9092",
        "avro.skip_bytes": 5,
        "avro.schema_registry_url": "https://schema-registry.example.com/subjects/my-topic/versions/latest"
      }
    }
  }
}
```

## Configuration Reference

| Parameter | Description |
|---|---|
| `avro.schema` | Inline Avro JSON schema |
| `avro.schema_registry_url` | URL to fetch the schema from at index creation |
| `avro.schema_registry_headers.<name>` | HTTP request headers for the registry (e.g. auth tokens) |
| `avro.schema_registry_connect_timeout_ms` | Connect timeout in ms (default: 10 000) |
| `avro.schema_registry_request_timeout_ms` | Request timeout in ms (default: 10 000) |
| `avro.skip_bytes` | Header bytes to strip before decoding (default: 0) |
| `avro.wrapper_schema` / `avro.wrapper_field` | Envelope schema with a substituted inner schema |
| `avro.msg_field` | Record field to use as the document source |

One of `avro.schema` or `avro.schema_registry_url` is required when using Avro decoding.

### Related component

_No response_

Tried using a pre-processing service which converted the avro bytes input into raw json and use that for opensearch

### Additional context

_No response_

Tarun-kishore:
## Summary

Introduces a reusable payload-decoding layer between ingestion consumers and message mappers for pull-based ingestion.

This enables plugins to register custom wire-format decoders (such as Avro or Protobuf) that convert raw message bytes directly into a field map, eliminating the JSON serialization/deserialization round-trip currently required by the ingestion pipeline.

### New public interfaces (`org.opensearch.index`)

- **`IngestionPayloadDecoder`**
  - Per-shard decoder.
  - `decode(BytesReference) -> Map<String, Object>`
  - Implements `Closeable`.

- **`IngestionPayloadDecoderFactory`**
  - Node-scoped factory.
  - Exposes:
    - `validate(Settings)`
    - `create(IndexMetadata, shardId, Settings)`
  - Implements `Closeable`.

- **`IngestionPayloadDecodingException`**
  - Unchecked exception representing payload decoding failures.

### Registry (`org.opensearch.indices.pollingingest`)

- Added `IngestionPayloadDecoderRegistry`, a named registry constructed during node startup.
- Fails fast if duplicate decoder names are registered.
- Added `XContentIngestionPayloadDecoder`, which preserves the existing JSON/XContent ingestion behavior.

### Index settings

- `index.ingestion_source.decoder_type`
  - Default: `xcontent`
  - `Final` setting.
  - Selects the decoder implementation.

- `index.ingestion_source.decoder_settings.*`
  - `AffixSetting`
  - `Final`
  - Supplies decoder-specific configuration to the selected decoder factory.

### Plugin extension point

`IngestionConsumerPlugin` now exposes a new default method:

```java
Map<String, IngestionPayloadDecoderFactory> getIngestionPayloadDecoderFactories()
```

Since the method has a default implementation, existing plugins continue to work without modification.

### Mapper contract

`IngestionMessageMapper.mapAndProcess(...)` now accepts a pre-decoded payload:

```java
Map<String, Object> decodedPayload
```

instead of parsing the payload internally using `IngestionUtils.getParsedPayloadMap(...)`.

## Ingestion flow

```text
Consumer (raw bytes)
        │
        ▼
IngestionPayloadDecoder.decode(BytesReference)
        │
        ▼
Map<String, Object>
        │
        ▼
IngestionMessageMapper.mapAndProcess(pointer, message, decodedPayload)
        │
        ▼
ShardUpdateMessage
        │
        ▼
PartitionedBlockingQueueContainer
```

Payload decoding occurs **before queue insertion**, since partition selection depends on the document ID generated by the mapper.

## Backward compatibility

- Existing indices continue using the built-in `xcontent` decoder without any configuration changes.
- `MetadataCreateIndexService.validateIngestionSourceSettings()` rejects non-`xcontent` decoder types on mixed-version clusters containing nodes older than **3.8.0**.
- `IngestionPayloadDecoderFactory.close()` is invoked during node shutdown, allowing implementations to clean up shared resources (for example, HTTP clients or schema caches).
- Per-shard decoder instances are closed when the corresponding shard poller shuts down.

## Related Issues

Resolves #22363

## Checklist

For more information on the Developer Certificate of Origin and signing off commits, see the [contributing guide](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


plugins/ingestion-kafka/build.gradle
```diff
@@ -22,13 +23,77 @@ versions << [
   'testcontainers': '1.19.7',
   'ducttape': '1.0.8',
   'snappy': '1.1.10.7',
+  'avro': '1.12.0',
```
varunbharadwaj: https://github.com/opensearch-project/OpenSearch/blob/main/DEVELOPER_GUIDE.md#adding-dependencies 

We have to copy the license and library SHAs when we add a new dependency. I see avro licenses added [here](https://github.com/opensearch-project/OpenSearch/tree/95ece9bad72ac507f654c94209bd08e9c6fd9d47/plugins/repository-hdfs/licenses) for reference. Let us do this for all new dependencies.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,503 @@
...
+        try {
+            HttpClient client = HttpClient.newBuilder()
+                .connectTimeout(Duration.ofMillis(connectTimeoutMs))
+                .build();
+            HttpRequest.Builder builder = HttpRequest.newBuilder()
+                .uri(URI.create(url))
+                .timeout(Duration.ofMillis(requestTimeoutMs))
+                .GET();
+            headers.forEach(builder::header);
+            HttpResponse<String> response = client.send(builder.build(), HttpResponse.BodyHandlers.ofString());
```
varunbharadwaj: Just curious - have we tested this connection if it works? What is the expectation on failures, do we retry?

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
```diff
@@ -254,7 +263,7 @@ private synchronized List<ReadResult<KafkaOffset, KafkaMessage>> fetch(long star
             long currentOffset = messageAndOffset.offset();
             lastFetchedOffset = currentOffset;
             KafkaOffset kafkaOffset = new KafkaOffset(currentOffset);
-            KafkaMessage message = new KafkaMessage(messageAndOffset.key(), messageAndOffset.value(), messageAndOffset.timestamp());
+            KafkaMessage message = new KafkaMessage(messageAndOffset.key(), payloadDecoder.decode(messageAndOffset.value()), messageAndOffset.timestamp());
```
varunbharadwaj: If `payloadDecoder.decode` throws an exception, we skip the remaining messages in the batch. On the next poller loop, it looks like we will start from the next batch resulting in skipping messages. Can we double check this?

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
```diff
@@ -73,6 +74,14 @@ protected KafkaPartitionConsumer(String clientId, KafkaSourceConfig config, int
         this.consumer = consumer;
         this.config = config;
         this.partitionId = partitionId;
+        Map<String, Object> avroParams = config.getAvroParams();
+        this.payloadDecoder = avroParams.isEmpty() ? KafkaPayloadDecoder.PASSTHROUGH : new AvroPayloadDecoder(avroParams);
```
varunbharadwaj: If `AvroPayloadDecoder` fails to be initialized, we can leak Kafka connections. There was a recent fix to handle this for `initialize` failures ([ref](https://github.com/opensearch-project/OpenSearch/blob/main/plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaConsumerFactory.java#L33)), but failures here in the constructor is not handled. We will have to move this to the initialize method or think or a better way outside the constructor.

plugins/ingestion-kafka/src/internalClusterTest/java/org/opensearch/plugin/kafka/AvroIngestionFromKafkaIT.java
```diff
@@ -0,0 +1,209 @@
...
+ * searchable with the expected field values.
+ *
+ * <p>The Avro schema encodes the full ingestion envelope:
+ * {@code _id}, {@code _op_type}, and {@code _source} (a nested record).
+ * After decoding, the JSON produced by {@link AvroPayloadDecoder} has the same
+ * structure as the plain-JSON messages used in {@link IngestFromKafkaIT}, so no
+ * changes to the ingestion pipeline are required.
+ */
+@OpenSearchIntegTestCase.ClusterScope(scope = OpenSearchIntegTestCase.Scope.TEST, numDataNodes = 0)
+public class AvroIngestionFromKafkaIT extends KafkaIngestionBaseIT {
```
varunbharadwaj: Let us add some tests on schema evolution, error handling (invalid message, null fields). Let's focus more on schema evolution, testing the different supported formats (including the confluent style format) with field addition cases. Along with this, we can also add a test to ensure ingestion works when the kafka topic has 2 different schema messages (without newly added fields + few with new fields) to test backward compatibility.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,503 @@
...
+        if (value instanceof Map) {
+            Map<String, Object> result = new HashMap<>();
+            ((Map<Object, Object>) value).forEach((k, v) -> result.put(String.valueOf(k), convertAvroValue(v)));
+            return result;
+        }
+        if (value instanceof CharSequence) return value.toString();
+        return value;
+    }
+
+    private static byte[] toJsonBytes(Map<String, Object> map) {
```
varunbharadwaj: We could reuse XContentBuilder, something like

```
private static byte[] toJsonBytes(Map<String, Object> map) {
      try {
          return BytesReference.toBytes(BytesReference.bytes(XContentFactory.jsonBuilder().map(map)));
      } catch (IOException e) {
          throw new IllegalArgumentException("Failed to serialize decoded Avro record as JSON", e);
      }
  }
```

We also do a similar mapping [here](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/indices/pollingingest/MessageProcessorRunnable.java#L323).

We can explore if this cleans up the code, combining with `convertAvroValue` being schema aware and coverting the avro types to plain Java types.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,503 @@
...
+            logger.debug("Parsed Avro schema directly, name=[{}]", schema.getFullName());
+            return schema;
+        } catch (Throwable e) {
+            if (!(e instanceof SchemaParseException)) {
+                logger.error("AvroPayloadDecoder: schema parse threw unexpected error: {}", e.getMessage(), e);
+                throw new IllegalStateException("Schema.Parser failed with unexpected error: " + e.getMessage(), e);
+            }
+            // Fall back to Confluent-style {"schema":"<escaped-json>"} wrapper
+            logger.debug("Direct parse failed ({}), trying Confluent schema-field extraction", e.getMessage());
+            int keyIdx = body.indexOf("\"schema\"");
```
varunbharadwaj: Is it possible to use a JSONParser or XContentParser here instead of doing it manually?

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,503 @@
...
+                if (nested instanceof Map == false) {
+                    throw new IllegalArgumentException(
+                        "avro.msg_field [" + msgField + "] must be a record, found: " + nested.getClass().getSimpleName()
+                    );
+                }
+                map = (Map<String, Object>) nested;
+                logger.debug("Extracted msg_field=[{}] with {} fields", msgField, map.size());
+            }
+
+            byte[] jsonBytes = toJsonBytes(map);
```
varunbharadwaj: It looks like we are converting the avro record into a Map, then back to json bytes here, and later down in the pipeline again back to a Map. This might need changes downstream as well, but wonder if we can avoid the serialize/parse back and forth? We can add a note if not easy to implement as a TODO.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,503 @@
...
+                    i++;
+                }
+            }
+            Schema schema = new Schema.Parser().parse(sb.toString());
+            logger.debug("Parsed Avro schema via Confluent extraction, name=[{}]", schema.getFullName());
+            return schema;
+        }
+    }
+
+    static Schema substituteInnerSchema(Schema wrapper, Schema inner, String wrapperField) {
```
varunbharadwaj: Wonder where this will be useful? I was thinking the avro payload will follow the exact format required by the PBI flow, with same field name requirements. Later the message mappers will be able to do the transformations?

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: We have tested the connection and the node was able to connect to schema registry. We do not retry the error as the error can be because of bad configs too. Should we retry in case of failure?

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: Some Kafka topics use an outer envelope record that carries routing/operational metadata alongside the actual document payload. An envelope may be like:

```
  {
    "type": "record",
    "name": "kafka_envelope",
    "fields": [
      { "name": "ts",        "type": "double"          },
      { "name": "host",      "type": "string"          },
      { "name": "schema_id", "type": ["null", "int"]   },
      { "name": "msg",       "type": "null"            }
    ]
  }
```

 The msg field is a placeholder — its actual schema is the document schema, which may be owned and versioned separately (e.g. registered independently in the schema registry). `avro.wrapper_schema` + `avro.wrapper_field` lets the user configure the envelope and the inner document schema independently. At index creation time they are merged by substituting the inner schema into the named wrapper_field. After decoding, avro.msg_field extracts just the inner record as the document source, discarding the envelope metadata fields like ts, host, dc etc.
Without this, the user would have to maintain one monolithic schema that hardcodes the full inner document definition inside the envelope, which breaks cleanly when the inner schema evolves independently under its own registry subject.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: Added a TODO comment, we can take up this optimization in future

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: Yes, updated to use XContentParser

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: Updated to use XContentFactory. Will explore change to convert avro type to Java type

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
Tarun-kishore: Moved the logic to `initialize` method which closes the consumer on failure.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
Tarun-kishore: Updated to have per message try-catch, so only bad message will be skipped

plugins/ingestion-kafka/build.gradle
Tarun-kishore: Added licenses for dependencies

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
```diff
@@ -254,7 +263,14 @@ private synchronized List<ReadResult<KafkaOffset, KafkaMessage>> fetch(long star
             long currentOffset = messageAndOffset.offset();
             lastFetchedOffset = currentOffset;
             KafkaOffset kafkaOffset = new KafkaOffset(currentOffset);
-            KafkaMessage message = new KafkaMessage(messageAndOffset.key(), messageAndOffset.value(), messageAndOffset.timestamp());
+            byte[] value;
+            try {
+                value = payloadDecoder.decode(messageAndOffset.value());
+            } catch (Exception e) {
+                logger.error("Failed to decode message at offset {}, skipping: {}", currentOffset, e.getMessage(), e);
```
varunbharadwaj: Skipping failed messages here does not look correct, as this decision is left to the poller and processor depending on user configured DROP/BLOCK strategies. I'm thinking one way to handle this could be to always return the KafkaMessage (marked as failure if there are decoding errors). Later when the payload is read from KafkaMessage, we can throw an exception. Alternatively, we could lazily decode the payload when KafkaMessage.getPayload() is called (with caching). This however needs more thinking and handle thread visibility requirements. Both these approaches should follow the existing DROP/BLOCK strategies. 

Please take a look and we can discuss if there are any challenges with this approach or if it doesn't make sense.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
```diff
@@ -73,6 +74,8 @@ protected KafkaPartitionConsumer(String clientId, KafkaSourceConfig config, int
         this.consumer = consumer;
         this.config = config;
         this.partitionId = partitionId;
+        this.payloadDecoder = KafkaPayloadDecoder.PASSTHROUGH;
+        logger.info("KafkaPartitionConsumer: topic={} partition={}", config.getTopic(), partitionId);
```
philiplhchan: nit: might as well add `payloadDecoder=passthrough` here for consistency and clarity

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
philiplhchan: i'm also not sure it's a good idea to log errors in the hot loop here. if 1 message errors, chances are all the messages will error, this will probably blow up your system just from the logging

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPayloadDecoder.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Strategy for transforming raw Kafka message bytes before they are wrapped in a {@link KafkaMessage}.
+ *
+ * <p>Implementations can strip wire-format headers, decode binary formats (Avro, Protobuf, …),
+ * and convert the result to whatever byte representation the downstream mapper expects.
+ *
+ * <p>The default no-op implementation is {@link #PASSTHROUGH}, which returns the raw bytes unchanged.
+ */
+@FunctionalInterface
+public interface KafkaPayloadDecoder {
```
philiplhchan: does this need to be limited to kafka?
what if someone wants to use this for kinesis?

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+     */
+    public AvroPayloadDecoder(Map<String, Object> params) {
+        int rawSkipBytes = params.containsKey(PARAM_SKIP_BYTES)
+            ? Integer.parseInt(String.valueOf(params.get(PARAM_SKIP_BYTES)))
+            : 0;
+        if (rawSkipBytes < 0) {
+            throw new IllegalArgumentException("[avro.skip_bytes] must be non-negative, got: " + rawSkipBytes);
+        }
+        this.skipBytes = rawSkipBytes;
+        this.msgField = params.containsKey(PARAM_MSG_FIELD) ? String.valueOf(params.get(PARAM_MSG_FIELD)) : null;
```
philiplhchan: i feel like all these `if contains then string else null` can be cleaned up by having a simple helper method

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+        String wrapperSchemaJson = params.containsKey(PARAM_WRAPPER_SCHEMA)
+            ? String.valueOf(params.get(PARAM_WRAPPER_SCHEMA))
+            : null;
+        String wrapperField = params.containsKey(PARAM_WRAPPER_FIELD) ? String.valueOf(params.get(PARAM_WRAPPER_FIELD)) : null;
+
+        if (registryUrl == null && inlineSchema == null) {
+            throw new IllegalArgumentException(
+                "avro params present but neither [avro.schema_registry_url] nor [avro.schema] is configured"
+            );
+        }
```
philiplhchan: suggest moving this before the wrapperSchema parsing

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+            try {
+                innerSchema = new Schema.Parser().parse(inlineSchema);
+                logger.debug("AvroPayloadDecoder: parsed inline schema [{}]", innerSchema.getFullName());
+            } catch (Throwable t) {
+                logger.error("AvroPayloadDecoder: inline schema parse FAILED: {}", t.getMessage(), t);
+                throw new IllegalArgumentException("Failed to parse avro.schema: " + t.getMessage(), t);
+            }
+        }
+
+        if (wrapperSchemaJson != null) {
```
philiplhchan: IMO it's cleaner to move all the schema parsing to a static method in this class

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+        if (registryUrl != null) {
+            Map<String, String> headers = resolveHeaders(params);
+            int connectTimeoutMs = params.containsKey(PARAM_SCHEMA_REGISTRY_CONNECT_TIMEOUT_MS)
+                ? Integer.parseInt(String.valueOf(params.get(PARAM_SCHEMA_REGISTRY_CONNECT_TIMEOUT_MS)))
+                : DEFAULT_REGISTRY_TIMEOUT_MS;
+            int requestTimeoutMs = params.containsKey(PARAM_SCHEMA_REGISTRY_REQUEST_TIMEOUT_MS)
+                ? Integer.parseInt(String.valueOf(params.get(PARAM_SCHEMA_REGISTRY_REQUEST_TIMEOUT_MS)))
+                : DEFAULT_REGISTRY_TIMEOUT_MS;
+            logger.info("AvroPayloadDecoder: fetching schema from [{}] headers={}", registryUrl, headers.keySet());
+            innerSchema = fetchSchema(registryUrl, headers, connectTimeoutMs, requestTimeoutMs);
```
philiplhchan: it's generally not good practice to have a constructor do something that can fail, doing http request even more so.
i would suggest moving this out of the constructor and making it an explicit action that the caller can decide what to do if there are failures.
this is an example of froh fixing a similar pattern: https://github.com/opensearch-project/OpenSearch/pull/21974

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+            throw new IllegalArgumentException(
+                "Avro payload too short after skipping " + skipBytes + " bytes (" + remaining + " remaining)"
+            );
+        }
+
+        if (logger.isDebugEnabled()) {
+            int previewLen = Math.min(remaining, 16);
+            StringBuilder hex = new StringBuilder();
+            for (int i = 0; i < previewLen; i++) {
+                hex.append(String.format("%02x ", raw[skipBytes + i]));
```
philiplhchan: nit: just start the loop for `skipBytes`?

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+            return null;
+        }
+        int remaining = raw.length - skipBytes;
+        if (remaining < 1) {
+            throw new IllegalArgumentException(
+                "Avro payload too short after skipping " + skipBytes + " bytes (" + remaining + " remaining)"
+            );
+        }
+
+        if (logger.isDebugEnabled()) {
```
philiplhchan: little bit cleaner to use a Runnable like
```
logger.debug(() -> {
  // ...
  return new ParameterizedMessage("failed to delete file [{}]", origFile), ex);
});
```

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+            }
+
+            // TODO: the decoded record is serialized to JSON bytes here and then parsed back to a Map
+            // in MessageProcessorRunnable. A future optimization could pass the Map directly through
+            // the ingestion pipeline to avoid the serialize/parse round-trip.
+            byte[] jsonBytes = toJsonBytes(map);
+            logger.debug("Avro decode complete: json_bytes={}", jsonBytes.length);
+            return jsonBytes;
+        } catch (IOException e) {
+            logger.error("Failed to decode Avro message: schema=[{}] payload_length={}", schema.getFullName(), raw.length, e);
```
philiplhchan: don't recommend logging and throwing exception, it will usually result in double logging. make this a debug if required

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+                skipBytes,
+                remaining,
+                hex.toString().trim()
+            );
+        }
+
+        try {
+            BinaryDecoder decoder = DecoderFactory.get().binaryDecoder(new ByteArrayInputStream(raw, skipBytes, remaining), null);
+            GenericRecord record = reader.read(null, decoder);
+            logger.debug("Decoded Avro record: schema=[{}]", record.getSchema().getFullName());
```
philiplhchan: you probably don't want to do `record.getSchema().getFullName()` in the kafka hot loop.
use the runnable method suggested above to skip it

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+                "Decoding Avro payload: total={} skip={} remaining={} first_bytes=[{}]",
+                raw.length,
+                skipBytes,
+                remaining,
+                hex.toString().trim()
+            );
+        }
+
+        try {
+            BinaryDecoder decoder = DecoderFactory.get().binaryDecoder(new ByteArrayInputStream(raw, skipBytes, remaining), null);
```
philiplhchan: would you not want to reuse the decoder for better efficiency? 

also it seems like you can just make the factory a static field

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+            GenericRecord record = reader.read(null, decoder);
```
philiplhchan: same question here on opportunity for re-use

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    static String extractSchemaJson(String body) {
+        try {
+            // Validate it's parseable Avro JSON — if so, use body directly
+            new Schema.Parser().parse(body);
+            logger.debug("Registry response is a valid Avro schema JSON");
+            return body;
+        } catch (Throwable e) {
+            if (!(e instanceof SchemaParseException)) {
```
philiplhchan: why not just catch `SchemaParseException` instead?

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+            throw new IllegalStateException("Interrupted while fetching Avro schema from [" + url + "]", e);
+        } catch (IOException e) {
+            throw new IllegalStateException("Failed to fetch Avro schema from [" + url + "]: " + e.getMessage(), e);
+        }
+    }
+
+    static String extractSchemaJson(String body) {
+        try {
+            // Validate it's parseable Avro JSON — if so, use body directly
+            new Schema.Parser().parse(body);
```
philiplhchan: double parsing here seems unnecessary, why not just actually parse it in this method for both paths

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+                }
+            } else if (schemaVal != null) {
+                inlineSchema = String.valueOf(schemaVal);
+            }
+        }
+
+        String wrapperSchemaJson = params.containsKey(PARAM_WRAPPER_SCHEMA)
+            ? String.valueOf(params.get(PARAM_WRAPPER_SCHEMA))
+            : null;
+        String wrapperField = params.containsKey(PARAM_WRAPPER_FIELD) ? String.valueOf(params.get(PARAM_WRAPPER_FIELD)) : null;
```
philiplhchan: seems a bit weird that wrapper schema must be defined inline, but "inner schema" can be inline or from remote registry

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+        for (Schema.Field field : record.getSchema().getFields()) {
+            map.put(field.name(), convertAvroValue(record.get(field.name())));
+        }
+        return map;
+    }
+
+    @SuppressWarnings("unchecked")
+    private static Object convertAvroValue(Object value) {
+        if (value == null) return null;
+        if (value instanceof GenericRecord) return recordToMap((GenericRecord) value);
```
philiplhchan: for all the `instanceOf`, they can be cast inline 
```
if (value instanceof GenericRecord recordVal) {
  return recordToMap(recordVal);
}
```
also for consistency, recommend added brackets even if single line

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
```diff
@@ -0,0 +1,411 @@
...
+            logger.debug("Decoded Avro record: schema=[{}]", record.getSchema().getFullName());
+
+            Map<String, Object> map = recordToMap(record);
+
+            if (msgField != null) {
+                Object nested = map.get(msgField);
+                if (nested == null) {
+                    logger.debug("avro.msg_field [{}] is null — tombstone, returning null", msgField);
+                    return null;
+                }
```
philiplhchan: if msgField != null, then `recordToMap` doesn't need to parse all the other fields that are going to get thrown away

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
philiplhchan: have we done any benchmarks on this avro decoder?
my gut feeling/concern is that this is going to be so slow it won't actually be usable in any real world application

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPartitionConsumer.java
Tarun-kishore: Updated to move the decoding in `KafkaMessage.getPayload()`. and the logging for failed message would have the same behaviour as other failures in pull based ingestion. As for thread visibility, GenericDatumReader is not thread-safe, so I've removed the cached reader field from AvroPayloadDecoder and it's created one per decode() call instead.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: Updated to move this call to a static `.create()` method.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: Will update wrapper schema to have support to fetch from remote registry too.

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/KafkaPayloadDecoder.java
Tarun-kishore: Moved a decoder to poller. So now if we want to extend similar approach to kinesis, PayloadDecoder can be extended

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: Updated to use static field

plugins/ingestion-kafka/src/main/java/org/opensearch/plugin/kafka/AvroPayloadDecoder.java
Tarun-kishore: Updated to parse only msg_field if it's non-null.

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
```diff
@@ -1313,6 +1313,27 @@ static void validateIngestionSourceSettings(Settings settings, ClusterState stat
             // (multi-partition consumer factory). The check will be wired here once available.
         }
 
+        // Decoder type validation. Any non-default decoder_type (i.e. anything other than "xcontent")
+        // requires all nodes to support the payload-decoder registry introduced in V_3_8_0. Older nodes
+        // do not recognise the setting and cannot load the index or create the decoder.
+        if (IndexMetadata.INGESTION_SOURCE_DECODER_TYPE_SETTING.exists(settings)) {
+            String decoderType = IndexMetadata.INGESTION_SOURCE_DECODER_TYPE_SETTING.get(settings);
+            if ("xcontent".equals(decoderType) == false) {
```
varunbharadwaj: Should we prevent decoder_type and decoder_settings to be added altogether instead of only allowing `xcontent` type for mixed version scenario?

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
```diff
@@ -159,6 +167,11 @@ private void initializeStreamPoller(
             .pipelineExecutor(pipelineExecutor)
             .warmupConfig(ingestionSource.getWarmupConfig())
             .indexMetadata(indexMetadata)
+            .payloadDecoder(payloadDecoderFactory.create(
```
varunbharadwaj: Do we need to close the payloadDecoder? We need to close it when poller initialization fails by moving the payloadDecoderFactory.create outside the builder.build()?

server/src/main/java/org/opensearch/index/IngestionPayloadDecoder.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.io.Closeable;
+import java.util.Map;
+
+/**
+ * Converts a raw ingestion message payload into a field map that the message mapper consumes.
+ *
+ * <p>A decoder is created once per shard poller by its {@link IngestionPayloadDecoderFactory}
+ * and is called exclusively from the poller's single consumer thread. Implementations are
+ * therefore NOT required to be thread-safe; they may freely use thread-local or instance-level
```
varunbharadwaj: >Implementations are therefore NOT required to be thread-safe

We can have multiple poller instances per shard (this is ongoing work). The payloadDecoder can be local to a poller, but they should ensure there is no cross-poller leaks when using shared state. We can highlight this point.

server/src/main/java/org/opensearch/index/IngestionPayloadDecoderFactory.java
```diff
@@ -0,0 +1,61 @@
...
+     * setting that is unrecognized, malformed, or required-but-absent.
+     *
+     * @param settings the {@code index.ingestion_source.decoder_settings.*} map
+     */
+    default void validate(Map<String, Object> settings) {}
+
+    /**
+     * Creates a new decoder for the given shard.
+     *
+     * <p>Called once per shard when the stream poller initializes. The returned decoder is used
```
varunbharadwaj: We can rephrase it as - called once per poller initialization, as we are also working on concurrent poller support.

server/src/main/java/org/opensearch/index/IngestionPayloadDecoderFactory.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Validates {@code settings} at index-creation time.
+     *
+     * <p>Called before the index is committed so users receive immediate feedback on
+     * invalid or missing configuration. Throw {@link IllegalArgumentException} for any
+     * setting that is unrecognized, malformed, or required-but-absent.
+     *
+     * @param settings the {@code index.ingestion_source.decoder_settings.*} map
+     */
+    default void validate(Map<String, Object> settings) {}
```
varunbharadwaj, line 37: Where will this be called?

server/src/main/java/org/opensearch/indices/pollingingest/mappers/RawPayloadIngestionMessageMapper.java
```diff
@@ -37,19 +36,11 @@ public RawPayloadIngestionMessageMapper(int shardId) {
...
+        throws IllegalArgumentException {
         String id = shardId + "-" + pointer.asString();
-
-        // No auto-generated ID timestamp since we're using the pointer as ID
         long autoGeneratedIdTimestamp = UNSET_AUTO_GENERATED_TIMESTAMP;
-
-        // Construct the full payload map with metadata fields and source
-        Map<String, Object> payloadMap = Map.of(ID, id, OP_TYPE, OP_TYPE_INDEX, SOURCE, sourceMap);
-
+        Map<String, Object> payloadMap = Map.of(ID, id, OP_TYPE, OP_TYPE_INDEX, SOURCE, decodedPayload);
```
varunbharadwaj, line 46: Shall we retain the comments for clarity?

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -389,6 +404,11 @@ public void close() {
         }
         consumerThread.shutdown();
         blockingQueueContainer.close();
+        try {
+            payloadDecoder.close();
```
varunbharadwaj, line 415: This should be called irrespective of whether poller is started/polling or not, right? Should we move this to the top before `started` check?

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
```diff
@@ -315,8 +322,16 @@ private IngestionShardPointer processRecords(
             try {
                 totalPolledCount.inc();
 
+                // Decode payload bytes to a field map before the mapper sees the message.
+                // Errors here are caught below and handled by the configured DROP/BLOCK strategy.
+                @SuppressWarnings("unchecked")
+                BytesArray payload = new BytesArray((byte[]) result.getMessage().getPayload());
```
varunbharadwaj: Let us discuss if there is a better way to handle Message\<?\> which is the current contract, maybe by accepting Message\<?\> in the decoder.

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -955,6 +957,13 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
             pluginsService.filterPlugins(IngestionConsumerPlugin.class)
                 .forEach(plugin -> ingestionConsumerFactories.putAll(plugin.getIngestionConsumerFactories()));
 
+            // build ingestion payload decoder registry
+            final IngestionPayloadDecoderRegistry.Builder registryBuilder = IngestionPayloadDecoderRegistry.builder()
+                .register("xcontent", XContentIngestionPayloadDecoder.Factory.INSTANCE);
+            pluginsService.filterPlugins(IngestionConsumerPlugin.class)
+                .forEach(plugin -> plugin.getIngestionPayloadDecoderFactories().forEach(registryBuilder::register));
+            final IngestionPayloadDecoderRegistry payloadDecoderRegistry = registryBuilder.build();
```
varunbharadwaj, line 965: Do we need to add them to `resourcesToClose` if we want to close the resources?

varunbharadwaj:
Looks like the gradle check failed. We can run `./gradlew spotlessApply` to fix the formatting.

server/src/main/java/org/opensearch/cluster/metadata/MetadataCreateIndexService.java
Tarun-kishore: Updated to prevent decoder configs at all

server/src/main/java/org/opensearch/index/IngestionPayloadDecoder.java
Tarun-kishore: Updated javadoc to clarify each poller should have its own decoder

server/src/main/java/org/opensearch/index/IngestionPayloadDecoderFactory.java
Tarun-kishore: updated

server/src/main/java/org/opensearch/indices/pollingingest/mappers/RawPayloadIngestionMessageMapper.java
Tarun-kishore: Done

server/src/main/java/org/opensearch/index/IngestionPayloadDecoderFactory.java
Tarun-kishore: It will be called in IndicesService after decoderFactory is created.

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
Tarun-kishore: Decoder now accepts Message<?> as per current contract

server/src/main/java/org/opensearch/indices/pollingingest/DefaultStreamPoller.java
Tarun-kishore: Yes, moved

server/src/main/java/org/opensearch/node/Node.java
Tarun-kishore: Yes, updated so that factory resources are closed.

server/src/main/java/org/opensearch/index/engine/IngestionEngine.java
Tarun-kishore: moved create method outside

server/src/main/java/org/opensearch/indices/pollingingest/XContentIngestionPayloadDecoder.java
```diff
@@ -0,0 +1,65 @@
...
+import org.opensearch.index.IngestionPayloadDecodingException;
+import org.opensearch.index.Message;
+
+import java.util.Map;
+
+/**
+ * Built-in decoder that parses JSON (or any XContent format) payloads.
+ *
+ * <p>This is the default decoder used when {@code index.ingestion_source.decoder_type} is
+ * {@code xcontent} or is not set. It preserves the existing ingestion behavior.
```
varunbharadwaj: >It preserves the existing ingestion behavior.

Let us remove this line

server/src/main/java/org/opensearch/index/IngestionPayloadDecoder.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.Closeable;
+import java.util.Map;
+
+/**
+ * Converts a raw ingestion message into a field map that the message mapper consumes.
+ *
+ * <p>A decoder is created once per poller initialization by its {@link IngestionPayloadDecoderFactory}.
+ * Multiple poller instances may exist per shard (e.g. for concurrent ingestion support), so each
+ * poller owns its own decoder instance. Implementations must not share mutable state across poller
```
varunbharadwaj, line 21: >so each poller owns its own decoder instance

We are using singleton decoder instance today so this statement is not correct. We need to update the singleton logic.

server/src/main/java/org/opensearch/index/IngestionPayloadDecoder.java
```diff
@@ -0,0 +1,53 @@
...
+     * current ingestion consumers (Kafka, Kinesis, etc.) produce byte-array payloads.
+     *
+     * @param message the raw ingestion message
+     * @return a fresh, mutable, XContent-compatible field map
+     * @throws IngestionPayloadDecodingException if the payload cannot be decoded
+     */
+    Map<String, Object> decode(Message<?> message);
+
+    /**
+     * Releases resources held by this decoder. Called once when the shard poller shuts down,
```
varunbharadwaj: Close can be called more than once currently, for example if reset fails. This might need another fix/change, but we can update the comment to remove "Called once".

server/src/main/java/org/opensearch/indices/pollingingest/XContentIngestionPayloadDecoder.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public class XContentIngestionPayloadDecoder implements IngestionPayloadDecoder {
+
+    public static final XContentIngestionPayloadDecoder INSTANCE = new XContentIngestionPayloadDecoder();
```
varunbharadwaj: In the comments above (registry and other classes), we mention each poller must own the decoder. Using singleton here may not match that statement. Let us discuss about this.

server/src/main/java/org/opensearch/indices/pollingingest/IngestionPayloadDecoderRegistry.java
```diff
@@ -0,0 +1,86 @@
...
+        IOUtils.close(factories.values());
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    /**
+     * Builds the registry. Core decoders should be registered before plugin decoders.
+     * Duplicate names cause an {@link IllegalStateException} at build time.
```
varunbharadwaj: >Duplicate names cause an {@link IllegalStateException} at build time

We can remove the `build time`

server/src/main/java/org/opensearch/indices/pollingingest/XContentIngestionPayloadDecoder.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public Map<String, Object> decode(Message<?> message) {
+        try {
+            BytesReference payload = new BytesArray((byte[]) message.getPayload());
```
varunbharadwaj: I think `IngestionUtils.getParsedPayloadMap()` can be cleaned up and tests updated to use this instead for validation.

server/src/main/java/org/opensearch/indices/pollingingest/XContentIngestionPayloadDecoder.java
Tarun-kishore: Yes, updated singleton implementation

varunbharadwaj:
Thanks for the changes @Tarun-kishore. We can add some basic tests like duplicate decoder registration throwing error and decoder settings rejection. Let us also add a simple integration test in IngestFromKafkaIT.java explicitly passing the new settings.

server/src/main/java/org/opensearch/node/Node.java
```diff
@@ -956,6 +957,16 @@ protected Node(final Environment initialEnvironment, Collection<PluginInfo> clas
             pluginsService.filterPlugins(IngestionConsumerPlugin.class)
                 .forEach(plugin -> ingestionConsumerFactories.putAll(plugin.getIngestionConsumerFactories()));
 
+            // build ingestion payload decoder registry
+            final IngestionPayloadDecoderRegistry.Builder registryBuilder = IngestionPayloadDecoderRegistry.builder()
+                .register("xcontent", XContentIngestionPayloadDecoder.Factory.INSTANCE);
```
varunbharadwaj: We can define the key as a constant in XContentIngestionPayloadDecoder

varunbharadwaj (approved):

PR #22370, opened 2026-07-01
fix(tiering): Track force merges in activeMerges to prevent tiering before force merge completion

KhishorekumarBS:
## Description

MergeScheduler.forceMerge() bypasses the activeMerges counter, making in-flight force merges invisible to onMergesDrained(). When AutoForceMergeManager triggers a force merge and tiering starts before it completes, tiering proceeds without waiting — causing the merged segment to never reach remote store and leaving the
replica permanently diverged. 

## Changes

- **MergeScheduler.java**: Increment/decrement activeMerges in forceMerge() + fire drain listeners on completion
- **IndexShard.java**: Add waitForReplicaSync(TimeValue) as belt-and-suspenders check
- **TransportPrepareTieringAction.java**: Call waitForReplicaSync after waitForRemoteStoreSync
- **Tests**: Verify force merges block drain, activeMerges includes force merges, waitForReplicaSync timeout propagates

## Testing

- Unit tests: MergeSchedulerOnDrainedTests, TransportPrepareTieringActionTests, IndexShardTests
- Integration test: Added

### Description

### Related Issues

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
+++ b/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
@@ -169,20 +169,7 @@ public class MergeScheduler {
                 runMerge(oneMerge);
             }
         } finally {
-            activeMerges.decrementAndGet();
-            // Fire all drain listeners if all merges completed and none pending.
-            // This ensures tiering's onMergesDrained callback fires after a force merge completes.
-            if (isFrozen() && activeMerges.get() == 0 && !mergeHandler.hasPendingMerges() && !onDrainedListeners.isEmpty()) {
-                List<Runnable> listeners = List.copyOf(onDrainedListeners);
-                onDrainedListeners.clear();
-                for (Runnable listener : listeners) {
-                    try {
-                        listener.run();
-                    } catch (Exception ex) {
-                        logger.warn("Exception in onDrained listener", ex);
-                    }
-                }
-            }
+            decrementAndFireDrainListeners();
             forceMergeLock.release();
         }
     }
@@ -357,19 +344,7 @@ public class MergeScheduler {
                 // runMerge already invoked onMergeFailureCleanup; swallow to prevent
                 // uncaught exception on the merge thread pool.
             } finally {
-                activeMerges.decrementAndGet();
-                // Fire all drain listeners if all merges completed and none pending
-                if (isFrozen() && activeMerges.get() == 0 && !mergeHandler.hasPendingMerges() && !onDrainedListeners.isEmpty()) {
-                    List<Runnable> listeners = List.copyOf(onDrainedListeners);
-                    onDrainedListeners.clear();
-                    for (Runnable listener : listeners) {
-                        try {
-                            listener.run();
-                        } catch (Exception ex) {
-                            logger.warn("Exception in onDrained listener", ex);
-                        }
-                    }
-                }
+                decrementAndFireDrainListeners();
                 // A completed merge may free up capacity for new merges, so check again.
                 executeMerge();
             }
@@ -407,6 +382,26 @@ public class MergeScheduler {
             throw e instanceof IOException ? (IOException) e : new IOException(e);
         } finally {
             mergeStatsTracker.afterMerge(tookMS, totalNumDocs, totalSizeInBytes);
+        }
+    }
+
+    /**
+     * Decrements the active merge count and fires all registered drain listeners if the scheduler
+     * is frozen and no merges (active or pending) remain. Called from both the background merge
+     * ({@link #submitMergeTask}) and force merge ({@link #forceMerge}) completion paths.
+     */
+    private void decrementAndFireDrainListeners() {
+        activeMerges.decrementAndGet();
+        if (isFrozen() && activeMerges.get() == 0 && !mergeHandler.hasPendingMerges() && !onDrainedListeners.isEmpty()) {
+            List<Runnable> listeners = List.copyOf(onDrainedListeners);
+            onDrainedListeners.clear();
+            for (Runnable listener : listeners) {
+                try {
+                    listener.run();
+                } catch (Exception ex) {
+                    logger.warn("Exception in onDrained listener", ex);
+                }
+            }
         }
     }
 }
--- a/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
+++ b/server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
@@ -154,6 +154,7 @@ public class MergeScheduler {
         assert Thread.currentThread().getName().contains(ThreadPool.Names.FORCE_MERGE)
             : "forceMerge must be called on FORCE_MERGE thread but was: " + Thread.currentThread().getName();
         forceMergeLock.acquireUninterruptibly();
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -3843,6 +3843,60 @@ public Set<SegmentReplicationShardStats> getReplicationStatsForTrackedReplicas()
...
+                behindReplicas,
+                maxBehind
+            );
+            try {
+                Thread.sleep(500);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new OpenSearchException("Interrupted waiting for replica sync on shard [" + shardId + "]", e);
+            }
+        }
```
Bukhtawar, line 3899: Sleeping is not a good practice as we burn unnecessary CPU cycles, maybe use `LatchedActionListener` instead?

KhishorekumarBS: We're using a similar pattern as `waitForRemoteStoreSync`. If we go with `LatchedActionListener` , we'll need to call `getSegmentReplicationStats` inside `updateVisibleCheckpointForShard` and both are synchronized methods. Though deadlock may not happen, trying to avoid this. Resolving these might include lot of refactoring and can increase blast radius. As discussed offline, adding a TODO to remove the sleep and move to listeners.

alchemist51:
```
Thread.sleep(500);
```

How did we arrive on this number?

KhishorekumarBS: Our total timeout is 30 seconds. So, 500 ms polling looks reasonable for replicas to sync right

server/src/main/java/org/opensearch/index/engine/dataformat/merge/MergeScheduler.java
```diff
@@ -168,6 +169,20 @@ public void forceMerge(int maxNumSegment) throws IOException {
                 runMerge(oneMerge);
             }
         } finally {
+            activeMerges.decrementAndGet();
```
chaitanya588: This is completely a duplicate of what we have in the existing merge flow. Can we pull this out to a private method?

server/src/main/java/org/opensearch/index/shard/IndexShard.java
chaitanya588: +1 to @alchemist51  comment. In the waitForRemoteStoreSync, we have 1 hour as a total time out (indices.recovery.internal_remote_upload_timeout) and we are keeping 30 seconds sleep in the loop. As our sync can also take that much time (because the last diff segment the shard is going to see is a merged segment which is of GBs in size as well. Keeping 30 seconds timeout is going to be too quick).

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -3843,6 +3845,77 @@ public Set<SegmentReplicationShardStats> getReplicationStatsForTrackedReplicas()
...
+                        s -> s.getCheckpointsBehindCount() == 0 && s.getBytesBehindCount() == 0 && s.getCurrentReplicationTimeMillis() == 0
+                    )) {
+                logger.info("All replicas in sync for shard [{}]", shardId);
+                return;
+            }
+            long behindReplicas = stats.stream().filter(s -> s.getCheckpointsBehindCount() > 0 || s.getBytesBehindCount() > 0).count();
+            long maxCheckpointsBehind = stats.stream().mapToLong(SegmentReplicationShardStats::getCheckpointsBehindCount).max().orElse(0);
+            long maxBytesBehind = stats.stream().mapToLong(SegmentReplicationShardStats::getBytesBehindCount).max().orElse(0);
+            logger.info(
+                "Waiting for replica sync on shard [{}]: {} replica(s) still behind, max checkpoints behind: {}, max bytes behind: {}",
```
chaitanya588, line 3887: Keep this logging at debug level only.

server/src/main/java/org/opensearch/index/shard/IndexShard.java
KhishorekumarBS: waitForRemoteStoreSync - the primary usecase to introduce the method is for recovery flows where 1 hour timeout is ok. The method is reused in our tiering flow but here, we have a transport timeout of 90 seconds and all of these sync and tiering should be complete before that. if waitForRemoteStoreSync doesn't complete in that 90 seconds, it anyways results in transport exception.

 waitForReplicaSync - the primary usecase to introduce the method is tiering. I agree that in worst case, it can also take 1 hour but for tiering, we anyway can't wait that much time. Hence went with small value.

chaitanya588 (approved):

server/src/main/java/org/opensearch/storage/action/tiering/TransportPrepareTieringAction.java
```diff
@@ -59,6 +59,7 @@ public class TransportPrepareTieringAction extends TransportBroadcastByNodeActio
 
     private static final Logger logger = LogManager.getLogger(TransportPrepareTieringAction.class);
     private static final TimeValue PERMITS_ACQUIRE_TIMEOUT = TimeValue.timeValueSeconds(30);
+    private static final TimeValue REPLICA_SYNC_TIMEOUT = TimeValue.timeValueSeconds(30);
```
Bukhtawar: Lets make this dynamic

server/src/main/java/org/opensearch/index/shard/IndexShard.java
```diff
@@ -3843,6 +3845,77 @@ public Set<SegmentReplicationShardStats> getReplicationStatsForTrackedReplicas()
...
+     * or the timeout is exceeded.
+     * <p>
+     * Used during tiering preparation to verify replicas are in sync after
+     * {@code waitForRemoteStoreSync()} — ensures replicas have downloaded the latest
+     * segments before shard relocation begins.
+     *
+     * @param timeout maximum time to wait for replicas to sync
+     * @throws IOException if replicas fail to sync within the timeout
+     */
+    public void waitForReplicaSync(TimeValue timeout) throws IOException {
```
Bukhtawar, line 3867: What happens if we don't wait long enough, wouldn't after tiering to warm replica catches up?

server/src/main/java/org/opensearch/storage/action/tiering/TransportPrepareTieringAction.java
KhishorekumarBS: Added in new revision

server/src/main/java/org/opensearch/index/shard/IndexShard.java
KhishorekumarBS: As no refresh on readonly engine, replica will not know that it's out of sync. This check before tiering will make sure we are all caught up.

Bukhtawar: Waiting for replica sync is wasteful espl if we are tiering, as the full files would anyways not be needed, we only care about metadata to be in-sync. I would have liked for the replica to re-sync once tiering completes.

Bukhtawar (approved):
Thanks, lets see how we can avoid blocking on replica sync, add a TODO

PR #22371, opened 2026-07-01
Fix ClassCastException on malformed mappings in create index

vishdivs:
### Description

  Added type validation for the `mappings` field in `CreateIndexRequest#source()`. Previously, passing a non-object value (e.g., a string like `"{}"`) for the `mappings` key resulted in a 500 Internal Server
  Error with a `ClassCastException`. This change adds explicit type checks that return a proper 400 Bad Request with a clear error message.

  ### Related Issues

  ### Check List

  ### Testing

  **Before (500 ClassCastException):**
  curl -X PUT "localhost:9200/vishdivs_doc"
      -H "Content-Type: application/json"
      -d '{
        "mappings": "{}"
    }'
  {"error":{"root_cause":[{"type":"class_cast_exception","reason":"class java.lang.String cannot be cast to class java.util.Map (java.lang.String and java.util.Map are in module java.base of loader
  'bootstrap')"}],"type":"class_cast_exception","reason":"class java.lang.String cannot be cast to class java.util.Map (java.lang.String and java.util.Map are in module java.base of loader
  'bootstrap')"},"status":500}

  **After (400 with clear error message):**
  curl -X PUT "localhost:9200/vishdivs_doc"
      -H "Content-Type: application/json"
      -d '{
        "mappings": "{}"
    }'
  {"error":{"root_cause":[{"type":"parse_exception","reason":"key [mapping] must be an object"}],"type":"parse_exception","reason":"key [mapping] must be an object"},"status":400}

  For more information on following Developer Certificate of Origin and signing off your commits, please check
  [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

  ---


server/src/main/java/org/opensearch/action/admin/indices/create/CreateIndexRequest.java
```diff
@@ -527,8 +527,14 @@ public CreateIndexRequest source(Map<String, ?> source, DeprecationHandler depre
...
                 settings((Map<String, Object>) entry.getValue());
             } else if (MAPPINGS.match(name, deprecationHandler)) {
+                if (entry.getValue() instanceof Map == false) {
+                    throw new OpenSearchParseException("key [mappings] must be an object");
+                }
                 Map<String, Object> mappings = (Map<String, Object>) entry.getValue();
                 for (Map.Entry<String, Object> entry1 : mappings.entrySet()) {
+                    if (entry1.getValue() instanceof Map == false) {
+                        throw new OpenSearchParseException("values inside key [mappings] must be an object");
+                    }
```
mgodwan: If rest action adds a `_doc` key here which ensures this is always true, do we need this additional check?

vishdivs: Thanks for the review @mgodwan. You're right, I checked this and confirmed the inner check is not needed.
The `prepareMappings()` method always wraps the mappings content under _doc before it reaches source(Map, handler): newSource.put("mappings", singletonMap(MapperService.SINGLE_MAPPING_NAME, mappings));
So by the time the loop iterates, entry1 is always _doc -> Map<>, which means entry1.getValue() will always be a Map instance.

I tested this with,
```
  curl -X PUT "localhost:9200/test_nested_scalar" \
    -d '{"mappings": {"properties": "invalid"}}'
```

  Without the inner check, this still returns a proper 400:
```
{"error":{"root_cause":[{"type":"parse_exception","reason":"properties must be a map type"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: properties must be a map type","caused_by":{"type":"parse_exception","reason":"properties must be a map type"}},"status":400}
```
The mapper parsing layer already handles invalid nested values gracefully. I'll remove the inner check and keep only the outer validation.

mgodwan (approved):

PR #22378, opened 2026-07-02
[Analytics Engine] Error on BIGINT arithmetic overflow instead of wrapping

ahkcs:
### Description

BIGINT (`long` / Int64) arithmetic (`+`, `-`, `*`) on the analytics-engine DataFusion backend silently **wraps** on 64-bit overflow instead of erroring. For example `eval x = long_col * long_col` where the product exceeds `9,223,372,036,854,775,807` returns a wrapped negative value with HTTP 200, rather than a clear client error. DataFusion lowers integer `Plus`/`Minus`/`Multiply` to arrow's *wrapping* kernels (`add_wrapping`/`sub_wrapping`/`mul_wrapping`) by default.

This makes BIGINT overflow an explicit **HTTP 400** on the DataFusion route.

#### Mechanism

A logical-plan rewrite (`checked_arith_rewrite`) runs between `from_substrait_plan` and `create_physical_plan` on every executor. It walks the plan with `LogicalPlan::map_expressions` + `Expr::transform_up` and rewrites every `BinaryExpr { op: Plus | Minus | Multiply }` **whose both operands resolve to `Int64`** into a call to an overflow-checked scalar UDF (`checked_add_i64` / `checked_sub_i64` / `checked_mul_i64`). The UDFs invoke arrow's *checked* numeric kernels (`add`/`sub`/`mul`), which return an error on overflow. The error carries a stable, self-authored phrase (`BIGINT arithmetic overflow`) that `NativeErrorConverter` maps to an `IllegalArgumentException` → HTTP 400.

Why a logical rewrite (rather than flipping DataFusion's physical `BinaryExpr::fail_on_overflow`): walking the *logical* plan reaches arithmetic in **every** node uniformly — projection exprs, filter predicates, aggregate group/argument exprs (`sum(a*b)`), window args, sort keys, join conditions — because those are all first-class top-level `Expr`s at the logical level. The physical-plan approach would depend on per-node expression accessors that do not uniformly surface aggregate/window argument expressions, leaving silent overflow gaps.

#### Type gate (what is and isn't affected)

Only `Int64 op Int64` is rewritten. Everything else passes through unchanged:

| operands | behavior |
|---|---|
| `Int64` `{+,-,*}` `Int64` | **checked → 400 on overflow** |
| `Float32`/`Float64` | unchanged (wrap to ±Infinity, PPL/Calcite parity) |
| `Decimal*` | unchanged (own overflow path) |
| narrower ints (`Int8/16/32`), `UInt64` | unchanged |
| `/`, `%` | unchanged (not `+/-/*`) |

The analytics scan boundary already narrows the only `UInt64` source (`unsigned_long`) to `Int64` before planning (`schema_coerce`), and the SQL-plugin widening (opensearch-project/sql#5603) widens all sub-BIGINT integer arithmetic to `Int64` **before** the engine fork — so `Int64 op Int64` is the only integer arithmetic that reaches DataFusion and can still overflow. The gate is therefore both correct and sufficient.

Type resolution failures (e.g. a correlated column absent from the merged input schema) fall through to "leave unchanged" rather than propagate, so the rewrite can never turn a previously-working query into a planning error.

The UDFs are `Volatility::Immutable`, and their `return_type` is `Int64` — identical to the `BinaryExpr` they replace — so plan schemas are byte-identical (no `recompute_schema`, no `relabel_exec` churn), and predicate pushdown / CSE still apply. Const-folding an *overflowing* literal pair does not lose the error: DataFusion's `ConstEvaluator` preserves the original expr on a UDF runtime error (it only fails-fast for `CAST`), so overflow still surfaces at execution with the stable message; a non-overflowing constant simply folds to its exact value. Each rewritten expression's output name is preserved (`NamePreserver`) so a later `recompute_schema` can't rename an aggregate measure like `sum(a * b)` and break a parent node that references it by the derived name.

#### Distributed / aggregate path (transport)

Flight transport does not serialize the Java exception type — only a `StreamErrorCode` + message survive. A client-error `IllegalArgumentException` with no HTTP status would cross as `INTERNAL` and be redacted to a generic 500 on the coordinator (this is what `stats sum(a*b)` overflow originally did). `AnalyticsTransportErrors` now tags the overflow as `StreamErrorCode.INVALID_ARGUMENT` on the data-node send side and rebuilds it as a `BAD_REQUEST` (400) `OpenSearchStatusException` on the coordinator receive side — mirroring the existing breaker→429 and cancel→CANCELLED mappings — so the 400 survives the distributed reduce.

#### Indexed path

On the data-node indexed path the rewrite is applied to both the executed plan **and** the plan used for filter extraction, so arithmetic inside a `WHERE` predicate that seeds the parquet pushdown predicate is checked too (otherwise `where long1 * long2 > k` would be scan-filtered with wrapping arithmetic and drop overflowing rows silently). Checked arithmetic only ever appears nested inside a comparison operand — never as a top-level `AND`/`OR` or delegation marker — so predicate delegation classification is unaffected.

### Scope

This is the DataFusion-backend half of the fix for the residual `long × long` overflow that operand-widening cannot address (there is no wider primitive than `long`). The Calcite/v2 half — rewriting `long` `+`/`-`/`*` to `Math.*Exact` and surfacing a 4xx — landed in opensearch-project/sql#5604 (merged), so both engines now **error consistently** on `long` overflow rather than one wrapping and one erroring. This PR closes the gap on the analytics route.

Resolves the reported analytics-engine `long`-arithmetic overflow (see opensearch-project/sql#5164).

### Testing

| test | before | after |
|---|---|---|
| `checked_arith` Rust UDF unit tests (overflow errors, null-propagation, exact values, batch-fail) | — | 7/7 pass |
| `checked_arith_rewrite` Rust unit tests (Int64 `+/-/*` rewritten; float/mixed/divide untouched; nested bottom-up; Filter + Sort node coverage) | — | 8/8 pass |
| full `analytics-backend-datafusion` Rust lib suite (regression) | 1284 pass | 1284 pass |
| `NativeErrorConverterTests` (raw + controlled `BIGINT arithmetic overflow` → `IllegalArgumentException`) | — | 19/19 pass |
| `AnalyticsTransportErrorsTests` (BIGINT overflow → `INVALID_ARGUMENT` → 400 round-trip) | — | 20/20 pass |
| `BigintOverflowIT` QA (column `*`, literal+column, `where`-arith, `stats sum(arith)` → 400; non-overflow → 200; double → 200 no-error) | overflow → 200 wrapped | overflow → **400**, non-overflow → 200 |

### Check List
- [x] New functionality has been documented (module docs + PR).
- [x] Commits are signed per the DCO using `--signoff` or `-s`.

### DIFF
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
@@ -149,6 +149,8 @@ async fn build_dataframe(
     let substrait_plan = Plan::decode(plan_bytes)
         .map_err(|e| DataFusionError::Execution(format!("Failed to decode Substrait: {}", e)))?;
     let logical_plan = from_substrait_plan(&ctx.state(), &substrait_plan).await?;
+    // Make BIGINT +,-,* overflow error instead of silently wrapping (see checked_arith_rewrite).
+    let logical_plan = crate::checked_arith_rewrite::rewrite_checked_int64_arith(logical_plan)?;
     ctx.execute_logical_plan(logical_plan).await
 }
 
@@ -251,6 +253,8 @@ pub async fn execute_with_context(
 
         // Union schema widening was applied at table registration (session_context::widen_to_union_schema).
         let logical_plan = from_substrait_plan(&handle.ctx.state(), &substrait_plan).await?;
+        // Make BIGINT +,-,* overflow error instead of silently wrapping (see checked_arith_rewrite).
+        let logical_plan = crate::checked_arith_rewrite::rewrite_checked_int64_arith(logical_plan)?;
         log_debug!(
             "DataFusion logical plan:\n{}",
             logical_plan.display_indent()
--- a/sandbox/plugins/analytics-backend-datafusion/rust/src/udf/checked_arith.rs
+++ b/sandbox/plugins/analytics-backend-datafusion/rust/src/udf/checked_arith.rs
@@ -1,0 +1,246 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+//! Overflow-checked BIGINT arithmetic UDFs: `checked_add_i64`, `checked_sub_i64`,
+//! `checked_mul_i64`.
+//!
+//! DataFusion's default integer arithmetic (`Operator::Plus/Minus/Multiply`) uses arrow's
+//! *wrapping* kernels, so `9223372036854775807 * 2` silently wraps into a negative i64 instead
+//! of erroring. PPL callers expect an error, not a wrapped value. [`crate::checked_arith_rewrite`]
+//! rewrites every `Int64 {+,-,*} Int64` logical `BinaryExpr` into a call to one of these UDFs,
+//! which invoke arrow's *checked* numeric kernels (`add`/`sub`/`mul`) and surface a stable,
+//! self-authored error on overflow.
+//!
+//! The overflow message carries [`OVERFLOW_KEYPHRASE`] verbatim; `NativeErrorConverter` (Java)
+//! matches on that substring and maps it to an `IllegalArgumentException` (HTTP 400).
+
+use std::hash::{Hash, Hasher};
+use std::sync::Arc;
+
+use datafusion::arrow::array::ArrayRef;
+use datafusion::arrow::compute::kernels::numeric::{add, mul, sub};
+use datafusion::arrow::datatypes::DataType;
+use datafusion::common::{exec_datafusion_err, DataFusionError, Result};
+use datafusion::execution::context::SessionContext;
+use datafusion::logical_expr::{
+    ColumnarValue, ScalarFunctionArgs, ScalarUDF, ScalarUDFImpl, Signature, Volatility,
+};
+
+/// Stable Rust→Java contract phrase. `NativeErrorConverter.PATTERNS` matches on this exact
+/// substring to classify the error as a client (400) error. Changing it requires a coordinated
+/// edit in `NativeErrorConverter.java`.
+pub const OVERFLOW_KEYPHRASE: &str = "BIGINT arithmetic overflow";
+
+#[derive(Copy, Clone, Debug)]
+enum Op {
+    Add,
+    Sub,
+    Mul,
+}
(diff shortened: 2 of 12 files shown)


sandbox/plugins/analytics-backend-datafusion/rust/src/query_executor.rs
```diff
@@ -251,6 +253,8 @@ pub async fn execute_with_context(
 
         // Union schema widening was applied at table registration (session_context::widen_to_union_schema).
         let logical_plan = from_substrait_plan(&handle.ctx.state(), &substrait_plan).await?;
+        // Make BIGINT +,-,* overflow error instead of silently wrapping (see checked_arith_rewrite).
+        let logical_plan = crate::checked_arith_rewrite::rewrite_checked_int64_arith(logical_plan)?;
```
LantaoJin, line 257: Could we apply `rewrite_checked_int64_arith` inside `prepare_partial_plan` before creating the physical plan? The partial-aggregate execution path runs the cached plan directly (`SETUP_PARTIAL_AGGREGATE` is present), so multi-shard queries such as `stats sum(long_val * long_val)` currently bypass the new overflow checks and can still return a wrapped result. Please also cover this with a multi-shard test.

sandbox/plugins/analytics-backend-datafusion/rust/src/udf/checked_arith.rs
```diff
@@ -0,0 +1,246 @@
...
+        let lhs: ArrayRef = args.args[0].to_array(n)?;
+        let rhs: ArrayRef = args.args[1].to_array(n)?;
+
+        // arrow's checked kernels null-propagate (null in either operand -> null out) and return
+        // Err(ArithmeticOverflow) only for non-null overflowing pairs. `&dyn Array` is a `Datum`.
+        let result = match self.op {
+            Op::Add => add(&lhs, &rhs),
+            Op::Sub => sub(&lhs, &rhs),
+            Op::Mul => mul(&lhs, &rhs),
+        };
```
LantaoJin, line 132: nit: How about preserve `ColumnarValue::Scalar` operands and pass them to the checked Arrow kernels as scalar `Datum`s, similar to DataFusion's binary `apply` helper? Expanding literals to batch-sized arrays adds an allocation on every batch for common expressions like `long_val * 2`. (The `long_val * 2` allocates and fills an extra Int64 buffer for every batch, whereas DataFusion's binary-expression implementation passes scalar values as Arrow `Datum` scalars without expansion.)

PR #19933, opened 2025-11-08
Adding bulk collect API for filter rewrite sub aggregation collection

jainankitk:
### Description
Filter rewrite optimization for numeric bucket aggregation is able to multi range traverse over BKD, and provide matching documents bitset for each top level bucket. Individually invoking sub collectors for each docId for same bucket is wasteful, and can be done more efficient using bulk collection API. This PR adds such an API to `BucketLeafCollector` and invokes the same from few places

### Related Issues
Related #19324

### Check List
- [x] Functionality includes testing.
- ~[ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.~
- ~[ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #22390, opened 2026-07-06
Fix filter rewrite crash with sub-aggregations

### LINKED ISSUE #22342: [BUG] OpenSearch 3.x — date_histogram + sub-terms aggregation throws array_index_out_of_bounds_exception
## Summary
 
After upgrading OpenSearch from 2.19 to 3.7, a `date_histogram` aggregation with a
`terms` sub-aggregation began failing with:
 
```
search_phase_execution_exception
  caused_by: array_index_out_of_bounds_exception
  reason: "Index 31 out of bounds for length 31"
```
 
The plain `date_histogram` worked; the error only appeared once the `terms`
sub-aggregation was added.
 
The index was created when OpenSearch version was 2.11

Managed to fix it by disabling the filter rewrite optimization in cluster settings (`"search.max_aggregation_rewrite_filters": 0`), but was wondering if this is a system bug in the first place waiting to be patched.

### Related component

Search:Aggregations

### To Reproduce

An aggregation query with around 50K documents spanning from 1970 to 2025. (A few records dated 1970, and the rest in the recent years 2020-2024) Created a `date_histogram` aggregation query with `calendar_interval` set to month/year,  and a sub `terms` aggregation with an array of strings. (filtered to documents LTE 2024, matching around 8K records)

### Expected behavior

The buckets to show up like it did previously prior to the upgrade

### Additional Details

_No response_

Adam-Horse:
### Description

Fixes an `ArrayIndexOutOfBoundsException` thrown by `date_histogram` (and other bucket aggregations using the filter rewrite optimization) when a sub-aggregation is present and the segment contains documents beyond the query's upper bound.

**Root cause:** when BKD traversal encounters values beyond the last range, `AbstractRangeCollector.iterateRangeEnd` advances `activeIndex` one past the end of the ranges array as its "ranges exhausted" signal and traversal terminates early via `CollectionTerminatedException`. `PointTreeTraversal.multiRangesTraverse` then always makes one final `finalizePreviousRange()` call to flush the in-progress bucket. `SimpleRangeCollector` is safe there because its flush is guarded by `counter > 0`, but `SubAggRangeCollector.finalizePreviousRange()` unconditionally looked up the bucket ordinal for `activeIndex`, reading `ranges.lowers[activeIndex]` on a full array and throwing `ArrayIndexOutOfBoundsException: Index N out of bounds for length N`.

**Fix:** a bounds check at the top of `SubAggRangeCollector.finalizePreviousRange()`. When `activeIndex >= ranges.getSize()`, traversal ended past the last range; no documents can be collected beyond the last range, so there is nothing to flush for the sub-aggregation.

This matches the reporter's observations in #22342: the plain `date_histogram` works (the no-sub-agg path is guarded), the failure appears only when a sub-aggregation is added, and disabling the optimization (`search.max_aggregation_rewrite_filters: 0`) works around it.

**Regression test:** one doc per month of 2024 plus two docs in early 2025 sharing a segment, a range query bounded at 2024-12-31, and a `date_histogram` with a `stats` sub-aggregation. Throws `ArrayIndexOutOfBoundsException` before the fix; returns 12 correct buckets with correct per-bucket stats after.

@bowenlan-amzn @jainankitk tagging you as the authors of this code path (#17447, #19933) in case you'd like to review.

First contribution, happy to adjust if I've missed any process steps (labels, backports).

### Related Issues
Resolves #22342

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/search/aggregations/bucket/filterrewrite/rangecollector/SubAggRangeCollector.java
+++ b/server/src/main/java/org/opensearch/search/aggregations/bucket/filterrewrite/rangecollector/SubAggRangeCollector.java
@@ -99,14 +99,6 @@ public class SubAggRangeCollector extends SimpleRangeCollector {
     public void finalizePreviousRange() {
         super.finalizePreviousRange();
 
-        // Traversal can terminate early once values exceed the last range, leaving
-        // activeIndex past the end of the ranges array. No documents are collected
-        // beyond the last range, so there is nothing to flush for the sub-agg.
-        if (activeIndex >= ranges.getSize()) {
-            bitSet.clear();
-            return;
-        }
-
         long bucketOrd = getBucketOrd.apply(activeIndex);
         logger.trace("finalize range {} with bucket ordinal {}", activeIndex, bucketOrd);
 
--- a/server/src/main/java/org/opensearch/search/aggregations/bucket/filterrewrite/rangecollector/SubAggRangeCollector.java
+++ b/server/src/main/java/org/opensearch/search/aggregations/bucket/filterrewrite/rangecollector/SubAggRangeCollector.java
@@ -99,6 +99,14 @@ public class SubAggRangeCollector extends SimpleRangeCollector {
     public void finalizePreviousRange() {
         super.finalizePreviousRange();
 
+        // Traversal can terminate early once values exceed the last range, leaving
+        // activeIndex past the end of the ranges array. No documents are collected
+        // beyond the last range, so there is nothing to flush for the sub-agg.
+        if (activeIndex >= ranges.getSize()) {
+            bitSet.clear();
+            return;
+        }
+
         long bucketOrd = getBucketOrd.apply(activeIndex);
         logger.trace("finalize range {} with bucket ordinal {}", activeIndex, bucketOrd);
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/opensearch/search/aggregations/bucket/filterrewrite/rangecollector/SubAggRangeCollector.java
```diff
@@ -99,6 +99,14 @@ public void collectDocIdSet(DocIdSetIterator iter) throws IOException {
     public void finalizePreviousRange() {
         super.finalizePreviousRange();
 
+        // Traversal can terminate early once values exceed the last range, leaving
+        // activeIndex past the end of the ranges array. No documents are collected
+        // beyond the last range, so there is nothing to flush for the sub-agg.
+        if (activeIndex >= ranges.getSize()) {
```
bowenlan-amzn: Thanks for fixing this. The regression test is very helpful!
I think moving this finalize call up to line 81, right after intersectWithRanges will do the job
https://github.com/opensearch-project/OpenSearch/blob/feb6941351c2063f26a8d87b32c1adad8ac9c4f7/server/src/main/java/org/opensearch/search/aggregations/bucket/filterrewrite/PointTreeTraversal.java#L84 

The early terminate situations are followed with a finalize call right in place.

Adam-Horse: Done in 3db8442 - moved the finalize call up right after intersectWithRanges inside the try block, and removed the bounds check in SubAggRangeCollector.finalizePreviousRange since it's no longer reachable with activeIndex past the end. The early-terminate paths in the visitor each finalize in place before throwing, so nothing is left to flush after CollectionTerminatedException. Regression test still passes.

bowenlan-amzn (approved):

jainankitk (approved):
Thanks @Adam-Horse for fixing this issue!

PR #22392, opened 2026-07-06
Add BWC tests to prevent throttling key removal

mohit10011999:
### Description
Add BWC tests for persisted throttling task keys

Add rolling-upgrade test for throttling settings

### Related Issues

### Check List


qa/rolling-upgrade/src/test/java/org/opensearch/upgrades/ClusterManagerThrottlingSettingsIT.java
```diff
@@ -0,0 +1,82 @@
...
+                trySetThrottlingSetting("put-mapping", 5000);
+                trySetThrottlingSetting("create-index", 100);
+            }
+            // Verify cluster is healthy in old state
+            verifyClusterHealth();
+        } else {
+            // After upgrade (mixed or fully upgraded), verify the cluster is healthy.
+            // If settings restoration failed due to missing task key registrations,
+            // nodes would have crashed and the cluster would not be green/yellow.
+            verifyClusterHealth();
```
cwperks: I think this `else` block can be dropped altogether and have a single call to `verifyClusterHealth` if there is no different in each branch of logic.

server/src/test/java/org/opensearch/cluster/service/ClusterManagerTaskThrottlerTests.java
```diff
@@ -635,4 +638,108 @@ public String describeTasks(List tasks) {
...
+            )
+        )
+    );
+
+    /**
+     * Verifies that every historically persisted task key still exists in the ClusterManagerTask enum.
+     * If this test fails, a key was removed that is persisted in cluster state by prior versions,
+     * which will cause fatal node startup failures during upgrades.
+     */
+    public void testBWCAllHistoricallyPersistedKeysExistInEnum() {
```
cwperks: IMO this test is too brittle. I think we can mark ClusterManagerTask enum at `@PublicApi` which would trigger failure in `Detect Breaking Changes` check of core which enforces that any public interfaces cannot be changed in breaking ways from one version to the next.

mohit10011999: @PublicApi + japicmp protects against API-level removal (deleting the enum value). It does not protect against registration-level removal (keeping the enum value but not registering it at runtime).

 if someone removes a `registerClusterManagerTask()` call for a task that's still in the enum (and potentially persisted in cluster state), this test will fail because it proves the registration path works. It's not about maintaining a hardcoded list; it's about proving the enum → registration → validation pipeline is intact

server/src/test/java/org/opensearch/cluster/service/ClusterManagerTaskThrottlerTests.java
```diff
@@ -635,4 +635,36 @@ public String describeTasks(List tasks) {
...
+            assertTrue(
+                "ClusterManagerTask."
+                    + task.name()
+                    + " (key='"
+                    + task.getKey()
+                    + "') was not found in "
+                    + "THROTTLING_TASK_KEYS after registration. If this task's registerClusterManagerTask() "
+                    + "call is removed, persisted cluster state containing its throttling threshold will cause "
+                    + "a fatal node startup failure during upgrades.",
+                throttler.THROTTLING_TASK_KEYS.containsKey(task.getKey())
```
cwperks: I know this isn't introduced in this PR, but I thinking SCREAMING_SNAKE_CASE should be reserved for constants.

server/src/test/java/org/opensearch/cluster/service/ClusterManagerTaskThrottlerTests.java
```diff
@@ -635,4 +635,36 @@ public String describeTasks(List tasks) {
...
+     * validateSetting() to throw IllegalArgumentException during cluster state restoration.
+     */
+    public void testAllEnumTasksCanBeRegistered() {
+        ClusterSettings clusterSettings = new ClusterSettings(Settings.builder().build(), ClusterSettings.BUILT_IN_CLUSTER_SETTINGS);
+        ClusterManagerTaskThrottler throttler = new ClusterManagerTaskThrottler(Settings.EMPTY, clusterSettings, () -> {
+            return clusterService.getClusterManagerService().getMinNodeVersion();
+        }, new ClusterManagerThrottlingStats());
+
+        for (ClusterManagerTask task : ClusterManagerTask.values()) {
+            throttler.registerClusterManagerTask(task, true);
```
cwperks: When would it be the case that a value in this enum cannot be registered? 

Based on the comment 

> ...scenario where a task key exists in the enum (and may be persisted in cluster state)

should we be testing the case where enum value doesn't have corresponding `registerClusterManagerTask` call?

btw what PR corresponds to the fix for this issue? Are we retaining old keys in perpetuity within cluster state or do they get cleaned up if no longer needed?

server/src/test/java/org/opensearch/cluster/service/ClusterManagerTaskThrottlerTests.java
mohit10011999: Agreed, that's pre-existing naming. Won't address in this PR

server/src/test/java/org/opensearch/cluster/service/ClusterManagerTaskThrottlerTests.java
mohit10011999: **When would it be the case that a value in this enum cannot be registered?**

It can always be registered - the issue is when a `registerClusterManagerTask()` call is removed from production code. In the incident, the three persistent task keys (create-persistent-task, finish-persistent-task, remove-persistent-task) still existed in the enum, but PersistentTasksClusterService stopped calling registerClusterManagerTask() for them after the batching refactor. So at runtime `THROTTLING_TASK_KEYS` didn't contain them, and `validateSetting()` threw during cluster state restoration.

**btw what PR corresponds to the fix for this issue?**
That fix commit is in managed service only currently. https://github.com/opensearch-project/OpenSearch/pull/22656 for opensource


**Are we retaining old keys in perpetuity within cluster state or do they get cleaned up if no longer needed?**
The keys stay in cluster state until someone explicitly removes the persistent setting. There's no automatic cleanup. In managed service, got to know that we can archive the setting

cwperks: ok I see the issue fully now, do you think we should consider changing https://github.com/opensearch-project/OpenSearch/blob/f128e1d712b7c97858522fe5e527f9a2ec9dce13/server/src/main/java/org/opensearch/cluster/service/ClusterManagerTaskThrottler.java#L183-L185 to a warning log? 

Even if it changes to a warning log if someone uses `PUT _cluster/settings` to modify `cluster_manager.throttling.thresholds.create-persistent-task.value` then we may have some phantom settings around that _should_ be allowed since [THRESHOLD_SETTINGS is a group setting](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/cluster/service/ClusterManagerTaskThrottler.java#L51-L55).

cwperks: ah nvm `validateSetting` is used to validate that user supplies a valid throttling key when using PUT _cluster/settings API. I think the alternative is version guards that get removed in next major release, but the solution you put forward makes sense as well. Marking resolved.

cwperks (approved):

PR #5199, opened 2022-11-10
Adding depth check in doc parser for deep nested document

nkumar04:
### Description
Added a depth level check within the document parser context. The depth level is incremented for each nested object parsing and decremented when parsing is done. If the depth level crosses a threshold, document parser throws a parsing exception and an error is returned in the response, stating the reason of the failure

### Issues Resolved
Closes issue (#5195)

### Check List
- [ ] New functionality includes testing.
  - [ ] All tests pass
- [ ] New functionality has been documented.
  - [ ] New functionality has javadoc added
- [ ] Commits are signed per the DCO using --signoff
- [ ] Commit changes are listed out in CHANGELOG.md file (See: [Changelog](../blob/main/CONTRIBUTING.md#changelog))

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #22404, opened 2026-07-07
Fix unbounded recursion in deserialization that can cause StackOverflowError

DarshitChanpura:
### Description
Fix unbounded recursion in deserialization that can cause StackOverflowError.
 Deeply nested JSON or binary payloads can trigger unbounded recursion in two deserialization paths, causing a `StackOverflowError` that kills the JVM rather than producing a catchable exception.
  
This PR adds depth tracking to:
  - **`AbstractXContentParser`** (`readListUnsafe`/`readValueUnsafe`/`readMapEntries`) — the XContent (JSON/YAML/CBOR/Smile) parsing layer
  - **`StreamInput`** (`readGenericValue`/`readArrayList`/`readArray`/`readLinkedHashMap`/`readHashMap`) — the binary transport/serialization layer
  
Both layers now enforce `MAX_DESERIALIZATION_DEPTH=100` (configurable via `-Dopensearch.xcontent.depth.max`), throwing a catchable `IOException` or `IllegalStateException` instead of allowing the JVM to crash.
  
Additionally lowers the Jackson `StreamReadConstraints` `maxNestingDepth` from 1000 to 100 as defense-in-depth, aligning all three layers to the same limit.
  
### Related Issues
- N/A
  
### Check List

### DIFF
--- a/libs/core/src/main/java/org/opensearch/core/common/io/stream/StreamInput.java
+++ b/libs/core/src/main/java/org/opensearch/core/common/io/stream/StreamInput.java
@@ -723,11 +723,24 @@ public abstract class StreamInput extends InputStream {
     }
 
     /**
+     * Maximum depth for recursive deserialization of nested generic values.
+     * Protects against StackOverflowError from deeply nested binary payloads.
+     */
+    private static final int MAX_DESERIALIZATION_DEPTH = Integer.parseInt(System.getProperty("opensearch.xcontent.depth.max", "100"));
+
+    /**
      * Reads a value of unspecified type. If a collection is read then the collection will be mutable if it contains any entry but might
      * be immutable if it is empty.
      */
     @Nullable
     public Object readGenericValue() throws IOException {
+        return readGenericValue(0);
+    }
+
+    private Object readGenericValue(int depth) throws IOException {
+        if (depth > MAX_DESERIALIZATION_DEPTH) {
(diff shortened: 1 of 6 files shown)


reta:
@DarshitChanpura I believe we have the check on the `DocumentParser` side (see please https://github.com/opensearch-project/OpenSearch/pull/5199 as an example), aren't these being enforced properly?

DarshitChanpura:
@reta Based on my understanding of DocumentParser and the changes in that PR, it only covers mapped document parsing path whereas changes in this PR cover generic deserialization path as well as other paths that don't go through DocumentParser.

The primary gap this PR addresses is StreamInput.readGenericValue() — the binary transport deserialization path has no Jackson involvement and had zero depth protection, so a crafted binary payload could cause an unrecoverable StackOverflowError.

Additionally, this lowers the Jackson maxNestingDepth from 1000 to 100 as hardening — 1000 levels of nesting is far beyond any legitimate use case.

cwperks:
@reta as @DarshitChanpura mentioned there's a gap particularly at https://github.com/opensearch-project/OpenSearch/blob/main/libs/core/src/main/java/org/opensearch/core/common/io/stream/StreamInput.java#L722 which doesn't go through DocumentParser.

cwperks (approved):

PR #22405, opened 2026-07-08
Add low_cardinality_enable field-level property for keyword/text fields

cocosz:
# Add `low_cardinality_enable` field-level property for parquet composite indices

### Description

Adds a new field-level index setting `low_cardinality_enable` for the `parquet-data-format` sandbox plugin. When set to `true` for a `keyword` or `text` field on a parquet composite index, it automatically applies two optimizations at index creation time:

1. **Suppresses the Lucene inverted index** (`index=false`) — avoids write amplification and reduces index size for fields that will only be queried via Parquet bloom filters rather than Lucene term queries. The suppression is applied at mapper build time using the standard `IndexOptions.NONE` mechanism, so it integrates cleanly with all field mapper types and the composite engine's capability routing without requiring changes to the core indexing path.
2. **Enables a Parquet bloom filter** for that column — so point-lookup equality queries can skip entire row groups efficiently at the Parquet layer.

This is designed for low-cardinality string fields (e.g. `status`, `region`, `event_type`) in analytical parquet composite indices where exact-match filtering via bloom filters is the primary access pattern and the Lucene inverted index overhead is wasteful.

#### Usage

```json
PUT /my-index
{
  "settings": {
    "index.composite.secondary_data_formats": ["lucene"],
    "index.parquet.low_cardinality_enable.field": ["status", "region"],
    "index.parquet.low_cardinality_enable.value": [true, true]
  },
  "mappings": {
    "properties": {
      "status": { "type": "keyword" },
      "region": { "type": "keyword" }
    }
  }
}
```

This is equivalent to setting `"index": false` on those fields in the mapping **plus** enabling `index.parquet.bloom_filter_enabled` for the same fields — but expressed as a single declarative intent.

---

#### `ParquetSettings.java`
- Added `LOW_CARDINALITY_ENABLE_FIELD_SETTING` and `LOW_CARDINALITY_ENABLE_VALUE_SETTING` — parallel-array index settings (`IndexScope`, `Final`), following the same pattern as existing `bloom_filter_enabled` and `encoding` settings.
- Added `getLowCardinalityEnabledFields(Settings)` — returns the set of field names whose value is `true`.
- Extended `validateFieldConfigurations()` with a new overload that accepts `Set<String> lowCardinalityEnabledFields` and `MapperService` to enforce the keyword/text-only restriction at index creation time.
- Kept a `@Deprecated` 4-arg overload for backward compatibility with existing callers.
- Registered both new settings in `getSettings()`.

#### `ParquetIndexCreationValidator.java`
- Reads `lowCardinalityEnabledFields`, includes it in the `hasParquetSettings` guard (so a non-parquet index with this setting is rejected), and passes it to the new `validateFieldConfigurations` overload with `MapperService`.

#### `NativeSettings.java`
- Added `Set<String> lowCardinalityEnabledFields` field, `getLowCardinalityEnabledFields()` getter, and `Builder.lowCardinalityEnabledFields(Set<String>)` method to carry the field set through to the Rust layer.

#### `RustBridge.java`
- In `onSettingsUpdate()`, merges `lowCardinalityEnabledFields` into the effective `fieldBloomFilterEnabled` map via `putIfAbsent(field, true)` before passing to the Rust FFI call. Explicit `bloom_filter_enabled` entries take precedence over the implicit bloom from `low_cardinality_enable`.

#### `ParquetIndexingEngine.java`
- Adds `.lowCardinalityEnabledFields(ParquetSettings.getLowCardinalityEnabledFields(settings))` to the `NativeSettings` builder chain in `pushSettingsToRust()`.

#### `KeywordFieldMapper.java` and `TextFieldMapper.java` (server module)
- Added `isLowCardinalityEnabled(BuilderContext)` private helper in each mapper's `Builder` class. Reads `index.parquet.low_cardinality_enable.field/value` from `BuilderContext.indexSettings()` by raw key. This keeps the server module free of any compile-time dependency on sandbox plugin classes — the mapper simply checks whether its fully-qualified field name appears in the configured list, regardless of which plugin registered the setting.
- In `build()`, calls `isLowCardinalityEnabled()` and sets `indexed.setValue(false)` / `index.setValue(false)` if matched. This causes the Lucene `FieldType` to use `IndexOptions.NONE` — the standard mechanism for suppressing the inverted index on any field mapper — which propagates through `MappedFieldType.isSearchable()=false` into the composite engine's capability assignment, ensuring no posting lists are written for that field in the Lucene segment.

---

### Validation

All error paths enforced at index creation time via `ParquetIndexCreationValidator`:

| Condition | Error |
|---|---|
| Field type is not `keyword` or `text` | `low_cardinality_enable is only supported for keyword and text fields` |
| Field name not present in mappings | `does not exist in mappings` |
| Setting used on a non-parquet index | `does not use parquet data format` |
| Duplicate field name in the array | `Duplicate field '...'` |
| Field/value arrays have different lengths | `must have the same size` |
| Invalid boolean value (e.g. `"yes"`) | `Must be true or false` |
| `value=false` | Accepted, no-op (field stays indexed) |

---

### Storage and accuracy characteristics

This setting is a storage-vs-access-path trade-off. It does not change query results — it changes *what is written on disk* and *which path answers a lookup*.

#### Storage impact

Two independent effects combine:

- **Lucene inverted index removed (saves space).** With `index=false` the field no longer contributes a term dictionary or postings lists to the Lucene segment. The saving scales with document count and, for the term dictionary, with cardinality. Low-cardinality string fields are the ideal case: the postings (one entry per document) dominate the on-disk cost, so suppressing them removes the bulk of the field's Lucene footprint while the term dictionary that would have been saved is small anyway. This also removes the per-document indexing work (write amplification) at ingest time.
- **Parquet bloom filter added (costs space).** A bloom filter is written per column chunk (per row group). Its size is governed by the target false-positive probability and the distinct-value hint: `index.parquet.bloom_filter_fpp` (default `0.1`) and `index.parquet.bloom_filter_ndv` (default `100000`). Lower FPP or higher NDV produce a larger filter. For genuinely low-cardinality fields the distinct-value count is small, so the filter stays compact relative to the column data.

Net effect for the intended use case (low-cardinality strings queried by exact match): the Lucene postings removed are typically much larger than the bloom filter added, so total index size goes down. If you set this on a high-cardinality field, the bloom filter grows and the FPP degrades (see below), which is why the feature is restricted in intent to low-cardinality fields even though it is only *type*-restricted (keyword/text) at validation time.

Tuning note: the default `ndv=100000` is a global hint. For a field with far fewer distinct values you can lower `bloom_filter_ndv` to shrink the filter, or lower `bloom_filter_fpp` to trade space for fewer false positives.

#### Accuracy impact

- **Query correctness is unchanged.** Parquet bloom filters are a probabilistic *skip* structure: they can yield false positives (a row group is scanned when it held no match) but never false negatives (a matching row group is never skipped). A false positive only costs an unnecessary row-group scan; it never drops a matching document. Equality/point-lookup results remain exact.
- **False-positive rate is bounded and configurable.** The effective FPP tracks `bloom_filter_fpp` as long as the actual distinct-value count stays at or below the `ndv` hint. If real cardinality substantially exceeds the hint, the observed false-positive rate rises above the target — meaning fewer row groups are skipped (a performance regression), not incorrect results.
- **Functional trade-off from suppressing Lucene.** With `index=false` the field reports `searchable=false` and is no longer answerable via the Lucene inverted index; lookups route through the Parquet/bloom path. Exact-match filtering stays correct, but Lucene-side term/full-text querying on that field is intentionally given up. Choose this only for fields whose access pattern is exact-match skipping at the Parquet layer.

---

### Testing

Manually verified on a live OpenSearch 3.8.0-SNAPSHOT node with the full sandbox plugin stack (`composite-engine`, `parquet-data-format`, `analytics-backend-datafusion`).

**Lucene index suppression**

The `_field_caps` API reports `searchable=false` for fields with `low_cardinality_enable=true`, confirming no inverted index is written. The result is identical to explicitly setting `"index": false` in the field mapping:

```
status (keyword): searchable=False   ← low_cardinality_enable=true
category (keyword): searchable=True  ← plain keyword, unaffected
```

**Parquet bloom filter**

Inspected the written Parquet file using `parquet-bloom-info`. A bloom filter is present only for the `low_cardinality_enable=true` field, and absent for unaffected fields — identical behavior to using `index.parquet.bloom_filter_enabled.field/value` explicitly:

```
column        bloom_filter_offset  bloom_filter_length
status                        630                   47   ← bloom written
category                     None                 None   ← no bloom (expected)
```

**Correctness cases (11 total)**

| # | Case | Result |
|---|---|---|
| 1 | `keyword` fields with `value=true` | Accepted; `searchable=false`, bloom written |
| 2 | `text` field with `value=true` | Accepted; `searchable=false`, bloom written |
| 3 | `value=false` (explicit opt-out) | Accepted; field stays searchable, no bloom |
| 4 | `low_cardinality_enable` coexists with `bloom_filter_enabled` + `encoding` | All settings applied correctly; explicit entries take precedence |
| 5 | Settings round-trip via `_settings` API | `field` and `value` arrays persisted correctly |
| 6 | Non-text/keyword field type (`double`, `integer`, …) | Rejected at index creation |
| 7 | Field name not present in mappings | Rejected at index creation |
| 8 | Setting used on a non-parquet index | Rejected at index creation |
| 9 | Duplicate field name in the array | Rejected at index creation |
| 10 | Field/value arrays have different lengths | Rejected at index creation |
| 11 | Invalid boolean value (e.g. `"yes"`) | Rejected at index creation |

---

### Check List


Bukhtawar (commented):
The one thing I need to understand is what if the cardinality is very low like HTTP status codes, will `low_cardinality` be any significant value? It definitely is for `high_cardinality` but at the expense of low performance. How do we highlight this positiong?

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -208,6 +208,24 @@ public Builder docValues(boolean hasDocValues) {
...
+        private boolean isLowCardinalityEnabled(BuilderContext context) {
+            if (!context.indexSettings().getAsBoolean(IndexSettings.PLUGGABLE_DATAFORMAT_ENABLED_SETTING.getKey(), false)) {
+                return false;
+            }
+            List<String> fields = context.indexSettings().getAsList("index.parquet.low_cardinality_enable.field", Collections.emptyList());
+            if (fields.isEmpty()) return false;
+            List<String> values = context.indexSettings().getAsList("index.parquet.low_cardinality_enable.value", Collections.emptyList());
+            String fullName = buildFullName(context);
+            for (int i = 0; i < fields.size(); i++) {
+                if (fullName.equals(fields.get(i)) && i < values.size() && "true".equalsIgnoreCase(values.get(i))) {
```
mgodwan: Can this be a mapping parameter?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetSettings.java
```diff
@@ -480,6 +481,24 @@ public static boolean isEncodingValidForArrowType(String encoding, ArrowType arr
...
+    );
+
+    public static final Setting<List<Boolean>> LOW_CARDINALITY_ENABLE_VALUE_SETTING = Setting.listSetting(
+        "index.parquet.low_cardinality_enable.value",
+        Collections.emptyList(),
+        ParquetSettings::validateBoolean,
+        Setting.Property.IndexScope,
+        Setting.Property.Final
+    );
+
```
mgodwan, line 488: The correct experience here would be a mapping parameter. I think we should expose a way for plugins to provide params into the mappers

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
cocosz: We can do that I will cover that in the next commit .

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetDataFormatPlugin.java
```diff
@@ -197,6 +199,20 @@ public DataFormat getDataFormat() {
         return PARQUET_DATA_FORMAT;
     }
 
+    /** Contributes the {@code low_cardinality} parameter to {@code keyword} and {@code text} fields, suppressing Lucene indexing and enabling a Parquet column bloom filter. */
+    @Override
+    public List<PluginMappingParameter> getPluginMappingParameters() {
```
mgodwan: Can we pass the content type here?

server/src/main/java/org/opensearch/index/mapper/PluginMappingParameter.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * A typed mapping parameter contributed by a data-format plugin to core field mappers (currently
+ * {@code keyword} and {@code text}). The {@link Type} determines how the value is parsed and serialized.
+ * A boolean parameter may additionally request core-defined {@link FieldBuildEffect build-time effects}
+ * (e.g. disabling Lucene indexing) that the mapper applies when the value resolves to {@code true}.
+ *
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public final class PluginMappingParameter {
```
mgodwan: Can we not use the Parameter class directly?

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
```diff
@@ -1018,6 +1049,8 @@ protected String contentType() {
 
     @Override
     public ParametrizedFieldMapper.Builder getMergeBuilder() {
-        return new Builder(simpleName(), indexAnalyzers, canConsumeRawValueForSource).init(this);
+        Builder builder = new Builder(simpleName(), indexAnalyzers, canConsumeRawValueForSource);
+        builder.addPluginMappingParameters(pluginMappingParameterSpecs);
```
mgodwan: Can we make this a ctor arg instead of `addPluginMappingParameters`  method?

server/src/main/java/org/opensearch/index/mapper/PluginMappingParameter.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * A build-time effect a boolean parameter can request on its field. The vocabulary is owned by core:
+     * plugins select from it but cannot define new effects, since effects mutate core mapper behaviour.
+     *
+     * @opensearch.experimental
+     */
+    @ExperimentalApi
+    public enum FieldBuildEffect {
+        /** Suppresses Lucene indexing for the field (equivalent to {@code index: false}). */
+        DISABLE_INDEXING
```
mgodwan: Instead of enum, can this be a function?

server/src/main/java/org/opensearch/index/mapper/PluginMappingParameter.java
```diff
@@ -0,0 +1,116 @@
...
+    public boolean updateable() {
+        return updateable;
+    }
+
+    public Set<String> applicableContentTypes() {
+        return applicableContentTypes;
+    }
+
+    /** The build-time effects to apply when this (boolean) parameter resolves to {@code true}. */
+    public Set<FieldBuildEffect> effectsWhenEnabled() {
```
mgodwan: All of these can be record methods

server/src/main/java/org/opensearch/index/mapper/TextFieldMapper.java
```diff
@@ -375,25 +375,29 @@ public Builder addMultiField(Mapper.Builder<?> builder) {
...
-                similarity,
-                positionIncrementGap,
-                fieldData,
-                freqFilter,
-                eagerGlobalOrdinals,
-                indexPhrases,
-                indexPrefixes,
-                boost,
-                meta
+            List<Parameter<?>> parameters = new ArrayList<>(
```
mgodwan, line 383: This method should return an immutable list

server/src/main/java/org/opensearch/index/mapper/PluginMappingParametersAware.java
```diff
@@ -0,0 +1,30 @@
...
+ * @opensearch.experimental
+ */
+@ExperimentalApi
+public interface PluginMappingParametersAware {
+
+    /** Returns the resolved values of the plugin-contributed parameters on this mapper, keyed by parameter name. */
+    Map<String, Object> pluginMappingParameterValues();
+
+    /** Returns the plugin-contributed parameter definitions this mapper was built with. */
+    List<PluginMappingParameter> pluginMappingParameterSpecs();
```
mgodwan: Remove this if not used.

server/src/main/java/org/opensearch/index/mapper/PluginMappingParameter.java
mgodwan: For side effects, we can make the parameter class sealed, and have another SideEffectParameter class which can apply overrides for any existing values

mgodwan (approved):

mgodwan:
Please add an end to end test before we merge this

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeLowCardinalityIT.java
```diff
@@ -0,0 +1,195 @@
...
+
+        // Control field: no low_cardinality → parameter absent and indexing not disabled.
+        Map<String, Object> userId = fieldMapping(properties, "user_id");
+        assertNull("control field must not expose low_cardinality", userId.get("low_cardinality"));
+        assertNotEquals("control field indexing must not be disabled", Boolean.FALSE, userId.get("index"));
+
+        // Write path still works with the suppressed-indexing fields.
+        indexLowCardinalityDocs(INDEX_NAME, 5);
+        refreshIndex(INDEX_NAME);
+        flushIndex(INDEX_NAME);
```
mgodwan, line 66: Can you add search flow as well to the test?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/ParquetDataFormatPlugin.java
cocosz: I have updated in latest version

server/src/main/java/org/opensearch/index/mapper/PluginMappingParameter.java
cocosz: Done

server/src/main/java/org/opensearch/index/mapper/KeywordFieldMapper.java
cocosz: Yes changed

server/src/main/java/org/opensearch/index/mapper/PluginMappingParameter.java
cocosz: ok sure

mgodwan (approved):

PR #22406, opened 2026-07-08
Integrated updates and deletes for DataFormatAwareEngine

thorkous:
### Description

Adds **update and delete** support to `DataFormatAwareEngine` (DFAE) via **hard deletes** — no soft-delete tombstones. A delete/update marks the row deleted in Lucene's `liveDocs`; the row stays physically present in the columnar segment (1:1 by row id) and is filtered at query time. `refresh()` reconciles liveness across both formats, so deleted rows — and generations that become fully deleted — drop out of both search and the catalog snapshot.

**Delete execution layer (format-agnostic) — `server/…/index/engine/dataformat/`**

- **`Deleter` / `DeleterImpl`** — per-generation delete primitive. `recordBufferedDeletes(id)` buffers an `_id` for a Term-delete applied when the generation is checked out at refresh; `recordPositionalDelete(rowId)` supersedes a specific prior row. Stays format-agnostic: owns the buffered/active lifecycle, delegates the positional delete to its paired `Writer`.
- **`DeleteExecutionEngine`** — contract a format implements to execute deletes when it can't delete natively: `createDeleter`/`getDeleter` per generation, `recordWrite(id, generation, rowId)`, `deleteDocument(...)`, and `onWriterCheckedOut(...)` to drain buffered deletes at refresh.
- **`DeleteResult`** — `Success(version, primaryTerm, seqNo)` / `Failure(cause)`. `Success` fields are reserved for a future storage-layer consumer; the engine layer is authoritative for version/seqNo/term today.
- **`Writer.recordPositionalDelete(long)`** — new optional writer capability (default throws), mirroring `deleteDocument`, so a deleter can forward a row-id delete to whichever writer supports it.
- **`RefreshResult.droppedGenerations`** — refresh now reports generations that became fully deleted, so the engine can drop them from the catalog snapshot.

**Engine wiring — `DataFormatAwareEngine.java`**

- **`delete()`** — full delete path mirroring `InternalEngine.delete()`: plan via `DeletionStrategyPlanner` (version-conflict detection), assign seqNo on a primary, reserve/release against the in-flight-doc budget, route to the format's `DeleteExecutionEngine`, record a version-map tombstone, append `Translog.Delete`. Handles PRIMARY and replica/recovery origins; decomposed into `planDelete` / `generateSeqNoForDelete` / `executeDeletePlan` / `finalizeDelete` / `deleteInEngine`.
- **`prepareDelete()`** + version resolution wired into both the indexing and new deletion planners; expired tombstones pruned after refresh (`maybePruneDeletes`).
- **`refresh()`** — determines live vs fully-deleted generations, drops fully-deleted ones from the catalog snapshot, and reopens reader managers when deletes were applied so **pure deletes become visible after refresh**.

**Lucene format — `sandbox/plugins/analytics-backend-lucene/…`**

- **`LuceneDeleteExecutionEngine`** — the Lucene-backed `DeleteExecutionEngine`. Tracks `id → (generation, rowId)`; on delete it buffers a Term-delete on the current generation and, when a prior copy still lives in an active earlier generation, also records a positional (row-id) delete on that generation's deleter. Buffered Term-deletes drain into the parent writer as `deleteDocuments(new Term(_id, …))` when the generation is checked out at refresh. Takes no second writer-holder lock → deadlock-free against `refresh().checkoutAll()` (the earlier `writerByGenSupplier` design AB-BA-deadlocked here) and recovery-robust (id-based, idempotent).
- **`LuceneWriter`** — buffers row-id deletes (`recordPositionalDelete`) and applies them as `liveDocs`-only deletes during flush (`applyPositionalDeletes`), mapping insertion row ids to post-reorder positions so the row stays 1:1 with the columnar segment; `deleteDocument(Term)` delegates to `IndexWriter.deleteDocuments`.
- **`LuceneIndexingExecutionEngine.refresh()`** — derives `liveGenerations` from the reopened reader's leaves and reports `droppedGenerations` (live before the refresh, leaf now gone ⇒ fully deleted). This lets the engine drop the whole generation, including its columnar segment, which has no `liveDocs` of its own to detect the full delete.

**Composite format — `CompositeIndexingExecutionEngine.java`**

- Unions `droppedGenerations` across the primary and secondary formats and propagates it on both the merge-on-refresh and normal refresh paths.

**`IndexMetadata.java`**

- Clarifying comment on append-only setting precedence (explicit `index.append_only.enabled` vs the value auto-derived from pluggable-dataformat).

### Related Issues

Part of the DataFusion update/delete support initiative; builds on #21597 (filter live docs / remove deleted rows during parquet merge).

### Check List

### DIFF
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneReaderManager.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneReaderManager.java
@@ -28,11 +28,9 @@ import org.opensearch.index.engine.exec.coord.CatalogSnapshot;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Set;
 
 import static org.opensearch.be.lucene.index.LuceneWriter.WRITER_GENERATION_ATTRIBUTE;
 
@@ -44,8 +42,8 @@ import static org.opensearch.be.lucene.index.LuceneWriter.WRITER_GENERATION_ATTR
  * to {@link LuceneReader} so each snapshot gets the reader that was current
  * at the time of its refresh. On each {@link #afterRefresh}, the current reader is
  * refreshed via {@link DirectoryReader#openIfChanged} and paired with a
- * {@code writer_generation → leaf index} map built by matching the catalog's
- * {@link WriterFileSet#files()} against each leaf's {@code SegmentCommitInfo.files()}.
+ * {@code writer_generation → segment name} map built from each leaf's writer-generation
+ * stamp (stable across the liveDocs/doc-values updates a drained delete writes onto a segment).
  *
  * @opensearch.experimental
  */
@@ -131,28 +129,35 @@ public class LuceneReaderManager implements EngineReaderManager<LuceneReader> {
     }
 
     private static Map<Long, String> buildGenerationToSegmentName(CatalogSnapshot catalogSnapshot, List<LeafReaderContext> leaves) {
-        // Index leaves by their file set → segment name
-        Map<Set<String>, String> filesToSegName = new HashMap<>(leaves.size());
-        for (int i = 0; i < leaves.size(); i++) {
-            SegmentReader sr = (SegmentReader) leaves.get(i).reader();
-            try {
-                filesToSegName.put(new HashSet<>(sr.getSegmentInfo().files()), sr.getSegmentInfo().info.name);
-            } catch (IOException e) {
-                throw new IllegalStateException("Failed to read files for leaf " + i, e);
+        // Index leaves by writer generation → segment name.
+        Map<Long, String> generationToLeafSegment = new HashMap<>(leaves.size());
+        for (LeafReaderContext lrc : leaves) {
+            SegmentReader sr = (SegmentReader) lrc.reader();
+            SegmentCommitInfo sci = sr.getSegmentInfo();
+            String genAttr = sci.info.getAttribute(WRITER_GENERATION_ATTRIBUTE);
+            if (genAttr == null) {
+                throw new IllegalStateException(
+                    "Lucene leaf segment=" + sci.info.name + " is missing the " + WRITER_GENERATION_ATTRIBUTE + " attribute"
+                );
             }
+            generationToLeafSegment.put(Long.parseLong(genAttr), sci.info.name);
         }
 
-        // Match catalog segments to leaves via file sets
+        // Resolve each catalog segment carrying Lucene data to its leaf's segment name by generation.
         Map<Long, String> out = new HashMap<>();
         for (Segment seg : catalogSnapshot.getSegments()) {
             WriterFileSet wfs = seg.dfGroupedSearchableFiles().get(LuceneDataFormat.LUCENE_FORMAT_NAME);
             if (wfs == null) {
                 continue;
             }
-            String segName = filesToSegName.get(wfs.files());
+            String segName = generationToLeafSegment.get(seg.generation());
             if (segName == null) {
                 throw new IllegalStateException(
-                    "Catalog segment gen=" + seg.generation() + " files=" + wfs.files() + " has no matching Lucene leaf"
+                    "Catalog segment gen="
+                        + seg.generation()
+                        + " has no matching Lucene leaf (leaf generations="
+                        + generationToLeafSegment.keySet()
+                        + ")"
                 );
             }
             out.put(seg.generation(), segName);
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
@@ -30,14 +30,12 @@ import org.opensearch.index.mapper.Uid;
 import org.opensearch.index.store.Store;
 import org.opensearch.plugin.stats.DataFormatStatsProviderRegistry;
 
-import java.io.Closeable;
 import java.io.IOException;
 import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
-import java.util.function.LongFunction;
 
 /**
  * Lucene-based implementation of {@link DeleteExecutionEngine} that tracks per-generation
@@ -53,8 +51,12 @@ public class LuceneDeleteExecutionEngine implements DeleteExecutionEngine<DataFo
     private final Map<Long, Deleter> generationToDeleterMap;
     private final DataFormat dataFormat;
     private final IndexWriter parentWriter;
-    private final ConcurrentMap<String, Long> idToGen;
+    private final ConcurrentMap<String, GenRow> idToGen;
     private final Store store;
+
+    /** Generation + insertion rowId where a document currently lives in an active child writer. */
+    private record GenRow(long generation, long rowId) {
+    }
 
     public LuceneDeleteExecutionEngine(DataFormat dataFormat, Committer committer) {
         this.generationToDeleterMap = new ConcurrentHashMap<>();
@@ -67,14 +69,12 @@ public class LuceneDeleteExecutionEngine implements DeleteExecutionEngine<DataFo
 
     @Override
     public Deleter createDeleter(Writer<?> writer) {
-        LuceneWriter luceneWriter = writer.getWriterForFormat(LuceneDataFormat.LUCENE_FORMAT_NAME)
-            .map(w -> (LuceneWriter) w)
-            .orElseThrow(
-                () -> new IllegalArgumentException("Cannot create deleter: no Lucene writer found for generation=" + writer.generation())
-            );
-        Deleter deleter = new DeleterImpl<>(luceneWriter);
-        generationToDeleterMap.put(writer.generation(), deleter);
-        return deleter;
+        return writer.getWriterForFormat(LuceneDataFormat.LUCENE_FORMAT_NAME).map(w -> {
+            LuceneWriter luceneWriter = (LuceneWriter) w;
+            Deleter deleter = new DeleterImpl<>(luceneWriter);
+            generationToDeleterMap.put(writer.generation(), deleter);
+            return deleter;
+        }).orElse(null);
     }
 
     @Override
@@ -83,29 +83,28 @@ public class LuceneDeleteExecutionEngine implements DeleteExecutionEngine<DataFo
     }
 
     @Override
-    public DeleteResult deleteDocument(DeleteInput deleteInput, LongFunction<Closeable> writerByGenSupplier) throws IOException {
+    public DeleteResult deleteDocument(DeleteInput deleteInput) throws IOException {
         long start = System.nanoTime();
         try {
             Deleter currentDeleter = generationToDeleterMap.get(deleteInput.generation());
-            assert currentDeleter != null && currentDeleter.isActive()
-                : "current-gen deleter must exist and be active while caller holds the writer lock; gen=" + deleteInput.generation();
+            if (currentDeleter == null) {
+                throw new IllegalArgumentException(
+                    "Update/delete is not supported for this index: no delete-applicable data format "
+                        + "(requires a format such as Lucene)"
+                );
+            }
+            assert currentDeleter.isActive() : "current-gen deleter must be active while caller holds the writer lock; gen="
+                + deleteInput.generation();
 
-            // TODO: If not present then record buffered deletes.
             currentDeleter.recordBufferedDeletes(deleteInput.id());
-            Long previousGen = lookupGen(deleteInput.id());
-            if (previousGen != null) {
-                Closeable previousWriterLock = writerByGenSupplier.apply(previousGen);
-                if (previousWriterLock != null) {
-                    // It means previous writer is active here.
-                    try {
-                        Deleter deleter = generationToDeleterMap.get(previousGen);
-                        return deleter.deleteDoc(deleteInput);
-                    } finally {
-                        previousWriterLock.close();
-                    }
+            GenRow previous = idToGen.get(deleteInput.id());
+            if (previous != null) {
+                Deleter previousDeleter = generationToDeleterMap.get(previous.generation());
+                if (previousDeleter != null) {
+                    previousDeleter.recordPositionalDelete(previous.rowId());
+                    return new DeleteResult.Success(1L, 1L, 1L);
                 }
             }
-
             return new DeleteResult.Success(1L, 1L, 1L);
         } finally {
             LuceneStatsProvider provider = (LuceneStatsProvider) DataFormatStatsProviderRegistry.INSTANCE.get(
@@ -128,8 +127,7 @@ public class LuceneDeleteExecutionEngine implements DeleteExecutionEngine<DataFo
 
     @Override
     public void close() throws IOException {
-        // TODO: Fix this.
-
+        // Engine shutdown: close all deleters and drop tracking.
         for (Deleter deleter : generationToDeleterMap.values()) {
             deleter.close();
         }
@@ -138,18 +136,14 @@ public class LuceneDeleteExecutionEngine implements DeleteExecutionEngine<DataFo
         idToGen.clear();
     }
 
-    private Long lookupGen(String id) {
-        return idToGen.get(id);
-    }
-
     @Override
-    public void recordWrite(String id, long generation) {
-        idToGen.put(id, generation);
+    public void recordWrite(String id, long generation, long rowId) {
+        idToGen.put(id, new GenRow(generation, rowId));
     }
 
     @Override
     public boolean onWriterCheckedOut(long generation) throws IOException {
-        idToGen.entrySet().removeIf(e -> e.getValue() == generation);
+        idToGen.entrySet().removeIf(e -> e.getValue().generation() == generation);
 
         Deleter deleter = generationToDeleterMap.remove(generation);
         if (deleter == null) {
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
@@ -353,25 +353,23 @@ public class LuceneIndexingExecutionEngine implements IndexingExecutionEngine<Lu
                         }
                     }
                 }
-
-                // After addIndexes, open an NRT reader to discover the actual file names
-                // for the newly added segments. Lucene renames files during addIndexes,
-                // so the original temp directory file names are no longer valid.
+            }
+            // After addIndexes, open an NRT reader to discover the actual file names
+            // for the newly added segments. Lucene renames files during addIndexes,
+            // so the original temp directory file names are no longer valid.
+            Set<Long> liveGenerations = new HashSet<>();
+            if (sourceDirectories.isEmpty() == false || refreshInput.existingSegments().isEmpty() == false) {
                 Path sharedDir = store.shardPath().resolveIndex();
-
                 try (DirectoryReader reader = DirectoryReader.open(sharedWriter)) {
-                    List<LeafReaderContext> leaves = reader.leaves();
-
-                    for (int i = 0; i < leaves.size(); i++) {
-                        LeafReaderContext ctx = leaves.get(i);
+                    for (LeafReaderContext ctx : reader.leaves()) {
                         if (ctx.reader() instanceof SegmentReader segReader) {
                             SegmentCommitInfo segInfo = segReader.getSegmentInfo();
                             String genAttr = segInfo.info.getAttribute(LuceneWriter.WRITER_GENERATION_ATTRIBUTE);
                             if (genAttr == null) {
                                 continue;
                             }
-
                             long writerGen = Long.parseLong(genAttr);
+                            liveGenerations.add(writerGen);
                             if (!writerGenerations.contains(writerGen)) {
                                 continue;
                             }
@@ -392,7 +390,6 @@ public class LuceneIndexingExecutionEngine implements IndexingExecutionEngine<Lu
                         }
                     }
                 }
-                assert writerGenerations.isEmpty() : "Could not get segments from all writers";
             }
             assert writerGenerations.isEmpty() : "Could not get segments from all writers";
 
@@ -408,7 +405,14 @@ public class LuceneIndexingExecutionEngine implements IndexingExecutionEngine<Lu
                 }
             }
 
-            return new RefreshResult(List.copyOf(resultSegments));
+            Set<Long> droppedGenerations = new HashSet<>();
+            for (Segment segment : refreshInput.existingSegments()) {
+                if (liveGenerations.contains(segment.generation()) == false) {
+                    droppedGenerations.add(segment.generation());
+                }
+            }
+
+            return new RefreshResult(List.copyOf(resultSegments), droppedGenerations);
         } finally {
             stats.incRefreshTotal();
             stats.addRefreshTimeMillis(TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - refreshStart));
--- a/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
+++ b/sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
@@ -60,7 +60,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
 
 /**
@@ -104,6 +106,8 @@ public class LuceneWriter implements Writer<LuceneDocumentInput> {
     private final Directory directory;
     private final IndexWriter indexWriter;
     private final Set<LuceneWriter> registry;
+    /** Row ids to mark deleted during flush. */
+    private final Queue<Long> positionalDeletes = new ConcurrentLinkedQueue<>();
     private long mappingVersion;
     private volatile long docCount;
     private volatile boolean flushed;
@@ -279,6 +283,14 @@ public class LuceneWriter implements Writer<LuceneDocumentInput> {
     }
 
     /**
+     * Buffers an insertion row id to be marked deleted during this writer's flush.
+     */
+    @Override
+    public void recordPositionalDelete(long insertionRowId) {
+        positionalDeletes.add(insertionRowId);
+    }
+
+    /**
      * Force-merges all buffered documents into exactly one segment, commits the IndexWriter,
      * and returns a {@link FileInfos} describing the resulting segment files in the temp directory.
      * <p>
@@ -313,6 +325,10 @@ public class LuceneWriter implements Writer<LuceneDocumentInput> {
             );
             indexWriter.flush();
 
+            // Sort permutation from the primary (null if unsorted), used by applyPositionalDeletes to
+            // map insertion row ids to their post-reorder positions.
+            RowIdMapping mapping = null;
+
             // If sort permutation is provided, configure the reorder merge policy
             if (flushInput.hasRowIdMapping()) {
                 // RowIdMapping shouldn't be available if index has sort configurations.
@@ -326,7 +342,7 @@ public class LuceneWriter implements Writer<LuceneDocumentInput> {
                             + "]"
                     );
                 }
-                RowIdMapping mapping = flushInput.rowIdMapping();
+                mapping = flushInput.rowIdMapping();
                 if (mapping.size() != docCount) {
                     throw new IllegalStateException(
                         "RowIdMapping size ["
@@ -361,6 +377,9 @@ public class LuceneWriter implements Writer<LuceneDocumentInput> {
                 docCount,
                 forceMergeDurationMs
             );
+
+            // Apply the buffered positional deletes to the reordered segment; the commit below persists them with the reorder.
+            applyPositionalDeletes(mapping);
 
             long commitStartNanos = System.nanoTime();
             indexWriter.commit();
@@ -450,6 +469,48 @@ public class LuceneWriter implements Writer<LuceneDocumentInput> {
         if (currentCodec instanceof LuceneWriterCodec lwc) {
             lwc.enableRowIdRewrite();
         }
+    }
+
+    /**
+     * Applies the row ids buffered by {@link #recordPositionalDelete(long)} as {@code liveDocs}-only deletes: rows
+     * are marked not-live but stay physically present, keeping this segment 1:1 (offset-aligned) with Parquet.
+     * No-op when nothing was recorded.
+     *
+     * @param mapping insertion-order to post-reorder permutation for a sorted flush, or {@code null} when unsorted
+     *     (the recorded position is already the doc id)
+     * @throws IOException if opening the NRT reader or buffering a delete fails
+     * @throws IllegalStateException if the single-segment invariant is broken or a doc id cannot be resolved
+     *     ({@link IndexWriter#tryDeleteDocument} returned {@code -1})
+     */
+    private void applyPositionalDeletes(RowIdMapping mapping) throws IOException {
+        if (positionalDeletes.isEmpty()) {
+            return;
+        }
+        indexWriter.getConfig().setMergePolicy(NoMergePolicy.INSTANCE);
+        int deleted = 0;
+        try (DirectoryReader reader = DirectoryReader.open(indexWriter)) {
+            assert reader.leaves().size() == 1 : "expected exactly 1 segment after forceMerge, got " + reader.leaves().size();
+            LeafReader leaf = reader.leaves().get(0).reader();
+            Long insertionRowId;
+            while ((insertionRowId = positionalDeletes.poll()) != null) {
+                // Post-reorder position (identity when unsorted); equals the doc id (see assertRowIdsSequential).
+                long position = mapping != null ? mapping.getNewRowId(insertionRowId, RowIdMapping.SINGLE_GEN) : insertionRowId;
+                assert position >= 0 : "positional delete resolved to a negative row id (insertionRowId=" + insertionRowId + ")";
+                if (indexWriter.tryDeleteDocument(leaf, (int) position) == -1) {
+                    throw new IllegalStateException(
+                        "tryDeleteDocument failed for position="
+                            + position
+                            + " (insertionRowId="
+                            + insertionRowId
+                            + ") on writer generation ["
+                            + writerGeneration
+                            + "]; segment must not have merged under NoMergePolicy"
+                    );
+                }
+                deleted++;
+            }
+        }
+        logger.debug("applyPositionalDeletes: generation={}, marked {} row(s) deleted (liveDocs-only)", writerGeneration, deleted);
     }
 
     /**
@@ -685,10 +746,9 @@ public class LuceneWriter implements Writer<LuceneDocumentInput> {
 
         @Override
         public void setMergeInfo(SegmentCommitInfo info) {
+            // No generation stamp here — LuceneWriterCodec stamps the real one on every .si write (after
+            // setMergeInfo), so one here would be overwritten. Override kept only to guard against a bogus re-stamp.
             super.setMergeInfo(info);
-            if (info != null) {
-                info.info.putAttribute(WRITER_GENERATION_ATTRIBUTE, String.valueOf(0));
-            }
         }
     }
 
--- a/sandbox/qa/analytics-engine-coordinator/src/internalClusterTest/java/org/opensearch/analytics/cancellation/AnalyticsQueryTaskCleanupIT.java
+++ b/sandbox/qa/analytics-engine-coordinator/src/internalClusterTest/java/org/opensearch/analytics/cancellation/AnalyticsQueryTaskCleanupIT.java
@@ -31,6 +31,7 @@ import org.opensearch.composite.CompositeDataFormatPlugin;
 import org.opensearch.core.rest.RestStatus;
 import org.opensearch.core.tasks.TaskCancelledException;
 import org.opensearch.index.engine.dataformat.stub.MockCommitterEnginePlugin;
+import org.opensearch.index.engine.dataformat.stub.MockDataFormatPlugin;
 import org.opensearch.parquet.ParquetOnlyDataFormatPlugin;
 import org.opensearch.plugins.Plugin;
 import org.opensearch.plugins.PluginInfo;
@@ -95,7 +96,8 @@ public class AnalyticsQueryTaskCleanupIT extends OpenSearchIntegTestCase {
             TestPPLPlugin.class,
             CompositeDataFormatPlugin.class,
             MockTransportService.TestPlugin.class,
-            MockCommitterEnginePlugin.class
+            MockCommitterEnginePlugin.class,
+            MockDataFormatPlugin.class
         );
     }
 
--- a/server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
+++ b/server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
@@ -38,6 +38,9 @@ import org.opensearch.index.IndexModule;
 import org.opensearch.index.VersionType;
 import org.opensearch.index.engine.dataformat.DataFormat;
 import org.opensearch.index.engine.dataformat.DataFormatRegistry;
+import org.opensearch.index.engine.dataformat.DeleteExecutionEngine;
+import org.opensearch.index.engine.dataformat.DeleteInput;
+import org.opensearch.index.engine.dataformat.DeleteResult;
 import org.opensearch.index.engine.dataformat.FileInfos;
 import org.opensearch.index.engine.dataformat.FlushInput;
 import org.opensearch.index.engine.dataformat.IndexingEngineConfig;
@@ -103,7 +106,6 @@ import org.opensearch.search.suggest.completion.CompletionStats;
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -154,6 +156,7 @@ public class DataFormatAwareEngine implements Indexer {
 
     private final IndexingExecutionEngine indexingExecutionEngine;
     private final IndexingStrategyPlanner indexingStrategyPlanner;
+    private final DeletionStrategyPlanner deletionStrategyPlanner;
     private final DocumentCountTracker documentCountTracker;
     private final AtomicLong pendingRowCount = new AtomicLong();
     private final LockablePool<DefaultLockableHolder<Writer<?>>> writerPool;
@@ -163,6 +166,7 @@ public class DataFormatAwareEngine implements Indexer {
 
     private final CatalogSnapshotManager catalogSnapshotManager;
     private final Committer committer;
+    private final DeleteExecutionEngine<?> deleteExecutionEngine;
     private final List<ReferenceManager.RefreshListener> refreshListeners;
     private final CatalogSnapshotStatsCache statsCache;
 
@@ -349,6 +353,7 @@ public class DataFormatAwareEngine implements Indexer {
 
             // 5. Create IndexingExecutionEngine and ReaderManagers
             DataFormatRegistry registry = engineConfig.getDataFormatRegistry();
+            this.deleteExecutionEngine = registry.getDeleteExecutionEngine(committer);
             this.indexingExecutionEngine = registry.getIndexingEngine(
                 new IndexingEngineConfig(
                     committer,
@@ -377,6 +382,7 @@ public class DataFormatAwareEngine implements Indexer {
                 long gen = writerGenerationCounter.incrementAndGet();
                 assert gen > 0 : "writer generation must be positive but was: " + gen;
                 Writer<?> writer = indexingExecutionEngine.createWriter(new WriterConfig(gen));
+                deleteExecutionEngine.createDeleter(writer);
                 return DefaultLockableHolder.of(new RowIdAwareWriter<>(writer));
             }, LinkedList::new, Runtime.getRuntime().availableProcessors());
             // Create Reader managers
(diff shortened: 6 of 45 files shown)


alchemist51:
Can we add e2e tests for Search with deleted docs as well? Interested in all three filter shapes: parquet, lucene and combination of both. We will need tests with dynamic mappings. Also few of the datafusion queries are derived from the stats, like min, max(without delegation), how are we making sure those are handled?

thorkous:
> Can we add e2e tests for Search with deleted docs as well? Interested in all three filter shapes: parquet, lucene and combination of both. We will need tests with dynamic mappings. Also few of the datafusion queries are derived from the stats, like min, max(without delegation), how are we making sure those are handled?

The search changes are pending. These are just indexing changes. Since the PR was getting big and hence wanted to raise search changes in different PR but let me raise the search changes in this pr only

alchemist51:
> > Can we add e2e tests for Search with deleted docs as well? Interested in all three filter shapes: parquet, lucene and combination of both. We will need tests with dynamic mappings. Also few of the datafusion queries are derived from the stats, like min, max(without delegation), how are we making sure those are handled?
> 
> The search changes are pending. These are just indexing changes. Since the PR was getting big and hence wanted to raise search changes in different PR but let me raise the search changes in this pr only

We can do in a separate PR as long as we make sure the delete API is also blocked right now.

RS146BIJAY:
I guess this change has dependency on these two PRs: https://github.com/opensearch-project/OpenSearch/pull/21597 and https://github.com/opensearch-project/OpenSearch/pull/21937

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -83,29 +83,28 @@ public RefreshResult refresh(RefreshInput refreshInput) throws IOException {
     }
 
     @Override
-    public DeleteResult deleteDocument(DeleteInput deleteInput, LongFunction<Closeable> writerByGenSupplier) throws IOException {
```
RS146BIJAY, line 86: Why are we removing writerByGenSupplier here? This was passed intentionally to ensure that delete does not happen on a checked out IndexWriter and we have an active lock. Say by the time we get the previous deleter, the previous writer gets closed and synced by parent. This will cause a Already closed Exception while performing deletion on previous deleter (or IndexWriter). `writerByGenSupplier` ensure that lock is obtained first before we perform any operation on previous writer so that it does not get checked out during ongoing deletion.

thorkous: `writerByGenSupplier` was removed because in this PR we are not performing delete on the previous generation's IndexWriter at delete time and hence it is not required. Reason to apply delete at flush time was we wanted to keep 1:1 mapping with parquet for sorted index which was not getting maintained if we applied delete at delete time. 

Also there was a deadlock condition which we have tested in `testUpdateHoldingCurrentWriterDoesNotDeadlockAgainstRefresh` 

Thread 1 : updates a doc. IW1 is the writer which adds the doc and IW2 is the writer which has to delete prev gen document. 

Thread 2 : refresh() is called and in refresh() we are calling writers.checkoutAll(),  in no fixed (generation) order, where we can get a lock on IW2. 

Each side holds the exclusive holder lock the other needs, so both stall forever

thorkous:
> I guess this change has dependency on these two PRs: #21597 and #21937

Yes. I think the above two PRs must get merged before this PR can be merged.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
RS146BIJAY: @thorkous Can u separate out core changes like LuceneDeleteExecutionEngine etc from integration change inside DataFormatAwareEngine? I think anything we are modifying in the overall approach of deletes are handled needs to be reviewed separately from how deletes will be integrated with DataFormatAwareEngine (It will make things slightly easier for me to review). Also ensure that there are enough test cases in the core PR which will give us idea around how we plan to integrate core changes with DataFormatAwareEngine.

RS146BIJAY: > keep 1:1 mapping with parquet for sorted index which was not getting maintained if we applied delete at delete time

Can u elaborate on this part? Is the issue here because of rowId getting changed due to sort function call, in which case should'nt the issue be due to recordBufferedDeletes? Since Lucene will be internally able to handle sort, unable to understand how applying the delete on IndexWriter during ingestion time will cause issue and applying it during flush will fix it?

thorkous: > > keep 1:1 mapping with parquet for sorted index which was not getting maintained if we applied delete at delete time
> 
> Can u elaborate on this part? Is the issue here because of rowId getting changed due to sort function call, in which case should'nt the issue be due to recordBufferedDeletes? Unable to understand how applying the delete on IndexWriter during ingestion time caused issue here and applying it during flush will fix it?

The issue is caused by the force-merge during flush, not by the delete operation itself.

For sorted generations, flush() does not disable force-merge. Instead, it configures ReorderingMergePolicy to reorder documents into the Parquet sort order, rewrites __row_id to a sequential 0..N-1, and then executes forceMerge(1, true). Since forceMerge expunges pending deletes, any document deleted during ingestion is physically removed from the merged segment. As a result, the Lucene segment contains N-1 rows while the corresponding Parquet file still contains N rows, breaking the required 1:1 row and offset mapping between Lucene and Parquet.

Applying positional deletes after flush avoids this issue. We first allow the reorder and forceMerge(1, true) to complete with all N rows intact. Once the final document ordering is established, applyPositionalDeletes() sets NoMergePolicy.INSTANCE and issues deleteDocuments(...) using the post-sort positions. This creates only a liveDocs (.liv) update without physically removing documents. Because NoMergePolicy prevents any subsequent merge from expunging deletes, Lucene retains all N physical rows in the same order as Parquet, preserving the 1:1 mapping while deleted rows are filtered at query time through liveDocs.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -452,6 +472,32 @@ private void configureSortedMerge(RowIdMapping mapping, boolean hasRolledBack) {
...
+        if (positionalDeletes.isEmpty()) {
+            return;
+        }
+        indexWriter.getConfig().setMergePolicy(NoMergePolicy.INSTANCE);
+        List<Query> deleteQueries = new ArrayList<>(positionalDeletes.size());
+        Long insertionRowId;
+        while ((insertionRowId = positionalDeletes.poll()) != null) {
+            long position = mapping != null ? mapping.getNewRowId(insertionRowId, RowIdMapping.SINGLE_GEN) : insertionRowId;
+            assert position >= 0 : "positional delete resolved to a negative row id (insertionRowId=" + insertionRowId + ")";
+            deleteQueries.add(NumericDocValuesField.newSlowExactQuery(DocumentInput.ROW_ID_FIELD, position));
```
RS146BIJAY: DeleteByQuery will be very costly operation

thorkous: Thanks, I have updated the code changes.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -83,29 +83,28 @@ public RefreshResult refresh(RefreshInput refreshInput) throws IOException {
...
-                        Deleter deleter = generationToDeleterMap.get(previousGen);
-                        return deleter.deleteDoc(deleteInput);
-                    } finally {
-                        previousWriterLock.close();
-                    }
+            GenRow previous = idToGen.get(deleteInput.id());
+            if (previous != null) {
+                Deleter previousDeleter = generationToDeleterMap.get(previous.generation());
+                if (previousDeleter != null) {
+                    previousDeleter.recordPositionalDelete(previous.rowId());
```
RS146BIJAY, line 104: Won't sort change row id and doc id mappings? I think we may need to handle this during refresh

RS146BIJAY: On checking this further this will be handled by this line:
```
long position = mapping != null ? mapping.getNewRowId(insertionRowId, RowIdMapping.SINGLE_GEN) : insertionRowId;
```

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -452,6 +471,48 @@ private void configureSortedMerge(RowIdMapping mapping, boolean hasRolledBack) {
...
+     */
+    private void applyPositionalDeletes(RowIdMapping mapping) throws IOException {
+        if (positionalDeletes.isEmpty()) {
+            return;
+        }
+        indexWriter.getConfig().setMergePolicy(NoMergePolicy.INSTANCE);
+        int deleted = 0;
+        try (DirectoryReader reader = DirectoryReader.open(indexWriter)) {
+            assert reader.leaves().size() == 1 : "expected exactly 1 segment after forceMerge, got " + reader.leaves().size();
+            LeafReader leaf = reader.leaves().get(0).reader();
```
RS146BIJAY, line 493: Can you elaborate on why delete will be on the first segment only? Is it because we are doing a force merge before flush?

thorkous: Yes, that's the reason, the force merge happens inside flush(), right before the deletes are applied. So by the time applyPositionalDeletes does DirectoryReader.open(indexWriter), everything has already been merged into one segment.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
mgodwan, line 152: When can this be null? Should we throw if this is null?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -53,9 +51,13 @@ public class LuceneDeleteExecutionEngine implements DeleteExecutionEngine<DataFo
     private final Map<Long, Deleter> generationToDeleterMap;
     private final DataFormat dataFormat;
     private final IndexWriter parentWriter;
-    private final ConcurrentMap<String, Long> idToGen;
+    private final ConcurrentMap<String, GenRow> idToGen;
```
mgodwan, line 54: Can we expose a stat on this?

server/src/main/java/org/opensearch/index/engine/dataformat/Writer.java
mgodwan, line 139: Should this be a property of composite writer? Ideally, when LDEE calls this, it should either check for instanceof LuceneWriter or CompositeWriter only?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -138,18 +136,14 @@ public void close() throws IOException {
...
-    public void recordWrite(String id, long generation) {
-        idToGen.put(id, generation);
+    public void recordWrite(String id, long generation, long rowId) {
+        idToGen.put(id, new GenRow(generation, rowId));
     }
 
     @Override
     public boolean onWriterCheckedOut(long generation) throws IOException {
-        idToGen.entrySet().removeIf(e -> e.getValue() == generation);
+        idToGen.entrySet().removeIf(e -> e.getValue().generation() == generation);
```
mgodwan, line 146: Do we have any perf measurement on this flow at high loads? I wanted to understand the refresh lag this adds to the overall flow.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1042,6 +1222,12 @@ public void refresh(String source) throws EngineException {
                             toClose.add(pendingWriter);
                         }
 
+                        for (Closeable retiring : toClose) {
+                            if (retiring instanceof Writer<?> retiringWriter) {
+                                deletesApplied |= deleteExecutionEngine.onWriterCheckedOut(retiringWriter.generation());
```
mgodwan, line 1227: Does this need to be called when failure happens on some writers as well?

mgodwan (commented):
During indexing, there are cases where we may end up flushing a writer before other writers in the same cycle are flushed. Can you write tests for the cases when such failure modes occur and ensure there are no races where a document's consistent view is not reflected.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
RS146BIJAY: I guess number of active writers should not be that high, unless refresh is already lagging. But yeah we can validate this via some form benchmark.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
RS146BIJAY: I guess we need to remove deleters in `retireWriterIfNeeded()` and `preIndex()` as well where we are removing writers (seems like recently added code flow).

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
RS146BIJAY, line 104: I think we need to handle rollback correctly here. Say if a writer is rolled back we need to clear all the corresponding pending position deletes from the queue as well.

RS146BIJAY: Also ensure we do add failure test case scenarios on this as well.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
thorkous: This can be null when a data format doesn't have delete execution engine, like parquet only data format. We dont need to throw exception here as we would be handling that at the time of delete. deleteDocument throws IllegalArgumentException("no delete-applicable data format")

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -128,41 +148,83 @@ public DataFormat getDataFormat() {
...
+        Deleter previousDeleter = generationToDeleterMap.get(previous.generation());
+        if (previousDeleter != null) {
+            previousDeleter.recordPositionalDelete(previous.rowId());
+        }
+    }
 
-        return totalApplied > 0;
+    /** Raises the retired-generation mark, which never moves backwards. */
+    private void markCheckedOut(long generation) {
+        maxCheckedOutGeneration.accumulateAndGet(generation, Math::max);
```
rayshrey: Writers can be checkout out in any random order, not necessarily in an increasing monotonic way.
In that case how are we even relying on this `maxCheckoutOutGeneration` ?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDocumentInput.java
```diff
@@ -89,14 +98,17 @@ public Document getFinalInput() {
     public void addField(MappedFieldType fieldType, Object value) {
         Set<FieldTypeCapabilities.Capability> capabilities = fieldType.getCapabilityMap().getOrDefault(LucenePlugin.DATA_FORMAT, Set.of());
         if (capabilities.isEmpty()) {
-            // nothing to support on this format for this field.
+            // Mirror version metadata as doc values for update-time resolution.
```
rayshrey: Let's have these gated behind settings/feature flag as needed.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -53,9 +55,25 @@ public class LuceneDeleteExecutionEngine implements DeleteExecutionEngine<DataFo
...
+     * Highest generation whose writer has been checked out. Generations increase monotonically, so a
+     * delete at or below this mark arrives after its generation retired, and a higher one targets a
+     * writer that is still live.
+     */
+    private final AtomicLong maxCheckedOutGeneration = new AtomicLong(-1L);
+
+    private static final long BASE_BYTES_PER_ID_TO_GEN_ENTRY = RamUsageEstimator.HASHTABLE_RAM_BYTES_PER_ENTRY + RamUsageEstimator
+        .shallowSizeOfInstance(GenRow.class);
+    /** Running footprint of {@link #idToGen}, maintained incrementally so reads never scan the map. */
+    private final AtomicLong idToGenRamBytesUsed = new AtomicLong();
```
rayshrey: What would be the max size of the idToGen map ?
Do we really need to track it's usage ?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -97,13 +101,20 @@ public class LuceneWriter implements Writer<LuceneDocumentInput> {
...
     private final long writerGeneration;
     private final LuceneDataFormat dataFormat;
     private final LuceneShardStatsTracker stats;
     private final Path tempDirectory;
     private final Directory directory;
     private final IndexWriter indexWriter;
     private final Set<LuceneWriter> registry;
+    /** Row ids to mark deleted during flush. */
+    private final Queue<Long> positionalDeletes = new ConcurrentLinkedQueue<>();
+    private final AtomicLong positionalDeletesRamBytesUsed = new AtomicLong();
```
rayshrey: Can't we use the size of the positionalDeletes to determine the bytes used ?
Why are we maintaining a separate variable for that ?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
rayshrey: The interfaces are a bit confusing here - we have a Deleter interface and it's implementation DeleterImpl which have a `deleteDoc` method, but we are using the Writer's deleteDocument method everywhere and methods like recordPositionalDelete are also part of the Writer only now ? 

Either we need to have these methods in Deleter and call them or remove the unused methods from Deleter interface itself to avoid any confusions and maintain readability.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
rayshrey, line 316: Can we have a mechanism to check here if any deletes were done or not and then based on that branch out - old flow if no deletes otherwise new flow if deletes are there ?

Don't want to have unnecessary computations in case deletes are not happening.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDataFormat.java
```diff
@@ -63,6 +64,7 @@ public class LuceneDataFormat extends DataFormat {
         new FieldTypeCapabilities(IdFieldMapper.CONTENT_TYPE, Set.of(STORED_FIELDS, FULL_TEXT_SEARCH)),
         new FieldTypeCapabilities(SeqNoFieldMapper.CONTENT_TYPE, Set.of(COLUMNAR_STORAGE, POINT_RANGE)),
         new FieldTypeCapabilities(SeqNoFieldMapper.PRIMARY_TERM_NAME, Set.of(COLUMNAR_STORAGE)),
+        new FieldTypeCapabilities(VersionFieldMapper.CONTENT_TYPE, Set.of(COLUMNAR_STORAGE)),
```
rayshrey, line 67: Why is this required ?

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneReaderManager.java
```diff
@@ -131,28 +129,35 @@ public void afterRefresh(boolean didRefresh, CatalogSnapshot catalogSnapshot) th
     }
 
     private static Map<Long, String> buildGenerationToSegmentName(CatalogSnapshot catalogSnapshot, List<LeafReaderContext> leaves) {
```
rayshrey, line 131: Why are we doing changes here ?

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -1076,35 +1259,47 @@ public void refresh(String source) throws EngineException {
...
+                                    + " dropped "
+                                    + result.droppedGenerations().size()
                                     + " but got "
                                     + result.refreshedSegments().size();
 
+                            // Drop fully-deleted generations so the catalog stays consistent with the reader.
+                            final List<Segment> finalSegments = result.refreshedSegments()
+                                .stream()
+                                .filter(seg -> result.droppedGenerations().contains(seg.generation()) == false)
+                                .toList();
```
rayshrey, line 1312: If we have a new segment that contains all deleted documents, then that segment will never be a part of any CatalogSnapshot and will eventually not be picked up for deletion since the IndexFileDeleter will not be aware of it's existence - leading to a file leak. We need to handle these scenarios as well.

rayshrey (commented):
Can we have a benchmark comparing append only pluggable dataformat indices with the current set of changes and verify there is no regression in throughput/storage. Let's have the benchmark with a larger set such as clickbench (along with different refresh intervals as well such as 1s, 5s, 10s, 30s).

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
thorkous: Done

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
thorkous: There is no cap on max size. We need to track usage for heap size. We can't do idToGen.size() because that's O(N)

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
thorkous: Done.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Done.

server/src/main/java/org/opensearch/index/engine/dataformat/Writer.java
thorkous: Added

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
thorkous: I have removed this section.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDocumentInput.java
thorkous: Added

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
thorkous: We can't because .size() for concurrentLinkedQueue is O(N)

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
thorkous: Removed and simplified the interface.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneIndexingExecutionEngine.java
thorkous: We already have this where we check if there are deletes or not. `refreshed || deletesApplied`

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneDataFormat.java
thorkous: It makes _version land as NumericDocValues in the Lucene segment

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
thorkous: Made code changes for this and this case is handled now.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneReaderManager.java
thorkous: because of .liv files

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
```diff
@@ -815,13 +838,152 @@ private Engine.IndexResult indexIntoEngine(Engine.Index index, IndexingStrategy
...
+        if (plan.executeOpOnEngine) {
+            versionMap.putDeleteUnderLock(
+                delete.uid().bytes(),
+                new DeleteVersionValue(
+                    plan.version,
+                    delete.seqNo(),
+                    delete.primaryTerm(),
+                    engineConfig.getThreadPool().relativeTimeInMillis()
+                )
+            );
```
LantaoJin, line 935: Should we persist delete tombstones, including the version, sequence number, and primary term, and restore them on startup as `InternalEngine` does. Otherwise, a flush and translog trim followed by a restart loses the deleted ID’s history, allowing re-indexing to reset its version or bypass conflict checks.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
```diff
@@ -452,6 +491,46 @@ private void configureSortedMerge(RowIdMapping mapping, boolean hasRolledBack) {
...
 
+    /**
+     * Applies buffered insertion-row deletes as liveDocs-only deletes, preserving physical row
+     * alignment with Parquet.
+     *
+     * @param mapping insertion-to-final-doc-id mapping, or {@code null} when unsorted
+     * @throws IOException if opening the NRT reader or applying a delete fails
+     * @throws IllegalStateException if a mapped doc id cannot be resolved
+     */
+    private void applyPositionalDeletes(RowIdMapping mapping) throws IOException {
```
LantaoJin, line 504: What if the positional-delete is across all child segments? With `NoMergePolicy`, `forceMerge(1)` may leave multiple leaves, so treating a generation-wide row ID as a local doc ID in the first leaf can fail refresh and close the shard; either merge to one leaf or map each row ID to the correct leaf and local doc ID.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
RS146BIJAY: Actually this needs to happen during failures as well. I think my earlier similar comments seems to have resolved without addressing the failure handling correctly: https://github.com/opensearch-project/OpenSearch/pull/22406#discussion_r3709564201

thorkous: Hi @RS146BIJAY ,

This is not the same issue. LantaoJin’s tombstone-persistence concern is still pending and will be addressed separately.

The comment you linked was about pending positional deletes during writer rollback, and that path has already been fixed and tested:

- A positional delete is recorded only after addDoc succeeds. A failed update therefore does not supersede the previous document.
- rollbackTo(...) explicitly removes positional deletes for rolled-back rows and updates positionalDeleteCount.
- If the writer cannot be safely retired, the engine is failed and recovery replays the translog.

Failure coverage was added in testFailedUpdateDoesNotSupersedePreviousCopy, testIndexAddDocFailureReturnsFailureResult, and the Lucene writer rollback tests.
Reopening an addressed rollback issue by linking it from an unrelated pending tombstone-persistence discussion is incorrect. If you believe a failure path still leaves a pending positional delete behind, identify the exact path and explain how it bypasses the existing sequencing, rollback cleanup, and tests. Otherwise, this comment should remain scoped to LantaoJin’s pending tombstone-persistence concern.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneWriter.java
thorkous: I have addressed this and made the code changes.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
RS146BIJAY: > This is not the same issue

Yeah i think i got confused earlier.

So restart will also do a rollback on the IndexWriter which will revert to last commit, so delete post last commit will not be useful. I am just thinking if flush should persist deletes instead. Let me start a separate thread on this as this comment is unrelated.

RS146BIJAY:
Can you elaborate how does following scenario is handled here: delete followed by commit followed by a node reboot?  Lucene creates a liv file to handle deletes before merges. So if merge cannot get triggered before reboot, .liv files can be used restored for deletes.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
rayshrey: @LantaoJin @RS146BIJAY not able to clearly understand why we need to persist tombstone information.

A flush followed by a restart will have the `.liv` file persisted (containing the deleted docs prior to the flush) and the operations after the flush will anyways be replayed via translog. So after the restart we will eventually land into the same scenario as pre-restart, right ? Let me know if I am missing something.

server/src/main/java/org/opensearch/index/IndexSettings.java
```diff
@@ -1287,7 +1287,8 @@ public IndexSettings(final IndexMetadata indexMetadata, final Settings nodeSetti
         defaultSearchPipeline = scopedSettings.get(DEFAULT_SEARCH_PIPELINE);
         pluggableDataFormatEnabled = FeatureFlags.isEnabled(FeatureFlags.PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG)
             && scopedSettings.get(PLUGGABLE_DATAFORMAT_ENABLED_SETTING);
-        derivedSourceEnabled = scopedSettings.get(INDEX_DERIVED_SOURCE_SETTING) || pluggableDataFormatEnabled;
+        derivedSourceEnabled = scopedSettings.get(INDEX_DERIVED_SOURCE_SETTING)
+            || (pluggableDataFormatEnabled && scopedSettings.get(IndexMetadata.INDEX_APPEND_ONLY_ENABLED_SETTING));
```
rayshrey, line 1291: What is changing for derivedSource here ?
Is this change unintended ?

server/src/main/java/org/opensearch/index/mapper/SourceFieldMapper.java
```diff
@@ -305,6 +305,10 @@ public void preParse(ParseContext context) throws IOException {
         if (adaptedSource != null) {
             final BytesRef ref = adaptedSource.toBytesRef();
             context.doc().add(new StoredField(fieldType().name(), ref.bytes, ref.offset, ref.length));
+            if (isPluggableDataFormatFeatureEnabled(context)) {
+                // Preserve the original source because column reconstruction is not lossless.
+                context.documentInput().addField(fieldType(), ref);
+            }
```
rayshrey, line 311: Why is this needed ?
Please remove this as this will store the source field separately which we do not want for pluggable dataformat indices.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/fields/ArrowSchemaBuilder.java
```diff
@@ -109,8 +108,7 @@ private static boolean isMultiValued(Mapper mapper) {
     }
 
     private static boolean isUnsupportedMetadataField(Mapper mapper) {
-        return mapper instanceof SourceFieldMapper
-            || mapper instanceof FieldNamesFieldMapper
+        return mapper instanceof FieldNamesFieldMapper
```
rayshrey, line 111: Same comment as the stored field here - remove this.

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/fields/plugins/MetadataFieldPlugin.java
```diff
@@ -58,6 +62,14 @@ public Map<String, ParquetField> getParquetFields() {
...
                 return Set.of(STORED_FIELDS, COLUMNAR_STORAGE);
             }
+
+            /** Writes the {@link BytesRef} slice, not its potentially larger backing array. */
+            @Override
+            protected void addToGroup(MappedFieldType mappedFieldType, ManagedVSR managedVSR, Object parseValue) {
+                BytesRef ref = (BytesRef) parseValue;
+                VarBinaryVector vector = (VarBinaryVector) managedVSR.getVector(mappedFieldType.name());
+                vector.setSafe(managedVSR.getRowCount(), ref.bytes, ref.offset, ref.length);
+            }
```
rayshrey, line 72: Same - unintended ?
Please take a pass and remove all stored fields related changes that are not required in this PR.

sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/index/LuceneDeleteExecutionEngine.java
```diff
@@ -138,31 +150,64 @@ public void close() throws IOException {
...
         }
+        parentWriter.deleteDocuments(terms);
+
+        return true;
+    }
 
-        return totalApplied > 0;
+    /** Records a positional delete for the tracked previous copy, if its generation is active. */
+    private void recordPreviousPositionalDelete(String id) {
+        GenRow previous = idToGen.get(id);
```
RS146BIJAY: I am not sure if it make sense to tie idToGen needs to be tied to a generation as well? As I am just thinking if keeping idToGen global across generation can cause deletes to get lost due to stale delete.
For eg, consider the scenario:
```
1. X indexed into gen5 → recordWrite sets idToGen[X] = (gen5, 42).                                                                                                                                             
2. Refresh starts, checkoutAll() pulls gen5 out and unlocks it.                                                                                                                                                
3. Refresh's flush loop runs gen5.flush() → applyPositionalDeletes runs (nothing to apply) → commit() → close(). gen5's liveDocs are now frozen. But idToGen[X] is untouched — still (gen5, 42).               
4. Before refresh reaches onWriterCheckedOut(gen5), a delete X arrives on another thread. Pool is empty → it makes gen6.                                                                              
5. The delete calls recordPreviousPositionalDelete(X) → idToGen.get(X) → returns the stale (gen5, 42) → it routes a liveDocs delete at gen5's now-closed writer → no-op / lost.                                
6. Refresh finally runs onWriterCheckedOut(gen5), clearing idToGen[X] — but the damage (a lost positional delete) is already done, and gen5's segment is published with X live.
```

RS146BIJAY: On checking this further, this should not become an issue in data integrity as we are performing deletes on parent writer as well and even if delete is performed on stale writer, seq no for delete is never propogated to parent writer. So this should be fine.

RS146BIJAY: But one thing that still need to be fixed is positionalDeleteCount gets never decremented for the previous generation writer as recordPreviousPositionalDelete never gets called for this scenario, causing getHeapBytesUsed() to over-report. Can you ensure positionalDeleteCount is always decremented even when a delete is recorded on a writer that's already been checked out.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
RS146BIJAY: One scenario what I can think of is issue that may happens due to hard deletes when say any stale operation comes in probably due to coordinator retry. Like what will happen for a following scenario:
1. Ingestion happens for a id say "test"
2. This is followed by update which got timed out.
3. Then a delete operation comes in which got succeded.
4. Then flush happens.
5. Followed by node reboot.
6. Now node comes up, the timed out update operation is retried. Since this operation is no longer either in the version map or in the shard actual data (due to hard delete), this operation succeds here creating a new version of this document. 

With soft delete this issue would not have happened because tombstone entry for that id is already present in the shard, which return a conflict exception for that retry. But now since there is no trace of this document in the shard or in the live version map, there is no sequence this operation against.

I don't know though if saving the tombstone entry is the right approach here though. May need to think this through

thorkous:
> Can you elaborate how does following scenario is handled here: delete followed by commit followed by a node reboot? Lucene creates a liv file to handle deletes before merges. So if merge cannot get triggered before reboot, .liv files can be used restored for deletes.

Yes, this scenario is handled without requiring a merge. Lucene live-doc files are included in the committed catalog state.

For a delete against an active child writer, `LuceneWriter.flush()` applies the positional delete before calling `IndexWriter.commit()`. Lucene writes the resulting `.liv` file, and that file is included in the writer’s `WriterFileSet`.

For a delete against an already-published segment, the delete is applied to the shared Lucene writer. Refresh receives `deletesApplied=true` and opens the writer with:

```java
DirectoryReader.open(sharedWriter, true, true)
```

Here, `writeAllDeletes=true` forces Lucene to write pending live-doc changes to index files instead of retaining them only in memory. Refresh then rebuilds the Lucene `WriterFileSet` from `SegmentCommitInfo.files()`, so the new `.liv` file becomes part of the catalog snapshot.

`DataFormatAwareEngine.flush()` performs this refresh before committing. `LuceneCommitter` fsyncs all files referenced by the catalog snapshot, including `.liv`, before `IndexWriter.commit()` publishes the `segments_N` commit point.

After a node reboot, Lucene restores each segment’s deletion generation from `segments_N` and loads the corresponding `.liv` file. The restored catalog snapshot references the same file, so the delete remains visible even when no merge occurred.

If every document in a segment is deleted, Lucene drops the segment and the engine removes that generation from the catalog instead. In either case, a merge is needed only to reclaim disk space; it is not required to preserve delete visibility across a restart.

server/src/main/java/org/opensearch/index/engine/dataformat/DeleteExecutionEngine.java
```diff
@@ -58,27 +47,42 @@ public interface DeleteExecutionEngine<T extends DataFormat> extends Closeable {
...
+    /**
+     * Records that document {@code id} now lives at {@code rowId} within the active writer
+     * {@code generation}. The rowId is the insertion position (0-based) within that
+     * generation's segment.
+     *
+     * @param id         the document id
+     * @param generation the writer generation the document was written to
+     * @param rowId      the insertion row id within that generation
+     */
+    void recordWrite(String id, long generation, long rowId);
```
rayshrey: Nit: GenRow as parameter here instead of generation and rowId (and GenRow defined in core instead of lucene plugin)

server/src/main/java/org/opensearch/index/engine/dataformat/RowIdAwareWriter.java
```diff
@@ -61,6 +62,11 @@ public long docCount() {
         return rowIdCounter.get();
     }
 
+    /** Returns the last assigned insertion row ID, or {@code -1} if empty. */
+    public long lastAssignedRowId() {
+        return rowIdCounter.get() - 1;
+    }
+
```
rayshrey: Nit - Can we re-use the `docCount` method itself ?

sandbox/qa/analytics-engine-coordinator/src/internalClusterTest/java/org/opensearch/analytics/cancellation/AnalyticsQueryTaskCleanupIT.java
rayshrey, line 1: The root cause of adding MockDataPlugin to all these ITs seems to be due to DataFormatRegistry throwing exceptions when no DeleteExecutionEngine is being provided, which is wrong behaviour actually. When our index is append only, we should allow no DeleteExecutionEngine case.

Let's add a NO_OP for now and on delete related operations if our DeleteExecutionEngine is NO_OP, the throw. We shuld throw at update/delete operations instead of throwing when DFAE is initializing.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
rayshrey: Which coordinator retries are we talking about here ?
I don't think the co-ordinator node (where the bulk request lands on) does any retires - it simply batches the bulks to bulk shards and routes it to the respective shards. So the update retry we are talking about won't be hit. 

If there are retries at client level, then Engine anyways won't be able to differentiate whether that update was pre-delete or post-delete and will accept or reject the update based in whether it is an upsert or not.

Even with vanilla Lucene we would have the same behavior of the 'lost' delete we are talking about.

RS146BIJAY: > I don't think the co-ordinator node (where the bulk request lands on) does any retires.

This statement is incorrect. Currently [coordinator retries](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/replication/TransportReplicationAction.java#L1169) incase of ConnectTransportException, NodeClosedException, etc.,.

> vanilla Lucene we would have the same behavior of the 'lost' delete we are talking about.

I am not sure I understood what do you mean by "vanilla Lucene" here. I am assuming you mean InternalEngine or regular index with soft Delete enabled. InternalEngine uses soft deletes and keeps a delete tombstone (with version/seqNo/primaryTerm) in the version map / Lucene index within the soft-delete retention window. A stale retried update would be detected against that tombstone and rejected with a VersionConflictEngineException — it would not silently resurrect the doc.

RS146BIJAY:
> > Can you elaborate how does following scenario is handled here: delete followed by commit followed by a node reboot? Lucene creates a liv file to handle deletes before merges. So if merge cannot get triggered before reboot, .liv files can be used restored for deletes.
> 
> Yes, this scenario is handled without requiring a merge. Lucene live-doc files are included in the committed catalog state.
> 
> For a delete against an active child writer, `LuceneWriter.flush()` applies the positional delete before calling `IndexWriter.commit()`. Lucene writes the resulting `.liv` file, and that file is included in the writer’s `WriterFileSet`.
> 
> For a delete against an already-published segment, the delete is applied to the shared Lucene writer. Refresh receives `deletesApplied=true` and opens the writer with:
> 
> ```java
> DirectoryReader.open(sharedWriter, true, true)
> ```
> 
> Here, `writeAllDeletes=true` forces Lucene to write pending live-doc changes to index files instead of retaining them only in memory. Refresh then rebuilds the Lucene `WriterFileSet` from `SegmentCommitInfo.files()`, so the new `.liv` file becomes part of the catalog snapshot.
> 
> `DataFormatAwareEngine.flush()` performs this refresh before committing. `LuceneCommitter` fsyncs all files referenced by the catalog snapshot, including `.liv`, before `IndexWriter.commit()` publishes the `segments_N` commit point.
> 
> After a node reboot, Lucene restores each segment’s deletion generation from `segments_N` and loads the corresponding `.liv` file. The restored catalog snapshot references the same file, so the delete remains visible even when no merge occurred.
> 
> If every document in a segment is deleted, Lucene drops the segment and the engine removes that generation from the catalog instead. In either case, a merge is needed only to reclaim disk space; it is not required to preserve delete visibility across a restart.

Yeah I think this scenario is handled correctly.

server/src/main/java/org/opensearch/index/engine/DataFormatAwareEngine.java
rayshrey: >Currently [coordinator retries](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/replication/TransportReplicationAction.java#L1169) incase of ConnectTransportException, NodeClosedException, etc.

These retries are for the whole bulk shard request and not the single bulk item.

>A stale retried update would be detected against that tombstone and rejected with a VersionConflictEngineException — it would not silently resurrect the doc.

If the update never started executing, it will never be stale to begin with. The update would be stale only when it got a sequence number assigned which is older than the delete's sequence number.

I know that since we are not explicitly adding a tombstone entry, there is a possibility we might be missing something, but I want to be very clear on the reproduction steps so that we can clearly reprodice it and fix it later on.

thorkous:
Benchmark numbers : 

| | InternalEngine(Without my changes) | DFAE(with my changes and append only off)| DFAE (with my changes but append only on) |
|---|---|---|---|
| Mean throughput | 13,189 docs/s | 13,962 | **15,912** |
| p50 latency | 2,722 ms | 2,501 | **2,045** |
| p99 latency | 8,575 ms | 7,977 | 5,534 |
| Segment count | 1,707 | 1,153 | 1,142 |
| Cumulative indexing time | 244.2 min | 226.0 min | 199.8 min |
| Refresh time / count | 55.8 min / 486 | 65.3 min / 581 | 38.5 min / 575 |
| Flush time / count | 108.8 min / 378 | 68.6 min / 475 | 43.4 min / 476 |
| Young-gen GC | 55.1 s | 61.2 s | 49.1 s |

rayshrey (approved):
Thanks for the changes, LGTM
Approved.

rayshrey:
Merging this PR as it is a dependency for the merge and search support PRs for deleted docs.
Some open items that can be discussed as follow-ups:
- https://github.com/opensearch-project/OpenSearch/pull/22406#discussion_r3994378683
- https://github.com/opensearch-project/OpenSearch/pull/22406#discussion_r3996479261
- https://github.com/opensearch-project/OpenSearch/pull/22406#discussion_r3910750232

PR #22408, opened 2026-07-08
Fix negative os.cpu.percent on cgroup v2 containers

### LINKED ISSUE #19120: [BUG] (v3.2.0) `_cat/nodes` API shows negative values in cpu stats

suraj-soni-glean:
### Description

On containers that use **cgroup v2** (e.g. GKE nodes on 1.26+, which default to cgroup v2), `_cat/nodes` and `_nodes/stats` report `os.cpu.percent: -1` even though the JVM fully supports cgroup v2:

```
"os":{"cpu":{"percent":-1,"load_average":{...}}, ...}
```

`os.cpu.percent` is sourced from `com.sun.management.OperatingSystemMXBean.getCpuLoad()` (via `OsProbe.getSystemCpuPercent()` → `Probes.getLoadAndScaleToPercent`). On a container with a CPU quota (`cpu.max` = `"<quota> <period>"`), the JVM's container-aware `getCpuLoad()` takes the quota branch and reads **`/sys/fs/cgroup/cpu.stat`** (`usage_usec`, `nr_periods`) to compute the load.

**Root cause:** the bundled JDK performs that cgroup read through a plain `Files.lines(...)` with no `doPrivileged` wrapping (the old wrappers were dropped when the SecurityManager was removed in JEP 486). OpenSearch's agent-based policy (`FileInterceptor`) intercepts the file open and walks the caller protection-domain chain; with no `doPrivileged` frame to stop at, the OpenSearch frames (`OsProbe`/`Probes`) remain in the chain and must hold the `FilePermission`. `security.policy` granted `cpu.max` but **not `cpu.stat`**, so the read is denied, `getCpuLoad()` throws, `Probes.getLoadAndScaleToPercent` swallows the exception, and `os.cpu.percent` becomes `-1`.

This was verified by running `getCpuLoad()` standalone (no agent) on the same bundled JDK in the same container — it returns valid values — while the node process (agent + policy) returns `-1`.

**Fix:** grant read access to `/sys/fs/cgroup/cpu.stat`. This is the cgroup v2 sibling of the existing `cpu,cpuacct` grant used for the cgroup v1 process-CPU path.

### Related Issues
Resolves #19120

### Check List

### DIFF
--- a/server/src/main/resources/org/opensearch/bootstrap/security.policy
+++ b/server/src/main/resources/org/opensearch/bootstrap/security.policy
@@ -246,6 +246,8 @@ grant {
   permission java.io.FilePermission "/sys/fs/cgroup/cpu", "read";
   permission java.io.FilePermission "/sys/fs/cgroup/cpu/-", "read";
   permission java.io.FilePermission "/sys/fs/cgroup/cpu.max", "read";
+  // cgroup v2: read by the JVM's container-aware OperatingSystemMXBean.getCpuLoad() (usage_usec/nr_periods)
+  permission java.io.FilePermission "/sys/fs/cgroup/cpu.stat", "read";
   permission java.io.FilePermission "/sys/fs/cgroup/cpu.weight", "read";
   permission java.io.FilePermission "/sys/fs/cgroup/cpuset.cpus", "read";
   permission java.io.FilePermission "/sys/fs/cgroup/cpuset.cpus.effective", "read";


reta (approved):
Thanks a lot @suraj-soni-glean !

PR #22415, opened 2026-07-08
[DFAE] Fix memory allocation failure in write request path to fail request instead of shard

mgodwan:
Fix memory allocation failure in write request path to fail request instead of shard

### Description

### Related Issues

### Check List


Bukhtawar (approved):
Thanks lgtm

alchemist51 (approved):

sandbox/plugins/composite-engine/src/internalClusterTest/java/org/opensearch/composite/CompositeEngineParquetOutOfMemoryIT.java
```diff
@@ -0,0 +1,132 @@
...
+            .prepareCreate(INDEX_NAME)
+            .setSettings(compositeParquetIndexSettings())
+            .setMapping("f1", "type=keyword")
+            .get();
+        ensureGreen(INDEX_NAME);
+
+        // Constrain the ingest pool so hard that the first VSR buffer allocation for a single
+        // document throws an Arrow OutOfMemoryException. No writer/VSR exists yet, so no ingest
+        // memory is in use at this point and lowering the limit is safe.
+        setIngestPoolLimit(0L, 1L);
```
rayshrey, line 106: Nit - can we also disable rebalancer to avoid flakiness ?

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/ParquetWriter.java
```diff
@@ -151,7 +152,7 @@ public WriteResult addDoc(ParquetDocumentInput d) throws IOException {
             // caller-driven rollback no-ops in the VSR and restores ACTIVE.
             try {
                 vsrManager.addDocument(d);
-            } catch (MismatchedInputException e) {
+            } catch (MismatchedInputException | OutOfMemoryException e) {
```
rayshrey, line 155: While going through the rollback flow, noticed that we might enter into a corrupt state, as we are just setting row counts while rolling back our VSR. Any stale values are not cleared and if the same VSR is re-used (which it is as we change the status back to ACTIVE from PENDING_ROLLBACK after the reset the row count), values from previous document might remain (in case those fields are empty in the new doc).

Trying to validate this scenario using an IT, but since it is an existing issue, not blocking this PR - will open an issue once able to validate using IT.

rayshrey (approved):

sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/writer/ParquetWriter.java
rayshrey: Created an issue here - https://github.com/opensearch-project/OpenSearch/issues/22417

PR #22421, opened 2026-07-09
Fix NPE when creating a match-all composable index template without an inline template

### LINKED ISSUE #22420: [BUG] NPE when creating a match-all composable index template without an inline template block
### Describe the bug

Creating a composable index template with a match-all pattern (`*`) in index_patterns and without an inline `template` block results in a 500 null_pointer_exception.
Since the `template` block is optional per the API spec, this request should either succeed or fail with a 400 validation error.
Using a non-match-all pattern such as `["logs-*"]` succeeds.

### Related component

Cluster Manager

### To Reproduce

Verified on an OpenSearch 3.7.0 cluster.

```
PUT /_component_template/my-defaults
{
  "template": {
    "settings": { "index.number_of_replicas": 2 }
  }
}

PUT /_index_template/my-catch-all
{
  "index_patterns": ["*"],
  "priority": 0,
  "composed_of": ["my-defaults"]
}
```

The second request returns a 500 with the following NPE:

```
{
  "error": {
    "type": "null_pointer_exception",
    "reason": "Cannot invoke \"org.opensearch.cluster.metadata.Template.settings()\" because the return value of \"org.opensearch.cluster.metadata.ComposableIndexTemplate.template()\" is null"
  },
  "status": 500
}
```

### Expected behavior

The request should succeed. A match-all, composed_of-only template should be usable. Since the template block is optional in the OpenSearch API, the index.hidden validation for global templates should only be performed when an inline template with settings actually exists.

- Index templates guide: https://docs.opensearch.org/latest/im-plugin/index-templates/
- Create index template API: https://docs.opensearch.org/latest/api-reference/index-apis/create-index-template/

### Additional Details

**Additional context**

This bug also existed in Elasticsearch, where it was fixed and released in 7.10.3.
- Elasticsearch fix: https://github.com/elastic/elasticsearch/pull/67310

llilyy:
### Description

Fixes a `NullPointerException` when creating a composable index template with a match-all pattern (`*`) in `index_patterns` and without an inline `template` block:

```
PUT /_index_template/my-catch-all
{
  "index_patterns": ["*"],
  "priority": 0,
  "composed_of": ["my-defaults"]
}
```

```
{
  "error": {
    "type": "null_pointer_exception",
    "reason": "Cannot invoke \"org.opensearch.cluster.metadata.Template.settings()\" because the return value of \"org.opensearch.cluster.metadata.ComposableIndexTemplate.template()\" is null"
  },
  "status": 500
}
```

`Request#validate()` checked the `index.hidden` restriction for global templates by calling `indexTemplate.template().settings()` without null checks. Since the `template` block is optional, both `template()` and its `settings()` can be null, which caused the NPE. With this change, the `index.hidden` validation only runs when an inline template with settings is actually present, so a match-all, `composed_of`-only template can be created.

The same bug existed in Elasticsearch and was fixed in 7.10.3 (elastic/elasticsearch#67310).

Added unit tests covering a global template without an inline template and one with an inline template but null settings.

### Related Issues

Resolves #22420 

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/action/admin/indices/template/put/PutComposableIndexTemplateAction.java
+++ b/server/src/main/java/org/opensearch/action/admin/indices/template/put/PutComposableIndexTemplateAction.java
@@ -118,13 +118,14 @@ public class PutComposableIndexTemplateAction extends ActionType<AcknowledgedRes
             if (indexTemplate == null) {
                 validationException = addValidationError("an index template is required", validationException);
             } else {
-                if (indexTemplate.indexPatterns().stream().anyMatch(Regex::isMatchAllPattern)) {
-                    if (IndexMetadata.INDEX_HIDDEN_SETTING.exists(indexTemplate.template().settings())) {
-                        validationException = addValidationError(
-                            "global composable templates may not specify the setting " + IndexMetadata.INDEX_HIDDEN_SETTING.getKey(),
-                            validationException
-                        );
-                    }
+                if (indexTemplate.indexPatterns().stream().anyMatch(Regex::isMatchAllPattern)
+                    && indexTemplate.template() != null
+                    && indexTemplate.template().settings() != null
+                    && IndexMetadata.INDEX_HIDDEN_SETTING.exists(indexTemplate.template().settings())) {
+                    validationException = addValidationError(
+                        "global composable templates may not specify the setting " + IndexMetadata.INDEX_HIDDEN_SETTING.getKey(),
+                        validationException
+                    );
                 }
                 if (indexTemplate.priority() != null && indexTemplate.priority() < 0) {
                     validationException = addValidationError("index template priority must be >= 0", validationException);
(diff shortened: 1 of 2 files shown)


llilyy:
Hi @msfroh @gbbafna @cwperks, tagging you since you've recently reviewed changes in this area (MetadataIndexTemplateService). Gentle ping on this PR, which has been open for a while without a review.

Quick summary:

- Fixes a NullPointerException when creating a match-all composable index template without an inline template block (#22420)
- Gradle check is passing and the change includes tests
- Small, focused bug fix limited to the cluster manager path

I'd appreciate a review when you have a chance. Happy to address any feedback. Thank you!

msfroh (approved):
This seems like a very reasonable change. In particular, when I open the existing code in my IDE, it flags the possible NPE. Adding the null-checks makes a lot of sense.

Thank you, @llilyy!

PR #22423, opened 2026-07-09
Add bulkscorer script processing interface in ScriptScore query so that underline ScriptEngines can take advantage of bulk scoring.

navneet1v:
### Description
Add bulkscorer script processing interface in ScriptScore query so that underline ScriptEngines can take advantage of bulk scoring.

### Related Issues
NA

### Check List
- [X] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

### DIFF
--- a/server/src/internalClusterTest/java/org/opensearch/search/query/ScriptScoreQueryIT.java
+++ b/server/src/internalClusterTest/java/org/opensearch/search/query/ScriptScoreQueryIT.java
@@ -34,6 +34,13 @@ package org.opensearch.search.query;
 
 import com.carrotsearch.randomizedtesting.annotations.ParametersFactory;
 
+import org.apache.lucene.index.LeafReaderContext;
+import org.apache.lucene.search.BulkScorer;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.LeafCollector;
+import org.apache.lucene.search.Scorable;
+import org.apache.lucene.search.ScoreMode;
+import org.apache.lucene.util.Bits;
 import org.opensearch.OpenSearchException;
 import org.opensearch.action.admin.cluster.settings.ClusterUpdateSettingsRequest;
 import org.opensearch.action.search.SearchResponse;
@@ -42,20 +49,28 @@ import org.opensearch.index.fielddata.ScriptDocValues;
 import org.opensearch.index.query.QueryBuilder;
 import org.opensearch.index.query.RangeQueryBuilder;
 import org.opensearch.plugins.Plugin;
+import org.opensearch.plugins.ScriptPlugin;
 import org.opensearch.script.MockScriptPlugin;
+import org.opensearch.script.ScoreScript;
 import org.opensearch.script.Script;
+import org.opensearch.script.ScriptContext;
+import org.opensearch.script.ScriptEngine;
 import org.opensearch.script.ScriptType;
+import org.opensearch.search.lookup.SearchLookup;
 import org.opensearch.test.ParameterizedStaticSettingsOpenSearchIntegTestCase;
 
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Function;
 
 import static org.opensearch.index.query.QueryBuilders.boolQuery;
 import static org.opensearch.index.query.QueryBuilders.idsQuery;
+import static org.opensearch.index.query.QueryBuilders.matchAllQuery;
 import static org.opensearch.index.query.QueryBuilders.matchQuery;
 import static org.opensearch.index.query.QueryBuilders.scriptScoreQuery;
 import static org.opensearch.search.SearchService.CLUSTER_CONCURRENT_SEGMENT_SEARCH_SETTING;
@@ -83,7 +98,7 @@ public class ScriptScoreQueryIT extends ParameterizedStaticSettingsOpenSearchInt
 
     @Override
     protected Collection<Class<? extends Plugin>> nodePlugins() {
-        return Collections.singleton(CustomScriptPlugin.class);
+        return Arrays.asList(CustomScriptPlugin.class, BulkScoringScriptPlugin.class);
     }
 
     public static class CustomScriptPlugin extends MockScriptPlugin {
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/opensearch/common/lucene/search/function/ScriptScoreQuery.java
```diff
@@ -137,6 +137,15 @@ public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOExcepti
...
                             final BulkScorer subQueryBulkScorer = subQueryScorerSupplier.bulkScorer();
+                            if (subQueryBulkScorer == null) {
+                                return null;
+                            }
+                            if (scriptBuilder instanceof ScoreScript.BulkScoringLeafFactory bulkFactory) {
+                                final BulkScorer custom = bulkFactory.bulkScorer(context, subQueryBulkScorer, subQueryScoreMode, boost);
+                                if (custom != null) {
+                                    return custom;
+                                }
+                            }
```
owaiskazi19, line 148: [explain](https://github.com/opensearch-project/OpenSearch/pull/22423/files#diff-9d4b4f755d23235ec5bd75b1a32ced38e6421271561522097f09c35fc1124112R180) on L180 still shows the explanation through per doc. We should add the custom bulk scorer in the explanation as well

server/src/internalClusterTest/java/org/opensearch/search/query/ScriptScoreQueryIT.java
```diff
@@ -238,4 +253,212 @@ public void testScriptScoreNotExistingQuery() throws Exception {
...
+        SearchResponse resp = client().prepareSearch("bulk-test-minscore")
+            .setQuery(scriptScoreQuery(matchAllQuery(), script).setMinScore(1.5f))
+            .get();
+        assertNoFailures(resp);
+        assertEquals(2, resp.getHits().getTotalHits().value());
+        for (var hit : resp.getHits().getHits()) {
+            assertEquals(2.0f, hit.getScore(), 0.001f);
+        }
+    }
+
```
owaiskazi19, line 301: We should add one IT with explain path as well

server/src/internalClusterTest/java/org/opensearch/search/query/ScriptScoreQueryIT.java
```diff
@@ -238,4 +253,212 @@ public void testScriptScoreNotExistingQuery() throws Exception {
...
+                public double execute(ExplanationHolder explanation) {
+                    return 2.0;
+                }
+            };
+        }
+
+        @Override
+        public BulkScorer bulkScorer(LeafReaderContext context, BulkScorer subQueryBulkScorer, ScoreMode subQueryScoreMode, float boost)
+            throws IOException {
+            if (!useBulk) {
```
owaiskazi19: nit: OpenSearch style
```suggestion
            if (useBulk == false) {
```

owaiskazi19 (commented):
Overall LG with few comments and from the PR reviewer should be addressed

server/src/internalClusterTest/java/org/opensearch/search/query/ScriptScoreQueryIT.java
navneet1v: will fix it.

server/src/main/java/org/opensearch/common/lucene/search/function/ScriptScoreQuery.java
navneet1v: why we would need to do that? BulkScoring explanation should be same as that of single score explanation. Trying to understand why we need a change?

owaiskazi19: Agreed, if the bulk explanation should equal the single-score explanation, then per-doc explain is fine and we don't need to touch explain(). My only ask is that we make that an explicit contract, because right now nothing enforces it and the IT actually models the two paths diverging (newInstance() returns 2.0, bulk returns (doc+1)*boost). Could we:
Add a line to the `BulkScoringLeafFactory` javadoc stating that implementers must keep `newInstance().execute()` consistent with their bulk scorer, otherwise _explain will disagree with _score

navneet1v: @owaiskazi19 updated the java docs. Let me know if that helps or something else is needed.

owaiskazi19:
> [**Behavior Change**](https://github.com/opensearch-project/OpenSearch/pull/22423/files#diff-9d4b4f755d23235ec5bd75b1a32ced38e6421271561522097f09c35fc1124112R140-R142)
> 
> The added `if (subQueryBulkScorer == null) return null;` short-circuits the previous behavior where `ScriptScoreBulkScorer` would be constructed regardless. While the underlying `ScriptScoreBulkScorer` likely cannot function without a sub-query bulk scorer, this early null return changes the contract of `bulkScorer()` in the null case. Confirm this does not regress behavior when sub-query has no matches for a segment but the caller expected a non-null (empty) BulkScorer.
> 
> [**Contract Inconsistency**](https://github.com/opensearch-project/OpenSearch/pull/22423/files#diff-b460a929836e49463eacd268c32075f4659d3f32d08f48db45ee9632c53b003bR284-R293)
> 
> The javadoc states the returned BulkScorer "fully replaces" the default and is responsible for score validation (non-negative, non-NaN) and applying boost. However, the caller in `ScriptScoreQuery` does not enforce this—if a script engine returns a BulkScorer that emits negative or NaN scores, they will propagate silently, unlike the default `ScriptScoreBulkScorer` path which validates scores. Consider either wrapping the custom BulkScorer with score validation or documenting this as a caller responsibility more prominently.

@navneet1v can we also address these especially the second one since validation is needed for negative/Nan scores?

navneet1v:
> > [**Behavior Change**](https://github.com/opensearch-project/OpenSearch/pull/22423/files#diff-9d4b4f755d23235ec5bd75b1a32ced38e6421271561522097f09c35fc1124112R140-R142)
> > The added `if (subQueryBulkScorer == null) return null;` short-circuits the previous behavior where `ScriptScoreBulkScorer` would be constructed regardless. While the underlying `ScriptScoreBulkScorer` likely cannot function without a sub-query bulk scorer, this early null return changes the contract of `bulkScorer()` in the null case. Confirm this does not regress behavior when sub-query has no matches for a segment but the caller expected a non-null (empty) BulkScorer.
> > [**Contract Inconsistency**](https://github.com/opensearch-project/OpenSearch/pull/22423/files#diff-b460a929836e49463eacd268c32075f4659d3f32d08f48db45ee9632c53b003bR284-R293)
> > The javadoc states the returned BulkScorer "fully replaces" the default and is responsible for score validation (non-negative, non-NaN) and applying boost. However, the caller in `ScriptScoreQuery` does not enforce this—if a script engine returns a BulkScorer that emits negative or NaN scores, they will propagate silently, unlike the default `ScriptScoreBulkScorer` path which validates scores. Consider either wrapping the custom BulkScorer with score validation or documenting this as a caller responsibility more prominently.
> 
> @navneet1v can we also address these especially the second one since validation is needed for negative/Nan scores?

I am pretty surprised that we have such kind of validations. Since who ever is generating the scores in this case BulkScorer should be doing this validation. It was fine when docs were getting scored one at a time. But since all the responsibility of score creation is going in the hands of ScriptEngine that creates the Score using BulkScorer so we don't need those validations now.

owaiskazi19:
> BulkScorer should be doing this validation. 

If plugin owns the validation then that should be enough but if validation is missed there then there would be no backstop. Fine for now!

owaiskazi19 (approved):
Thanks for addressing the comments

reta:
@navneet1v we have new flaky tests introduced by this change:

```
java.lang.AssertionError: expected:<3.0> but was:<165.0>
	at __randomizedtesting.SeedInfo.seed([714BC37BA2B32A8F:645B4EDC09AB107F]:0)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:577)
	at org.junit.Assert.assertEquals(Assert.java:701)
	at org.opensearch.search.query.ScriptScoreQueryIT.testBulkScoringWithBoost(ScriptScoreQueryIT.java:324)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1750)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:938)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:974)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:988)
	at org.opensearch.test.OpenSearchTestClusterRule$1.evaluate(OpenSearchTestClusterRule.java:369)
```

See please https://build.ci.opensearch.org/job/gradle-check/81991/, could you please take a look? Thank you

navneet1v:
> @navneet1v we have new flaky tests introduced by this change:
> 
> ```
> java.lang.AssertionError: expected:<3.0> but was:<165.0>
> 	at __randomizedtesting.SeedInfo.seed([714BC37BA2B32A8F:645B4EDC09AB107F]:0)
> 	at org.junit.Assert.fail(Assert.java:89)
> 	at org.junit.Assert.failNotEquals(Assert.java:835)
> 	at org.junit.Assert.assertEquals(Assert.java:577)
> 	at org.junit.Assert.assertEquals(Assert.java:701)
> 	at org.opensearch.search.query.ScriptScoreQueryIT.testBulkScoringWithBoost(ScriptScoreQueryIT.java:324)
> 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
> 	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
> 	at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1750)
> 	at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:938)
> 	at com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:974)
> 	at com.carrotsearch.randomizedtesting.RandomizedRunner$10.evaluate(RandomizedRunner.java:988)
> 	at org.opensearch.test.OpenSearchTestClusterRule$1.evaluate(OpenSearchTestClusterRule.java:369)
> 	at com.carrotsearch.randomizedtesting.rules.StatementAdapter.evaluate(StatementAdapter.java:36)
> 	at org.junit.rules.RunRules.evaluate(RunRules.java:20)
> ```
> 
> See please https://build.ci.opensearch.org/job/gradle-check/81991/, could you please take a look? Thank you

Will check on this.

andrross:
@navneet1v Still not fixed: https://build.ci.opensearch.org/job/gradle-check/82263/

I think the test is sensitive to segment topology.

PR #17178, opened 2025-01-29
Gradle plugin support for dependencies in plugin descriptor properties

abseth-amzn:
### Description
https://github.com/opensearch-project/OpenSearch/pull/11441 added support for semver ranges in dependencies of plugins. This commit updates the `opensearchplugin` (from `org.opensearch.gradle:build-tools`) to support `dependencies` property when building plugins. Without this support, plugin's plugin-descriptor.properties file would require manual update to add `dependencies` property. With this support, it would be possible to build plugins with semver range support for opensearch using gradle code. Example usage:

```
opensearchplugin {
  description = '<Plugin description>'
  classname = '<Plugin class name>'
  dependencies = '{opensearch: "3.0.0"}'
}
```
Since no plugin is currently using `dependencies` property, this change would simply add empty `dependencies` property to plugin-descriptor.properties file. Example:
```
description=The Mapper Murmur3 plugin allows to compute hashes of a field's values at index-time and to store them in the index.
version=3.0.0-SNAPSHOT
name=mapper-murmur3
classname=org.opensearch.plugin.mapper.MapperMurmur3Plugin
java.version=21
opensearch.version=3.0.0
custom.foldername=
extended.plugins=
has.native.controller=false
dependencies=
```

Note - before this commit, opensearch process would reject above mentioned plugin-descriptor.properties file since it contains both `opensearch.version` and `dependencies` property. With this commit, PluginInfo is being updated to allow files like above since one of these properties is empty. 
Hence this is a breaking change in opensearchplugin. The plugin-descriptor.properties file created by this plugin would not be accepted by older opensearch versions.

Following is an example of plugin-descriptor.properties file with `dependencies` property specified by the plugin in build.gradle (note empty `opensearch.version` property):

```
description=The Mapper Murmur3 plugin allows to compute hashes of a field's values at index-time and to store them in the index.
version=3.0.0-SNAPSHOT
name=mapper-murmur3
classname=org.opensearch.plugin.mapper.MapperMurmur3Plugin
java.version=21
opensearch.version=
custom.foldername=
extended.plugins=
has.native.controller=false
dependencies={opensearch: "3.0.0"}
```
### Related Issues
Resolves #13187 

### Check List
- [X] Functionality includes testing.
- [ ] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md), if applicable.
[...]

PR #11441, opened 2023-12-04
Add support for dependencies in plugin descriptor properties with semver range 

abseth-amzn:
### Description
As a first step towards supporting semVer range of compatible versions for plugins, we are starting with support for specifying 'dependencies' for a plugin in its plugin-descriptor.properties file. The change currently supports specification of only one dependency (opensearch). Following notations are supported for specifying the compatible dependency version:
* dependencies={ opensearch: x.y.z }
* dependencies={ opensearch: =x.y.z }
* dependencies={ opensearch: ~x.y.z } 
(specification of major, minor and patch version is required in above notations)

### Related Issues
- Resolves #1707 

### Check List
- [X] New functionality includes testing.
  - [X] All tests pass
- [X] New functionality has been documented.
  - [X] New functionality has javadoc added
- [ ] Failing checks are inspected and point to the corresponding known issue(s) (See: [Troubleshooting Failing Builds](../blob/main/CONTRIBUTING.md#troubleshooting-failing-builds))
- [X] Commits are signed per the DCO using --signoff
- [x] Commit changes are listed out in CHANGELOG.md file (See: [Changelog](../blob/main/CONTRIBUTING.md#changelog))
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #19680, opened 2025-10-17
Allow plugins to declare core version ranges

cwperks:
### Description

OpenSearch already supports semantic version ranges in the plugin descriptor `dependencies` field through #11441. This PR adds an opt-in Gradle DSL property so plugin authors can generate that metadata without maintaining `plugin-descriptor.properties` manually:

```gradle
opensearchplugin {
    name = "opensearch-security"
    description = "Provide access control related features for OpenSearch"
    classname = "org.opensearch.security.OpenSearchSecurityPlugin"
    opensearchVersionRange = "^3.7.0"
}
```

When `opensearchVersionRange` is configured, the generated descriptor contains:

```properties
dependencies={ opensearch: "^3.7.0" }
```

Otherwise it retains the existing exact-version descriptor:

```properties
opensearch.version=3.9.0
```

The generator deliberately emits exactly one compatibility property. This lets a range-based plugin ZIP load on existing OpenSearch versions that understand `dependencies` but reject descriptors containing both property names, even when the inactive property is blank. It also leaves descriptors unchanged for plugins that do not opt in.

The declared range states the plugin author's compatibility contract; it does not itself make incompatible plugin APIs binary compatible.

This overlaps with #22426 in adding Gradle support for the existing `dependencies` metadata. This PR is narrower and focused specifically on core-version ranges and preserving compatibility with older cores through one-of descriptor generation.

### Related Issues

Resolves https://github.com/opensearch-project/opensearch-build/issues/5720#issuecomment-3324591582

### Validation

- `./gradlew :build-tools:test --tests org.opensearch.gradle.plugin.PluginPropertiesExtensionTests`
- `./gradlew :build-tools:precommit`
- Generated the `analysis-common` descriptor with `opensearchVersionRange = "^3.7.0"` and verified it contains `dependencies` without `opensearch.version`
- Generated the default `analysis-common` descriptor and verified it contains `opensearch.version` without `dependencies`

### Check List

- [x] Functionality includes testing.
- [ ] API changes companion pull request created, if applicable.
- [ ] Public documentation issue/PR created, if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #22426, opened 2026-07-09
Gradle plugin support for dependencies in plugin descriptor properties

### LINKED ISSUE #13187: [Feature Request] missing gradle plugin support for dependencies in plugin descriptor properties with semver range

#11441 added support for semver ranges in dependencies of plugins. however, there's no way to use this with the gradle plugin used to build plugins (`opensearchplugin` from `org.opensearch.gradle:build-tools`).

* it must be possible to specify this new kind of dependencies also using the gradle plugin. i.e. it should be easily possible to just say "this OpenSearch plugin is compatible with OpenSearch >= 2.13" using gradle code without having to manually write the plugin descriptor file (since the gradle plugin is supposed to take care of it).
* [opensearch-plugin-template-java](https://github.com/opensearch-project/opensearch-plugin-template-java) should be updated to use this and show it as best practices.
* other OpenSearch plugins may or may not be updated to match this (depends on whether it makes sense for them - though for patch releases it might make sense so that in the future you're not forced to re-release each and every one of them even if nothing changed there?)

### Related component

Plugins

the plugin descriptor file could be written manually, but then a lot of the automation coming from the gradle plugin would no longer be used, which makes no sense.

### Additional context

see https://github.com/opensearch-project/OpenSearch/pull/11441#issuecomment-2042814075 and following comments

rursprung:
### Description
this is #17178 rebased and split in two commits, with better commit messages. all credits for the actual commit content go to @abseth-amzn (who doesn't seem to be working on OpenSearch anymore, thus the original PR has been abandoned for a long time).

the rest is copy & paste from the commit message:
PR #11441 added support for semver ranges in dependencies of plugins.
this now updates `opensearchplugin` (from
`org.opensearch.gradle:build-tools`) to support `dependencies` property
when building plugins. since most plugins do not create the
`plugin-descriptor.properties` file manually but instead use this gradle
plugin for it this means that so far `dependencies` was not really used.

`dependencies` also allows depending on `opensearch` (i.e. the core) and
it allows depending on a semver range - thus this finally allows
decoupling plugin releases from OpenSearch releases, i.e. plugins using
this instead of specifying an `opensearch.version` do not need a new
release for every new OpenSearch minor/patch release.

for existing plugins updating to the new gradle plugin version will just
generate an empty `dependencies` entry. so far, this would have then
caused the plugin to fail to load, however the previous commit has
modified this behaviour.
an alternative would be to not generate the entry at all if no
dependencies are specified, but then `dependencies` would be handled
differently to all other entries in the properties file. since the
gradle plugin is generally only updated together with the OpenSearch
version which the plugins support this should be ok to do (it's unlikely
that a plugin will use the gradle plugin 3.8.0 and generate a plugin
release for e.g. 3.7.0).

---

plugin info: support empty `dependencies`

so far, if a plugin provides a `plugin-descriptor.properties` with an
empty `dependencies` entry this will fail if `opensearch.version` is
also specified. however, an empty `dependencies` entry should not fail
this - it should behave as if it were not present.

### Related Issues
resolves #13187
related to #11441

### Check List

### DIFF
--- a/buildSrc/src/main/groovy/org/opensearch/gradle/plugin/PluginBuildPlugin.groovy
+++ b/buildSrc/src/main/groovy/org/opensearch/gradle/plugin/PluginBuildPlugin.groovy
@@ -113,8 +113,15 @@ class PluginBuildPlugin implements Plugin<Project> {
                     'customFolderName'    : extension1.customFolderName,
                     'extendedPlugins'     : extension1.extendedPlugins.join(','),
                     'hasNativeController' : extension1.hasNativeController,
-                    'requiresKeystore'    : extension1.requiresKeystore
+                    'requiresKeystore'    : extension1.requiresKeystore,
+                    'dependencies'        : extension1.dependencies
             ]
+
+            // Clear opensearch version if dependencies are specified
+            if (extension1.dependencies != null && !extension1.dependencies.trim().isEmpty()) {
+                properties.put('opensearchVersion', '');
+            }
+
             project.tasks.named('pluginProperties').configure {
                 expand(properties)
                 inputs.properties(properties)
--- a/buildSrc/src/main/java/org/opensearch/gradle/plugin/PluginPropertiesExtension.java
+++ b/buildSrc/src/main/java/org/opensearch/gradle/plugin/PluginPropertiesExtension.java
@@ -53,6 +53,8 @@ public class PluginPropertiesExtension {
 
     private String customFolderName = "";
 
+    private String dependencies = "";
+
     /** Other plugins this plugin extends through SPI */
     private List<String> extendedPlugins = new ArrayList<>();
 
@@ -84,6 +86,14 @@ public class PluginPropertiesExtension {
 
     public void setCustomFolderName(String customFolderName) {
         this.customFolderName = customFolderName;
+    }
+
+    public String getDependencies() {
+        return dependencies;
+    }
+
+    public void setDependencies(String dependencies) {
+        this.dependencies = dependencies;
     }
 
     public String getName() {
--- a/buildSrc/src/main/resources/plugin-descriptor.properties
+++ b/buildSrc/src/main/resources/plugin-descriptor.properties
@@ -59,3 +59,6 @@ extended.plugins=${extendedPlugins}
 #
 # 'has.native.controller': whether or not the plugin has a native controller
 has.native.controller=${hasNativeController}
+#
+# 'dependencies': any dependencies specified by the plugin
+dependencies=${dependencies}
--- a/buildSrc/src/test/java/org/opensearch/gradle/plugin/PluginBuildPluginTests.java
+++ b/buildSrc/src/test/java/org/opensearch/gradle/plugin/PluginBuildPluginTests.java
@@ -39,8 +39,8 @@ import org.gradle.api.internal.project.ProjectInternal;
 import org.gradle.api.tasks.bundling.AbstractArchiveTask;
 import org.gradle.testfixtures.ProjectBuilder;
 import org.junit.Before;
-import org.junit.Ignore;
 
+import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.mockito.Mockito;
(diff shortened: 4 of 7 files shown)


cwperks:
Thank you for this PR @rursprung ! I have something similar open at https://github.com/opensearch-project/OpenSearch/pull/19680 and I think between the 2 of these they can help ease the maintenance burden on plugins.

cwperks:
@rursprung I will try to go through this in earnest soon bc I've been trying to push similar changes for some time. Please see the following bot comments as well

## My take

I like the goal and the implementation is fairly small, but **I would request changes before approving it**. There is one compatibility problem that cuts directly against the purpose of introducing semantic-version ranges.

I reviewed the current head, `f013b87`, covering 7 files with 165 additions and 9 deletions.

### 1. Blocking: the generated descriptor should omit the inactive property

The Gradle plugin always generates both of these lines:

```properties
opensearch.version=${opensearchVersion}
dependencies=${dependencies}
```

When `dependencies` is configured, the code changes `opensearchVersion` to an empty string rather than removing that property. The resulting descriptor is therefore:

```properties
opensearch.version=
dependencies={opensearch:"~3.8.0"}
```

That behavior is visible in the template and in the branch that clears the version.

The problem is that semantic-range support originally shipped in OpenSearch 2.13.0, but the 2.13 parser—and subsequent versions before this PR—rejects a descriptor whenever both property names are present. It checks for `null`, not blankness, so an empty value still counts as a supplied property.

Consequently, a plugin generated by this PR with a range such as:

```properties
dependencies={opensearch:"^3.4.0"}
```

would not install on 3.4 through whatever release first contains the new `isBlank` behavior, even though those versions are inside its declared range. It would only work from the release containing this PR forward.

I would generate **exactly one** of the two properties:

```properties
# Ordinary plugin
opensearch.version=3.8.0

# Range-based plugin
dependencies={opensearch:"^3.4.0"}
```

That preserves compatibility with every runtime that already understands `dependencies`. It also avoids adding `dependencies=` to ordinary descriptors consumed by older OpenSearch versions or external tooling.

The `PluginInfo.isBlank` change can remain as defensive leniency for manually authored descriptors, but the Gradle generator should not rely on it. Alternatively, that first commit could be dropped entirely once the generator cleanly implements the one-of relationship.

### 2. Strongly recommended: test the generated file, not just task registration

The new Gradle test only checks that the `pluginProperties` task has an input named `dependencies`. It does not set a range, execute the generation task, or inspect the generated descriptor.

The runtime tests are good: they cover both blank, one blank and one populated, and both populated.  The CLI fixtures also manually construct descriptors containing the blank alternate property.  But none of those proves that the Gradle plugin produces the intended descriptor.

I would add two generation tests:

1. With no `dependencies`, assert that the output contains a populated `opensearch.version` and **does not contain** a `dependencies` property.
2. With `dependencies` configured, assert that the output contains the exact range and **does not contain** `opensearch.version`.

Ideally, feed the generated descriptor into `PluginInfo.readFromProperties` as an integration assertion. Those tests would also cover the currently untested setter and generation branch. Codecov’s 60% patch coverage and four uncovered changed lines are consistent with this gap.

### 3. Recommended: document the public Gradle API

The new API requires plugin authors to know and manually serialize an implementation-specific value such as:

```groovy
opensearchplugin {
    dependencies = '{opensearch: "^3.4.0"}'
}
```

There is no Javadoc, DSL example, or linked plugin-template/documentation update. The issue being resolved explicitly asked for the Java plugin template to demonstrate the feature. Also, the descriptor comment says “any dependencies,” while the parser currently allows exactly one entry and requires that entry to be `opensearch`.

At minimum, I would add a documented example and clarify the accepted format. A nicer API would expose an OpenSearch version-range property or typed map and have the build plugin handle serialization, but that is an API-design improvement rather than a blocker.

## One bot concern I would **not** block on

The automated review suggested warning because `opensearchVersion` is silently cleared. I do not think that is materially problematic in the current design: `opensearchVersion` is not supplied through `PluginPropertiesExtension`; it is derived internally from `VersionProperties`. Selecting `dependencies` is therefore intentionally selecting the alternative representation, rather than overriding a separately configured user value.

PR #22432, opened 2026-07-10
Fix query cache heap leak from wildcard field queries

### LINKED ISSUE #22419: [BUG] WildcardMatchingQuery causes query cache heap exhaustion via SearchLookup pinning
#### Summary

`WildcardMatchingQuery` stores a `SearchLookup` (per-request state) as a field. When the query is cached by the node-level LRU query cache, the `SearchLookup` — and everything it transitively retains — is pinned for the lifetime of the cache entry. Combined with severely broken size accounting, this leads to unbounded heap growth.

Observed: 26.2 GB retained by the query cache on a 30.5 GB heap (96%).

#### Impact

Any workload with moderate wildcard query cardinality against a `wildcard` field type will eventually exhaust heap. The cache never self-corrects because it believes it's within budget.

<details><summary><b>Root Cause</b></summary>
<p>

`WildcardMatchingQuery` captures `SearchLookup` at construction time:

```java
// WildcardFieldMapper.java:760
this.searchLookup = context.lookup();
```

The `SearchLookup.fieldDataLookup` lambda implicitly captures the entire `QueryShardContext`, which holds the `IndexSearcher` current at request time. When the query enters the cache as a map key:

```
LRU cache key → WildcardMatchingQuery → SearchLookup → lambda → QueryShardContext
    → IndexSearcher → DirectoryReader → SegmentReader[] (20-29 per shard)
```

As segments merge, old DirectoryReaders become unreachable from the active shard — but remain pinned through the cache. GC cannot collect them.

**Size accounting:** `WildcardMatchingQuery.visit()` only delegates to `firstPhaseQuery.visit()`, hiding `searchLookup`, `valueFetcher`, and `secondPhaseMatcher` from the `RamUsageQueryVisitor`. The cache estimates ~500 bytes per query; actual retained size is 8-17 MB.

</p>
</details> 

<details><summary>Heap Dump Screenshot</summary>
<p>

Image 

Image 

</p>
</details> 

### Related component

Search

### To Reproduce

- Use a `wildcard` field type
- Run many distinct wildcard patterns (e.g. `*foo*`, `*bar*`, `*baz*`...)
- Observe query cache `ramBytesUsed` stays well below budget while actual heap grows unboundedly
- Eventually OOM or severe GC pressure

## Proposed Fix

<details><summary>Details</summary>
<p>

### 1. Prevent caching (short-term, safe)

```java
// WildcardFieldMapper.java — in createWeight() ConstantScoreWeight
@Override
public boolean isCacheable(LeafReaderContext leafReaderContext) {
    return searchLookup == null;
}
```

Production queries always have `searchLookup != null` and will not be cached. This eliminates both the pinning and the size accounting problem.

**Precedent:** `DerivedFieldQuery` (same pattern — holds SearchLookup) already returns `isCacheable = false`.

### 2. Cache phase 1 independently (performance recovery)

The current code calls `firstPhaseQuery.createWeight(searcher, ...)` directly, bypassing the cache layer. Changing to:

```java
Weight firstPhaseWeight = searcher.createWeight(firstPhaseQuery, scoreMode, boost);
```

routes the inner BooleanQuery (ngram intersection) through `IndexSearcher.createWeight()` → `queryCache.doCache()`. This BooleanQuery holds only TermQuery clauses — no hidden state, correctly sized, safely cacheable.

Result: the expensive postings intersection is cached; only the cheap source verification re-runs per request.

### 3. Remove SearchLookup from Query fields (ideal long-term)

Resolve `SearchLookup` at `createWeight()`/`scorerSupplier()` time instead of storing it as a Query field. This makes the Query lightweight and safely cacheable as a whole — bitset caching works correctly since segments are immutable.

Requires API to obtain `SearchLookup` from the searcher at weight-creation time.

</p>
</details> 

### Additional Details

- OpenSearch 3.3
- Latent since wildcard field type introduction (not a regression)
- Lucene [15097](https://github.com/apache/lucene/pull/15097) / PR [15124](https://github.com/apache/lucene/pull/15124) (already on 3.3) does not help in this case

Adam-Horse:
### Description
Fixes a heap exhaustion caused by `WildcardMatchingQuery` (the two-phase query behind the `wildcard` field type) being admitted to the LRU query cache while holding shard-scoped state.

When built from a `QueryShardContext`, the query captures the shard's `SearchLookup` and a `ValueFetcher` supplier for second-phase verification, but its weight unconditionally reported `isCacheable() == true`. Cache entries outlive the search, so each cached wildcard query pinned its captured object graph on the heap. Because `equals`/`hashCode` ignore the captured lookup, every unique pattern pinned a distinct graph, and the cache's RAM accounting (via `Query#visit`, which only sees the first-phase terms) missed the pinned state entirely, so memory-based eviction never triggered. Under steady wildcard query traffic this accumulates until the node runs out of heap.

Changes:
- The weight now reports `isCacheable() == false` whenever a `SearchLookup` was captured, keeping the stateful two-phase query out of the cache.
- The first-phase trigram query is created through `searcher.createWeight(firstPhaseQuery, ScoreMode.COMPLETE_NO_SCORES, 1f)` instead of `firstPhaseQuery.createWeight(...)`, so the cheap, stateless part of the query remains independently cacheable (its scores were never used; the outer weight is constant-score).

Tests:
- `testSearchLookupNotPinnedByQueryCache` — regression test for the reported leak: runs a context-bound query through an `LRUQueryCache`, retains only a `WeakReference` to the captured `SearchLookup`, releases the query, and asserts the reference clears under GC. On `main` it fails with `SearchLookup is still strongly reachable: the query cache is pinning shard state`; with this fix it passes.
- `testCacheabilityDependsOnSearchLookup` — the weight is cacheable without a `QueryShardContext` and not cacheable with one.
- `testFirstPhaseQueryCachedIndependently` — the first-phase query still enters the cache and serves hits, while the outer `WildcardMatchingQuery` never does.

Verified on a live node built from source (1 GB heap, 20k-doc single-segment `wildcard` index, 2000 unique patterns x 6 repetitions in filter context with `request_cache=false`, then forced GC + `jcmd GC.class_histogram`):

| After GC | main | with fix |
|---|---|---|
| Retained `SearchLookup` / `QueryShardContext` | 4000 / 2000 | 0 / 0 |
| Heap used | 134 MB | 47 MB (baseline) |
| Query cache entries / hits | 2000 / 0 | 2000 / 4000 |

On `main` the cache self-reported ~2.4 MB while actually pinning ~88 MB, even with a near-empty mapping — consistent with the tens-of-GB retention reported in the issue at production scale.

A follow-up restructuring `WildcardMatchingQuery` so the second phase does not capture shard-scoped state at all (making the full query safely cacheable again) is discussed in the linked issue.

### Related Issues
Resolves #22419

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/index/mapper/WildcardFieldMapper.java
+++ b/server/src/main/java/org/opensearch/index/mapper/WildcardFieldMapper.java
@@ -839,7 +839,9 @@ public class WildcardFieldMapper extends ParametrizedFieldMapper {
 
         @Override
         public Weight createWeight(IndexSearcher searcher, ScoreMode scoreMode, float boost) throws IOException {
-            Weight firstPhaseWeight = firstPhaseQuery.createWeight(searcher, scoreMode, boost);
+            // Delegate to the searcher so the (cheap, stateless) first-phase query remains eligible for the
+            // query cache. Scores from the first phase are never used, since this weight produces constant scores.
+            Weight firstPhaseWeight = searcher.createWeight(firstPhaseQuery, ScoreMode.COMPLETE_NO_SCORES, 1f);
             return new ConstantScoreWeight(this, boost) {
                 @Override
                 public ScorerSupplier scorerSupplier(LeafReaderContext context) throws IOException {
@@ -892,7 +894,12 @@ public class WildcardFieldMapper extends ParametrizedFieldMapper {
 
                 @Override
                 public boolean isCacheable(LeafReaderContext leafReaderContext) {
-                    return true;
+                    // This weight references the shard-scoped SearchLookup and ValueFetcher through the
+                    // second-phase matching logic. Caching it would pin those (and everything they reference,
+                    // such as the QueryShardContext) in the query cache long after the search completes,
+                    // while equals/hashCode ignores them, so a cache hit could resurrect a stale lookup.
+                    // The rewritten first-phase query is cached independently via searcher.createWeight above.
+                    return searchLookup == null;
                 }
             };
         }
--- a/server/src/test/java/org/opensearch/index/mapper/WildcardFieldTypeTests.java
+++ b/server/src/test/java/org/opensearch/index/mapper/WildcardFieldTypeTests.java
@@ -272,10 +272,10 @@ public class WildcardFieldTypeTests extends FieldTypeTestCase {
         );
 
         try (Directory dir = newDirectory()) {
-            IndexWriter iw = new IndexWriter(dir, new IndexWriterConfig(Lucene.KEYWORD_ANALYZER));
-            iw.addDocument(appleTrigramDocument());
-            try (IndexReader reader = DirectoryReader.open(iw)) {
-                iw.close();
+            try (IndexWriter iw = new IndexWriter(dir, new IndexWriterConfig(Lucene.KEYWORD_ANALYZER))) {
+                iw.addDocument(appleTrigramDocument());
+            }
+            try (IndexReader reader = DirectoryReader.open(dir)) {
                 IndexSearcher searcher = new IndexSearcher(reader);
                 searcher.setQueryCache(null);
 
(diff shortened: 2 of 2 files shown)


bowenlan-amzn:
@Adam-Horse Thanks for the change here. I made a deeper fix to remove the references of request level objects from wildcard query https://github.com/opensearch-project/OpenSearch/pull/22433. We have merged that in, pls take a look if it looks interesting.

Issue #979, opened 2021-07-16
Tests required to verify changes in opensearch-env


xuezhou25:
About PR #875, tests(starting multiple instances with deb/rpm packages after override the value of `OPENSEARCH_PATH_CONF` manually) are needed to verify this PR changes and edge cases.

reta:
@xuezhou25  @dblock the distribution tests cover this particular functionality pretty well but we don't run them now, see please [1]

[1] https://github.com/opensearch-project/opensearch-build/issues/1572

swatichauhan814:
Hi @alchemist51 @andrross 
I would like to work on this issue. Could you please assign it to me?
Thanks!

PR #22436, opened 2026-07-10
tests/979-path-conf-env-override

swatichauhan814:
### Description
Add packaging tests to verify that a pre-set OPENSEARCH_PATH_CONF environment variable is not overridden when opensearch-env is sourced

- **ArchiveTests.java** - test60CustomPathConfOverride: sets OPENSEARCH_PATH_CONF via shell env before starting an archive-based node; asserts the custom config directory (and its node.name) is used instead of the default $OPENSEARCH_HOME/config.

- **PackageTests.java** - test85CustomPathConfViaSystemdDropIn: injects OPENSEARCH_PATH_CONF via a systemd drop-in Environment= directive (the correct mechanism since sh.getEnv() does not propagate into systemd services); asserts the custom config directory is used, not /etc/opensearch.

- **PackageTests.java** - test86DefaultEnvFileSourcedWhenPathConfAbsent: regression guard confirming that when no external override exists, the env file is still sourced and OpenSearch starts normally with the package default config path.

### Related Issues
Resolves #[https://github.com/opensearch-project/OpenSearch/issues/979]

### Check List


swatichauhan814:
Hi @alchemist51 @andrross
Requesting review on this PR.
Thanks!

PR #22439, opened 2026-07-10
Add local-node in-process dispatch for streaming transport (StreamTransportService)

Bukhtawar:
### Description

Adds an **in-process (local) streaming transport path** centrally in `StreamTransportService`. When a
streaming request's target is the coordinator's own node, the registered handler is run in-process and
its response batches are handed to the caller **by reference**, skipping the loopback
serialize → gRPC frame → deserialize round-trip for data that never leaves the JVM. Remote targets are
unaffected — they take the existing wire path unchanged.

### How it works

`StreamTransportService.getConnection(node)` returns an in-process `LocalStreamConnection` for the local
node instead of a loopback wire connection. `sendRequest` on that connection runs
the registered handler against a `LocalStreamChannel` — a `TransportChannel` whose
`sendResponseBatch`/`completeStream`/`sendResponse(Exception)` place items on a bounded queue, and whose
`responseStream()` feeds the caller's `handleStreamResponse(...)` by draining that queue. Batches are
passed by reference; ownership transfers to the consumer exactly as on the wire path (the consumer
closes each response it receives).

Three design points that were essential to get right (each caught and fixed during single-node
validation — see the root-cause notes below):

1. **Local detection matches on the persistent node id, not `isLocalNode()`.** The stream transport
   resolves its own local node via a separate `LocalNodeFactory` (different random `ephemeralId`), and
   `isLocalNode()` compares `ephemeralId`, so it never matches the cluster-state node here. Matching by
   node id also keeps this decision independent of `isLocalNode()`'s other role in `connectToNode()`,
   which suppresses establishing a loopback self-connection — the very connection non-local code paths
   still rely on.

2. **Dispatch is asynchronous.** `sendRequest` forks the stream drain onto `GENERIC` and returns
   immediately, mirroring the async wire path where `connection.sendRequest` hands off and returns.
   Draining inline would block the caller's dispatch thread for the whole stream and **serialize**
   sibling per-shard requests that the wire path runs concurrently.

3. **`handleStreamResponse` is invoked on the plain `TransportResponseHandler`** (it is a default method
   there); the `StreamTransportResponseHandler` marker sub-interface is **not** required, since
   streaming consumers such as the analytics fragment handler implement streaming on a plain
   `TransportResponseHandler`.

Backpressure: batches flow through a bounded `ArrayBlockingQueue`, so the producer blocks when full,
mirroring the wire path's gRPC-readiness flow control. The terminal state is sticky — once the consumer
drains end-of-stream / error / cancel, `nextResponse()` returns null without re-blocking on
`queue.take()`.

**Cancellation / exception / leak safety.** Ownership mirrors the wire path: the consumer owns and
closes every batch it takes from the queue. Batches left **undelivered** when the consumer cancels or
closes the stream (or that the producer sends after cancel) are released by the channel via the generic
`AutoCloseable` contract, so off-heap Arrow buffers don't leak — the counterpart to
`FlightServerChannel.releaseUnsentSource`. `ArrowBatchResponse` now implements `Closeable` (delegating
to its root close) so the core transport can free these without depending on the Arrow types. The queue
is the single ownership arbiter (`poll()` removes exactly once), so a batch is never both delivered and
released. If the consumer's `handleStreamResponse` throws, the channel is cancelled (draining/releasing
the remainder) and the failure is routed to `handleException`; producer-side failures enqueue a terminal
that surfaces as a `TransportException` on the next `nextResponse()`.

### Testing

**Unit tests** (`LocalStreamChannelTests`) — 8 tests covering in-order drain + end-of-stream, error
propagation to the consumer, bounded-queue backpressure (producer blocks when full, unblocks on drain),
consumer cancel dropping batches, single `sendResponse` → one batch then end, and the ownership/leak
contract: undelivered batches are `close()`d on cancel and on stream close, while a batch the consumer
already took is never closed by the channel (no double-free).

**End-to-end on a single-node cluster** with the full ClickBench `hits` dataset (99.9M rows, 5 shards).
The optimization is now unconditional; these numbers were gathered with a temporary toggle (since
removed) so the local path could be A/B'd against the wire path on the same node:

- **Correctness:** local path returns results identical to the wire path, verified per-query
  (interleaved local vs wire).
- **Local dispatch confirmed engaging:** instrumented `getConnection` fired the in-process path once per
  shard (5×) on the local path, 0× on the wire path.
- **Latency:** interleaved A/B (wire/local alternating, 10 iterations, 0 mismatches):

  | query (distinct keys) | wire | local | speedup |
  |---|---|---|---|
  | group-by UserID (17.7M) | 2204 ms | 1999 ms | **−9.3%** |
  | group-by WatchID (100M) | 3378 ms | 3166 ms | **−6.3%** |
  | group-by ClientIP (9.9M) | 1479 ms | 1411 ms | **−4.6%** |

  Pooled: **−6.9% mean / −10% median.** Distributions are non-overlapping per query (stdev 32–85 ms).

  > An earlier measurement of the central path showed inconsistent, direction-flipping deltas
  > (clientip +18%, watchid −8%). Root cause: the drain ran **inline on the dispatch thread**, serializing
  > the 5 shards instead of overlapping them — a fixed per-shard penalty that swamped the serde savings
  > on lighter queries. Forking the drain to `GENERIC` (design point 2 above) restored shard concurrency
  > and the win now shows cleanly across all three queries.

### Notes

- The benefit is topology-dependent. Single-node "wire" traffic is already Flight loopback, so the win
  here (~5-9%) comes from skipping serialize/deserialize + a copy; a multi-node cluster (real gRPC +
  cross-node transfer) should show a larger benefit.
- The path is always on for same-node streaming requests; there is no setting. Remote targets are
  unaffected.

### DIFF
--- a/plugins/arrow-base/src/main/java/org/opensearch/arrow/transport/ArrowBatchResponse.java
+++ b/plugins/arrow-base/src/main/java/org/opensearch/arrow/transport/ArrowBatchResponse.java
@@ -15,6 +15,7 @@ import org.opensearch.core.action.ActionResponse;
 import org.opensearch.core.common.io.stream.StreamInput;
 import org.opensearch.core.common.io.stream.StreamOutput;
 
+import java.io.Closeable;
 import java.io.IOException;
 
 /**
@@ -76,7 +77,7 @@ import java.io.IOException;
  * @opensearch.experimental
  */
 @ExperimentalApi
-public abstract class ArrowBatchResponse extends ActionResponse {
+public abstract class ArrowBatchResponse extends ActionResponse implements Closeable {
 
     private final VectorSchemaRoot batchRoot;
     private final byte[] metadata;
@@ -134,6 +135,19 @@ public abstract class ArrowBatchResponse extends ActionResponse {
     /** Returns the application metadata attached to this batch, or {@code null} if none. */
     public byte[] getMetadata() {
         return metadata;
+    }
+
+    /**
+     * Releases the off-heap buffers backing this batch by closing its root. Lets a transport release
+     * a batch it never delivered (e.g. an in-process local stream that was cancelled) without knowing
+     * the concrete Arrow type — the generic counterpart to the wire path's {@code releaseUnsentSource}.
+     * Consumers that take ownership normally close the root themselves; this is for the undelivered case.
+     */
+    @Override
+    public void close() {
+        if (batchRoot != null) {
+            batchRoot.close();
+        }
     }
 
     @Override
--- a/server/src/main/java/org/opensearch/common/settings/ClusterSettings.java
+++ b/server/src/main/java/org/opensearch/common/settings/ClusterSettings.java
@@ -933,6 +933,7 @@ public final class ClusterSettings extends AbstractScopedSettings {
                 ForceMergeManagerSettings.JVM_THRESHOLD_PERCENTAGE_FOR_AUTO_FORCE_MERGE,
                 ForceMergeManagerSettings.CONCURRENCY_MULTIPLIER,
                 StreamTransportService.STREAM_TRANSPORT_REQ_TIMEOUT_SETTING,
+                StreamTransportService.STREAM_TRANSPORT_LOCAL_QUEUE_DEPTH_SETTING,
                 StreamSearchTransportService.STREAM_SEARCH_ENABLED,
                 TieredStoragePrefetchSettings.READ_AHEAD_BLOCK_COUNT,
                 TieredStoragePrefetchSettings.STORED_FIELDS_PREFETCH_ENABLED_SETTING,
--- a/server/src/main/java/org/opensearch/common/settings/ClusterSettings.java
+++ b/server/src/main/java/org/opensearch/common/settings/ClusterSettings.java
@@ -934,6 +934,7 @@ public final class ClusterSettings extends AbstractScopedSettings {
                 ForceMergeManagerSettings.CONCURRENCY_MULTIPLIER,
                 StreamTransportService.STREAM_TRANSPORT_REQ_TIMEOUT_SETTING,
                 StreamTransportService.STREAM_TRANSPORT_LOCAL_QUEUE_DEPTH_SETTING,
+                StreamTransportService.STREAM_TRANSPORT_LOCAL_DISPATCH_ENABLED_SETTING,
                 StreamSearchTransportService.STREAM_SEARCH_ENABLED,
                 TieredStoragePrefetchSettings.READ_AHEAD_BLOCK_COUNT,
                 TieredStoragePrefetchSettings.STORED_FIELDS_PREFETCH_ENABLED_SETTING,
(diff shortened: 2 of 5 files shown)


plugins/arrow-base/src/main/java/org/opensearch/arrow/transport/ArrowBatchResponse.java
```diff
@@ -136,6 +137,19 @@ public byte[] getMetadata() {
...
     }
 
+    /**
+     * Releases the off-heap buffers backing this batch by closing its root. Lets a transport release
+     * a batch it never delivered (e.g. an in-process local stream that was cancelled) without knowing
+     * the concrete Arrow type — the generic counterpart to the wire path's {@code releaseUnsentSource}.
+     * Consumers that take ownership normally close the root themselves; this is for the undelivered case.
+     */
+    @Override
+    public void close() {
```
rishabhmaurya, line 147: I'm not in favour of this close method here. I think I intentionally avoided it as `ArrowBatchResponse` is one of the response in stream of responses and shouldn't own closing of root - shared across multiple responses.

rishabhmaurya:
@Bukhtawar to me, this change looks risky due to its high maintainability and complexity. The DirectChannel for regular transport has simple semantics whereas streaming comes with challenges like - multiple responses, cancellation, backpressure, underlying buffer lifecycle/management and risks of leaks, header propagation to name a few.

Maybe we can explore the use of in-process server making use of in-process channel in gRPC, its mostly used for testing but looks stable and supports things we need. This should avoid all above mentioned complexities as its another channel and all abstractions above it continue to work as it is - https://github.com/opensearch-project/OpenSearch/commit/d2f8893064eb3aac8337aa02fdde2eaa029a90b9#diff-a71ea7a04beada03e9202dafb97e3cd8bf374faf30052115b4e0d6aebe764e71

I think we should see similar gains since we avoid tcp/socket altogether and buffer copies. We need to handle the lifecycle of VSR close a bit carefully here as well.

Bukhtawar:
> @Bukhtawar to me, this change looks risky due to its high maintainability and complexity. The DirectChannel for regular transport has simple semantics whereas streaming comes with challenges like - multiple responses, cancellation, backpressure, underlying buffer lifecycle/management and risks of leaks, header propagation to name a few.
> 
> Maybe we can explore the use of in-process server making use of in-process channel in gRPC, its mostly used for testing but looks stable and supports things we need. This should avoid all above mentioned complexities as its another channel and all abstractions above it continue to work as it is - [d2f8893#diff-a71ea7a04beada03e9202dafb97e3cd8bf374faf30052115b4e0d6aebe764e71](https://github.com/opensearch-project/OpenSearch/commit/d2f8893064eb3aac8337aa02fdde2eaa029a90b9#diff-a71ea7a04beada03e9202dafb97e3cd8bf374faf30052115b4e0d6aebe764e71)
> 
> I think we should see similar gains since we avoid tcp/socket altogether and buffer copies. We need to handle the lifecycle of VSR close a bit carefully here as well.


While I agree streaming has challenges they can be addressed as well(some of them are already handled). However if we have a simpler solution that doesn't diminish the gains, I am all up for it. Can we run some benchmarks on the same, I believe although simple it would have minor overhead from marshalling(metadata mostly) as the data would be just be referenced.

plugins/arrow-base/src/main/java/org/opensearch/arrow/transport/ArrowBatchResponse.java
Bukhtawar: This is just a defensive check at this point as we have encountered multiple leaks

PR #22448, opened 2026-07-11
 [analytics-engine] Fix native memory leak on Arrow C Data batch import under allocator pressure

sandeshkr419:
### Description

`DatafusionResultStream` and `LuceneResultStream` imported each native Arrow batch directly into the caller's `BufferAllocator` via `Data.importIntoVectorSchemaRoot`, which charges every imported buffer against that allocator as it walks the array. When the allocator is bounded and its limit is hit part-way through a wide batch (e.g. chained-`concat` `Utf8View` columns under concurrent load), the import throws `OutOfMemoryException`.

arrow-java's `ReferenceCountedArrowArray.unsafeAssociateAllocation` calls `retain()` **before** the throwing `wrapForeignAllocation` and does not roll it back, so the imported array's reference count stays elevated, the C Data release callback never fires, and the **entire native batch leaks in the producer's native allocator** — invisible to the JVM heap and the Java Arrow allocator. Under a concurrent chained-`concat` storm this accumulates multiple GB of native memory that never drains, eventually wedging the node into HTTP 429s until restart. It is visible only in `_plugins/arrow_base/stats → memory_pools.runtime.allocated_bytes`; the DataFusion pool stats and circuit breakers read 0.

**Fix:** import each batch into a per-batch staging allocator that is a fresh unbounded child of the root allocator, so a single import can never OOM part-way through the array and the release callback always fires. The batch is returned as-is (zero-copy, no buffer transfer) and released by the existing consumer close paths, which drives the C Data reference count to zero. Staging allocators are reclaimed once drained (per import and at stream close); on import failure the staging allocator is closed immediately, firing the native release for the whole batch. The same fix is applied to both result-stream import paths.

This is independent of whether `POOL_QUERY` is bounded: making the shared pool unbounded does **not** fix it, because many concurrent imports still pile onto the one shared pool and hit the node's overall memory ceiling, so an import still OOMs mid-batch and strands. A per-import staging allocator is what prevents the mid-array OOM. The underlying arrow-java bug is being reported upstream separately; this change works around it for the OpenSearch import paths.

### Testing

- `DatafusionImportLeakTests` (new, self-contained — no native runtime): exports a multi-buffer batch across the C Data Interface and imports it back through the production staging path under a caller allocator far too small to hold it; asserts the producer allocator (standing in for the native producer) fully drains. This test fails on the unfixed direct-import path (`Memory was leaked ... (8521000)`) and passes with the fix, so it is a genuine regression guard. A companion test pins the arrow-java mid-import-OOM leak deterministically.
- `DatafusionResultStreamTests.testStreamConsumeAndCloseDrainsAllocator` verifies the staging import path end-to-end against a real native stream under a small caller allocator and asserts the shared root allocator drains after consume + close.
- Validated on a live 1-node/3-shard m8g.large cluster (textbench_50m) under a 16-worker chained-`concat` storm: stock leaks ~3.1 GB that stays flat at idle forever; with the fix native `allocated_bytes` drains from ~2.2 GB back to ~13 MB after every storm, in both bounded and unbounded `POOL_QUERY` configurations.

### Check List


sandbox/plugins/analytics-backend-lucene/src/main/java/org/opensearch/be/lucene/LuceneResultStream.java
```diff
@@ -79,6 +80,7 @@ public void close() {
         try {
             if (iteratorInstance != null) {
                 iteratorInstance.closeLastBatch();
+                iteratorInstance.reclaimDrainedStaging();
```
mch2, line 83: this doesn't feel right - we are storing allocators for each batch and only clearing when last batch is consumed?  We should be releasing each batch allocator as its read?

sandeshkr419: Reclaim actually runs at the start of every importBatch(), not just at last-batch — each drained staging allocator is closed as soon as its batch's buffers are released (getAllocatedMemory() == 0), and the call in close() is only a final sweep. So it's incremental, not stored-till-the-end.

mch2 (approved):

PR #22450, opened 2026-07-12
perf: optimize bucket counting and tracking in LongKeyedBucketOrds

### LINKED ISSUE #22449: [Performance] Optimize LongKeyedBucketOrds by caching bucket counts and running maximum
### Description
During terms aggregation with many buckets, `LongKeyedBucketOrds.FromMany` performs linear scans (`O(N)`) over the underlying `LongLongHash` map whenever `bucketsInOrd(owningBucketOrd)` and `maxOwningBucketOrd()` are queried. 

Specifically, in `FromMany`:
1. `bucketsInOrd` iterates from `0` to `ords.size()` to count how many keys match the given `owningBucketOrd`.
2. `maxOwningBucketOrd` iterates from `0` to `ords.size()` to compute the maximum value of the first key.

Under high-cardinality searches with nested bucket structures, these repeated scans consume substantial CPU cycles and slow down the query execution path.

### Benchmark Performance Impact
We ran a JMH microbenchmark simulating 50,000 buckets and 10,000 queries.
- **Baseline (Linear scans)**: `2.364 ns/op`
- **Optimized (O(1) lookups)**: `0.593 ns/op`
- **Performance Improvement**: **~4x speedup** (approx. 75% reduction in lookup latency)

### Proposed Solution
1. Maintain `maxOwningBucketOrd` dynamically as a primitive field inside `FromMany` by updating it on every `add()` call ($O(1)$ complexity).
2. Maintain a dynamic `LongArray` `bucketOrdsCounts` managed by `BigArrays` to cache bucket cardinalities per `owningBucketOrd`, incrementing the count incrementally in `add()`. This converts the complexity of `bucketsInOrd` from $O(N)$ to $O(1)$.

### Related component

Search:Aggregations

rajat315315:
### Description
This PR optimizes `LongKeyedBucketOrds.FromMany` to replace $O(N)$ linear scans in `bucketsInOrd` and `maxOwningBucketOrd` with $O(1)$ operations:
1. Tracks `maxOwningBucketOrd` dynamically as a class field updated on insertions.
2. Caches bucket counts per `owningBucketOrd` in a `LongArray` managed by `BigArrays`, incrementing it incrementally.

### Tests and Verification
- **Unit Tests**: Running `./gradlew :server:test --tests "org.opensearch.search.aggregations.bucket.terms.LongKeyedBucketOrdsTests"` passed successfully.
- **Spotless Formatting**: Verified with `./gradlew spotlessJavaCheck`.
- **Microbenchmarking**: Microbenchmarks show a ~4x speedup in lookup operations (see the corresponding issue for details).

### Related Issues
Resolves #22449 

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/LongKeyedBucketOrds.java
+++ b/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/LongKeyedBucketOrds.java
@@ -52,8 +52,7 @@ public abstract class LongKeyedBucketOrds implements Releasable {
         return cardinality.map(estimate -> estimate < 2 ? new FromSingle(bigArrays) : new FromMany(bigArrays));
     }
 
-    private LongKeyedBucketOrds() {
-    }
+    private LongKeyedBucketOrds() {}
 
     /**
      * Add the {@code owningBucketOrd, value} pair. Return the ord for
@@ -108,7 +107,7 @@ public abstract class LongKeyedBucketOrds implements Releasable {
     public interface BucketOrdsEnum {
         /**
          * Advance to the next value.
-         * 
+         *
          * @return {@code true} if there *is* a next value,
          *         {@code false} if there isn't
          */
@@ -250,8 +249,10 @@ public abstract class LongKeyedBucketOrds implements Releasable {
             // performance.
             long ord = ords.add(owningBucketOrd, value);
             if (ord >= 0) {
-                maxOwningBucketOrd = Math.max(maxOwningBucketOrd, owningBucketOrd);
-                bucketOrdsCounts = bigArrays.grow(bucketOrdsCounts, owningBucketOrd + 1);
+                if (owningBucketOrd > maxOwningBucketOrd) {
+                    maxOwningBucketOrd = owningBucketOrd;
+                    bucketOrdsCounts = bigArrays.grow(bucketOrdsCounts, owningBucketOrd + 1);
+                }
                 bucketOrdsCounts.increment(owningBucketOrd, 1);
             }
             return ord;
--- a/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/LongKeyedBucketOrds.java
+++ b/server/src/main/java/org/opensearch/search/aggregations/bucket/terms/LongKeyedBucketOrds.java
@@ -52,7 +52,8 @@ public abstract class LongKeyedBucketOrds implements Releasable {
         return cardinality.map(estimate -> estimate < 2 ? new FromSingle(bigArrays) : new FromMany(bigArrays));
     }
 
-    private LongKeyedBucketOrds() {}
+    private LongKeyedBucketOrds() {
+    }
 
     /**
      * Add the {@code owningBucketOrd, value} pair. Return the ord for
@@ -107,6 +108,7 @@ public abstract class LongKeyedBucketOrds implements Releasable {
     public interface BucketOrdsEnum {
         /**
          * Advance to the next value.
+         * 
          * @return {@code true} if there *is* a next value,
          *         {@code false} if there isn't
          */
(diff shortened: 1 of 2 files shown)


server/src/main/java/org/opensearch/search/aggregations/bucket/terms/LongKeyedBucketOrds.java
```diff
@@ -229,16 +230,27 @@ public void close() {
...
 
         @Override
         public long add(long owningBucketOrd, long value) {
             // This is in the critical path for collecting most aggs. Be careful of performance.
-            return ords.add(owningBucketOrd, value);
+            long ord = ords.add(owningBucketOrd, value);
+            if (ord >= 0) {
+                maxOwningBucketOrd = Math.max(maxOwningBucketOrd, owningBucketOrd);
+                bucketOrdsCounts = bigArrays.grow(bucketOrdsCounts, owningBucketOrd + 1);
+                bucketOrdsCounts.set(owningBucketOrd, bucketOrdsCounts.get(owningBucketOrd) + 1);
```
msfroh: You can use `bucketOrdsCounts.increment(owningBucketOrd, 1)`. It might help performance just a little bit, but also IMO it's more readable.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/LongKeyedBucketOrds.java
```diff
@@ -229,16 +230,27 @@ public void close() {
...
         }
 
         @Override
         public long add(long owningBucketOrd, long value) {
             // This is in the critical path for collecting most aggs. Be careful of performance.
-            return ords.add(owningBucketOrd, value);
+            long ord = ords.add(owningBucketOrd, value);
+            if (ord >= 0) {
+                maxOwningBucketOrd = Math.max(maxOwningBucketOrd, owningBucketOrd);
+                bucketOrdsCounts = bigArrays.grow(bucketOrdsCounts, owningBucketOrd + 1);
```
msfroh: I feel like we can slightly reduce the number of comparisons here (which normally I wouldn't worry about, but it is in the critical path, so it might actually be worth it).

What if we make this:
```
if (owningBucketOrd > maxOwningBucketOrd) {
  maxOwningBucketOrd = owningBucketOrd;
  bucketOrdsCounts = bigArrays.grow(bucketOrdsCounts, owningBucketOrd + 1);
}
```

Basically, we get two comparisons for the price of one. What do you think?

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/LongKeyedBucketOrds.java
```diff
@@ -313,7 +316,13 @@ public long ord() {
...
         @Override
         public void close() {
-            ords.close();
+            try {
+                ords.close();
+            } finally {
+                if (bucketOrdsCounts != null) {
+                    bucketOrdsCounts.close();
+                }
+            }
```
msfroh: A couple of things here:
1. `ords.close()` should *never* throw an exception, so I don't believe the `try` is necessary.
2. `bucketOrdsCounts` should never be `null` (since it's allocated in the constructor). So I don't think we need the `null` check.

msfroh (commented):
Thanks @rajat315315 ! 

This looks good overall. I left some pretty minor cleanup comments.

server/src/main/java/org/opensearch/search/aggregations/bucket/terms/LongKeyedBucketOrds.java
rajat315315: Yes, I agree. Updated the code. :+1:

msfroh (approved):

msfroh:
@rajat315315 -- one last thing. I noticed that your commits are signed off with:

```
Signed-off-by: Your Name <rajatjain.ix@gmail.com>
```

Unless I'm mistaken, our DCO rules want each commit signed using a real name. Can you please run `git config --global user.name "Rajat Jain"`, then do a `git commit --amend -s` to update the signature on the last commit?

rajat315315:
Done.

PR #22452, opened 2026-07-12
Return 400 for user-facing planner validation errors

finnegancarroll:
## Description

Changes two specific `IllegalStateException` throws to `IllegalArgumentException` in the analytics engine planner. These are error paths triggered by user-caused issues that should return HTTP 400 (actionable error) rather than 500 (redacted internal error).

### Changes

| File | Error | Rationale |
|------|-------|-----------|
| `FieldStorageResolver.java` | "Field [X] not found in field storage for index" | User's query references a field that doesn't exist in the index mapping (e.g., typo in field name). |
| `OpenSearchFilterRule.java` | "Unrecognized scalar function [X] in call [Y]" | User used an unsupported UDF in a WHERE clause. |

### DIFF
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/FieldStorageResolver.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/FieldStorageResolver.java
@@ -122,7 +122,7 @@ public class FieldStorageResolver {
         for (String fieldName : fieldNames) {
             FieldStorageInfo info = fieldStorage.get(fieldName);
             if (info == null) {
-                throw new IllegalStateException("Field [" + fieldName + "] not found in field storage for index");
+                throw new IllegalArgumentException("Field [" + fieldName + "] not found in field storage for index");
             }
             result.add(info);
         }
--- a/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
+++ b/sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchFilterRule.java
@@ -283,7 +283,7 @@ public class OpenSearchFilterRule extends RelOptRule {
             }
             ScalarFunction scalarFunc = ScalarFunction.fromSqlOperatorWithFallback(scalarFunctionCall.getOperator());
             if (scalarFunc == null) {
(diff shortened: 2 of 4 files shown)


sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchValuesRule.java
```diff
@@ -45,7 +45,7 @@ public void onMatch(RelOptRuleCall call) {
         }
         List<String> viableBackends = context.getCapabilityRegistry().operatorBackends(EngineCapability.VALUES);
         if (viableBackends.isEmpty()) {
-            throw new IllegalStateException("No backend supports VALUES");
+            throw new IllegalArgumentException("No backend supports VALUES");
```
mch2: This indicates an engine misconfiguration and ISE is appropriate.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/CapabilityResolutionUtils.java
```diff
@@ -36,7 +36,7 @@ public static List<String> filterByReduceCapability(CapabilityRegistry registry,
             }
         }
         if (result.isEmpty()) {
-            throw new IllegalStateException("No viable backend supports coordinator reduce among " + viableBackends);
+            throw new IllegalArgumentException("No viable backend supports coordinator reduce among " + viableBackends);
```
mch2: same here with ISE.

sandbox/plugins/analytics-engine/src/main/java/org/opensearch/analytics/planner/rules/OpenSearchUnionRule.java
```diff
@@ -105,7 +105,7 @@ public void onMatch(RelOptRuleCall call) {
         viableBackends.retainAll(unionCapable);
 
         if (viableBackends.isEmpty()) {
-            throw new IllegalStateException("No backend supports UNION among viable backends after intersecting inputs");
+            throw new IllegalArgumentException("No backend supports UNION among viable backends after intersecting inputs");
```
mch2: I think this is also an engine misconfiguration, not request level 403.

mch2 (approved):

PR #22457, opened 2026-07-13
Add supportsWildcard capability to Attribute interface

### LINKED ISSUE #22456: [WLM] Support wildcard '*' in principal.username / principal.role rule values

WLM auto-tagging rules accept the wildcard character `*` in `principal.username` and `principal.role` values, but these subfields were matched **exactly** — the wildcard was silently stripped and never behaved as a wildcard. This let users create rules that look like prefix rules but match only the literal stem, with no error or warning.

**Example**

```json
PUT _rules/workload_group
{
  "description": "tag dev users",
  "principal": { "username": ["dev*"] },
  "workload_group": "<id>"
}
```

A user would reasonably expect this to match `dev`, `dev_john`, `dev_alice`, etc. In reality it matched only the literal username `dev`, because `PrincipalAttribute#findAttributeMatches` (opensearch-project/security) matched via `getExactMatch` rather than the prefix-based `getMatches`.

This is also inconsistent with `index_pattern` rules, which support a trailing `*` wildcard, and with roles mappings in the security plugin, which support wildcard/pattern matching for users and backend roles.

Support a trailing `*` wildcard in `principal.username` / `principal.role`, so `username: ["dev*"]` prefix-matches `dev`, `dev_john`, `dev_alice`, etc. — consistent with `index_pattern` rules and roles mappings.

This builds on the framework fix in #22461, which made the store's `getMatches` distinguish exact values (no `*`) from trailing-wildcard values (`<prefix>*`). With that in place, the only change needed is in the security plugin, where `PrincipalAttribute` is defined: switch `findAttributeMatches` from `getExactMatch` to the wildcard-aware `getMatches`. Values without a trailing `*` continue to match exactly, and the existing username/role subfield weighting is unchanged.

### Additional context

Affected components: `PrincipalAttribute` in opensearch-project/security (builds on `modules/autotagging-commons` in core).

Completed by opensearch-project/security#6324.

dzane17:
### Description
Adds a `supportsWildcard()` default method to the `Attribute` interface, returning `true`. This declares the capability contract: the framework's default `findAttributeMatches` implementation uses `getMatches`, which is wildcard-aware (trailing `*` = prefix match, no `*` = exact match, per #22461). Attributes that match via `getExactMatch` and do not honor wildcards can override to return `false`.

This is the core-side preparation for enabling wildcard support in `principal.username`/`principal.role` rules. The matching-side change (switching `PrincipalAttribute.findAttributeMatches` from `getExactMatch` to `getMatches`) will be a companion PR in opensearch-project/security.

### Related Issues
Resolves #22456

### Check List

### DIFF
--- a/modules/autotagging-commons/common/src/main/java/org/opensearch/rule/autotagging/Attribute.java
+++ b/modules/autotagging-commons/common/src/main/java/org/opensearch/rule/autotagging/Attribute.java
@@ -49,6 +49,19 @@ public interface Attribute extends Writeable {
     }
 
     /**
+     * Indicates whether this attribute supports the wildcard character ({@code *}) in its rule values.
+     * Attributes that perform prefix (wildcard) matching override their {@link #findAttributeMatches}
+     * implementation to use {@link org.opensearch.rule.storage.AttributeValueStore#getMatches} which
+     * is wildcard-aware. Defaults to {@code true} because the framework's default
+     * {@code findAttributeMatches} uses {@code getMatches}. Attributes that match via
+     * {@link org.opensearch.rule.storage.AttributeValueStore#getExactMatch} should override this
+     * to return {@code false} if they do not intend to honor wildcards.
+     */
+    default boolean supportsWildcard() {
+        return true;
+    }
+
+    /**
      * Ensure that `validateAttribute` is called in the constructor of attribute implementations
      * to prevent potential serialization issues.
      */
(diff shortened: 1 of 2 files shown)


kaushalmahi12 (approved):

cwperks:
@dzane17 what would it take to support wildcard here? I would prefer that route as it would be in line with behavior for roles mappings.

dzane17:
PR #22461 gave `index_pattern` rules the ability to do both exact and trailing-wildcard prefix matching. I've opened a PR in the security repo to grant the `principal` attribute the same semantics: opensearch-project/security#6324. After that, all WLM rules will be consistent.

This PR is no longer needed, since we're adding wildcard support to username/role rules rather than rejecting it.

Issue #16888, opened 2024-12-20
[Proposal] Rule Matching


kaushalmahi12:
### Please describe the end goal of this project

This document covers the Rule Matching for incoming search requests in depth. After reading this document, reader will have sound understanding of the following

1. **How the request attributes are extracted ?**
2. **How exactly does the matching algorithm works on the extracted attributes ?**
3. **What are the new constructs and how they work with each other ?**

### Assumptions

* Rule attributes (cardinality) will be limited in nature at feature level and aggregated count.
    * When creating Rules the control will first go to the feature to do the
        * Cardinality validation
        * Attributes validation
* Each feature will define their own attributes which the feature will be responsible for validating.

### New Rule insertion process

{
    
   "attribute1": ["value*"],
   "attribute2": ["value*"],
   "label": "fjagjag9243421_425285",
   "updatedAt": "12-03-2024T18:00:23Z",
   "feature": "WLM"
}

New rule insertion works pretty much how insertion in [compressed trie](https://en.wikipedia.org/wiki/Radix_tree) works. But given each attribute can take N(this will be limited e,g; 5 at max) number there is a chance that two or more rules could share one of the values in their respective attribute values list.
In this case whichever rule was created or modified last will override the label for that attribute in the trie. 

### Attribute extraction process

Given that the Rule attributes are specific to a feature, feature will provide a construct to extract the attributes. For example WLM feature would use threadContext to retrieve the authN info and request itself to retrieve the indices.

### Rule Matching Process

In a nutshell the rule matching will be per attribute trie (prefix tree) based matching. Hence ultimately we will be dealing with a list of values per attribute. How this per attribute based list is calculated, we will understand that with the help of some examples 
![RulesMatching](https://github.com/user-attachments/assets/d86b30a6-ae51-4c76-bd15-21baf7011ff7)

lets say above picture captures a subpart of the in-memory rules Trie. Now following scenarios can occur while calculating possible labels for an attribute value. 

#### Scenario - 1

When the search for the attribute value ends on a node which has a label 

`search_value: log_q3_2024`

In this case the resulting label would be abc.

#### Scenario - 2

When the search ends on a node which doesn’t have a label then there are following sub scenarios

* There is an empty prefix match for this value i,e the search ends at the root node which holds the empty value ( e,g; `order_q3_2024`)
    * In this case we assign nothing and downstream systems can treat this as default behavior
[...]

Issue #16797, opened 2024-12-06
[WLM] Automated labeling of search requests 


kaushalmahi12:
### Is your feature request related to a problem? Please describe

Recently we launched WLM subfeature i,e; multitenant search resiliency. But the feature still required an external hint to be sent along with each request via HTTP header. Hence this approach puts the burden on the user to apply these hints intelligently. 

This can become a real pain if the access is programmatic and if not planned properly the programmatic multitenant access can become unmanageable. Hence it would rather be great if user could just define some **rules** to determine what should be the right tenant for certain class (confirms to a rule) of request.

Though we have touch based on this idea in the following RFCs
* https://github.com/opensearch-project/OpenSearch/issues/14425
* https://github.com/opensearch-project/OpenSearch/issues/13341

In this issue I want to go over the high level approach to achieve this. 

### Describe the solution you'd like

Given that this tagging component will lie directly on the search path, we will keep efficient in memory snapshot of the rules for faster processing. The label assignment will only happen once for a request at co-ordinator node irrespective of number of shards it is going to hit.

### Rules schema and Storage options

Rule Schema
```
{
    
   "attribute1": ["value*"],
   "attribute2": ["value*"],
   "label": "fjagjag9243421_425285",
   "updatedAt": "12-03-2024T18:00:23Z"
}
```

* **Cluster State** -  If we use search pipelines to encapsulate rules for determining the label then soon enough the pipelines will explode in numbers which can be detrimental to cluster state processing and could become a bottleneck in clusters stability. In addition to this the cluster state is already quite bloated hence maybe it wouldn’t be such a great idea to use this option.
* **System Index** - This will definitely help us decouple the rules storage and processing from cluster manager related tasks. But since there is no mechanism in indices to propagate these changes to all nodes, it will compel us to either periodically refresh the rules on all nodes or define a custom request handlers to carry out the refresh.

### In-memory Structure for Rules

Since we want to hold all the rules in memory and do a fast prefix based string matching trie data structure becomes a natural choice for this problem.

We will keep per attribute trie in memory, each trie will give us a possible list of matching labels.

### Rules storage

Following diagram illustrates the rules storage process and how does the structure evolves over time on incremental rule additions 

![RulesStorageAndMatching](https://github.com/user-attachments/assets/a84ee92b-c52b-4351-ade2-afa4a78e7866)

### Rules Matching
[...]

PR #22461, opened 2026-07-13
Match index_pattern rules exactly unless value ends in wildcard

### LINKED ISSUE #22460: [WLM] index_pattern rules without a trailing '*' still match by prefix
### Describe the bug

An `index_pattern` rule value **without** a trailing `*` still matches request indices that merely share it as a prefix — the trailing wildcard is always implied. There is no way to write an exact-only index_pattern rule.

The reason: `index_pattern` matching always goes through the prefix-based trie lookup (`DefaultAttributeValueStore.getMatches`), which walks every prefix of the request's index. The `*` is stripped at storage time (`InMemoryRuleProcessingService.addOperation` does `value.replace("*", "")`), so `"my-index"` and `"my-index*"` become the same trie key and behave identically.

This contradicts the documentation, which states index_pattern supports exact matching for a plain value and prefix matching only when the value ends in `*` (workload group rules page: "Each element can be a full index name or a prefix ending in `*`"; rule-based auto-tagging page lists "Exact match: `logs-2025-04`" vs "Prefix pattern: `logs-2025-*`").

### Example

```
Rule: { "index_pattern": ["my-index"], "workload_group": "groupA" }
```

- Expected (per docs): matches only the index `my-index`.
- Actual: also matches `my-index-1`, `my-index-archive`, `my-index_2024`, etc.

So a request targeting `my-index-1` is tagged with `groupA` even though the rule named the exact index `my-index`.

### Expected behavior

- `index_pattern: ["my-index"]` matches only the index `my-index` (exact).
- `index_pattern: ["my-index*"]` matches `my-index`, `my-index-1`, etc. (prefix).

### Not affected: principal attributes

`principal.username` and `principal.role` do **not** have this bug. `PrincipalAttribute` overrides matching to use exact lookup (`getExactMatch`), so those subfields already match exactly and are unaffected.

### Additional context

This is also the root of a related duplicate-detection gap: because `"my-index"` and `"my-index*"` collapse to the same in-memory key but are stored as distinct literal strings, they are not detected as duplicates, yet a request matching the shared key can tie between them.

dzane17:
### Description
An `index_pattern` rule value without a trailing `*` still matched request indices that merely shared it as a prefix — the trailing wildcard was always implied, so there was no way to write an exact-only index rule. For example, a rule for `index_pattern: ["my-index"]` also matched `my-index-1`, `my-index-archive`, etc. This contradicts the documentation, which states a plain value is matched exactly and only a value ending in `*` matches by prefix.

Root cause: the `*` was stripped at storage time (`InMemoryRuleProcessingService`) so `"my-index"` and `"my-index*"` became the same trie key, and `DefaultAttributeValueStore.getMatches` always walked prefixes.

This change stores values verbatim and makes `getMatches` wildcard-aware:
- A value stored without a trailing `*` is matched exactly.
- A value of the form `<prefix>*` matches request keys that start with `<prefix>` (scored by matched length, as before).
- The empty-key "no constraint" behavior is unchanged.

The change is confined to two production files to keep it easy to backport to affected versions.

**Behavior changes to note:**
- `index_pattern` rules without a trailing `*` stop prefix-matching after upgrade. Users relying on the implicit prefix behavior must append `*` to their rule values. No reindex or migration is needed — matching is in-memory and rebuilt on each rule sync.
- `principal.username`/`principal.role` rules **without** `*` are unaffected. However, an existing principal rule **containing** `*` (e.g. `username: ["dev*"]`) previously matched the stripped literal value (`dev`); with verbatim storage it matches nothing, since principal matches via `getExactMatch`. Honoring the wildcard for principal attributes will be handled in a follow-up (#22457).

### Related Issues
Resolves #22460

### Check List

### DIFF
--- a/modules/autotagging-commons/common/src/main/java/org/opensearch/rule/storage/DefaultAttributeValueStore.java
+++ b/modules/autotagging-commons/common/src/main/java/org/opensearch/rule/storage/DefaultAttributeValueStore.java
@@ -25,6 +25,11 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
  * ref: https://commons.apache.org/proper/commons-collections/javadocs/api-4.4/org/apache/commons/collections4/trie/PatriciaTrie.html
  */
 public class DefaultAttributeValueStore<K extends String, V> implements AttributeValueStore<K, V> {
+    /**
+     * Trailing character in a stored value that marks it as a prefix (wildcard) pattern. A stored value
+     * without this character is matched exactly.
+     */
+    private static final String WILDCARD = "*";
     private final PatriciaTrie<Set<V>> trie;
     private static final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
     private static final ReentrantReadWriteLock.ReadLock readLock = lock.readLock();
@@ -96,18 +101,30 @@ public class DefaultAttributeValueStore<K extends String, V> implements Attribut
         readLock.lock();
         try {
             List<MatchLabel<V>> results = new ArrayList<>();
-            StringBuilder prefixBuilder = new StringBuilder(key);
 
+            // Exact match: a value stored without a trailing wildcard matches the request key exactly.
+            // The empty key is reserved for "no constraint" rules and is handled separately below.
+            if (!key.isEmpty()) {
+                addMatches(results, trie.get(key), 1f);
+            }
+
+            // Prefix (wildcard) matches: a stored value of the form "<prefix>*" matches the request key
+            // when the key starts with <prefix>. Walk each prefix of the key from longest to shortest.
+            StringBuilder prefixBuilder = new StringBuilder(key);
             for (int i = key.length(); i >= 0; i--) {
-                Set<V> values = trie.get(prefixBuilder.toString());
+                Set<V> values = trie.get(prefixBuilder.toString() + WILDCARD);
                 if (values != null && !values.isEmpty()) {
-                    float score = (float) prefixBuilder.length() / key.length();
+                    float score = key.isEmpty() ? 0f : (float) prefixBuilder.length() / key.length();
                     addMatches(results, values, score);
                 }
                 if (!prefixBuilder.isEmpty()) {
                     prefixBuilder.deleteCharAt(prefixBuilder.length() - 1);
                 }
             }
+
+            // Rules that do not constrain this attribute are stored under the empty key and match anything.
+            addMatches(results, trie.get(""), 0f);
+
             return results;
         } finally {
             readLock.unlock();
--- a/modules/autotagging-commons/common/src/main/java/org/opensearch/rule/storage/DefaultAttributeValueStore.java
+++ b/modules/autotagging-commons/common/src/main/java/org/opensearch/rule/storage/DefaultAttributeValueStore.java
@@ -103,22 +103,25 @@ public class DefaultAttributeValueStore<K extends String, V> implements Attribut
             List<MatchLabel<V>> results = new ArrayList<>();
 
             // Exact match: a value stored without a trailing wildcard matches the request key exactly.
-            // The empty key is reserved for "no constraint" rules and is handled separately below.
-            if (!key.isEmpty()) {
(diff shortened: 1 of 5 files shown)


modules/autotagging-commons/common/src/main/java/org/opensearch/rule/storage/DefaultAttributeValueStore.java
```diff
@@ -96,18 +101,33 @@ public List<MatchLabel<V>> getMatches(String key) {
...
 
-            for (int i = key.length(); i >= 0; i--) {
-                Set<V> values = trie.get(prefixBuilder.toString());
-                if (values != null && !values.isEmpty()) {
-                    float score = (float) prefixBuilder.length() / key.length();
-                    addMatches(results, values, score);
-                }
-                if (!prefixBuilder.isEmpty()) {
-                    prefixBuilder.deleteCharAt(prefixBuilder.length() - 1);
+            // Exact match: a value stored without a trailing wildcard matches the request key exactly.
```
cwperks, line 105: I think we should aim for consistency with roles mappings here to create a unified experience across the platform. IMO I think I would like to see [WildcardMatcher](https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/support/WildcardMatcher.java) moved to the core and used here as well. Why does wildcard only need to be trailing?

dzane17: I reviewed WildcardMatcher. It can work, but there are a few drawbacks:

1. Efficiency. `WildcardMatcher.getAllMatchingPatterns` loops through all N patterns (O(N)). The current trie method walks the index name string, so it only depends on the number of characters (O(L), independent of N). Both the WLM rule count (N) and index name length (L) are bounded, so this is not a major difference. However, WLM rule tagging occurs on the hot search path, so speed is critical.

2. No scoring. WildcardMatcher returns which patterns matched, not how well. WLM has to pick one winner (exact beats a longer prefix beats a shorter prefix), so ranking is an extra step on top of WildcardMatcher that the trie provides for "free."

3. No payload. WildcardMatcher returns the matching patterns, but not the workload group each belongs to. Using WildcardMatcher would require an additional map of pattern → workload group. The trie stores the workload groups as the value at each pattern key (`PatriciaTrie<Set<V>>`), so match and fetch happen in one lookup.

I lean towards the current Patricia trie implementation. I understand the preference for consistency across the project, but this case is not a direct fit.

dzane17: Trailing wildcard comes from the original WLM rules design: "rule matching will be per attribute trie (prefix tree)" #16888, #16797. I believe this was done for speed (using Trie which  only supports prefix matches) and tie-breaking simplicity.

```
Max WLM rules per domain = 500 (200 default)
Max index patterns per rule = 10
```

A search request index string can get matched against at most 5000 patterns. If all 5000 are complex glob patterns like `*-logs-*-books*` this could definitely add latency to lightweight searches. Benchmarking is important before lifting this requirement.

cwperks: ok, understood. I do think this is an area we can converge on in the future to make this experience consistent across the platform. Speaking for the security plugin, starting in 3.x it makes a tradeoff to have more latency when roles are created/updated vs latency at query time when evaluating permissions. At a high-level, it pre-computes a data structure like (index, action) -> roles that provide that combo. I think we can consider doing similar in WLM to ensure quick performance of computing matching query group(s).

cwperks: ok I see, I misunderstood this trie data structure and see how it relates to scoring now. 

btw I do still think WildcardMatcher would be suited here (in place of trie) and the same scoring logic could be used where exact match is score of 1 and otherwise its a value between 0 and 1 of how close the pattern from the rule matches the action value. With this trie in place, I think this PR lgtm.

cwperks (approved):

PR #22462, opened 2026-07-14
Fix non-deterministic aggregation profiler debug output under concurrent search

### LINKED ISSUE #22442: Aggregation profile debug output is non-deterministic under concurrent segment search
When concurrent segment search is enabled, the aggregation profiler produces arbitrary, non-deterministic values for the per-aggregator `debug` map. This has been a recurring source of CI flakiness (e.g. the `string profiler via global ordinals` and `numeric profiler` YAML tests in `rest-api-spec/.../search.aggregation/20_terms.yml`), most recently worked around in #20902 by relaxing the assertions rather than fixing the underlying behavior.

### Root cause

With concurrent search, a single segment's doc ID space is partitioned across multiple slices. Each slice gets its own aggregator instance with independent profiling counters, producing one `ProfileResult` per slice. `ConcurrentAggregationProfiler.reduceProfileResultsTree()` reduces these per-slice results into a single result, and it correctly aggregates timing and breakdown counts — but it does **not** merge the `debug` maps. Inside the per-slice loop it simply does:

```java
debug = profileResult.getDebugInfo();
```

so the reduced result keeps only whichever slice happened to be iterated last. Iteration order is governed by a `HashMap` (`getSliceLevelAggregationMap`) and is effectively arbitrary, so the reported `debug` values change from run to run.

Source: [`ConcurrentAggregationProfiler.java`](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/search/profile/aggregation/ConcurrentAggregationProfiler.java#L112)

### Observed impact

- `GlobalOrdinalsStringTermsAggregator` debug: `segments_with_single_valued_ords` and `segments_with_multi_valued_ords` reflect only one slice's segments instead of the shard total.
- Numeric terms aggregator debug: `total_buckets` reports one slice's bucket-ordinal count rather than the true total across slices.
- Any aggregator that reports numeric debug info (counts, cardinalities, thresholds) is affected. The values are not just imprecise — they are non-deterministic.
- Recurring CI flakiness; the assertions in `20_terms.yml` had to be weakened to `gte: 0` / `gte: 1`, which removes meaningful validation (e.g. the single-valued vs multi-valued ordinal distinction and the exact bucket count are no longer verified).

### Why there's no trivial fix

`debug` is an untyped `Map<String, Object>`. There is no metadata describing how a given key should combine across slices — a counter should be summed, a cardinality might need max or a proper cardinality merge, a threshold or strategy flag should be carried through unchanged, and some values are booleans or strings. A generic "merge the maps" is therefore unsafe.

### Proposed direction

Two options discussed on the PR:

1. **Preserve per-slice debug info (preferred by reviewers).** Since the Profile API is experimental (`@ExperimentalApi`), its shape can change. Under concurrent search, expose `debug` as a list of per-slice debug maps (or a `slices` sub-object) rather than a single map, so no information is discarded and consumers can see the true per-slice breakdown.
2. **Typed merge strategy.** Extend the debug-info API so aggregators declare a merge strategy per key (sum, max, last, cardinality-merge, passthrough, etc.), and have the reducer apply it. More invasive but keeps the single-map shape.

Whichever shape is chosen, the reducer must stop silently discarding all but the last slice's debug map, and the `20_terms.yml` assertions should be restored to meaningful checks (exact `total_buckets`, and the single/multi-valued ordinal distinction) once the output is deterministic.

### Additional context

- Follow-up to #20902 (which only relaxed the flaky assertions).
- Part of the concurrent search profiling effort, #14319.
- Example CI failure: https://build.ci.opensearch.org/job/gradle-check/74123/testReport/junit/org.opensearch.test.rest/ClientYamlTestSuiteIT/test__p0_search_aggregation_20_terms_numeric_profiler_/

RahilOp:
### Description

`ConcurrentAggregationProfiler.reduceProfileResultsTree()` correctly reduces timing and breakdown stats across slices, but for the per-aggregator `debug` map it simply kept whichever slice happened to be iterated last:

```java
debug = profileResult.getDebugInfo();
```

Since slice iteration order comes from a `HashMap` (`getSliceLevelAggregationMap`), the reported debug values (e.g. `total_buckets`, `segments_with_single_valued_ords`/`segments_with_multi_valued_ords`) changed arbitrarily from run to run — the root cause described in the issue.

**Chosen approach vs. the issue's proposed direction:** the issue suggests restructuring `debug` into a list of per-slice maps, reasoning that the Profile API is `@ExperimentalApi` and so free to reshape. I want to flag a discrepancy I found while implementing: `ProfileResult` (which owns the `debug` field) is actually annotated `@PublicApi(since = "1.0.0")`, not `@ExperimentalApi`. Reshaping `debug`'s JSON structure would therefore be a breaking change to a long-stable public field.

Digging into which debug keys are actually affected, they fall into two categories:
- Per-slice numeric counters (`total_buckets`, `segments_with_single_valued_ords`, `segments_with_multi_valued_ords`) — these are the ones that actually vary and cause the reported bug. Since each Lucene segment is assigned to exactly one slice under concurrent search, summing these across slices gives the correct, deterministic shard-level total with no double-counting.
- Static configuration flags (`collection_strategy`, `result_strategy`, `has_filter`, `deferred_aggregators`) — these reflect the aggregator's configured behavior and are identical on every slice already, so they were never actually a source of non-determinism, just incidentally overwritten by the same buggy assignment.

Given that, this PR takes a narrower fix: `mergeDebugInfo()` sums numeric values across slices and takes the first-seen value for everything else. This resolves the actual non-determinism without changing `debug`'s existing JSON shape, so it's a lower-risk change than restructuring a `@PublicApi` field. Happy to switch to the per-slice-list approach if there's a reason the `@ExperimentalApi` framing was intentional that I'm missing (e.g. if there's a plan to reclassify this class) — let me know and I'll rework it.

I left the `20_terms.yml` YAML assertions (`gte: 0` / `gte: 1`) as-is rather than tightening them to exact values, since the actual slice count (and therefore whether e.g. `total_buckets` reflects 1 slice or is summed across N) depends on concurrent-search slicing decisions that vary with core count / segment layout across CI machines — I didn't want to trade one source of flakiness for another without being able to verify against a real multi-core CI run.

### Related Issues
Resolves #22442

### Testing

- Added unit tests in `ConcurrentAggregationProfilerTests` covering: numeric values summed across slices, determinism regardless of slice iteration order, static non-numeric values preserved, and empty debug maps handled correctly.
- Ran the full existing `search.aggregation/20_terms.yml` YAML REST suite (26 tests, including the three profiler tests this bug affects: `string profiler via global ordinals`, `string profiler via map`, `numeric profiler`) — all pass.
- `./gradlew :server:test --tests "*ConcurrentAggregationProfilerTests*"` — 8/8 pass (4 new + 4 existing, no regressions).
- `./gradlew :server:spotlessJavaCheck` — clean.

### Check List
- [ ] API changes companion pull request created, if applicable. (no API/wire shape change)
- [ ] Public documentation issue/PR created, if applicable. (internal bug fix, no user-facing behavior change beyond correctness)


RahilOp:
Thanks for the review — both addressed in `048f9d58`:
- Numeric values are now normalized to `Long` up front in `mergeDebugInfo()`, so a debug key's Java type no longer depends on how many slices reported it.
- Added an assertion in `mergeDebugValue()` that non-numeric values agree across slices, since the merge relies on that assumption.

RahilOp:
Gradle check caught a real regression from the last commit: normalizing every numeric debug value to `Long` broke `AggregationProfilerIT` under concurrent search (`testComplexProfile`, `testMultiLevelProfile`, `testMultiLevelProfileBreadthFirst`), all with `ClassCastException: Long cannot be cast to Integer` — `assertRemapTermsDebugInfo` unboxes `segments_with_single_valued_ords`/`segments_with_multi_valued_ords` with a narrowing `(int)` cast, and those fields are always populated as `Integer` by the aggregators.

Fixed in `bac94abe`: removed the up-front `Long` coercion, and `mergeDebugValue` now sums two `Integer` operands as `Integer` (matching what these two fields have always reported), only widening to `Long` when a value is already a wider `Number` — which is what `total_buckets` (populated as `long`) already was.

This also turned out to be existing evidence for the `@PublicApi`-vs-`@ExperimentalApi` question I raised in the PR description: real, existing test code already depends on these debug values keeping a stable Java type, which is exactly the kind of consumer-visible break that comes with treating `debug`'s value types as free to change.

Verified against the actual failing suite this time, not just unit tests: `AggregationProfilerIT` (18/18, all 3 previously-failing concurrent-search cases now pass) and the full `20_terms.yml` YAML suite (26/26), both clean-rebuilt end to end.

RahilOp:
Checking on the `gradle-check` failure for `cbdd2bfd`: I don't have access to the Jenkins console log (`build.ci.opensearch.org` returns 403 for external contributors), and the GitHub Actions side doesn't auto-extract the failing test name for this run either.

That commit only adds two unit tests — no production code changed. Locally, clean-rebuilt:
- `ConcurrentAggregationProfilerTests`: 12/12 pass
- The entire `search.profile.*` package (68 tests, includes the `AggregationProfilerIT` cases that failed two commits ago): 0 failures
- `search.aggregation/20_terms.yml` YAML suite: 26/26 pass
- `spotlessJavaCheck`: clean

Given the diff is test-only and everything touching this code passes locally, this looks like it's very likely an unrelated flaky test elsewhere in the full suite rather than something introduced by this commit. If a maintainer can see the actual failing test name (or trigger a rerun), happy to look into it further if it turns out to be related.

RahilOp:
Rebased onto latest  to retrigger  after the previous run timed out.

RahilOp:
The latest gradle-check run is now green. Could a maintainer please review when you have a moment?

PR #22464, opened 2026-07-14
Instrument legacy java.io file streams in java agent

cwperks:
## Description

The Java agent previously instrumented `java.nio.file.*` classes (`Files`, `FileChannel`, `FileSystemProvider`) but did not cover legacy `java.io` stream constructors. That left file access paths such as `URL.openStream()`, XML entity resolution, direct `FileInputStream` usage, and direct `FileOutputStream` usage outside the agent's file permission checks.

This change instruments legacy stream constructors so the Java agent enforces policy checks for both read and write access.

## Changes

- Add `FileInputStreamInterceptor` to enforce `FilePermission(..., "read")` on `FileInputStream(String|File)` construction.
- Add `FileOutputStreamInterceptor` to enforce `FilePermission(..., "write")` on `FileOutputStream(String|File)` construction.
- Register both interceptors using named type matching for bootstrap classes.
- Switch agent redefinition strategy to `RETRANSFORMATION` and disable class format changes for already-loaded bootstrap classes.
- Normalize legacy stream paths before permission checks.
- Treat the exact built-in class-resource loading path as a privileged boundary so sandboxed code can trigger lazy class loading without gaining direct file-read access.
- Pass each Gradle test task's JaCoCo output directory through the existing `jacoco.dir` test-policy grant.
- Add positive and negative coverage for direct streams, file URLs, XML external entities, and the class-loader boundary.

## Testing

```bash
./gradlew :libs:opensearch-agent-sm:bootstrap:check :libs:opensearch-agent-sm:agent:check

JAVA25_HOME="$HOME/.sdkman/candidates/java/25.0.2-amzn" \
  ./gradlew ':modules:lang-painless:test' \
  -Dtests.security.manager=true \
  -Druntime.java=25
```

The Jenkins `AdditionTests.testDef` seed from build 82119 was also reproduced before the fix and passed afterward.

### DIFF
--- a/gradle/code-coverage.gradle
+++ b/gradle/code-coverage.gradle
@@ -34,6 +34,11 @@ repositories {
 allprojects {
   plugins.withId('jacoco') {
     jacoco.toolVersion = '0.8.14'
+
+    tasks.withType(Test).configureEach {
+      // BootstrapForTesting uses this property to grant the JaCoCo agent access to its execution-data directory.
+      systemProperty 'jacoco.dir', layout.buildDirectory.dir('jacoco').get().asFile.absolutePath
+    }
   }
 }
 
--- a/libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
+++ b/libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
@@ -12,6 +12,8 @@ import org.opensearch.javaagent.bootstrap.internal.SubjectInterceptor;
 
 import javax.security.auth.Subject;
 
+import java.io.File;
+import java.io.FileInputStream;
 import java.lang.instrument.Instrumentation;
 import java.net.Socket;
 import java.nio.channels.FileChannel;
@@ -79,6 +81,7 @@ public class Agent {
         final Junction<TypeDescription> pathType = ElementMatchers.isSubTypeOf(Files.class);
         final Junction<TypeDescription> fileChannelType = ElementMatchers.isSubTypeOf(FileChannel.class);
         final Junction<TypeDescription> fileSystemProviderType = ElementMatchers.isSubTypeOf(FileSystemProvider.class);
+        final Junction<TypeDescription> fileInputStreamType = ElementMatchers.named("java.io.FileInputStream");
 
         final AgentBuilder.Transformer socketTransformer = (b, typeDescription, classLoader, module, pd) -> b.visit(
             Advice.to(SocketChannelInterceptor.class)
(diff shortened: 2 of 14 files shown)


server/src/main/resources/org/opensearch/bootstrap/security.policy
```diff
@@ -275,6 +275,9 @@ grant {
   // needed by RestClientBuilder
   permission java.io.FilePermission "${java.home}/lib/security/cacerts", "read";
   permission java.io.FilePermission "${java.home}/lib/security/jssecacerts", "read";
+  permission java.io.FilePermission "${java.home}/lib/security/blocked.certs", "read";
```
cwperks, line 278: @reta ty for all the reviews lately. I came across this issue in the agent recently and would like to get your thoughts on this PR.

btw I noticed we have some paths referencing `${java.home}` across the policy files. Is there a way we can do this from the agent itself? These weren't needed back with JSM.

libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
```diff
@@ -79,6 +80,8 @@ private static AgentBuilder createAgentBuilder() throws Exception {
         final Junction<TypeDescription> pathType = ElementMatchers.isSubTypeOf(Files.class);
         final Junction<TypeDescription> fileChannelType = ElementMatchers.isSubTypeOf(FileChannel.class);
         final Junction<TypeDescription> fileSystemProviderType = ElementMatchers.isSubTypeOf(FileSystemProvider.class);
+        final Junction<TypeDescription> fileInputStreamType = ElementMatchers.named("java.io.FileInputStream");
```
reta, line 83: @cwperks I don't think we should instrument those, they are explicitly listed as forbidden APIs [1] and should not pass build checks. Or I am missing something?

[1] https://github.com/opensearch-project/OpenSearch/blob/main/buildSrc/src/main/resources/forbidden/jdk-signatures.txt#L28

libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
```diff
@@ -89,19 +92,40 @@ private static AgentBuilder createAgentBuilder() throws Exception {
...
+        );
+
         final AgentBuilder.Transformer subjectTransformer = (b, typeDescription, classLoader, module, pd) -> b.method(
             ElementMatchers.named("getSubject")
         ).intercept(MethodDelegation.to(SubjectInterceptor.class));
 
         final ByteBuddy byteBuddy = new ByteBuddy().with(Implementation.Context.Disabled.Factory.INSTANCE);
         var builder = new AgentBuilder.Default(byteBuddy).with(AgentBuilder.InitializationStrategy.NoOp.INSTANCE)
-            .with(AgentBuilder.RedefinitionStrategy.REDEFINITION)
+            .with(AgentBuilder.RedefinitionStrategy.RETRANSFORMATION)
```
reta: I think it does not matter for this particular change for redef vs retransform, right? The latter looks preferable as per https://github.com/raphw/byte-buddy/issues/676, probably would be good to have a comment

libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
andrross: @reta There's still value in preventing a dependency from a accessing these code paths, right? i.e. a hypothetical log4j bug that exposes a code path that reads files using FileInputStream.

libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/FileInputStreamInterceptor.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        final StackWalker walker = StackWalker.getInstance(StackWalker.Option.RETAIN_CLASS_REFERENCE);
+        final List<StackWalker.StackFrame> frames = walker.walk(BuiltinClassLoaderResourceBoundary.INSTANCE);
+
+        // Loading a class from a class-path directory is infrastructure performed by the built-in class loader.
+        // The protection domain that triggered lazy class loading must not be treated as directly reading the
+        // underlying .class file. Keep this boundary deliberately narrow so direct FileInputStream, URL, and XML
+        // reads from an untrusted domain continue through the policy check below.
+        if (BuiltinClassLoaderResourceBoundary.matches(frames)) {
```
reta, line 67: So the `FileInterceptor` does quite a few checks (filesystem provider, caller frames) to handle similar cases, could we reuse that instead?

gradle/code-coverage.gradle
```diff
@@ -34,6 +34,11 @@ repositories {
 allprojects {
   plugins.withId('jacoco') {
     jacoco.toolVersion = '0.8.14'
+
+    tasks.withType(Test).configureEach {
```
reta, line 38: 👍

server/src/main/resources/org/opensearch/bootstrap/security.policy
reta: > [@reta](https://github.com/reta) ty for all the reviews lately. I came across this issue in the agent recently and would like to get your thoughts on this PR.

Thanks @cwperks , those seems to be a new files. We probably could do this inside the agent, if that becomes painful to maintain, indeed.

libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
reta: Thanks @andrross , yeah, talked to Craig offline, got the context

libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/Agent.java
cwperks: Restored this to redefinition.

libs/agent-sm/agent/src/main/java/org/opensearch/javaagent/FileInputStreamInterceptor.java
cwperks: Looking for opportunities for re-use now...

PR #22477, opened 2026-07-15
Add query_string nesting depth limit

finnegancarroll:
## Description

Adds a new dynamic cluster setting `search.query.max_query_nesting_depth` (default 200) that pre-scans query strings for parenthesis nesting depth before delegating to Lucene's parser. Queries exceeding the limit are rejected with a `ParseException`.

## Testing

Unit tests and integ tests for deeply nested queryies.

### DIFF
--- a/server/src/test/java/org/opensearch/index/search/QueryStringNestedParenthesesOverflowTests.java
+++ b/server/src/test/java/org/opensearch/index/search/QueryStringNestedParenthesesOverflowTests.java
@@ -1,0 +1,133 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.index.search;
+
+import org.apache.lucene.queryparser.classic.ParseException;
+import org.apache.lucene.search.Query;
+import org.opensearch.index.query.QueryShardContext;
+import org.opensearch.test.OpenSearchSingleNodeTestCase;
+
+import static org.hamcrest.Matchers.containsString;
+
+/**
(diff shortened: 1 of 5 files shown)


finnegancarroll:
Hi @cwperks @nibix , i'm trying to guard against a stack overflow which can be triggered in lucene query parser. I'm thinking the long term fix should come in lucene but in the meantime OpenSearch can introduce a nesting depth limit. Can I get your thoughts on these changes when convenient?

DarshitChanpura (approved):

cwperks:
> Hi @cwperks @nibix , i'm trying to guard against a stack overflow which can be triggered in lucene query parser. I'm thinking the long term fix should come in lucene but in the meantime OpenSearch can introduce a nesting depth limit. Can I get your thoughts on these changes when convenient?

sorry Finn, I missed this message. I will take a look ASAP.

cwperks (approved):

cwperks:
Changes lgtm, ty @finnegancarroll

server/src/test/java/org/opensearch/index/search/QueryStringNestedParenthesesOverflowTests.java
```diff
@@ -0,0 +1,133 @@
...
+
+        // Should parse without error since parens are inside quotes
+        Query result = parser.parse(sb.toString());
+        assertNotNull("Parentheses inside quotes should not count toward nesting depth", result);
+    }
+
+    /**
+     * Verifies that the helper method correctly computes nesting depth.
+     */
+    public void testMaxParenthesisNestingDepthCalculation() {
```
cwperks, line 112: I do think some more permutations within this test could be useful, but this is a good base.

cwperks:
@finnegancarroll can you create a backport for this?

PR #22479, opened 2026-07-16
Wire request_timeout into S3 sync client

Sukriti0717:
### Description
The S3 repository plugin's `request_timeout` setting (default 5m) was only applied to the async S3 client. The sync client—used by all snapshot finalize and delete paths—had no call-level timeout, allowing requests to hang indefinitely on a degraded store.

This adds `apiCallAttemptTimeout` to the sync client's `ClientOverrideConfiguration`, mirroring what the async client already does.

### Testing
  - `S3ClientSettingsTests.testRequestTimeoutAppliedToSyncClientOverrideConfiguration`: verifies default (5m) and custom timeout values.
  - Existing repository-s3 test suites pass unchanged.

### Check List

### DIFF
--- a/plugins/repository-s3/src/internalClusterTest/java/org/opensearch/repositories/s3/S3SyncClientRequestTimeoutIT.java
+++ b/plugins/repository-s3/src/internalClusterTest/java/org/opensearch/repositories/s3/S3SyncClientRequestTimeoutIT.java
@@ -1,0 +1,207 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.repositories.s3;
+
+import com.carrotsearch.randomizedtesting.annotations.ThreadLeakFilters;
+import com.carrotsearch.randomizedtesting.annotations.ThreadLeakScope;
+
+import com.sun.net.httpserver.HttpServer;
+
+import org.opensearch.action.admin.cluster.repositories.put.PutRepositoryRequest;
+import org.opensearch.action.admin.cluster.snapshots.create.CreateSnapshotResponse;
+import org.opensearch.cluster.metadata.IndexMetadata;
+import org.opensearch.common.SuppressForbidden;
+import org.opensearch.common.network.InetAddresses;
+import org.opensearch.common.settings.MockSecureSettings;
+import org.opensearch.common.settings.Setting;
+import org.opensearch.common.settings.Settings;
+import org.opensearch.plugins.Plugin;
+import org.opensearch.repositories.s3.utils.AwsRequestSigner;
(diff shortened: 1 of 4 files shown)


shourya035:
Let's have an integration test for this change as well

plugins/repository-s3/src/internalClusterTest/java/org/opensearch/repositories/s3/S3SyncClientRequestTimeoutIT.java
```diff
@@ -0,0 +1,234 @@
...
+                .cluster()
+                .prepareCreateSnapshot("test-repo", "test-snap")
+                .setWaitForCompletion(true)
+                .setIndices(index)
+                .get();
+
+            SnapshotState state = response.getSnapshotInfo().state();
+            assertTrue(
+                "Snapshot should fail due to request timeout, but got state: " + state,
+                state == SnapshotState.FAILED || state == SnapshotState.PARTIAL
```
shourya035: This should be `FAILED` only? Is there any reason why it would move to `PARTIAL` state?

plugins/repository-s3/src/internalClusterTest/java/org/opensearch/repositories/s3/S3SyncClientRequestTimeoutIT.java
```diff
@@ -0,0 +1,234 @@
...
+                cause = cause.getCause();
+            }
+            assertTrue("Expected timeout-related exception but got: " + e.getMessage(), foundTimeout);
+        }
+
+        assertThat("Expected requests to the delayed S3 handler", delayedRequestCount.get(), greaterThan(0));
+    }
+
+    @SuppressForbidden(reason = "this test uses a HttpServer to emulate an S3 endpoint")
+    private static class DelayedS3HttpHandler extends S3HttpHandler {
```
shourya035: Let's move this to a helper class? This can be re-used across multiple other tests that we might introduce in the future

plugins/repository-s3/src/internalClusterTest/java/org/opensearch/repositories/s3/S3SyncClientRequestTimeoutIT.java
Sukriti0717: Right, no reason for PARTIAL, timeout hits during finalization so the snapshot always ends as FAILED. Fixed.

plugins/repository-s3/src/internalClusterTest/java/org/opensearch/repositories/s3/S3SyncClientRequestTimeoutIT.java
Sukriti0717: Done, moved to test/fixtures/s3-fixture alongside S3HttpHandler. Made it configurable for reuse.

shourya035 (approved):

linuxpi (approved):

PR #22482, opened 2026-07-16
ParquetDataFormat plugin - Fix document corruption bug due to dirty writes from rolled back document in VSR

### LINKED ISSUE #22417: [BUG] [DataFormatAwareEngine] Document corruption (edge case) after document rollback in CompositeWriter/ParquetWriter
### Describe the bug

**CompositeWriter reuses a Parquet writer with a corrupted VSR after a mid-document ingest failure**

When a document fails partway through ParquetWriter.addDoc (e.g., an Arrow OutOfMemoryException mid-VSRManager.addDocument), some field vectors have already been written for that row. CompositeWriter treats this as a recoverable per-doc failure, rolls back the primary, and keeps the same writer ACTIVE in the pool for the next document. But the rollback is a no-op (the row count never advanced), so the partially-written row is never scrubbed — and because CompositeWriter reuses that writer, the next document lands on the same VSR row and silently inherits the stale column values from the failed doc.

### Related component

Indexing

### To Reproduce

Steps to reproduce:

1. doc1 — indexes successfully with fields {clean, a_leak} → lands at row 0.
2. doc2 — sets {a_leak, z_big} and fails midway through field writing, after a_leak is written into the row but before the row is completed. The doc is rejected (recoverable per-doc failure), the writer is rolled back but stays ACTIVE, and the stale a_leak value remains in the (uncommitted) row.
3. doc3 — indexes successfully but sets only {clean} (does not set a_leak) → reuses the same row slot.
4. Flush and read the Parquet columns: doc3's row shows a_leak = doc2's value even though doc3 never set it → corruption.

Concrete IT simulating this issue - https://github.com/rayshrey/OpenSearch/commit/1bb43a4c98732b8068c7c29d88ebe3ce35053064

Verified in the above IT that doc3 (clean=d3) reads back a_leak=999999999 from the failed doc2, though doc3 never set it. (The test induces the mid-document failure via an ingest-pool limit, but any partial-write failure produces the same result.)

### Expected behavior

There shouldn't be any corruption in the data.

**Suggested Fix**

Before CompositeWriter returns the writer to the pool for reuse after a per-doc failure, ensure the partial row is scrubbed — either have the composite rollback path reset the current row index across all vectors (setNull + reset lastSet for variable-width vectors), or retire/replace the writer instead of reusing it (as is already done for the irreversible Lucene-secondary failure case).

### Additional Details

N/A

rayshrey:
### Description

### Related Issues
Resolves #22417 

### Check List


sandbox/plugins/parquet-data-format/src/main/java/org/opensearch/parquet/vsr/VSRManager.java
```diff
@@ -209,41 +217,103 @@ public void addDocument(ParquetDocumentInput doc) throws IOException {
...
-                // Defense-in-depth: schema reconciliation is supposed to happen in
-                // ParquetWriter.updateMappingVersion before any addDocument with a new
-                // field type. If we still see an unmapped type here, the writer is
-                // out of sync with the mapping — surface as a recoverable failure.
-                // TODO:: we can remove this post the validation on mapping update
-                throw new MismatchedInputException(
-                    "No ParquetField mapping for field [" + fieldType.name() + "] of type [" + fieldType.typeName() + "]"
-                );
+        final int rowIndex = activeVSR.getRowCount();
+        final List<FieldVector> writtenVectors = new ArrayList<>();
```
mgodwan: instead of always collecting writtenVectors, can we not just not re iterate only on failure?

mgodwan (approved):

PR #22485, opened 2026-07-16
Add virtual thread-per-task executor support

andrross:
Introduce plumbing for thread pools backed by virtual threads:

- Add ContextPreservingExecutorService, an ExecutorService wrapper that preserves ThreadContext across all task submission paths.
- Add OpenSearchExecutors.newVirtualThreadPerTaskExecutor to create a context-preserving virtual thread-per-task executor with named threads.
- Add the VIRTUAL thread pool type to ThreadPool. It is serialized as SCALING when communicating with nodes older than 3.8.0, which do not know about the "virtual" type.
- Generalize ThreadPool shutdown handling to cover any executor type other than the direct executor.

Nothing in this code base uses virtual threads yet; this is groundwork for future efforts to build on VTs.

### Check List

### DIFF
--- a/server/src/main/java/org/opensearch/common/util/concurrent/ContextPreservingExecutorService.java
+++ b/server/src/main/java/org/opensearch/common/util/concurrent/ContextPreservingExecutorService.java
@@ -1,0 +1,81 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * The OpenSearch Contributors require contributions made to
+ * this file be licensed under the Apache-2.0 license or a
+ * compatible open source license.
+ */
+
+package org.opensearch.common.util.concurrent;
+
+import org.opensearch.common.SuppressForbidden;
+
+import java.util.List;
+import java.util.concurrent.AbstractExecutorService;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * An {@link ExecutorService} wrapper that preserves {@link ThreadContext} across task submissions.
+ *
+ * @opensearch.internal
+ */
+class ContextPreservingExecutorService extends AbstractExecutorService {
+
+    private final ExecutorService delegate;
+    private final ThreadContext threadContext;
+
+    @SuppressForbidden(reason = "properly rethrowing errors, see OpenSearchExecutors.rethrowErrors")
+    ContextPreservingExecutorService(ExecutorService delegate, ThreadContext threadContext) {
+        super();
+        this.delegate = delegate;
+        this.threadContext = threadContext;
+    }
+
+    @Override
+    public void execute(Runnable command) {
+        final Runnable contextPreserving = threadContext.preserveContext(command);
+        delegate.execute(() -> {
+            try {
+                contextPreserving.run();
+                // Tasks submitted via submit()/invokeAll()/invokeAny() are RunnableFutures that capture any thrown
+                // Error instead of letting it propagate, so unwrap and rethrow it here the way
+                // OpenSearchThreadPoolExecutor does.
+                OpenSearchExecutors.rethrowErrors(threadContext.unwrap(contextPreserving));
+            } finally {
+                onTaskFinished();
+            }
+        });
+    }
+
+    /**
+     * Invoked on the executing thread once a task has finished, whether it completed normally or threw. Rethrowing a
+     * fatal {@link Error} happens before this, so it is called from a finally block. Does nothing by default.
+     */
+    protected void onTaskFinished() {}
+
+    @Override
+    public void shutdown() {
+        delegate.shutdown();
+    }
+
+    @Override
+    public List<Runnable> shutdownNow() {
+        return delegate.shutdownNow();
+    }
+
+    @Override
+    public boolean isShutdown() {
+        return delegate.isShutdown();
+    }
+
+    @Override
+    public boolean isTerminated() {
+        return delegate.isTerminated();
+    }
+
+    @Override
+    public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
+        return delegate.awaitTermination(timeout, unit);
+    }
+}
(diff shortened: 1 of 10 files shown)


andrross:
FYI @rishabhmaurya

server/src/main/java/org/opensearch/common/util/concurrent/OpenSearchExecutors.java
```diff
@@ -264,6 +265,20 @@ public static OpenSearchThreadPoolExecutor newAutoQueueFixed(
         );
     }
 
+    /**
```
rishabhmaurya, line 268: Could you also expose the underlying named virtual-thread `ThreadFactory`? Call sites could own the thread lifecycle directly (`start() / interrupt() / join(), cooperative shutdown, restart-in-place`) and so can't use an `ExecutorService`. 

This mirrors the existing platform-thread API: `OpenSearchExecutors.daemonThreadFactory(name)` already exposes a named platform-thread factory for exactly these lifecycle-owning cases, and every platform pool (`newScaling/newFixed/…`) is built from a ThreadFactory. The virtual side could be symmetric too:

```java
  public static ThreadFactory virtualThreadFactory(String name) {
      return Thread.ofVirtual().name(name + "#", 0).factory();
  }
  public static ExecutorService newVirtualThreadPerTaskExecutor(String name, ThreadContext ctx) {
       return new ContextPreservingExecutorService(Executors.newThreadPerTaskExecutor(virtualThreadFactory(name)), ctx);
}
```

server/src/main/java/org/opensearch/common/util/concurrent/ContextPreservingExecutorService.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * An {@link ExecutorService} wrapper that preserves {@link ThreadContext} across task submissions.
+ * Each submitted task is wrapped via {@link ThreadContext#preserveContext(Runnable)} so that the
+ * thread context at submission time is restored on the executing thread.
+ *
+ * @opensearch.internal
+ */
+class ContextPreservingExecutorService implements ExecutorService {
```
rishabhmaurya: we could instead do this right?
```java
  class ContextPreservingExecutorService extends AbstractExecutorService {

      private final ExecutorService delegate;
      private final ThreadContext threadContext;

      ContextPreservingExecutorService(ExecutorService delegate, ThreadContext threadContext) {
          this.delegate = delegate;
          this.threadContext = threadContext;
      }

      @Override
      public void execute(Runnable command) {
          delegate.execute(threadContext.preserveContext(command));   // single context-preservation point
      }

      // submit(Runnable), submit(Callable), invokeAll, invokeAny are inherited from
      // AbstractExecutorService and funnel through execute() above.

      @Override public void shutdown()              { delegate.shutdown(); }
      @Override public List<Runnable> shutdownNow() { return delegate.shutdownNow(); }
      @Override public boolean isShutdown()         { return delegate.isShutdown(); }
      @Override public boolean isTerminated()       { return delegate.isTerminated(); }
      @Override public boolean awaitTermination(long timeout, TimeUnit unit) throws InterruptedException {
          return delegate.awaitTermination(timeout, unit);
      }
  }
```
This removes the Callable-specific preserveContext helper and all five extra overrides, reuses the existing preserveContext primitive instead of a second capture, and makes every submission path behave matching `OpenSearchThreadPoolExecutor`.

server/src/main/java/org/opensearch/threadpool/ThreadPool.java
rishabhmaurya: stats() only extracts metrics when the executor is an OpenSearchThreadPoolExecutor. `newVirtualThreadPerTaskExecutor` returns a `ContextPreservingExecutorService`, so a registered VIRTUAL pool falls through and reports all -1 (`threads/queue/active/completed`) in thread_pool stats. If VIRTUAL pools are meant to be observable in _nodes/stats, they need a stats path; if not, worth a note that they're intentionally invisible there.

server/src/main/java/org/opensearch/threadpool/ThreadPool.java
```diff
@@ -707,9 +707,8 @@ public void shutdown() {
         stopCachedTimeThread();
         scheduler.shutdown();
         for (ExecutorHolder executor : executors.values()) {
-            ExecutorService es = executor.executor();
-            if (es instanceof ThreadPoolExecutor || es instanceof ForkJoinPool) {
-                es.shutdown();
+            if (executor.executor() != DIRECT_EXECUTOR) {
```
rishabhmaurya, line 720: A virtual-thread-per-task executor's shutdown() does not interrupt running tasks. For long-lived blocking VTs (e.g. a stream drain parked in a blocking take()), shutdown() + awaitTermination() hangs until the task returns on its own; only shutdownNow() interrupts. Worth documenting that a VIRTUAL pool carrying blocking tasks needs shutdownNow() (or explicit
  cancellation), else node shutdown can block.

server/src/main/java/org/opensearch/threadpool/ThreadPool.java
```diff
@@ -718,17 +717,16 @@ public void shutdownNow() {
         stopCachedTimeThread();
         scheduler.shutdownNow();
         for (ExecutorHolder executor : executors.values()) {
-            ExecutorService es = executor.executor();
-            if (es instanceof ThreadPoolExecutor || es instanceof ForkJoinPool) {
-                es.shutdownNow();
+            if (executor.executor() != DIRECT_EXECUTOR) {
```
rishabhmaurya, line 731: i hope changing from allow to denylist here isn't risky?
also, maybe we can use `executor.info().getThreadPoolType() != ThreadPoolType.DIRECT`

andrross:
@rishabhmaurya I addressed your comments, as well as expanded this PR to add a `VirtualExecutorBuilder` which allows plugins to define a VT-based executor. It also reports stats on active and completed (the pool-based stats are not relevant).

rishabhmaurya (approved):

PR #22487, opened 2026-07-16
Add API to modify a data stream's backing indices

### LINKED ISSUE #8271: Ability to add a backing index to a data stream
It doesn't seem to be possible from what I can see. We had an issue recently where we tried to restore several datastream indices that were in cold storage back to warm all at once. Not all of them were added back to be a backing index for the datastream, and it's unclear how I can manually add them.

elasticsearch has this API for reference https://www.elastic.co/guide/en/elasticsearch/reference/current/modify-data-streams-api.html

I would like to be able to add backing indices to a datastream.

**Describe alternatives you've considered**
None

**Additional context**
image

jainankitk:
### Description
Adds `POST /_data_stream/_modify`, a metadata-only API to add or remove backing indices of a data stream. All actions in a request are applied atomically in a single cluster-state update and never create, delete, restore, open, close, or relocate shards.
  
  The request body is a list of actions:

      { "actions": [
          { "remove_backing_index": { "data_stream": "logs-foo", "index": ".ds-logs-foo-000001" } },
          { "add_backing_index":    { "data_stream": "logs-foo", "index": "legacy-logs-2023" } }
      ] }
  
  Behavior and guardrails:

  * Multiple add/remove actions per request, composed atomically.
  * Generation is derived from the backing indices (highest counter = write index), never set directly.
  * The write index cannot be removed, and removing the last backing index is rejected.
  * Added indices need not follow the `.ds-<data_stream>-NNNNNN` convention — arbitrary-named indices are attached as older, non-write backing indices, which supports migrating pre-existing regular indices into a data stream. The convention-following write index is preserved so rollover keeps working.
  * Attached indices are marked hidden to match all other backing indices; adding an index already in the stream is a no-op; adding one that belongs to another data stream is rejected.

  `MetadataDataStreamsService` is injectable, so components such as cross-cluster replication can call it directly instead of routing through the transport action.

  ### Related Issues

  Resolves #8271
  
  ### Check List

  * [x] Functionality includes testing.
  * [ ] API changes companion pull request created, if applicable.
  * [ ] Public documentation issue/PR created, if applicable.
  

andrross:
```
REPRODUCE WITH: ./gradlew ':client:rest-high-level:test' --tests 'org.opensearch.client.RestHighLevelClientTests.testApiNamingConventions' -Dtests.seed=2B9A3FC53E859617 -Dtests.security.manager=true -Dtests.jvm.argline="-XX:TieredStopAtLevel=1 -XX:ReservedCodeCacheSize=64m" -Dtests.locale=sr-Cyrl-RS -Dtests.timezone=Atlantic/Bermuda -Druntime.java=25

RestHighLevelClientTests > testApiNamingConventions FAILED
    java.lang.AssertionError: Some API are not supported but they should be: [indices.modify_data_stream]
    Expected: <0>
         but: was <1>
        at __randomizedtesting.SeedInfo.seed([2B9A3FC53E859617:344A7240E705C0C6]:0)
        at org.apache.lucene.tests.util.LuceneTestCase.assertThat(LuceneTestCase.java:2123)
        at org.opensearch.client.RestHighLevelClientTests.testApiNamingConventions(RestHighLevelClientTests.java:975)
```

server/src/main/java/org/opensearch/cluster/metadata/MetadataDataStreamsService.java
```diff
@@ -0,0 +1,303 @@
...
+        // The index must follow the data stream's backing-index naming convention (.ds-<dataStream>-NNNNNN). This keeps
+        // the backing-index set consistent with the generation, which is derived from the highest counter below.
+        long addedCounter = parseBackingIndexCounter(dataStream.getName(), indexName);
+
+        // Backing indices are hidden; mark the newly attached index hidden if it is not already, matching how data
+        // stream creation and rollover create backing indices.
+        if (IndexMetadata.INDEX_HIDDEN_SETTING.get(indexMetadata.getSettings()) == false) {
+            IndexMetadata hiddenIndexMetadata = IndexMetadata.builder(indexMetadata)
+                .settings(Settings.builder().put(indexMetadata.getSettings()).put(IndexMetadata.SETTING_INDEX_HIDDEN, true))
+                .settingsVersion(indexMetadata.getSettingsVersion() + 1)
```
andrross: What about the metadata version? It should probably be bumped here too.

server/src/main/java/org/opensearch/action/admin/indices/datastream/DataStreamAction.java
```diff
@@ -0,0 +1,192 @@
...
+                    return REMOVE_BACKING_INDEX;
+                default:
+                    throw new IllegalArgumentException("no data stream action type for [" + value + "]");
+            }
+        }
+    }
+
+    private final Type type;
+    private final String dataStream;
+    private String index;
```
andrross: Can this be `final`? If so, I'd consider making this a Java `record` since it's a pure data class.

andrross (commented):
> Added indices must follow the .ds-<data_stream>-NNNNNN naming convention

Why must indices follow the name convention? Is it not valid to add arbitrary indexes to a data stream? This would allow for use cases where you have implemented a data stream pattern on regular indexes but want to move to data streams.

server/src/main/java/org/opensearch/cluster/metadata/MetadataDataStreamsService.java
jainankitk: Good catch! Switched to the `put(IndexMetadata.Builder)` which auto-increments the index version.

server/src/main/java/org/opensearch/action/admin/indices/datastream/DataStreamAction.java
jainankitk: Updated this class to be a Java `record`

jainankitk:
> Why must indices follow the name convention? Is it not valid to add arbitrary indexes to a data stream?

I was slightly confused whether those arbitrary indices can be write target as well and how to handle them. But, I guess they would be read-only, and the writes always go through the latest generated data stream index.

jainankitk:
Unrelated flaky test failure:

```
java.lang.AssertionError: Failed to get all expected responses: 150 left
	at __randomizedtesting.SeedInfo.seed([2124928713CBBF8B:650D3EDBA12F4ED1]:0)
	at org.opensearch.http.netty4.Netty4HttpClient.sendRequests(Netty4HttpClient.java:273)
	at org.opensearch.http.netty4.Netty4HttpClient.processRequestsWithBody(Netty4HttpClient.java:245)
	at org.opensearch.http.netty4.Netty4HttpClient.post(Netty4HttpClient.java:212)
	at org.opensearch.http.netty4.Netty4HttpRequestSizeLimitIT.testLimitsInFlightRequests(Netty4HttpRequestSizeLimitIT.java:110)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at com.carrotsearch.randomizedtesting.RandomizedRunner.invoke(RandomizedRunner.java:1750)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$8.evaluate(RandomizedRunner.java:938)
	at com.carrotsearch.randomizedtesting.RandomizedRunner$9.evaluate(RandomizedRunner.java:974)
```

Retried gradle check

jainankitk:
Unrelated test failures again:

```
[Test Result](https://build.ci.opensearch.org/job/gradle-check/82414/testReport/) (2 failures / +1)

    [org.opensearch.http.netty4.Netty4HttpRequestSizeLimitIT.testLimitsInFlightRequests](https://build.ci.opensearch.org/job/gradle-check/82414/testReport/junit/org.opensearch.http.netty4/Netty4HttpRequestSizeLimitIT/testLimitsInFlightRequests/)
    [org.opensearch.cluster.decommission.DecommissionControllerTests.testTimesOut](https://build.ci.opensearch.org/job/gradle-check/82414/testReport/junit/org.opensearch.cluster.decommission/DecommissionControllerTests/testTimesOut/)
```

Retrying gradle check

jainankitk:
Another flaky test:

```
[Test Result](https://build.ci.opensearch.org/job/gradle-check/82437/testReport/) (1 failure / +1)

    [org.opensearch.upgrades.IndexingIT.testIndexingWithSegRep](https://build.ci.opensearch.org/job/gradle-check/82437/testReport/junit/org.opensearch.upgrades/IndexingIT/testIndexingWithSegRep/)
```

Retrying gradle check

andrross:
> > Why must indices follow the name convention? Is it not valid to add arbitrary indexes to a data stream?
> 
> I was slightly confused whether those arbitrary indices can be write target as well and how to handle them. But, I guess they would be read-only, and the writes always go through the latest generated data stream index.

@jainankitk  I think we need at a minimum to validate there is a `@timestamp` field as that is the critical bit of schema that data streams rely upon at search time, otherwise searches won't work. I think the validation needs to ensure a field with the name `dataStream.getTimeStampField().getName()` exists and has the right mapping type.

server/src/main/java/org/opensearch/cluster/metadata/MetadataDataStreamsService.java
```diff
@@ -0,0 +1,277 @@
...
+        Index index = indexMetadata.getIndex();
+        if (dataStream.getIndices().contains(index)) {
+            // Idempotent: already a member.
+            return dataStream;
+        }
+
+        // Backing indices are hidden; mark the index hidden if it is not already, matching data stream creation and
+        // rollover. put(IndexMetadata.Builder) bumps the index metadata version; the top-level Metadata version is
+        // bumped by the cluster-manager service on publish.
+        if (IndexMetadata.INDEX_HIDDEN_SETTING.get(indexMetadata.getSettings()) == false) {
```
andrross, line 196: Should the hidden setting be removed upon detach? That would mirror the add-upon-attach behavior.

jainankitk: That is something which came up during development, and decided to leave it as is. There isn't a clean way to tell whether the index was hidden before attach. Data stream index => hidden, but index not being part of data stream does not imply it is unhidden

jainankitk: Thinking through this bit more, leaving index hidden supposed to be visible is worse than making index visible when it is supposed to be hidden. Will make this change!

jainankitk:
Regarding the comments `Code-Diff-Analyser` comments: 

1. server/src/main/java/org/opensearch/snapshots/RestoreService.java at 671
> The auto-attach runs inside the restore cluster-state update, which already creates indices, and — with include_global_state — writes templates, persistent settings, and cluster customs, all under the single `cluster:admin/snapshot/restore` permission without dispatching separately-authorized sub-actions. Attaching a restored `.ds-<stream>-NNNNNN` index to its data stream is metadata of the same kind and strictly narrower than the global-state powers restore already holds, so it does not introduce a new authorization boundary. The alternative (restore the index but leave it orphaned outside the stream) is worse operationally.

2. server/src/main/java/org/opensearch/cluster/metadata/DataStream.java at 163
> The generation advance here is intentional, not an unguarded side effect. `addBackingIndex` orders backing indices by counter and derives the generation as the highest counter, so the write index (last element) and the generation are consistent by construction.
The "high-counter index promoted to write index" behavior is the point: it's how a restored/replicated rollover index is attached atomically (the follower's stream is a generation behind the leader). There's no path to promote a lower-counter index — promotion is purely by counter ordering, which is the data-stream definition of the write index.

server/src/main/java/org/opensearch/snapshots/RestoreService.java
```diff
@@ -667,6 +668,33 @@ public ClusterState execute(ClusterState currentState) {
                                 .map(ds -> updateDataStream(ds, mdBuilder, request))
                                 .collect(Collectors.toMap(DataStream::getName, Function.identity()))
                         );
+                        // Auto-attach any restored backing index (".ds-<streamName>-NNNNNN") to a pre-existing data stream
```
andrross: This is a significant change in behavior of this API that will break any user that relies on the current behavior to restore backing indexes as top-level indexes. A behavior change like this has to be opt-in to not be breaking. This also seems unrelated to the modify API and should be a separate PR.

Also should probably not mention the cross-cluster replication implementation since the comment will likely not be accurate over time.

jainankitk: Sure, I have pulled the `RestoreService` changes into a separate PR. Will address the cross-cluster replication comment as part of that PR

server/src/main/java/org/opensearch/action/admin/indices/datastream/DataStreamAction.java
```diff
@@ -0,0 +1,143 @@
...
+
+import java.io.IOException;
+import java.util.Objects;
+
+/**
+ * A single add- or remove-backing-index operation within a {@link ModifyDataStreamsAction} request.
+ *
+ * @opensearch.api
+ */
+@PublicApi(since = "3.8.0")
```
andrross, line 30: The api spec says `"stability":"experimental"`, this says Public. Why aren't they the same?

server/src/main/java/org/opensearch/cluster/metadata/DataStream.java
```diff
@@ -148,6 +148,61 @@ public DataStream removeBackingIndex(Index index) {
         return new DataStream(name, timeStampField, backingIndices, generation);
     }
 
+    /**
+     * Adds the given index to the data stream's backing indices and returns a new {@code DataStream} instance. Backing
+     * indices are ordered oldest-to-newest so the write index remains last: convention-named indices
+     * ({@code .ds-<name>-NNNNNN}) order by their counter, and non-convention names sort first. The generation is derived
+     * as the highest backing-index counter, so adding an index whose counter exceeds the current generation advances
+     * the generation and makes it the new write index; adding a lower-counter or non-convention index leaves the
```
andrross, line 156: > adding an index whose counter exceeds the current generation advances the generation and makes it the new write index

Is this true? MetadataDataStreamsService.applyAddBackingIndex says the added index is always the non-write index

andrross (approved):

server/src/main/java/org/opensearch/action/admin/indices/datastream/DataStreamAction.java
jainankitk: Good catch! Let me change this to `@ExperimentalApi`

server/src/main/java/org/opensearch/cluster/metadata/DataStream.java
jainankitk: Both are true.

From `MetadataDataStreamsService.applyAddBackingIndex` perspective, an index with generation id > data stream generation id could never have been created. So, it can never be write index.

`DataStream.addBackingIndex` is a general primitive which can advance generation (that capability exists for the restore path that will be in the followup PR).

server/src/main/java/org/opensearch/cluster/metadata/DataStream.java
```diff
@@ -148,6 +148,61 @@ public DataStream removeBackingIndex(Index index) {
...
+     * generation unchanged.
+     *
+     * @param index the backing index to add
+     * @return new {@code DataStream} instance with the index added
+     */
+    public DataStream addBackingIndex(Index index) {
```
sandeshkr419, line 162: Quick Question:

Are we validating any mappings compatibility before adding the index to data stream. Basically when you hit search on a data-stream, we shouldn't have a failing case because a field doesn't exists on few backing indices or the field types are different across different backing indices.

sandeshkr419 (commented):
Thanks for working on this feature, much needed.

Just quickly wanted to check if we have tested the scenarios:
- adding an already data-stream backed index to another data-stream
- incompatible mapping for an index to a data stream you are trying to attach. I glanced through tests & logic and could see timestamp field validations, but other fields also need to be compatible in some way.

server/src/main/java/org/opensearch/cluster/metadata/DataStream.java
jainankitk: The only mapping constraint being enforced by data streams is the existence of `@timestamp` field, which is being validated before adding the index to data stream

jainankitk:
> adding an already data-stream backed index to another data-stream

Yes, `testCannotAddSameIndexToTwoStreamsInOneRequest`, `testAddWithoutRemoveStillRejectedRegardlessOfOrder` tests covers this scenario.

server/src/main/java/org/opensearch/action/admin/indices/datastream/DataStreamAction.java
jainankitk: Resolved as part of https://github.com/opensearch-project/OpenSearch/pull/22539

PR #19116, opened 2025-08-20
Remove unnecessary looping in field data cache clear

peteralfonsi:
### Description
Previously, clearing a shard from the field data cache (which happens on shard closure like in a B/G, or during manual cache clear) unnecessarily looped through each of the N keys in the whole cache, once per field that uses field data on that index, which can potentially be very large. This takes O(FN) time for F the number of fields on that index. For real user scenarios this could get as long as minutes which caused node drops. 

This PR moves around the cleanup logic a bit, so that the IndicesFieldDataCache itself is the one that iterates through its keys when a cleanup is requested. This takes us from O(FN) to O(N) time to clear a shard.

In the next PR, the iteration itself will be moved to a scheduled cleaner thread, as in the request cache, and the clear() methods on IndicesFieldDataCache will mark those indices/fields for cleanup in that thread, rather than iterating through them then and there. I purposefully split these, because in the case of a whole-cache clean via API call, the existing code in the request cache actually has a similar issue, and I wanted to raise an issue to get feedback on that and then apply the same solution to both the FDC and the RC. 

A future PR will also remove the synchronized blocks in IndexFieldDataService. This depends on the scheduled cleaner thread to work without iteration issues on the cache keys iterator, so it's also separate.

Benchmark results: 
On a single-node OS cluster running on a c5.2xl EC2 instance, I set up indices with 999 field data-enabled fields each. 999 is the max number of fields per index without adjusting settings; it's reasonable to pick such a high value because the users who saw issues were actually using ~20k fields each. Then I searched each field for each index to populate the field data cache with many keys. Finally, I cleared one index (which has 1 shard) using the cache clear API and timed it on the client side (end-to-end). Note this hits the same code path as when the shard closes in a B/G. Scripts to do this are [here](https://github.com/peteralfonsi/misc-scripts/tree/main/cache-clear-timing/fielddata). 

|Number of indices | Baseline clear time (sec) | Contender clear time (sec) | 
|--|--|--|
|100|59.09|0.0993|
|1|0.0153|0.0186|

I purposefully did not time clearing the whole cache at once as this will be changed to the scheduled thread in the next PR. 

We see for 100 indices the speedup is very large. This is because N is very large so checking all N keys 999 times is slow. There isn't a speedup when there's just 1 index; I think the 0.01 sec is mostly overhead of OS processing the request rather than actual cache clear time. 

### Related Issues
Part of https://github.com/opensearch-project/OpenSearch/issues/13862

### Check List
- [x] Functionality includes testing.
[...]

PR #22491, opened 2026-07-17
Fix stale fielddata retention by invalidating exact key on reader close

jwils:
### Description
Since #19116, reader-close cleanup of fielddata cache entries has been deferred: `IndexFieldCache.onClose` only marked the reader's `CacheKey` in a `cacheKeysToClear` set, and the periodic cache cleaner later consumed the marks while scanning `cache.keys()` for matching entries.

That scan can silently miss entries. `Cache.keys()` walks the LRU linked list without holding `lruLock`, and its Javadoc states that iteration under concurrent mutation is undefined. The cleaner's `synchronized (this)` block only serializes concurrent scans — it does not stop search threads from reading the cache, and every cache hit relinks the accessed entry to the LRU head. If the entry the iterator is positioned on gets promoted mid-scan, `next = current.after` jumps the cursor back to the old head, skipping every entry in between.

Because the marks are removed from `cacheKeysToClear` *before* the scan and a reader closes exactly once, a dead entry skipped by that one scan is never marked again. It remains strongly reachable until size-based eviction happens to reach it, which on an under-capacity node is effectively forever.

The cache size limits the sum of accounted entry weights, not the complete object graph retained by those entries. A cached `GlobalOrdinalsIndexFieldData` strongly retains its leaf fielddata through `segmentAfd`, while its own `ramBytesUsed()` reports only the global ordinal map because the leaves are normally accounted as separate cache entries. If a stale global-ordinals entry is missed during reader-close cleanup, its leaf entries can later be removed normally and their bytes deducted from cache and circuit-breaker accounting. The leaf objects nevertheless remain strongly reachable through the stale global entry. Repeated reader turnover can therefore accumulate substantial heap usage beyond `indices.fielddata.cache.size` even while the cache remains within its accounted weight limit and fielddata breaker usage no longer reflects all of the retained objects.

This change restores the pre-#19116 behavior for the reader-close path: `onClose` invalidates the exact key synchronously. Exact-key invalidation is O(1), so there is nothing to gain from batching it into the scan — the batching in #19116 was only needed for the per-index/per-field clear paths, which previously looped over the whole cache and which this change leaves as-is. This also makes reader-close a reliable backstop for the remaining deferred paths: an entry missed by an index/field scan now lives only until its reader closes, instead of indefinitely.

The `assertBusy` in `AbstractStringFieldDataTestCase.testGlobalOrdinalsGetRemovedOnceIndexReaderCloses` is tightened to a direct assertion, since reader-close cleanup is synchronous again.

### Related Issues
N/A

### Check List

---
**Update Jul 20, 2026:** Clarified how retained heap can exceed the configured fielddata cache size.


andrross (approved):
Thanks @jwils. This looks good to me. My reading of the code is that you're right that this part was not critical for the performance issue that #19116 fixed. It would be nice to verify that though. Did you happen to run a benchmark similar to the setup described in #19116?

sgup432:
@jwils Thanks for the PR. I see your PR handles the case where stale segment reader entries are evicted instantly rather than being missed due to LRU reordering during iteration.

But the per-index and per-field still uses the async flow using the same `Cache.keys()` approach, so the same risk of missed entries remains there. 

I wonder whether it would be better for fielddata cache to create a snapshot of cache keys under the lock before iterating? That would address the issue for all cleanup paths, not just this one.

sgup432:
For additional context, the fix in https://github.com/opensearch-project/OpenSearch/pull/19116 was primarily made to avoid node-drop issues. Previously, all cleanups were done synchronously, and in some cases these ran on the cluster manager thread when an index was closed. This sometimes took too long when the cache had a large number of entries to iterate over, causing node-drop issues. That led us to move all cleanup paths to async for consistency (including the segment reader case), with the tradeoff that some stale entries might remain in the cache for longer(but still not overflowing the cache beyond its size limit).

jwils:
> Did you happen to run a benchmark similar to the setup described in https://github.com/opensearch-project/OpenSearch/pull/19116?

Quick script locally on my M3 Max mac, but should show that there isn't a regression in perf.

| Timing (client-side e2e) | Baseline | Contender (PR) |
|---|---|---|
| Hot index delete, median of 10 | 40.7 ms | 35.2 ms |
| Hot index delete, mean / min / max | 41.0 / 31.2 / 53.5 | 36.6 / 31.1 / 46.7 |
| Cold index delete (no fielddata), 2 samples | 55.6, 38.4 ms | 47.3, 31.1 ms |
| Per-index `_cache/clear?fielddata=true` | 1.7, 0.7 ms | 1.1, 0.3 ms |


> But the per-index and per-field still uses the async flow using the same Cache.keys() approach, so the same risk of missed entries remains there.

I want to be clear on this fix. The reader-close path was the permanent leak. If missed one pass it would never get closed  again. Since this remaining leak bounds to the reader lifetime, my preference would be to do a minimal fix first as I'd like to request a backport of this. I'm happy to do a followup with the larger fix. I think it might require a bit more testing to ensure it performs well.

> (but still not overflowing the cache beyond its size limit)

Understood, this is where we are seeing issues though. We're seeing 10s of GB of retained fielddata on our heap that never gets cleared. I'm optimistic that this was the issue causing it.

sandeshkr419:

sandeshkr419:

sgup432:
>I want to be clear on this fix. The reader-close path was the permanent leak. If missed one pass it would never get closed again. Since this remaining leak bounds to the reader lifetime, my preference would be to do a minimal fix first as I'd like to request a backport of this.

Got it. When you say that this is a permanent leak, I assume you mean such stale entries remain in the cache forever, never cleaned up and eventually taking cache's size beyond its max limit configured. If yes, then I do not think this is right IMO. Field data cache can occupy 35% of the heap by default and uses Cache.java as its underlying cache implementation as you know. Cache.java does track the max size of the cache, and once it is breached it uses LRU based policy to evict older items, so these stale entries will be cleared at some point. I think you can prove/reproduce this by writing a simple UT/IT for your case. I am happy to be proven wrong if thats not the case. 


>Understood, this is where we are seeing issues though. We're seeing 10s of GB of retained fielddata on our heap that never gets cleared. I'm optimistic that this was the issue causing it.

As mentioned above, field data cache is configured to take upto 35% of the heap and its not unusual to see this cache taking 10s of GB data. If you have a 64gb heap, this cache can take upto ~20gb. If you perform a lot of terms/keyword based aggregation, have fielddata enabled as true, this can happen pretty easily. This can probably be verified by calling `/_cat/fielddata` and see how much heap each field is taking up, and also in total. This stats would exclude the leaks, so might give you some idea as to what is going on.

I would like to know what was the heap configured on your node, and did field data cache size go beyond the configured limit(either 35% default or custom one)?

jwils:
The cache size limits the sum of accounted entry weights, not the complete object graph retained by those entries. A cached GlobalOrdinalsIndexFieldData strongly retains its leaf fielddata through segmentAfd, while its own ramBytesUsed() reports only the global ordinal map because the leaves are normally accounted as separate cache entries. If a stale global-ordinals entry is missed during reader-close cleanup, its leaf entries can later be removed normally and their bytes deducted from cache and circuit-breaker accounting. The leaf objects nevertheless remain strongly reachable through the stale global entry. Repeated reader turnover can therefore accumulate substantial heap usage beyond indices.fielddata.cache.size even while the cache remains within its accounted weight limit and fielddata breaker usage no longer reflects all of the retained objects.


In practice we see the field data cache only believing it retains 10s of MB of data, but the heap shows it is linked to GB of data. I imagine this is a issue for many users?

jwils:
> I would like to know what was the heap configured on your node, and did field data cache size go beyond the configured limit(either 35% default or custom one)?

Our field data cache is configured to max 20% of heap. Max heap is 45 gb. This means field data cache can get to 9GB before eviction. Largest field data cache is 181 MB through the api, so will never get to eviction threshold. 

Here is a graph after clearing the ~181 MB fielddata cache. You can probably tell where the cache clear happened.
Screenshot 2026-07-20 at 5 15 04 PM

sgup432:
Can we write some form of tests to reproduce this issue? This would help us validate with actual data.

jwils:
> Can we write some form of tests to reproduce this issue? This would help us validate with actual data.

Happy to do that. Can we do it post merge and get this merged? Given this is a real issue either way would love to get it deployed as soon as possible.

sgup432:
>Happy to do that. Can we do it post merge and get this merged? Given this is a real issue either way would love to get it deployed as soon as possible.

From my perspective, a reproducible test is a must-have, it lets us prove that our assumptions are right and gives us confidence that the fix is correct. 
I believe it should be quick?

jwils:
> gives us confidence that the fix is correct.

Included [here](https://github.com/opensearch-project/OpenSearch/pull/22491/changes#diff-cc09f66510ae3b5c2e1b8879ee15ba9b29983c8900dc92dc2ab8883df464bc8bR77).

sandeshkr419:

jwils:
I have opened https://github.com/opensearch-project/OpenSearch/issues/22522 to clearly state the problem that I am attempting to fix. I am very optimistic this fixes the bug, but if not then there is some other data leak.

sgup432 (approved):

sgup432:
@andrross Can you help with merging this change?

jwils:
@andrross Is it possible to backport this fix to the 3.3 branch? Anything you need from me to do that?

jwils:
Sharing the production results here as well (posted on the other PR first). After ~2 days deployed we see no indication of a leak. Unexpectedly, we're also seeing ~20% lower steady-state old-gen heap usage (first annotation is a cache clear, second is the deploy).

The second chart compares fielddata memory under production query load: the patched cluster holds steady around ~50 MB, while stock 3.3 shows a sawtooth between ~70–100 MB. The accounted fielddata size difference is far too small to explain the old-gen improvement on its own. Our working theory is that entries surviving until the sweep retain significantly more heap than they account for.

Since invalidation now happens synchronously on reader close, we also watched for write-path overhead: average indexing latency is unchanged (maybe slightly improved) across the deploy.

Finally, Refresh latency (delta of refresh.total_time / delta of refresh.total from node stats) is unchanged across the deploy at ~65ms per refresh, so the synchronous invalidation isn't adding measurable cost to the refresh path.


Screenshot 2026-07-24 at 9 30 17 AM


fielddata_steady_compare


Screenshot 2026-07-24 at 9 39 52 AM

Screenshot 2026-07-24 at 9 50 20 AM

PR #22498, opened 2026-07-18
Invalidate exact bitset filter cache keys on reader close

jwils:
### Description

The node-level bitset filter cache introduced in #21179 defers cleanup of entries from closed readers to a periodic sweep over `Cache#keys()`. That sweep is unreliable: the live `keys()` iterator walks the LRU linked list without holding the LRU lock, and iteration under concurrent mutation is undefined. Every cache hit relinks the accessed entry to the head of the LRU list, so an entry promoted mid-sweep can end up behind the cursor and never be visited. Because `purgeStaleEntries()` unconditionally drains `staleCacheKeys` after each pass and a reader closes only once, a skipped entry is never revisited: it stays in the cache until size-based eviction reaches it — potentially forever — pinning its bitset, its query, and the `Value.listener → IndexService` graph on heap while being invisible to shard bitset stats.

This applies the same fix as the fielddata cache in #22491: invalidate the exact keys synchronously when the reader closes, instead of marking the reader stale and relying on a sweep.

- `getAndLoadIfNotPresent()` records every key cached for a reader in a per-reader registry (`keysByReader`); the closed listener is registered at most once per reader via the registry's atomic `computeIfAbsent`. Keys are only unregistered on reader close: trimming on eviction would race with a concurrent reload of the same key re-registering it, and a lost registration would leave the entry with no cleanup path. A registered key whose entry was evicted costs two references, is bounded by the reader's lifetime, and invalidating it on close is a no-op.
- `onClose()` invalidates exactly the keys registered for the closed reader, synchronously. A reader closes only once and exact-key invalidation is O(#queries cached for the reader), so there is no need to batch it.
- With no sweep left, the periodic `BitsetCacheCleaner`, the `indices.cache.bitset.cleanup_interval` setting, `purgeStaleEntries()`, and the stale-mark bookkeeping are removed, and the constructor no longer needs a `ThreadPool`.
- New regression tests cover exact per-reader invalidation across multiple cached queries and the eviction-then-reload case that motivates never unregistering keys on eviction; existing purge-based tests now assert entries are gone immediately after reader close.

### Related Issues

Related to #21179. Companion to #22491, which made the same change for the fielddata cache. No longer depends on #22499 (`Cache#keysSnapshot()`): with exact-key invalidation there is no sweep left in this cache.

### Check List


jwils:
@sgup432 please review this as well.

jwils:
@andrross I believe this is another concerning case that can cause permeant misses (until the cache is full) if hit. Since these also link to unaccounted for objects like Query, ValueWrapper, etc  I think it may also be able to grow larger than the max cache size.

kkhatua:
Thanks for the PR,  @jwils 

We recently did some minor improvements ( #21179  ) to contain this cache size limit, but a rewrite of the cache (like field data cache) is probably what is needed but that's not going to be trivial. 

We'll need some time to review this PR as there are competing priorities at the moment.

jwils:
@kkhatua our clusters are not on 3.7 yet so I can't say for sure of how large the impact is, but https://github.com/opensearch-project/OpenSearch/pull/21179 introduced a bug that this attempts to fix. I am concerned the cache size is no longer bounded without either a fix like this or reverting https://github.com/opensearch-project/OpenSearch/pull/21179

jwils:
Similar to the other pr. I believe removing the async `BitsetCacheCleaner` flow is safe, but I think the fix is important either way, so more than happy to refactor this to still retain the `BitsetCacheCleaner` if that makes it easier to review.

server/src/main/java/org/opensearch/indices/IndicesBitsetFilterCache.java
```diff
@@ -158,19 +152,26 @@ BitSet getAndLoadIfNotPresent(final Query query, final LeafReaderContext context
...
+        // Invalidate the exact keys synchronously rather than deferring to a periodic cache
+        // sweep: a reader closes only once, and exact-key invalidation is O(#queries cached for
+        // the reader), so there is no need to batch it. A key whose entry was already evicted is
+        // a no-op to invalidate.
+        Set<BitsetCacheKey> keys = keysByReader.remove(ownerCoreCacheKey);
+        if (keys != null) {
+            for (BitsetCacheKey key : keys) {
+                cache.invalidate(key);
+            }
+        }
```
sgup432, line 164: Should we instead do this in a different thread rather than doing in a sync manner? As I worry that this method might get triggered on a cluster applier thread while trying to close a shard(and eventually its readers), and might unnecessarily block it if it takes a lot of time (like we have seen in the past) and causes node drops as a result.

jwils: My preference would be to do it on one thread. I think it simplifies the logic where it doesn't need to be async. I don't think this can block for very long as these invalidate operations should be O(1). I think the max number of keys here is bounded by the number of nested fields so I can't imagine this list to invalidate being large enough to cause any issues.

I don't know the history of unnecessary blocks causing cluster blocks, so let me know if you still prefer this async and I'll make the change.

sgup432: >I think the max number of keys here is bounded by the number of nested fields so I can't imagine this list to invalidate being large enough to cause any issues.

Oh yeah! I missed the fact that for this cache, we won't have much queries cached per segment reader as it is bounded by the setting `index.mapping.nested_fields.limit` which is 50. Someone can set higher ofcourse to a very large value as there is no max defined it seems, but that's not a good practice and will break something else. So I agree that we can keep this as a sync operation.

sgup432 (approved):

sgup432:
@andrross @jainankitk Can you folks also review/approve and help in merging it in?

jainankitk (approved):

